PR #38, opened 2017-06-22
up the minimum required boost level to 1.64 to fix problem with strin…

pmesnier:
…g::data() only returning a const char *
This is a result of a discussion I had with Dan on Telegram. The money quote:
 Daniel Larimer, [Jun 21, 2017, 9:47 PM]:
actually

it requires boost 1.64

they added a non-const data() method

we should update cmake lists to require that


nathanhourt:
Is all the network code supposed to be in this pull req? I'm just adding 7b3584d3e978adb3f2c1bee534d0b641d07a6b82 for now...

PR #41, opened 2017-06-25
Make necessary changes to make it compile with g++

danielfrndz:
Change code to avoid g++ compiler bug: 
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67274
More info here: https://stackoverflow.com/questions/32097759/calling-this-member-function-from-generic-lambda-clang-vs-gcc


bytemaster:
What version of g++ are you using?

danielfrndz:
g++ version 5.4.0

PR #44, opened 2017-06-25
Net_plugin first feature

pmesnier:
With this PR a couple of eosd nods can connect, and  do an initial sync up. There's lots more to do but I wanted to get this out.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -255,15 +326,151 @@ class net_plugin_impl {
...
+    chain_controller& cc = chain_plug->chain();
+    uint32_t head = cc.head_block_num ();
+    if ( msg.last_irreversible_block_num > head) {
+      uint32_t delta = msg.last_irreversible_block_num - head;
+      uint32_t count = connections.size();
+      uint32_t span = delta / count;
+      uint32_t lastSpan = delta - (span * (count-1));
+      ilog ("peer is ahead of head by ${d}, count = ${c}, span = ${s}, lastspan = ${ls} ",
+            ("d",delta)("c",count)("s",span)("ls",lastSpan));
+      for (auto &cx: connections) {
```
pmesnier: I was following the guideline you wrote in protocols.hpp:
    if my head block < the LIB of a peer and my head block age > block interval * round_size/2 then
    enter sync mode...
        divide the block numbers you need to fetch among peers and send fetch request
        if peer does not respond to request in a timely manner then make request to another peer
        ensure that there is a constant queue of requests in flight and everytime a request is filled
        send of another request.

PR #46, opened 2017-06-26
Docker compile and docker-compose  start eos

peterwillcn:
Easy to compile and run with docker-compose.


educatedwarrior:
@peterwillcn ... thanks for building the docker image but why not use a dockerfile to build the base image.   No one knows what is in the base image.   This is dangerous from a security perspective.

peterwillcn:
@educatedwarrior Base image is only a compile environment in clang that reduces the build time.

PR #47, opened 2017-06-26
Fix dereference of DataStream struct

elmato:
Fix dereference of DataStream struct and use memoryArrayPtr for strings with length


bytemaster:
Thanks for this update :)

PR #61, opened 2017-07-03
Incremental p2p networking

pmesnier:
Adding in broadcast of newly generated blocks. There is some problem with the forked_database base interaction that causes an unhandled exception when receiving an out-of-order block.


bytemaster:
Blocks should never be sent "out of order" it is the responsibility of the sender to ensure that the receiver has the prior block. Either by knowing that it has already sent the prior block or receiving notice from the receiver that they already know about it.

PR #67, opened 2017-07-06
Limiting WASM exectuion time

brianjohnson5972:
by injecting checktime call into wasm function calls, loops, and branch.  Currently only loops have tests.


libraries/chain/include/eos/chain/wasm_interface.hpp
```diff
@@ -6,6 +6,10 @@
 
 namespace eos { namespace chain {
 
+struct checktime_exceeded : public fc::exception
```
bytemaster: We should define exceptions in eos/chain/exceptions.hpp

libraries/chain/wasm_interface.cpp
```diff
@@ -17,6 +18,22 @@ namespace eos { namespace chain {
    wasm_interface::wasm_interface() {
    }
 
+   std::chrono::time_point<std::chrono::system_clock> checktimeStart;
```
bytemaster: This global needs to be moved into the wasm_interface member data

libraries/chain/wasm_interface.cpp
```diff
@@ -245,6 +262,11 @@ DEFINE_INTRINSIC_FUNCTION1(env,toUpper,toUpper,none,i32,charptr) {
 //   return 0;
 }
 
+DEFINE_INTRINSIC_FUNCTION1(env,loopControl,loopControl,i32,i32,i) {
+   usleep(50);
```
bytemaster: What is this all about?    We shouldn't be sleeping.  I think this entire intrinsic can be removed.

libraries/chain/wasm_interface.cpp
```diff
@@ -17,6 +18,22 @@ namespace eos { namespace chain {
...
+   std::chrono::time_point<std::chrono::system_clock> checktimeStart;
+
+#ifdef NDEBUG
+   const int CHECKTIME_LIMIT = 2000;
+#else
+   const int CHECKTIME_LIMIT = 12000;
+#endif
+
+DEFINE_INTRINSIC_FUNCTION0(env,checktime,checktime,none) {
+   auto dur = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - checktimeStart);
```
bytemaster: fc::time_point::now() is a shorter way to write this.

libraries/chain/wasm_interface.cpp
bytemaster: also it should be an fc::time_point

PR #79, opened 2017-07-13
Create special accounts

elmato:


libraries/chain/include/eos/chain/config.hpp
```diff
@@ -53,6 +61,7 @@ const static ShareType DefaultElectedPay = Asset(100).amount;
 const static ShareType DefaultRunnerUpPay = Asset(75).amount;
 const static ShareType DefaultMinEosBalance = Asset(100).amount;
 const static UInt32 DefaultMaxTrxLifetime = 60*60;
+const static double ProducersAuthorityThreshold = 2.0/3.0;
```
bytemaster: we should never use double's in our code, especially any logic that is run as part of consensus.

libraries/chain/chain_controller.cpp
```diff
@@ -717,6 +718,16 @@ void chain_controller::update_global_properties(const signed_block& b) {
          gpo.active_producers = std::move(schedule);
          gpo.configuration = std::move(config);
       });
+
+      auto active_producers_authority = types::Authority(uint32_t(gpo.active_producers.size()*config::ProducersAuthorityThreshold), {}, {});
```
bytemaster: This like will break once you replace the double constant with a constant number of producers threshold.

The active producers size should always be 21.

tests/tests/special_accounts_tests.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+      chain.produce_blocks(config::BlocksPerRound - chain.head_block_num() );
+
+      auto& gpo = chain_db.get<global_property_object>();
+
+      BOOST_REQUIRE(boost::find(gpo.active_producers, "newproducer1") != gpo.active_producers.end());
+      BOOST_REQUIRE(boost::find(gpo.active_producers, "newproducer2") != gpo.active_producers.end());
+      BOOST_REQUIRE(boost::find(gpo.active_producers, "newproducer3") != gpo.active_producers.end());
+
+      auto threshold = uint32_t(gpo.active_producers.size()*config::ProducersAuthorityThreshold);
```
bytemaster: This line will also break when constant changed.

PR #98, opened 2017-07-26
Contract test api

elmato:


CMakeModules/wasm.cmake
```diff
@@ -39,7 +39,7 @@ macro(add_wast_target target SOURCE_FILES INCLUDE_FOLDERS DESTINATION_FOLDER)
 
     add_custom_command(OUTPUT ${outfile}.bc
       DEPENDS ${infile}
-      COMMAND ${WASM_CLANG} -emit-llvm -O3 --std=c++14 --target=wasm32 -I ${INCLUDE_FOLDERS} -fno-threadsafe-statics -fno-rtti -fno-exceptions -c ${infile} -o ${outfile}.bc
+      COMMAND ${WASM_CLANG} -emit-llvm -O3 --std=c++14 --target=wasm32 -fno-builtin -ffreestanding -nostdlib -I ${INCLUDE_FOLDERS} -fno-threadsafe-statics -fno-rtti -fno-exceptions -c ${infile} -o ${outfile}.bc
```
bytemaster: Why is this necessary?  Could we document this wasm.cmake file to explain why each and every argument is listed?

PR #100, opened 2017-07-27
Build on Travis CI

mitza-oci:
Travis build doesn't have WASM support yet, but builds all other code and runs tests/chain_test (currently failing 4 tests).
Results from running on my fork: https://travis-ci.org/mitza-oci/eos
Once this is merged the eos repository owners can enable Travis integration for master and for pull requests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 *.a
 *.sw*
-*.cmake
 *.dylib
 *.ll
 *.bc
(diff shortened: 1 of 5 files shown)


mitza-oci:
WASM support now implemented on Travis CI

mitza-oci:
Adding tests/slow_test caused "The log length has exceeded the limit of 4 MB".  Is there a chance that the very verbose logging (see below) will be turned off or should we drop slow_test from Travis CI?

> chain_controller.cpp:571      process_message

Issue #97, opened 2017-07-25
Edits needed on eosc.dox 


clayop:
1. `eosc create key` should be done twice to show both owner key and active key.
Current
~~~
./eosc create key
public: EOS4toFS3YXEQCkuuw1aqDLrtHim86Gz9u3hBdcBw5KNPZcursVHq
private: 5JKbLfCXgcafDQVwHMm3shHt6iRWgrr9adcmt6vX3FNjAEtJGaT
~~~
Suggested
~~~
./eosc create key
public: EOS4toFS3YXEQCkuuw1aqDLrtHim86Gz9u3hBdcBw5KNPZcursVHq
private: 5JKbLfCXgcafDQVwHMm3shHt6iRWgrr9adcmt6vX3FNjAEtJGaT
./eosc create key
public: EOS6KdkmwhPyc2wxN9SAFwo2PU2h74nWs7urN1uRduAwkcns2uXsa
private: 5WhichIsMatchedToTheExample
~~~

2. `eosc create account inita tester ...` does not work. Use `eos` instead of `inita` or place **transfer** instruction before **create account** since account `inita` is empty.
Suggestion 1
~~~
./eosc create account eos tester EOS4toFS3YXEQCkuuw1aqDLrtHim86Gz9u3hBdcBw5KNPZcursVHq EOS6KdkmwhPyc2wxN9SAFwo2PU2h74nWs7urN1uRduAwkcns2uXsa
~~~
Suggestion 2
~~~
./eosc transfer eos inita 1000
./eosc create account init tester EOS4toFS3YXEQCkuuw1aqDLrtHim86Gz9u3hBdcBw5KNPZcursVHq EOS6KdkmwhPyc2wxN9SAFwo2PU2h74nWs7urN1uRduAwkcns2uXsa
~~~

bytemaster:
Can you generate pull request?

clayop:
Sure I will do.

heifner:
eosc.dox has been updated with these changes and much more. Closing.

PR #108, opened 2017-07-31
Added Active Key Creation #97

clayop:


nathanhourt:
Is docs/group__eosc.html not autogenerated? It would be better to fix the source file that it is generated from, then regenerate the file. I think the file you're looking for is contracts/eoslib/eosc.dox

bytemaster:
Yes, it is auto-generated so we need to fix the source.

clayop:
Okay I will close and redo.

PR #110, opened 2017-07-31
fix return value.

philsong:


libraries/chain/message_handling_contexts.cpp
```diff
@@ -18,7 +18,7 @@ void message_validate_context::require_authorization(const types::AccountName& a
 
 void message_validate_context::require_scope(const types::AccountName& account)const {
    auto itr = boost::find_if(trx.scope, [&account](const auto& scope) {
-      return scope == account;
```
bytemaster: This change is returns an undefined value.

http://www.boost.org/doc/libs/1_45_0/libs/range/doc/html/range/reference/algorithms/non_mutating/find_if.html

philsong: as you defined, the require_scope function's return type is "void", but "return scope == account;" means return a bool value, so if some wrong happen, should raise a exception, otherwise should just "return;" instead of "return true/false", pls confirm again.thx.

Issue #90, opened 2017-07-20
Complete the ABI conversion of Binary/JSON


bytemaster:
The methods on AbiSerializer are barely functional for a few types necessary to work with `transfer` messages.  

https://github.com/EOSIO/eos/blob/master/libraries/types/AbiSerializer.cpp#L78

This is currently being tested here:

https://github.com/EOSIO/eos/blob/master/tests/tests/block_tests.cpp#L139

PR #118, opened 2017-08-01
complete abi conversion

elmato:
Ref #90


bytemaster:
This still lacks support for Array types (indicated as Type[]) and deserialized like vector<Type>().  I will merge it anyway, but #90 is not yet complete until it supports arrays and also the Bytes type.

Other built in types include uint256, PublicKey, Signature, etc.

See libraries/types/include/eos/types/native.hpp

Issue #128, opened 2017-08-02
Transaction API lookup by id


brianjohnson5972:
Adding Account History plugin to track transactions for future lookup in the blockchain

brianjohnson5972:
Provides lookup of transactions based on:
transaction id
account name -> all transactons that have account name in its scope

brianjohnson5972:
Working on adding lookup of account names in scope.

Issue is that multi_index only provides indexing on "flat" data (can use vector as key, but not entries in vector), so planning to have block_id, transaction_id, and account_name, which will mean duplicate id storage.

brianjohnson5972:
Merged #130

PR #130, opened 2017-08-02
Initial Transaction API

brianjohnson5972:
#128 
First commit.  Lookup of Transactions that have already been added to a block.


nathanhourt:
Hmm... This looks like a simple boost multi_index just sitting in RAM with no regard to blockchain history. That would be volatile storage, and it won't be resilient to forking.

Should this be stored in the database instead, so it's persistent and kept in sync through fork switches?

brianjohnson5972:
Yes, it is a simplistic implementation.  I was mostly wanting to get the plugins in there and then make the implementation actually correct.  Do you want me to fix that before committing or is it ok to have a naive implementation for now(with further work coming)?

The plugin also needs tracking based on scope/authorities which will really affect memory (I think, since multi_index doesn't support matching on contents in a key that is a vector), needs to add filtering AccountNames (to only track accounts you care about to limit size) and also thinking that the plugin will need to just copy the contents of the block and store it in the database on another thread to prevent blocking the chain production.

brianjohnson5972:
(Excuse my accidental closing of the pull request)

I'll make the change to using the database to hold the multi_index, at the time I was holding off all the complexities, but the complexities are not there yet.

bytemaster:
Yes, please implement with chainbase.

brianjohnson5972:
@nathanhourt - I presume that github sends you an email when something is added to the pull request, but just in case, I figured I would manually send this.

bytemaster:
@brianjohnson5972 there were conflicts between this and Phil's work, can you please resolve them so that I can merge?

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -0,0 +1,157 @@
...
+struct by_trx_id;
+using transaction_history_multi_index = chainbase::shared_multi_index_container<
+   transaction_history_object,
+   indexed_by<
+      ordered_unique<tag<by_id>, BOOST_MULTI_INDEX_MEMBER(transaction_history_object, transaction_history_object::id_type, id)>,
+      hashed_unique<tag<by_trx_id>, BOOST_MULTI_INDEX_MEMBER(transaction_history_object, transaction_id_type, transaction_id), std::hash<transaction_id_type>>
+   >
+>;
+
+typedef chainbase::generic_index<transaction_history_multi_index> transaction_history_index;
```
nathanhourt: Typically we keep the object and index definitions in a header file. Please move to a header for consistency

libraries/chain/include/eos/chain/types.hpp
```diff
@@ -171,6 +171,7 @@ namespace eos { namespace chain {
       transaction_object_type,
       producer_object_type,
       chain_property_object_type,
+      transaction_history_object_type,
```
nathanhourt: A comment here noting that this object is defined in the account_history_plugin would be good (like in the lines below). With modern IDEs, it's not as necessary, but it's still helpful to note which object types are declared here, but implemented in a separate library/plugin.

libraries/chain/include/eos/chain/types.hpp
```diff
@@ -171,7 +171,7 @@ namespace eos { namespace chain {
       transaction_object_type,
       producer_object_type,
       chain_property_object_type,
-      transaction_history_object_type,
+      transaction_history_object_type, ///< Defined by account history plugin library
```
nathanhourt: "Defined by account_history_plugin"

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -0,0 +1,136 @@
...
+                     return trx;
+      }
+
+      // ERROR in indexing logic
+      std::string msg = "transaction_id=" + transaction_id.str() + " indexed with block_id=" + block_id->str() + ", but ";
+      if (!block)
+         msg += "block was not found";
+      else
+         msg += "transaction was not found in the block";
+      BOOST_THROW_EXCEPTION( std::runtime_error( msg ) );
```
nathanhourt: Don't use a `std::runtime_error` -- we want to exclusively use our own exception types in this codebase.

This is throwing for a pretty weird case that, theoretically, should never actually occur, so we don't need to define an exception type for it specifically. We can just use assertions. Something like this should suffice:
```
FC_ASSERT(block, "Transaction with ID ${tid} was indexed as being in block ID ${bid}, but no such block was found", ("tid", transaction_id")("bid", block_id));
FC_THROW("Transaction with ID ${tid} was indexed as being in block ID ${bid}, but was not found in that block", ("tid", transaction_id")("bid", block_id));
```

libraries/chain/include/eos/chain/types.hpp
```diff
@@ -171,7 +171,7 @@ namespace eos { namespace chain {
       transaction_object_type,
       producer_object_type,
       chain_property_object_type,
-      transaction_history_object_type, ///< Defined by account history plugin library
+      transaction_history_object_type, ///< Defined by account_history_plugin library
```
nathanhourt: I apologize, I have been unclear. This comment claims the object is defined by a plugin library, but it's not a library, it's a plugin. The purpose of the comment is to make it easier to find the object without an IDE tracking references of `transaction_history_object_type`, so if it says "library" and "plugin", it's not obvious whether to look in the `plugins` or `libraries` directory. This comes from a plugin, so it should only say plugin and not library.

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -0,0 +1,132 @@
...
+private:
+
+   optional<block_id_type> find_block_id(const transaction_id_type& transaction_id) const;
+};
+
+optional<block_id_type> account_history_plugin_impl::find_block_id(const transaction_id_type& transaction_id) const
+{
+   const auto& db = chain_plug->chain().get_database();
+   optional<block_id_type> block_id;
+   db.with_read_lock( [&]() {
```
nathanhourt: Does not compile:
```
../../plugins/account_history_plugin/account_history_plugin.cpp:42:7: error: no matching member function for call to 'with_read_lock'
   db.with_read_lock( [&]() {
   ~~~^~~~~~~~~~~~~~
../../libraries/chainbase/include/chainbase/chainbase.hpp:881:15: note: candidate function not viable: 'this' argument has type 'const chainbase::database', but method is not marked const
         auto with_read_lock( Lambda&& callback, uint64_t wait_micro = 1000000 ) -> decltype( (*(Lambda*)nullptr)() )
              ^
```
db is a const ref, but `with_read_lock` is not a const method. I wonder... should `with_read_lock` be const?

brianjohnson5972: I believe I "fixed" that (Otherwise Travis would complain).  see commit d48ebabf56b4115753fcabb7648a0ffcf3b0f5e9.

libraries/chain/include/eos/chain/types.hpp
brianjohnson5972: No I apologize for not paying as close of attention to your original comment.

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: Were you able to get it to work, or is there an issue with my commit?  I had a heck of a time getting the submodule commit correct, so it would not be a big shocker if it really didn't work.

nathanhourt: Yep, just an old submodule. Oops :]

0x83043:
Hello, I caught the same error:
`
[ 97%] Building CXX object plugins/producer_plugin/CMakeFiles/producer_plugin.dir/producer_plugin.cpp.o
/root/eos/plugins/account_history_plugin/account_history_plugin.cpp:42:7: error: no matching member function for call to 'with_read_lock'
   db.with_read_lock( [&]() {
   ~~~^~~~~~~~~~~~~~
/root/eos/libraries/chainbase/include/chainbase/chainbase.hpp:881:15: note: candidate function not viable: 'this' argument has type 'const chainbase::database', but method is not marked const
         auto with_read_lock( Lambda&& callback, uint64_t wait_micro = 1000000 ) -> decltype( (*(Lambda*)nullptr)() )
              ^
/root/eos/plugins/account_history_plugin/account_history_plugin.cpp:71:2: warning: TODO: lookup of recent transactions [-W#warnings]
#warning TODO: lookup of recent transactions
 ^
1 warning and 1 error generated.
plugins/account_history_plugin/CMakeFiles/account_history_plugin.dir/build.make:62: recipe for target 'plugins/account_history_plugin/CMakeFiles/account_history_plugin.dir/account_history_plugin.cpp.o' failed
make[2]: *** [plugins/account_history_plugin/CMakeFiles/account_history_plugin.dir/account_history_plugin.cpp.o] Error 1
CMakeFiles/Makefile2:3396: recipe for target 'plugins/account_history_plugin/CMakeFiles/account_history_plugin.dir/all' failed
make[1]: *** [plugins/account_history_plugin/CMakeFiles/account_history_plugin.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
[ 97%] Linking CXX static library libproducer_plugin.a
[ 97%] Built target producer_plugin
[ 98%] Linking CXX static library libchain_api_plugin.a
[ 98%] Built target chain_api_plugin
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
`
How you fixed it?

brianjohnson5972:
That means that you need to run "git submodule update".  with_read_lock was changed to be const, but since it is a submodule it doesn't get updated when you do your git fetch or pull.

0x83043:
It's didn't update with "git submodule update". But the same result (error) after force update
```
root@4c5c874888cf:~/eos# git submodule update
root@4c5c874888cf:~/eos# git submodule update -f
Submodule path 'libraries/appbase': checked out '754b9d19783e008aa783a12465594f1fe2a49c29'
Submodule path 'libraries/binaryen': checked out '3f0db5a7aafaaa4de713ff3ba3c3bbeb59fe566e'
Submodule path 'libraries/chainbase': checked out 'f1a150f97bfde1b10113b77c22206e5c00158205'
```

brianjohnson5972:
Sorry, you are correct, the submodule got changed to an earlier revision.  We will get that fixed.

0x83043:
Ok. Please let me know, when I can try more time.

brianjohnson5972:
What is on master now should build.

0x83043:
It's ok. Thanx )

PR #133, opened 2017-08-03
assign tag to image in dockerfile

tooooolong:
For users who pulled ubuntu image long ago, it will not update this image version.

And the commands in this dockerfile seems like only works in 16.04 or above.

By assign a tag to the docker image, every one will use the same version of ubuntu.


tooooolong:
111

Issue #112, opened 2017-07-31
Merge Message Contexts


bytemaster:
Currently we have apply, precondition, and validate contexts but they are now irrelevant and can be merged into one simple apply_context.

bytemaster:
https://github.com/EOSIO/eos/blob/master/libraries/chain/include/eos/chain/message_handling_contexts.hpp

nathanhourt:
Resolved by #134

PR #134, opened 2017-08-03
merged apply, precondition and validate contexts into apply_context

iamaprotoss:


nathanhourt:
This change resolves #112

PR #151, opened 2017-08-09
Progress on #7, upgrades to testing framework

nathanhourt:
We now check that transactions don't have any unnecessary signatures.

Upgrade AuthorityChecker to support tracking which keys were used. This makes it possible to check for irrelevant sigs, and also makes it possible to automatically sign transactions by taking a list of available keys and determining which ones are necessary to authorize the transaction.

Upgrade testing framework to support automatically signing transactions (not yet working as of 7eb2013)

### DIFF
--- a/libraries/chain/include/eos/chain/authority.hpp
+++ b/libraries/chain/include/eos/chain/authority.hpp
@@ -30,55 +30,8 @@ struct shared_authority {
    shared_vector<types::KeyPermissionWeight>     keys;
 };
 
-/**
- * @brief This class determines whether a set of signing keys are sufficient to satisfy an authority or not
- *
- * To determine whether an authority is satisfied or not, we first determine which keys have approved of a message, and
- * then determine whether that list of keys is sufficient to satisfy the authority. This class takes a list of keys and
- * provides the @ref satisfied method to determine whether that list of keys satisfies a provided authority.
- *
- * @tparam F A callable which takes a single argument of type @ref AccountPermission and returns the corresponding
- * authority
- */
-template<typename F>
-class AuthorityChecker {
-   F PermissionToAuthority;
-   flat_set<public_key_type> signingKeys;
-
-public:
-   AuthorityChecker(F PermissionToAuthority, const flat_set<public_key_type>& signingKeys)
-      : PermissionToAuthority(PermissionToAuthority), signingKeys(signingKeys) {}
-
-   bool satisfied(const types::AccountPermission& permission) const {
-      return satisfied(PermissionToAuthority(permission));
-   }
-   template<typename AuthorityType>
-   bool satisfied(const AuthorityType& authority) const {
-      UInt32 weight = 0;
-      for (const auto& kpw : authority.keys) {
-         if (signingKeys.count(kpw.key)) {
-            weight += kpw.weight;
-            if (weight >= authority.threshold)
-               return true;
-         }
-      }
-      for (const auto& apw : authority.accounts)
-//#warning TODO: Recursion limit? Yes: implement as producer-configurable parameter 
-         if (satisfied(apw.permission)) {
-            weight += apw.weight;
-            if (weight >= authority.threshold)
-               return true;
-         }
-      return false;
-   }
-};
-
-inline bool operator < ( const types::AccountPermission& a, const types::AccountPermission& b ) {
-   return std::tie( a.account, a.permission ) < std::tie( b.account, b.permission );
-}
-template<typename F>
-AuthorityChecker<F> MakeAuthorityChecker(F&& pta, const flat_set<public_key_type>& signingKeys) {
-   return AuthorityChecker<F>(std::forward<F>(pta), signingKeys);
+inline bool operator< (const types::AccountPermission& a, const types::AccountPermission& b) {
+   return std::tie(a.account, a.permission) < std::tie(b.account, b.permission);
 }
 
 /**
--- a/libraries/chain/include/eos/chain/authority.hpp
+++ b/libraries/chain/include/eos/chain/authority.hpp
@@ -38,7 +38,7 @@ inline bool operator< (const types::AccountPermission& a, const types::AccountPe
  * Makes sure all keys are unique and sorted and all account permissions are unique and sorted and that authority can
  * be satisfied
  */
-inline bool validate( types::Authority& auth ) {
+inline bool validate(const types::Authority& auth) {
    const types::KeyPermissionWeight* prev = nullptr;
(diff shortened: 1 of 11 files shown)


libraries/chain/include/eos/chain/authority_checker.hpp
```diff
@@ -0,0 +1,127 @@
...
+ * @tparam F A callable which takes a single argument of type @ref AccountPermission and returns the corresponding
+ * authority
+ */
+template<typename F>
+class AuthorityChecker {
+   F PermissionToAuthority;
+   vector<public_key_type> signingKeys;
+   vector<bool> usedKeys;
+
+   struct WeightTallyVisitor {
```
wanderingbort: Minor style note: I lean towards stateless visitors when I can in order to improve the colocation of code and intent.  Below in `AuthorityChecker::satisfied` we loop over all the `permissions` mutating the internal state of a visitor.  The only hint that this is happening in the context of that method's scope is the use of Tally in the visitor's type. 

I am not going to ding a review for it but, I would consider making the visitor simply produce the weight and then putting the aggregation of weights in the `::satisfied` method where the result is important/used.

wanderingbort: Hrm, on second blush there are a few hidden side effects in this visitor, like `usedKeys` mutation.  So, making it "pure" from an aggregation point of view is rather hollow.

libraries/chain/include/eos/chain/authority_checker.hpp
```diff
@@ -0,0 +1,127 @@
...
+      permissions.insert(authority.keys.begin(), authority.keys.end());
+      permissions.insert(authority.accounts.begin(), authority.accounts.end());
+
+      // Check all permissions, from highest weight to lowest, seeing if signingKeys satisfies them or not
+      WeightTallyVisitor visitor(*this);
+      for (const auto& permission : permissions)
+         // If we've got enough weight, to satisfy the authority, return!
+         if (permission.visit(visitor) >= authority.threshold) {
+            KeyReverter.cancel();
+            return true;
```
wanderingbort: are we concerned that this early out may lead to problems when permissions are over-authorized?
For instance, if a user were to have 2 messages that required (2 of 3: A, B, C) and (1 of 3: A, B)  and had signatures from A, B and C.  Depending on the order of the checks this will either throw for unecessary sigs (A,B pass, A pass) or not (B, C pass, A pass).

libraries/chain/include/eos/chain/authority_checker.hpp
nathanhourt: Yes, the whole AuthorityChecker class got pretty complex, but what it's doing is also pretty bloody complex... I didn't come up with a simpler approach, but perhaps one exists. I'm open to suggestions. In the meantime, I designed it to hide all the details inside.

libraries/chain/include/eos/chain/authority_checker.hpp
nathanhourt: I don't see what affect breaking early has on that... but we definitely want to break early, some of these authorities can be quite expensive to check (many recursions down an account multisig tree, for instance) and we don't want to do that if we already know the answer.

As to the case where two messages might let the user slip some extra sigs in... Can they, though? Authorities should be sorted, so we should always check the keys within them in the same order... That should prevent the situation you described, yes?

libraries/chain/include/eos/chain/authority_checker.hpp
nathanhourt: > I would consider making the visitor simply produce the weight and then putting the aggregation of weights in the ::satisfied method where the result is important/used.

Not sure I understand what you mean here. Do you mean not taking the early out you mentioned below?

libraries/chain/include/eos/chain/authority.hpp
```diff
@@ -30,62 +30,15 @@ struct shared_authority {
...
-      UInt32 weight = 0;
-      for (const auto& kpw : authority.keys) {
-         if (signingKeys.count(kpw.key)) {
-            weight += kpw.weight;
-            if (weight >= authority.threshold)
-               return true;
-         }
-      }
-      for (const auto& apw : authority.accounts)
-//#warning TODO: Recursion limit? Yes: implement as producer-configurable parameter
```
bytemaster: From now on all such TODO's need to reference a specific github issue number and should be tagged as "bug" and assigned to a final release milestone (June 2018) or earlier. We should also not flood our builds with warnings like this, especially in header files.

libraries/chain/include/eos/chain/authority_checker.hpp
nathanhourt: Ohh, you mean add up the weights in `satisfied`... Seems like 6 of one, half dozen of the other to me, but sure, I can change it.

libraries/chain/include/eos/chain/authority_checker.hpp
wanderingbort: the sorting certainly means that for any transaction this always goes the same way.  My concern is the burden on wallets to match this algorithm in their signature collection and pruning.  There should be a usable set of rules for this pruning but it seems excessive.  

A wallet must prune any signatures made irrelevant by this evaluation algorithm or it risks being rejected.  Likely this means it will collect signatures and re-run this algorithm with the early out to determine which of the obtained signatures will be marked as "unused" in the case of over authorization. 

I understand that over-authorization has costs but the rules for a properly formed transaction are made more complex by this early out.  Without it, we may exceed the authority threshold but we will mark all signatures that *could* be part of a message authorization as used.  Relieving the burden of wallets but putting potentially more stress on the chain/processing.

nathanhourt: I guess the ordering requirement only means that we have consistent behavior across messages, not that we would detect that oversigning issue you described. I suppose to address that, I can make `WeightTallyVisitor` check if keys that are already used would satisfy the authority first, and only then check unused keys/marking them used if they help.

bytemaster: I think we can make the over-authorization a "soft" consensus rather than a "hard" consensus.  In other words, block producers may choose to include over-authorized transactions, but in practice will reject them. This will allow us to update / refine this without hard forking the chain.

nathanhourt: Then again, I can't figure out a way to make `WeightTallyVisitor` mark a key as used when it didn't need to be used. Since it always checks keys in the same order, it always prefers using already-used keys to unused ones without me explicitly checking for it.

PR #156, opened 2017-08-10
Resolve #66, more auth tests

### LINKED ISSUE #66: Authority Check Recursion Depth Limit Check
AuthorityChecker::satisfied() has a #warning TODO that is spamming output because it is in header.  Moving that TODO to a this issue.

nathanhourt:


nathanhourt:
The thing I'm least sure of here is how to handle recursion depth violations. Right now I'm just silently failing the authority, which works fine as long as the user wasn't relying on the deep recursing tree to authorize his transaction. If he was, however, it fails without any explanation as to why.

If we want to give some feedback as to why authorization failed due to overly deep recursion, I could set a flag when the check trips, and then mention that in the exception that results if the authorization ultimately does fail, but does it matter? Ultimately the wallet should have caught this before the transaction was signed anyways.

PR #159, opened 2017-08-10
Resolve #152: Disable auth updates in pending transactions

### LINKED ISSUE #152: Don't update authorities for pending transactions
All authority checks are done at the beginning of block processing, prior to evaluating any transactions. Stated another way, updates to authorities do not take effect until the end of the block that updated them. This means that pending transactions, which are not yet in a block, should not store their authority updates in the database. These updates will only hit the database when the pending transaction is included in a block and processed while applying that block.

nathanhourt:
Affected contract methods:
 - newaccount
 - updateauth
 - deleteauth
 - linkauth
 - unlinkauth

Also, blockchain initialization (the genesis block) behaves like it is
applying a block, even though the genesis block isn't really a block.


nathanhourt:
Yeah, I think that's best. I'll add that in.

Issue #161, opened 2017-08-11
Add slow_test to Travis build


brianjohnson5972:
Need to fix failing tests tapos_wrap and staked and then can turn on Travis build.  Also need to fix create_exchange test.

Need to fix api_test and run for Travis build.

PR #163, opened 2017-08-11
Fix slow_test stake test

brianjohnson5972:
Fixed multiple errors preventing slow_test from passing and added slow_test to Travis build.
#161

### DIFF
--- a/tests/api_tests/api_tests.cpp
+++ b/tests/api_tests/api_tests.cpp
@@ -224,7 +224,7 @@ BOOST_FIXTURE_TEST_CASE(test_all, testing_fixture)
       //Set test code
       types::setcode handler;
       handler.code.resize(wasm.size());
-      memcpy( handler.code.data(), wasm.data(), wasm.size() );
+      handler.code.assign(wasm.begin(), wasm.end());
 
       send_set_code_message(chain, handler, "testapi");
       send_set_code_message(chain, handler, "acc1");
--- a/tests/api_tests/api_tests.cpp
+++ b/tests/api_tests/api_tests.cpp
@@ -99,7 +99,7 @@ void SetCode( testing_blockchain& chain, AccountName account, const char* wast )
       }
 } FC_LOG_AND_RETHROW( ) }
 
-uint32_t CallFunction( testing_blockchain& chain, const types::Message& msg, const vector<char>& data, const vector<AccountName>& scope = {N(test_api)}) {
+uint32_t CallFunction( testing_blockchain& chain, const types::Message& msg, const vector<char>& data, const vector<AccountName>& scope = {N(testapi)}) {
    static int expiration = 1;
    eos::chain::SignedTransaction trx;
    trx.scope = scope;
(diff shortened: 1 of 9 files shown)


tests/common/macro_support.hpp
```diff
@@ -146,8 +146,8 @@ inline std::vector<Name> sort_names( std::vector<Name>&& names ) {
 #define STAKE4(chain, sender, recipient, amount) \
    { \
       eos::chain::SignedTransaction trx; \
-      trx.scope = sort_names( { #sender, #recipient } ); \
-      trx.emplaceMessage(config::EosContractName, vector<types::AccountPermission>{{#sender, "active"}}, \
+      trx.scope = sort_names( { #sender, #recipient, "eos" } ); \
```
nathanhourt: It's better to use `config::EosContractName` rather than the literal "eos". I'd like to minimize the number of places that name the contract directly.

nathanhourt: Same for lines 162 and 175... perhaps others

tests/common/macro_support.hpp
```diff
@@ -173,7 +173,7 @@ inline std::vector<Name> sort_names( std::vector<Name>&& names ) {
    { \
       eos::chain::SignedTransaction trx; \
       trx.scope = sort_names( { "eos", #account } ); \
-      trx.emplaceMessage(config::EosContractName, vector<types::AccountPermission>{{#account, "active"}}, \
+      trx.emplaceMessage(config::EosContractName, vector<types::AccountPermission>{}, \
```
nathanhourt: Also lines 150, 164... Why remove the declared authorizations? Doesn't this make tests break? If not, something is wrong.

brianjohnson5972: I removed it because it indicates irrelevant authority included and fails the test.  I thought it was odd, but figured  the question would be answered in this review.  So how do we want to handle this?  Keep this code and write a separate Issue, or should I expand this to fix the check to correctly identify the correct authorities?

brianjohnson5972:
I don't know if it makes it obvious, but also notice I added fixes for test_api.

tests/common/macro_support.hpp
brianjohnson5972: So what is the verdict on this?

tests/api_tests/api_tests.cpp
```diff
@@ -211,9 +222,13 @@ BOOST_FIXTURE_TEST_CASE(test_all, testing_fixture)
       chain.produce_blocks(1);
 
       //Set test code
-      SetCode(chain, "test_api", test_api_wast_str.c_str());
-      SetCode(chain, "acc1", test_api_wast_str.c_str());
-      SetCode(chain, "acc2", test_api_wast_str.c_str());
+      types::setcode handler;
+      handler.code.resize(wasm.size());
+      memcpy( handler.code.data(), wasm.data(), wasm.size() );
```
nathanhourt: Why memcpy if you don't have to? `handler.code.assign(wasm.begin(), wasm.end());` does the resize and assign without worrying about buffer boundaries. :)

brianjohnson5972: Was just copying code from create_script, so I'll clean this up.

brianjohnson5972:
@nathanhourt - Not sure if I did the right thing in eos_contract.cpp for apply_eos_setcode's setup of new apply_context for call to wasm_interface::get().init, so please comment on that.

nathanhourt:
Looks good to me. Travis is failing because something in api_tests is exceeding a time limit, but it seems to pass fine on my computer. Do we want to do something to fix that?... CI is much less useful if it fails spuriously.

brianjohnson5972:
It passed Travis now.

PR #164, opened 2017-08-13
Fix template specialization issue

### LINKED ISSUE #140: error when  'make'
I use ubuntu mate,but got the following error when doing make :
```
libraries/chain/CMakeFiles/eos_chain.dir/build.make:62: recipe for target 'libraries/chain/CMakeFiles/eos_chain.dir/chain_controller.cpp.o' failed
make[2]: *** [libraries/chain/CMakeFiles/eos_chain.dir/chain_controller.cpp.o] Error 1
CMakeFiles/Makefile2:1082: recipe for target 'libraries/chain/CMakeFiles/eos_chain.dir/all' failed
make[1]: *** [libraries/chain/CMakeFiles/eos_chain.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```
How can I fix it ?

sim31:
According to C++14 template specialization has to be more specialized than
the primary template. get_typename defined in static_variant.hpp was less
specialized than the primary template in typename.hpp.
Because of that build was failing on GCC 7.1.1. Fixes #140

### DIFF
--- a/libraries/fc/include/fc/reflect/typename.hpp
+++ b/libraries/fc/include/fc/reflect/typename.hpp
@@ -15,7 +15,7 @@ namespace fc {
   class exception;
   namespace ip { class address; }
 
-  template<typename T> class get_typename{};
+  template<typename... T> struct get_typename;
   template<> struct get_typename<int32_t>  { static const char* name()  { return "int32_t";  } };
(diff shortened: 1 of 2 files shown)


Apercu:
Might be related to #140

Apercu:
Confirming that this works, thanks!

PR #177, opened 2017-08-15
Fix for issue 173, --skip-transaction-signatures to eosd

heifner:


contracts/eoslib/eosc.dox
```diff
@@ -6,7 +6,12 @@
   @section intro Introduction to EOSC
   
   `eosc` is a command line tool that interfaces with the REST api exposed by @ref eosd. In order to use `eosc` you will need to
-  have a local copy of `eosd` running and configured to load the 'eos::chain_api_plugin'.  
+  have a local copy of `eosd` running and configured to load the 'eos::chain_api_plugin'. Currently, `eosc` does not
```
bytemaster: This documentation misrepresents what is going on.  It has nothing to do with eosc support for signature checking and everything to do with making it easy for developers to test functions without keys.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -107,6 +110,20 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
       app().get_plugin<database_plugin>().wipe_database();
       fc::remove_all(my->block_log_dir);
    }
+   if (options.at("skip-transaction-signatures").as<bool>()) {
+      ilog("Setting skip_transaction_signatures");
+      elog("Setting skip_transaction_signatures\n"
+           "\n"
+           "***************************************\n"
+           "*                                     *\n"
+           "*   -- CHAIN IGNORING SIGNATURES --   *\n"
```
bytemaster: CHAIN isn't ignoring it, this EOSD is ignoring it.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -166,7 +187,10 @@ bool chain_plugin::accept_block(const chain::signed_block& block, bool currently
 }
 
 void chain_plugin::accept_transaction(const chain::SignedTransaction& trx) {
-   chain().push_transaction(trx);
+   if (my->skip_flags & chain_controller::skip_transaction_signatures) {
```
bytemaster: This might be overly verbose for flood testing.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -301,7 +329,10 @@ read_write::push_block_results read_write::push_block(const read_write::push_blo
 }
 
 read_write::push_transaction_results read_write::push_transaction(const read_write::push_transaction_params& params) {
-   auto ptrx = db.push_transaction(params);
+   if (skip_flags & chain_controller::skip_transaction_signatures) {
```
bytemaster: rather than duplicating the bitwise logic, this should call chain_plugin::skip_transaction_signatures.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -177,6 +201,10 @@ bool chain_plugin::block_is_on_preferred_chain(const chain::block_id_type& block
    return chain().get_block_id_for_num(chain::block_header::num_from_id(block_id)) == block_id;
 }
 
+bool chain_plugin::skip_transaction_signatures() const {
```
bytemaster: This function name looks like an "action", eg start skipping transaction signatures, rather than the answer to a question, "is_skipping_trx_sigs" or "should_skip_sigs".

heifner: is_skip_transaction_signatures() since all it does is check the flag?

nathanhourt: `is_skipping_transaction_signatures()`

pmesnier:
What about adding a compiled-in "test" mode that could allow commands to set the various skip- flags to be issued through the eosc? These commands would propagate as transactions through all networked nodes. Networked nodes would check for this capacity in peers upon startup and would only connect to similarly build nodes. 

This way we could allow for such things as running tests across many nodes using varied levels of validation but without worrying about stopping and restarting all the connected nodes.

heifner:
Ultimately you want all verifications in place I would think. For example, once the command line wallet is up and running it can be used in the p2p tests to provide the signatures. That way you are testing signature verification and performance along with everything else.

Issue #144, opened 2017-08-08
Implement floating point API functions


elmato:
- double_add
- double_sub
- double_mult
- double_div
- double_lt
- double_eq
- double_gt
- double_to_i64
- i64_to_double

API should use the boost multi-precision implementation

nathanhourt:
So we're exposing a deterministic double type to the VM? Cool :)

PR #183, opened 2017-08-16
Implement floating point API

elmato:
Ref #144

### DIFF
--- a/libraries/chain/wasm_interface.cpp
+++ b/libraries/chain/wasm_interface.cpp
@@ -1,4 +1,5 @@
 #include <boost/function.hpp>
+#include <boost/multiprecision/cpp_dec_float.hpp>
 #include <eos/chain/wasm_interface.hpp>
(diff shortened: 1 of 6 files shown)


libraries/chain/wasm_interface.cpp
```diff
@@ -16,6 +17,7 @@
 namespace eos { namespace chain {
    using namespace IR;
    using namespace Runtime;
+   typedef boost::multiprecision::cpp_dec_float_50 DOUBLE;
```
bytemaster: The actual precision of a cpp_dec_float is always slightly higher than the number of digits specified in the template parameter, actually how much higher is an **implementation detail but is always at least 8 decimal** digits.

This indicates that it may not be deterministic. 

The documentation for cpp_bin_float_50 does not have such a disclaimer, that said we should verify that under the hood it only uses integer maths.

bytemaster:
I'm going to merge this now, but we will create an issue to have it fixed.

PR #192, opened 2017-08-16
Make irrelevant sigs a soft check

nathanhourt:
Previously, irrelevant signatures were always rejected. Now, they're always rejected unless we're processing a block which includes a transaction with irrelevant signatures.

Question: Should we instead allow irrelevant sigs for pending transactions as well as when applying a block, but only reject them when generating a block? If we do, then these transactions will move through the P2P network, but never get added to a block unless a producer ignores the soft check. It is likely that these transactions would circulate in nodes' pending state until they expire.

To prevent this, I opted to make pending transactions apply the soft checks as well.

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -318,8 +318,8 @@ signed_block chain_controller::_generate_block(
 
    uint64_t postponed_tx_count = 0;
    // pop pending state (reset to head block state)
-   for( const SignedTransaction& tx : _pending_transactions )
-   {
+   for (auto itr = _pending_transactions.begin(); itr != _pending_transactions.end(); ) {
+      auto& tx = *itr;
       size_t new_total_size = total_block_size + fc::raw::pack_size( tx );
 
       // postpone transaction if it would make block too big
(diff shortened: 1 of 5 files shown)


libraries/chain/chain_controller.cpp
```diff
@@ -519,7 +524,7 @@ void chain_controller::check_transaction_authorization(const SignedTransaction&
          }
       }
 
-   if ((_skip_flags & skip_transaction_signatures) == false)
+   if (!allow_unused_signatures && (_skip_flags & skip_transaction_signatures) == false)
```
bytemaster: We should never use _skip_flags and/or bit math in IF statements, instead we should define a function that returns true/false and call that function.  It will make the code more readable and also eliminate errors if we need the same check in multiple places.

libraries/chain/chain_controller.cpp
```diff
@@ -484,7 +489,7 @@ void chain_controller::_apply_block(const signed_block& next_block)
 
 } FC_CAPTURE_AND_RETHROW( (next_block.block_num()) )  }
 
-void chain_controller::check_transaction_authorization(const SignedTransaction& trx)const {
+void chain_controller::check_transaction_authorization(const SignedTransaction& trx, bool allow_unused_signatures)const {
    if ((_skip_flags & skip_transaction_signatures) && (_skip_flags & skip_authority_check)) {
```
bytemaster: these skip flags need to be moved into named methods that return true or false. Too many errors can occur by doing bit and/or boolean math in if statements.

PR #205, opened 2017-08-18
while exec the currency contract，error json led to parser error

OracleChain:
"from":"tester","to":"inita","amount":50 --> "from":"tester","to":"inita","amount":"50"

### DIFF
--- a/contracts/eoslib/eosc.dox
+++ b/contracts/eoslib/eosc.dox
@@ -254,7 +254,7 @@ The next step is to publish the contract (.wast) and its abi (.abi)
 After the contract is published it initially allocates all of the currency to the `currency` account.  So lets transfer some of it to our tester.
 
 ```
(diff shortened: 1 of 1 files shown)


OracleChain:
while exec the currency contract,  ./eosc exec currency transfer '{"from":"currency","to":"oraclechain","amount":50}' '["currency","oraclechain"]' '[{"account":"currency","permission":"active"}]', 

the params {"from":"tester","to":"inita","amount":50} will led to parse error:

10 assert_exception: Assert Exception
status_code == 200: Error
: 7 bad_cast_exception: Bad Cast
Invalid cast from type 'uint64_type' to Object
    {"type":"uint64_type"}
    thread-1  variant.cpp:565 get_string

{"from":"tester","to":"inita","amount":"50"} is the right params

elmato:
@OracleChain thanks for this PR, please see #210 that fixes the problem

PR #208, opened 2017-08-18
Transaction api find by account name

brianjohnson5972:
#208 
Added retrieving transactions based on an account name being in its scope.
Limit tracking to specific accounts via configuration.
Limit time spent collecting transactions (returns the transactions collected before time exceeded).


plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -99,10 +266,24 @@ account_history_plugin::~account_history_plugin()
 
 void account_history_plugin::set_program_options(options_description& cli, options_description& cfg)
 {
+   cfg.add_options()
+         ("filter_on_accounts,f", bpo::value<vector<string>>()->composing(),
+          "Track only transactions whose scopes involve the listed accounts. Default is to track all transactions.")
+         ("get-transactions-time-limit", bpo::value<int>()->default_value(account_history_plugin_impl::DEFAULT_TRANSACTION_TIME_LIMIT),
+          "Limits the time to wait before returning for get_transactions")
```
nathanhourt: The time to wait before returning? Or the maximum time to spend processing a single call?

nathanhourt: Also, what's the limit measured in? Milliseconds?

plugins/account_history_plugin/include/eos/account_history_plugin/account_history_object.hpp
```diff
@@ -13,15 +26,18 @@ class transaction_history_object : public chainbase::object<chain::transaction_h
...
 struct by_id;
 struct by_trx_id;
+struct by_account_name;
 using transaction_history_multi_index = chainbase::shared_multi_index_container<
    transaction_history_object,
    indexed_by<
       ordered_unique<tag<by_id>, BOOST_MULTI_INDEX_MEMBER(transaction_history_object, transaction_history_object::id_type, id)>,
-      hashed_unique<tag<by_trx_id>, BOOST_MULTI_INDEX_MEMBER(transaction_history_object, transaction_id_type, transaction_id), std::hash<transaction_id_type>>
+      hashed_non_unique<tag<by_trx_id>, BOOST_MULTI_INDEX_MEMBER(transaction_history_object, transaction_id_type, transaction_id), std::hash<transaction_id_type>>,
+      hashed_non_unique<tag<by_account_name>, BOOST_MULTI_INDEX_MEMBER(transaction_history_object, AccountName, account_name), std::hash<AccountName>>
```
nathanhourt: Fwiw, we avoid non unique indexes in core validation, because they're inconsistent across machines. This isn't core validation, it's just an API call, so it shouldn't matter so much here.

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: How about "Limits the maximum time (in milliseconds) processing a single get_transactions call."?Also changing code to take milliseconds (was micro).

plugins/account_history_plugin/include/eos/account_history_plugin/account_history_object.hpp
brianjohnson5972: Yeah, I had seen that on other indexes and evaluated that inconsistent ordering wouldn't matter in these cases.

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -48,45 +70,190 @@ optional<block_id_type> account_history_plugin_impl::find_block_id(const transac
...
-      FC_THROW("Transaction with ID ${tid} was indexed as being in block ID ${bid}, but was not found in that block", ("tid", transaction_id)("bid", block_id));
+      return find_transaction(transaction_id, *block_id);
    }
 
 #warning TODO: lookup of recent transactions
    FC_THROW_EXCEPTION(chain::unknown_transaction_exception,
                       "Could not find transaction for: ${id}", ("id", transaction_id.str()));
 }
 
+get_transactions_results account_history_plugin_impl::get_transactions(const AccountName&  account_name, const optional<uint32_t>& start_seq, const optional<uint32_t>& stop_seq) const
```
nathanhourt: Historically, we've implemented result pagination of results using a skip count (the number of records to skip returning) and a limit (the maximum number of records to return). That structure worked pretty well, I think, and I'm inclined to stick with it for consistency's sake, unless there's some reason not to.

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -270,13 +270,13 @@ void account_history_plugin::set_program_options(options_description& cli, optio
...
          ("get-transactions-time-limit", bpo::value<int>()->default_value(account_history_plugin_impl::DEFAULT_TRANSACTION_TIME_LIMIT),
-          "Limits the time to wait before returning for get_transactions")
+          "Limits the maximum time (in milliseconds) processing a single get_transactions call.")
          ;
 }
 
 void account_history_plugin::plugin_initialize(const variables_map& options)
 {
-   my->transactions_time_limit = options.at("get-transactions-time-limit").as<int>();
+   my->transactions_time_limit = options.at("get-transactions-time-limit").as<int>() * 1000;
```
nathanhourt: Docs say it's in milliseconds now, but the value is treated as seconds and multiplied by 1000? Just take the value in milliseconds directly and don't convert.

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -48,45 +70,190 @@ optional<block_id_type> account_history_plugin_impl::find_block_id(const transac
...
+   auto block_transaction = block_transaction_ids.cbegin();
+
+   uint32_t current = 0;
+   // keep iterating through each equal range
+   while (block_transaction != block_transaction_ids.cend() && !time_exceeded(start_time))
+   {
+      optional<signed_block> block;
+      db.with_read_lock( [&]() {
+         block = chain_plug->chain().fetch_block_by_id(block_transaction->first);
+      } );
```
nathanhourt: You don't need a read lock for this. You shouldn't need a read lock for any call you're making on `chain_controller` -- the controller should handle it's own locking. Locks should only be necessary when calling the database directly.

plugins/account_history_plugin/include/eos/account_history_plugin/account_history_object.hpp
```diff
@@ -13,15 +26,18 @@ class transaction_history_object : public chainbase::object<chain::transaction_h
    id_type             id;
    block_id_type       block_id;
    transaction_id_type transaction_id;
+   AccountName         account_name;
```
nathanhourt: I'm pretty sure normalizing this will greatly simplify your lookup logic as well.

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: It is taking milliseconds and is converting to microseconds, for the comparison to time_point.count()

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -24,16 +24,38 @@ using chain::signed_block;
...
+      {
+         return chain::block_header::num_from_id(a) > chain::block_header::num_from_id(b);
+      }
+   };
+   typedef std::multimap<block_id_type, transaction_id_type, block_comp> block_transaction_id_map;
 
    optional<block_id_type> find_block_id(const transaction_id_type& transaction_id) const;
+   ProcessedTransaction find_transaction(const chain::transaction_id_type&  transaction_id, const signed_block& block) const;
+   ProcessedTransaction find_transaction(const chain::transaction_id_type&  transaction_id, const block_id_type& block_id) const;
+   bool scope_relevant(const eos::types::Vector<AccountName>& scope);
```
brianjohnson5972: you are correct, but how about is_scope_relevant?

nathanhourt: Sure

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: So start_seq just needs a name change (like skip_seq) and then instead of end_seq we want num_seq?

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: fetch_block_by_id doesn't take a lock (which means get_transaction is wrong).  If we want to add taking the lock to fetch_block_by_id, we would need to make a public version that takes the lock and a private one that doesn't, since it is used internal to the class when the lock is already held.  It looks like most of the chain_controller public functions do not take the lock (or at least not that I can see).

nathanhourt: `fetch_block_by_id` doesn't need a lock, since it's not hitting the database at all. Blocks aren't stored in the database. We don't want those in RAM.

brianjohnson5972: isn't the db's lock also used to protect the fork_db, which the block could be retrieved from?

plugins/account_history_plugin/include/eos/account_history_plugin/account_history_object.hpp
brianjohnson5972: I think I initially collapsed it to avoid the second lookup (transaction id to block id), but I agree now since it will be much smaller (block_id not repeated for every account name in scope and account name only once with vector storage for transaction ids), but in terms of the logic it will just simplify calling equal_range to find and then will need some more logic to find the block_id.  Either way I agree that I need to separate the storage.

plugins/account_history_plugin/include/eos/account_history_plugin/account_history_object.hpp
```diff
@@ -13,15 +26,18 @@ class transaction_history_object : public chainbase::object<chain::transaction_h
...
 };
 
 struct by_id;
 struct by_trx_id;
+struct by_account_name;
 using transaction_history_multi_index = chainbase::shared_multi_index_container<
    transaction_history_object,
```
brianjohnson5972: I made this change for my next task branch, but I'll make it here.

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: What do you mean?  This is what fetch_block_by_id does:

optional<signed_block> chain_controller::fetch_block_by_id(const block_id_type& id)const
{
   auto b = _fork_db.fetch_block(id);
   if(b) return b->data;
   return _block_log.read_block_by_id(id);
}

nathanhourt: Right, that function never touches chainbase::database at all. It checks the forkdb, and if the block isn't recent enough to be in there (i.e. it's not irreversible yet) then it checks the block log, which is its own thing, not associated with chainbase. If chainbase were involved, you'd see a reference to _db somewhere in that function.

brianjohnson5972: My point was I thought that the db lock is what was protecting _fork_db and _block_log reader from being read in an inconsistent state.  What is to prevent me from hitting _fork_db's _index at the same time that chain_controller is telling _fork_db to pop the most recent block?

bytemaster: The fact that only once process has the ability to write and only one thread within that process.   All access to the chain controller should be via the main appbase thread.  

We do not use locks for ANYTHING.  If we need to do interthread communication we post events to a local ASIO handler.

brianjohnson5972:
All review comments should be addressed

bytemaster:
This looks OK for now, in the future we should attempt to figure out how to eliminate the need to change anything in the eos/chain library when writing a plugin.

PR #218, opened 2017-08-23
Use cpp_bin_float instead of cpp_dec_float

elmato:


bytemaster:
Since your research indicated that these methods both use integer math, I think cpp_dec_float would be better for currency applications.  What are your thoughts?

PR #220, opened 2017-08-23
CLI wallet issue 165

heifner:
* I still have the outstanding question about which wallet keys to use for the signatures. Currently using all unlocked wallet keys.
* I have not updated docs since I'm not certain on the exact final interface.
* I have one TODO about restricting to localhost which is waiting on http_plugin feature.
* The eosc is rough as I'm waiting for the eosc re-write.


nathanhourt:
> I still have the outstanding question about which wallet keys to use for the signatures. Currently using all unlocked wallet keys.

I'm think this is referring to determining which keys are necessary to authorize the transaction? If so, I'm happy to say that you can use the same logic the backend does!

The `AuthorityChecker` can take a list of all the keys your wallet can sign with, and tell you which ones are necessary and sufficient to authorize the transaction. The test framework does exactly that [here](https://github.com/EOSIO/eos/blob/366e70f800f849ee718fde7e0307a2168ec755a3/tests/common/database_fixture.cpp#L175).

It's pretty straightforward: Make an authority checker with all the keys you have available, then crunch all the messages in the transaction through it to make sure you actually do have keys sufficient to authorize them all, and then call `checker.used_keys()` to get the set of keys you need to sign with.

heifner:
> The AuthorityChecker can take a list of all the keys your wallet can sign with, and tell you which ones are necessary and sufficient to authorize the transaction. The test framework does exactly that here.

That appears to interact with the chainbase db to get the key associated with a given account and permission. However, I'm not sure we want the wallet interacting with the chainbase db. Also thinking about how the web wallet would sign a transaction.

In the steemit web code I found [this](https://github.com/steemit/steem-js/blob/53f503e73d8b55b8a1baa5432e2ec48e47046a22/src/broadcast/index.js#L87)
`keys[operation.roles[0]] = wif; // TODO - Automatically pick a role? Send all?`

So maybe I should store keys in the wallet via account/permission and grab account/permission from the message and use the keys for each account/permission.

nathanhourt:
> That appears to interact with the chainbase db to get the key associated with a given account and permission.

`AuthorityChecker` was intentionally designed to have no knowledge of `chainbase`, `chain_controller`, or anything else "big." To make this possible, it requires a callable capable of converting `{accountName, permissionName}` into an `Authority`.

This conversion is unavoidable, and you'll have to be capable of doing it if you're going to sign transactions; however, you can implement that callable however you want, and there is no dependency on having a full database around somewhere.

bytemaster:
We can provide an API hook that will do this function on EOSD.  Eg, a CLI wallet sends a list of public keys to EOSD along with a transaction, and gets back a list of used keys.

heifner:
Going to close this pull request now and will open a new one once the cli-wallet-165 has the additional changes mentioned on issue 165.

heifner:
- [ ] Add API hook to EOSD for CLI wallet to send list of public keys with transaction and get back a list of used keys
- [ ] Update wallet manager to use new api to sign transaction
- [ ] Determine mechanism to specify walletd endpoint in eosc
- [ ] Merge/add wallet CLI commands to eosc with inline doc
- [ ] Remove unneeded wallet_data attributes

plugins/wallet_plugin/wallet_manager.cpp
```diff
@@ -0,0 +1,164 @@
...
+   }
+   auto& w = wallets.at(name);
+   if (w->is_locked()) {
+      FC_THROW("Wallet is locked: ${w}", ("w", name));
+   }
+   w->import_key(wif_key);
+}
+
+chain::SignedTransaction
+wallet_manager::sign_transaction(const chain::SignedTransaction& txn, const chain::chain_id_type& id) {
```
nathanhourt: It appears that this function is just blindly signing the transaction with all keys in all unlocked wallets? A transaction is only valid if signed with no irrelevant keys, so this is inadequate. The `AuthorityChecker` class can tell you exactly which keys you need to sign with so you sign the transaction correctly.

Another question is: should we be grabbing keys from all wallets? Or should the user specify a particular wallet to get keys from?

heifner: This was before you pointed me to the `AuthorityChecker` and Dan said to add an api to EOSD to provide the keys.  I'm thinking that interface on `chain_controller` will look like this:

```C++
         /**
          * Determine which public keys are needed to sign the given transaction.
          * @param trx Transaction that requires signature
          * @param candidateKeys Set of public keys to examine for applicability
          * @return Subset of candidateKeys whose private keys should be used to sign transaction
          * @throws fc::exception if candidateKeys does not contain all required keys
          */
         flat_set<public_key_type> get_required_keys( const SignedTransaction& trx, const flat_set<public_key_type>& candidateKeys)const;
```
eosc will get keys from all **unlocked** wallets, call above, then ask walletd to sign with the subset of keys.

My thought is all unlocked wallets. But could add a command for which wallets. thoughts?

nathanhourt: > eosc will get keys from all unlocked wallets, call above, then ask walletd to sign with the subset of keys.

Makes sense.

What use case is the multiple wallet support aimed at?

heifner: > What use case is the multiple wallet support aimed at?

Good question. No idea. Just doing as I'm told ;)

plugins/wallet_plugin/wallet_manager.cpp
```diff
@@ -0,0 +1,164 @@
...
+namespace eos {
+namespace wallet {
+
+constexpr auto file_ext = ".wallet";
+constexpr auto password_prefix = "PW";
+
+std::string gen_password() {
+   auto key = fc::ecc::private_key::generate();
+   return password_prefix + utilities::key_to_wif(key);
+
```
heifner: Note here I'm creating a password for the user by creating a private key and prefixing it with "PW" so a user can easily distinguish it from other keys.

plugins/wallet_plugin/wallet_manager.cpp
bytemaster: Who is telling you to support multiple wallets, what is your definition of a wallet?

heifner: > Who is telling you to support multiple wallets, what is your definition of a wallet?
You did. But apparently, I misunderstood.  I understand a wallet to be a named file of encrypted private keys.

I assumed you wanted multiple wallets so a user could segment his keys and only unlock a particular wallet at a time.

The interface allows a user to create a named wallet. Open a named wallet. Unlock a particular wallet, etc. If there is only one wallet, a name would not be needed.

Here is the comment from the old eosc I was using as a guide:
* eosc unlock walletname PASSWORD
* eosc create wallet walletname PASS1 PASS2
* eosc unlock walletname PASSWORD
* eosc wallets -> prints list of wallets with * next to currently unlocked
* eosc keys -> prints list of private keys
* eosc importkey privatekey -> loads keys
* eosc accounts -> prints list of accounts that reference key
* eosc lock

programs/eosc/main.cpp
```diff
@@ -160,13 +187,15 @@ int main( int argc, char** argv ) {
    string name;
    string ownerKey;
    string activeKey;
+   bool sign = false;
    auto createAccount = create->add_subcommand("account", "Create a new account on the blockchain", false);
    createAccount->add_option("creator", creator, "The name of the account creating the new account")->required();
    createAccount->add_option("name", name, "The name of the new account")->required();
    createAccount->add_option("OwnerKey", ownerKey, "The owner public key for the account")->required();
    createAccount->add_option("ActiveKey", activeKey, "The active public key for the account")->required();
+   createAccount->add_option("sign", sign, "Specify if unlocked wallet keys should be used to sign transaction");
```
heifner: I really don't like how this add_option works for bool. It treats it as an int. So you have to pass 1 on the command line. But I just did what was done for the other existing bool options.

programs/eosc/main.cpp
```diff
@@ -275,10 +306,87 @@ int main( int argc, char** argv ) {
...
+
+   // list keys
+   auto listKeys = wallet->add_subcommand("keys", "List of private keys from all unlocked wallets in wif format.", false);
+   listKeys->set_callback([] {
+      const auto& v = call(wallet_host, wallet_port, wallet_list_keys);
+      std::cout << fc::json::to_pretty_string(v) << std::endl;
+   });
+
    // Benchmark subcommand
    auto benchmark = app.add_subcommand( "benchmark", "Configure and execute benchmarks", false );
```
heifner: I didn't add sign to benchmark. Didn't know if that was desired.

programs/eosc/main.cpp
heifner: Figured it out.` add_flag()` with "-s,--sign"

heifner:
I made a couple of small changes after putting this into needs review. But it is ready, minus any peer review issues.

Issue #203, opened 2017-08-17
Make eosc easier to use


nathanhourt:
OK, so eosc is really rough right now, both from an architecture perspective, and from a usability perspective. The former isn't super critical, but it's still nice to improve; however, the latter is a showstopper.

Improve eosc usability by making clear and complete help information available.

 - [x] Update eosc to be fully self-documenting
 - [x] Update any docs impacted by these changes

pmesnier:
Since I had a hand in both of those conditions, please let me know if there is anything I can do to help out. Or maybe just to explain why something was done the way it was.

nathanhourt:
Thanks, Phil! Mostly I want to make sure I don't leave behind any use cases it currently supports. I think that probably means the batch processing mode, which I don't fully understand the purpose or effect of. What is that feature for, and why do we need it?

pmesnier:
As I was starting to make some scripted tests that would exercise the p2p network code, I found myself wanting to feed commands to it via more traditional unix streaming services, with the command line arguments being reserved for traditional command line options. Just like you are not fond of virtual inheritance, I'm not fond of unnecessarily starting serial copies of a program. 

In fact, batch mode isn't operational. I couldn't come up with a use case, so I stopped short of having to test and document it. The streaming feature is line-oriented, although multi-line commands are supported though the use of quoted strings. I also preserved the original behavior as well. The major refactoring I did was to move the command processing out of main to make it easier to deal with alternate input sources.

Given your concern about usability I think it would be reasonable to take what I did and make a differently named program. maybe eosst?

brianjohnson5972:
FYI: I will be doing a pull request soon with changes to the "transactions" command in eosc.

nathanhourt:
OK, this should be pretty much complete at this point, thought there may be other docs to update that I don't know about.

nathanhourt:
Preq #221

pmesnier:
I'm looking for the most appropriate way to override the default host/port to which eosc sends the messages. I've thought of: 
eosc --server host:port command ...,  
eosc \@host:port commmand ...
eosc command@host:port ....

in each case, an omitted host or port value is defaulted. In the case of streaming mode, I'd like this value to be changeable

@nathanhourt are you still actively refactoring eosc?

thomasbcox:
cleaning up old assignments

heifner:
This was merged, closing.

PR #221, opened 2017-08-24
Eosc overhaul

nathanhourt:
See discussion in #203


nathanhourt:
Are there any docs to update other than https://eosio.github.io/eos/group__eosc.html?

bytemaster:
This pull request needs to be rebased on current master with fixed submodule.

programs/eosc/main.cpp
```diff
@@ -21,6 +22,8 @@
 
 #include <fc/io/fstream.hpp>
 
+#include "CLI11.hpp"
```
heifner: Comment with verison of CLI11 used might be helpful, I assume version 1.1.

nathanhourt: At this point we're using the HEAD version, as 1.1 contained a bug [which I reported](https://github.com/CLIUtils/CLI11/issues/23) and he fixed, slated for release in version 1.2. I've included the header in our source distribution so there's no worries about having to go find the right version.

PR #225, opened 2017-08-25
Reverting unintended revert of chainbase submodule.

brianjohnson5972:
Last commit moved chainbase back to where it was in May, so bringing it back to most recent changes.


dskvr:
Sorry, not sure what happened there.

PR #227, opened 2017-08-25
process generated transactions

### LINKED ISSUE #174: Process Generated Transactions
`GeneratedTransaction` s can be scheduled but they cannot be processed at the moment.  

This task will refactor processing of `SignedTransaction` s that is currently part of `chain_controller` to allow for processing `GenerateTransaction` s as well with as much shared code as possible.  

NB: this task will not provide API or WASM-VM support for generating transactions.  That will come in another issue.

wanderingbort:
resolves #174 

test cases will be easier to write and more meaningful once there is contract support for sync and async transactions.  However I wanted to go ahead and review what was working without impacting the existing code/tests.


libraries/chain/chain_controller.cpp
```diff
@@ -490,6 +497,132 @@ void chain_controller::apply_block(const signed_block& next_block, uint32_t skip
...
+   } FC_RETHROW_EXCEPTIONS(warn, "in .account");
+   try {
+      check_output(expected.permission, actual.permission);
+   } FC_RETHROW_EXCEPTIONS(warn, "in .permission");
+}
+
+template<>
+void check_output(const types::Message& expected, const types::Message& actual) {
+   try {
+      check_output(expected.code, actual.code);
```
bytemaster: why don't you move the try{ } catch{} . into check_output and pass the "in .x" as a const char* 3rd argument?  I think it would make the code more readable.

libraries/chain/include/eos/chain/transaction.hpp
```diff
@@ -52,6 +52,53 @@ namespace eos { namespace chain {
...
+
+      template <typename... Args>
+      void emplace_message(Transaction& t, Args&&... a) {
+         t.messages.emplace_back(Message(std::forward<Args>(a)...));
+      }
+
+      /**
+       * clear all common data
+       */
+      inline
```
bytemaster: put inline on same line as return statement.

programs/eosc/main.cpp
```diff
@@ -322,11 +322,11 @@ int send_command (const vector<string> &cmd_line)
 
     SignedTransaction trx;
     trx.scope = sort_names({sender,recipient});
-    trx.emplaceMessage(config::EosContractName, vector<types::AccountPermission>{{sender,"active"}}, "transfer",
+    transaction_helpers::emplace_message(trx, config::EosContractName, vector<types::AccountPermission>{{sender,"active"}}, "transfer",
                        types::transfer{sender, recipient, amount});
     auto info = get_info();
     trx.expiration = info.head_block_time + 100; //chain.head_block_time() + 100;
-    trx.set_reference_block(info.head_block_id);
+    transaction_helpers::set_reference_block(trx, info.head_block_id);
```
bytemaster: I find the name transaction_helpers long, I feel we should say "using transaction_helpers" unless there is clear possibility of name conflict.

tests/slow_tests/slow_tests.cpp
```diff
@@ -1116,9 +1117,9 @@ R"(
       trx.messages.resize(1);
       trx.messages[0].code = config::EosContractName;
       trx.messages[0].authorization.emplace_back(types::AccountPermission{"simplecoin","active"});
-      trx.setMessage(0, "setcode", handler);
+      transaction_helpers::set_message(trx, 0, "setcode", handler);
```
bytemaster: perhaps renaming the methods .  transaction_set_message rather than transaction_helpers::set_message if we are going to be verbose everywhere.  If follows the naming convention one would use for a object-oriented C api.

Issue #167, opened 2017-08-15
Account History API lookup of controlled accounts


brianjohnson5972:
All accounts controlled by the requesting account will be returned.

brianjohnson5972:
Will add index for tracking each account listed in a named Authority for the controlled account, either from newaccount or updateauth messages.  Will update/remove all accounts for the controlled account's named Authority based on updateauth or deleteauth messages.

brianjohnson5972:
Should this interface just be send in an account name and return a list of account names that account controls, or should the permission names also be included?  (Either or both of the controlling account's permission name and the controlled account's permission name?)

PR #247, opened 2017-08-29
Added lookup of controlled accounts by controlling account name.

brianjohnson5972:
When get_controlled_accounts is called the passed in account name is used to retrieve all accounts that have provided an Authority that has given permissions to the passed in account .


brianjohnson5972:
#167

Issue #248, opened 2017-08-29
Native Transfer message Memo


bytemaster:
Transfer's require a memo for things like exchange deposits.  It was originally intended that the user could embed arbitrary data at the end of the "hex" string for the defined message.  Making the memo an explicit field enables it to work with the JSON ABI when it would otherwise be ignored.

Issue #70, opened 2017-07-08
Transfer example .. Unit test typo?


jcalfee:
Is this a typo?
```c++
trx.messages[0].recipients = {"inita", config::EosContractName}
```
This value is not used then re-assigned before the transaction push:
```c++ 
trx.messages[0].recipients = {"inita", "initb"};
```

jcalfee:
https://github.com/EOSIO/eos/blob/422c40f7237552c539f65a2c0ebfe54936831dfe/tests/tests/native_contract_tests.cpp#L93

nathanhourt:
Doesn't line 107 test that the first set of recipients is invalid, then set them correctly so the transaction can be pushed?

jcalfee:
yes, I see it .. thanks

PR #249, opened 2017-08-29
Adding memo to transfer type #248

bytemaster:
- also adding readscope to Transaction header #70


bytemaster:
I'm working on fixing the build, will push in a moment.

libraries/types/types.eos
```diff
@@ -71,6 +72,7 @@ struct transfer
    from      AccountName  # may not be the message.sender if message.sender has delegated authority by from
    to        AccountName
    amount    UInt64
+   memo      String
```
heifner: Seems like we should limit the size of the string. Currently, this is a std::string.

PR #251, opened 2017-08-29
Log 244

MatthewPappas:
All done, please approve for merge to master.


libraries/chain/chain_controller.cpp
```diff
@@ -210,7 +212,26 @@ bool chain_controller::_push_block(const signed_block& new_block)
...
+      if( (fc::time_point::now() - new_block.timestamp) < fc::seconds(60) )
+      {
+         auto exec_stop = std::chrono::high_resolution_clock::now();
+         auto exec_ms = std::chrono::duration_cast<std::chrono::milliseconds>(exec_stop - exec_start);      
+         size_t trxcount = 0;
+         for (const auto& cycle : new_block.cycles)
+   	      for (const auto& thread : cycle)
+   		      for (const auto& trx : thread.user_input) 
+                  trxcount++;
+         ilog( "producer=[${a}], blocktime=${b}, blocknum=${c}, trxcount=${d}, pendingcount=${e}, exectime_ms=${f}",
```
bytemaster: ${A} ${c} ${b} |  ${d} trx  | ${e} pending | ${f} ms

The output you generated is too verbose and syntax heavy... 
Also rather than using "a", "b", "c"... you should use sensical names.

libraries/utilities/generic.cpp
```diff
@@ -0,0 +1,16 @@
+#include <string>
```
bytemaster: These methods should not be defined or used, where ever they are used we need to do something different.  I do not even se their use in this commit.

pmesnier:
The generic code is not actually being used, it should not be included in this PR. Please do a new commit to your branch.

MatthewPappas:
The requested changes are made.

libraries/chain/chain_controller.cpp
```diff
@@ -54,6 +54,7 @@
 #include <fstream>
 #include <functional>
 #include <iostream>
+#include <chrono>
 
 //#include <Wren++.h>
```
nathanhourt: Lol, go ahead and delete this line. :)

libraries/chain/chain_controller.cpp
```diff
@@ -210,7 +211,26 @@ bool chain_controller::_push_block(const signed_block& new_block)
...
       apply_block(new_block, skip);
+      if( (fc::time_point::now() - new_block.timestamp) < fc::seconds(60) )
+      {
+         auto exec_stop = std::chrono::high_resolution_clock::now();
+         auto exec_ms = std::chrono::duration_cast<std::chrono::milliseconds>(exec_stop - exec_start);      
+         size_t trxcount = 0;
+         for (const auto& cycle : new_block.cycles)
+   	      for (const auto& thread : cycle)
+   		      for (const auto& trx : thread.user_input) 
+                  trxcount++;
```
nathanhourt: Bad indentation on above lines... Just replace the innermost loop with `trxcount += thread.user_input.size()` -- might as well skip a few iterations, ya?

MatthewPappas: Will do

libraries/chain/chain_controller.cpp
MatthewPappas: Will do

MatthewPappas:
Changes have been pushed, please proceed.

Issue #243, opened 2017-08-28
Eosc/eosd multi-transaction mode


nathanhourt:

bytemaster:
push_transactions has been implemented and tested in eosc on the benchmark branch.

PR #259, opened 2017-08-29
implementing push_transactions api #243

bytemaster:


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -263,9 +263,23 @@ read_write::push_block_results read_write::push_block(const read_write::push_blo
...
-   auto ptrx = db.push_transaction(params, skip_flags);
+   auto pretty_input = db.transaction_from_variant( params );
+   auto ptrx = db.push_transaction(pretty_input, skip_flags);
    auto pretty_trx = db.transaction_to_variant( ptrx );
-   return read_write::push_transaction_results{ params.id(), pretty_trx };
+   return read_write::push_transaction_results{ pretty_input.id(), pretty_trx };
+}
+
+read_write::push_transactions_results read_write::push_transactions(const vector<read_write::push_transaction_params>& params) {
+   push_transactions_results result;
```
heifner: Could reserve `result.reserve(params.size());`

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -263,9 +263,23 @@ read_write::push_block_results read_write::push_block(const read_write::push_blo
...
+   for( const auto& item : params ) {
+      try {
+        result.push_back( push_transaction( item ) ); 
+      } catch ( const fc::exception& e ) {
+        result.push_back( read_write::push_transaction_results{ chain::transaction_id_type(), 
+                          fc::mutable_variant_object( "error", e.to_detail_string() ) } );
+      }
```
heifner: Could use emplace_back()

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -263,9 +263,23 @@ read_write::push_block_results read_write::push_block(const read_write::push_blo
...
 read_write::push_transaction_results read_write::push_transaction(const read_write::push_transaction_params& params) {
-   auto ptrx = db.push_transaction(params, skip_flags);
+   auto pretty_input = db.transaction_from_variant( params );
+   auto ptrx = db.push_transaction(pretty_input, skip_flags);
    auto pretty_trx = db.transaction_to_variant( ptrx );
-   return read_write::push_transaction_results{ params.id(), pretty_trx };
+   return read_write::push_transaction_results{ pretty_input.id(), pretty_trx };
+}
+
+read_write::push_transactions_results read_write::push_transactions(const vector<read_write::push_transaction_params>& params) {
```
heifner: `const push_transactions_results& params` would match the declaration.

wanderingbort: I think you mean `push_transaction_params` but otherwise 👍

heifner: yes, of course, thanks

PR #265, opened 2017-08-29
Add a single threaded scheduler and make it the default

### LINKED ISSUE #263: Add single threaded scheduler for reference

wanderingbort:
closes #263

### DIFF
--- a/libraries/chain/block_schedule.cpp
+++ b/libraries/chain/block_schedule.cpp
@@ -289,4 +289,23 @@ block_schedule block_schedule::by_cycling_conflicts(
    return from_entries(schedule);
 }
 
+block_schedule block_schedule::in_single_thread(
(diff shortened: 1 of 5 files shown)


libraries/chain/block_schedule.cpp
```diff
@@ -289,4 +289,23 @@ block_schedule block_schedule::by_cycling_conflicts(
...
+   thread.transactions.reserve(transactions.size());
+   for(const auto& t: transactions) {
+      if (skipper.should_skip(t)) {
+         break;
+      }
+
+      thread.transactions.push_back(t);
+   }
+
+   return block_schedule { { { thread } } };
```
bytemaster: perhaps this should be moved into block_schedule?  It looks like it will create a copy of the thread which could be quite expensive at load.

PR #266, opened 2017-08-30
Add doxygen guidelines

andriantolie:
Credit to @ChristianDunst

### DIFF
--- a/Doxygen guidelines.md
+++ b/Doxygen guidelines.md
@@ -1,0 +1,302 @@
+#Doxygen Guide for EOS
+<br/>
+
+## Index
+- [Introduction](#Introduction)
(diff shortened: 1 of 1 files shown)


thomasbcox:
I'm wondering if we need the file "Doxygen" when we're supposed to use the file "eos.doxygen" as the One True Config File. Can someone review this and either add commentary to clarify the situation or delete the extraneous file? My fear is someone will use the default file "Doxygen" simply because it's there. And my sense is that having the default file lying around not being used is a problem waiting to happen. An alternative is to merge the two somehow, or move the contents of "eos.doxygen" into "Doxygen". Is there a design pattern we should follow here, or a convention, that would clarify?  @ChristianDunst ?

Issue #239, opened 2017-08-28
Create index with string key sorted alphabetically


bytemaster:
Ensure it works with get table rows and other APIs for tables.

elmato:
#859aa75
The str index has been refactored to use shared_string instead of Name type.
Need more testing and to solve a problem with the wasm-vm that for some reason is not finding some intrinsic functions related to this index.

elmato:
Pull request ready to review and merge EOSIO/eos#274

PR #274, opened 2017-08-30
Create index with string key sorted alphabetically

elmato:
ref  #239


contracts/eoslib/db.h
```diff
@@ -113,7 +113,65 @@ int32_t remove_i64( AccountName scope, TableName table, void* data );
...
+ * @pre *((Name*)data) stores the primary key
+ * @pre scope is declared by the current transaction
+ * @pre this method is being called from an apply context (not validate or precondition)
+ *
+ * @post a record is either created or updated with the given scope and table.
+ *
+ * @throw if called with an invalid precondition execution will be aborted
+ *
+ */
+ int32_t store_str( AccountName scope, TableName table, const void* data, uint32_t datalen );
```
bytemaster: This should be a KEY / VALUE database (eg a mapping) not a set.

contracts/eoslib/db.h
```diff
@@ -113,7 +113,65 @@ int32_t remove_i64( AccountName scope, TableName table, void* data );
...
 
+/**
+ * @param scope - the account scope that will be read, must exist in the transaction scopes list
+ * @param table - the ID/name of the table within the current scope/code context to modify
+ *
+ * @return 1 if a new record was created, 0 if an existing record was updated
+ *
+ * @pre datalen >= sizeof(uint64_t)
+ * @pre data is a valid pointer to a range of memory at least datalen bytes long
+ * @pre *((Name*)data) stores the primary key
```
bytemaster: the primary key should be a variable length string, not a Name*.

libraries/chain/include/eos/chain/key_value_object.hpp
```diff
@@ -55,9 +55,39 @@ namespace eos { namespace chain {
...
+      OBJECT_CTOR(keystr_value_object, (value))
+      
+      typedef Name key_type;
+      static const int number_of_keys = 1;
+
+      id_type               id;
+      AccountName           scope; 
+      AccountName           code;
+      AccountName           table;
+      Name                  primary_key;
```
bytemaster: Name should be a shared_string type.

PR #284, opened 2017-08-31
Cache `WASM_LLVM_CONFIG` value

jgiszczak:
`WASM_LLVM_CONFIG` will now be a properly cached variable, settable with `cmake -DWASM_LLVM_CONFIG=/path/to/llvm-config` and persisting across runs of `cmake` and editable with `ccmake`|`cmake-gui`.  If the value was not specified on the command line or retrieved from the cache, it will resort to the environment variable.


jgiszczak:
Unexpected commits included.

PR #292, opened 2017-08-31
Improved docker building

### LINKED ISSUE #289: Dockerfile does not have to clone repo
Right now the Dockerfile clones in the git repo through a RUN call. Replacing this with a COPY will not only marginally speed up the build, it will also enable proper caching of the build tools. I think right now the if you don't use --no-cache the build will just skip the whole building process due to RUNs being cached by Docker.

With this we will also be able to build different branches.

The dockerfile will be moved to the Dockerfile to the project root, because `COPY ../ <dest>` is otherwise out of the build context.

toonsevrin:
build tools properly cache and you can build specific branches/commits with docker.

Fixes  #289

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,7 @@
+image: docker:latest
+
+build:
+  script:
+      - docker login -u gitlab-ci-token -p $CI_JOB_TOKEN $CI_REGISTRY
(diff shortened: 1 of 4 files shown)


toonsevrin:
I'm not sure why the travis check failed, could someone verify this/rebuild this pr?

toonsevrin:
This speeds up my docker builds by about a 10x factor (from 29 minutes to 3 minutes).

PR #297, opened 2017-09-01
Make binaryen an external dependency

brianjohnson5972:
Removed the binaryen submodule, added instructions to compile and install binaryen, and changed cmake to find binaryen binaries using BINARYEN_ROOT var or else find them on path.


brianjohnson5972:
Took 19 minutes.  Could knock off another 2 minutes if we prebuild boost.

https://travis-ci.org/EOSIO/eos/builds/270846270

brianjohnson5972:
#297

Issue #214, opened 2017-08-22
DAWN-496 ⁃ Initialize Genesis from ERC-20 Snapshot


bytemaster:
1. Add snapshot script to /eos/programs/snapshot
2. Provide instructions on using this script in /eos/progams/snapshot
3. Create a program to convert snapshot.csv into snapshot.json file that can be used with eosd to start a properly configured blockchain
4. Check in an snapshot.json file that is ready to go for the test network

┆Attachments: <a href="https:&#x2F;&#x2F;blockone.atlassian.net&#x2F;secure&#x2F;attachment&#x2F;10122&#x2F;snapshot+%28period+118%29.json">snapshot (period 118).json</a> | <a href="https:&#x2F;&#x2F;blockone.atlassian.net&#x2F;secure&#x2F;attachment&#x2F;10121&#x2F;snapshot+118_csv.zip">snapshot 118_csv.zip</a>

dskvr:
We will need to determine a the final block for the testnet snapshot.

nathanhourt:
The EOS ERC-20 freezes all balances at the end of the sale, does it not? So knowing the exact ethereum block wouldn't matter, you just need some block after the sale ends?

dskvr:
@nathanhourt Yes and no. The Token is frozen at end of crowdsale, but registration transactions could still happen, causing mismatch between different snapshots. 

Alice wants to verify the snapshot and runs generator. Meanwhile, Roger sends a Registration TX. Bob wants to verify the snapshot and runs generator. Alive and Bob compare snapshots but they are different.

However, this is mostly in regards to testnet screenshot. Ending block is important because no tokens are frozen, if multiple people generated a snapshot, they would all have different versions. That's not easy to verify.

rangershi:
grate

dskvr:
What is the deadline for snapshot.json?

dskvr:
[ ] Add snapshot script to /eos/programs/snapshot
[ ] Provide instructions on using this script in /eos/progams/snapshot
[ ] Create a program to convert snapshot.csv into snapshot.json file that can be used with eosd to start a properly configured blockchain
[ ] Check in an snapshot.json file that is ready to go for the test network

dskvr:
Snapshot specifications have changed since this issue was initially opened. Unregistered addresses will now be included in snapshot through a fallback strategy which is currently being tested. This change of strategy has a completely different approach from initial requirements, as it requires full supply monitoring. 

[...]

PR #314, opened 2017-09-03
/Programs/Snapshot directory as genesis submodule

dskvr:
Submodule did not register in last pull request.

#214


.gitmodules
```diff
@@ -6,3 +6,6 @@
 	path = libraries/appbase
 	url = https://github.com/eosio/appbase
         ignore = dirty
+[submodule "programs/snapshot"]
+	path = programs/snapshot
+	url = git@github.com:EOSIO/genesis.git
```
nathanhourt: Using an SSH URL is nice for us devs, but it breaks for people who don't have SSH keys set up on github. Best to use HTTPS URLs instead.

dskvr: You got it

dskvr:
@nathanhourt Fixed

PR #322, opened 2017-09-04
Skipsig 317

MatthewPappas:
Please commit this to master, it fixes --skip-transaction-signatures

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -755,7 +755,7 @@ flat_set<public_key_type> chain_controller::get_required_keys(const SignedTransa
 }
 
 void chain_controller::check_transaction_authorization(const SignedTransaction& trx, bool allow_unused_signatures)const {
-   if ((_skip_flags & skip_transaction_signatures) && (_skip_flags & skip_authority_check)) {
+   if ((_skip_flags & skip_transaction_signatures) /* && (_skip_flags & skip_authority_check)*/ ) {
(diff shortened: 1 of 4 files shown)


libraries/chain/chain_controller.cpp
```diff
@@ -755,7 +755,7 @@ flat_set<public_key_type> chain_controller::get_required_keys(const SignedTransa
 }
 
 void chain_controller::check_transaction_authorization(const SignedTransaction& trx, bool allow_unused_signatures)const {
-   if ((_skip_flags & skip_transaction_signatures) && (_skip_flags & skip_authority_check)) {
+   if ((_skip_flags & skip_transaction_signatures) /* && (_skip_flags & skip_authority_check)*/ ) {
```
bytemaster: && should be || this "fix" creates a new bug.

libraries/chain/chain_controller.cpp
```diff
@@ -1140,9 +1140,9 @@ void chain_controller::initialize_chain(chain_initializer_interface& starter)
 } FC_CAPTURE_AND_RETHROW() }
 
 chain_controller::chain_controller(database& database, fork_database& fork_db, block_log& blocklog,
-                                   chain_initializer_interface& starter, unique_ptr<chain_administration_interface> admin)
+                                   chain_initializer_interface& starter, unique_ptr<chain_administration_interface> admin, uint64_t skip_flags)
```
bytemaster: skip flags are not constant, they are determined by the caller.

Issue #318, opened 2017-09-03
p2p transaction cache purging


pmesnier:
This is the last of the implementation basics for the net plugin. The cache holds transaction information for quick verification and forwarding. This issue covers expunging old data from the cache to avoid consuming all available memory.

Issue #291, opened 2017-08-31
STAT-138 ⁃ P2P Network Improvements


bytemaster:
## Easy
- Clean up Console Spam

## Stability
1. Reconnect every 30 seconds to every peer in our config file
6. ping pong latency measurement protocol and keep alive 
7. on connection check latency / clock skew
8. Disable Naggle algorithm on connections (may introduce lag)

## Usability
11. RPC API for dynamically adding / removing peers and checking their status
8. block only mode...send full blocks not pending trx (with delay for last irreversible)

## Security
2. Limit the maximum number of incoming connections
9. Error reporting and abuse detection
10. Authentication on handshake

## Performance

3. Don't relay messages to peers that already sent us the message
4. Send block recipe to reconstruct from trx previously sent
5. For large messages (block or trx) send notice hash + query to fetch
      - query from only one peer at a time
      - timeout quickly 
6. Multi-Threading of network processing 

TODO: create per-item issues on github and link to issue number.

pmesnier:
I'll bet this in place tonight* (* for some timezone :-)

bytemaster:
In terms of priority:
1. Stability
2. Usability
3. Security
4. Performance

bytemaster:
Please refine this into individual tickets and mark the parts that are actually in progress as in progress.

pmesnier:
I've gone through all of the sub-issues and marked them as either done or pending a particular issue  number. In a couple of cases I have additional questions and a couple I just feel wont get done in the next day or two.

## Easy
- Clean up Console Spam  ***DONE***

## Stability
1. Reconnect every 30 seconds to every peer in our config file ***DONE***
6. ping pong latency measurement protocol and keep alive ***in progress #395***
7. on connection check latency / clock skew ***Pending #395***
8. Disable Naggle algorithm on connections (may introduce lag) ***DONE #394***
9. [not previously listed] synchronizing across multiple peers didn't work ***DONE***
9. [not previously listed] synchronizing newly started nodes needs to survive a slow or crashed supplier ***Pending #406***

## Usability
11. RPC API for dynamically adding / removing peers and checking their status ***Need more information***
8. block only mode...send full blocks not pending trx (with delay for last irreversible) ***DONE, sort of, see question***

Question: with multiple producer nodes, isn't it alwasys necessary for pending transactions to be communicated to a block producer? or is it sufficient for transactions to be held as pending until it's node becomes the current block producer?

## Security
[...]

PR #329, opened 2017-09-04
purger of entries in the local transaction cache. Ref #291 #318

pmesnier:
I left the timer in, doing the expire check on every received txn seened a little excessive. I did set the tier to expire every 3 seconds to be basically per block


bytemaster:
Why don't you just use the block message to trigger it rather than creating a timer?

pmesnier:
Because I had the timer code in place and while I have it at 3 seconds now, we may want to adjust the timeout to be maybe 1 second or 10 seconds later.

pmesnier:
Also, the timer decouples the reception of the block from the action of purging the cache, which will be advantageous when we add MT to the net plugin.

PR #346, opened 2017-09-05
Small print.hpp documentation update

toonsevrin:
According to https://stackoverflow.com/questions/20944784/why-is-conversion-from-string-constant-to-char-valid-in-c-but-invalid-in-c. As I don't write a lot of cpp myself I'm not 100% sure why this is, but at least the code compiles with that.


nathanhourt:
Yep, const is correct there, but we use `const char *` rather than `char const *` -- please switch those around.

PR #348, opened 2017-09-05
Add Access-Control-Allow-Origin to http

heifner:
Added Add Access-Control-Allow-Origin : *


bytemaster:
Can we make setting the origin policy a config parameter rather than defaulting to something less secure?

PR #352, opened 2017-09-05
Add network topology diagrams

jgiszczak:
Add network topology diagrams, including graphviz source and a script to
regenerate them.  Integrate into testnet markdown documentation.


jgiszczak:
Added as siblings of testnet.md for the time being.  Should be relocated when `docs` is revisited.

PR #361, opened 2017-09-06
Updates for SDK

nathanhourt:
Some updates I made in the process of preparing the SDK


bytemaster:
I previously implemented https://github.com/EOSIO/eos/blob/master/libraries/chain/wast_to_wasm.cpp

I would suggest removing eos::utilities::assembl_wasm in favor of the wast_to_wasm and wasm_to_wast functions defined in chain.

I will merge for now, but we should remove this kind of duplicate code.

nathanhourt:
Ahh, whoops, I missed it. I'll fix that.

Issue #374, opened 2017-09-07
Add Access-Control-Allow options to eosd


heifner:
Need to add another header along with the `Access-Control-Allow-Origin`, which is the `Access-Control-Allow-Headers`, which should let us specify a list of accepted request headers from api clients.

See:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers

Also add `Access-Control-Allow-Credentials: true` as well.
See:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

These should be added to config.ini.

heifner:
This has been merged. Closing.

PR #376, opened 2017-09-07
Add Access-Control-Allow options

heifner:
Issue #374.

To host eosd behind ssl, some additional Access-Control-Allow options where required. This adds:
* Access-Control-Allow-Headers
* Access-Control-Allow-Credentials

Added as options to the config.ini

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -80,7 +80,9 @@ namespace eos {
          //asio::io_service         http_ios;
          map<string,url_handler>  url_handlers;
          optional<tcp::endpoint>  listen_endpoint;
-         string access_control_allow_origin;
+         string                   access_control_allow_origin;
+         string                   access_control_allow_headers;
+         bool                     access_control_allow_credentials = false;
 
          websocket_server_type    server;
    };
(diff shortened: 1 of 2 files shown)


heifner:
Not sure why there is no travis build. Also it doesn't say travis build is required.

heifner:
Travis was having issues earlier today. It is obviously back up and running now.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -144,6 +159,12 @@ namespace eos {
                      if (!my->access_control_allow_origin.empty()) {
                         con->append_header("Access-Control-Allow-Origin", my->access_control_allow_origin);
                      }
+                     if (!my->access_control_allow_headers.empty()) {
```
wanderingbort: While it does no harm to include `Access-Control-Allow-Headers` on every response, the standards are such that it is  only meaningful as part of an `OPTIONS` request. 

https://www.w3.org/TR/cors/#resource-requests (section 6.2.4)

this does not apply to `Access-Control-Allow-Origin` and `Access-Control-Allow-Credentials` as those are included on preflighted, simple CORS, regular and redirect responses.

wanderingbort: perhaps conditionally include the `OPTIONS` headers based on the method?

for reference, there are other headers we may want to consider that are `OPTIONS` only:

- `Access-Control-Request-Method` to limit to `POST` `GET` and `OPTIONS`?
- `Access-Control-Max-Age` to some resonable limit to reduce future preflights from the same client?

heifner: I believe the web wallet team is going to propose some changes to the http interface. These sort of items would likely be part of that. These were added as the minimum needed for the web wallet demo.

Issue #175, opened 2017-08-15
Contract API support for generating asynchronous transactions


wanderingbort:
Related to #174

This task is for implementation of the contract API and related support for producing `GeneratedTransaction` for asynchronous calls to other contract methods.

wanderingbort:
### Existing Stubs
At the moment we have  a stub from several months ago which defines this intrinsic in the wasm-vm:
```c
int send(void *tx_buffer, int tx_buffer_size);
```
It seems that the idea was to serialize a whole transaction into the proper packed form and then pass it to the native code. 
PROS: 

- dirt simple ABI

CONS:

- guarantees that the wasm must understand `fc::pack` binary format for things like `vector`
- probably requires some concept of `malloc` or at least `alloca` for message sizes that cannot be determined at compile time

### Managed Transaction Alternative
alternatively, if we make a slightly more complex api.  We may be able to simplify what have to implement in the more restrictive wasm-vm environment

```c
typedef uint32_t TransactionHandle;
typedef uint64_t FuncName;
 
// create a transaction and auto-determine refBlockPrefix, refBlockNum and Expiration
TransactionHandletransaction_create(AccountName *scopes, int scope_count);

// add a message to the pending transaction
void transaction_add_message(TransactionHandle trx, AccountName code, FuncName type, PermissionName*permissions, int permission_count, void *data, int data_size);

// publish transaction (defaults to async, but can be sync)
void transaction_send(TransactionHandle trx, bool immediate = true);
```

PROS: 

- unless the `data` for a message is dynamically sized, we may not have to re-implement `fc::pack` in wasm-vm OR provide `malloc`, `alloca`, etc .  
- parallels `readMessage` which is currently given a raw buffer which must be de-serialized in the VM 
- provides some insulation for maintaining backwards compatibility as the native data structures and procedures are abstracted
- less native-side de/re-serialization of things to check validity of things like `transaction::scope`, `message::code` and `message::authorization`

CONS:

- fat ABI (more like a C-style library than intrinsics)
- retaining the transaction on the native side requires more book keeping in the application context

###### Other thoughts:
in the above API, `scope` is set once when the transaction is created.  It may be more developer friendly to have an API endpoint to add them ad-hoc (as you collect messages and realize you need more scopes):
```c
// add a scope to the pending transaction
void transaction_add_scope(TransactionHandle trx, AccountName scope);
```

bytemaster:
[...]

PR #389, opened 2017-09-08
Feature/175 async transactions

wanderingbort:
See #175 for notes about implementation and other fun stuff.


wanderingbort:
Sorry, I'm shooting int the dark for the build errors.  my local environment is not the same as travis in some strange way.  Nothing travis has brought up has been incorrect, it is just frustrating to have to bounce off travis for changes.  😠

wanderingbort:
Wow, its literally the difference between `clang-4.0` from Ubuntu 16.04 official repositories and `clang-4.0` from llvm's release repository.  

granted that is effectively `4.0.0` vs `4.0.1` but still.  Anyway... now that I have a local build environment I can agree with Travis... its building fine 😜

Issue #129, opened 2017-08-02
Implement malloc, free, and realloc for WASM


bytemaster:
The basic idea is that we should make it easy for developers to do some dynamic memory allocation for certain applications.

Currently a process is limited to 64KB of memory including stack and static data. This memory limit creates a defacto limit on the size of a generated transaction which would likely be less than 32KB.

The way I envision malloc being implemented is via a global static char[] buffer that uses pre-processor macro to define the reserved memory.  

We need to ensure that compiled programs do not attempt to reserve more than the 64KB, including the reserved stack space.

64KB should be enough for anybody... :)

xh3:
Famous last words

376726756:
hi,very nice

brianjohnson5972:
@bytemaster

I have some questions on these requirements.
If I am reading this correctly, I am adding intrinsic functions for malloc, realloc, and free, to provide memory for use during WASM execution.
This looks like it should be allocated on the MemoryInstance.

How much effort are we wanting to put into memory pool management to avoid ending up with a fragmented address space? (i.e. would we rather have a naive approach that will definitely be completed in time, vs. a more useful solution for the long run?)

bytemaster:
We are not adding intrinsic functions, we are implementing this entirely within C++ headers of EOSLIB.

brianjohnson5972:
Will need to add a memory.h file to EOSLIB and allocate a static char buffer.

Initial iteration will just have alloc and realloc grabbing the next unallocated region, free will be a no-op.
Next iteration will implement adding another 64K page and a new block of memory (and zero-ing the page when finished with exectution)
Depending on performance of allocating 64K page we will add tracking freed memory.

Along with first iteration need to cerate a test to print out location of buffer as well as create a method to print out the location of a local variable in the method and recursively call itself to identify how the wasm C++ stack grows and if we will need to worry about it eventually overwriting our buffer.  Also need to identify how big of an initial buffer we can allocate.

brianjohnson5972:
It looks like our block of memory is being allocated early in the memory space (offset 12 is the start, for the create_script test I ran).  The C++ stack seems to start somewhere around 10000 (printing the address of my local variable in my recursive function is offset 9884) and there is nothing preventing either of them from walking over the other.  If I allocate 8K create_script runs fine, anything close to 9884 causes bad behavior.  Similarly if I have my recursive function recurse enough times to overwrite past 8K, no system errors are reported, just system instability.

[...]

PR #400, opened 2017-09-11
Initial Wasm memory allocation

brianjohnson5972:
#129 
Added malloc, realloc, and free to eoslib.  Currently does not free memory, just allocates/reallocates what is available.  Changed user stack to 8K.


tests/slow_tests/memory_test/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+file(GLOB SOURCE_FILES "*.cpp")
+add_wast_target(memory_test "${SOURCE_FILES}" "${CMAKE_SOURCE_DIR}/contracts" ${CMAKE_CURRENT_SOURCE_DIR})
```
wanderingbort: This directive will build the memory test contract code intp `contracts/` however you later attempt to include the `.wast.hpp` file relative to the slow tests source.

we have precedence for storing "test" contracts in the contracts tree that are used in slow tests.  the `memory_test` contract should be placed there (along side `api_test`).  Then you can include the `.wast.hpp` using brackets and it should work just fine.

tests/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ add_executable( chain_test ${UNIT_TESTS} ${COMMON_SOURCES} )
 target_link_libraries( chain_test eos_native_contract eos_chain chainbase eos_utilities eos_egenesis_none wallet_plugin fc ${PLATFORM_SPECIFIC_LIBS} )
 
 if(WASM_TOOLCHAIN)
+  add_subdirectory(slow_tests/memory_test)
```
wanderingbort: This should not be necessary once the memory test contract is moved to the contracts tree.

tests/slow_tests/slow_tests.cpp
```diff
@@ -47,6 +47,7 @@
 #include <currency/currency.wast.hpp>
 #include <exchange/exchange.wast.hpp>
 #include <infinite/infinite.wast.hpp>
+#include "memory_test/memory_test.wast.hpp"
```
wanderingbort: once memory_test is moved to the `contracts/` tree, this should become `#include <memory_test/memory_test.wast.hpp>`

CMakeModules/wasm.cmake
```diff
@@ -121,7 +121,7 @@ macro(add_wast_target target SOURCE_FILES INCLUDE_FOLDERS DESTINATION_FOLDER)
 
   add_custom_command(OUTPUT ${DESTINATION_FOLDER}/${target}.wast
     DEPENDS ${target}.s
-    COMMAND ${BINARYEN_BIN}/s2wasm -o ${DESTINATION_FOLDER}/${target}.wast -s 1024 ${target}.s
+    COMMAND ${BINARYEN_BIN}/s2wasm -o ${DESTINATION_FOLDER}/${target}.wast -s 8192 ${target}.s
```
wanderingbort: We should consider making this something each contract can override.  Setting 1 true stack size for all contracts seems overly limiting.

brianjohnson5972: Yeah I agree, but I'd like @bytemaster 's input, since I am not sure how we want to handle this.  Any wasm *.cpp file that included memory.hpp would need to have it set, or else we need to set the memory space allocated based on the user stack.  So I'll keep this as an open issue for now (while I fix everything else), unless you have more input on this.

tests/CMakeLists.txt
brianjohnson5972: See my previous response.

tests/slow_tests/slow_tests.cpp
brianjohnson5972: See my previous response.

brianjohnson5972:
@wanderingbort I'm fine with moving memory_test in with contracts, if you still want it there, just wanted to indicate why I did it initially and that I think eventually it would be good to have a place for usable or emulatable contracts and a place for test contracts that would cause confusion or would would be examples of what not to do (since they are test) .  But with the short time I will defer to you and move it, just let me know.

wanderingbort:
I don't mean to block this, I agree @bytemaster should weigh in.  given that this falls into the category of "unit tests for contract API", I think it should be with the other unit tests.  If those tests should be under the test/ tree rather than contracts, then I am fine with it.  We can do that move in another issue.

I'm going to change my review to an approval for now and start a new issue to unify the memory tests and existing API tests and determine the correct directory for them.

Issue #393, opened 2017-09-10
p2p network txn handling, new constants, socket options


pmesnier:
Relates to #291, cleans up txn handling, define constants for new magic numbers, add setting the NODELAY socket option.

pmesnier:
I've named several magic numbers in net_plugin.cpp. This adds 9 potentially configurable values to the four existing. I think it would be a good idea to add either section names or prefixes to config options to avoid potential namespace collisions. For instance, change listen-endpoint, public-endpoint, and remote-endpoint to be p2p-listen-endpoint, etc. then the new values when added could be p2p-<option-name>.

In fact, we don't need to do that to all config.ini options, http-server-endpoint is clear, as is plugin and producer-name.

pmesnier:
closing, solution was merged in pr #403.

PR #403, opened 2017-09-11
P2p txn handling 393

pmesnier:
For issue #393, addresses the odd transaction handling, properly enables TCP NoDelay for all sockets and replaces magic numbers with constants. It does not yet define new options for overriding these default values but I have a comment on issue 393 discussing that need.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -47,7 +62,7 @@ namespace eos {
     transaction_id_type id;
     bool                is_known_by_peer = false; ///< true if we sent or received this trx to this peer or received notice from peer
     bool                is_noticed_to_peer = false; ///< have we sent peer notice we know it (true if we receive from this peer)
-    uint32_t            block_num = -1; ///< the block number the transaction was included in
+    UInt16              block_num = -1; ///< the block number the transaction was included in
```
bytemaster: Why is this 16 bit?

pmesnier: because that's what was in the generated Transaction  type.  See 
libraries/types/include/.../generated.hpp

PR #422, opened 2017-09-13
help messaging and transaction options to reduce duplicate transaction errors

### LINKED ISSUE #384: Duplicate transaction error
When I attempt to do the following twice it throws up a duplicate transaction error:

./eosc --wallet-port 7777 transfer inita currency 10000

If I run the same command with a different amount it works fine but only once.

./eosc --wallet-port 7777 transfer inita currency 10001

The error output is as below:

2274982ms thread-0   main.cpp:768                  main                 ] Failed with error: 10 assert_exception: Assert Exception
status_code == 200: Error
: 3030011 tx_duplicate: duplicate transaction
Transaction is not unique
    {}
    thread-0  chain_controller.cpp:828 validate_uniqueness

wanderingbort:
closes #384

- [x] more informative error messages
  - [x] client side system for translating non-human-friendly error spew
  - [x] optional `-v,--verbose` option to get back the error spew of olde
- [x] standard transaction options composable on any sub-command
  - [x] expiration support `-x,--expiration[=]<milliseconds til expiration>`
  - [x] force unique transaction support `-f,--force-unique`
    - uses the transfer memo when it can
    - uses a superfluous message to `eos` with type `nonce` otherwise
    - do we want to just pad out the last messages payload bytes instead?
- [x] BONUS: human readable error messaging for:
  - permissions you dont have the keys for
  - referring to accounts that don't exist
  - using `push message` for an unknown ABI
  - opening an unknown wallet
  - incorrect password when unlocking wallet
  - trying to access a locked wallet

### DIFF
--- a/programs/eosc/CMakeLists.txt
+++ b/programs/eosc/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_executable( eosc main.cpp httpc.cpp )
+add_executable( eosc main.cpp httpc.cpp help_text.cpp )
 if( UNIX AND NOT APPLE )
   set(rt_library rt )
 endif()
--- a/programs/eosc/help_text.cpp
+++ b/programs/eosc/help_text.cpp
@@ -1,0 +1,129 @@
+/*
+ * Copyright (c) 2017, Respective Authors.
+ *
+ * The MIT License
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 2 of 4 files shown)


wanderingbort:
examples of the new error messaging:
```
$ ./eosc transfer inita user1 100 "The Memo"
An error occurred while submitting the transaction for this command!

The transaction references an account which does not exist.
Unknown accounts:
  - user1

Please check the account names and try again!   

$ ./eosc wallet open -n foo
Unable to find a wallet named "foo", are you sure you typed the name correctly?

$ ./eosc wallet unlock
password: <obviously bad password>
Invalid password for wallet named "default"

$ ./eosc transfer inita user1 100 "The Memo"
An error occurred while submitting the transaction for this command!

The transaction requires permissions that could not be authorized by the wallet.
Missing authrizations:
  - inita@active

Please make sure the proper keys are imported into an unlocked wallet and try again!    

```

bytemaster:
This looks amazing, your bonus efforts also look good except it may invalidate one of the steps in my tutorial!  Need to check to see if I can use push message with hex data or if your new checks cause that to fail.

Issue #333, opened 2017-09-05
Wallet API restricted to localhost


toonsevrin:
This is not fun for a dockerized environment, can this check be made ignorable?
https://github.com/EOSIO/eos/blob/master/plugins/wallet_api_plugin/wallet_api_plugin.cpp#L103

heifner:
@bytemaster Do you want to remove the localhost restriction on eos-walletd?

heifner:
@bytemaster Indicated that this restriction can be removed and replaced with a warning.

heifner:
So I'm thinking the warning should be:
```c++
              "*************************************\n"
              "*                                   *\n"
              "*  --   Wallet NOT on localhost  -- *\n"
              "*  - Password and/or Private Keys - *\n"
              "*  - are transferred unencrypted. - *\n"
              "*                                   *\n"
              "*************************************\n");
```

Will put this small change in as a PR tomorrow morning.

Issue #334, opened 2017-09-05
walletd for docker


toonsevrin:
docker-compose should automatically setup a walletd instance.

toonsevrin:
working on this over at https://github.com/toonsevrin/eos/tree/patch-4

toonsevrin:
@peterwillcn

heifner:
Is this still desired? Can we close this issue?

noprom:
should be closed now.

heifner:
Wallet added to docker a long time ago.

PR #465, opened 2017-09-15
Remove localhost restriction #333

heifner:
Issue #333.

Removed localhost restriction and replaced it with a warning.


dskvr:
Resolves #333 
Resolves #334

dskvr:
This resolves #334 as well

PR #468, opened 2017-09-15
localization support for eosc

### LINKED ISSUE #419: Add localization support to eosc
while working on help text for #384 it became obvious that there needs to be support for localization in our developer/consumer facing tools.  

This should include the ability to localize:
- [x] all options/subcommand descriptions
- [x] all usage information
- [x] any help text displayed on errors

I don't think we need to localize the raw errors off the wire so long as we create more descriptive help messages which are localized.

wanderingbort:
This adds support for localization via `boost::locale` to `eosc`. Translations are specified in the GNU gettext format (`.po` and `.mo` files). For now, localizations are stored local to the binary but conceivably they can be stored in the system's default localization directory allowing orthogonal updates to the translation dictionaries.  

For now, only en_US is provided and if `eosc` detects that there is no available dictionary it will default to en_US so that our pre-translated identifiers don't come pouring out. 

NB: this is a "quick" extraction of the strings.  No effort was made to consolidate similar strings. 

a `.pot` file is provided for translators to use when generating a new localization.   There are not many comments in the `.pot` file so, translators will have to refer to the en_US translation to understand what templating parameters are coming in to each translated string. 

Lastly, while there is a naming convention implied by the quick extraction of strings, there is no requirement that it be held to for new strings.  However, we should probably at least have a plan lest the poor translators have heart attacks trying to localize. 

closes #419


wanderingbort:
I've added some "not completely accurate" code to get the `locale` directory that is adjacent to the executable.  As it is dependent on `arg[0]` being correct, it is easily fooled.  The "real" way to allow local localizations would rely on OS specific calls to determine the path to the "real" executing module/binary.   

I don't like this solution but, for now it works.   We should probably endeavor to install these localizations in an appropriate system wide path on unix and mac AND use proper methods of determining local localizations at some point.

bytemaster:
Does this have a sane fall back if the files are missing?  I would hate the merge something that causes default use to break more often.

wanderingbort:
we can implement "sane" fall-back for the case where even the `en-US` files are missing by using the same strings for the ID as we do for the text of the translation.  Considering some of the strings are multi-line and interpolated with fc's string formatting, I didn't use that convention.

I could see if it would be possible to "statically" embed at least the `en-US` translation so that the binary always has at least that.

bytemaster:
@wanderingbort yes, please embed the en-US statically and resolve the conflicts and then I will approve.

PR #486, opened 2017-09-21
use clang/clang++ to build EOS in build.sh

elmato:


elmato:
@pavybez can you verify if ``clang++-4.0`` and ``clang-4.0`` are in the PATH after installing llvm@4 on OSX?

Issue #493, opened 2017-09-25
Extend db C++ API to define tables with string keys.


pavybez:
We now support storage of Strings as keys in database. But db API lacks utilities to define tables using strings as primary and secondary indices.

pavybez:
Pull request https://github.com/EOSIO/eos/pull/501 in review.

brianjohnson5972:
This was against the old table designe and it looks like this was initially completed  but not marked as complete.   Current string as secondary index deficiency is addressed in #2402

Issue #492, opened 2017-09-25
Extend message API to read variable length strings.


pavybez:
At the moment, eos::currentMessage<T> only supports reading structures with fixed length types. This should be extended to support reading structures with String types now that we support Strings in ABI and database.

pavybez:
For example, the "String" is prefixed with length followed by data. But it is not easy to tell how many bytes represent length field. We should have utilities to unpack these.

heifner:
variable length `string`s are supported in actions now.

PR #494, opened 2017-09-25
Storage

pavybez:
A first cut implementation of storage contract. Does not do anything yet. Still needs implementation of reserving tokens for storage and freeing tokens when storage is released.

Also methods to read structures from message are yet to be implemented. Issues https://github.com/EOSIO/eos/issues/493 and https://github.com/EOSIO/eos/issues/492

### DIFF
--- a/contracts/storage/storage.cpp
+++ b/contracts/storage/storage.cpp
@@ -1,0 +1,130 @@
+#include "storage.hpp"
+
+namespace TOKEN_NAME {
+   void storeAccount( AccountName name, const Account& account ) {
+      if ( account.isEmpty() ) {
+         Accounts::remove( account, name );
+      } else {
+         Accounts::store(account, name );
+      } 
+   }
+
+   void apply_storage_transfer( const TOKEN_NAME::Transfer& transfer ) {
+      eos::requireNotice( transfer.to, transfer.from );
+      eos::requireAuth( transfer.from );
+
+      Account from = getAccount( transfer.from );
+      Account to   = getAccount( transfer.to );
+
+      from.balance -= transfer.quantity; /// token subtraction has underflow assertion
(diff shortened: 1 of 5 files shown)


contracts/storage/storage.cpp
```diff
@@ -0,0 +1,130 @@
...
+namespace TOKEN_NAME {
+   void storeAccount( AccountName name, const Account& account ) {
+      if ( account.isEmpty() ) {
+         Accounts::remove( account, name );
+      } else {
+         Accounts::store(account, name );
+      } 
+   }
+
+   void apply_storage_transfer( const TOKEN_NAME::Transfer& transfer ) {
```
bytemaster: storage doesn't implement a currency...

contracts/storage/storage.cpp
```diff
@@ -0,0 +1,130 @@
...
+extern "C" {
+    void init()  {
+       // How do we initialize the storage capacity? By how much here?
+       Accounts::store( Account( StorageTokens(1000ll*1000ll*1000ll) ), N(storage) );
+    }
+
+    /// The apply method implements the dispatch of events to this contract
+    void apply( uint64_t code, uint64_t action ) {
+       if( code == N(storage) ) {
+          if( action == N(transfer) ) {
```
bytemaster: this should be code == EOS and action == transfer... apply_eos_transfer(...)

contracts/storage/storage.hpp
```diff
@@ -0,0 +1,170 @@
...
+
+        /**
+        * Account to transfer to
+        */
+        AccountName to;
+
+        /**
+        * quantity to transfer
+        */
+        StorageTokens quantity;
```
bytemaster: We don't create StorageTokens and storage, instead we have storage quota and storage consumed and its measured in bytes.

PR #501, opened 2017-09-26
Updated db C++ API for string type primary key

pavybez:
Updated db C++ API with new implementation for defining VarTable that can take String as primary key or any other object in future.

### DIFF
--- a/contracts/eoslib/db.hpp
+++ b/contracts/eoslib/db.hpp
@@ -7,8 +7,105 @@
 /**
  *  @ingroup databaseCpp
  *  Cpp implementation of database API. It is based on pimpl idiom.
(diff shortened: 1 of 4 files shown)


contracts/eoslib/db.hpp
```diff
@@ -7,9 +7,106 @@
...
  *  Cpp implementation of database API. It is based on pimpl idiom.
- *  @see Table class and table_impl class
+ *  @see Table class and table_impl and table_impl_str class
  */
 
+template<typename T>
+struct table_impl_obj {};
+
+template<>
+struct table_impl_obj<char*> {
```
bytemaster: We should define a string class and not use char* as strings in c++ apis.

Issue #498, opened 2017-09-25
EOS build failure on Ubuntu 16.10


signitary:
Hello,

I had to locate the clang-4.0 and lldb-4.0 sources for Yakkety since they were not available from the website. That solved that failure, but now I'm getting the following after all previous dependencies were downloaded and installed:

[ 63%] Building CXX object programs/launcher/CMakeFiles/launcher.dir/main.cpp.o
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata+0x70): undefined reference to `typeinfo for llvm::JITSymbolResolver'
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata._ZTIN7LLVMJIT17UnitMemoryManagerE[_ZTIN7LLVMJIT17UnitMemoryManagerE]+0x10): undefined reference to `typeinfo for llvm::RTDyldMemoryManager'
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata._ZTIN4llvm18MCJITMemoryManagerE[_ZTIN4llvm18MCJITMemoryManagerE]+0x10): undefined reference to `typeinfo for llvm::RuntimeDyld::MemoryManager'
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata._ZTIN4llvm17raw_pwrite_streamE[_ZTIN4llvm17raw_pwrite_streamE]+0x10): undefined reference to `typeinfo for llvm::raw_ostream'
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata._ZTIN4llvm18ObjectMemoryBufferE[_ZTIN4llvm18ObjectMemoryBufferE]+0x10): undefined reference to `typeinfo for llvm::MemoryBuffer'
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata._ZTIN4llvm9ErrorInfoINS_9ErrorListENS_13ErrorInfoBaseEEE[_ZTIN4llvm9ErrorInfoINS_9ErrorListENS_13ErrorInfoBaseEEE]+0x10): undefined reference to `typeinfo for llvm::ErrorInfoBase'
../Runtime/libRuntime.a(LLVMJIT.cpp.o):(.rodata+0x70): undefined reference to `typeinfo for llvm::JITSymbolResolver'
[...]

PR #509, opened 2017-09-27
Add curl to dependency list

andriantolie:
Xenial (Ubuntu 16.04) image that is downloaded from the official ubuntu website doesn't have curl installed by default


andriantolie:
Related to issue https://github.com/EOSIO/eos/issues/498

Issue #504, opened 2017-09-26
Implement function for adding and removing WASM memory


brianjohnson5972:
Implement logic similar to Emscripten's sbrk function in wasm_interface.  Will need to impose a max size of 1M.  Also need to perform checktime after memory is grown or shrunk to ensure time allotment isn't exceeded.  Function will return the offset of the new memory that can be used for the  new buffer.  This will be used by #129 to allocate new pages when existing buffers are exhausted.

brianjohnson5972:
Logic is written and now writing unit tests.

brianjohnson5972:
@bytemaster for the emscrypten sbrk function the return value is useful when you pass in a positive num_bytes to allocate, since it will point to the previous number of bytes, which would be the offset of the start of the new block of memory, but I'm not sure how a user could use the return value if a negative num_bytes is passed, since it will be returning the offset of the end of the buffer that is to be removed.  So I am thinking that the eoslib provided function will only take positive num_bytes and only the code that returns the MemoryInstance to its original size (a single 64K page) will be able to shrink the memory.  Does that make sense?

PR #514, opened 2017-09-27
WASM Add Memory Pages

brianjohnson5972:
#504 
Added sbrk to intrinsic functions to allow new page memory to be added in WASM.  All pages added during execution of WASM will be removed when WASM execution is completed.


brianjohnson5972:
I kept the emscripten function name, sbrk, since I didn't know what would be a good name to indicate what it was used for and figured it could just be discussed when the pull request was being reviewed.

PR #517, opened 2017-09-28
419 localization support redux

### LINKED ISSUE #419: Add localization support to eosc

wanderingbort:
This is a re-do for localization and closes #419 

The solution is still based on GNU gettext (which is the underlying translation format for `boost::locale`) however this is a simpler integration directly with that tooling.  

In addition, the translation for `en_US` is embedded in the binary itself leaving no need to produce any translation files for a standard build.  Instead, optional translations can be downloaded/packaged separately and installed in the standard location for the given OS/distro (eg `{installation prefix}/share/locale` ).   There is a `.pot` file that will need to be updated as new translatable strings are added, and sent to translators to process.

### DIFF
--- a/programs/eosc/CMakeLists.txt
+++ b/programs/eosc/CMakeLists.txt
@@ -9,8 +9,16 @@ if( GPERFTOOLS_FOUND )
     list( APPEND PLATFORM_SPECIFIC_LIBS tcmalloc )
 endif()
 
+find_package(Intl)
+
+set(LOCALEDIR ${CMAKE_INSTALL_PREFIX}/share/locale)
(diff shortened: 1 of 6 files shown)


testzcrypto:
I like this way!

Early translators can request languages and start working on translation here: https://crowdin.com/project/eosz
I will make translation pull request when this pull request will be merged.

PR #532, opened 2017-10-02
Add Page Memory to malloc/realloc and add free logic

brianjohnson5972:
#129 
Refactored to add 64K page memory as new heaps for allocating.  If new page memory is adjacent to an existing heap, then the heap is expanded to limit fragmentation of the memory.

Free logic was also added.  Free memory is just tracked by using the highest bit in the ptr header.  So free processing costs very little at time of calling free at the cost of searching when the freed memory is required (after 1M of memory has been allocated).  Also, this implementation helps to not have reserve extra memory to maintain a separate list and the logic to manage those list (like allocating it at the end of a heap and growing backwards and preventing memory collisions).

### DIFF
--- a/contracts/eoslib/memory.hpp
+++ b/contracts/eoslib/memory.hpp
@@ -34,6 +34,12 @@ namespace eos {
    private:
       void* malloc(uint32_t size)
       {
+         if (size == 0)
+            return nullptr;
+
+         eos::print("malloc ", size, "\n");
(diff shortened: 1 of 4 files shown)


bytemaster:
Please resolve conflicts.

brianjohnson5972:
Conflict is resolved

bytemaster:
Conflicts created again.... please resolve.

Issue #524, opened 2017-09-29
Different Time Based Execution Checks for Production& Validation


bytemaster:
When building a block we need to have stricter validation than when validating a block previously built or slower nodes might reject valid blocks. 

Building a block includes:
1. push_transaction
2. produce_block 

Validating a block occurs when we receive a block from someone else, in this case we assume that our machine is 4x slower and therefore give transactions 4x the runtime before we consider rejecting the block.

- [ ] add command line argument to change runtime length checks on push_block 
- [ ] add command line argument to change default runtime length checks for push_transaction
- [ ] add command line argument to set runtime length checks on produce_block 

These settings will be helpful for producers to tune their filtering without risking unnecessary forks.

brianjohnson5972:
Putting in plumbing for configuring and passing different execution times.

Received block transaction exectuion time and transaction execution time will be configured in chain_plugin,  create block transaction execution time will be configured in producer_plugin.
Assuming initialize and replay will use the received block execution time.
Not currently giving control to set execution time for init (i.e. setcode), it is still set to the original values(since it wasn't mentioned here).

Should be pushing shortly.

PR #540, opened 2017-10-04
Different, configurable contract execution times

brianjohnson5972:
#524 

Added 3 configuration parameters:

trans-exectution-time - in chain_plugin, used for push_transaction and accept_transaction

rcvd-block-trans-exectution-time - in chain_plugin, used for push_block, accept_block, and chain_controller initialization of block and replay

create-block-trans-execution-time - in producer_plugin for maybe_produce_block

wasm_interface::init still uses fixed value - (figured you would comment in this review if that should have its own config value)

Made default values to be used in unit tests.


bytemaster:
I had envisioned these would be properties on the controller and not something passed into every push()/apply() call.

I would have used the skip_flags (which shouldn't really be called skip flags) to indicate the "mode" and then based on the mode the WASM interface can pick the proper runtime.

brianjohnson5972:
I will put all the different execution times in chain_plugin and pass into chain_controller.

brianjohnson5972:
@bytemaster If I use the skip flags, the issue is that I have 3 states to convey in a flag, so I will have the ambiguity of being able to set 3 flags, where only one of them will be used.

Also, I'm assuming that you want _skip_flags used until wasm_interface needs to be called.

I was passing the time to address the eventual issue that we will have once we add multi-threading, but I guess we will need to address the same issue with _skip_flags anyways.

brianjohnson5972:
Another ambiguity with the flags is that when we get to calling wasm_interface::get().apply(...), at least one of those flags needs to be set, so I guess I will default to the shortest duration if one isn't set.

brianjohnson5972:
Made requested changes and squashed commits.

PR #541, opened 2017-10-04
added skeleton to CMakeLists.txt

ioleksiu:


brianjohnson5972:
You need to have a CMakeLists.txt in contracts/skeleton, for this cmake instruction to work, did you forget to commit that?

PR #542, opened 2017-10-04
Update include

ioleksiu:
#include "skeleton.hpp"


bytemaster:
Why isn't "skeleton.hpp" in the include path and therefore available via <skeleton.hpp> ?

brianjohnson5972:
it should be <skeleton/skeleton.hpp>

PR #556, opened 2017-10-11
P2p authenticate peer on handshake 505

jgiszczak:


plugins/net_plugin/include/eos/net_plugin/protocol.hpp
```diff
@@ -11,6 +11,9 @@ namespace eos {
       int16_t         network_version = 0;
       chain_id_type   chain_id; ///< used to identify chain
       fc::sha256      node_id; ///< used to identify peers and prevent self-connect
+      Name            producer; ///< first producer name on this node
+      fc::sha256      token; ///< digest to prove we own the private key of the producer named above
```
bytemaster: This system is subject to replay attack, you must perform a ECDH key exchange and sign the shared secret.

plugins/net_plugin/include/eos/net_plugin/protocol.hpp
```diff
@@ -11,6 +11,9 @@ namespace eos {
       int16_t         network_version = 0;
       chain_id_type   chain_id; ///< used to identify chain
       fc::sha256      node_id; ///< used to identify peers and prevent self-connect
+      Name            producer; ///< first producer name on this node
```
bytemaster: Rather than authenticating by name of producer, authenticate by public key.  We may want to authenticate non-producing nodes.

jgiszczak:
Now supports authenticating producers, non-producing peers, and/or not authenticating.  Authentication now occurs by public key rather than producer name.  Protocol extended to eliminate replay attack: packet now contains a PGP-style message, hash, and hash signature, not just hash and hash signature.  Only the peer configured to authenticate will check the handshake.  If both sides are configured to authenticate, the connection is validated from both sides.  ECDH shared secrets could not be used while supporting non-authenticated peers.

plugins/net_plugin/include/eos/net_plugin/protocol.hpp
```diff
@@ -7,23 +7,26 @@ namespace eos {
...
-
+   static_assert(sizeof(std::chrono::system_clock::duration::rep) >= 8, "system_clock is expected to be at least 64 bits");
    typedef std::chrono::system_clock::duration::rep tstamp;
-   typedef int32_t                                  tdist;
 
-   static_assert(sizeof(std::chrono::system_clock::duration::rep) >= 8, "system_clock is expected to be at least 64 bits");
+   struct handshake_message {
+      int16_t                    network_version = 0;
+      chain_id_type              chain_id; ///< used to identify chain
+      fc::sha256                 node_id; ///< used to identify peers and prevent self-connect
```
bytemaster: peers should be identified by their authenticated public key even if randomly generated.

jgiszczak: Sorry, you'll need to be a little less terse than that.  I don't know what that means.

jgiszczak: Per Dan, node_id is to be removed at a later date.  I left it in in order to ease the merge with Phil's ongoing sync work.  Authentication makes it redundant.

heifner:
@bytemaster Can you review to verify this now meets your requirements.

jgiszczak:
Corrected default network shape back to "star" from "ring".  That was a merge error.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2017,11 +2074,97 @@ namespace eosio {
...
+          elog( "Peer ${peer} sent a handshake with an unauthorized key: ${key}.",
+                ("peer", msg.p2p_address)("key", msg.key));
+          return false;
+        }
+      }
+
+      namespace sc = std::chrono;
+      sc::system_clock::duration msg_time(msg.time);
+      auto time = sc::system_clock::now().time_since_epoch();
+      if(time - msg_time > peer_authentication_interval) {
```
alice-bob-dlfs: What about the case of `msg_time - time > peer_authentication_interval`?

Issue #578, opened 2017-10-19
cannot include memory.hpp twice


pavybez:
memory.hpp defines a variable memory_heap which causes linker issues due to multiple definitions if this file in included in more than one cpp file. This prevents us from using string.hpp (which internally uses memory.hpp). Problem does not occur in smart contracts if we have only one source file but test_api does not link anymore because it has multiple sources.

bytemaster:
My understanding is that we would not build multiple .o files from multiple cpp files.  Instead, all CPP files for a contract should be built as a single compilation unit. If developers want to divide up their cpp files into "libraries" then it would look like this:

```
/** @file library.cpp */
#include <memory.cpp>
#include <contract.cpp>
 ```

Then you would simply compile library.cpp and be done with it.

Smart contracts shouldn't be all that big where multiple compilation units are required.

brianjohnson5972:
If you look at api_test, there are multiple cpp files that end up creating
.bc files, which feed into one .wast file.

bytemaster:
yeah, we need to make sure there is only one .bc

brianjohnson5972:
Well my fix is fine too be in the code, but we should make a new task to
change was.cmake to make only one .bc file

On Oct 19, 2017 4:50 PM, "Daniel Larimer" <notifications@github.com> wrote:

yeah, we need to make sure there is only one .bc

—
You are receiving this because you were assigned.

Reply to this email directly, view it on GitHub
<https://github.com/EOSIO/eos/issues/578#issuecomment-338048056>, or mute
the thread

brianjohnson5972:
Is this necessary for the STAT release? If it is, is it a nice to have, or
is it a must have?

brianjohnson5972:
I should have looked more closely before I responded above.  So each *.cpp file is compiled to bitcode in a *.bc file, then all of the *.bc files are linked into a <target>.bc file (which for most of our contracts ends up compiling <target>.cpp to <target>.bc, then linking <target>.bc and overwriting it)

@bytemaster Is this what you were thinking should happen?

heifner:
I took some time to look into this and I could not find a way to create a true global in wasm via c++. Extern doesn't work, static local in a function doesn't work, etc. However, if we go with Dan's recommendation above:

> My understanding is that we would not build multiple .o files from multiple cpp files. Instead, all CPP files for a contract should be built as a single compilation unit.

[...]

PR #580, opened 2017-10-19
Fixed error of not being able to include memory.hpp in multiple cpp files

brianjohnson5972:
#578 
Defined memory_heap as static.


pavybez:
Wouldn't it still create one copy per compilation unit? Is that what is intended?

bytemaster:
Declaring static will create multiple incompatible heaps.

brianjohnson5972:
Changed to only compile the ${target}.cpp and require any subsequent cpp files be included in ${target}.cpp.  @thomasbcox  this affects documentation for contract creation.

PR #581, opened 2017-10-20
add validation for utf_8 string

andriantolie:
Based on definition of UTF-8 in https://en.wikipedia.org/wiki/UTF-8

### DIFF
--- a/contracts/eoslib/string.h
+++ b/contracts/eoslib/string.h
@@ -1,0 +1,10 @@
+/**
+ *  @file
+ *  @copyright defined in eos/LICENSE.txt
+ */
+#pragma once
+#include <eoslib/types.h>
+
+extern "C" {
+    uint32_t is_valid_utf8_str( const char* str, uint32_t len);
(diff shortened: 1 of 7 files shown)


contracts/eoslib/string.h
```diff
@@ -0,0 +1,10 @@
+/**
+ *  @file
+ *  @copyright defined in eos/LICENSE.txt
+ */
+#pragma once
+#include <eoslib/types.h>
+
+extern "C" {
+    uint32_t is_valid_utf8_str( const char* str, uint32_t len);
```
bytemaster: I prefer for this to be "assert_is_utf8" and have it throw an exception on failure. I cannot think of a reason why someone would need "if not utf8 then do..." type logic and any "assert..." type APIs can be optimized out during replay.

andriantolie: Right, I think this makes more sense.

andriantolie:
Updated

PR #592, opened 2017-10-26
Initial version of fixedpoint template for 32,64,128 bits

dhaneshvb:


contracts/eoslib/fixedpoint.hpp
```diff
@@ -0,0 +1,565 @@
...
+    * Example:
+    * @code
+    * fixed_point128<6> a(123232.455667233)
+    * fixed_point128<0> a(123424)
+    * fixed_point128<18> c = a*b;
+    * fixed_point128<24> d = a+b+c;
+    * fixed_point128<24> e = b/a;
+    * @endcode
+    */
+    template <uint32_t q>
```
bytemaster: q, as a template parameter, should be capital and its meaning should be documented.

bytemaster: Also it should be uint8_t and there should be a static_assert that it is less than 128

PR #599, opened 2017-10-27
The error log is transaction check.

Ben002:


Ben002:
Is it wrong?

heifner:
Could be either.

PR #604, opened 2017-10-30
Update group__eosc.html

yuelipeng:
missing keyword of "wallet"


heifner:
This is a generated file from eosc.dox.

PR #606, opened 2017-10-30
next_pending_message_serial is not initialized in apply_context

learnforpractice:


learnforpractice:
I think it will be better to initialize next_pending_transaction_serial to 1 to avoid a 0 value or change
   pending_message::handle_type handle = next_pending_message_serial++;
to
   pending_message::handle_type handle = ++next_pending_message_serial;
since next_pending_message_serial is used as a message handle in smart contracts.

Can anyone give a comment about this.

PR #612, opened 2017-10-31
Add string util to convert int to string

andriantolie:
Add utility function to convert int to string. One use case that will be useful for this is for hashing an integer in a contract (since sha256 takes string as input). 

STAT-112


bytemaster:
sha256 doesn't take a string as input, it takes a binary buffer as input.

Generally contracts should not be doing string to int operations, and it certainly doesn't belong as part of the intrinsic API.

PR #617, opened 2017-11-01
STAT-64 Chain to mongo 172

heifner:
**STAT-64 - Current implementation:**

* chain_controller
  * Added {{signal<void(const signed_block&)> applied_irreversible_block}}

* db_plugin
  * Adds to mongodb
    * Schema: https://docs.google.com/document/d/1E6BQVX1E8jsCzJIpFWV_B3hfA7Br6iiFlUiWZJpJ3zM/edit?ts=59d29be1#heading=h.sgz9di1jhkvk
    * Block, Transaction, Message
    * Account eos_balance, abi

**Future implementation:**

* eosd running in delay mode so that it only processes irreversible blocks
  * Options:
    1. eosd setup so that it only pushes block from P2P to chain_controller when the block is irreversible
       * eosd would need to cache blocks until the irreversible block num indicates they can be pushed to chain_controller
       * eosd would be in a state of constant "syncing"
       * chainbase would be set up to push to mongodb
       * Simpler configuration since only one special eosd required
    2. a trusted eosd setup so that it only broadcasts irreversible blocks
       * delayed eosd connected to trusted eosd would then simply run as normal
       * the delayed eosd would be set up to push to mongodb

* Add hook to chainbase so that commit pushes into mongodb
  * Either a hook given to chainbase OR a replacement of chainbase


libraries/chain/chain_controller.cpp
```diff
@@ -1224,15 +1227,25 @@ void chain_controller::initialize_chain(chain_initializer_interface& starter)
          for (int i = 0; i < 0x10000; i++)
             _db.create<block_summary_object>([&](block_summary_object&) {});
 
+         signed_block block{};
```
brianjohnson5972: Please add a comment to indicate what you are doing here.

libraries/chain/chain_controller.cpp
```diff
@@ -1273,6 +1290,12 @@ chain_controller::~chain_controller() {
 void chain_controller::replay() {
    ilog("Replaying blockchain");
    auto start = fc::time_point::now();
+
+   auto on_exit = fc::make_scoped_exit([this](){
```
brianjohnson5972: Limit lambda capturing to just one needed parameter.

heifner: Done.

libraries/chain/chain_controller.cpp
heifner: Added comment:
`         // create a dummy block and cycle for our dummy transactions to send to applied_irreversible_block below
`

Issue #614, opened 2017-10-31
STAT-84 ⁃ Add better error handling/messages to push transaction


blockone-syncclient:
Currently, the error messages returned from push transaction are not proper JSON. The error messages need to be improved, error codes added, and JSON structure defined.

Work with the WebUI team to determine best user experience.

blockone-syncclient:
➤ Kevin Heifner commented:

Outline of better error generation at the end of this document.
https://docs.google.com/document/d/1E6BQVX1E8jsCzJIpFWV_B3hfA7Br6iiFlUiWZJpJ3zM/edit#

Need to improve http_plugin with correct error codes and formatted json error responds. Update eosc and eosd_run_test.sh as needed.

blockone-syncclient:
➤ Jonathan Giszczak commented:

Incorporated response code 202 into Chain API as appropriate.  Still need to revise wallet and account history plugins as needed.

blockone-syncclient:
➤ Jonathan Giszczak commented:

Incorporated response code 201 into Wallet API as appropriate.  Account history plugin required no changes.

heifner:
Resolved by #628.

blockone-syncclient:
➤ Joshua Lavin commented:

There is no test plan on this issue; cannot be tested.

blockone-syncclient:
➤ Andrianto Lie commented:

Checked for the following HTTP Status Code

  * 200 OK
  * 202 Accepted
  * 400 Bad Request
  * 401 Unauthorized
  * 404 Not Found
  * 409 Conflict
  * 500 Internal Service Error

That their response's header is `Content-type: application/json` and has valid JSON response in form of

{
    "code": 404, 
    "message": "Not found"
    “details”: “a more detailed string “
}

blockone-syncclient:
➤ Andrianto Lie commented:

Details:

200 OK
Command:

 curl  -v  http://localhost:8888/v1/chain/get_block -d '{"block_num_or_id":1}'Result:

*   Trying ::1...
* TCP_NODELAY set
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8888 (#0)
> POST /v1/chain/get_block HTTP/1.1
> Host: localhost:8888
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 21
> Content-Type: application/x-www-form-urlencoded
>
* upload completely sent off: 21 out of 21 bytes
[...]

PR #628, opened 2017-11-02
Http error handling 614

jgiszczak:
Http errors are now application/json, with Content-type header.  Response codes have been enhanced to better match typical RESTful APIs.


heifner:
Resolves #614

PR #646, opened 2017-11-07
Use local loggers in net_plugin classes and enable cli logging level.

jgiszczak:
Created --log-level-net-plugin command line option with possible values corresponding to the logging levels in `fc::logger`.  Net_plugin logging can now be configured without rebuilding.


heifner:
STAT-138 Github #291

PR #649, opened 2017-11-08
 Formatting

gguoss:


heifner:
Thank you, but I already have this fix in another large coding style change that is in work.

PR #650, opened 2017-11-08
fix for corner condition of last irreversible block number 0

pmesnier:
Here's the fix for the sync stink. Turns out there were 2 errors related to the corner case of a new producer trying to synch while the last irreversible block is 0. First, when trying to initialize a "pending" block chain from LIB to current head, trying to fetch the block id for block 0, an exception was tossed, which caused a second lookup to be skipped. 

The other error was that when processing the handshake message, I was excluding a catch up step if the peer's head block was also 0,  fixing these make syncing work like butter.

I also added a startup delay argument to the launcher command line and refactored the p2p sync test to forward a value to that. 

As a final manual test I started 10 producers with empty block data directories and slowly started them in a random order that included a couple that had no direct connections between them. They all synched up just fine.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1597,7 +1597,7 @@ namespace eos {
     }
     bool accepted = false;
     // dlog ("last irrevesible block = ${lib}", ("lib", cc.last_irreversible_block_num()));
-    if( !syncing || num == cc.head_block_num()+1 ){ //  || num > cc.last_irreversible_block_num()) {
+    if( !syncing || num == cc.head_block_num()+1 ) {
       try {
         chain_plug->accept_block(msg, syncing);
         accepted = true;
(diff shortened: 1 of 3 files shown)


heifner:
@pmesnier Please resolve merge conflicts.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1219,7 +1220,7 @@ namespace eos {
       if( msg.last_irreversible_block_num  > head || sync_master->syncing() ) {
         sync_master->start_sync( c, peer_lib );
       }
-      else if( msg.head_id != head_id && msg.head_id != block_id_type( )) {
+      else if( msg.head_id != head_id ) { // && msg.head_id != block_id_type( )) {
```
heifner: Remove the commented out code.

pmesnier:
I'm abandoning this PR. I'll submit a new one deconflicted, and properly identified shortly.

Issue #651, opened 2017-11-08
STAT-144 ⁃ eosd sync failure in some cases


blockone-syncclient:
The problem is manifest when a multinode testnet is started manually. This induces a delay wherein one instance of eosd will create blocks the best it can, but  not advancing the last irreversible block. Then sometime later a new eosd is started, which should add more producers and should also synchronize with the existing producers, but this sometimes doesn't happen.

The actual problem is a corner case where the last irreversible block number is 0. In this case, the chain controller throws an exception when you try to get the block id for block number 0. In the p2p code, this exception is caught, but it caused another block id request to be skipped, leading to the observed failure. 

A second problem is that the handshake handling code was skipping the update notice if the peer's reported head block had an id value of all 0.

heifner:
Resolved by #652.

blockone-syncclient:
➤ Bezwada Satyapravin commented:

How do we test this tomorrow? Is there an ACT?

PR #652, opened 2017-11-08
fix for corner condition of last irreversible block number 0 #651

pmesnier:
This is the same change as #650 but applied to the current master so as to avoid a massive rebase effort. Here's the original comment, which still applies.

Here's the fix for the sync stink. Turns out there were 2 errors related to the corner case of a new producer trying to synch while the last irreversible block is 0. First, when trying to initialize a "pending" block chain from LIB to current head, trying to fetch the block id for block 0, an exception was tossed, which caused a second lookup to be skipped.

The other error was that when processing the handshake message, I was excluding a catch up step if the peer's head block was also 0, fixing these make syncing work like butter.

I also added a startup delay argument to the launcher command line and refactored the p2p sync test to forward a value to that.

As a final manual test I started 10 producers with empty block data directories and slowly started them in a random order that included a couple that had no direct connections between them. They all synched up just fine.

### DIFF
--- a/tests/p2p_tests/sync/test.sh
+++ b/tests/p2p_tests/sync/test.sh
@@ -3,20 +3,43 @@
 pnodes=10
 npnodes=0
 topo=star
-if [ -n "$1" ]; then
-    pnodes=$1
-    if [ -n "$2" ]; then
-        topo=$2
-        if [ -n "$3" ]; then
-            npnodes=$3
+delay=0
+
+args=`getopt p:n:t:d: $*`
+if [ $? == 0 ]; then
+
+    set -- $args
+    for i; do
+        case "$i"
+        in
+            -p) pnodes=$2;
+                shift; shift;;
+            -n) nodes=$2;
+                shift; shift;;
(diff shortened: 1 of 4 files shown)


tests/p2p_tests/sync/test.sh
```diff
@@ -3,20 +3,43 @@
 pnodes=10
 npnodes=0
 topo=star
-if [ -n "$1" ]; then
-    pnodes=$1
-    if [ -n "$2" ]; then
-        topo=$2
-        if [ -n "$3" ]; then
-            npnodes=$3
+delay=0
```
heifner: Should we test with a delay > 0 ?

pmesnier: Sure. I can add a variation to the travis script that does something like: test.sh -p 2 -d 10 
to start two producer instances with a 10 second delay.

pmesnier:
Just threw in an updated travis config file that includes a second p2p test. This second test adds a delay between launching instances of eosd to insure synchronization.

PR #655, opened 2017-11-09
Delete duplicate function.

gguoss:


plugins/chain_plugin/include/eos/chain_plugin/chain_plugin.hpp
```diff
@@ -160,7 +160,7 @@ class read_only {
       name        scope;
       name        code;
       name        table;
-//      string      table_type;
+
```
spicynoodle: `table_type` is used to specify the whether the table is `Dual Index Table` or `Single Index Table`

gguoss: Accept

heifner:
I would rather see a "duplication" then introduce a const cast.

Issue #618, opened 2017-11-01
DAWN-118 ⁃ Modify P2P to use scatter/gather for reads


blockone-syncclient:
Modify the P2P code to do the following:

* Use a sequence of fixed size buffers to store data read from the TCP socket
* Grow and shrink the chain of buffers as necessary
* Use a boost pool to manage the individual buffers

This change should result in better memory usage for eosd and make it easier to do further time/space optimizations.

ATC

There is no externally visible functional changes related to this issue, so normal regression testing with a multi-node network is all that is required.

bytemaster:
Is this change about making things faster or using less space?

Does this change represent pre-mature optimization?

pmesnier:
space reduction is one issue, but this also addresses a situation Matthew had encountered, which is the case where a signed block message exceeded the fixed buffer size.  With the fixed buffers we are sunk if that occurs.  Prior to this patch we didn't have a strategy to support fractional message reads. 

So maybe in the context of STAT this is premature optimization, but I believe it improves overall robustness and scalability.

PaulCalabrese:
I typed a longer message but Phil beat me to most of it.

My last change eliminated the fixed buffer size and this one uses scatter/gather to avoid large individual buffers (which are required for each connection). They are really both part of the larger changes Phil discussed and allow us to scale better for connections, message size, and message frequency.

blockone-syncclient:
➤ Dhanesh Valappil commented:

I am not too sure how exactly should this be tested.
As the changes are in p2p layer, will do basic tests on p2p communication and syncs

blockone-syncclient:
➤ Dhanesh Valappil commented:

Tested a few p2p scenarios and looks good

PR #663, opened 2017-11-10
STAT-118 (GH-618): Use Scatter/Gather for reads

PaulCalabrese:
This changes the P2P code to use scatter/gather reads to and a sequence of buffers in place of one small buffer. It uses boost object_pool to manage the individual buffers.

### DIFF
--- a/plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
+++ b/plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
@@ -1,8 +1,11 @@
+/**
+ *  @file
+ *  @copyright defined in eos/LICENSE.txt
+ */
 #pragma once
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/pool/object_pool.hpp>
 #include <fc/io/raw.hpp>
-#include <iostream>
 #include <deque>
 #include <array>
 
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
```diff
@@ -0,0 +1,278 @@
+#pragma once
```
heifner: Add copyright comment.

plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
```diff
@@ -0,0 +1,278 @@
...
+    void advance_write_ptr(uint32_t bytes) {
+      advance_index(write_ind, bytes);
+    }
+
+    /*
+     *  Debug function that will print the state of the message buffer
+     *  to standard out.
+     */
+    void show_state() const {
+      std::cout << "Number of buffers = " << buffers.size() << "\n"
```
heifner: Should use the logger like in net_plugin.

tests/tests/message_buffer_tests.cpp
```diff
@@ -0,0 +1,195 @@
+
```
heifner: Copyright comment.

plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
PaulCalabrese: I was using this member function to debug some of the unit tests, I couldn't decide whether to leave it in or not. I might just take it out.

bytemaster:
See my comments on: https://github.com/eosio/eos/issues/618

plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
pmesnier: I agree with Kevin on this. We need to be consistent with our logging strategy. No other plugin or library code uses std::cout or std::cerr.

pmesnier: I agree with Kevin. We need to be consistent with our logging.

Issue #667, opened 2017-11-12
eosd "--public-endpoint arg" need to be elaborated


spicynoodle:
the `--public-endpoint arg` under `eos::network_plugin` is commented as "_overrides the **advertised** listen endpoint ip address_". 
The definition of **advertised listen endpoint** may need to be elaborated.
![image](https://user-images.githubusercontent.com/16319106/32696304-c23d4e60-c7af-11e7-8999-17688732e8a7.png)

pmesnier:
Here's a revised description:
  --listen-endpoint arg (=0.0.0.0:9876) The actual host:port used to listen for
                                        incoming p2p connections.
  --public-endpoint arg                 An externally accessible host:port for 
                                        identifying this node. Defaults to 
                                        listen-endpoint.
  --remote-endpoint arg                 The public endpoint of a peer. Repeat 
                                        this option to build a list.
  --agent-name arg (="EOS Test Agent")  The name supplied to identify this node
                                        amongst the peers.

spicynoodle:
What's the meaning of "Repeat this option to build a list.", may elaborate "option" and "list"

pmesnier:
no, but I trimmed too much from the "remote-endpoint" description. Is something like this more clear:
  -- remote-endpoint arg   The public endpoint of a peer node to connect to. Use multiple remote-endpoint options as needed to compose a network.

heifner:
Resolved by #698.

PR #669, opened 2017-11-12
 fix the comments for --public-testnet arg

spicynoodle:
fix the comments for --public-testnet arg


pmesnier:
I've got some other command line option changes as we discussed in issue #667 So I'm going to close this PR, but I willl refernfe your issue. Thank you.

PR #670, opened 2017-11-12
Update README.md

gluneau:
the command `./eosc get table currency currency account` now returns a key instead of an account

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,9 +319,9 @@ Next verify the currency contract has the proper initial balance:
 ./eosc get table currency currency account
 {
   "rows": [{
-     "account": "account",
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -319,9 +319,9 @@ Next verify the currency contract has the proper initial balance:
 ./eosc get table currency currency account
 {
   "rows": [{
-     "account": "account",
-     "balance": 1000000000
-     }
+      "key": "3607749778735104000",
```
spicynoodle: may explain the rationale of this number

heifner:
The code is actually wrong, See #702

PR #675, opened 2017-11-13
fixed few error

asvirido:
I fixed error in smart-contract skeleton and change namespace in "fixedpoint.hpp" and "real.hpp" from "eos" to "eosio"


heifner:
@asvirido This is not building.

brianjohnson5972:
STAT-182

PR #684, opened 2017-11-13
Travis: set CMAKE_BUILD_TYPE for eos

mitza-oci:


heifner:
How about adding -j 4 to the mongo builds. That could speed it up.

PR #689, opened 2017-11-13
Embed git version in eosc/eosd/launcher.  Log in eosd.  Resolves #664

### LINKED ISSUE #664: STAT-146 ⁃ Add Release / Build number to eosd logging
When we are testing, releasing, and/or debugging a eosd release, it would be great that on startup of the process that it logs it's build/release number to the top of the log on startup.

It would be nice if this was the very first line the process logs so it is easy to find.

ATC:

Run eosd and eosc with --version option to echo version to output.

jgiszczak:
Not strictly necessary for launcher since it does not interact so directly with eosc/eosd, but included for the sake of completeness.  Launcher has been getting new features that correspond with new features in eosd, so being able to verify a version match could be useful.

### DIFF
--- a/programs/launcher/CMakeLists.txt
+++ b/programs/launcher/CMakeLists.txt
@@ -5,12 +5,31 @@ endif()
 
 find_package( Gperftools QUIET )
 if( GPERFTOOLS_FOUND )
-    message( STATUS "Found gperftools; compiling steemd with TCMalloc")
+    message( STATUS "Found gperftools; compiling with TCMalloc")
     list( APPEND PLATFORM_SPECIFIC_LIBS tcmalloc )
 endif()
 
-target_link_libraries( launcher
-                       PRIVATE fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../../.git)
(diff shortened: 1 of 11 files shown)


programs/launcher/CMakeLists.txt
```diff
@@ -5,12 +5,31 @@ endif()
...
 find_package( Gperftools QUIET )
 if( GPERFTOOLS_FOUND )
-    message( STATUS "Found gperftools; compiling steemd with TCMalloc")
+    message( STATUS "Found gperftools; compiling with TCMalloc")
     list( APPEND PLATFORM_SPECIFIC_LIBS tcmalloc )
 endif()
 
-target_link_libraries( launcher
-                       PRIVATE fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
+if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../../.git)
```
brianjohnson5972: Are these 3 cmake blocks the same?  Can we make a git.cmake file and use a macro/function?

jgiszczak: They are very similar, but they each set a different variable.  It's probably possible to make a cmake function type thing that accepts the variable name as a parameter, but I didn't take the time to figure out how.

PR #720, opened 2017-11-16
Add missing set of integer version.

### LINKED ISSUE #687: STAT-151 ⁃ Node Startup Versioning Checks
Is there any way to add code that checks the versions of the other nodes and logs stderr warnings if the releases do not match?

Or even a way to ask a node what version it's running?

If the node release versions do not match, that node should not start.

ATC:
Compile a diff version of eosd. Verify it will not connect to testnet.

jgiszczak:
Knew it was going to be a problem when I shadowed the version variable.
Fix it later.

STAT-151. Resolves #687

### DIFF
--- a/programs/eosd/main.cpp
+++ b/programs/eosd/main.cpp
@@ -28,6 +28,7 @@ int main(int argc, char** argv)
 {
    try {
       app().set_version(eosio::eosd::config::version_str);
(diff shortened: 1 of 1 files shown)


heifner:
As we discussed, this would be better as a diff method.

PR #721, opened 2017-11-16
Stop having two types of version.

jgiszczak:
Will get rid of support for two types in appbase later.

### DIFF
--- a/programs/eosd/main.cpp
+++ b/programs/eosd/main.cpp
@@ -27,9 +27,8 @@ using namespace eosio;
 int main(int argc, char** argv)
 {
    try {
-      app().set_version(eosio::eosd::config::version_str);
-      app().set_version(eosio::eosd::config::version_int);
-      ilog("eosd version ${ver}", ("ver", app().version()));
+      app().set_version(eosio::eosd::config::version);
+      ilog("eosd version ${ver}", ("ver", eosio::eosd::config::itoh((uint32_t)app().version_int())));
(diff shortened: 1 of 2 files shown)


programs/eosd/main.cpp
```diff
@@ -27,9 +27,8 @@ using namespace eosio;
 int main(int argc, char** argv)
 {
    try {
-      app().set_version(eosio::eosd::config::version_str);
-      app().set_version(eosio::eosd::config::version_int);
-      ilog("eosd version ${ver}", ("ver", app().version()));
+      app().set_version(eosio::eosd::config::version);
+      ilog("eosd version ${ver}", ("ver", eosio::eosd::config::itoh((uint32_t)app().version_int())));
```
heifner: I would prefer a `static_cast<uint32_t>()`

jgiszczak: I'm converting it to hex in order to provide an easily human readable match between the reported version from a running daemon and git's hash.  Forcing people running the daemon to manually convert from decimal to hex to do that check is very unfriendly.

heifner: Right. I was just saying this:
`eosio::eosd::config::itoh((uint32_t)app().version_int())`
to
`eosio::eosd::config::itoh(static_cast<uint32_t>(app().version_int()))`

jgiszczak: Oh.  Sure, will do.

Issue #733, opened 2017-11-18
STAT-180 ⁃ Launch a parallel testnet with 1 second per block


blockone-syncclient:
As per Dan,

We need a new testnet configured with 1 sec per block and run it upto a million blocks.

ATC:

* Run eosd and watch output to see that a new block is produced once a second.

blockone-syncclient:
➤ Kevin Heifner commented:

* Default block interval changed to 1 sec.
    * Use ./eosd --resync --block-interval-seconds 3 to run with previous default.

  * Block interval made configurable via a command line or config.ini option to chain_plugin.

blockone-syncclient:
➤ Dhanesh Valappil commented:

The eosd instance, without specifying the blok-interval-seconds generates block every second
68009ms thread-0   chain_controller.cpp:208      pushblock          ] initj #1116 @2017-11-22T09:02:48  | 0 trx, 0 pending, exectime_ms=1
168009ms thread-0   producer_plugin.cpp:244       block_production_loo ] initj generated block #1116 @ 2017-11-22T09:02:48 with 0 trxs  0 pending
169004ms thread-0   chain_controller.cpp:208      pushblock          ] inith #1117 @2017-11-22T09:02:49  | 0 trx, 0 pending, exectime_ms=0
169004ms thread-0   producer_plugin.cpp:244       block_production_loo ] inith generated block #1117 @ 2017-11-22T09:02:49 with 0 trxs  0 pending
170002ms thread-0   chain_controller.cpp:208      pushblock          ] initm #1118 @2017-11-22T09:02:50  | 0 trx, 0 pending, exectime_ms=0
170002ms thread-0   producer_plugin.cpp:244       block_production_loo ] initm generated block #1118 @ 2017-11-22T09:02:50 with 0 trxs  0 pending
1

blockone-syncclient:
➤ Dhanesh Valappil commented:

Looks good as context switching among nodes happen every second. Blocks are being generated by the same node on every 21 seconds
397001ms thread-0   producer_plugin.cpp:244       block_production_loo ] initi generated block #2952 @ 2017-11-23T09:06:37 with 0 trxs  0 pending
409002ms thread-0   chain_controller.cpp:208      pushblock          ] initk #2953 @2017-11-23T09:06:49  | 0 trx, 0 pending, exectime_ms=0
411714ms thread-0   net_plugin.cpp:1132           connect              ] host: ip-10-160-12-231 port: 9876 
[...]

PR #734, opened 2017-11-18
Make block interval configurable

heifner:
#733

* Default block interval changed to 1 sec.
* Block interval made configurable via a command line or config.ini option to chain_plugin.


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -119,16 +122,20 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
 void chain_plugin::plugin_initialize(const variables_map& options) {
    ilog("initializing chain plugin");
 
+   my->block_interval_seconds = config::default_block_interval_seconds;
```
brianjohnson5972: This line and the next are redundant, since line 90 uses default_value, options will perform the same check.

PR #735, opened 2017-11-19
Update readme with correct config changes required

suryagh:
The readme says to add the `key` to existing config file. But, some of the keys already exist, and only need to be edited with required value. I stumbled on this while trying to get `eosd` run on my mac


suryagh:
retest this please

suryagh:
@heifner the readme says to add keys: `enable-stale-production` & `genesis-json`. Which if fillowed causes `./esod` to not start-up. So, that is a documentation bug.

PR #741, opened 2017-11-20
STAT-153(GH-682): Remove signed_transaction from transaction_object s…

### LINKED ISSUE #682: STAT-153 ⁃ SignedTransaction should not be stored in shared memory
From [~dan.larimer]

https://github.com/EOSIO/eos/blame/66bcdbb06c9700f72b6f19762f77412d20866809/libraries/chain/include/eos/chain/transaction_object.hpp.  line 29..

SignedTransaction should not exist in an object stored in shared memory
this will cause crashes

ATC

1) Start a multi-eosd system
2) Make some basic transactions (something like the wallet transaction in eos/README.md)
3) Wait until the transaction moves to irreversible blocks.
4) Kill one of the eosd processes.
5) Restart the eosd process without either --resync-blockchain or --replay-blockchain.
6) Verify that the nodes are in sync (using eosc get info)
7) Make another transaction (like a currency transfer) and ensure it succeeds

Limitations:
A) If there are generated transactions in the blockchain, the resync may not work if STAT-156 has not been fixed.
B) Any transactions in the reversible blocks may cause issues syncing with an exceptions in the log of "could not insert object, most likely a uniqueness constraint was violated". A separate issue will be filed for this. See STAT-188.

PaulCalabrese:
The transaction_object previously contained a signed_transaction which has various vector data members, making it unsuitable to store in shared memory. This change removes the signed_transaction and stores its id in the transaction_object. Various other locations were corrected to use this and one usage was commented out in handle_message(block_summary) as this is not currently supported.
Resolves #682

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -86,16 +86,6 @@ optional<signed_block> chain_controller::fetch_block_by_number(uint32_t num)cons
    return optional<signed_block>();
 }
 
-/*
-const signed_transaction& chain_controller::get_recent_transaction(const transaction_id_type& trx_id) const
-{
-   auto& index = _db.get_index<transaction_multi_index, by_trx_id>();
-   auto itr = index.find(trx_id);
-   FC_ASSERT(itr != index.end());
-   return itr->trx;
-}
-*/
-
 std::vector<block_id_type> chain_controller::get_block_ids_on_fork(block_id_type head_of_fork) const
 {
   pair<fork_database::branch_type, fork_database::branch_type> branches = _fork_db.fetch_branch_from(head_block_id(), head_of_fork);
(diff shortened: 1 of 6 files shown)


libraries/chain/chain_controller.cpp
```diff
@@ -86,13 +86,15 @@ optional<signed_block> chain_controller::fetch_block_by_number(uint32_t num)cons
    return optional<signed_block>();
 }
 
+/*
```
heifner: Add to this comment why the code is commented out.

libraries/chain/include/eos/chain/chain_controller.hpp
```diff
@@ -121,7 +121,7 @@ namespace eosio { namespace chain {
          block_id_type               get_block_id_for_num( uint32_t block_num )const;
          optional<signed_block>      fetch_block_by_id( const block_id_type& id )const;
          optional<signed_block>      fetch_block_by_number( uint32_t num )const;
-         const signed_transaction&    get_recent_transaction( const transaction_id_type& trx_id )const;
+         //const signed_transaction&    get_recent_transaction( const transaction_id_type& trx_id )const;
```
heifner: Add why commented out, Maybe just // see definition for why this is commented out.

PR #750, opened 2017-11-21
Launcher hostconf 701

pmesnier:
Added command line options:

--nogen     makes the launcher start programs without rewriting the config files.
--host-map  provides an alternate filename for reading or writing the hosts file as a
            companion to the network topology file

Refactored the network definition to separate the host configuration from the network layout so that different host configurations can be used to deploy a common configuration over multiple sets of servers.

By default, -o <filename.json> will also create a <filename_hosts.json> as a dummy filename for the hosts. likewise to deploy, use -s <filename.json> which will read from <filename_hosts.json>. for an alternate host map file, use -s <filename.json> --host-map <althosts.json>.

### DIFF
--- a/programs/launcher/main.cpp
+++ b/programs/launcher/main.cpp
@@ -24,6 +24,7 @@
 #include <sys/types.h>
 #include <netinet/in.h>
 #include <net/if.h>
+#include <eos/native_contract/native_contract_chain_initializer.hpp>
 
 #include "config.hpp"
 
@@ -36,12 +37,14 @@ using boost::asio::ip::host_name;
 using bpo::options_description;
 using bpo::variables_map;
 
-struct localIdentity {
+struct local_identity {
   vector <fc::ip::address> addrs;
   vector <string> names;
 
   void initialize () {
     names.push_back ("localhost");
+    names.push_back ("127.0.0.1");
+
     boost::system::error_code ec;
     string hn = host_name (ec);
     if (ec.value() != boost::system::errc::success) {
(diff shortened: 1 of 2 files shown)


programs/launcher/main.cpp
```diff
@@ -478,43 +623,47 @@ launcher_def::write_config_file (eosd_def &node) {
...
     }
 
-    bf::path dpath = bf::path (node.eos_root_dir) / node.data_dir / "config.ini";
-    scp_cmd_line += node.hostname + ":" + dpath.string();
+    bf::path dpath = bf::path (host->eos_root_dir) / instance.data_dir / "config.ini";
+    scp_cmd_line += host->host_name + ":" + dpath.string();
 
     cerr << "cmdline = " << scp_cmd_line << endl;
-    int res = boost::process::system (scp_cmd_line);
+    int res = 1; //boost::process::system (scp_cmd_line);
```
jgiszczak: scp command line not actually executed anymore?

programs/launcher/main.cpp
```diff
@@ -747,22 +905,35 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
...
-          continue;
+  case LM_VERIFY:
+    //validate configuration, report findings, exit
+    return;
+  case LM_NAMED : {
+    try {
+      auto node = network.nodes.find(launch_name);
+      launch(*node->second.instance, gts);
+    } catch (...) {
+      // failed to start.
```
jgiszczak: Does something visibly report the launch failure and say why?

programs/launcher/main.cpp
```diff
@@ -747,22 +905,35 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
...
+    break;
+  }
+  case LM_ALL:
+  case LM_REMOTE:
+  case LM_LOCAL: {
+    for (auto &h : host_map.bindings ) {
+      if (mode == LM_ALL ||
+          (h.second.is_local() ? mode == LM_LOCAL : mode == LM_REMOTE)) {
+        for (auto &inst : h.second.instances) {
+          launch (inst, gts);
```
jgiszczak: `launch()` above is wrapped in try{}catch{}.  Shouldn't this invocation also be wrapped?

programs/launcher/main.cpp
```diff
@@ -24,6 +24,7 @@
 #include <sys/types.h>
 #include <netinet/in.h>
 #include <net/if.h>
+#include <eos/native_contract/native_contract_chain_initializer.hpp>
```
heifner: Why was this added? I don't see it being used.

pmesnier: I was planning on pulling producer names from the genesis block rather than just using "inita" etc. but I ran out of time.

programs/launcher/main.cpp
pmesnier: I can add that.

programs/launcher/main.cpp
pmesnier: Yikes! Good catch, update for this on its way.

PR #755, opened 2017-11-22
Benchmark fixes for signature

pavybez:
STAT-189

Fixed benchmark to a state similar to last time. There are other bugs in benchmark that still require attention.


programs/eosc/main.cpp
```diff
@@ -972,10 +977,11 @@ int main( int argc, char** argv ) {
          transaction_emplace_message(trx, config::eos_contract_name,
                                               vector<types::account_permission>{{sender,"active"}},
                                               "transfer", types::transfer{sender, recipient, amount, memo});
-         trx.expiration = info.head_block_time + tx_expiration; 
+         trx.expiration = info.head_block_time + fc::seconds(1800);
```
heifner: Please add this as an argument to benchmark like it is for other command line options. Also use below.

programs/eosc/main.cpp
```diff
@@ -955,9 +955,11 @@ int main( int argc, char** argv ) {
 
    auto benchmark_transfer = benchmark->add_subcommand( "transfer", localized("executes random transfers among accounts") );
    uint64_t number_of_transfers = 0;
+   uint64_t expiration = 0;
    bool loop = false;
    benchmark_transfer->add_option("accounts", number_of_accounts, localized("the number of accounts in transfer among"))->required();
    benchmark_transfer->add_option("count", number_of_transfers, localized("the number of transfers to execute"))->required();
+   benchmark_transfer->add_option("expiration", expiration, localized("the number of seconds to expire"))->required();
```
heifner: Please use `add_standard_transaction_options()`.

PR #770, opened 2017-11-24
Malicious contract

dhaneshvb:


contracts/CMakeLists.txt
```diff
@@ -6,6 +6,8 @@ add_subdirectory(test_api)
 add_subdirectory(simpledb)
```
heifner: Please move test contracts to `tests/tests/contracts`

contracts/dummy/dummy.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <eoslib/eos.hpp>
+#include <eoslib/token.hpp>
+#include <eoslib/db.hpp>
+
+/**
+ *  @defgroup examplecontract Example Contract
+ *  @brief Example Contract
+ *  @ingroup contractapi
+ *
```
heifner: Update doxygen with correct info. Also indicate what the purpose of this contract is for.

contracts/dummy/dummy.hpp
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+
+/**
+ * Make it easy to change the account name the currency is deployed to.
+ */
+#ifndef TOKEN_NAME
+#define TOKEN_NAME currency
```
heifner: I think it would be better to call this something other than currency so it is not confused with the other example currency contract.

contracts/malicious/malicious.cpp
```diff
@@ -0,0 +1,108 @@
...
+ *  @copyright defined in eos/LICENSE.txt
+ */
+
+#include <malicious/malicious.hpp>
+#include <eoslib/memory.hpp>
+#include <eoslib/account.hpp>
+#include <eoslib/eos.hpp>
+#include <eoslib/system.h>
+
+#define EOS_SYMBOL  (int64_t(4) | (uint64_t('E') << 8) | (uint64_t('O') << 16) | (uint64_t('S') << 24))
```
heifner: Remove. Not used.

contracts/malicious/malicious.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <eoslib/eos.hpp>
+#include <eoslib/token.hpp>
+#include <eoslib/db.hpp>
+
+/**
+ *  @defgroup examplecontract Example Contract
+ *  @brief Example Contract
+ *  @ingroup contractapi
+ *
```
heifner: Update with useful doxygen. Explain the purpose of this contract.

contracts/malicious/malicious.hpp
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+
+/**
+ * Make it easy to change the account name the currency is deployed to.
+ */
+#ifndef TOKEN_NAME
+#define TOKEN_NAME currency
+#endif
```
heifner: Naming this currency is confusing. This has nothing to do with TOKENs. Please remove use of TOKEN_NAME.

contracts/malicious/malicious.cpp
```diff
@@ -0,0 +1,108 @@
...
+    using namespace eosio;
+
+    ///  When storing accounts, check for empty balance and remove account
+    void store_account(account_name account_to_store, const account &a) {
+
+    }
+
+    // Create memory leak , try to allocate huge chunk of memory
+    void apply_malicious_resource() {
+        // allocate memory in an infinite loop
```
heifner: I don't see a loop here. Is the comment wrong, or did you forget the loop?

contracts/malicious/malicious.cpp
```diff
@@ -0,0 +1,108 @@
...
+        eosio::account::get(b);
+        eosio::print("\n\tAccount=", b.account);
+        eosio::print("\n\tBalance amount=", (uint64_t) b.eos_balance.amount);
+        eosio::print("\n\tBalance symbol=",b.eos_balance.symbol);
+        eosio::print("\n\tStaked balance=", (uint64_t) b.staked_balance.amount);
+        eosio::print("\n\tStaked balance symbol=", b.staked_balance.symbol);
+        eosio::print("\n\tUnstaking balance amount=", (uint64_t) b.unstaking_balance.amount);
+        eosio::print("\n\tUnstaking balance symbol=", b.unstaking_balance.symbol);
+        eosio::print("\n\tLast unstaking time=", b.last_unstaking_time);
+    }
```
heifner: What is malicious about this? Looks more like just a simple test of getting the account balance. If your plan is to use malicious for misc testing. Then rename to misctest or something.

contracts/malicious/malicious.cpp
```diff
@@ -0,0 +1,108 @@
...
+
+    // Infinite messaging between Malicious and Dummy contract
+    void apply_malicious_notify(const TOKEN_NAME::transfer &transfer_msg) {
```
heifner: The coding style is for '&' to be with transfer:  `const TOKEN_NAME::transfer& transfer_msg`

contracts/dummy/dummy.cpp
```diff
@@ -0,0 +1,43 @@
+/**
+ *  @file
+ *  @copyright defined in eos/LICENSE.txt
+ */
+
+#include <dummy/dummy.hpp>
+#include <eoslib/memory.hpp>
+
+
```
heifner: Instead of calling this dummy. How about malicioushelp.

tests/eos_bashrc
```diff
@@ -0,0 +1,152 @@
+# Note: you need to configure your eosroot according to your folder structure
```
heifner: Having a bashrc here seems odd. I don't think this is a good way to maintain a multi-developer project.  Please create a `tests/script` directory and create individual scripts in there instead of having one large bashrc file.

contracts/CMakeLists.txt
dhaneshvb: Sure, will do that

PR #776, opened 2017-11-24
Refactor block production to have limited execution time and block size

bytemaster:
This should contain block size to 1000 trx.

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -323,7 +323,6 @@ signed_block chain_controller::_generate_block(
       pending_block.cycles[0].resize(1); // single thread
 
       for( const auto& pending_trx : _pending_transactions ) {
-         idump((pending_trx.id()));
          if( (fc::time_point::now() - start) > fc::milliseconds(200) ||
              pending_block_size > gprops.configuration.max_blk_size)
          {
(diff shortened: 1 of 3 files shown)


libraries/chain/chain_controller.cpp
```diff
@@ -241,6 +241,8 @@ processed_transaction chain_controller::_push_transaction(const signed_transacti
    if (!_pending_tx_session.valid())
       _pending_tx_session = _db.start_undo_session(true);
 
+   FC_ASSERT( _pending_transactions.size() < 1000, "too many pending transactions, try again later" );
```
brianjohnson5972: If we make 1000 a config parameter in chain_controller, then we can adjust these during testing and not have to wait for a tag and recompile for the test net

heifner: @brianjohnson5972 go ahead and prepare a PR with making these two configurable.

libraries/chain/chain_controller.cpp
```diff
@@ -429,16 +366,14 @@ signed_block chain_controller::_generate_block(
       pending_block.producer_changes = get_global_properties().active_producers - new_schedule;
    }
 
+   auto end = fc::time_point::now();
+   ilog( "generation took ${x} ms", ("x", (end-start).count()/1000) );
```
heifner: This is generating too many output statements causing the travis build to fail.

heifner:
@brianjohnson5972 Since I added to this PR, please review at least my portion. I modified the timings on your rate limiting tests.

Issue #777, opened 2017-11-24
STAT-200 ⁃ deployment optimizations


blockone-syncclient:
on net_plugin, increased the limit on client connections and made it configurable. also change the behavior of initial connections so that if it has never connected to a peer, keep trying asap to connect to at least one, then back off for retrying others. 

on launcher, give it a list of host IPs for it to use for remote configuration. Second, format the aliases to use leading 0's as necessary so that these names string-sort as expected. Third, increase the number of connections used in the star configuration and allow the mesh configuration to specify redundant connections to insure as quick a connection as possible.

pmesnier:
all features implemented and merged via #779

blockone-syncclient:
➤ Andrianto Lie commented:

Pass.
All nodes show 21 for the given shell command in a stable testnet network.

PR #779, opened 2017-11-25
#777 deployment improvements STAT-200

pmesnier:
Fixes for testnet deployment issues.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -157,6 +157,7 @@ namespace eosio {
     chain_plugin*                 chain_plug;
     size_t                        just_send_it_max;
     bool                          send_whole_blocks;
+    int                           started_sessions;
 
     node_transaction_index        local_txns;
     ordered_txn_ids               pending_notify;
@@ -263,8 +264,8 @@ namespace eosio {
    */
   constexpr auto     def_send_buffer_size_mb = 4;
   constexpr auto     def_send_buffer_size = 1024*1024*def_send_buffer_size_mb;
-  constexpr auto     def_max_clients = 20; // 0 for unlimited clients
-  constexpr auto     def_conn_retry_wait = std::chrono::seconds (30);
+  constexpr auto     def_max_clients = 25; // 0 for unlimited clients
+  constexpr auto     def_conn_retry_wait = std::chrono::seconds (1);
   constexpr auto     def_txn_expire_wait = std::chrono::seconds (3);
   constexpr auto     def_resp_expected_wait = std::chrono::seconds (1);
   constexpr auto     def_sync_rec_span = 10;
@@ -1184,6 +1185,9 @@ namespace eosio {
                                        ( "peer", c->peer_name())("error",err.message()));
                                  c->connecting = false;
                                  c->close();
+                                 if (started_sessions == 0) {
+                                   connect (c);
+                                 }
                                }
                              }
                            } );
@@ -1193,6 +1197,7 @@ namespace eosio {
       boost::asio::ip::tcp::no_delay option( true );
       con->socket->set_option( option );
       start_read_message( con );
+      ++started_sessions;
       con->send_handshake( );
 
       // for now, we can just use the application main loop.
@@ -2166,6 +2171,8 @@ namespace eosio {
      ( "peer-private-key", boost::program_options::value<vector<string>>()->composing()->multitoken(),
        "Tuple of [PublicKey, WIF private key] (may specify multiple times)")
      ( "log-level-net-plugin", bpo::value<string>()->default_value("info"), "Log level: one of 'all', 'debug', 'info', 'warn', 'error', or 'off'")
+      ( "max-clients", bpo::value<int>()->default_value(def_max_clients), "Maximum number of clients from which connections are accepted, use 0 for no limit")
+      ( "connection-cleanup-period", bpo::value<int>()->default_value(30), "number of seconds to wait before cleaning up dead connections")
      ;
   }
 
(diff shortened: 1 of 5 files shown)


heifner:
You need to update:
`tests/eosd_run_test.sh `
`tests/eosd_run_mongodb_test.sh`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2206,12 +2213,15 @@ namespace eosio {
 
     my->sync_master.reset( new sync_manager( def_sync_rec_span ) );
 
-    my->connector_period = def_conn_retry_wait;
+    my->connector_period = options.count( "connection-cleanup-period") ? std::chrono::seconds(options.at("connection-cleanup-period").as<int>()) : def_conn_retry_wait;
```
heifner: Since you specify a default this will always be true. So if not specified it will be 30, NOT def_conn_retry_wait.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2166,6 +2171,8 @@ namespace eosio {
      ( "peer-private-key", boost::program_options::value<vector<string>>()->composing()->multitoken(),
        "Tuple of [PublicKey, WIF private key] (may specify multiple times)")
      ( "log-level-net-plugin", bpo::value<string>()->default_value("info"), "Log level: one of 'all', 'debug', 'info', 'warn', 'error', or 'off'")
+      ( "max-clients", bpo::value<int>()->default_value(def_max_clients), "Maximum number of clients from which connections are accepted, use 0 for no limit")
+      ( "connection-cleanup-period", bpo::value<int>()->default_value(30), "number of seconds to wait before cleaning up dead connections")
      ;
```
heifner: Did you mean to use default of def_conn_retry_wait here instead of 30?

programs/launcher/main.cpp
```diff
@@ -402,19 +434,53 @@ launcher_def::initialize (const variables_map &vmap) {
...
+      for (auto &s : servers.observer) {
+        nodes += s.instances;
+      }
+
+      total_nodes = prod_nodes + nodes;
+      per_host = 1;
+      network.ssh_helper = servers.ssh;
+    }
+    catch (const exception &ex) {
+      elog( "caught assert on fetch_block_by_id, ${ex}",("ex",ex.what()));
```
heifner: Looks like a copy/paste error. Also please use explicit `std::exception` here as `fc::exception` and `std::exception` are both in play and it is not immediately obvious which this is. Note that `fc::exception` does not derive from `std::exception` which means we normally need a catch for each one.

programs/launcher/main.cpp
```diff
@@ -402,19 +434,53 @@ launcher_def::initialize (const variables_map &vmap) {
...
+      exit (-1);
+      }
+    catch (...) {
+      cerr << "unable to load server identity file " << server_ident_file << endl;
```
heifner: Use elog

programs/launcher/main.cpp
```diff
@@ -482,102 +546,224 @@ launcher_def::write_dot_file () {
...
   char * erd = getenv ("EOS_ROOT_DIR");
   if (erd == 0) {
-    cerr << "environment variable \"EOS_ROOT_DIR\" unset. defaulting to current dir" << endl;
     erd = getenv ("PWD");
   }
-  local_host.eos_root_dir = erd;
-  local_host.genesis = genesis.string();
+  stage = bf::path(erd);
+  if (!bf::exists(stage)) {
+    cerr << erd << " is not a valid path" << endl;
```
heifner: elog

programs/launcher/main.cpp
```diff
@@ -482,102 +546,224 @@ launcher_def::write_dot_file () {
...
+      assign_name(eosd);
+      aliases.push_back(eosd.name);
+      eosd.set_host (&local_host);
+      local_host.instances.emplace_back(move(eosd));
+    }
+    bindings.emplace_back(move(local_host));
+  }
+  else {
+    int ph_count = 0;
+    host_def *lhost = 0;
```
heifner: nullptr would be better. Even better would be to use unique_ptr.

programs/launcher/main.cpp
```diff
@@ -482,102 +546,224 @@ launcher_def::write_dot_file () {
...
-      network.nodes[node.name] = move(node);
-      inst.node = &network.nodes[inst.name];
-      i++;
+  }
+}
+
+host_def *
+launcher_def::find_host (const string &name)
+{
+  host_def *host = 0;
```
heifner: nullptr

programs/launcher/main.cpp
```diff
@@ -482,102 +546,224 @@ launcher_def::write_dot_file () {
...
+  for (auto &h : bindings) {
+    if (h.host_name == name) {
+      host = &h;
+      break;
     }
   }
+  if (host == 0) {
+    cerr << "could not find host for " << name << endl;
```
heifner: elog

programs/launcher/main.cpp
```diff
@@ -482,102 +546,224 @@ launcher_def::write_dot_file () {
...
         cerr << "count = " << count << " could not remove old directory: " << dd
              << " " << strerror(ec.value()) << endl;
         exit (-1);
       }
-    }
-    if (!bf::create_directory (instance.data_dir, ec) && ec.value()) {
+      count = bf::remove_all (dd / "blockchain", ec);
+      if (ec.value() != 0) {
+        cerr << "count = " << count << " could not remove old directory: " << dd
+             << " " << strerror(ec.value()) << endl;
```
heifner: elog. Just search file for cerr and fix each one.

programs/launcher/main.cpp
```diff
@@ -684,7 +846,7 @@ launcher_def::make_star () {
 
   size_t links = 3;
   if (total_nodes > 12) {
-    links = (size_t)sqrt(total_nodes);
+    links = (size_t)sqrt(total_nodes) + 2;
```
heifner: Prefer static_cast to C-style casts.

programs/launcher/main.cpp
pmesnier: this was a copy/paste, but the error was not pulling it before committing. At this moment I would prefer to stick with the minimalist approach to error processing. We can switch to more rigorous error handling later, but this is a command-line utility that does its business and then terminates, so I'd rather make enhanced error reporting a post-12/4 activity.

programs/launcher/main.cpp
pmesnier: For the same reason stated above, I'd like to defer refactoring these error messages until later.

Issue #790, opened 2017-11-27
STAT-203 ⁃ Update start_all.sh script


blockone-syncclient:
Per Phil:  I need to make the start_all.sh script a little more generic / portable.

Please post to GitHub in ./eos/scripts when completed.

blockone-syncclient:
➤ Phil Mesnier commented:

in addition to the scripts, I'm going to fix the bug in the launcher that was causing restart headaches last night.

heifner:
Resolved by #791

PR #791, opened 2017-11-27
testnet startup fixes STAT-203

pmesnier:
update the launcher to address failure restarting locally and add testnet scripts to repo, #790

### DIFF
--- a/programs/launcher/main.cpp
+++ b/programs/launcher/main.cpp
@@ -1080,9 +1080,9 @@ launcher_def::start_all (string &gts, launch_modes mode) {
       auto node = network.nodes.find(launch_name);
       launch(*node->second.instance, gts);
     } catch (fc::exception& fce) {
-       cerr << "unable to launch " << launch_name << "fc::exception=" << fce.to_detail_string() << endl;
+       cerr << "unable to launch " << launch_name << " fc::exception=" << fce.to_detail_string() << endl;
     } catch (std::exception& stde) {
-       cerr << "unable to launch " << launch_name << "std::exception=" << stde.what() << endl;
+       cerr << "unable to launch " << launch_name << " std::exception=" << stde.what() << endl;
     } catch (...) {
       cerr << "Unable to launch " << launch_name << endl;
       exit (-1);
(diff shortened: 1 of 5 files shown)


programs/launcher/main.cpp
```diff
@@ -987,16 +986,19 @@ launcher_def::launch (eosd_def &instance, string &gts) {
...
     eosdcmd += "--skip-transaction-signatures ";
   }
-  eosdcmd += eosd_extra_args + " ";
-  eosdcmd += "--enable-stale-production true ";
+  if (!eosd_extra_args.empty()) {
+    eosdcmd += eosd_extra_args + " ";
+  }
+  else {
+    eosdcmd += "--enable-stale-production true ";
+  }
```
heifner: So if extra args are given you don't want the `--enable-stale-production true` ?

pmesnier: correct. the user can add --enable-stale-production true along with any other arguments if it is desired, but in cases where a server is being bounced, --enable-stale-production should be replaced with --resync.

programs/launcher/main.cpp
```diff
@@ -716,8 +716,7 @@ launcher_def::deploy_config_file (tn_node_def &node) {
            << " errno " << ec.value() << " " << strerror(ec.value()) << endl;
```
brianjohnson5972: Above this, at line 468, can we add:

          } catch (fc::exception& fce) {
            cerr << "unable to launch " << inst.name << "fc::exception=" << fce.to_detail_string() << endl;
          } catch (std::exception& stde) {
             cerr << "unable to launch " << inst.name << "std::exception=" << stde.what() << endl;

brianjohnson5972: It helped me out a lot when it was failing because the test directory wasn't cleaned up previously for some reason.

programs/launcher/main.cpp
```diff
@@ -1092,6 +1098,10 @@ launcher_def::start_all (string &gts, launch_modes mode) {
         for (auto &inst : h.instances) {
           try {
             launch (inst, gts);
+          } catch (fc::exception& fce) {
+             cerr << "unable to launch " << inst.name << "fc::exception=" << fce.to_detail_string() << endl;
+          } catch (std::exception& stde) {
+             cerr << "unable to launch " << inst.name << "std::exception=" << stde.what() << endl;
           } catch (...) {
```
heifner: Can we get a space added. `" fc::exception="` and `" std::exception="`. Both here and above.

Issue #796, opened 2017-11-28
DAWN-215 ⁃ Remove Block Summary code from P2P


PaulCalabrese:
Since we are not using block summaries for STAT, this change removes any code related to these messages from the P2P plug-in

blockone-syncclient:
➤ Paul Calabrese commented:

I am cancelling this issue as we don't really want to do this now that the testnet has launched. My previous pull request deleted the block_summary_message and all associated code. Unfortunately, doing this now would break backward-compatibility in the testnet, so we can't apply this change. In addition, the p2p code is being reworked on the noon branch, so there is really no advantage to doing this.

PR #797, opened 2017-11-28
GH-796: Remove block summary support

PaulCalabrese:
Remove the P2P support for block summary as it is not supported in STAT. Also removes the send-whole-blocks option from the launcher.


PaulCalabrese:
I am withdrawing this pull request as this needs to be reworked because of concerns about "breaking" changes to the protocol.

PR #800, opened 2017-11-28
clean up network code, remove reconnect tight loop, default to allow …

bytemaster:
…any connections, fix style issues


plugins/net_api_plugin/net_api_plugin.cpp
```diff
@@ -0,0 +1,110 @@
...
+      if (host != "localhost" && host != "127.0.0.1") {
+         wlog("\n"
+              "*************************************\n"
+              "*                                   *\n"
+              "*  --   Wallet NOT on localhost  -- *\n"
+              "*  - Password and/or Private Keys - *\n"
+              "*  - are transferred unencrypted. - *\n"
+              "*                                   *\n"
+              "*************************************\n");
+      }
```
heifner: There are a number of remnants from the wallet_api_plugin left in here. This message and the commented out code above.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1186,20 +1198,20 @@ namespace eosio {
                                        ( "peer", c->peer_name())("error",err.message()));
                                  c->connecting = false;
                                  c->close();
-                                 if (started_sessions == 0) {
-                                   connect (c);
-                                 }
+                                 /*if(started_sessions == 0) {
+                                   connect(c);
```
pmesnier: you should just delete these lines to avoid a merge conflict.

PR #804, opened 2017-11-28
Fix memory issue with full message_buffer

PaulCalabrese:
When the message_buffer's write index was advanced to the end of the buffer, the get_buffer_sequence_for_boost_async_read() function would return invalid pointers. An assert was added to that function to detect this, but the real problem was fixed in advance_write_ptr() to always assure that the write_ptr is pointing to allocated memory.

### DIFF
--- a/plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
+++ b/plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
@@ -35,6 +35,13 @@ namespace eosio {
 
     message_buffer() : buffers{pool().malloc()}, read_ind{0,0}, write_ind{0,0} { }
 
+    ~message_buffer() {
+      while (buffers.size() > 0) {
+        pool().destroy(buffers.back());
+        buffers.pop_back();
+      }
+    }
+
     /*
      *  Returns the current read index referencing the byte in the buffer
      *  that is next to be read.
@@ -84,6 +91,19 @@ namespace eosio {
     }
 
     /*
+     *  Resets the message buffer to the initial state. Any unread data is
+     *  discarded.
+     */
+    void reset() {
+      read_ind = { 0, 0 };
+      write_ind = { 0, 0 };
+      while (buffers.size() > 1) {
+        pool().destroy(buffers.back());
+        buffers.pop_back();
+      }
+    }
+
+    /*
      *  Returns the current number of bytes remaining to be read.
      *  Logically, this is the different between where the read and write pointers are.
      */
@@ -119,12 +139,7 @@ namespace eosio {
     void advance_read_ptr(uint32_t bytes) {
       advance_index(read_ind, bytes);
       if (read_ind == write_ind) {
-        read_ind = { 0, 0 };
-        write_ind = { 0, 0 };
-        while (buffers.size() > 1) {
-          pool().destroy(buffers.back());
-          buffers.pop_back();
-        }
+        reset();
       } else if (read_ind.first > 0) {
         while (read_ind.first > 0) {
           pool().destroy(buffers.front());
--- a/plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
+++ b/plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
@@ -141,6 +141,9 @@ namespace eosio {
      */
     void advance_write_ptr(uint32_t bytes) {
       advance_index(write_ind, bytes);
+      while (write_ind.first >= buffers.size()) {
+        buffers.push_back(pool().malloc());
+      }
     }
 
     /*
@@ -151,6 +154,7 @@ namespace eosio {
      */
     std::vector<boost::asio::mutable_buffer> get_buffer_sequence_for_boost_async_read() {
       std::vector<boost::asio::mutable_buffer> seq;
+      FC_ASSERT(write_ind.first < buffers.size());
       seq.push_back(boost::asio::buffer(&buffers[write_ind.first]->at(write_ind.second),
                                                 buffer_len - write_ind.second));
       for (std::size_t i = write_ind.first + 1; i < buffers.size(); i++) {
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
```diff
@@ -141,6 +141,9 @@ namespace eosio {
      */
     void advance_write_ptr(uint32_t bytes) {
       advance_index(write_ind, bytes);
+      while (write_ind.first >= buffers.size()) {
+        buffers.push_back(pool().malloc());
```
wanderingbort: This seems "too late".

Our usage pattern is to pass the boost buffers to `async_read_some` and then in the resulting handler, call `advance_write_ptr` to account for the data that was written in the async call.  

By this time, the data that would have been written has already over flowed any buffer it thought it had.

wanderingbort: I agree as an API this makes sense to advance, then write.  So this change is good outright but not exercised in our code base

plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
```diff
@@ -151,6 +154,7 @@ namespace eosio {
      */
     std::vector<boost::asio::mutable_buffer> get_buffer_sequence_for_boost_async_read() {
       std::vector<boost::asio::mutable_buffer> seq;
+      FC_ASSERT(write_ind.first < buffers.size());
```
wanderingbort: This is a good assert, I have it locally and am hitting it.  It would be more useful if it printed some data out like the value of `write_ind.first` and `buffers.size()`

Also, the code that calls this doesn't properly catch and recover from thrown errors.  See net_plugin.cpp:1230

plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
PaulCalabrese: There will not be any overflow as async_read_some() will stop when the allocated buffers are full. The problem here was that you could end up in a situation where write_ind pointed past the end of buffers which caused the problems when generating the boost mutable buffers.

wanderingbort: > you could end up in a situation where write_ind pointed past the end of buffers

have we identified this situation though?  The net-plugin never "directly writes" it only indirectly writes through these boost buffers and then returns how much it had written. In turn this is the only call to `advance_write_ptr`.  So, either the calling code is not respecting the conventions of boost and boost was giving us a larger bytes read than it actually wrote to, or there is some other problem going on.

PaulCalabrese: Let me be clearer, i wanted to maintain the invariant that write_ptr() always points to allocated memory. This change assures that.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1228,44 +1229,48 @@ namespace eosio {
...
-                auto index = conn->pending_message_buffer.read_index();
-                conn->pending_message_buffer.peek(&message_length, sizeof(message_length), index);
-                if (bytes_in_buffer >= message_length + message_header_size) {
-                  conn->pending_message_buffer.advance_read_ptr(message_header_size);
-                  if (!conn->process_next_message(*this, message_length)) {
-                    return;
-                  }
-                } else {
-                  conn->pending_message_buffer.add_space(message_length - bytes_in_buffer);
+              FC_ASSERT(bytes_transferred <= conn->pending_message_buffer.bytes_to_write());
```
wanderingbort: I realize I put this in with an earlier commit, but this assert will bubble out of this async callback and potentially wreak havok on the asio system.  We should wrap the contents of the handler lambda with a try-catch that guarantees this handler doesn't throw

plugins/net_plugin/include/eos/net_plugin/message_buffer.hpp
wanderingbort: If that is the case, then I suggest that `advance_write_ptr` which is used to record a write _that has already happened_  ASSERT that it is not advancing past the end.   Its invariant is that you aren't trying to advance passed your requested space.

`write_ptr()` should be the method that maintains about what it can/cannot return.  So, detect a write pointer at the end of the buffer and expand it there. 

This way, 
 * you don't waste a segment of memory just to store a "Full" buffer. 
 * you notify the callers when they are doing something the shouldn't be doing (like advancing a write pointer past where they have requested space via `add_space` or `add_buffer_to_chain`)
 * and you maintain the invariant you are worried about

In all honesty, we already have 2 methods for explicitly requesting space and doing it implicitly on either `write_ptr()` or `advance_write_ptr()`  just seems unnecessary.  That we are adding this to either as opposed to throwing asserts for mis-use of the explicit API is just going to cover up caller errors creating a tight coupling between the callers and some undocumented implicit behavior.

PaulCalabrese: I added a unit test that exposed the problem (and would cause your assert when generating boost buffers). If boost async was filling the current buffers (as it would in a stress test) and we did not read all the data, we would pass invalid pointers on the next async_read_some() call.

wanderingbort: I am not interested in delaying the fix for the test net any longer.

PaulCalabrese: I could add the buffer growth in two other places: write_ptr() and get_buffer_sequence_for_boost_async_read() which are the two places that the pointer indicated by write_ind can escape the class, but doing it in advance_write_ptr() seemed to be more consistent. If you have read enough to fill your buffer, you're probably going to need another buffer in the sequence to contain the rest of a partial message (unless you were lucky and got a whole message). So in the nominal case it won't be an extra buffer allocation.

Issue #802, opened 2017-11-28
DAWN-213 ⁃ add more information to net log messages


pmesnier:
Enhance elog messages and exception reports.

heifner:
Note FC_CAPTURE_AND_LOG macro can be very useful for this.

andriantolie:
@pmesnier is this issue still relevant?

pmesnier:
no

PR #805, opened 2017-11-28
error message enhancements.

pmesnier:
added connection identifier and block id prefix as appropriate to error messages. #802 
also commented out the 0-delay connect retry code.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -599,6 +599,7 @@ namespace eosio {
       if (response_expected) {
         response_expected->cancel();
       }
+      pending_message_buffer.reset();
     }
 
   void connection::txn_send_pending (const vector<transaction_id_type> &ids) {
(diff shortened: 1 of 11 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1185,9 +1185,9 @@ namespace eosio {
                                        ( "peer", c->peer_name())("error",err.message()));
                                  c->connecting = false;
                                  c->close();
-                                 if (started_sessions == 0) {
+                                 /*if (started_sessions == 0) {
                                    connect (c);
-                                 }
+                                 }*/
```
heifner: This was also removed in Dan's changes. Do you want to remove it with this PR?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1493,7 +1493,7 @@ namespace eosio {
           try {
             b = cc.fetch_block_by_id(blkid);
           } catch (const assert_exception &ex) {
-            elog( "caught assert on fetch_block_by_id, ${ex}",("ex",ex.what()));
+            elog( "caught assert on fetch_block_by_id, ${ex}",("ex",ex.to_string()));
```
heifner: I think you would be better served using FC_CAPTURE_AND_LOG macro for here and all these catch blocks.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1232,8 +1232,8 @@ namespace eosio {
       conn->socket->async_read_some(
         conn->pending_message_buffer.get_buffer_sequence_for_boost_async_read(),
         [this,c]( boost::system::error_code ec, std::size_t bytes_transferred ) {
+          connection_ptr conn = c.lock();
           if( !ec ) {
-            connection_ptr conn = c.lock();
             if (!conn) {
```
heifner: This `!conn` needs to be moved out of if as well otherwise your `conn-peer_name()` below could be a problem.

plugins/net_plugin/net_plugin.cpp
pmesnier: I added it here to avoid the spinning on startup. I'm trying to take the smallest set of changes that I can put on the testnet right away to hopefully resolve the network stability issues. I do not want to mess with the new net_api code he also added to his PR. Not at this moment at least.

plugins/net_plugin/net_plugin.cpp
pmesnier: no, if conn is false, the function returns immediately and won't hit my peer_name.

plugins/net_plugin/net_plugin.cpp
pmesnier: right. and after I am done testing I am happy to make this change.

plugins/net_plugin/net_plugin.cpp
pmesnier: nevermind. I get what you mean.

heifner:
What happened here? Looks like you pulled in a number of already merged changes.

pmesnier:
yes. due to the merge conflict, I had to do a pull to synchronize my branch with github's. Since I did a rebase origin/master, that's why this looks crazy. I'm not going to apply this PR to the testnet now.
I'll probably  abandon this branch and try again later.

PR #814, opened 2017-11-29
STAT-214: Add plugin for generating test transactions inside of eosd

spoonincode:
txn_test_gen_plugin is a plugin that can generate test transactions at a configurable rate in transactions/sec. Since the transactions are generated internal to eosd, there is very little overhead compared to generating the transactions via eosc or other RPC mechanisms. This current implementation is a little rough; only having a rather coarse adjustment for transactions/sec and limiting it to a maximum of 400.


programs/eosd/main.cpp
```diff
@@ -36,6 +37,8 @@ int main(int argc, char** argv)
       app().register_plugin<producer_plugin>();
       app().register_plugin<account_history_api_plugin>();
       app().register_plugin<db_plugin>();
+      app().register_plugin<wallet_api_plugin>();
```
heifner: This line is a duplicate.

Issue #813, opened 2017-11-29
STAT-216 ⁃ net code crash on server side after quick reconnection


blockone-syncclient:
Have two nodes running eosd, node A and node B. Node A is running as a producer and listening for connections. Node B is connected to node A. Node B begins to generate transactions and send them to node A. I Ctrl-C node B and it cleanly shuts down. However, if I restart node B and have it begin to generate transactions within a few seconds, node A crashes.

The backtrace tends to always look like the attached.

ATC:
First, run the ATC from STAT-214. leave it running for a few minutes.

While waiting, prepare a script that terminates the node B instance, waits 2 seconds, then restarts node B with --resync.
```
kill `cat tn_data_01/eosd.pid`
sleep 2
programs/eosd/eosd --resync --data-dir tn_data_01
```
use the tail command from stat-214 to verify transactions are still being pumped. if not, use 

curl --data-binary '["one", 400]' http://localhost:8889/v1/txn_test_gen/start_generation

to get them going. Then run your script from above.

In a passing test, node A will stilll be running.

┆Attachments: <a href="https:&#x2F;&#x2F;blockone.atlassian.net&#x2F;secure&#x2F;attachment&#x2F;10065&#x2F;reconnectcrash.txt">reconnectcrash.txt</a>

blockone-syncclient:
➤ Phil Mesnier commented:

Phil Mesnier, [Nov 28, 2017 at 5:47:35 PM]:
Did that test run on the most recently (tagged 11_28_2017) changed code? I just want to confirm that the line numbers in the stack line up with the current code.

one more thing. can you add “—log-level-net-plugin debug” on your eosd A command line, or in its config.ini, and run this again? then attach the resulting stderr output to the issue.

blockone-syncclient:
➤ Phil Mesnier commented:

What was happening and how I fixed it:

Synchronizing a new peer involves three fundamental steps:

  * send blocks up to the last irreversible block, iterating as many times as necessary to get caught up.
  * send blocks from the LIB to the head block. This is handled differently since different peers may have different chains of speculative blocks. So in this case each supplying peer works back from its own head to the LIB and supplying these blocks to the syncing peer.
  * Finally send any unblocked transactions, so that they are ready to be blocked by the new producer if necessary.

[...]

PR #820, opened 2017-11-29
safe this captures

wanderingbort:
this is related to #813 and STAT-216


plugins/net_plugin/net_plugin.cpp
```diff
@@ -887,12 +900,22 @@ namespace eosio {
...
+
+                                      my_impl->local_txns.modify(tx, update_in_flight(-1));
+                                      if (ec) {
+                                         elog("Error sending txn to ${p}: ${msg}", ("p", conn->peer_name())("msg", ec.message()));
+                                      }
+                                      conn->send_next_message();
+                                   } catch (...) {
+
+                                   }
                                 });
```
heifner: Line 928 and 933 also.

Issue #743, opened 2017-11-20
STAT-188 ⁃ Restart of eosd with transactions in reversible blocks causes problems


blockone-syncclient:
When restarting eosd without --resync-blockchain or --replay-blockchain, eosd will fail to get back in sync because it is trying to create transaction_history_object objects that already exist. This will result in exceptions in the log of "could not insert object, most likely a uniqueness constraint was violated".

This only happens if the transactions are in the reversible portion of the blockchain when the process is killed.

ATC

This is probably best tested on a local network.
Bring up a multi-node network.
Perform some activities on the network.
Kill on of the nodes (kill <pid>)
Restart the killed node without either of --resync-blockchain or --replay-blockchain.
Verify that the nodes come back into sync
Verify that things are functional by performing more activities.

heifner:
Verified by QA.

PR #825, opened 2017-11-30
STAT-188: GH-743: Fix eosd restart issues

PaulCalabrese:
Fix multiple problems that were preventing restart of eosd without --resync:
- transaction_history_object and account_transaction_history_object were both trying to create duplicates for transactions in reversible blocks. For now, we simply ignore the exceptions and keep the objects in shared memory
- fix multiple problems with signed_transaction marshalling
- correct handling of signed_blocks

### DIFF
--- a/plugins/account_history_plugin/account_history_plugin.cpp
+++ b/plugins/account_history_plugin/account_history_plugin.cpp
@@ -301,17 +301,23 @@ void account_history_plugin_impl::applied_block(const signed_block& block)
             if (check_relevance && !is_scope_relevant(trx.scope))
                continue;
 
-            db.create<transaction_history_object>([&block_id,&trx](transaction_history_object& transaction_history) {
-               transaction_history.block_id = block_id;
-               transaction_history.transaction_id = trx.id();
-            });
+            // TODO: Find a better way of avoiding duplicates on restart. For now swallow exception.
+            try {
+              db.create<transaction_history_object>([&block_id,&trx](transaction_history_object& transaction_history) {
+                 transaction_history.block_id = block_id;
+                 transaction_history.transaction_id = trx.id();
+              });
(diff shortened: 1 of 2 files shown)


plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -301,17 +301,23 @@ void account_history_plugin_impl::applied_block(const signed_block& block)
             if (check_relevance && !is_scope_relevant(trx.scope))
                continue;
 
-            db.create<transaction_history_object>([&block_id,&trx](transaction_history_object& transaction_history) {
-               transaction_history.block_id = block_id;
-               transaction_history.transaction_id = trx.id();
-            });
+            // TODO: Find a better way of avoiding duplicates on restart. For now swallow exception.
+            try {
```
heifner:
```
auto trx_obj_ptr = db.find<transaction_history_object, by_trx_id>(trx.id());
if (trx_obj_ptr == nullptr) {
...
}
```

plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -301,17 +301,23 @@ void account_history_plugin_impl::applied_block(const signed_block& block)
...
             {
-               db.create<account_transaction_history_object>([&trx,&account_name](account_transaction_history_object& account_transaction_history) {
+              // TODO: Find a better way of avoiding duplicates on restart. For now swallow exception.
+              try {
+                db.create<account_transaction_history_object>([&trx,&account_name](account_transaction_history_object& account_transaction_history) {
                   account_transaction_history.name = account_name;
                   account_transaction_history.transaction_id = trx.id();
-               });
+                });
+              } catch (std::logic_error&) { }
```
heifner: With the above change, I don't think you need this one.

PR #829, opened 2017-11-30
condenced -> condensed

yithump:


heifner:
The html is generated. Please submit PR from source.

PR #832, opened 2017-11-30
potential fix for de-sync due to subjective differences in time

bytemaster:


libraries/chain/chain_controller.cpp
```diff
@@ -971,8 +973,7 @@ void chain_controller::apply_message(apply_context& context)
              : _skip_flags & created_block
                ? _create_block_txn_execution_time
                : _txn_execution_time;
-       const bool is_received_block = _skip_flags & received_block;
-       wasm_interface::get().apply(context, execution_time, is_received_block);
+       wasm_interface::get().apply(context, execution_time, is_producing() );
```
brianjohnson5972: This change to use is_producing means that table validation logic will NOT be performed for pushed_transactions, so it can pass, then when the block is produced it could then be rejected.  Is there a reason we would want to not fail a bad table access and then fail it?

bytemaster: what do you mean by bad table access?  If it is objective then it shouldn't be based on is_receiving_block.

brianjohnson5972: ABI table verification.  When we discussed the requirements you said that for received blocks we could skip the check.  If not then the flag should always be false or removed and we perform the check always.

bytemaster: table verification should only occur when subjective time is considered, so my change is good.

PR #835, opened 2017-11-30
fix for notice forwarding

pmesnier:
OK, this should prevent notice forwarding all together. Which is OK for now since the mesh connects all producers to each other.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1516,6 +1516,7 @@ namespace eosio {
       // notices of previously unknown blocks or txns,
       //
       fc_dlog(logger, "got a notice_message from ${p}", ("p",c->peer_name()));
+      bool can_fwd = false;
```
brianjohnson5972: can we get a comment here or below, indicating we are not currently setting it?

pmesnier: It's going to take a little more time to figure out the safe way to determine whether or not to forward notifications.

PR #838, opened 2017-12-01
Don't copy entire send_buffer when sending a network message

spoonincode:
Fix a huge performance penalty.


heifner:
No longer needed since included in #834

PR #840, opened 2017-12-01
Test design no longer consistent with ABI Table validation only on bl…

brianjohnson5972:
…ock generation after commit to prevent ABI Table verification on pushed transactions.


brianjohnson5972:
Closing this and fixing tests with #839 that also has failing tests.

PR #843, opened 2017-12-01
Check the incoming message length before allocating memory for the me…

spoonincode:
…ssage

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1311,6 +1311,11 @@ namespace eosio {
                     uint32_t message_length;
                     auto index = conn->pending_message_buffer.read_index();
                     conn->pending_message_buffer.peek(&message_length, sizeof(message_length), index);
+                    if(message_length > def_send_buffer_size*2) {
+                        elog("incoming message length unexpected (${i})", ("i", message_length));
+                        close(conn);
+                        return;
+                    }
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1311,6 +1311,11 @@ namespace eosio {
                     uint32_t message_length;
                     auto index = conn->pending_message_buffer.read_index();
                     conn->pending_message_buffer.peek(&message_length, sizeof(message_length), index);
+                    if(message_length > def_send_buffer_size*2) {
```
wanderingbort: instead of `def_send_buffer_size*2` can we use `app().find_plugin<chain_plugin>()->chain().get_global_properties().max_blk_size * SAFETY_FACTOR` (or some terse version thereof)

A block should be the largest message we receive, and this would make this tolerance "elastic" based on the consensus of producing nodes.

Once voting is enabled, or if we wanted to up the block size and redeploy the test net in the future this would "automatically adjust" as opposed to needing manual tweaks

wanderingbort: I should say, I'm fine with the fix as is.

spoonincode: Agreed, but def_send_buffer_size is similarly disjoint from the max_blk_size. Since the RX buffer is unbounded, all that's important today is that I can atleast receive as much as I can transmit.

PR #849, opened 2017-12-02
P2p launcher odds ends 842

pmesnier:
Add's fix for STAT 220, 228, plus named instance launch to support round robin upgrades

### DIFF
--- a/programs/launcher/main.cpp
+++ b/programs/launcher/main.cpp
@@ -338,6 +338,7 @@ struct launcher_def {
   last_run_def last_run;
   int start_delay;
   bool nogen;
+  bool add_enable_stale_production;
   string launch_name;
   string launch_time;
   server_identities servers;
@@ -409,6 +410,7 @@ launcher_def::initialize (const variables_map &vmap) {
         exit (-1);
       }
     }
+    add_enable_stale_production = true;
   }
 
   using namespace std::chrono;
@@ -571,6 +573,7 @@ launcher_def::define_network () {
   }
   stage /= bf::path("staging");
   bf::create_directory (stage);
+  cout << "staging path = " << stage << endl;
 
   if (per_host == 0) {
     host_def local_host;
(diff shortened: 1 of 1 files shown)


programs/launcher/main.cpp
```diff
@@ -326,6 +339,7 @@ struct launcher_def {
   last_run_def last_run;
   int start_delay;
   bool nogen;
+  bool add_enable_stale_production;
```
heifner: These uninitialized bools and ints worry me. add_enable_stale_production is not set unless mode is passed. Which means checking it is undefined and could crash or assume true or false.

programs/launcher/main.cpp
```diff
@@ -397,6 +411,7 @@ launcher_def::initialize (const variables_map &vmap) {
         exit (-1);
       }
     }
+    add_enable_stale_production = true;
```
heifner: Why is this true only if mode is passed?

programs/launcher/main.cpp
```diff
@@ -1044,27 +1060,49 @@ launcher_def::launch (eosd_def &instance, string &gts) {
   last_run.running_nodes.emplace_back (move(info));
 }
 
+#if 0
+void
+launcher_def::kill_instance(eosd_def, string sig_opt) {
+}
+#endif
```
heifner: Can we just remove this?

programs/launcher/main.cpp
pmesnier: I only want this to be true when we are starting a new testnet from scratch. Otherwise it should use --replay or --resync.

programs/launcher/main.cpp
pmesnier: that's a placeholder for the new feature. I'll replace these commented out code later. Since you've already tagged, let's just leave this be and I'll update the pr with new code soon,

PR #869, opened 2017-12-05
STAT 247 - Remove shadowed variable in launcher.

jgiszczak:
(Incorrectly tied to STAT 250, STAT 247 is the appropriate ticket)
Remove shadowed variable in launcher.
Remove end of string in stream that is causing nohup to not see the full command with & to put process in background.

Branch number is Jira ticket number.  No corresponding GH number.

### DIFF
--- a/programs/launcher/main.cpp
+++ b/programs/launcher/main.cpp
@@ -458,8 +458,8 @@ launcher_def::initialize (const variables_map &vmap) {
   if (prod_nodes > total_nodes)
     total_nodes = prod_nodes;
 
-  char * erd = getenv ("EOS_ROOT_DIR");
-  if (erd == 0) {
+  erd = getenv ("EOS_ROOT_DIR");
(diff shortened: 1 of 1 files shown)


programs/launcher/main.cpp
```diff
@@ -458,8 +458,8 @@ launcher_def::initialize (const variables_map &vmap) {
   if (prod_nodes > total_nodes)
     total_nodes = prod_nodes;
 
-  char * erd = getenv ("EOS_ROOT_DIR");
-  if (erd == 0) {
+  erd = getenv ("EOS_ROOT_DIR");
```
heifner: `getenv()` can return NULL. And storing NULL in `std::string` is undefined: 

> basic_string( const CharT* s, const Allocator& alloc = Allocator() );
The behavior is undefined if s does not point at an array of at least Traits::length(s)+1 elements of CharT, including the case when s is a null pointer.

PR #871, opened 2017-12-05
fix build.sh and install_dependencies.sh on macos

piguin:
1) set stop on error, so that the script won't continue to run when error happens
2) added missing dependencies for mac: cmake libmongoc mongo-cxx-driver doxygen
3) deleting /usr/local/binaryen before copy, so that the script runs idempotently
4) link files under /usr/local/binaryen/bin to /usr/local/bin, instead of /usr/local (assuming that was a mistake)


pacificcode:
These issues have been addressed by PR #1370

Issue #864, opened 2017-12-04
DAWN-238 ⁃ Connecting an external eosd to testnet does not work.


blockone-syncclient:
Need the ability of an external eosd to connect to the testnet.

requires a fix to net_plugin.cpp to fix a segv when restarting synchronization after a connectivity break.

ATC - find config file in the attachment, copy it to ./data-dir
run programs/eosd/eosd --resync --data-dir ./data-dir
once it is caught up, disrupt its connection to its peer, either by pulling the ethernet wire or turning off wifi.
At some point, the console output shows an "operation timed out" message. 
Reconnect the cable or turn on wifi.
Observe that the eosd will catch back up with the testnet without crashing.

┆Attachments: <a href="https:&#x2F;&#x2F;blockone.atlassian.net&#x2F;secure&#x2F;attachment&#x2F;10097&#x2F;config.ini">config.ini</a> | <a href="https:&#x2F;&#x2F;blockone.atlassian.net&#x2F;secure&#x2F;attachment&#x2F;10101&#x2F;genesis.json">genesis.json</a>

blockone-syncclient:
➤ Phil Mesnier commented:

from Telegram:
Phil Mesnier, [Dec 5, 2017 at 5:49:11 PM]:
OK. So there is now a code change for the external p2p connectivity.

tuns out that when a connection closed, it was canceling and resetting the response_expected timer, but when the connection restarted, the timer was not re-initialized so the first time a timer was required, blamo.

blockone-syncclient:
➤ Andrianto Lie commented:

Pass.

By the way, genesis.json that is the same as the one in testnet is required for this to pass. So, if public wants to connect to the testnet, the genesis.json in github will need to be updated.  [^genesis.json]

PR #880, opened 2017-12-06
#864 fix reconnection sync crash and bump the sync fetch span to 100 …

pmesnier:
…blocks


pmesnier:
note my editor is set to nuke trailing whitespace

plugins/net_plugin/net_plugin.cpp
```diff
@@ -632,7 +632,6 @@ namespace eosio {
       out_queue.clear();
       if(response_expected) {
         response_expected->cancel();
-        response_expected.reset();
```
heifner: This was added by: PR #831 .

Should it instead be:
`response_expected.reset(new boost::asio::steady_timer(app().get_io_service()));`

pmesnier: no, I am simply not getting rid of the current timer instance.

PR #883, opened 2017-12-06
Extend launcher to generate logging.json.

jgiszczak:
Revise later to deploy to data directory when eosd supports it.

Does what it says on the tin.

### DIFF
--- a/programs/launcher/main.cpp
+++ b/programs/launcher/main.cpp
@@ -21,6 +21,7 @@
 #include <fc/io/json.hpp>
 #include <fc/network/ip.hpp>
 #include <fc/reflect/variant.hpp>
+#include <fc/log/logger_config.hpp>
 #include <ifaddrs.h>
 #include <sys/types.h>
 #include <netinet/in.h>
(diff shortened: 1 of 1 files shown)


programs/launcher/main.cpp
```diff
@@ -353,8 +354,11 @@ struct launcher_def {
   void define_network ();
   void bind_nodes ();
   host_def *find_host (const string &name);
-  host_def *deploy_config_file (tn_node_def &node);
+  host_def *deploy_config_files (tn_node_def &node);
+  string compose_scp_command (const host_def &host, const bf::path &source,
+                              const bf::path &destination);
```
heifner: At some point need to clean up these files so they follow the style of the rest of the code base. e.g. `const host_def& host, const bf::path& source, const bf::path& destination`

programs/launcher/main.cpp
```diff
@@ -840,6 +866,38 @@ launcher_def::write_config_file (tn_node_def &node) {
...
+  if (!cfg.good()) {
+    cerr << "unable to open " << filename << " " << strerror(errno) << "\n";
+    exit (9);
+  }
+
+  auto log_config = fc::logging_config::default_config();
+  log_config.appenders.push_back(
+        fc::appender_config( "net", "gelf",
+            fc::mutable_variant_object()
+                ( "endpoint", "10.160.11.21:12201" )
```
heifner: Hardcoded? This needs to be configurable.

jgiszczak: Next version.  There's only one gelf server today.

heifner: Hardcoded value will be addressed by: STAT-268

PR #887, opened 2017-12-07
Remove redundant page in doxygen

andriantolie:
EDOCS-73


contracts/eoslib/message.h
```diff
@@ -7,7 +7,7 @@
 
 extern "C" {
    /**
-    * @defgroup messageapi message API
+    * @defgroup messageapi Message API
```
heifner: Do we really want this upper case? the type is 'message' not 'Message'.

contracts/eoslib/transaction.h
```diff
@@ -7,7 +7,7 @@
 
 extern "C" {
    /**
-    * @defgroup transactionapi transaction API
+    * @defgroup transactionapi Transaction API
```
heifner: Do we really want this upper case? the type is 'transaction' not 'Transaction'.

contracts/eoslib/message.h
andriantolie: I'm trying to make it consistent for the smart contract API, currently it looks like this in doxygen page (https://eosio.github.io/eos/):

- Database API - APIs that store and retreive data on the blockchainEOS.IO organizes data according to the following broad structure
- Math API - Defines common math function
- message API - Define API for querying message properties
- Console API - Enables applications to log/print text messages
- Token API - Defines the ABI for interfacing with standard-compatible token messages and database tables
- Builtin Types - Specifies typedefs and aliases
- transaction API - Define API for sending transactions and inline messages
- System API - Define API for interating with system level intrinsics

Should it be all lowercase instead then?

heifner: Thanks for this clarification, I see now what you mean. Since all types are now lower case it is probably fine that we upper case them when talking generically about them as opposed to talking explicitly about the struct. So lets go with your changes on this PR.

PR #889, opened 2017-12-07
STAT-205, STAT-198, STAT-197

pavybez:
Simple fixes for benchmark funcationality.

### DIFF
--- a/programs/eosc/main.cpp
+++ b/programs/eosc/main.cpp
@@ -65,6 +65,7 @@ Options:
 #include <string>
 #include <vector>
 #include <boost/asio.hpp>
+#include <boost/format.hpp>
 #include <iostream>
 #include <fc/variant.hpp>
 #include <fc/io/json.hpp>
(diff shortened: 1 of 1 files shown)


programs/eosc/main.cpp
```diff
@@ -959,15 +960,33 @@ int main( int argc, char** argv ) {
    benchmark_setup->add_option("owner", owner_key, localized("The owner key to use for account creation"))->required();
    benchmark_setup->add_option("active", active_key, localized("The active key to use for account creation"))->required();
    add_standard_transaction_options(benchmark_setup);
+
    benchmark_setup->set_callback([&]{
+      auto controlling_account_arg = fc::mutable_variant_object( "controlling_account", string("inita"));
```
heifner: I think `inita` needs to be an argument since you prompt for owner and active keys. As it is now owner and active keys would have to be the keys of inita.

programs/eosc/main.cpp
```diff
@@ -988,6 +1007,7 @@ int main( int argc, char** argv ) {
         transaction_set_reference_block(trx, info.head_block_id);
         sign_transaction(trx);
         batch.emplace_back(trx);
+	info = get_info();
```
heifner: Please cleanup indentation here and a few other places.

Issue #903, opened 2017-12-08
STAT-274 ⁃ eosd sync issues for hotfix 1


blockone-syncclient:
Phil has been working on some sync issues for the december hotfix #1
On Dec. 7, testnet B node 4 went out of sync. several attempts to restart it proved problematic as the sync feature continually got out-of-order blocks, or in some cases simply stopped dead.

After log analysis, code review, and localized testing, It was determined that the sync manager object was still requesting concurrent sets of blocks from multiple peers, rather than the intended behavior of receiving one chunk at a time but varying the intended requester so as to avoid excessive burden to a single peer.

blockone-syncclient:
➤ Corey Lederer commented:

Phil and I decided to close this and add the description elements directly to STAT-276, as they were overlapping tickets.

coreylederer:
Closing.  This is now a duplicate of #916

PR #904, opened 2017-12-09
#903 STAT 274

pmesnier:
refactor the sync manager to do a round-robin block retrieval from only one peer at a time. added more logging too.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -468,6 +468,8 @@ namespace eosio {
     void stop_send();
 
     void enqueue( const net_message &msg );
+    void cancel_sync(const string &reason);
+    void cancel_fetch();
     bool enqueue_sync_block();
     void send_next_message();
     void send_next_txn();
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1071,58 +1115,70 @@ namespace eosio {
...
+        if( ci->current()) {
+          hello.generation = ++ci->sent_handshake_count;
+          fc_dlog(logger, "send to ${p}", ("p",ci->peer_name()));
+          ci->enqueue( hello );
+        }
       }
-      assign_chunk( c);
-      request_next_chunk();
     }
+    ilog( "Catching up with chain, our last req is ${cc}, theirs is ${t}", ( "cc",sync_last_requested_num)("t",target));
```
jgiszczak: fc_dlog(logger

pmesnier: No, I want the ilog to be used to report all these occurrences. They should be infrequent.

PR #907, opened 2017-12-10
Fix reasons to add plugins of EOS Wallet Process

SheldonHH:
clearer explanation and fix confusion
image


heifner:
This is generated html and we are working on moving this information into the wiki.

PR #920, opened 2017-12-11
Refactor install_dependencies.sh

liuchengxu:
- Split install_dependencies.sh into install_dependencies_macos.sh and
install_dependencies_ubuntu.sh.

- Use a set of functions to install the whole dependecies, which fits
into the instructions in README.md

- Only install the missing dependecies in macOS. If the tool has been
installed, there is no need to install again.


kesar:
👍 nice refactoring!

it would be interesting to get here the problems solved by https://github.com/EOSIO/eos/pull/871

liuchengxu:
Thanks for your review @kesar . I'd love to update this PR if allowed.

pacificcode:
This request has been addressed by PR #1370

PR #940, opened 2017-12-13
Review Smart Contract API Doxygen

andriantolie:
Revisit Smart Contract API Doxygen and update them

EDOCS-86


andriantolie:
PR Updated.
Apologize for the compile error. I was trying to switch the order of table in db.hpp so single index table appears before dual index table but I didn't notice that single index table depends on dual index table template. Thanks goodness the compilation error is caught by travis.

PR #944, opened 2017-12-13
Testnet eosd script 282

jgiszczak:
Script and accompanying configuration for connecting a local eosd to the public testnet.


README.md
```diff
@@ -417,10 +418,55 @@ To confirm the nodes are running, run following `eosc` commands:
...
 
 You can read more on launcher and its settings [here](https://github.com/EOSIO/eos/blob/master/testnet.md)
 
+<a name="publictestnet"></a>
+## Running a local node connected to the public testnet
+
+To run a local node connected to the public testnet operated by block.one, a script is provided.
+
+```bash
+cd ~/eos/build/scripts
```
brianjohnson5972: I think this should assume that the user is in the current directory.  If not that then there should be something like <EOS_REPO> or something, not an expected directory to have checked out to.

jgiszczak: The entire README is written dictating the directory to check out to.  My addition is in keeping with that assumption.

thomasbcox: Yes, issuing a specific `cd <dir>` command in the documentation is our awkward standard of the moment, pending the work of STAT-227 to clean up directories and directory references.

jgiszczak: I actually prefer specific, literally correct instructions in cases like this.  Inserting <EOS_REPO> all over the place, then having to explain about substituting your actual path, is a worse solution, to my mind.  If you need to follow the instructions slavishly, you're better off with instructions that work without modifications.  If you're comfortable enough with Linux that you're using the instructions merely as a guide, making your own substitutions to paths of personal preference, then having <EOS_REPO> splattered throughout the text won't help you, and will hurt the people who are less familiar with the Linux command line.

PR #952, opened 2017-12-14
Add subcommands in eosc

SheldonHH:
- `-p` replace port to permission
- `-S` add scope
- `--password` add password


SheldonHH:
partially solve #953, wiki still need to be update for `eosc push message`

PR #960, opened 2017-12-15
Adding distributed transfer transaction validation test.

### LINKED ISSUE #739: STAT-187 ⁃ Distributed transfer on multiple node test
Need to add test script to:
1) launch N nodes, create N accounts
2) send a transfer message, from account 1 to account 2, to Node 1
3) repeat incrementing account number and node number (finally wrapping back to account 1, and each message having a different transferred balance)
4) after block production has been performed, verify that each account has the expected balance based on the transfers

Ideally this test will be configurable, but if it takes too long, than just 3 accounts and 3 Nodes will be sufficient.

ATC:

* Verify travis build is passing test.
* Verify test `tests/distributed-transactions-test.py` passes.

cj-oci:
Adding distributed transfer transaction validation test (STAT 187). Add python3 dependency to travis script. Explicitly invoke python3 in test script. Cmake copies test script to builds directory. Test driver (run_tests.sh) tracks test failures and returns failures count.

Resolves #739

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
             - ninja-build
             - libgmp-dev
             - libclang-4.0-dev
+            - python3
     - os: osx
       osx_image: xcode9.1
 before_install:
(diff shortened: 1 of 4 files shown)


heifner:
Close/open to trigger a fresh merged travis build.

PR #978, opened 2017-12-19
Update config.ini

sergmetelin:
Update config.ini to reflect the latest name change from endpoint to address. Was previously breaking the RPC API access from Docker build.


Docker/config.ini
```diff
@@ -30,13 +30,13 @@ http-server-endpoint = 0.0.0.0:8888
 # access-control-allow-credentials = true
 
 # The local IP address and port to listen for incoming connections.
-listen-endpoint = 0.0.0.0:9876
+listen-address = 0.0.0.0:9876
```
heifner: This should be `p2p-listen-endpoint`

Docker/config.ini
```diff
@@ -30,13 +30,13 @@ http-server-endpoint = 0.0.0.0:8888
...
 
 # The IP address and port of a remote peer to sync with.
 # remote-endpoint =
 
-# The public IP address and port that should be advertised to peers.
-public-endpoint = 0.0.0.0:9876
+# The public IP address and port that should be advertized to peers.
+public-address = 0.0.0.0:9876
```
heifner: Should be `p2p-server-address`

PR #983, opened 2017-12-20
p2p stat 291 gh916 3

pmesnier:
address an unfortunate side-effect of a notification storm that was caused by the fix for other sync issues. Also added a new test. The message_storm.sh test needs a success/failure evaluation based on the quantity and repetition of "Catching up" messages.

### DIFF
--- a/tests/p2p_tests/sync/message_storm.sh
+++ b/tests/p2p_tests/sync/message_storm.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+
+###############################################################
+# -p <producing nodes count>
+# -n <total nodes>
+# -s <topology>
+# -d <delay between nodes startup>
+###############################################################
+
+pnodes=21
+topo=mesh
+delay=5
+
+args=`getopt p:n:s:d: $*`
+if [ $? == 0 ]; then
+
+    set -- $args
+    for i; do
+        case "$i"
+        in
+            -p) pnodes=$2;
+                shift; shift;;
+            -n) total_nodes=$2;
+                shift; shift;;
+            -d) delay=$2;
+                shift; shift;;
+            -s) topo="$2";
+                shift; shift;;
+            --) shift;
+                break;;
+        esac
+    done
+else
+    echo "huh we got err $?"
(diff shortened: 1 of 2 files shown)


tests/p2p_tests/sync/message_storm.sh
```diff
@@ -0,0 +1,107 @@
...
+    port=`expr $port + 1`
+done
+
+grep 'producer"' test.out | tee summary | sort -u -k2 | tee unique
+prodsfound=`wc -l < unique`
+lines=`wc -l < summary`
+if [ $lines -eq $total_nodes -a $prodsfound -eq 1 ]; then
+    echo all synced
+#    programs/launcher/launcher -k 15
+#    exit
```
heifner: Did you mean to leave this commented out?

tests/p2p_tests/sync/message_storm.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
+
```
heifner: Can you just make the --eosd "--log-level-net-plugin debug" as an argument to test.sh so we don't have two almost identical scripts?

tests/p2p_tests/sync/message_storm.sh
pmesnier: yes. To get enough run time for properly detecting the message storm.

tests/p2p_tests/sync/message_storm.sh
pmesnier: that's not the only change. The pnodes, topo, and delay values are also modified, and the setting of debug level output is not optional for this test.  I thought about parameters, but it seemed easier and more sure to just make a new test, Perhaps we can parameterize the test.sh more then make message_storm.sh call that.

heifner: pnodes, total_nodes, delay, topo are all already parameters.

heifner: How about you have Ciju combine these tomorrow.

PR #986, opened 2017-12-21
p2p stat 271 gh 916, 4 

pmesnier:
one more round of performance tuning of the synchronization feature, and also dead code removal.
This seems to be working fairly well in testnet 2.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -519,14 +519,15 @@ namespace eosio {
     uint32_t            sync_last_requested_num;
     uint32_t            sync_req_span;
     connection_ptr      source;
+    bool                active;
```
heifner: Need to initialize in the constructor or better yet initialize right here.

pmesnier: got it.

PR #989, opened 2017-12-21
Test script for STAT 300

cj-oci:
This script tests various replay scenarios. It runs a set of transactions across a local cluster followed by transactions validation. Some of the cluster nodes are then killed. A set of transactions is run on the remaining nodes and validated. The previously killed nodes are resurrected with various chain sync arguments "replay-blockchain", "--resync-blockchain" and default. The nodes are allowed to synchronize. A transaction set is run on the cluster and results validated.


heifner:
Looks like these failed on Travis. Please take a look.

PR #1011, opened 2017-12-28
Restore fc thread logging

spoonincode:
Some thread constructs in fc's logging were removed on master. Restore them.

### DIFF
--- a/libraries/fc/src/log/logger_config.cpp
+++ b/libraries/fc/src/log/logger_config.cpp
@@ -91,7 +91,7 @@ namespace fc {
    }
    const string& get_thread_name() {
       static int thread_count = 0;
-      if( thread_name == string() )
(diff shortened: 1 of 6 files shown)


libraries/fc/src/log/logger_config.cpp
```diff
@@ -84,4 +84,15 @@ namespace fc {
...
       return cfg;
    }
+
+   static thread_local std::string thread_name;
+   void set_thread_name( const string& name ) {
+      thread_name = name;
+   }
+   const string& get_thread_name() {
+      static int thread_count = 0;
+      if( thread_name == string() )
```
wanderingbort: While I'm sure the compiler will handle this well enough, `thread_name.empty()` is a more canonical way to perform this check and it won't allocate any temporaries in un-optimized builds.

PR #1023, opened 2018-01-03
Boost 1.66 compatibility, same as master.

jgiszczak:
No changes to tests not present in the noon branch.

### DIFF
--- a/libraries/fc/include/fc/log/appender.hpp
+++ b/libraries/fc/include/fc/log/appender.hpp
@@ -3,7 +3,11 @@
 #include <fc/shared_ptr.hpp>
 #include <fc/string.hpp>
 
+#if BOOST_VERSION >= 106600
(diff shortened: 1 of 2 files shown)


libraries/fc/include/fc/log/appender.hpp
```diff
@@ -3,7 +3,11 @@
 #include <fc/shared_ptr.hpp>
 #include <fc/string.hpp>
 
+#if BOOST_VERSION >= 106600
```
wanderingbort: This sets a pretty fragile precedent however, I think a better long term solution is lower priority at this point than compatibility with boost 1.66.  I will put up a lower priority issue for a refactor and approve this.

spoonincode:
blerg, sorry guys I didn't realize you were doing two PRs for this -- I reapplied the master merge on to noon earlier

heifner:
Closing since already applied on eos-noon.

PR #1032, opened 2018-01-03
Improve transaction synchronization in tests stat301

cj-oci:
The tests would occasionally fail on slower machines due to the nodes being out of sync. This changes here explicitly validate that the nodes state matches expectation. Also refactored the test code into reusable classes.


brianjohnson5972:
Ciju is going to add a flag for running in travis so that if there is an error then it will print out the log files, since you cannot get to them otherwise.  Without the flag it leaves the log files there for you to investigate later.

Issue #1024, opened 2018-01-03
DAWN-397 ⁃ Most Hated Block (MHB) Fix


blockone-syncclient:
Fix for the Most Hated Block (MHB) issues encountered over the holiday break. In block 936299.

Some historical context. On Dec. 26, 2017, a message involving the transfer of currency value between two accounts was submitted. This was not a unique event, nor was it ill formed. Yet the WASM runtime infrastructure is unable to process this message, raising a "integer divide by 0 or integer overflow" exception. The problem is that when this first occurred, 15 of the block producers accepted the original block and made it irreversible. However the remaining 6 producer nodes, 4 11 12 14 15 21, all failed to accept the block. Subsequent efforts to restart those nodes also failed, hence the new designation of "Most Hated Block." Eventually I got the testnet restarted by performing a "brain transplant" copying the blockchain database and log files from one of the working nodes to the failing nodes.

The proposed patch provides a workaround which is to catch the specific exception when thrown by the WASM runtime while not producing new blocks. Then just log that the exception occurred but otherwise don't do anything about it.

ATP: start a eosd instance and connect to testnet 1.
Observe that your node syncs past block \#936299, the most hated block.
Break into your happy dance. \:-)

blockone-syncclient:
➤ Phil Mesnier commented:

Code for review is in github pull request \#1034

PR #1034, opened 2018-01-04
Addressing the MHB issue

pmesnier:
#1024 Fix to allow at least block 936299 to be successfully added to a new node's chain, Also addresses a secondary issue involving blocks with messages that take very close to the max txn processing time, sometimes the deadline is exceeded which then causes a fork and there is no recovering.

First thing is to allow the existing indication of syncing, provided by the net plugin to the chain plugin, to flow into the chain controller's context. This is done by defining a new skip flag and extending the is_producing test. 

Next thing is to use this flag to indicate when to consume wasm-generated exceptions to avoid problems with otherwise irreversible blocks.

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -979,13 +979,15 @@ void chain_controller::apply_message(apply_context& context)
                ? _create_block_txn_execution_time
                : _txn_execution_time;
        try {
-         wasm_interface::get().apply(context, execution_time, is_producing() );
+          wasm_interface::get().apply(context, execution_time, is_producing() );
        } catch (const fc::exception &ex) {
-         if (!is_producing()) { // && ex.cause == Runtime::Exception::Cause::integerDivideByZeroOrIntegerOverflow) {
-           wlog ("apply_message ignoring exception while not producing");
-         }
-         else
-           throw;
+          // The particular exception that is causing havoc is generated in the wasm layer, and is
+          // Runtime::Exception::Cause::integerDivideByZeroOrIntegerOverflow)
+          if (!is_producing()) {
+             wlog ("apply_message ignoring exception while not producing");
+          }
+          else
+             throw;
        }
     }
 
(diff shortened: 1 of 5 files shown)


libraries/chain/chain_controller.cpp
```diff
@@ -977,7 +978,15 @@ void chain_controller::apply_message(apply_context& context)
              : _skip_flags & created_block
                ? _create_block_txn_execution_time
                : _txn_execution_time;
-       wasm_interface::get().apply(context, execution_time, is_producing() );
+       try {
+         wasm_interface::get().apply(context, execution_time, is_producing() );
```
brianjohnson5972: Please fix the indentation to be 3 spaces.

libraries/chain/include/eos/chain/chain_controller.hpp
```diff
@@ -109,7 +109,8 @@ namespace eosio { namespace chain {
             skip_output_check           = 1 << 13, ///< used to skip checks for outputs in block exactly matching those created from apply
             pushed_transaction          = 1 << 14, ///< used to indicate that the origination of the call was from a push_transaction, to determine time allotment
             created_block               = 1 << 15, ///< used to indicate that the origination of the call was for creating a block, to determine time allotment
-            received_block              = 1 << 16  ///< used to indicate that the origination of the call was for a received block, to determine time allotment
+            received_block              = 1 << 16, ///< used to indicate that the origination of the call was for a received block, to determine time allotment
+            irreversible                = 1 << 17  ///< indicates the blcok was received while catching up and is already considered irreversible.
```
brianjohnson5972: blcok should be block

libraries/chain/include/eos/chain/chain_controller.hpp
```diff
@@ -269,8 +270,7 @@ namespace eosio { namespace chain {
          chainbase::database& get_mutable_database() { return _db; }
 
          bool should_check_scope()const                      { return !(_skip_flags&skip_scope_check);                     }
-         bool is_producing()const                            { return _skip_flags & (received_block | pushed_transaction); }
-
+     bool is_producing()const                            { return (_skip_flags & (received_block | pushed_transaction)) && !(_skip_flags & irreversible); }
```
brianjohnson5972: We need to at least have a comment here as to what this method is about, because this indicates that we are "not in the act of producing a block", right?
Also, please fix indentation.

dskvr:
Resolves #1024

PR #1038, opened 2018-01-05
Add set account stake subcommand to eosioc

jgiszczak:
Includes documentation updates (which also require a major search and replace for eosc/eosd).
Also correct error in create account write scope.
Also automated whitespace cleanup and some spelling corrections.


jgiszczak:
To be superseded.

PR #1042, opened 2018-01-05
Eos noon master cherry pick

heifner:
Catching up eos-noon to master.
Does not include:
* tests/*
* scripts/*
* programs/launcher/*
* plugins/net*plugin/*
* .gitignore


heifner:
Not ready for review...

heifner:
Now ready for review.

Docker/eosc.sh
```diff
@@ -4,7 +4,7 @@
 # Go into cmd loop: sudo ./eosc.sh
```
wanderingbort: we should probably rename this and all of its in docker dependencies to `eosioc` to match the binaries

README.md
```diff
@@ -258,15 +258,15 @@ Run the `create` command where `inita` is the account authorizing the creation o
 ./eosc create account inita currency PUBLIC_KEY_1 PUBLIC_KEY_2
```
wanderingbort: need to rename `eos*` binaries to `eosio*` in the docs

plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
```diff
@@ -127,14 +127,14 @@ struct txn_test_gen_plugin_impl {
          if(ec)
             return;
          try {
-            send_transaction();
+            this->send_transaction();
```
wanderingbort: we are pretty sure that the situation causing the need for the explicit `this->` was due to the use of `auto ec` at the time this change was made to master.  Can we reconfirm that this is necessary now that we are explicitly typing the input parameter?  

refer to: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67274

Docker/eosc.sh
heifner: I updated the Docker/* with eosio. I will have our docker guy test and fix as needed.

plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
heifner: Ok. I removed this.

heifner:
Updates ready for review.

Issue #1046, opened 2018-01-08
DAWN-376 ⁃ migrate account_history_plugin to eos-noon branch


blockone-syncclient:
(This ATC needs to be fixed once DAWN-466 has been merged)

ATC:
Can be run on local or public testnet (if public testnet, then node2 will need to be a local eosiod node that connects to the public testnet)
After each interaction using eosioc to communicate with the testnet, record the transaction_id that is returned.
Your eosioc will either connect with node1 or node2.  Steps #1 - #7 can be performed on either node. node1's config.ini file should not have any "filter_on_accounts = ..." lines.  node2's config.ini should have "filter_on_accounts = tester1" and "filter_on_accounts = tester11"
All corresponding private keys need to be imported into wallet for below used public keys.
1) create an account "tester1" (or choose your own name) with different owner and active public keys, using the existing "inita" account
2) create an account "tester2" (or choose your own name) with different owner and active public keys, using the existing "inita" account
3) create an account "tester11" (or choose your own name) with different owner and active key pair, using the existing "inita" account
4) create an account "tester21" (or choose your own name) with the same owner key pair used for "tester1” and the same active key pair used for "tester2”, using the existing "inita" account
5) transfer 100 EOS from "inita" to "tester1"
6) transfer 200 EOS from "inita" to "tester2"
7) transfer 100 EOS from "tester2" to "tester11"
8) run eosioc multiple times passing in ”get transaction " followed by the transaction_ids for #1-#7 above, and verify the returned transaction contains data consistent with the corresponding activity when connecting to node1, and should only return the transactions from #5 and #7 for node2
9) run eosioc passing in "get accounts ” followed by the owner public key for "tester1" and verify that an array with "tester1" and "tester21" is returned for node1, and nothing is returned for node2
10) run eosioc passing in "get accounts ” followed by the active public key for "tester1" and verify that an array with just "tester1" is returned for node1 and nothing for node2
11) run eosioc passing in "get accounts ” followed by the owner public key for "tester2" and verify that an array with just "tester2" is returned for node1 and nothing is returned for node2
12) run eosioc passing in "get accounts ” followed by the active public key for "tester2" and verify that an array with "tester2" and "tester21" is returned for node1 and nothing is returned for node2
13) run eosioc passing in "get accounts ” followed by the owner public key for "tester11" and verify that an array with just "tester11" is returned for node1 and nothing is returned for node2
[...]

PR #1048, opened 2018-01-08
DAWN-376 account history plugin migration

brianjohnson5972:
#1046 
Migrate account history plugin to the eos-noon branch


plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -385,6 +384,25 @@ bool account_history_plugin_impl::is_scope_relevant(const eosio::types::vector<a
...
+      }
+
+      return optional<abi_serializer>();
+   };
+
+   fc::variant pretty_output;
+   abi_serializer::to_variant(txn, pretty_output, resolver);
+   return pretty_output;
+}
+
```
heifner: @wanderingbort What do you think about moving this to abi_serializer?

plugins/account_history_plugin/include/eosio/account_history_plugin/account_transaction_history_object.hpp
```diff
@@ -40,13 +40,12 @@ using account_transaction_history_multi_index = chainbase::shared_multi_index_co
    account_transaction_history_object,
    indexed_by<
       ordered_unique<tag<by_id>, BOOST_MULTI_INDEX_MEMBER(account_transaction_history_object, account_transaction_history_object::id_type, id)>,
-      hashed_non_unique<tag<by_account_name>, BOOST_MULTI_INDEX_MEMBER(account_transaction_history_object, account_name, account_name), std::hash<account_name>>,
-      hashed_unique<tag<by_account_name_trx_id>,
+      ordered_non_unique<tag<by_account_name>, BOOST_MULTI_INDEX_MEMBER(account_transaction_history_object, account_name, name)>,
```
wanderingbort: Even though this isn't touched by contracts, we may as well maintain the best practices for these tables.  This means that all indices need to be deterministic.  In other tables, we have achieved this by only using ordered unique indices and adding in the `id` field as a means to distinguish between two otherwise identical entries.

wanderingbort: same goes for all the indices.

plugins/account_history_plugin/account_history_plugin.cpp
brianjohnson5972: Decision was to handle this in a separate ticket that will make sure that all transaction types can be converted to and from variants.

brianjohnson5972:
Made ordered_non_unique indexes into ordere_unique.  Note that there were 2 that were just renamed, since the account_name and permission_name pair were already unique.

PR #1054, opened 2018-01-09
allow passing binary WASM to eosioc set contract (in addtion to WAST)

spoonincode:
Some toolchains output binary WASM directly; it's nice to not need WASM->WAST translation before setting the contract via eosioc

I know this is a little kludgy since I left in the "reading WAST..." print which isn't always correct now. But this keep the changeset very minimal

### DIFF
--- a/programs/eosioc/main.cpp
+++ b/programs/eosioc/main.cpp
@@ -684,8 +684,11 @@ int main( int argc, char** argv ) {
       fc::read_file_contents(wastPath, wast);
 
       vector<uint8_t> wasm;
-      if(wast.compare(0, 4, "\x00\x61\x73\x6d"))
+      const string binary_wasm_header = "\x00\x61\x73\x6d";
+      if(wast.compare(0, 4, binary_wasm_header)) {
+         std::cout << localized("Using already assembled WASM...") << std::endl;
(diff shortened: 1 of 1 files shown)


programs/eosioc/main.cpp
```diff
@@ -682,8 +682,14 @@ int main( int argc, char** argv ) {
       std::string wast;
       std::cout << localized("Reading WAST...") << std::endl;
       fc::read_file_contents(wastPath, wast);
-      std::cout << localized("Assembling WASM...") << std::endl;
-      auto wasm = assemble_wast(wast);
+
+      vector<uint8_t> wasm;
+      if(wast.compare(0, 4, "\x00\x61\x73\x6d"))
```
wanderingbort: Can we pull `"\x00\x61\x73\x6d"` into something more descriptively named (maybe: `const char wasm_magic_bytes[4]`)  and put a comment on it that points to: 
https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#high-level-structure

spoonincode:
See if the new change is sufficient. I think that's self explanatory enough without citing the WASM spec for every nook and cranny.

PR #1055, opened 2018-01-09
Dev cmake refactoring + deb package

asiniscalchi:
- extract cmake installer code into installer.cmake module
- extract cmake doxygen code into doxygen.cmake module
- reorder for elegance
- installer cmake script is always parsed cause and execution needs an explicit call (cpack) so nothing changes.
- added travis cpack command
- added deb package creation


asiniscalchi:
Failing on linux clang because of:
E: Failed to fetch http://apt.llvm.org/trusty/pool/main/l/llvm-toolchain-4.0/llvm-4.0-dev_4.0~svn305264-1~exp1_amd64.deb  Connection failed

PR #1058, opened 2018-01-09
Reload logging.json on receipt of SIGHUP.

jgiszczak:
logging_conf_loop() duplicates everything initialize_logging() does but
with error handling intentionally omitted from initialize_logging().
Startup requires a good logging.json, but reload tolerates a bad one.
Good daemons always keep running as best they can once started, but can
insist on a good startup environment.

### DIFF
--- a/programs/eosiod/main.cpp
+++ b/programs/eosiod/main.cpp
@@ -31,6 +31,30 @@ namespace fc {
    std::unordered_map<std::string,appender::ptr>& get_appender_map();
 }
 
+namespace detail {
+
+void configure_logging(const bfs::path& config_path)
+{
+  try {
+    try {
+      fc::configure_logging(config_path);
+    } catch (...) {
+      elog("Error reloading logging.json");
+      throw;
+    }
+  } catch (const fc::exception& e) {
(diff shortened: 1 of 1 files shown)


programs/eosiod/main.cpp
```diff
@@ -31,13 +31,47 @@ namespace fc {
...
+    if(!err)
+    {
+      ilog("Received HUP.  Reloading logging configuration.");
+      auto config_path = app().get_logging_conf();
+      if(fc::exists(config_path))
+      {
+        try {
+          fc::configure_logging(config_path);
+        } catch (const fc::exception& e) {
+          elog("Error reloading logging.json");
```
wanderingbort: can we refactor all of the catch-and-log logic into a static function called in a `catch (...)` block that rethrows the exception and translates the type-specific error after printing out the common information of "Error reloading logging.json" ; uses a nested block to deduplicate those messages; OR anything that effectively removes the long-tail of repetitive error handling from what we did `fc::configure_logging(config_path)` from what we need to do next: re-initialize the appenders and re-register the handler.

PR #1059, opened 2018-01-10
Restore test cases of abi_tests.cpp

andriantolie:
DAWN-377

The following pull request update abi_tests.cpp:
- Remove dependency with database_fixture
- Update obsolete reference (native_contract_chain_initializer -> contract/chain_initializer, types/abi_serializer -> contracts/abi_serializer)
- Update the abi types to match the new types available in abi_serializer
- Update abi's field to the new format (from `field: {  name: type }` to `field: [{ name: name1, type: type1 }]`

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -27,8 +27,7 @@ set( CMAKE_CXX_STANDARD 14 )
 
 
 
-file(GLOB UNIT_TESTS "chain_tests/*.cpp")
-file(GLOB UNIT_TESTS "tests/abi_tests.cpp")
+file(GLOB UNIT_TESTS "chain_tests/*.cpp" "tests/abi_tests.cpp")
 if(WASM_TOOLCHAIN)
     file(GLOB WASM_UNIT_TESTS "wasm_tests/*.cpp")
 endif()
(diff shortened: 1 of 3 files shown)


tests/CMakeLists.txt
```diff
@@ -28,12 +28,13 @@ set( CMAKE_CXX_STANDARD 14 )
 
 
 file(GLOB UNIT_TESTS "chain_tests/*.cpp")
+file(GLOB UNIT_TESTS "tests/abi_tests.cpp")
```
wanderingbort: this should be merged with the line above.  `file(GLOB ...` can accept multiple patterns.  See: https://cmake.org/cmake/help/latest/command/file.html

andriantolie:
Pull Request Updated.
Btw, in the documentation of GLOB https://cmake.org/cmake/help/latest/command/file.html, I saw
> Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.
Should we avoid using GLOB and list the file explicitly?

wanderingbort:
> Should we avoid using GLOB and list the file explicitly?
I think we are undecided on that for now.  It will be a question for another issue/PR and shouldn't hold this up any longer.

andriantolie:
Alright since this PR already got approved, I will merge it

PR #1064, opened 2018-01-10
Import eos-walletd from master and rename it to eosio-walletd.

cj-oci:


programs/eosio-walletd/main.cpp
```diff
@@ -0,0 +1,39 @@
+/**
+ *  @file
+ *  @copyright defined in eos/LICENSE.txt
+ */
+#include <appbase/application.hpp>
+
+#include <eosio/http_plugin/http_plugin.hpp>
+#include "eosio/wallet_plugin/wallet_plugin.hpp"
```
wanderingbort: If there is no reason to have these quoted vs angle-bracketed, lets fix that in this import

cj-oci:
Made the requested change regards import quote style.

PR #1066, opened 2018-01-10
@DAWN-391  added sbrk and fixed the function signature of memset

larryk85:
/libraries/chain/wasm_interface.cpp and /libraries/chain/include/eosio/chain/wasm_interface_private.hpp updated.  These will be needed by anyone adding missing api intrinsics.

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -411,7 +411,6 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, array_ptr<U>, size_t
    }
 };
 
-#if 1
 /**
  * Specialization for transcribing memset parameters
  *
@@ -437,7 +436,6 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<char>, int, size_t>, std
       return next_step::template fn<translate_one<Then>>();
    }
 };
-#endif
 
 /**
  * Specialization for transcribing  a pointer type in the native method signature
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -421,9 +421,7 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, array_ptr<U>, size_t
  */
 template<typename Ret>
 struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<char>, int, size_t>, std::tuple<>> {
-   //using next_step = intrinsic_invoker_impl<void_type, std::tuple<>, std::tuple<I32, I32, I32>>;
    using next_step = intrinsic_invoker_impl<Ret, std::tuple<>, std::tuple<I32, I32, I32>>;
-   //using then_type = void_type(*)(wasm_interface &, array_ptr<char>, int, size_t);
    using then_type = Ret(*)(wasm_interface &, array_ptr<char>, int, size_t);
 
    template<then_type Then>
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -101,7 +101,7 @@ struct native_to_wasm {
 };
 
 /**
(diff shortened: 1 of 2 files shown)


libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -411,20 +411,23 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, array_ptr<U>, size_t
...
  * @tparam Inputs - the remaining native parameters to transcribe
  * @tparam Translated - the list of transcribed wasm parameters
  */
-template<>
-struct intrinsic_invoker_impl<void_type, std::tuple<array_ptr<char>, int, size_t>, std::tuple<>> {
-   using next_step = intrinsic_invoker_impl<void_type, std::tuple<>, std::tuple<I32, I32, I32>>;
-   using then_type = void_type(*)(wasm_interface &, array_ptr<char>, int, size_t);
+template<typename Ret>
+struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<char>, int, size_t>, std::tuple<>> {
+   //using next_step = intrinsic_invoker_impl<void_type, std::tuple<>, std::tuple<I32, I32, I32>>;
```
wanderingbort: go ahead and clean up the commented old `using` statements

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -411,20 +411,23 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, array_ptr<U>, size_t
    }
 };
 
+#if 1
```
wanderingbort: please clean up this preprocessor conditional

libraries/chain/wasm_interface.cpp
```diff
@@ -23,6 +23,7 @@
 #include <mutex>
 #include <thread>
 #include <condition_variable>
+//#include <iostream>
```
wanderingbort: no need to keep this around

libraries/chain/wasm_interface.cpp
```diff
@@ -622,11 +623,12 @@ DEFINE_INTRINSIC_FUNCTION0(env,checktime,checktime,none) {
 class context_aware_api {
    public:
       context_aware_api(wasm_interface& wasm)
-      :context(intrinsics_accessor::get_context(wasm).context)
+      :context(intrinsics_accessor::get_context(wasm).context), code(intrinsics_accessor::get_context(wasm).code)
       {}
 
    protected:
-      apply_context& context;
+		wasm_cache::entry& code;
```
wanderingbort: make sure you have "tabs" set to spaces

libraries/chain/wasm_interface.cpp
```diff
@@ -819,10 +822,44 @@ class memory_api {
...
+			constexpr uint32_t 	NBPPL2 = IR::numBytesPerPageLog2;
+			constexpr uint32_t max_mem = 1024 * 1024;
+			const auto 			 default_mem = Runtime::getDefaultMemory(code.instance);
+			const uint32_t		 num_pages = Runtime::getMemoryNumPages(default_mem);
+			// limit this min to 32 bit space
+			const uint32_t 	 min_bytes = (num_pages << NBPPL2) > UINT32_MAX ? UINT32_MAX : num_pages << NBPPL2;
+			static uint32_t 	 _num_bytes = min_bytes;
+			const uint32_t 	 prev_num_bytes = _num_bytes;
+
+			if (Runtime::getMemoryNumPages(default_mem) != num_pages)
```
wanderingbort: this seems redundant given that we get this value as const.

spoonincode:
I think we need to update eoslib/memory.h for the new memset() prototype too

spoonincode:
@wanderingbort you are right -- the design assumes that you're using malloc() to call sbrk() -- the memset occurs in malloc(). This looks like a massive bug since someone could use sbrk() without malloc(). IMO the WASM runtime needs to be refactored to always ensure you're given zeroed memory. I'm going to have to touch that code significantly when changing the virtual mem allocation strategy so I wouldn't bother fixing any of that logic now.

wanderingbort:
Just waiting on a fix from @larryk85 that will deallocate the pages properly before returning the code to the cache

PR #1077, opened 2018-01-12
Scope refactoring and block changes

### LINKED ISSUE #1062: DAWN-466 ⁃ Remove explicit scoping from transactions
# Scope is not the User's concern. 

This closes #1040 

Previously a user would have to guess or have tools infer "scopes" for a transactions.  On the occasion that this inference or guesswork would fail, users would have a pretty bad time trying to figure out the proper arcane scopes to make things work.  There was basically no way for scopes to create anything other than a negative user experience if the user experienced them at all.  As a result, we have move the determination of scopes into the responsibility of a block producer.  This simplifies the guarantees that each party in the network is making:
  * Users: guarantee that this is the set of actions they desire and that the supplied signatures + permissions grant authority to perform those actions
  * Producers: guarantee that blocks are scheduled such that no parallel execution unit (shard) conflicts with another shared in terms of read/write scopes

## Tasks

 - [x] Remove read/write_scope from `transaction`
 - [x] create a first class `shard` type instead of typedef'ing as `vector<transaction_receipt>`
 - [x] add read/write scopes to `shard`
 - [x] Add ability to "record" scopes used as an alternative to enforcing scopes are known when applying transactions
 - [x] produce new blocks such that scopes can be recorded into shards properly, determining them through the "record" feature
 - [x] validate incoming blocks by only allowing access to scopes as defined by the `shard` under which the transactions are being executed
    * if a transaction accesses an unlisted scope (read or write) this should be considered an objective failure of the incoming block
 - [x] remove scope arguments/handling from `eosioc`
 - [x] clean up test cases that were manually setting scopes
 - [ ] clean up documentation and tutorials that mention scopes

wanderingbort:
this resolves #1062

made some questionable changes to account history plugin given that it depended on scopes being account names.


plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -283,45 +283,59 @@ vector<account_name> account_history_plugin_impl::get_controlled_accounts(const
...
       if (trx_obj_ptr != nullptr)
-         continue; // on restart may already have block
+         return; // on restart may already have block
 
-      db.create<transaction_history_object>([&block_id,&trx](transaction_history_object& transaction_history) {
+      db.create<transaction_history_object>([&block_id,&trx_trace](transaction_history_object& transaction_history) {
          transaction_history.block_id = block_id;
-         transaction_history.transaction_id = trx.id();
+         transaction_history.transaction_id = trx_trace.id;
+         transaction_history.transaction_status = trx_trace.status;
```
brianjohnson5972: Do we need to add functionality around transaction_status now, or is it something that is already written up to be added?

brianjohnson5972:
account_history_plugin looks good, just the one question about transaction_status.

libraries/chain/chain_controller.cpp
```diff
@@ -557,6 +548,17 @@ void chain_controller::_apply_block(const signed_block& next_block, uint32_t ski
...
+static void validate_shard_locks(const vector<shard_lock>& locks, const string& tag) {
+   if (locks.size() < 2) {
+      return;
+   }
+
+   for (auto cur = locks.begin() + 1; cur < locks.end(); cur++) {
+      auto prev = cur - 1;
+      FC_ASSERT(*prev != *cur, "${tag} lock \"${a}::${s}\" is not unique", ("tag",tag)("a",cur->account)("s",cur->scope));
+      FC_ASSERT(*prev < *cur,  "${tag} locks are not sorted", ("tag",tag));
+   }
```
heifner: I see `FC_ASSERT` here and `EOS_ASSERT` elsewhere.  They do have slightly diff behavior. Could you give us guidance on when to use each?

plugins/account_history_plugin/account_history_plugin.cpp
wanderingbort: I tried to make the minimal changes necessary, we may want some indexing around transaction status as part of a future issue.

libraries/chain/chain_controller.cpp
```diff
@@ -1368,10 +1355,10 @@ static void log_handled_exceptions(const transaction& trx) {
    } FC_CAPTURE_AND_LOG((trx));
 }
 
-transaction_trace chain_controller::__apply_transaction( transaction_metadata& meta ) {
+transaction_trace chain_controller::__apply_transaction( transaction_metadata& meta, const vector<shard_lock>* allowed_read_locks, const vector<shard_lock>* allowed_write_locks ) {
```
wanderingbort: consolidate parameters

libraries/chain/chain_controller.cpp
```diff
@@ -625,7 +651,6 @@ void chain_controller::__apply_block(const signed_block& next_block)
                // check_transaction_authorization(trx, true);
             } /// for each transaction id
 
-            ++shard_index;
             s_trace.calculate_root();
             c_trace.shard_traces.emplace_back(move(s_trace));
```
wanderingbort: Need to validate that the list of shard locks was concise

libraries/chain/chain_controller.cpp
wanderingbort: I think we should probably try to use EOS_ASSERT for anything which may trip as part of validating the block chain.   Its debatable, but code paths that are only exercised when a producer is producing a block don't benefit as much from EOS_ASSERT.  I have changed many of my FC_ASSERTs to EOS_ASSERTs

PR #1085, opened 2018-01-12
Mongo db plugin to eos-noon

### LINKED ISSUE #1047: DAWN-449 ⁃ Move mongo db_plugin to eos-nxxn branch
To support the web-wallet and faucet for the next testnet move mongodb db_plugin over to eos-nxxn branch.

* Investigate making mongo_db_plugin an optional plugin

ATC:

* Build with -DBUILD_MONGO_DB_PLUGIN=true
* Run tests/eosiod_run_test.py --mongodb test

heifner:
* Moved mongo_db_plugin to eos-noon for use with web-wallet
* Plugins setup to self-register
* mongo_db_plugin not included in build by default, must be enabled via `-DBUILD_MONGO_DB_PLUGIN=true`
* Scripted test will be moved over with other test scripts

Resolves #1047


libraries/chain/chain_controller.cpp
```diff
@@ -1074,6 +1081,11 @@ void chain_controller::_initialize_chain(contracts::chain_initializer& starter)
             transaction_metadata tmeta( genesis_setup_transaction );
             __apply_transaction( tmeta );
          });
+
+         // TODO: Should we write this genesis block instead of faking it on startup?
+         block.input_transactions.emplace_back(genesis_setup_transaction);
+         applied_irreversible_block(block);
```
wanderingbort: in order to help indexers be generic we should have a guarantee that every block has at least `applied_block` called on it and only those that become irreversible eventually have `applied_irreversible_block` called.  

The take away here is that we should be building a full false block complete with trace of transactions to pass to indexers in both callbacks (in the proper order).  This would also allow the construction of merkle proofs that includ genesis transactions.

heifner: Added `block_trace` and call to `applied_block`. Also called `calculate_root()` on `shard_trace`. Anything missing?

heifner:
Ready for review.

PR #1094, opened 2018-01-15
Patch (from Brian J) to enable transfering funds between new accounts.

cj-oci:
This patch is required for newly created accounts to be able to transfer funds to other new accounts. Without this patch only transfers from genesis accounts are possible.


programs/eosioc/main.cpp
```diff
@@ -291,7 +291,7 @@ void create_account(name creator, name newaccount, public_key_type owner, public
       auto active_auth  = eosio::chain::authority{1, {{active, 1}}, {}};
       auto recovery_auth = eosio::chain::authority{1, {}, {{{creator, "active"}, 1}}};
 
-      uint64_t deposit = 1;
+      uint64_t deposit = 1000;
```
wanderingbort: we should add a `deposit` option which defaults to 1000

PR #1101, opened 2018-01-16
use nproc to accelerate the speed.

bamboowind:
nproc - print the number of processing units available. 
we can use this to match different processors.


build.sh
```diff
@@ -99,4 +99,4 @@ fi
 
 # Build EOS
 cmake -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DCMAKE_CXX_COMPILER=${CXX_COMPILER} -DCMAKE_C_COMPILER=${C_COMPILER} -DWASM_LLVM_CONFIG=${WASM_LLVM_CONFIG} -DBINARYEN_BIN=${BINARYEN_BIN} -DOPENSSL_ROOT_DIR=${OPENSSL_ROOT_DIR} -DOPENSSL_LIBRARIES=${OPENSSL_LIBRARIES} ..
-make -j4
+make -j$(nproc)
```
venediktov: For Mac Os this command is invalid 
```sysctl -n hw.physicalcpu```
So , it has to be some logic if/else in this script

pacificcode:
These issues and requests have been addressed by PR #1370

PR #1111, opened 2018-01-16
DAWN-376 account history scope refactor fix

brianjohnson5972:
Fixed refactor of account history processing of applied blocks.
#1046


plugins/account_history_plugin/account_history_plugin.cpp
```diff
@@ -467,7 +467,7 @@ void account_history_plugin::plugin_startup()
    db.add_index<transaction_history_multi_index>();
 
    my->chain_plug->chain().applied_block.connect ([&impl = my](const chain::block_trace& trace) {
-      impl->applied_block(trace.block);
+      impl->applied_block(trace);
```
heifner: I vote we add `explicit` to block.hpp `block_trace(const signed_block& s);` constructor.

brianjohnson5972: I agree, added.

Issue #1071, opened 2018-01-11
DAWN-469 ⁃ Standardizing Currencies


wanderingbort:
# Standard Currencies

In order to facilitate a wealth of contracts, interfaces, wallets, and tools it makes sense for us to have a standardized framework for implementing a currency.  The base EOS.IO tokens should implement this standard.

## Goals

 * Standard identification of a currency
    * which allows for a single contract to operate multiple currencies
    * which optionally allows guarantee of globally unique "ticker" 
 * Standardized way of constructing a `transfer` from an account to another account
    * This may enforce serialization formats
 * Standardized notifications of to/from when `transfer` s happen
 * Standardized access of balances for an account
    * This may enforce serialization formats
 * Standardized access to meta-information about a currency (perhaps in global registry)
    * This may enforce serialization formats

## Implementation Details

There is a reference implementation of a currency which can be used to define the spec [here](https://github.com/EOSIO/eos/blob/c3d336208bab2d081342ae8070f3ada7aad27282/contracts/eoslib/generic_currency.hpp). 

## Sub Tasks
- [ ] Derive standard document
- [ ] apply standard to the base chain tokens

Dexaran:
It should be noted that [ERC20](https://github.com/ethereum/EIPs/issues/20) token standard has a critical issue. This led to the loss of millions of dollars for the Ethereum ecosystem. You can read more about the problem here: https://gist.github.com/Dexaran/ddb3e89fe64bf2e06ed15fbd5679bd20

I strongly recommend to avoid using the ERC20 standard as a prototype model when developing the standard for EOS digital assets.

jgiszczak:
Implemented as eosio.token.

PR #1113, opened 2018-01-17
minimal changes to restore proxy and currency contracts

wanderingbort:
PLEASE NOTE: This does not bring the native currency up to spec with the current contract level generic currency.  It merely restores the unit tests for deferred transactions and currency as simply as possible.  Additional, issues will need to be created to handle further clean-up and the implied refactor to the native chain currency as part of #1071


spoonincode:
I checked this out, compiled it, played with the eoslib changes a little. Don't have a lot of constructive things to say, but it didn't seem to break any thing that I played with.

PR #1114, opened 2018-01-17
make token accept smaller/larger unsigned integral types for NumberType

### LINKED ISSUE #1103: DAWN-502 ⁃ Can't use any type other then uint64_t in eosio::token<uint64_t,N(currency)>
Should it be  allowed using uint32_t and uint128_t and other unsigned values to represent token quantity ? 
Currently only uint64_t compiles .

If that's the case I can make a pull request to address the issue with following features

- Add type traits to smart contract library as a separate file contracts/eoslib/type_traits.hpp ( currently some type traits exist in contracts/eoslib/type.h)
- add static assertion in the body of template class 
- add SFINAE exclusion of constructor if type is not unsigned integral   
- don't allow implicit conversions of integral value in the ctor of eosio::token , same type but smaller length will be allowed though 
- change hardcoded asserts in contracts library 

```
template<typename NumberType, uint64_t currency = N(eos)>
struct token {
static_assert(eosio::is_unsigned<NumberType>::value, "NumberType can only be unsigned number");
.....
.....  
token() : quantity{} {}

template<typename T, typename = typename eosio::enable_if<eosio::is_unsigned<T>::value && !(sizeof(T)>sizeof(NumberType))>::type>
 explicit token( T v ):quantity(v){};


 //NumberType quantity = 0; //use default value initialization  in ctor 
NumberType quantity;
.....
```

A much stricter would be even better  but I think above proposal would suffice ?

```
template<typename T, typename = typename eosio::enable_if<eosio::is_unsigned<T>::value && eosio::is_same<T,NumberType>::type>
 explicit token( T v ):quantity(v){};
```

the proposed  change will require fixing  hardcoded asserts in multiple places as when we change token NumberType compiler will have  to automatically calculate size, see below  
```
-   static_assert( sizeof(bid) == 32+12, "unexpected padding" );
+   static_assert( sizeof(bid) == sizeof(order_id)+sizeof(price)+sizeof(eosio::tokens)+sizeof(time), "unexpected padding" );
```
Let me know what you think.

venediktov:
This pull request should close #1103


contracts/currency/currency.hpp
```diff
@@ -87,7 +87,8 @@ namespace TOKEN_NAME {
    /**
    Assert statement to verify structure packing for account
    **/
-   static_assert( sizeof(account) == sizeof(uint64_t)+sizeof(currency_tokens), "unexpected packing" );
+   //This structure is *not* packed the check will not work for different sizes of currency_tockens 
+   //static_assert( sizeof(account) == sizeof(uint64_t)+sizeof(currency_tokens), "unexpected packing" );
```
wanderingbort: the `master` branch doesn't currently deal with ambiguously packed structure very well.  Does packing the account/transfer etc structs not resolve this without relaxing this static_assert?

contracts/eoslib/token.hpp
```diff
@@ -68,20 +70,22 @@ namespace eosio {
...
-    token(){}
+    token() : quantity{} {}
 
     /**
     * Constructor for token given quantity of tokens available
     * @brief Constructor for token given quantity of tokens available
     * @param v - quantity of tokens available
     */
-    explicit token( NumberType v ):quantity(v){};
+//    template<typename T, typename = typename eosio::enable_if<eosio::is_unsigned<T>::value && eosio::is_same<T,NumberType>::value>::type>
```
wanderingbort: Please remove commented out code

contracts/eoslib/types.hpp
```diff
@@ -89,4 +89,71 @@ namespace  eosio {
    template<typename T> struct remove_reference<const T&> { typedef T type; };
    ///@}
 
+    /**
```
wanderingbort: Please move this to a file specifically for compliance with standard lib features.   And add it to its own Doxygen group whose description describes what these do, _not_ why they are needed.  For instance, "support for compile time C++ standard features"

contracts/currency/currency.hpp
venediktov: Hi @wanderingbort , Yes first though was to pack it as the assert implies it checks for packed structure. However , packed structures are a bit unsafe when using them ( outside of WAST domain ) 
https://stackoverflow.com/questions/8568432/is-gccs-attribute-packed-pragma-pack-unsafe
I just did not want to introduce a bug , however I am ready to look into that and see if in existing code at least it will produce same webassembly.

contracts/eoslib/types.hpp
venediktov: how about eos/contracts/eoslib/type_traits.hpp ?

contracts/eoslib/token.hpp
venediktov: will do

contracts/eoslib/types.hpp
wanderingbort: works for me.

contracts/currency/currency.hpp
wanderingbort: This is a wart for the current release that we have largely addressed in the future releases.   Don't feel too bad about it for now, we will handle merging the intent of your changes into the new version of our standard currency contract.

venediktov: As I suspected when I added struct PACKED(account) LLVM produced extra align instruction align=1 for WAST

before 
```
(i64.load offset=8
```
after packing 

``` 
(i64.load offset=8 align=1
```
So if you don't mind I will not PACK account or transfer for now and just change my comment to 
//TODO: reintroduce static_assert if account and transfer is packed in the future 

This way it's safe with my pull request  as I see quite few use cases when from.ballance and to.ballance are used , currently token<uint64_t , N(currency)> will be aligned correctly but if other types are mixed together then it could lead to a bug with PACKED(account)
Thank you for letting me work on EOS stack.

PR #1117, opened 2018-01-17
write locks can be read locks too

wanderingbort:
when a shard locks a scope for write that should be sufficient for read


spoonincode:
Let's pull it since it clearly fixes a significant problem in noon.

heifner:
This fixes test:  `./tests/trans_sync_across_mixed_cluster_test.sh -p 1 -n 2`

PR #1121, opened 2018-01-17
Dawn 250 gh 1081 Add roll and bounce scripts and ability to call from launcher

PaulCalabrese:
This change adds new roll and bounce scripts which can be remotely invoked via the launcher. Using the --bounce and --roll options, you can now bounce all the nodes in a network or roll them to a new version.

### DIFF
--- a/scripts/tn_roll.sh
+++ b/scripts/tn_roll.sh
@@ -58,15 +58,19 @@ if [ \( -z "$prog" \) -o \( -z "$RD" \) ]; then
 fi
 
 SDIR=staging/eos
+if [ ! -e $SDIR/$RD/$prog ]; then
+    echo $SDIR/$RD/$prog does not exist
+    exit 1
+fi
+
 if [ -e $RD/$prog ]; then
-    s1=`md5sum $RD/$prog`
-    s2=`md5sum $SDIR/$prog`
+    s1=`md5sum $RD/$prog | sed "s/ .*$//"`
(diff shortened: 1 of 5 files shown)


scripts/tn_bounce.sh
```diff
@@ -1,10 +1,47 @@
...
+# tn_bounce is used to restart a node that is acting badly or is down.
+# usage: tn_bounce.sh [arglist]
+# arglist will be passed to the node's command line. First with no modifiers
+# then with --replay and then a third time with --resync
+#
+# the data directory and log file are set by this script. Do not pass them on
+# the command line.
+#
+# in most cases, simply running ./tn_bounce.sh is sufficient.
+#
```
heifner: I would note that this is called from launcher.

scripts/tn_roll.sh
```diff
@@ -0,0 +1,91 @@
...
+# all instances are restarted.
+# usage: tn_roll.sh [arglist]
+# arglist will be passed to the node's command line. First with no modifiers
+# then with --replay and then a third time with --resync
+#
+# the data directory and log file are set by this script. Do not pass them on
+# the command line.
+#
+# in most cases, simply running ./tn_roll.sh is sufficient.
+#
```
heifner: I would add a note that this is used by launcher.

scripts/tn_bounce.sh
PaulCalabrese: Both of these scripts can also be called directly.

PR #1122, opened 2018-01-17
Support Multiple Currencies in a single contract

bytemaster:


contracts/eoslib/vector.hpp
```diff
@@ -13,6 +13,11 @@ namespace eosio {
       public:
          vector(){}
 
+         vector( const T& value ) {
```
wanderingbort: we should consider pulling from the spec for `std::vector` for this constructor so that developers who are used to a certain meaning for "vector" are more apt to find what they expect.

There is this constructor defined in the spec which would be able to provide the functionality :
```
vector( size_type count, const T& value, const Allocator& alloc = Allocator());
```
Naturally we don't have an allocator. 

We do however, have the stubs now for `std::initializer_list<T>` meaning that all calls to this constructor can be changed from `vector(X)` to `vector({X})` and then this constructor could be dropped.

spoonincode:
I ran a test with a two node network (producer & non-producer). Submitted a currency setcode, then a currency issue, then a currency transfer all via non-producing node. Was able to read currency value via get table.

PR #1128, opened 2018-01-18
introducing ctest: test runner by cmake toolset

asiniscalchi:
- refactoring tests/eosd_run_test.sh to be called from every location in filesystem
- tests/eosd_run_test.sh is run by ctest runner
- travis will run it with ctest --output-on-failure

NB. tests can be added to the ctest runner by:
add_test(name executable) in any CMakeLists.txt

https://cmake.org/Wiki/CMake/Testing_With_CTest

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,6 @@ script:
   - mkdir build && cd build
   - ${CMAKE_DIR}cmake -G Ninja -DWASM_LLVM_CONFIG=$TRAVIS_BUILD_DIR/ext/wasm-compiler/bin/llvm-config -DSecp256k1_ROOT_DIR=$TRAVIS_BUILD_DIR/ext -DBINARYEN_ROOT=$TRAVIS_BUILD_DIR/ext/wasm-compiler -DCMAKE_PREFIX_PATH=$TRAVIS_BUILD_DIR/ext -DCMAKE_BUILD_TYPE=Release $EOS_CMAKE_OPTIONS ..
   - ninja -j4
-  - tests/eosd_run_test.sh --host=localhost --port=8888
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/eosd_run_test.py --host=localhost --port=8888'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || EOSLIB=../contracts tests/chain_test'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/slow_test'
@@ -86,4 +85,5 @@ script:
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/p2p_tests/sync/test.sh -p 2 -d 10'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/run_tests.sh'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/eosd_run_mongodb_test.sh'
+  - ctest --output-on-failure
   - cpack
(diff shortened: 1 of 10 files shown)


asiniscalchi:
output:

image

.travis.yml
```diff
@@ -77,13 +77,10 @@ script:
...
-  - tests/eosd_run_test.sh --host=localhost --port=8888
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/eosd_run_test.py --host=localhost --port=8888'
-  - '[ "$TRAVIS_OS_NAME" == "osx" ] || EOSLIB=../contracts tests/chain_test'
-  - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/slow_test'
-  - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/api_test'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/p2p_tests/sync/test.sh'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/p2p_tests/sync/test.sh -p 2 -d 10'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/run_tests.sh'
   - '[ "$TRAVIS_OS_NAME" == "osx" ] || tests/eosd_run_mongodb_test.sh'
+  - ctest --output-on-failure
```
wanderingbort: Does this preserve the early out for OSX: `[ "$TRAVIS_OS_NAME" == "osx" ] || ...` for the encapsulated tests in another place?

programs/launcher/main.cpp
```diff
@@ -372,12 +372,12 @@ struct launcher_def {
   void format_ssh (const string &cmd, const string &host_name, string &ssh_cmd_line);
   bool do_ssh (const string &cmd, const string &host_name);
   void prep_remote_config_dir (eosd_def &node, host_def *host);
-  void launch (eosd_def &node, string &gts);
+  void launch (const string &eosdBinary, eosd_def &node, string &gts);
```
wanderingbort: stylistically we prefer under_scores to camelCase

.travis.yml
asiniscalchi: fixed

asiniscalchi:
I added 2 unittests to expose a good pattern to test the code
- every lib/module with its own test directory
- cmake code is trivial
- add unit tests is trivial

programs/launcher/main.cpp
asiniscalchi: fixed

tests/eosd_run_test.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+DIR="$(cd "$(dirname "$0")" && pwd)/.."
```
jgiszczak: While these changes are very good, this whole file is obsolete and slated for removal.  It's being replaced with a suite of python scripts that are much more modular and easy to manage.  You could easily revert this file and submit a new commit, but it's not necessary.  Just be aware that the changes in this file will not survive for long.

jgiszczak:
We have a second branch that is slated to become `master` in the near future: `eos-noon`.  It would be helpful if you could also submit similar changes to that branch.

programs/launcher/main.cpp
```diff
@@ -1110,24 +1110,27 @@ launcher_def::launch (eosd_def &instance, string &gts) {
   node_rt_info info;
   info.remote = !host->is_local();
 
-  string eosdcmd = "programs/eosd/eosd ";
+  string eosdcmd(eosd_binary);
+
   if (skip_transaction_signatures) {
-    eosdcmd += "--skip-transaction-signatures ";
+    eosdcmd += " --skip-transaction-signatures";
```
jgiszczak: There are other command line options that also need this change.  Per the Travis build, at least `--plugin` needs to be similarly updated.  There may be others.

PR #1142, opened 2018-01-19
Error and exit on missing explicit config file. Resolves #1109.

### LINKED ISSUE #1109: DAWN-500 ⁃ eosd and custom configs
you can pass anything to eosd as config file, which leads to a situation where it is not clear if a custom config.ini was loaded:

```
$ eosd --config foo
667010ms            chain_plugin.cpp:126          plugin_initialize    ] initializing chain plugin
667010ms            wasm_interface.cpp:757        get                  ] Runtime::init
667011ms            producer_plugin.cpp:170       plugin_initialize    ] Public Key: EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV
667011ms            wallet_plugin.cpp:34          plugin_initialize    ] initializing wallet plugin
667011ms            http_plugin.cpp:138           plugin_initialize    ] host: 127.0.0.1 port: 8888
667012ms            http_plugin.cpp:141           plugin_initialize    ] configured http to listen on 127.0.0.1:8888
667012ms            net_plugin.cpp:2379           plugin_initialize    ] Initialize net plugin
667012ms            net_plugin.cpp:2399           plugin_initialize    ] Setting net_plugin logging level to info
667013ms            net_plugin.cpp:2424           plugin_initialize    ] host: 0.0.0.0 port: 9876
667013ms            net_plugin.cpp:2503           plugin_initialize    ] my node_id is 4ccad4efaac16f50dd96fda1f66f065a5087bb7bf221086e7b09982529c0e005
667013ms            main.cpp:59                   main                 ] eosd version 148f2363
```

expected behaviour: invalid config files / not existing config files produce an error.

jgiszczak:


jgiszczak:
Unfortunately because the option "config" has a default, `options.count("config")` is always > 0.

Yes, if someone passes in an absolute path ending in `config.ini` that doesn't exist, one will be created at that location (assuming write permissions for the destination).  The assumption when the user goes to that much trouble to give a full path is that they know what they're doing, and either the file exists or they know it doesn't and want a default generated.  At least at the moment, we're only trying to protect against casual mis-generation.  As part of becoming more standard Unix citizens, it may be necessary to stop being so helpful entirely, and require an explicit "generate" argument if you want a default config file generated.

If someone passes in "config.ini" explicitly when they actually mean "./config.ini" (which is subject to the same "helpful" repathing), then yes, they will be surprised.  This is probably the main impetus for eliminating the unhelpful helpfulness of implicit config file generation entirely.  I'm willing to take bigger steps along the path if you think that's advisable.  I was trying to be incremental about the changes, in an effort not to yank the rug out from under every single developer, including our own.

PR #1144, opened 2018-01-19
 Check wasm->native pointer validity based on active pages

spoonincode:
Previously the wasm->native invoke wrapper would validate wasm pointers based on their validity in the mapped memory pages. This only means the pointer lied within the 4/8/12GB virtual memory map. So it was trivial to hand off a pointer from wasm->native that would segv when native code accessed it. This is caught (and handled) but preference is for there to be no way to segv inside of native code to prevent any oddities from longjmping.

Now, check that the pointer lies within the "active" memory pages -- pages that are R/Wable

Extra review note: There are a few places that can still pass pointers than can segv native code. I'd like to get those cleaned up in the future.

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -180,7 +180,12 @@ auto convert_native_to_wasm(const wasm_interface &wasm, const fc::time_point_sec
 
 auto convert_native_to_wasm(wasm_interface &wasm, char* ptr) {
    auto mem = getDefaultMemory(intrinsics_accessor::get_context(wasm).code.instance);
-   char* base = &memoryRef<char>(mem, 0);
+   if(!mem)
+      Runtime::causeException(Exception::Cause::accessViolation);
+   char* base = (char*)getMemoryBaseAddress(mem);
+   char* top_of_memory = base + IR::numBytesPerPage*Runtime::getMemoryNumPages(mem);
+   if(ptr < base || ptr >= top_of_memory)
+      Runtime::causeException(Exception::Cause::accessViolation);
    return (int)(ptr - base);
 }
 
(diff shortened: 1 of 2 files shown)


libraries/wasm-jit/Source/Runtime/Memory.cpp
```diff
@@ -134,11 +134,10 @@ namespace Runtime
...
 	U8* getValidatedMemoryOffsetRange(MemoryInstance* memory,Uptr offset,Uptr numBytes)
 	{
-		if(!memory)
-			causeException(Exception::Cause::accessViolation);
 		// Validate that the range [offset..offset+numBytes) is contained by the memory's reserved pages.
 		U8* address = memory->baseAddress + offset;
-		if(	address < memory->reservedBaseAddress
+		if(	!memory
+		||	address < memory->reservedBaseAddress
 		||	address + numBytes < address
```
heifner: Line 138 uses `memory` before it is checked for null on line 139.

spoonincode: Yep, I know, I fixed this previously because our code called through this path. But now that we no longer go through this path, I am changing it back to upstream to make future merges easier & reduce the changeset between us and upstream.

https://github.com/AndrewScheidecker/WAVM/blob/master/Source/Runtime/Memory.cpp#L135

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -380,7 +385,9 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, size_t, Inputs...>,
    static Ret translate_one(wasm_interface &wasm, Inputs... rest, Translated... translated, I32 ptr, I32 size) {
       auto mem = getDefaultMemory(intrinsics_accessor::get_context(wasm).code.instance);
       size_t length = size_t(size);
-      T *base = memoryArrayPtr<T>(mem, ptr, length);
```
brianjohnson5972: How come we don't keep the use of memoryArrayPtr and pull these checks into it?

spoonincode: eh my thought was to not the change the library unless needed. In this case there is already an API there to do what is desired, so figured to use what is already there.

PR #1151, opened 2018-01-19
Validate "string pointers" for wasm are entirely in active pages

spoonincode:
There are a couple native methods callable from WASM where the methods expect a NULL terminated char array (a string). Validate that the string lives entirely within active pages before dispatching it. Otherwise native code could segv if given a non-null-terminated string on the end of active pages.

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -439,7 +439,7 @@ struct intrinsic_invoker_impl<Ret, std::tuple<null_terminated_ptr, Inputs...>, s
       char *base          = (char*)(getMemoryBaseAddress(mem) + ptr);
       char *p             = base;
       char *top_of_memory = (char*)(getMemoryBaseAddress(mem) + IR::numBytesPerPage*Runtime::getMemoryNumPages(mem));
-      while(p != top_of_memory)
+      while(p < top_of_memory)
          if(*p++ == '\0')
             return Then(wasm, null_terminated_ptr{base}, rest..., translated...);
       Runtime::causeException(Exception::Cause::accessViolation);
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -98,6 +98,26 @@ struct array_ptr {
    T *value;
 };
 
+/**
+ * class to represent an in-wasm-memory char array that must be null terminated
+ */
+struct null_terminated_ptr {
+   typename std::add_lvalue_reference<char>::type operator*() const {
+      return *value;
+   }
+
+   char *operator->() const noexcept {
+      return value;
+   }
+
+   template<typename U>
+   operator U *() const {
+      return static_cast<U *>(value);
+   }
+
+   char *value;
+};
+
 
 /**
  * template that maps native types to WASM VM types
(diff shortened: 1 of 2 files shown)


libraries/chain/wasm_interface.cpp
```diff
@@ -689,8 +689,8 @@ class console_api : public context_aware_api {
    public:
       using context_aware_api::context_aware_api;
 
-      void prints(const char *str) {
-         context.console_append(str);
+      void prints(null_terminated_ptr str) {
+         context.console_append((const char*)str);
```
spoonincode: halp! How can I get rid of this cast, I'm missing some template magic.......

heifner: `console_append` is a template and therefore requires an exact match on the argument. You could do this: `context.console_append<const char*>(str);` which is much better than the cast. Even better in my opinion would be to add a `.get()` to `null_terminated_ptr`. I am not a fan of `operator U*()`;

wanderingbort: I agree with @heifner about `context.console_append<const char*>(str);` being the best option at the moment.  

If we want to get crazy, we can use SFINAE to remove all the overloads for `console_append` which don't have a defined `operator <<` for the `std::ostringstream`.  This would remove the templated overload that accepts `null_terminated_ptr` and it looks like the compiler is smart enough to call the cast operator at that point.

at least according to this quickie I put together : https://godbolt.org/g/5LdbFB

wanderingbort: I want to be clear, the only reason we would do this insanity is if there is a high chance we will add more things like this. explicit template is pretty much ok 😄

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -397,6 +417,40 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, size_t, Inputs...>,
...
+struct intrinsic_invoker_impl<Ret, std::tuple<null_terminated_ptr, Inputs...>, std::tuple<Translated...>> {
+   using next_step = intrinsic_invoker_impl<Ret, std::tuple<Inputs...>, std::tuple<Translated..., I32>>;
+   using then_type = Ret(*)(wasm_interface &, null_terminated_ptr, Inputs..., Translated...);
+
+   template<then_type Then>
+   static Ret translate_one(wasm_interface &wasm, Inputs... rest, Translated... translated, I32 ptr) {
+      auto mem = getDefaultMemory(intrinsics_accessor::get_context(wasm).code.instance);
+      if(!mem)
+         Runtime::causeException(Exception::Cause::accessViolation);
+      char *base          = (char*)(getMemoryBaseAddress(mem) + ptr);
```
brianjohnson5972: Are we checking for ptr < 0 anywhere?

brianjohnson5972: And similarly for the array_ptr<...> commits that happened earlier, we need a check for ptr < 0?

heifner: Brian go ahead and make these updates with your changes. Since you need these for your work and they are approved, will merge now.

spoonincode: I agree, this needs to be fixed (in a new PR). However, I think I'd rather modify the invoker to treat "pointers" as unsigned.

PR #1156, opened 2018-01-20
Fix resync problem that results in spinning.

PaulCalabrese:
When restarting an eosiod node, the while loop changed can get caught endlessly spinning.

There is also cleanup some launcher issues related to the shared memory directory.

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -1306,7 +1306,7 @@ void chain_controller::update_last_irreversible_block()
       if( new_producer_schedule ) {
          _db.modify( gpo, [&]( auto& props ){
               boost::range::remove_erase_if(props.pending_active_producers,
-                                            [&](typename decltype(props.pending_active_producers)::value_type const& v) -> bool {
+                                            [new_last_irreversible_block_num](const typename decltype(props.pending_active_producers)::value_type& v) -> bool {
                                                return v.first < new_last_irreversible_block_num;
                                             });
               props.active_producers = *new_producer_schedule;
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -29,6 +29,7 @@
 #include <fc/crypto/digest.hpp>
 
 #include <boost/range/algorithm/copy.hpp>
+#include <boost/range/algorithm_ext/erase.hpp>
 #include <boost/range/algorithm_ext/is_sorted.hpp>
 #include <boost/range/adaptor/transformed.hpp>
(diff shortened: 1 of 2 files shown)


programs/launcher/main.cpp
```diff
@@ -763,7 +763,7 @@ launcher_def::deploy_config_files (tn_node_def &node) {
                << " " << strerror(ec.value()) << endl;
           exit (-1);
         }
-        count = bf::remove_all (dd / "blockchain", ec);
+        count = bf::remove_all (dd / "shared_mem", ec);
         if (ec.value() != 0) {
```
heifner: Line 760 has "blocks" and line 766 has "shared_mem". Both of these values are configurable via `config.ini`. Not sure why this is changed to "shared_mem". You removed the generation of this value in the `conifg.ini`. Seem like these values should be put in constants and used in the generation of `config.ini` and in the removal here. Otherwise there will be no way to know the values.

libraries/chain/chain_controller.cpp
```diff
@@ -1304,12 +1305,10 @@ void chain_controller::update_last_irreversible_block()
...
-             /// TODO: use upper / lower bound to remove range
-              while( gpo.pending_active_producers.size() ) {
-                 if( gpo.pending_active_producers.front().first < new_last_irreversible_block_num ) {
-                   props.pending_active_producers.erase(props.pending_active_producers.begin());
-                 }
-              }
+              boost::range::remove_erase_if(props.pending_active_producers,
+                                            [&](typename decltype(props.pending_active_producers)::value_type const& v) -> bool {
+                                               return v.first < new_last_irreversible_block_num;
+                                            });
```
heifner: Some style changes. Please avoid unqualified capture lists when possible. Also we use style of `const` before type not after. So this would become:

```
            boost::range::remove_erase_if(props.pending_active_producers,
                 [new_last_irreversible_block_num](const typename decltype(props.pending_active_producers)::value_type& v) -> bool {
                       return v.first < new_last_irreversible_block_num;
                 });
```

programs/launcher/main.cpp
PaulCalabrese: The shared memory directory is no longer configurable on the eos-noon branch (and was renamed from a default of "blockchain" to a constant of "shared_mem"). I was just trying to clean this up.

PaulCalabrese: The issue with "blocks" is a real one (the user can configure this, but the launcher doesn't use the value when using -f).  I will either write a separate JIRA issue for this or roll it in with the next set of launcher changes I am doing.

PR #1161, opened 2018-01-20
[eos-noon] extract eosc core into libeosioclient 

asiniscalchi:
GOALS:
- unit testing
- allow more frontends (EOS Graphical wallet?, python bindings, etc etc)
- eosioc will be command line frontend
 
I am supposing that a graphical Wallet (ex. Ethereum Mist) needs only the eosioc features so, please, stop me if I am wrong!


wanderingbort:
I like where this is going; I think it should go further (but perhaps this is why you had "in progress" in the title).

Can you comment on your design goals past the high level goals you mentioned?  

It seems to me that the final separation would have `eosioc` only perform command line parsing and help-text localization as a very thin shim.  The library would have equivalent methods for basically every current/future `eosioc` action.  Is that what you are going for?

asiniscalchi:
Just try to understand the code and refactoring it in the meanwhile.
I hope it could be useful.
As I am coding some lines for a graphical client in QML/Qt I needed something more OO.

asiniscalchi:
The design that is emerging is the following:

image

asiniscalchi:
moved to 
https://github.com/EOSIO/eos/pull/1195

PR #1168, opened 2018-01-22
repair ability to push transactions via eosioc on noon

spoonincode:
Review request -- should we add another ctor on chain::action? Or is construction of an action like this acceptable?

### DIFF
--- a/programs/eosioc/main.cpp
+++ b/programs/eosioc/main.cpp
@@ -1090,7 +1090,7 @@ int main( int argc, char** argv ) {
    string contract;
    string action;
    string data;
-   auto actionsSubcommand = push->add_subcommand("actions", localized("Push a transaction with a single actions"));
+   auto actionsSubcommand = push->add_subcommand("action", localized("Push a transaction with a single action"));
    actionsSubcommand->fallthrough(false);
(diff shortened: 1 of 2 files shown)


programs/eosioc/main.cpp
```diff
@@ -1112,10 +1112,12 @@ int main( int argc, char** argv ) {
...
-      /* TODO: restore this function
-      transaction_emplace_serialized_actions(trx, contract, action, accountPermissions,
-                                                      result.get_object()["binargs"].as<bytes>());
-      */
+      eosio::chain::action act;
+      act.account = contract;
+      act.name = action;
+      act.authorization = accountPermissions;
+      act.data = result.get_object()["binargs"].as<bytes>();
+      trx.actions.push_back(act);
```
heifner: Had same fix on my branch, but I added an `action` constructor:

```
      action( vector<permission_level> auth, account_name account, action_name name, const bytes& data )
            : account(account), name(name), authorization(move(auth)), data(data) {
      }
```

Also renamed command to `action` from `actions`

spoonincode: I did the actions->action change, but I'll leave the lack of new ctor for now

Issue #1080, opened 2018-01-12
DAWN-473 ⁃ Add in missing intrinsics to support eoslib


blockone-syncclient:
Also ensure intrinsic changes for eos-noon are reflected in eoslib

ATC:

* This will be tested as part of re-enabling api tests.

blockone-syncclient:
➤ Brian Johnson commented:

Code committed and initial review changes made.

PR #1169, opened 2018-01-22
DAWN-473 intrinsics support

brianjohnson5972:
#1080 
Added missing header definitions in eoslib and missing intrinsic definitions in wasm_interface.cpp.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -224,16 +224,13 @@ const contracts::table_id_object& apply_context::find_or_create_table( name scop
    });
 }
 
-size_t apply_context::get_active_producers(account_name* producers, size_t len) {
+vector<account_name> apply_context::get_active_producers() const {
    const auto& gpo = controller.get_global_properties();
-   int index = 0;
+   vector<account_name> accounts;
    for(const auto& producer : gpo.active_producers.producers)
-   {
-      *(producers + index) = producer.producer_name;
-      if (++index >= len)
-         break;
-   }
-   return index;
+      accounts.push_back(producer.producer_name);
+
+   return accounts;
 }
 
 } } /// eosio::chain
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -226,7 +226,13 @@ const contracts::table_id_object& apply_context::find_or_create_table( name scop
 
 void apply_context::get_active_producers(account_name* producers, size_t datalen) {
    const auto& gpo = controller.get_global_properties();
-   memcpy(producers, gpo.active_producers.producers.begin(), std::min(sizeof(account_name)*gpo.active_producers.producers.size(), datalen));
+   int index = 0;
+   for(const auto& producer : gpo.active_producers.producers)
+   {
+      *(producers + index) = producer.producer_name;
+      if (++index >= datalen)
+         break;
+   }
 }
 
 } } /// eosio::chain
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -224,15 +224,16 @@ const contracts::table_id_object& apply_context::find_or_create_table( name scop
    });
 }
 
-void apply_context::get_active_producers(account_name* producers, size_t datalen) {
+size_t apply_context::get_active_producers(account_name* producers, size_t len) {
    const auto& gpo = controller.get_global_properties();
    int index = 0;
    for(const auto& producer : gpo.active_producers.producers)
    {
       *(producers + index) = producer.producer_name;
-      if (++index >= datalen)
+      if (++index >= len)
          break;
    }
+   return index;
 }
 
 } } /// eosio::chain
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -224,4 +224,9 @@ const contracts::table_id_object& apply_context::find_or_create_table( name scop
    });
 }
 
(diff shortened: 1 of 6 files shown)


libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -467,14 +483,35 @@ struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, array_ptr<U>, size_t
...
+template<typename T, typename Ret, typename... Inputs, typename ...Translated>
+struct intrinsic_invoker_impl<Ret, std::tuple<array_ptr<T>, size_t, null_terminated_ptr, Inputs...>, std::tuple<Translated...>> {
+   using next_step = intrinsic_invoker_impl<Ret, std::tuple<Inputs...>, std::tuple<Translated..., I32, I32, I32>>;
+   using then_type = Ret(*)(wasm_interface &, array_ptr<T>, size_t, null_terminated_ptr, Inputs..., Translated...);
+
+   template<then_type Then>
+   static Ret translate_one(wasm_interface &wasm, Inputs... rest, Translated... translated, I32 ptr_t, I32 size, I32 ptr_u) {
+      const auto length = size_t(size);
+      return Then(wasm, array_ptr<T>(wasm, ptr_t, length), length, null_terminated_ptr(wasm, ptr_u), rest..., translated...);
    };
```
heifner: @wanderingbort mentioned and I confirmed this is not needed. You can remove it. The variadic templates take care of it.

libraries/chain/wasm_interface.cpp
```diff
@@ -638,11 +639,45 @@ class context_aware_api {
...
+      }
+};
+
 class system_api : public context_aware_api {
    public:
       using context_aware_api::context_aware_api;
 
-      void assert(bool condition, null_terminated_ptr str) {
+      void assert(bool condition, const char* str) {
          std::string message( str );
```
heifner: Don't see how you can remove this and still have the null check.

libraries/chain/wasm_interface.cpp
```diff
@@ -689,8 +724,8 @@ class console_api : public context_aware_api {
    public:
       using context_aware_api::context_aware_api;
 
-      void prints(null_terminated_ptr str) {
-         context.console_append<const char*>(str);
+      void prints(const char *str) {
+         context.console_append(str);
       }
```
heifner: Same comment as above, how does this check for null?

libraries/chain/apply_context.cpp
```diff
@@ -224,4 +224,9 @@ const contracts::table_id_object& apply_context::find_or_create_table( name scop
    });
 }
 
+void apply_context::get_active_producers(account_name* producers, size_t datalen) {
+   const auto& gpo = controller.get_global_properties();
+   memcpy(producers, gpo.active_producers.producers.begin(), std::min(sizeof(account_name)*gpo.active_producers.producers.size(), datalen));
+}
```
heifner: `gpo.active_producers.producers` is `fc::array<producer_key,config::producer_count> producers;` where `producer_key` is 
```
   struct producer_key {
      account_name      producer_name;
      public_key_type   block_signing_key;
```
So how can that be copied into an array of `account_name` ?

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
brianjohnson5972: So do you mean this whole specialization can go away?

libraries/chain/wasm_interface.cpp
brianjohnson5972: It will, when I change it back. Again, not sure what I did to get git to do this wrong.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
heifner: yes ;)

libraries/chain/wasm_interface.cpp
heifner: Oh good. Couldn't see how that was correct.

libraries/chain/apply_context.cpp
```diff
@@ -224,4 +224,15 @@ const contracts::table_id_object& apply_context::find_or_create_table( name scop
    });
 }
 
+void apply_context::get_active_producers(account_name* producers, size_t datalen) {
```
wanderingbort: Part of the purpose of the separation of `apply_context` and `wasm_interface` is that the `apply_context` should have implementations that look like they belong in our native code and the `wasm_interface` is responsible for the shim between that and the webassembly VM.  

In this case, I think I would prefer this to have the signature `vector<account_name> (void) const` (`void` for clarity).

This would mean that the logic for guarding against short arrays etc would be part of the `wasm_interface` glue that unfortunately has to deal with data buffers instead of proper containers.

spoonincode:
I'm not a fan of adding intrinsics that could be done on the WASM side (this include preexisting ones like memory, assert, the various prints). But the UTF8 one really worries me since it calls deep in to a "host" library. I can imagine a scenario where a couple years from now someone fixes an FC bug with its UTF8 validation and has no idea it will affect contracts. Not NAKing the PR because we need to get tests up and running ASAP but it's something I'm going to continue to harp on until release.

spoonincode:
Pathological, but the integer pointer checks effectively eliminate access tp the top half of memory. This is fixable at the invoker stage changing the I32s to U32s and the prototypes from ints to unsigns. But... we'll never support >2GB, probably.

PR #1185, opened 2018-01-23
eosioc currency commands

wanderingbort:
Added 2x commands to `eosioc`

### `get currency balance <currency contract> <account> [<symbol>]`

This command queries a standardized currency balance from the blockchain and puts it into a prettier output format than raw database access would allow.  The optional `symbol` parameter can be used to pick a single currency since contracts can operate multiple:

```
$ ./eosioc get currency balance my.token alice
[
  "1000.0000 MYT"
]
$ ./eosioc get currency balance my.token alice MYT
"1000.0000 MYT"
```

### `get currency stats <currency contract> [<symbol>]`

This command queries the information about a currency.  Once again the optional `symbol` parameter is used to select a specific currency on a contract which operates multiple

```
$ ./eosioc get currency stats my.token
{
  "MYT": {
    "supply": "1000.0000 MYT"
  }
}
$ ./eosioc get currency stats my.token MYT
{
  "supply": "1000.0000 MYT"
}
```

### DIFF
--- a/programs/eosioc/main.cpp
+++ b/programs/eosioc/main.cpp
@@ -136,6 +136,8 @@ const string get_block_func = chain_func_base + "/get_block";
 const string get_account_func = chain_func_base + "/get_account";
 const string get_table_func = chain_func_base + "/get_table_rows";
 const string get_code_func = chain_func_base + "/get_code";
+const string get_currency_balance_func = chain_func_base + "/get_currency_balance";
+const string get_currency_stats_func = chain_func_base + "/get_currency_stats";
 const string get_required_keys = chain_func_base + "/get_required_keys";
 
 const string account_history_func_base = "/v1/account_history";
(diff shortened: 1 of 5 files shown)


programs/eosioc/main.cpp
```diff
@@ -436,6 +436,25 @@ struct set_action_permission_subcommand {
    }
 };
 
+struct currency_stats {
+   asset supply;
+};
+
+FC_REFLECT(currency_stats, (supply));
+
```
heifner: For everything else in eosioc main the type is defined in the api plugin. Seems like prompting this to a full fledge understood json of the eosiod api would be appropriate. In this case chain_plugin.

programs/eosioc/main.cpp
```diff
@@ -436,6 +436,25 @@ struct set_action_permission_subcommand {
...
+static currency_stats parse_currency_stats( fc::variant const& row ) {
```
heifner: `const fc::variant& row` is style used elsewhere. Also in `parse_currency_balance`.

programs/eosioc/main.cpp
```diff
@@ -436,6 +436,25 @@ struct set_action_permission_subcommand {
...
+   uint64_t values[2];
+   fc::from_hex(row.as_string(), (char *) values, sizeof(uint64_t) * 2);
```
heifner: I dislike raw c-style casts. Would prefer `reinterpret_cast<char*>(values)`

heifner: `sizeof(values)` would be better in this case I think.

heifner:
Update to readme would be nice. But with other updates going in that can be a diff PR.

PR #1193, opened 2018-01-24
inline wasm compiling thread.

wanderingbort:
removed the utility thread which was in charge of compiling as we were seeing locks under load and a sufficiently multithreaded version would likely need a more robust solution


spoonincode:
I no longer see spooky deadlocks in this area of code, I think we should take it for now.

PR #1194, opened 2018-01-25
remove init() contract ABI call from setcode

bytemaster:
We no longer have the init() entry point for smart contracts, instead everything flows through apply().

We also no longer want to execute code within 30 seconds of an update to the code to give the node a chance to compile/cache the contract.

This update does not enforce the 30 second delay, but does remove the instant call to init() via inline notice of setcode.

### DIFF
--- a/contracts/exchange/exchange.cpp
+++ b/contracts/exchange/exchange.cpp
@@ -285,14 +285,6 @@ void apply_exchange_cancel_sell( order_id order ) {
 } // namespace exchange
 
 extern "C" {
-   void init() {
-      /*
-      setAuthority( "currencya", "transfer", "anyone" );
-      setAuthority( "currencyb", "transfer", "anyone" );
-      registerHandler( "apply", "currencya", "transfer" );
-      registerHandler( "apply", "currencyb", "transfer" );
-      */
-   }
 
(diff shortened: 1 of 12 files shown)


brianjohnson5972:
wasm_tests/memory_operators test is failing in travis.

@spoonincode should we change the wasm_tests.cpp memory_operators test to use current_memory and grow_memory in apply (instead of init), so that we are still verifying that they cannot be called?

../tests/wasm_tests/wasm_tests.cpp(510): error: in "wasm_tests/memory_operators": exception fc::unhandled_exception expected but not raised
../tests/wasm_tests/wasm_tests.cpp(513): error: in "wasm_tests/memory_operators": exception fc::unhandled_exception expected but not raised
*** 2 failures are detected in the test module "Master Test Suite"

spoonincode:
Yes but there are a few other tests in wasm_tests that are going to require a little more advanced changes.

brianjohnson5972:
How come they are not failing?

On Thu, Jan 25, 2018 at 10:28 AM, Matt Witherspoon <notifications@github.com

spoonincode:
Looks like it's only the mutable global test. And it passes because it relied on init() trashing a variable. Without init() nothing trashes that variable.

PR #1199, opened 2018-01-25
tic_tac_toe directory addition

CodyDeeds:
would like for tic_tac_toe to be added to the build


heifner:
The clean up of contracts is in work and tic_tac_toe will be added back in once it has been verified.

PR #1200, opened 2018-01-25
Dawn 249 launcher down node option

PaulCalabrese:
Adds --down option to the launcher that uses the tn_down.sh script to take down the listed nodes.
I also added the ability to pass "all" to the --down and --bounce options.
Cleaned up some launcher issues with the -f option.


programs/launcher/main.cpp
```diff
@@ -1210,48 +1214,91 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
...
-   boost::split(nodes, node_numbers, boost::is_any_of(","));
-   for (string node_number: nodes) {
-      uint16_t node = -1;
-      try {
-         node = boost::lexical_cast<uint16_t,string>(node_number);
+vector<pair<host_def, eosd_def>>
+launcher_def::get_nodes(const string& node_number_list) {
+   vector<pair<host_def, eosd_def>> node_list;
+   string nnl = node_number_list;
+   std::transform(nnl.begin(), nnl.end(), nnl.begin(), ::tolower);
```
heifner: fc has a to_lower in `fc/string.hpp`.

programs/launcher/main.cpp
```diff
@@ -1210,48 +1214,91 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
...
+            exit(-1);
+         }
+         string dex = node < 10 ? "0":"";
+         dex += boost::lexical_cast<string,uint16_t>(node);
+         string node_name = network.name + dex;
+         for (auto host: bindings) {
+            for (auto node: host.instances) {
+               if (node_name == node.name) {
+                  node_list.push_back(make_pair(host, node));
+                  goto continue_next_node;
```
heifner: Uhm, really? Lets go with a `break` and a couple of `node_list.empty()` checks.

programs/launcher/main.cpp
```diff
@@ -1210,48 +1214,91 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
...
-      try {
-         node = boost::lexical_cast<uint16_t,string>(node_number);
+vector<pair<host_def, eosd_def>>
+launcher_def::get_nodes(const string& node_number_list) {
+   vector<pair<host_def, eosd_def>> node_list;
+   string nnl = node_number_list;
+   std::transform(nnl.begin(), nnl.end(), nnl.begin(), ::tolower);
+   if (nnl == "all") {
+      for (auto host: bindings) {
+         for (auto node: host.instances) {
```
heifner: Performance doesn't matter here, but `const auto& x` is a better habit to get into.

programs/launcher/main.cpp
```diff
@@ -1210,48 +1214,91 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
...
+   for (auto node_pair: node_list) {
+      const host_def& host = node_pair.first;
+      const eosd_def& node = node_pair.second;
+      string node_num = node.name.substr( node.name.length() - 2 );
+      string cmd = "cd " + host.eos_root_dir + "; "
+                 + "export EOSIO_HOME=" + host.eos_root_dir + "; "
+                 + "export EOSIO_TN_NODE=" + node_num + "; "
+                 + "export EOSIO_TN_RESTART_DATA_DIR=" + node.data_dir + "; "
+                 + "./scripts/tn_down.sh";
+      cout << "Bouncing " << node.name << endl;
```
heifner: This is not "Bouncing".

PR #1223, opened 2018-01-26
DAWN-529 Reject invalid handshake messages

PaulCalabrese:
Add some simple message validation for handshake messages. If the incoming handshake validation does not pass these tests, reject it and close the connection.

Also get rid of the send_buffer and unneeded extra copy for sent p2p messages.

PR to dawn-2.x branch

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1477,11 +1477,11 @@ namespace eosio {
               ("i", msg.last_irreversible_block_num)("h", msg.head_num));
          valid = false;
       }
-      if (msg.p2p_address == "") {
+      if (msg.p2p_address.empty()) {
          wlog("Handshake message validation: p2p_address is null string");
          valid = false;
       }
-      if (msg.os == "") {
+      if (msg.os.empty()) {
          wlog("Handshake message validation: os field is null string");
          valid = false;
       }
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1465,8 +1461,38 @@ namespace eosio {
...
+       // Do some basic validation of an incoming handshake_message, so things
+       // that really aren't handshake messages can be quickly discarded without
+       // affecting state.
+       bool valid = true;
+       if (msg.last_irreversible_block_num > msg.head_num) {
+          wlog("Handshake message validation: last irreversible block (${i}) is greater than head block (${h})",
+               ("i", msg.last_irreversible_block_num)("h", msg.head_num));
+          valid = false;
+       }
+       if (msg.p2p_address == "") {
```
heifner: `msg.p2p_address.empty()`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1465,8 +1461,38 @@ namespace eosio {
...
+          wlog("Handshake message validation: p2p_address is null string");
+          valid = false;
+       }
+       if (msg.os == "") {
```
heifner: `msg.os.empty()`

PR #1225, opened 2018-01-26
Test for dawn515 stat537

cj-oci:
Stablize Phils test for DAWN 515. Add test to travis build.


cj-oci:
The test in its current state is hard to maintain cause the input data is a base64 string part of the script. After discussion with Kevin, have decided to re-write the test.

PR #1228, opened 2018-01-27
[travis] publish Linux/OSX packages on tag

asiniscalchi:
It will automatically upload the packages creted by travis when a tag is set in master branch.

you can find the result here: 
https://github.com/asiniscalchi/eos/releases

! You have to set the GITHUB OAUTH TOKEN in the .travis file. !

on linux:

$ sudo apt-get install ruby ruby-all-dev
$ sudo gem install travis -v 1.8.8 --no-rdoc --no-r
$ travis setup releases --force

last command will overwrite the .travis file. copy the GITHUB OAUTH TOKEN, revert .travis and past it. Commit.


kesar:
This PR will make everything way easier for newcomers :) 

I just tested it installing the .deb generated in an empty Ubuntu 17.10 and worked pretty smooth ( eoscpp, eosioc && eosio-walletd  ). With eosiod I think there is a missing dependency with mongo:

eosiod: error while loading shared libraries: libmongocxx.so._noabi: cannot open shared object file: No such file or directory

asiniscalchi:
Removed the commit to build mongo as a static lib.
! Remember to set the GITHUB OAUTH TOKEN !

jgiszczak:
Since I needed to add the oauth token, and since this is coming from an external repository, the path of least resistance was to create a new pull request with a branch of eos itself with the oauth token in it.  I gave you credit in both the commit message and the pull request description.  Thanks for the help.

fei-GitHub:
How do you solve this problem?   eosiod: error while loading shared libraries: libmongocxx.so._noabi: cannot open shared object file: No such file or directory @kesar

kesar:
@fei-GitHub there is actually a PR solving this: https://github.com/EOSIO/eos/pull/1890

PR #1246, opened 2018-01-30
Change txn_test_gen_plugin to use wasm currency contract

spoonincode:
Previously txn_test_gen_plugin used the native EOS token. Change it to use the wasm currency contract. This allows banging on the more frequent wasm execution path.


plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
```diff
@@ -10,16 +10,40 @@
...
+   WAST::parseModule(wast.c_str(),wast.size(),module,parseErrors);
+   if(parseErrors.size())
+      FC_THROW_EXCEPTION(fc::parse_error_exception, "wast parsing failure");
+
+   // Serialize the WebAssembly module.
+   Serialization::ArrayOutputStream stream;
+   WASM::serialize(stream,module);
+   return stream.getBytes();
+}
+
```
heifner: I think you can just use `chain::wast_to_wasm()` in `chain/wast_to_wasm.?pp`

spoonincode: Good call.

PR #1218, opened 2018-01-26
Update Dockerfile

JamesDrummond:
Docker build fail due to bin directory not being located in opt folder. Revised to copy from build directory instead.

`genesis-json` folder needed to be updated in config.ini .

PR #1251, opened 2018-01-31
Dawn 548 - Docker fails to build because of install directory doesn't match

andriantolie:
Since cmakelist gets refactored, CMAKE_INSTALL_PREFIX passed will always be replaced with CMAKE_BINARY_DIR/install

Also update the readme


wanderingbort:
@andriantolie if this is still applicable can we resolve merge conflicts?  We had a lot of build changes come through recently.

If its no longer needed please close it out.

andriantolie:
PR https://github.com/EOSIO/eos/pull/1218 has solved the same problem so this PR is obsolete now. I'm going to close it.

PR #1258, opened 2018-01-31
compressed transactions

### LINKED ISSUE #1255: DAWN-550 ⁃ add compression on transactions
In order to handle large transactions like `setcode` and future user driven transactions we are going to add the concept of compression to transactions

wanderingbort:
closes #1255

This enables compressed transaction support.

For now, the only compression schemes available are "none" and "zlib" aka deflate. 

ABI serializer is updated to decompress `packed_transaction` following the precedence set up with `action` which is that `data` will display uncompressed and pretty printed, and `hex_data` will contain the real `data` member's bytes. 

Most of the APIs now take `packed_transaction` where they used to take `signed_transaction` . For convenience, the `tester` has an API endpoint that still takes a `signed_transaction` and will pack it for you.

`eosioc` will automatically compress `setcode` and `setabi` actions.  All other actions are considered uncompressed.   `eosioc push transaction` and `eosioc push transactions` now expect JSONified `packed_transaction` as input.   For convenience, when given a JSON/variant `packed_transaction` it will respect `data` as a string of hex bytes OR a fully exploded pretty `transaction`.  This means that even a non ABI-aware JSONifiy of a `packed_transaction` is still loadable.

Added a few unit tests to make sure that inflate/deflate from known hex data (encoded via command line) works as well as our limiter which protects us from deflate bombs.

NOTE: as this potentially makes some operations like `id()` more expensive on a `packed_transaction` some of the core code was changed to cache values in proper places.  There are still opportunities for this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,8 @@ FIND_PACKAGE(Boost 1.64 REQUIRED COMPONENTS
     chrono
     unit_test_framework
     context
-    locale)
+    locale
+    iostreams)
 
 if( WIN32 )
 
--- a/libraries/chain/CMakeLists.txt
+++ b/libraries/chain/CMakeLists.txt
@@ -30,7 +30,7 @@ add_library( eosio_chain
 
 
              ${HEADERS}
-        )
+        transaction_metadata.cpp)
 
 target_link_libraries( eosio_chain eos_utilities fc chainbase Logging IR WAST WASM Runtime )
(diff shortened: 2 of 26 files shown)


wanderingbort:
This will meta-conflict with #1217 in that we will duplicate storing a packed representation of the transaction on the `transaction_metadata`

I will mark this as pending until that is merged to master and I can update this PR with appropriate resolutions.

PR #1259, opened 2018-01-31
STAT-538 Checktime Port from dawn-2.x

brianjohnson5972:
#1260 
First porting over checktime from dawn-2.x before adding code to inject instruction count.


libraries/chain/wasm_interface.cpp
```diff
@@ -219,7 +219,7 @@ namespace eosio { namespace chain {
                try {
                   Serialization::MemoryInputStream stream((const U8 *) wasm_binary, wasm_binary_size);
                   #warning TODO: restore checktime injection?
```
spoonincode: should this warning still be here?

brianjohnson5972: No, I'll remove it with the instruction counting changes.

PR #1262, opened 2018-01-31
Rename installed scripts and binaries with eosio- prefix

jgiszczak:


jgiszczak:
Python scripts updated.

PR #1269, opened 2018-02-01
Dawn2 genesis file for testnet3 stat-531

cj-oci:
Adding staked balance of 10000 to init accounts.


cj-oci:
Closing request as the genesis contents are no longer relevant.

PR #1275, opened 2018-02-01
packed tx fixes

wanderingbort:
There were a few subtle bugs with #1258 

moving a `transaction_metadata` constructed from a `packed_transaction` was creating a stale `const transaction&` value.  

`_pending_transaction_metas` was not getting properly cleared after block production.


wanderingbort:
as a bonus, we now have a test case in `chain_test` that actually exercises `push_block`

PR #1279, opened 2018-02-02
fix build, test failing

asiniscalchi:
test are not passing

### DIFF
--- a/libraries/chain/include/eosio/chain/transaction.hpp
+++ b/libraries/chain/include/eosio/chain/transaction.hpp
@@ -166,18 +166,18 @@ namespace eosio { namespace chain {
 
       packed_transaction() = default;
 
-      packed_transaction(const transaction& t, compression_type _compression = none)
+      explicit packed_transaction(const transaction& t, compression_type _compression = none)
       {
          set_transaction(t, _compression);
       }
 
-      packed_transaction(const signed_transaction& t, compression_type _compression = none)
+      explicit packed_transaction(const signed_transaction& t, compression_type _compression = none)
       :signatures(t.signatures)
       {
          set_transaction(t, _compression);
       }
 
-      packed_transaction(signed_transaction&& t, compression_type _compression = none)
+      explicit packed_transaction(signed_transaction&& t, compression_type _compression = none)
       :signatures(std::move(t.signatures))
       {
          set_transaction(t, _compression);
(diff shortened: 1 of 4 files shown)


kesar:
👍

asiniscalchi:
my build was using the header from a previous installation.
Files are missing... closing the pull

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -166,18 +166,18 @@ namespace eosio { namespace chain {
 
       packed_transaction() = default;
 
-      explicit packed_transaction(const transaction& t, compression_type _compression = none)
+      packed_transaction(const transaction& t, compression_type _compression = none)
```
wanderingbort: we shouldn't remove the explicit specifier.  It was added explicitly 😏 .  We should fix the invocations where this is the correct conversion by making those conversions explicit and, where this is not the correct conversion we should remove the need for the conversion.

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -166,18 +166,18 @@ namespace eosio { namespace chain {
...
       {
          set_transaction(t, _compression);
       }
 
-      explicit packed_transaction(const signed_transaction& t, compression_type _compression = none)
+      packed_transaction(const signed_transaction& t, compression_type _compression = none)
```
wanderingbort: same requirement for explicit

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -166,18 +166,18 @@ namespace eosio { namespace chain {
...
       :signatures(t.signatures)
       {
          set_transaction(t, _compression);
       }
 
-      explicit packed_transaction(signed_transaction&& t, compression_type _compression = none)
+      packed_transaction(signed_transaction&& t, compression_type _compression = none)
```
wanderingbort: same requirement for explicit.

libraries/chain/include/eosio/chain/transaction.hpp
wanderingbort: Originally, I had no `explicit` modifier expecting that the second parameter would keep this from being used as an implicit cast.  However, at least on llvm4 it would use the defaulted second parameter and deploy the implicit cast.  Learned something new that day.

asiniscalchi:
@wanderingbort done

PR #1280, opened 2018-02-02
Renaming eoslib path to eosiolib.

str:


heifner:
Fixed by @zorba80 recent PR.

PR #1286, opened 2018-02-02
fix for push_block with multiple transactions

wanderingbort:
fixed bugs that were only apparent when pushing a block with enough transactions to resize the `vector<transaction_metadata>` that we create out of the `input_transactions`


spoonincode:
Took 4df1b43, added 3a48574 and this 2e582fb on it. Works for txn_test_gen now.

PR #1287, opened 2018-02-02
Add more constraints on WASM execution environment

spoonincode:
Adding a few more constraints on the WASM execution environment. Some of these we’re well aware of but others are new constraints to aid performance and scalability.

* Limit initial linear memory size to 16 pages (1MB) maximum
* Limit linear memory initializers to the first page (first 64KB)
* Limit number of table elements to 1024
* Limit mutable globals to a maximum of 1KB
* Limit load/store offsets to a maximum of 1MB (matching maximum runtime size)

### DIFF
--- a/libraries/chain/CMakeLists.txt
+++ b/libraries/chain/CMakeLists.txt
@@ -9,6 +9,7 @@ add_library( eosio_chain
              block.cpp
              wast_to_wasm.cpp
              wasm_interface.cpp
+             wasm_eosio_constraints.cpp
              apply_context.cpp
              rate_limiting.cpp
 
--- a/libraries/chain/include/eosio/chain/wasm_eosio_constraints.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_eosio_constraints.hpp
@@ -1,0 +1,12 @@
+#pragma once
+
+namespace IR {
+    struct Module;
+};
+
+namespace eosio { namespace chain {
+
+//Throws if something in the module violates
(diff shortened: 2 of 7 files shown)


spoonincode:
Today, nothing would really stop us from giving 1GB or more as long as we account for time spent memset()ing it.

Going forward though, I think we will have trouble with scalability given how the current code handles growing memory: issuing multiple syscalls that implicitly cause CPU cache flushing, TLB flushing, and page faults. This doesn’t matter for initial release, but I don’t want to dig ourselves in to a scalability hole and needing to upheave memory behavior next release.

I have an untested theory on how to grow/shrink memory super fast with none of the above pitfalls. But I need to keep the total number of pages reasonably small for this to work. 256 64KB pages for 16MB is too much.

If moving beyond 1MB, what I’m considering is using graduated page sizes. For simple illustration at a 1MB max:
64KB (first page)
8KB
8KB
16KB
32KB
128KB
256KB
512KB
So a contract using a total of 65KB only has to memset() 72KB. But a heavy contract using 560KB will get burdened with the cost of memset()ing 1MB. But it's a heavy contract.. so Deal With It. This approach only used 8 "pages" which plays much better with the untested super fast memory grower idea.

If we can envision long running contracts (like 25ms) then extrapolating the above approach we could have some tail pages with huge sizes like 16MB or 32MB or more. It might burn 1ms to memset() a 64MB page but for a long running contract that's a very small time slice.

PR #1299, opened 2018-02-05
add more crypto apis to wasm interface

bytemaster:


libraries/chain/wasm_interface.cpp
```diff
@@ -534,14 +535,57 @@ class crypto_api : public context_aware_api {
...
+
+      int recover_key( fc::sha256& digest, 
+                        array_ptr<char> sig, size_t siglen,
+                        array_ptr<char> pub, size_t publen ) {
+         fc::crypto::signature s;
+         datastream<const char*> ds( sig, siglen );
+         datastream<char*> pubds( pub, publen );
+
+         fc::raw::unpack(ds, s);
+         fc::raw::pack( pubds, fc::crypto::public_key( s, digest, false ) );
```
spoonincode: Since this can throw, it'll kill a contract if key recovery fails. Is that the intention? Is there no use case where a contract may want to know if key recovery fails and act on it?

Issue #1222, opened 2018-01-26
DAWN-536 ⁃ Large Block notifications break synchronization


blockone-syncclient:
Matt discovered this situation. When a block producer has a block exceeding the "just send it" size, a notice message is sent and interested receivers have to request the block. but what can happen now is that while a large block is requested, a subsequent block can be produced which is pushed straight away. This new block ends up being received out of order and wackiness ensues.

In the short term we can get around this by always sending blocks regardless of size. 

A proper solution might be to cache "out of order" blocks while awaiting a requested block, or discard those blocks and fall back to a resync mode after receiving the requested block. 

A more exotic solution would be to build up a network node map that enables each node to be aware of which nodes are directly connected peers and thus be able to infer whether a message received from a given peer needs to be forwarded to any other peers.

blockone-syncclient:
➤ Phil Mesnier commented:

This was a significant topic of discussion at Blacksburg yesterday. Dan set ne right, explaining the details I had missed previously. The key factors being 1) don't send a notice for a message you don't have and 2) do send a notice when you receive a big message.

Adding in a new "Big Message Manager" to coordinate the notification and requesting of big messages, whether those are blocks or transactions.

blockone-syncclient:
➤ Phil Mesnier commented:

Given the turbulence hitting DAWN 3.0 with the advent of EOSIO-BIOS, I'm going to port my changes for this issue to Dawn 2.x. Then I think we will need to do a rollout to testnet 1. After I've verified it on testnet 2 of course.

pmesnier:
before I make a PR I need to address the deep recursion problem discovered by the community testnet users.

brianjohnson5972:
Are they actually related, or should they be separate PRs?

blockone-syncclient:
➤ Phil Mesnier commented:

Fix is working on testnet 2 but I want to add a rewrite of the recursive function used to gather "reversible" blocks - if there are too many, they will blow the stack.

pmesnier:
it's all part of synchronization and the program's failure to thrive, so yeah, they're related.

PR #1305, opened 2018-02-05
#1222, jira-536, import changes from the dawn 2.x based branch.

pmesnier:
This PR addresses the Big Message distribution issue, adding a new manager class. This new manager class brings together functionality related to the broadcasting of blocks or transactions that may exceed the maximum size of a message that may exceed the maximum size for messages to be broadcast as is. Too large messages are advertised with a notice message.

This PR also further simplifies the synchronization code, concentrating related functionality out of the connection and net_plugin_impl classes and into the sync manager.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -50,7 +50,7 @@ namespace eosio {
 
    class connection;
    class sync_manager;
-
+   class big_msg_manager;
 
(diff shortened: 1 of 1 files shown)


spoonincode:
This doesn't seem to cause any issues -- To get something stable for net testing on master I started from 4df1b43 then applied 3a48574, 2e582fb, and this c15ab66.

But, this doesn't seem to resolve the trivial large transaction propagation problems I've experienced in the past either. That can be solved later though.

PR #1308, opened 2018-02-05
Master bta 1307

zorba80:
* Added bill-to-account (bta) to C and C++ DB API's.
* Reversed scope-code order. Now the general parameter and template parameter order is code, scope, table_name, bta.
* Added bta_index to different key_value indexes.


contracts/eosiolib/db.h
```diff
@@ -229,7 +229,7 @@ int32_t upper_bound_i64( account_name scope, account_name code, table_name table
  *
  *  @return 1 if a record was removed, and 0 if no record with key was found
  */
-int32_t remove_i64( account_name scope, table_name table, void* data );
+int32_t remove_i64( account_name scope, table_name table, account_name bta, void* data );
```
wanderingbort: Do we need bta for remove?  are there any cases where this value would be something different than the one present on the existing record?

libraries/chain/include/eosio/chain/contracts/contract_table_objects.hpp
```diff
@@ -46,7 +46,18 @@ namespace eosio { namespace chain { namespace contracts {
...
    struct by_scope_tertiary;
-
+   struct by_bta;
+
+   template <typename ObjectType>
+   using bta_index = ordered_unique<tag<by_bta>,
+      composite_key<ObjectType,
+                    member<ObjectType, typename ObjectType::id_type, &ObjectType::id>,
+                    member<ObjectType, table_id, &ObjectType::t_id>,
+                    member<ObjectType, account_name, &ObjectType::bta>
```
wanderingbort: can we demote the `ObjectType::id_type` member of this index to the last member of the composite?

That would allow for queries in the form of "all rows in a given table that I am paying for".   I think the format as present means you have to have id values in the query

contracts/eosiolib/db.h
zorba80: Deleted bta from remove.

libraries/chain/include/eosio/chain/contracts/contract_table_objects.hpp
zorba80: Done.

PR #1310, opened 2018-02-05
Identity contract (part 2)

moskvanaft:
More tests, bugfixes, etc


contracts/eosiolib/singleton.hpp
```diff
@@ -27,12 +27,18 @@ namespace  eosio {
             char temp[1024+8];
             *reinterpret_cast<uint64_t *>(temp) = SingletonName;
             auto read = load_i64( scope, Code, SingletonName, temp, sizeof(temp) );
-            assert( read > 0, "singleton does not exist" );
-            datastream<const char*> ds(temp + sizeof(SingletonName), read);
+            assert( read < 0, "singleton does not exist" );
```
heifner: The assert was correct before.

moskvanaft: fixed. Thanks!

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -14,6 +15,8 @@ namespace eosio { namespace testing {
     */
    class tester {
       public:
+         typedef string ActionResult;
```
heifner: Do not use camel case for types.

moskvanaft: fixed

PR #1311, opened 2018-02-05
add new iterator based db api

bytemaster:


libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -185,7 +187,196 @@ class apply_context {
...
+
+         mutable_db.modify( obj, [&]( auto& o ) {
+           o.value.resize( buffer_size );
+           memcpy( o.value.data(), buffer, buffer_size );
+           o.payer = payer;
+         });
+      }
+
+      void db_remove_i64( int iterator ) {
+         const key_value_object& obj = keyval_cache.get( iterator );
```
brianjohnson5972: Not sure on whether we are worrying about locking yet, but we are using a reference to an object that could be removed by another thread calling db_remove_i64, right?

PR #1314, opened 2018-02-06
main.cpp

chenlian2015:
eosc  pamameter limit  not work

### DIFF
--- a/programs/eosioc/main.cpp
+++ b/programs/eosioc/main.cpp
@@ -560,6 +560,7 @@ int main( int argc, char** argv ) {
    string table;
    string lower;
    string upper;
+   string key;
    bool binary = false;
    uint32_t limit = 10;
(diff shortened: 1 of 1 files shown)


chenlian2015:
eosc  pamameter limit  not work

programs/eosioc/main.cpp
```diff
@@ -560,6 +560,7 @@ int main( int argc, char** argv ) {
    string table;
    string lower;
    string upper;
+   string key;
```
wanderingbort: This appears to go completely unused.

chenlian2015: already used!! 
eg.
eosc      get table oct oct account
eosc      get table songbo oct account
eosc      get table zhaobo oct account
when account table rows large enough, limit become import !!!

heifner:
This was fixed via DAWN-245.

PR #1315, opened 2018-02-06
more chain test fixes

wanderingbort:
fixes for the delegated transfer test.  

temp fixes for the irreversible block tests as we only have 1 producer at the moment and that is a pathological case for irreversibility.


libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -33,6 +33,7 @@ namespace eosio { namespace chain {
 
       shared_producer_schedule_type& operator=( const producer_schedule_type& a ) {
          version = a.version;
+         producers.clear();
          producers.reserve( a.producers.size() );
          for( const auto& p : a.producers )
             producers.push_back(p);
```
heifner: Github does not appear to let me add a comment outside of changed code...

Line 59, I think should have `if( a.producers.size() != b.producers.size()) return false;`

wanderingbort: agreed

PR #1316, opened 2018-02-06
fixes for wasm memory tests

wanderingbort:
the removal of `init` from `setcode` moved compilation AND validation of wasm to the first action.  We should do validation during `setcode` but doing this properly requires additional work.  This is a temporary fix

Additionally, the injector code was failing when hit with hand crafted wasm that had no types/import section(s)

### DIFF
--- a/libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
+++ b/libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
@@ -186,8 +186,10 @@ public:
    {}
 
 
-   void addCall(const Module& module, Serialization::OutputStream& inByteStream)
+   void addCall(Module& module, Serialization::OutputStream& inByteStream)
    {
+		// make sure the import is added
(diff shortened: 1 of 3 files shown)


libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
```diff
@@ -186,8 +186,10 @@ class ChecktimeInjection
    {}
 
 
-   void addCall(const Module& module, Serialization::OutputStream& inByteStream)
+   void addCall(Module& module, Serialization::OutputStream& inByteStream)
    {
+		// make sure the import is added
+		addImport(module);
       OpcodeAndImm<CallImm>* encodedOperator = (OpcodeAndImm<CallImm>*)inByteStream.advance(sizeof(OpcodeAndImm<CallImm>));
```
heifner: You might have some tab vs. spaces in this file if you want to fix.

PR #1320, opened 2018-02-07
feat(db): Contract interfaces did not include scope.

michaelbpaulson:
The contract interfaces for the database did not have scope plumbed through consistently.


michaelbpaulson:
@heifner do you want me to open up a pr against master?

heifner:
Yes, although that area is in flux. So make sure you are current and you checkout other PRs.

PR #1328, opened 2018-02-07
Reduce wasm table vmem usage

spoonincode:
Previously, a 68GB virtual memory map was created per wasm (contract) instance. This eats up the virtual memory space exceptionally quick.

Now, only allocate 12KB per wasm/contract instance. This is enough for 1024 table elements which is hopefully plenty. The side effect is that every indirect function call now has an additional runtime check on its value (checking if it is less than 1024). This seems like an acceptable tradeoff.

Also I have removed the 4GB alignment for the table and replaced it with 4KB. I wasn’t finding any benefit to that 4GB alignment. The generated x86_64 code was simply in the form of `*8(%base, %index)` so I probably don’t even need 4KB.

Review note: The unit tests fail due to a fault in checktime injection. Removing checktime injection causes them to pass.

### DIFF
--- a/libraries/wasm-jit/Source/Runtime/CMakeLists.txt
+++ b/libraries/wasm-jit/Source/Runtime/CMakeLists.txt
@@ -31,6 +31,8 @@ add_definitions(${LLVM_DEFINITIONS})
 
 add_definitions(-DRUNTIME_API=DLL_EXPORT)
 
+target_include_directories( Runtime PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../../../chain/include )
+
 # Link against the LLVM libraries
 llvm_map_components_to_libnames(LLVM_LIBS support core passes mcjit native DebugInfoDWARF)
 target_link_libraries(Runtime Platform Logging IR ${LLVM_LIBS})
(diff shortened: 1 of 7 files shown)


libraries/wasm-jit/Source/Runtime/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ add_definitions(${LLVM_DEFINITIONS})
 
 add_definitions(-DRUNTIME_API=DLL_EXPORT)
 
+target_include_directories( Runtime PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/../../../chain/include )
```
spoonincode: Gross. Any thoughts on improving this? chain library uses wasm-jit so I can't create some sort of cyclic dependency where wasm-jit uses chain, I don't think

wanderingbort: Can we reverse the direction?  It looks like we only include from this directory to use `eosio::chain::wasm_constraints::maximum_table_elements`

If we define that as a constant in say `Runtime.h` in wasm-jit, we can then include that in our chain level file and set our value to it.

Issue #1260, opened 2018-01-31
DAWN-538 ⁃ WASM Instruction Counting


blockone-syncclient:
1. move instruction insertion to JIT rather than WAST AST.
1. rather than call to "checktime" with 0 args, pass argument with number of instructions in current block.

Also need to:  Port this work over to the interpreter, see Kevin and Bart for clarification

ATC:
There is no outward evidence of this accumulated instruction count (to be implemented in a separate ticket), but a regression test should be performed.  Need to make a contract that takes a long time to perform so that a checktime exception is received for exceeding the execution time. Example: perform a while loop checking the content of a key stored in a table and writing data to another table, as long as the content for that key never changes, loop will run till time constraint is reached.

blockone-syncclient:
➤ Brian Johnson commented:

I'm  assuming that no one else is porting checktime from dawn-2.x to master.

blockone-syncclient:
➤ Brian Johnson commented:

[~dan.larimer] Was discussing with [~matt.witherspoon] how to handle not over counting instructions with nested loops/blocks and he suggested that we could do this in LLVMEmitIR.cpp and track the real instruction count and move the checktime call to EmitFunctionContext::call, EmitFunctionContext::block, and EmitFunctionContext::loop.  The difference would be that the instruction count at that point in time would be the current instruction count and not the expected count that we would have been calculating in counting the instructions in the function, block, or loop like we had discussed, but this would be more similar to the checktime time check, which is based off of the time that has currently expired.  Please indicate what you think of this solution.

blockone-syncclient:
➤ Bart Wyatt commented:

[~brian.johnson]has done the work to prove this technique.  Unfortunately, there will be enough superficial changes to that code when it is unified by [~matt.witherspoon] across all possible back-ends that we don't want to merge it into master directly. 

As a result, I'm considering this (and the PR) "done" even though the code will not be present in the master branch until after [~matt.witherspoon]'s task (which will link here later).

PR #1331, opened 2018-02-07
DAWN-538 Instruction Counting Passing Hard Coded Value

brianjohnson5972:
#1260 
Initially changing checktime to just pass a hard coded value, and will then work on counting the instructions and someone else can work on the apply_context implementation.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -238,7 +238,7 @@ void apply_context::checktime_start() {
    _checktime_start = fc::time_point::now();
 }
 
-void apply_context::checktime() const {
+void apply_context::checktime(uint32_t instruction_count) const {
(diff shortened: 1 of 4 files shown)


brianjohnson5972:
Is there anything malicious that a user could do with checktime, as long as we prevent a rollover of the instruction tally?  If we are worried about some one accidentally using it, they would have to add a c declaration for it without providing it, which seems like user would have to be pretty confused and also think they were cleverer than they really were.

spoonincode:
I don't think anything malicious but it exposes an implementation detail for determinism that could change in the future. For example, it may no longer be a function in the future.

brianjohnson5972:
I added DAWN-574 to address the quasi-Contract API inclusion of checktime.

PR #1332, opened 2018-02-07
Fixes Docker build to work with latest code.

ericiles:
@flux627  and I got docker builds working properly with the current codebase. This is using @heifner config.ini and genesis.json files. We have tested eosiod, and it now successfully creates blocks.


heifner:
This is overcome by events today. Hopefully, when master is more stable we can get an update for this.

PR #1333, opened 2018-02-07
submodulize musl and libcxx

wanderingbort:
moves musl and libcxx to their own repositories which mirror the appropriate upstreams and have branches named `eosio` which reflect the copied files we used to have. 

This will allow us to track/merge changes from upstream independently from the eos repository.  Hopefully this will make it easier to merge changes to the upstream as we will have access to the full history of those mirrored repository.


wanderingbort:
This seems to have failed Travis due to the non-determinisitic checktime failure funtimes reintroduced recently.  #1341 will (hopefully) solve this moving forward.

Issue #1338, opened 2018-02-08
intrinsic memset causes "access violation" in stltest


moskvanaft:
stltest works if memset is a libc function (on wasm side), but fails with ABI memset on native side.

PR #1339, opened 2018-02-08
#1338 fix boundary check

moskvanaft:


spoonincode:
This is accurate. @larryk85 and I discussed it awhile back too.

PR #1341, opened 2018-02-08
Deadlines for Checktime, not per-tx-times

wanderingbort:
Changed checktime to respect a deadline instead of an interval. 

This deadline is set when pushing a transaction or pushing a block.  When pushing a transaction it is set to some value which protects the producer/validator from being attacked.  When pushing a block it is set to the deadline for the _whole_ block.  Any transaction which exceeds this time is grounds for not accepting the block.

These are also configurable at `chain_controller` instantiation and plumbed in from `chain_plugin` and the `tester`.  `tester` in particular has no limits(*) which should help our non-deterministic CI failures.

(*)limits are set for one test case... which tests that limits are enforced.


wanderingbort:
this will no doubt conflict with #1331 I will push merge fixes when that goes in.

libraries/chain/chain_controller.cpp
```diff
@@ -586,6 +578,11 @@ static void validate_shard_locks(const vector<shard_lock>& locks, const string&
 
 void chain_controller::__apply_block(const signed_block& next_block)
 { try {
+   optional<fc::time_point> processing_deadline;
+   if (!_currently_replaying_blocks && _limits.max_push_block_us.count() > 0) {
+      processing_deadline = fc::time_point::now() + _limits.max_push_transaction_us;
```
brianjohnson5972: should this be:
+ _limits.max_push_block_us;

wanderingbort: yep, 11a6d1d fixes

wanderingbort:
Right now, we don't have a good avenue for "maximum time in a  block I'm producing".  But we need to have that.  We may have to use this deadline to set a per-tx max of the lesser of the per-tx max we already have and the remainder of some accumulated estimate of block time.  Or we may actually just use the wall clock inside your production window even though that would be a conservative estimate (production runs slower than validation).

heifner:
You put `in progress` on this. Is it ready to merge?

PR #1344, opened 2018-02-08
Eosio abi

heifner:
abi_serializer::to_variant failed because the eosio_system_abi was overwritten when set_abi for the eosio.system was set.  Added chain_initializer::eos_contract_abi() to eosio account abi.

Also fixed nonce action in eosioc.

### DIFF
--- a/tests/wasm_tests/wasm_tests.cpp
+++ b/tests/wasm_tests/wasm_tests.cpp
@@ -13,10 +13,13 @@
 
 #include <noop/noop.wast.hpp>
 #include <noop/noop.abi.hpp>
+#include <eosio.system/eosio.system.wast.hpp>
+#include <eosio.system/eosio.system.abi.hpp>
 
 #include <Runtime/Runtime.h>
 
 #include <fc/variant_object.hpp>
+#include <fc/io/json.hpp>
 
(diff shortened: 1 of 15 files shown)


tests/wasm_tests/wasm_tests.cpp
```diff
@@ -625,6 +628,47 @@ BOOST_FIXTURE_TEST_CASE(noop, tester) try {
 
  } FC_LOG_AND_RETHROW()
 
+// abi_serializer::to_variant failed because eosio_system_abi modified via set_abi.
+// This test also verifies that chain_initializer::eos_contract_abi() does not conflict
+// with eosio_system_abi as they are not allowed to contain duplicates.
+BOOST_FIXTURE_TEST_CASE(eosio_abi, tester) try {
```
wanderingbort: While I don't think this test case does anything bad, `tester::tester(true)` aka the default constructor used by the fixture test case.  Will call `set_code` and `set_abi` with the  `test.system` contract.  We may one day extend that to do more "bootstrapping".

We should make it easy to use a bootstrapped test chain and a fresh empty chain as a fixture.  

at this point I'm thinking out loud because we will put that in a new ticket :)

PR #1353, opened 2018-02-09
P2p queue unity d2x

pmesnier:
get rid of the separate "out_queue" and "txn_queue" from the connection class and have the all messages to be send go straight to the write_queue.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -299,18 +299,7 @@ namespace eosio {
       bool                is_known_by_peer = false; ///< true if we sent or received this trx to this peer or received notice from peer
       bool                is_noticed_to_peer = false; ///< have we sent peer notice we know it (true if we receive from this peer)
       uint32_t            block_num = 0; ///< the block number the transaction was included in
-      time_point          validated_time; ///< infinity for unvalidated
       time_point          requested_time; /// in case we fetch large trx
-#if 0
-      transaction_state(transaction_id_type &tid)
-         :id(tid)
-         ,is_known_by_peer(false)
-         ,is_moticed_to_peer(false)
-         ,block_num(0)
-         ,validated_time()
-         ,requested_time()
-         {}
-#endif
    };
 
    struct update_block_num {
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1496,14 +1451,14 @@ namespace eosio {
       ds.write( reinterpret_cast<char*>(&packsiz), sizeof(packsiz) );
       fc::raw::pack( ds, msg );
 
-      uint16_t bn = static_cast<uint16_t>(txn.ref_block_num);
+      uint32_t bn = static_cast<uint32_t>(txn.ref_block_num);
```
PaulCalabrese: Does this cast do what is intended? Is the 16 bit block number later treated like a full block number?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -684,28 +672,37 @@ namespace eosio {
...
 
    void connection::txn_send_pending(const vector<transaction_id_type> &ids) {
-      for(auto t : my_impl->local_txns){
-         if(t.packed_transaction.size()) {
+      for(auto tx = my_impl->local_txns.begin(); tx != my_impl->local_txns.end(); ++tx ){
+         if(tx->packed_transaction.size()) {
             bool found = false;
             for(auto l : ids) {
-               if( l == t.id) {
+               if( l == tx->id) {
```
PaulCalabrese: The variable "l" looks way too much like a one in the github code font; we should probably use something a little more descriptive.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1599,10 +1556,27 @@ namespace eosio {
...
+            note.known_blocks.ids.push_back( blk_id );
+            note.known_trx.mode = none;
+            my_impl->send_all(note, [blk_id](connection_ptr conn) -> bool {
+                  const auto& bs = conn->blk_state.find(blk_id);
+                  bool unknown = bs == conn->blk_state.end();
+                  if (unknown) {
+                     conn->blk_state.insert(block_state({blk_id,false,true,fc::time_point() }));
+                  }
+                  else {
+
```
PaulCalabrese: Git rid of the else.

PR #1354, opened 2018-02-09
Progress on Multi-Index API

bytemaster:
1. remove eosio::string and eosio::vector in favor of std::string and std::vector
2. updates to database_api for multi index use.


libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -134,54 +134,152 @@ class apply_context {
...
+               primary = itr->primary_key;
+               secondary = itr->secondary_key;
+
                itr_cache.cache_table( *tab );
                return itr_cache.add( *itr );
             }
 
-            int find_primary( uint64_t code, uint64_t scope, uint64_t table, uint64_t primary ) {
-               auto tab = context.find_table( scope, context.receiver, table );
+            int next_secondary( int iterator, uint64_t& primary ) {
```
wanderingbort: Since we are iterating until the end of the table, not the end of a particular secondary key value, shouldn't this emit the value of the secondary key too (to keep convention with the other iterator operations)?

tests/wasm_tests/wasm_tests.cpp
```diff
@@ -462,14 +462,20 @@ BOOST_FIXTURE_TEST_CASE( memory_init_border, tester ) try {
...
 
-   //this will fail to link but that's okay; mainly looking to make sure that the constraint
-   // system doesn't choke when memories and tables exist only as imports
-   BOOST_CHECK_THROW(set_code(N(imports), memory_table_import), fc::exception);
+      //this will fail to link but that's okay; mainly looking to make sure that the constraint
+      // system doesn't choke when memories and tables exist only as imports
+      BOOST_CHECK_THROW(set_code(N(imports), memory_table_import), fc::exception);
+   } catch ( const fc::exception& e ) {
+
+        edump((e.to_detail_string()));
```
wanderingbort: was there a case where `FC_LOG_AND_RETHROW()` was insufficient?

contracts/eosio.system/eosio.system.cpp
```diff
@@ -11,6 +11,6 @@ extern "C" {
 
     /// The apply method implements the dispatch of events to this contract
     void apply( uint64_t code, uint64_t act ) {
-       eosiosystem::contract<N(eosio)>::apply( code, act );
+//       eosiosystem::contract<N(eosio)>::apply( code, act );
```
wanderingbort: `eosio_run_test.py` is expecting this behavior to work, is there a reason why the `eosio.system` contract is currently disabled?

contracts/eosio.system/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ file(GLOB ABI_FILES "*.abi")
 configure_file("${ABI_FILES}" "${CMAKE_CURRENT_BINARY_DIR}" COPYONLY)
 
 add_wast_executable(TARGET eosio.system
-  INCLUDE_FOLDERS "${STANDARD_INCLUDE_FOLDERS}"
-  LIBRARIES eosiolib
+  INCLUDE_FOLDERS ${STANDARD_INCLUDE_FOLDERS}  /usr/local/include
```
wanderingbort: Shouldn't this be the CMake based macro instead of `/usr/local/include`?

contracts/proxy/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ file(GLOB ABI_FILES "*.abi")
 configure_file("${ABI_FILES}" "${CMAKE_CURRENT_BINARY_DIR}" COPYONLY)
 
 add_wast_executable(TARGET proxy
-  INCLUDE_FOLDERS "${STANDARD_INCLUDE_FOLDERS}"
-  LIBRARIES eosiolib
+  INCLUDE_FOLDERS "${STANDARD_INCLUDE_FOLDERS}" ${Boost_INCLUDE_DIR}
```
wanderingbort: This seems to be a result of including the `eosio.system` contract's include so that we can process EOS tokens in the proxy.  In order to keep dependencies on contracts low, maybe we should separate out the definition of the EOS currency from the rest of the system contract so that we don't have to pull in things like Boost HANA at build time just to reference the currency?

PR #1364, opened 2018-02-11
Consensus validation test - STAT372

cj-oci:
Add new test for consensus validation when malicious producers are present. Add test to travis build. Import test script changes from master branch.


heifner:
On mac this is getting: `tests/eosd_run_test.py: /usr/bin/python3: bad interpreter: No such file or directory`. I'm more concerned with this working on master than dawn-2.x.

PR #1371, opened 2018-02-12
Test porting

larryk85:
#DAWN-390,391,392

 Porting api tests, memory tests, and extended memory tests


contracts/eosiolib/crypto.h
```diff
@@ -14,10 +14,76 @@ extern "C" {
...
+/**
+ *  This method is implemented as:
+ *
+ *  checksum calc_hash;
+ *  ripemd160( data, length, &calc_hash );
+ *  eos_assert( calc_hash == hash, "invalid hash" );
+ *
+ *  This method is optimized to a NO-OP when in fast evaluation mode
+ */
+void eos_assert_ripemd160( char* data, uint32_t length, const checksum_base* hash );
```
heifner: These need to be all eosio_*

tests/wasm_tests/wasm_tests.cpp
```diff
@@ -264,6 +267,163 @@ BOOST_FIXTURE_TEST_CASE( test_api_bootstrap, tester ) try {
       BOOST_CHECK_EQUAL(transaction_receipt::executed, receipt.status);
    }
 } FC_LOG_AND_RETHROW() /// test_api_bootstrap
+#endif
+
+BOOST_FIXTURE_TEST_CASE( test_proxy, tester ) try {
```
elmato: proxy contract tests are now in currency_test.cpp

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -774,10 +774,9 @@ using apply_handler = std::function<void(apply_context&)>;
 
       if( itr == idx.end() ||
           itr->t_id != t_id.id ||
-          !impl::record_scope_compare<IndexType, Scope>::compare(*itr, keys)) return -1;
+          !impl::record_scope_compare<IndexType, Scope>::compare(*itr, keys)) return 0;
```
elmato: shouldn't we keep the -1 return value? 
How we can distinguish between reading a zero sized record (besides keys) and a "record not found" error.

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -799,7 +798,7 @@ using apply_handler = std::function<void(apply_context&)>;
 
       auto itr = idx.lower_bound( tuple );
       if( itr == idx.end() ||
-         itr->t_id != t_id.id ) return -1;
+         itr->t_id != t_id.id ) return 0;
```
elmato: Same as above

contracts/eosiolib/checktime.h
```diff
@@ -0,0 +1,19 @@
...
+
+extern "C" {
+  /**
+   *  @defgroup checktime API
+   *  @brief Declares checktime function. 
+   *  @ingroup checktimeapi
+   *
+   *  @{
+   */
+   void checktime();
```
wanderingbort: I think its preferable to not treat `chectime` as part of the ABI.  I don't think we want contracts calling this directly. @spoonincode may want to weigh in.

contracts/test_api/test_account.cpp
```diff
@@ -4,29 +4,43 @@
...
 
-unsigned int test_account::test_balance_acc1()
-{
+void test_account::test_balance_acc1() {
    eosio::print("test_balance_acc1\n");
-   eosio::account::account_balance b;
-
+   //eosio::account::account_balance b;
+   
+/*
```
wanderingbort: Go ahead and delete this test as opposed to merely commenting it out.

contracts/test_api/test_privileged.cpp
```diff
@@ -0,0 +1,21 @@
...
+#include "test_api.hpp"
+
+/*
+void test_privileged::set_privileged() {
+   prints("hello\n");
+}
+*/
+
+void test_privileged::test_is_privileged() {
+   prints("HELLO\n");
```
wanderingbort: we should probably have these tests run silently instead of printing out debugging totems.

contracts/test_api/test_transaction.cpp
```diff
@@ -4,106 +4,206 @@
...
-   for (int i = 0; i < 10; i++) {
-      transaction_create();
-   }
+void test_transaction::test_transaction_size() {
+   uint32_t trans_size = 0;
+   read_action( (char*)&trans_size, sizeof(uint32_t) );
+   
+   eosio_assert( trans_size == transaction_size(), "transaction size does not match" );
+}
+/*
```
wanderingbort: this looks like a stale copy of the test? can it be removed?

contracts/test_api_db/test_api_db.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+      WASM_TEST_HANDLER(test_db, primary_i64_general);
+      WASM_TEST_HANDLER(test_db, primary_i64_lowerbound);
+      WASM_TEST_HANDLER(test_db, primary_i64_upperbound);
+      WASM_TEST_HANDLER(test_db, idx64_general);
+      WASM_TEST_HANDLER(test_db, idx64_lowerbound);
+      WASM_TEST_HANDLER(test_db, idx64_upperbound);
+
+      WASM_TEST_HANDLER(test_db, key_str_general);
+#if 0
```
wanderingbort: since we are deprecating the old API, we can go ahead and remove these stubs too

contracts/test_api_db/test_db.cpp
```diff
@@ -64,114 +64,122 @@ extern "C" {
...
+     for (uint32_t i = 0; i < n; i++) {
+        if (c1[i] != c2[i]) {
+           return false;
+        }
+     }
+     return true;
+  }
 }
 
 #if 0
```
wanderingbort: remove stale/old tests

contracts/test_api_mem/test_memory.cpp
```diff
@@ -0,0 +1,313 @@
...
+void verify_mem(const void* const ptr, const uint32_t val, const uint32_t size)
+{
+   const char* char_ptr = (const char*)ptr;
+   for (uint32_t i = 0; i < size; ++i)
+   {
+      eosio_assert(static_cast<uint32_t>(static_cast<unsigned char>(char_ptr[i])) == val, "buf slot doesn't match");
+   }
+}
+
+/*
```
wanderingbort: this looks like a debugging helper function, can it be deleted?

libraries/chain/wasm_interface.cpp
```diff
@@ -43,27 +45,27 @@ using boost::asio::io_service;
       /**
       * Name of the account who's balance this is
       */
-      account_name account;
+      eosio::chain::account_name account;
```
wanderingbort: go ahead and remove this if 0'd section

libraries/test.err
```diff
@@ -0,0 +1 @@
+/Users/judgefudge/scripts/run_tests: line 7: /Users/judgefudge/bucky_eos/eos/libraries/chain_test: No such file or directory
```
wanderingbort: gonna guess this wasn't supposed to get checked in ... judgefudge :smirk:

wanderingbort: nor the test.out which is empty

tests/api_tests/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+#add_wast_target(test_api "${CMAKE_SOURCE_DIR}/contracts" ${CMAKE_CURRENT_BINARY_DIR})
```
wanderingbort: is this file necessary if they are all commented out?

tests/common/database_fixture.hpp
```diff
@@ -4,12 +4,14 @@
  */
 #pragma once
 
-#include <eos/chain/chain_controller.hpp>
-#include <eos/chain/producer_object.hpp>
-#include <eos/chain/exceptions.hpp>
+#include <eosio/chain/chain_controller.hpp>
```
wanderingbort: Is this file in use for anything?  can it be deleted.

tests/common/macro_support.hpp
```diff
@@ -10,7 +10,7 @@
  * @file Contains support macros for the testcase helper macros. These macros are implementation details, and thus
  * should not be used directly. Use their frontends instead.
  */
-
+#if 0
```
wanderingbort: same here, can we just delete this file now?   nothing shoudl be using the old Macros anymore

tests/wasm_tests/wasm_tests.cpp
```diff
@@ -226,44 +202,105 @@ BOOST_FIXTURE_TEST_CASE( abi_from_variant, tester ) try {
 
 } FC_LOG_AND_RETHROW() /// prove_mem_reset
 
-BOOST_FIXTURE_TEST_CASE( test_api_bootstrap, tester ) try {
+BOOST_FIXTURE_TEST_CASE( test_deferred_failure, tester ) try {
```
wanderingbort: pretty sure this test was moved to the currency tests as it required currency... is this file disabled from running?

contracts/eosiolib/checktime.h
spoonincode: Yeah I don't think we should expose it (we have a ticket somewhere where we intend to actively refuse linkage to it in the future). This call is implementation detail.

larryk85: Yeah, this was an oversight on my part.  I exposed it so that I could see what it was actually doing and forgot to delete it from the header.

contracts/test_api/test_privileged.cpp
larryk85: These tests are not finished yet, and I didn't mean to include these with this PR.

PR #1380, opened 2018-02-14
Test framework improvement - stat576

cj-oci:
Move chain initialization to sub class so tests can be written with chain in various states of initialization.


tests/chain_tests/block_tests.cpp
```diff
@@ -36,7 +36,8 @@ BOOST_AUTO_TEST_CASE( schedule_test ) { try {
 } FC_LOG_AND_RETHROW() }/// schedule_test
 
 BOOST_AUTO_TEST_CASE( push_block ) { try {
-   tester test1, test2(chain_controller::runtime_limits(), false);
+   tester test1;
+   base_tester test2(chain_controller::runtime_limits{});
```
wanderingbort: the explicit default of runtime_limits should be un-needed

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -101,6 +100,11 @@ namespace eosio { namespace testing {
          map<transaction_id_type, transaction_receipt> chain_transactions;
    };
 
+   class tester : public base_tester {
+   public:
```
wanderingbort: Should we move push_genesis_block to this class from base_tester?

cj-oci: Yes and done.

PR #1388, opened 2018-02-14
Unit tests for new db API's

zorba80:


libraries/chain/wasm_interface.cpp
```diff
@@ -727,13 +727,12 @@ class database_api : public context_aware_api {
       void db_idx64_remove( int iterator ) {
          return context.idx64.remove( iterator );
       }
-      int db_idx64_find_secondary( uint64_t code, uint64_t scope, uint64_t table, uint64_t& secondary, uint64_t& primary ) {
+      int db_idx64_find_secondary( uint64_t code, uint64_t scope, uint64_t table, const uint64_t& secondary, uint64_t& primary ) {
```
moskvanaft: Minor issue - here we are passing all read-only uint64_t parameters by value, except "secondary" which is passed by const reference. Seems inconsistent.

tests/wasm_tests/wasm_tests.cpp
```diff
@@ -830,5 +830,96 @@ BOOST_FIXTURE_TEST_CASE( check_table_maximum, tester ) try {
...
+      
+      set_tapos(trx);
+      trx.sign(get_private_key(N(tester), "active"), chain_id_type());
+      push_transaction(trx);
+      produce_block();
+      
+      BOOST_REQUIRE_EQUAL(true, chain_has_transaction(trx.id()));
+   }
+
+   {
```
moskvanaft: Minor issue - It seems like all the block down from this line are very similar. The only thing that different is function name that gets called. So, maybe it's possible to just do a loop over function names.
Also there is a function tester::push_action (libraries/testing/tester.cpp, usage example - tests/wasm_tests/identity_tests.cpp) which might be used here to reduce boilerplate code.

PR #1389, opened 2018-02-14
Make all wasm instances share same linear memory

spoonincode:
Previously, each wasm instance (smart contract instance) would allocate 12GB of virtual address space for it self. While not as dire as how table memory use to work, this still will exhaust virtual memory space at an alarmingly low number of contracts.

Since in this release execution is single threaded, for now just make every wasm instance share the same 12GB. In the future we can extend this to have a “thread pool of memories”.

Review notes: chain_test passes but I still feel a little queasy about the lack of extensive tests against this change. I think we still have a number of unit tests not merged for 3.0 changes though? So I was hesitant to work on making new unit tests that simply overlap with what we have non-merged anyways.

### DIFF
--- a/libraries/chain/wasm_interface.cpp
+++ b/libraries/chain/wasm_interface.cpp
@@ -125,13 +125,12 @@ namespace eosio { namespace chain {
        * the instance handed out to other threads
        */
       struct code_info {
-         code_info( size_t mem_end, vector<char>&& mem_image )
-         :mem_end(mem_end),mem_image(std::forward<vector<char>>(mem_image))
+         code_info( vector<char>&& mem_image )
+         :mem_image(std::forward<vector<char>>(mem_image))
          {}
 
          // a clean image of the memory used to sanitize things on checkin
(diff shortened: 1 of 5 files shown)


libraries/chain/wasm_interface.cpp
```diff
@@ -228,19 +226,22 @@ namespace eosio { namespace chain {
...
-                           mem_end = i + 1;
-                     }
-                     mem_image.resize(mem_end);
-                     memcpy(mem_image.data(), mem_ptr, mem_end);
+                  //populate the module's data segments in to a vector so the initial state can be
+                  // restored on each invocation
+                  //Be Warned, this may need to be revisited when module imports make sense. The
+                  // code won't handle data segments that initalize an imported memory which I think
+                  // is valid.
+                  for(const DataSegment& data_segment : module->dataSegments) {
```
spoonincode: If there is anything that could use some extra eyeballs, it's this area. Mainly want to double check no off by one.

PR #1394, opened 2018-02-15
Add Jenkinsfile, build dependencies with max jobs.

ericiles:
This PR contains an initial Jenkinsfile to start doing our builds on Jenkins. Additionally, I made small modifications to the build scripts for each platform to build dependencies with the maximum number of jobs in order to speed up build time. Finally, the Ubuntu build script was only doing a make for the LLVM/WASM dependency, causing builds to fail. I updated this to do a make install, and the builds are now successful.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,38 @@
+pipeline {
+    agent any
+    stages {
+        stage('Build') {
+            parallel {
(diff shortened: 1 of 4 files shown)


heifner:
Reviewed by @pacificcode

PR #1412, opened 2018-02-17
[cmake] add FindWasm.cmake and binaryen code into the repo

asiniscalchi:
it allows the cmake syntax:

find_package(Binaryen)
find_package(Wast)

and the cmake standard syntax
$ cmake .. -DWAST_ROOT=\<path\> -DBINARYEN_ROOT=\<path\>


CMakeModules/wasm.cmake
```diff
@@ -1,61 +1,8 @@
...
-  endif()
-
-  message(STATUS "BINARYEN_BIN => " ${BINARYEN_BIN})
-
-endif()
-
-# TODO: Check if compiler is able to generate wasm32
-if( NOT ("${WASM_CLANG}" STREQUAL "" OR "${WASM_LLC}" STREQUAL "" OR "${WASM_LLVM_LINK}" STREQUAL "" OR NOT BINARYEN_BIN) )
-  set(WASM_TOOLCHAIN TRUE)
+if(WASM_FOUND AND BINARYEN_FOUND)
```
wanderingbort: Given that `find_package` has Wasm and Binaryen as required, is there any purpose to this conditional or the dependent one in the root `CMakeLists.txt` that handles "soft" failure of the wasm toolchain?

plugins/txn_test_gen_plugin/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-if(WASM_TOOLCHAIN)
+if(WASM_TOOLCHAIN_FOUND)
```
wanderingbort: At this point, there is no "successful" cmake run where `WASM_TOOLCHAIN_FOUND` would be false.  Most of these conditionals can be removed in favor of running the side of the branch that corresponds with success.

CMakeModules/wasm.cmake
asiniscalchi: It dependes on what you want: master branch already fails if toolchain is not found.
What do you prefer ?

plugins/txn_test_gen_plugin/CMakeLists.txt
asiniscalchi: yes. I was trying to mimic the master branch behavior

asiniscalchi:
So in the end the question is:
Do you want the contracts compilation to be mandatory ?

wanderingbort:
For a while, it was mandatory as the chain library needed compilation for the system contract.  I suppose at this point it could return to optional but we have already leaned in to using WASM to exercise the chain in our integration/smoke tests.  It is impossible to test some parts of the chain functionality without contracts working.  

TL;DR we should assume its mandatory and strip out anything thats now unnecessary as a result.

Also, the OSX build needs love, even if its only changing the scripts to properly interface with your new modules

asiniscalchi:
Done.
You have the binaryen submodule into the repo.

asiniscalchi:
Done.
You have the binaryen submodule into the repo.

wanderingbort:
I tried to run this through our build scripts which rely on (among otherthings) WASM_LLVM_CONFIG instead of WASM_ROOT

please consider either maintaining WASM_LLVM_CONFIG as a single entry for configuring all endpoints or changing the build scripts to properly use WASM_ROOT

asiniscalchi:
I didn't find any other WASM_LLVM_CONFIG in the repo

PR #1423, opened 2018-02-19
P2p queue unity master

pmesnier:
I have yet to build a test with the EOS BIOS model that actually generates blocks from transactions. But at least this patch works with dawn-2.x. This patch includes all of the changes necessary to support block summaries and packed transactions. I also updated the connection objects to use a single output queue. This also includes further improvements to the catchup sync and big message managers.

### DIFF
--- a/libraries/chain/include/eosio/chain/block.hpp
+++ b/libraries/chain/include/eosio/chain/block.hpp
@@ -81,22 +81,22 @@ namespace eosio { namespace chain {
     *
     *  The primary purpose of a block is to define the order in which messages are processed.
     *
-    *  The secodnary purpose of a block is certify that the messages are valid according to 
+    *  The secodnary purpose of a block is certify that the messages are valid according to
     *  a group of 3rd party validators (producers).
     *
     *  The next purpose of a block is to enable light-weight proofs that a transaction occured
     *  and was considered valid.
     *
     *  The next purpose is to enable code to generate messages that are certified by the
-    *  producers to be authorized. 
+    *  producers to be authorized.
     *
     *  A block is therefore defined by the ordered set of executed and generated transactions,
     *  and the merkle proof is over set of messages delivered as a result of executing the
-    *  transactions. 
+    *  transactions.
     *
     *  A message is defined by { receiver, code, function, permission, data }, the merkle
     *  tree of a block should be generated over a set of message IDs rather than a set of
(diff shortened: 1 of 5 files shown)


libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -109,9 +109,16 @@ namespace eosio { namespace chain {
...
-    * transactions are not included.  
+    * transactions are not included.
     */
    struct signed_block : public signed_block_summary {
+      signed_block () = default;
+      signed_block (const signed_block& ) = default;
+      signed_block (const signed_block_summary&& base)
+         :signed_block_summary (base),
+          input_transactions()
+      {}
```
heifner: This is not moving `base`, it is copying it. You need `signed_block_summary(std::move(base))`.  Also it should not take the rvalue via const since you will not be able to move it. Also const rvalue reference is generally used to prevent binding to an rvalue which it would do if you also provided a non-const rvalue reference. You should also provide a default move constructor to prevent slicing when actually moving a signed_block.

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -109,9 +109,16 @@ namespace eosio { namespace chain {
     * what would be logged to disk to enable the regeneration of blockchain state.
     *
     * The transactions are grouped to mirror the cycles in block_summary, generated
-    * transactions are not included.  
+    * transactions are not included.
```
heifner: All these extra modifications cause a lot of noise.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2172,51 +2142,19 @@ namespace eosio {
...
-      auto entry = local_txns.get<by_id>().find( txnid );
-      if(entry != local_txns.end( ) ) {
-         if( entry->validated ) {
-            fc_dlog(logger, "the txnid is known and validated, so short circuit" );
-            return;
-         }
-      }
       if( sync_master->is_active(c) ) {
-         fc_dlog(logger, "got a txn during sync - dropping");
+         fc_dlog(logger, "got a txn during sync - droppXning");
```
PaulCalabrese: Typo.

heifner:
@spoonincode reviewed and is going to do a diff PR for a different Merkle proof calculation.

PR #1430, opened 2018-02-20
Fix typo & correct Boost URL

kirkins:
Hi just reading through the docs, I think this is a typo.

If not please disregard.


kirkins:
Going through the instructions for Ubuntu 16.04. The url was invalid for boost, looks like it was updated from 64, but only in one place.

PR #1434, opened 2018-02-20
Eosio darwin openssl 1.0

pacificcode:
This PR addresses an issue with Home Brew installing different versions of openssl on Darwin. The fix is to force home-brew to install openssl version 1.0.
Files changes:
scripts/eosio_build_darwin.sh


scripts/eosio_build_darwin.sh
```diff
@@ -110,6 +110,10 @@
 					pkg="llvm@4"
 				fi
 
+				if [ $pkg = "openssl" ]; then
+					pkg="openssl@1.0"
```
SheldonHH: version should be changed to 1.1

PR #1441, opened 2018-02-20
Cmake deps fix

gleehokie:
cmake changes from Venkata.

### DIFF
--- a/libraries/fc/CMakeLists.txt
+++ b/libraries/fc/CMakeLists.txt
@@ -47,6 +47,13 @@ IF(NOT "$ENV{OPENSSL_ROOT_DIR}" STREQUAL "")
   message(STATUS "Setting up OpenSSL root and include vars to ${OPENSSL_ROOT_DIR}, ${OPENSSL_INCLUDE_DIR}")
 ENDIF()
 
+IF( NOT OPENSSL_ROOT_DIR )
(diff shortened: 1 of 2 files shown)


libraries/fc/CMakeLists.txt
```diff
@@ -47,6 +47,13 @@ IF(NOT "$ENV{OPENSSL_ROOT_DIR}" STREQUAL "")
   message(STATUS "Setting up OpenSSL root and include vars to ${OPENSSL_ROOT_DIR}, ${OPENSSL_INCLUDE_DIR}")
 ENDIF()
 
+IF( NOT OPENSSL_ROOT_DIR )
+   set( OPENSSL_ROOT_DIR /usr/local/opt/openssl CACHE FILEPATH "" )
```
wanderingbort: This should use the standard cmake functions like `find_path` to find appropriate paths based on hints and _fail_ if it does not find the path instead of setting an arbitrary path and expecting it to fail later if it is wrong.

wanderingbort:
@gleehokie We should probably re-assign this and have someone fix the issues.

gleehokie:
> @gleehokie  We should probably re-assign this and have someone fix the issues.

I agree @wanderingbort, assigned to @larryk85.

wanderingbort:
Closing this, its tracked in JIRA

PR #1442, opened 2018-02-20
Del key64x64x64 value object

gleehokie:
key64x64x64 changes from Venkata.  @bytemaster please review.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -473,10 +473,4 @@ contracts::table_key_type apply_context::get_key_type<contracts::key64x64_value_
    return contracts::table_key_type::type_i64i64;
 }
 
-/*
-template<>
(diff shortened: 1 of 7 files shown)


moskvanaft:
If we marge this to master unit-tests will be broken. I was going to merge this to my local branch where I switched contracts to new DBAPI. And then create PR to merge both my and Venkata's changes to master.
Now I am working on radical changes than we have in Venkata's branch -  I want to remove entire old DBAPI (this PR is only about i64i64i64). So, maybe we will not need this PR after all.

moskvanaft:
I've removed entire DB API in my branch. So, we don't need this PR.
I have not created PR for my changes yet because my changes depend on fixes to new DB API made by Aarin, which are not in master yet. I would like to create my PR after his changes got merged to master. So I would not have changes which are not mine in my PR.

I think we can close this PR.

wanderingbort:
that API will be deprecated soon and these changes are not ready for master at this time.

PR #1445, opened 2018-02-20
RePR: Add a find_package(Clang) for abi_generator

wanderingbort:
@spoonincode wanted to investigate why this breaks travis


spoonincode:
Deep six this.

Nearly 100% sure I'm right on this but something seems busted with how the cmake modules are defined in the Ubuntu packages.

Issue #1446, opened 2018-02-21
DAWN-623 ⁃ fix for echoed transactions.


pmesnier:
This situation was discovered after merging in pr #1423 has a bug that causes some transactions to be echoed back to the source.

coreylederer:
Phil indicates this is completed

PR #1447, opened 2018-02-21
fix for the problem #1446

pmesnier:
the big_message_manager was incorrectly limiting the assignment of a connection to skip when forwarding a received transactions.


spoonincode:
Improved p2p for me -- no more duplicate transactions trying to be applied

PR #1459, opened 2018-02-22
Increase wasm (smart contract) max memory to 33MiB

spoonincode:
There is a requirement to support a 32MiB dataset in wasm memory. Make the maximum memory allowed 33MiB. 33 was chosen via fair dice roll; if we feel 32.064 is the most we want to push we can easily change it.

See each individual commit for a better description of changes. Most notable is that I had to disable the unit test that mixes sbrk() and malloc() because I modified malloc's heap growth to always create heaps on wasm page boundaries.


libraries/chain/include/eosio/chain/wasm_eosio_constraints.hpp
```diff
@@ -7,11 +7,17 @@ namespace IR {
 namespace eosio { namespace chain {
 
 namespace wasm_constraints {
-   //Be aware that some of these are required to be a multiple of some internal number
-   constexpr unsigned maximum_linear_memory      = 1024*1024;  //bytes
-   constexpr unsigned maximum_mutable_globals    = 1024;       //bytes
-   constexpr unsigned maximum_table_elements     = 1024;       //elements
-   constexpr unsigned maximum_linear_memory_init = 64*1024;    //bytes
+   constexpr unsigned maximum_linear_memory      = 33*1024*1024;//bytes
```
SheldonHH: under `wasm_eosio_constraints`, it increase the memory size to `33MB`

contracts/eosiolib/eosiolib.cpp
```diff
@@ -34,31 +34,39 @@ namespace eosio {
 
       memory* next_active_heap()
       {
+         constexpr uint32_t wasm_page_size = 64*1024;
```
SheldonHH: define the `wasm_page_size` to 64 Kibibyte

contracts/eosiolib/eosiolib.cpp
```diff
@@ -461,7 +469,6 @@ namespace eosio {
       static const uint32_t _mem_block = 8;
       static const uint32_t _rem_mem_block_mask = _mem_block - 1;
       static const uint32_t _initial_heap_size = 8192;//32768;
-      static const uint32_t _new_heap_size = 65536;
```
SheldonHH: `_new_heap_size` specified as `wasm_page_size` in the new commit

libraries/chain/wasm_interface.cpp
```diff
@@ -1030,17 +1030,17 @@ class memory_api : public context_aware_api {
...
 
-      uint32_t sbrk(int num_bytes) {
+      int sbrk(int num_bytes) {
          // sbrk should only allow for memory to grow
          if (num_bytes < 0)
             throw eosio::chain::page_memory_error();
          // TODO: omitted checktime function from previous version of sbrk, may need to be put back in at some point
          constexpr uint32_t NBPPL2  = IR::numBytesPerPageLog2;
-         constexpr uint32_t MAX_MEM = 1024 * 1024;
+         constexpr uint32_t MAX_MEM = wasm_constraints::maximum_linear_memory;
```
SheldonHH: increase the `MAX_MEM` from `1 Mib` to `33Mib`

libraries/chain/wasm_interface.cpp
```diff
@@ -1030,17 +1030,17 @@ class memory_api : public context_aware_api {
          return (char *)::memset( dest, value, length );
       }
 
-      uint32_t sbrk(int num_bytes) {
+      int sbrk(int num_bytes) {
```
SheldonHH: typo: the return variable `prev_num_bytes` is [uint32_t](https://github.com/EOSIO/eos/blob/709a68eaf7bf9f0c872ee94f20048fa131c911e5/libraries/chain/wasm_interface.cpp#L1047)

contracts/test_api_mem/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
+#set the MAX_MEMORY to 1MB for these tests; there were lots of memory unit tests that assume such
```
SheldonHH: 1048576 = 1024*1024, the comment should be 1MIB

CMakeModules/wasm.cmake
```diff
@@ -165,7 +165,7 @@ macro(add_wast_library)
 endmacro(add_wast_library)
 
 macro(add_wast_executable)
-  cmake_parse_arguments(ARG "NOWARNINGS" "TARGET;DESTINATION_FOLDER" "SOURCE_FILES;INCLUDE_FOLDERS;SYSTEM_INCLUDE_FOLDERS;LIBRARIES" ${ARGN})
+  cmake_parse_arguments(ARG "NOWARNINGS" "TARGET;DESTINATION_FOLDER;MAX_MEMORY" "SOURCE_FILES;INCLUDE_FOLDERS;SYSTEM_INCLUDE_FOLDERS;LIBRARIES" ${ARGN})
```
SheldonHH: `MAX_MEMORY` is assigned value on `test_api/CMakeLists.txt` and `test_api_mem/CMakeLists.txt`

spoonincode:
Just occurred to me that all of tests/wasm_tests have been disabled on the large test PR that recently went in (and was included in the sync to master on this branch). Recommending to delay this PR until the tests are reenabled as I think some of those tests may fail and need proper modifications.

libraries/chain/wasm_interface.cpp
SheldonHH: I think the return type of `sbrk` should still be `uint32_t`

spoonincode: I changed it to int to be a more clear indicator that `sbrk()` can now return a negative value. While I realize it could stay `uint32_t` and still return -1 or even 0xFFFFFFFF that doesn't seem as explicit.

PR #1464, opened 2018-02-22
DAWN-538 Added method to count instructions...

brianjohnson5972:
#1260 
 and inject as parameter to checktime.

### DIFF
--- a/libraries/chain/wasm_eosio_instruction_count.cpp
+++ b/libraries/chain/wasm_eosio_instruction_count.cpp
@@ -64,8 +64,6 @@ struct special_instruction_count_visitor : public opcode_instruction_count_visit
       if (instructionCounts.empty())
          FC_THROW_EXCEPTION(wasm_execution_error, "Smart Contract encountered end instruction with nothing to end");
 
-#warning TODO: need to determine what the trailing instructions are
-
       uint32_t residule_count = 0;
       {
          auto& instructionCount = instructionCounts.back();
@@ -90,7 +88,6 @@ struct special_instruction_count_visitor : public opcode_instruction_count_visit
 };
 
 void wasm_esio_instruction_count::inject_instruction_count(const IR::Module& module) {
-   //wlog("inject_instruction_count start");
    for(const FunctionDef& fd : module.functions.defs) {
       special_instruction_count_visitor visitor;
       OperatorDecoderStream decoder(fd.code);
--- a/libraries/chain/wasm_eosio_instruction_count.cpp
+++ b/libraries/chain/wasm_eosio_instruction_count.cpp
@@ -1,0 +1,102 @@
+#include <eosio/chain/wasm_eosio_instruction_count.hpp>
+#include <fc/exception/exception.hpp>
+#include <eosio/chain/exceptions.hpp>
+#include "IR/Module.h"
+#include "IR/Operators.h"
+
+namespace eosio { namespace chain {
+
+using namespace IR;
+
+struct InstructionCount {
+   InstructionCount(LiteralImm<I32>* i32_imm) : i32_imm(i32_imm) {
+   }
+
+   uint32_t count = 0;
+   LiteralImm<I32>* i32_imm;
+};
+
+typedef std::vector<InstructionCount> InstructionCounts;
+
+struct opcode_instruction_count_visitor {
+   typedef void Result;
+
+#define VISIT_OPCODE(opcode,name,nameString,Imm,...) \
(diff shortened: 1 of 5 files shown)


libraries/chain/wasm_eosio_instruction_count.cpp
```diff
@@ -0,0 +1,102 @@
...
+      }
+
+      opcode_instruction_count_visitor::i32_const(imm);
+   }
+
+   void end(NoImm& imm) override {
+      if (instructionCounts.empty())
+         FC_THROW_EXCEPTION(wasm_execution_error, "Smart Contract encountered end instruction with nothing to end");
+
+#warning TODO: need to determine what the trailing instructions are
```
spoonincode: What exactly is going on here? Sounds like we don't understand certain situations?

libraries/chain/wasm_eosio_instruction_count.cpp
```diff
@@ -0,0 +1,102 @@
...
+
+   void if_(ControlStructureImm& imm) {
+      opcode_instruction_count_visitor::if_(imm);
+      // no count structure, but still need a stack object to track the count
+      instructionCounts.emplace_back(nullptr);
+   }
+};
+
+void wasm_esio_instruction_count::inject_instruction_count(const IR::Module& module) {
+   //wlog("inject_instruction_count start");
```
spoonincode: Can we zap this

libraries/chain/wasm_eosio_instruction_count.cpp
brianjohnson5972: I forgot to remove the warning (was an issue I was tracking down and fixed), I will remove.

libraries/chain/wasm_eosio_instruction_count.cpp
brianjohnson5972: Yep, missed deleting it.

spoonincode:
This looks good to me though as you and others are aware we may need to change the location of the injection to earlier in the stack so that we can accommodate multiple runtime with a single implementation. This is work to be done later though.

A couple things I have noticed from playing around for a little--

The checktime() call injection gets counted as instructions; this is a little unfortunate since it leaks implementation details of how we implemented counting in to the resulting "wasm instructions" computation. But this isn't too bad since it means that a non-intrinsic-based-checktime() would just have to count each br/loop as 3 instructions in the future.

if/elses aren't computed fairly. I'm not sure how much we ultimately care though. Consider the following wasm--

```
(module
 (table 0 anyfunc)
 (memory $0 16)
 (export "memory" (memory $0))
 (export "apply" (func $apply))
 
 (func $apply (param $0 i64) (param $1 i64)
   (local i32)
   (set_local 2 (i32.const 100000))
   (loop
     (set_local 2 (i32.sub (get_local 2) (i32.const 1)))
     (if (get_local 2) (then br 1)
     (else
       (drop (i64.mul (i64.const 56) (i64.const 34)))
       (drop (i64.mul (i64.const 12) (i64.const 55)))
       (drop (i64.mul (i64.const 536) (i64.const 2333)))
       (drop (i64.mul (i64.const 23) (i64.const 12)))
       (drop (i64.mul (i64.const 543) (i64.const 76533)))
     ))
   )
 )
)
```
This loop takes the "simple" path for all but the last execution of the loop. Yet, for each loop it is billed for 30 instructions no matter what. Since block/loop/if/else are the only "structured instructions" in WASM this does make me wonder if we should promote if/else to be treated just like block/loop for purposes of checktime.

spoonincode:
Branches, yes. For some reason I was thinking if/else was the only issue but consider this wasm
```
(module
 (table 0 anyfunc)
 (memory $0 16)
 (export "memory" (memory $0))
 (export "apply" (func $apply))

 (func $complex_stuff_and_return_a_bool (result i32)
   (i32.const 0)
 )
 
 (func $apply (param $0 i64) (param $1 i64)
   (br_if 0 (call $complex_stuff_and_return_a_bool))

   (drop (i64.mul (i64.const 56) (i64.const 34)))
   (drop (i64.mul (i64.const 12) (i64.const 55)))
   (drop (i64.mul (i64.const 536) (i64.const 2333)))
   (drop (i64.mul (i64.const 23) (i64.const 12)))
   (drop (i64.mul (i64.const 543) (i64.const 76533)))
 )
)
```
It suffers from the same problem -- the apply call always bills as 24 instructions no mater what complex_stuff_and_return_a_bool returns.

PR #1466, opened 2018-02-23
Add multi-sig recovery tests. stat-291

cj-oci:
Add multi-sig recovery test.
Enable misc_tests.


wanderingbort:
@cj-oci please address any of the comments in a new PR if you agree with them.

PR #1467, opened 2018-02-23
Remove memset on malloc heap init

spoonincode:
The wasm runtime guarantees that fresh pages are zero filled. Not seeing any reason to do a memset on heap init since that memory will be zeroed already anyways.

### DIFF
--- a/contracts/eosiolib/eosiolib.cpp
+++ b/contracts/eosiolib/eosiolib.cpp
@@ -202,7 +202,6 @@ namespace eosio {
          {
             _heap_size = size;
             _heap = mem_heap;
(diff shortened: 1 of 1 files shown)


contracts/eosiolib/eosiolib.cpp
```diff
@@ -202,7 +202,6 @@ namespace eosio {
          {
             _heap_size = size;
             _heap = mem_heap;
-            memset(_heap, 0, _heap_size);
```
SheldonHH: `_heap` declared as `char*` in  method: `memory* next_active_heap()` under class`memory_manager` in memory.hpp together with uint32_t `_heap_size` and `_offset`

PR #1478, opened 2018-02-26
Fix/reintegrate wasm tests

larryk85:
Removed some files that made their way in in last PR, fixed one small oversight with new db api and re-enabled WASM tests.  I had to comment out most of identity tests for now, as these require the new db api to run correctly.


tests/wasm_tests/identity_tests.cpp
```diff
@@ -382,6 +382,8 @@ BOOST_FIXTURE_TEST_CASE( trust_untrust, identity_tester ) try {
 
 } FC_LOG_AND_RETHROW() //trust_untrust
 
+// TODO: Anton please re-enable when finished with new db api changes
```
wanderingbort: @moskvanaft just noting this for you

PR #1482, opened 2018-02-27
DAWN-551 Contract Debugging general approach for native contract debug

PaulCalabrese:
This PR is intended for review and discussion purposes only. It definitely needs some more work before merging. See the DAWN-551 issue in JIRA for my implementation notes and questions/issues.


spoonincode:
I have two high level things to think about --

One, have you considered using an .so for the contract instead? Using an .so (and some glue) would allow users to redeploy their debugging contract without having to rebuild and restart eosiod. And if they're at the point of trying to use the debugger, that's probably something they're going to do a lot of. Also, I can envision a time when contract developers never build eosiod -- they simply use prebuilts of eosiod and the "eosio sdk"

Secondly, I'm worried about the stark differences in runtime behavior of wasm vs this native environment --
- Different libc
- Different libc++
- Different malloc() behavior
-- Native is "unlimited" while wasm side is quite limited
-- Native probably is not zeroed while wasm side almost always is
- Wildly different stack sizes
- Type differences like size of pointers
- Initializers in wasm not called, will be in native
- Variables not reset between action execution
- Soft float vs hard float
- And of course other potential oddities like a misbehaving contract potentially able to corrupt eosiod and/or its database

This makes me concerned that debugging will end up fighting users as they'll try to debug their contract and end facing a whole swath of different behavior. But, all of the above is fixable (or can be nearly entirely mitigated) by placing these native debugging contracts in a sandboxed process instead.

PaulCalabrese:
Thanks for the comments, Matt, there is a lot to think about, here.

I wasn't involved in some of the initial discussions of contract debugging,
but it was my understanding that it was decided to investigate this
native-compiled contract approach as a relatively simple way to give
contract developers access to a debugger. It's obvious that somehow
debugging directly in the wasm environment would allow for more fidelity,
but would involve a much bigger development effort so the approach here is
something of a compromise. That being said, I'm trying to make the
debugging environment as valuable as possible.

I like the idea of loading the compiled contract via a shared library. It
might make it possible to not have to re-build the debug plugin. I'll
investigate this.

Your list of other potential issues is a good list of differences between
almost any native compilation approach and the wasm engines. A lot of these
should not impact many smart contracts (except those really pushing the
boundaries on things like memory usage) and I think are reasonable to
accept for the ability to verify/test the logic of your contract in the
debugger. It is important that we document all differences and warn people
about them.

A couple of the more important differences include:

   - Native malloc memory is not zeroed while wasm side almost always is
   - Type differences like size of pointers (and other types)
   - Initializers in wasm not called, will be in native
   - Variables not reset between action execution
   - Soft float vs hard float

We need to investigate these and determine what we can do about them (if
anything) and what the impact of each is.


I am not worried about misbehaving contracts potentially able to corrupt
eosiod and/or its database as I see this as strictly for use in a
standalone node that you are debugging your contract with.

Placing the native contract in a standalone process would solve some, but
not all of the above problems (there would still be different libraries,
type sizes, heap, stack, and other differences). Would we launch the
process each time an action occurs on the contract? This would make it hard
to debug its behavior.

Paul

Issue #895, opened 2017-12-07
make  Error 127***


boyue:
Hey ,My name is CrashHunter, I come from Youzan company in China.
I occur an error when run :
./build.sh darwin

the error is :
-- Build files have been written to: /Users/rancho/eos/build
[  1%] Generating WAST infinite.wast
[  1%] Generating WAST proxy.wast
[  1%] Generating WAST currency.wast
[  1%] Generating WAST test_api.wast
/bin/sh: /usr/local/binaryen/bin/s2wasm: No such file or directory
/bin/sh: /usr/local/binaryen/bin/s2wasm: No such file or directory
/bin/sh: /usr/local/binaryen/bin/s2wasm: No such file or directory
/bin/sh: /usr/local/binaryen/bin/s2wasm: No such file or directory
make[2]: make[2]: *** [contracts/test_api/test_api.wast] Error 127*** [contracts/proxy/proxy.wast] Error 127

make[2]: *** [contracts/currency/currency.wast] Error 127
make[2]: *** [contracts/infinite/infinite.wast] Error 127
make[1]: *** [contracts/proxy/CMakeFiles/proxy.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [contracts/test_api/CMakeFiles/test_api.dir/all] Error 2
make[1]: *** [contracts/currency/CMakeFiles/currency.dir/all] Error 2
make[1]: *** [contracts/infinite/CMakeFiles/infinite.dir/all] Error 2
make: *** [all] Error 2

can you help me?

heifner:
@boyue Please use Telegram `EOS Developers` for support.

You might take a look at: https://github.com/EOSIO/eos/pull/871/files and see if those changes fixes your issue. Either way, let us know on telegram.

Haley-Wong:
@boyue 你的问题解决了么？我也遇到跟你一样的问题

a-thomas:
Hi there, it seems that I've got the same problem with the last version of the `master` branch.

```
[ 58%] Generating WAST multi_index_test.wast
cd /Users/aleks/Dev/eos/eos/build/contracts/multi_index_test && /usr/local/binaryen/bin/s2wasm -o /Users/aleks/Dev/eos/eos/build/contracts/multi_index_test/multi_index_test.wast -s 4096 multi_index_test.s
/bin/sh: /usr/local/binaryen/bin/s2wasm: No such file or directory
[...]

PR #1485, opened 2018-02-27
Fix failing MacOS build

yumenohosi:
This is already proposed by PR #871

> 3. deleting /usr/local/binaryen before copy, so that the script runs idempotently
> 4. link files under /usr/local/binaryen/bin to /usr/local/bin, instead of /usr/local (assuming that was a mistake)

However, this file contains the same error.

I found those who are experiencing the same problem in Issue #895.


wanderingbort:
deprecated by #1479

PR #1487, opened 2018-02-27
Remove many useless cmake installs

spoonincode:
There are a bunch of files that get installed (via something like make install) that make absolutely no sense. No one can do anything useful with the static library to the net_plugin, for example. Removing all of the obvious offenders. I left stuff that still doesn't feel appropriate to me like fc and appbase. Technically those can be used standalone outside of eosio. Save those for a separate PR one day since it could be more contentious.

### DIFF
--- a/libraries/abi_generator/CMakeLists.txt
+++ b/libraries/abi_generator/CMakeLists.txt
@@ -62,13 +62,3 @@ if (USE_PCH)
   set_target_properties(abi_generator PROPERTIES COTIRE_ADD_UNITY_BUILD FALSE)
   cotire(eos_utilities)
 endif(USE_PCH)
-
-install( TARGETS
-   abi_generator
-   RUNTIME DESTINATION bin
-   LIBRARY DESTINATION lib
-   ARCHIVE DESTINATION lib
-)
-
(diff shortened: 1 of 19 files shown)


libraries/abi_generator/CMakeLists.txt
```diff
@@ -62,13 +62,3 @@ if (USE_PCH)
   set_target_properties(abi_generator PROPERTIES COTIRE_ADD_UNITY_BUILD FALSE)
   cotire(eos_utilities)
 endif(USE_PCH)
-
-install( TARGETS
-   abi_generator
-   RUNTIME DESTINATION bin
```
wanderingbort: Granted the ABI generator may not be operational right now, but I thought the point of it was to be a utility not unlike `eosiocpp` that could generate abi file from properly marked up contract code for the end-users.

heifner: This is the lib, not the abi_gen exec.

spoonincode: This is the abi_generator static library; eosio-abigen is the executable that users would use (that links to this static library).

wanderingbort: touche

PR #1488, opened 2018-02-27
fixes for multi-node testing with the test.system contract

wanderingbort:
* authentication keys were defaulting to the first producer key on a node if none was explicitly set however, the method of deriving this required that producer to be "active" creating a race condition for starting up a candidate producer node
 * `get_scheduled_producer` was not taking `config::producer_repititions` into account properly with its math, it is likely that we never had all of the producers working before
 * the ABI for `test.system` was not properly serializable for `setprods`

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2478,9 +2478,9 @@ namespace eosio {
    chain::public_key_type net_plugin_impl::get_authentication_key() const {
       if(!private_keys.empty())
          return private_keys.begin()->first;
-      producer_plugin* pp = app().find_plugin<producer_plugin>();
+      /*producer_plugin* pp = app().find_plugin<producer_plugin>();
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2478,9 +2478,9 @@ namespace eosio {
    chain::public_key_type net_plugin_impl::get_authentication_key() const {
       if(!private_keys.empty())
          return private_keys.begin()->first;
-      producer_plugin* pp = app().find_plugin<producer_plugin>();
+      /*producer_plugin* pp = app().find_plugin<producer_plugin>();
       if(pp != nullptr && pp->get_state() == abstract_plugin::started)
-         return pp->first_producer_public_key();
+         return pp->first_producer_public_key();*/
```
heifner: Seems to me we should make `peer-private-key` required so that private_keys is never empty. I think you are fine to remove this code for now.

PR #1490, opened 2018-02-27
Bypass all prompts for Jenkins

ericiles:


wanderingbort:
as this only touches the Jenkins file I'm going to put it through however, it would be nice to document somewhere which parts of the script were requiring user input so that @pacificcode can add a proper `-y` (apt lingo) style argument that automatically agrees to any user prompts with the default values.

pacificcode:
FYI - the package manager -y flag is set in the build scripts, however, the build scripts themselves prompt the user with a list of "NOT FOUND" required dependencies and requires the user to enter 0 or 1 to proceed with the installation.

PR #1497, opened 2018-02-28
Fixed issues with the Fedora build

ericiles:
Fedora builds were failing whenever there were package updates via yum (due to waiting for user input), and also due to LLVM not being compiled with RTTI enabled. With these changes, builds have been successful this morning.

### DIFF
--- a/scripts/eosio_build_fedora.sh
+++ b/scripts/eosio_build_fedora.sh
@@ -45,7 +45,7 @@
 	
 	printf "\tYum installation found at ${YUM}.\n"
 	printf "\tUpdating YUM.\n"
-	UPDATE=$( sudo yum update )
+	UPDATE=$( sudo yum -y update )
 	
 	if [ $? -ne 0 ]; then
 		printf "\n\tYUM update failed.\n"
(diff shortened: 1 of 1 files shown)


pacificcode:
This looks good.

scripts/eosio_build_fedora.sh
```diff
@@ -158,7 +158,7 @@
 		cd ..
 		mkdir build
 		cd build
-		cmake -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=${HOME}/opt/wasm -DLLVM_TARGETS_TO_BUILD= -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly -DCMAKE_BUILD_TYPE=Release ../
+		cmake -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX=${HOME}/opt/wasm -DLLVM_ENABLE_RTTI=1 -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly -DCMAKE_BUILD_TYPE=Release ../
```
wanderingbort: This seems to imply that we are using this LLVM install for 2 different purposes which we try not to.  Either it is being used to build something or it is being linked against for our wasm-jit library.  Lets figure out if there is a solution that has better isolation like our other platforms (hopefully) do.

wanderingbort:
Ok after talking with @spoonincode and @pacificcode etc, I think we need to document the need for an environment variable for `LLVM_DIR` as part of the fedora build that links together our web assembly compiler and the libraries we link to for JIT/ABI generation

# BIG NOTE

This link is something we want to remove in the future.   We should be able to upgrade the LLVM version for our WebAssembly toolchain and the LLVM version that backs our runtime JIT/tools independently.

PR #1502, opened 2018-02-28
Installation and startup cleanup

jgiszczak:
Support independent data and config directories in eosiod.
Start successfully on first launch whether installed or just built.
Honor EOSIO_ROOT environment variable in eosiod.
Regularize whitespace in eosiod main.cpp.
Add support for CMake v3.6 packaging features, and emulate in older versions.
Update skeleton contract if it changes, and install it in an eosio directory.
Add an eosio user and group in .deb package and set appropriate permissions.
Regularize whitespace in various CMakeLists.txt.
Add libmongoc dependency to Debian package if built with mongo support.
Add self-registration to template plugin.
Remove last reference to steemd in CMakeLists.txt.


spoonincode:
I think there is some abuse of cmake & packaging standards in our repo and this PR seems to dig us deeper.

For the installs, I think the generally it's recommended to do something like:
```
install( TARGETS
    eosio-abigen
   RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
 )
```
There are other defines like ${CMAKE_INSTALL_INCLUDEDIR} and ${CMAKE_INSTALL_SYSCONFDIR} (for /etc) that should be used.

If we insist on wanting to install to /usr instead of the default of /usr/local, then setting CMAKE_PREFIX_PATH to /usr is what we should do but this still seems like bad form as usually this is expected to be set by the user doing packaging/configurating.

If we want the build script to install to build/install then it should be setting DESTDIR=$PWD/install. But, I'm still skeptical this is what users expect or want. Think of any of other package -- it installs to system paths my default.

What we're doing now is actively hostile to packagers and anyone who wants to download the source, build it, and install it like any other package.

jgiszczak:
Withdrawing for rework.

PR #1504, opened 2018-02-28
EOSIO build amazon, RTTI support

pacificcode:
added LLVM_DIR variable to eosio_build.sh, added RTTI support to LLVM build.


wanderingbort:
Insert same caveat here as #1503

PR #1508, opened 2018-03-01
DAWN 646 - Fix problem where validation fails when creating new permission

andriantolie:
Inside the current updateauth action handler, the code tries to retrieve existing permission inside database using db.get function. Since the given permission name doesn't exist yet in the blockchain, this function which will always throw an error. This causes user to be unable to create new permission using updateauth action

### DIFF
--- a/libraries/chain/contracts/eosio_contract.cpp
+++ b/libraries/chain/contracts/eosio_contract.cpp
@@ -181,14 +181,15 @@ void apply_eosio_updateauth(apply_context& context) {
       if (act_auth.permission == config::owner_name || act_auth.permission == update.permission) {
          return true;
       }
+      const permission_object *current = db.find<permission_object, by_owner>(boost::make_tuple(update.account, update.permission));
+      // Permission doesn't exist yet
+      if (current == nullptr) return true;
 
-      auto current = db.get<permission_object, by_owner>(boost::make_tuple(update.account, update.permission));
-      while(current.name != config::owner_name) {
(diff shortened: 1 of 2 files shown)


libraries/chain/contracts/eosio_contract.cpp
```diff
@@ -181,14 +181,15 @@ void apply_eosio_updateauth(apply_context& context) {
       if (act_auth.permission == config::owner_name || act_auth.permission == update.permission) {
          return true;
       }
+      const permission_object *current = db.find<permission_object, by_owner>(boost::make_tuple(update.account, update.permission));
+      // Permission doesn't exist yet
+      if (current == nullptr) return true;
```
wanderingbort: I think this has the unintended side effect of allowing _any permission_ the ability to create a permission.  Even if that permission is linked into the tree _above_ it.  For instance, if we have:
* active
  * foo

we can authorize @foo to create bar whose _parent_ is active:

* active
  * foo
  * bar

This shouldn't be possible, but we are earlying out with `true` when the permission doesn't exist.

I think instead, we should handle the case where the permission doesn't exist by setting `current` to the parent-to-be and then processing the loop completely.  

This would create a situation where a permission level can only create permissions that attach to the tree of permissions below it.

andriantolie: Valid point.

PR Updated. Now, when creating a new permission, it will check whether the parent permission exists and has permission equal or lower to the carried auth.

PR #1515, opened 2018-03-01
Fixes for Docker builds

ericiles:


Docker/config.ini
```diff
@@ -1,15 +1,29 @@
...
 # the location of the block log (absolute path or relative to application data dir)
 block-log-dir = "blocks"
 
 # Pairs of [BLOCK_NUM,BLOCK_ID] that should be enforced as checkpoints.
 # checkpoint =
 
-# open the database in read only mode
-readonly = 0
+# Limits the maximum time (in milliseconds) that is allowed a transaction's code to execute from a received block.
+rcvd-block-trans-execution-time = 12
```
spoonincode: Where did these come from? I can't find `rcvd-block-trans-execution-time` in the code. (or am I just having a case of the mondays 🤔 )

ericiles: I had the config.ini provided to me from another developer. It looks like it may have been carried over from PR #540, which doesn't look like it made it into Dawn 3.x. I am going to remove it, and test further, and if all still works fine, I will update the PR.

PR #1521, opened 2018-03-02
Add support for mongo-cxx-driver stable release 3.2

cj-oci:
mongo-cxx-driver 3.2 modified cmake variable semantics. This change will extend support while maintaining backwards compatibility.


plugins/mongo_db_plugin/CMakeLists.txt
```diff
@@ -21,14 +21,30 @@ if(BUILD_MONGO_DB_PLUGIN)
...
       message(STATUS "Found bsoncxx library: ${EOS_LIBBSONCXX}")
 
       find_package(libmongocxx REQUIRED)
       message(STATUS "Found mongocxx headers: ${LIBMONGOCXX_INCLUDE_DIRS}")
-      find_library(EOS_LIBMONGOCXX ${LIBMONGOCXX_LIBRARIES}
-                   PATHS ${LIBMONGOCXX_LIBRARY_DIRS} NO_DEFAULT_PATH)
+
+      # mongo-cxx-driver 3.2 release altered LIBBSONCXX_LIBRARIES semantics. Instead of library names,
+      #  it now hold library paths.
+      if((LIBBSONCXX_VERSION_MAJOR LESS 3) OR ((LIBBSONCXX_VERSION_MAJOR EQUAL 3) AND (LIBBSONCXX_VERSION_MINOR LESS 2)))
```
spoonincode: shouldn't these be `LIBMONGOCXX_VERSION_MAJOR` etc instead?

cj-oci: Yes, For whatever reason previously they were undefined in my env. They are properly defined now, so I have updated my change.

Issue #1529, opened 2018-03-03
Error compiling contract in current master (MacOS) 


kesar:
```
➜  kesarCoin eosiocpp -o kesarCoin.wast kesarCoin.cpp
➜  kesarCoin eosioc set contract kesarcoin kesarCoin.wast kesarCoin.abi
Reading WAST...
Assembling WASM...
Publishing contract...
2980137ms thread-0   main.cpp:941                  main                 ] Failed with error: Assert Exception (10)
!"unresolvable": env._ZNKSt3__120__vector_base_commonILb1EE20__throw_length_errorEv
``` 

code:
```
#include <eosiolib/eosio.hpp>
#include <eosiolib/token.hpp>
#include <eosiolib/generic_currency.hpp>

typedef eosio::generic_currency<eosio::token<N(kesarcoin), S(8,KSR)>> kesarcoin;

extern "C" {
    void apply(uint64_t code, uint64_t action) {
        kesarcoin::apply(code, action);
    }
}
```

Abi:

```
{
  "types": [{
      "new_type_name": "account_name",
      "type": "name"
    }
  ],
  "structs": [{
      "name": "transfer",
      "base": "",
      "fields": [
        {"name":"from", "type":"account_name"},
        {"name":"to", "type":"account_name"},
        {"name":"quantity", "type":"asset"},
        {"name":"memo", "type":"string"}
      ]
    },{
     "name": "issue",
     "base": "",
     "fields": [
        {"name":"to", "type":"account_name"},
        {"name":"quantity", "type":"asset"}
     ]
  },{
      "name": "account",
      "base": "",
      "fields": [
        {"name":"currency", "type":"uint64"},
        {"name":"balance", "type":"uint64"}
      ]
    },{
      "name": "currency_stats",
      "base": "",
      "fields": [
        {"name":"currency", "type":"uint64"},
        {"name":"supply", "type":"uint64"}
      ]
    }
  ],
  "actions": [{
      "name": "transfer",
      "type": "transfer"
    },{
      "name": "issue",
      "type": "issue"
    }
  ],
  "tables": [{
      "name": "account",
      "type": "account",
      "index_type": "i64",
      "key_names" : ["currency"],
      "key_types" : ["uint64"]
    },{
      "name": "stat",
      "type": "currency_stats",
      "index_type": "i64",
      "key_names" : ["currency"],
      "key_types" : ["uint64"]
    }
  ]
}
```

Wast generated:

```
(module
[...]

PR #1534, opened 2018-03-04
fix eosiocpp so external contracts can be built

spoonincode:
Install the libraries and link against them like done in internal eosio cmakes


kesar:
fixes https://github.com/EOSIO/eos/issues/1529

PR #1538, opened 2018-03-05
Improve eosioc error message for get, set, wallet, push subcommand

andriantolie:
Fourth PR for DAWN-513

Improve eosioc error message for get, set, net, and wallet subcommand. After this, only push subcommand left.

Update:
As 16 March, this PR is updated to include error message improvement for push subcommand. This should be the last PR for DAWN-513

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -18,6 +18,7 @@
 #include <eosio/chain/scope_sequence_object.hpp>
 #include <eosio/chain/merkle.hpp>
 
+#include <eosio/chain/exceptions.hpp>
 #include <eosio/chain/wasm_interface.hpp>
 
 #include <eosio/utilities/rand.hpp>
@@ -248,13 +249,13 @@ bool chain_controller::_push_block(const signed_block& new_block)
  * queues.
  */
 transaction_trace chain_controller::push_transaction(const packed_transaction& trx, uint32_t skip)
-{ try {
-   return with_skip_flags(skip, [&]() {
-      return _db.with_write_lock([&]() {
-         return _push_transaction(trx);
-      });
-   });
-} EOS_CAPTURE_AND_RETHROW( tx_apply_exception, "Unable to apply transaction") }
+   { try {
+         return with_skip_flags(skip, [&]() {
+            return _db.with_write_lock([&]() {
+               return _push_transaction(trx);
+            });
+         });
+      } EOS_CAPTURE_AND_RETHROW( transaction_exception ) }
 
 transaction_trace chain_controller::_push_transaction(const packed_transaction& trx) 
 { try {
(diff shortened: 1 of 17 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -280,11 +280,14 @@ read_only::get_info_results read_only::get_info(const read_only::get_info_params
 }
 
 abi_def get_abi( const chain_controller& db, const name& account ) {
-   const auto& d = db.get_database();
-   const auto& code_accnt  = d.get<account_object,by_name>( account );
-
+   const auto &d = db.get_database();
+   const account_object *code_accnt = nullptr;
+   try {
+      code_accnt = &d.get<account_object, by_name>(account);
```
bytemaster: get returns a reference, therefore it can never be null. It will throw an exception if not found.  Perhaps you want d.find<...

andriantolie: Right. Thanks for pointing it out! PR Updated.

andriantolie:
PR Updated.

Now, this PR also includes error message improvement for eosioc push subcommand.

Furthermore, the RPC error structure is redefined to be
```
{
    "code": 400,
    "message": "Bad Request",
    "error": {
        "code": 3030025,
        "name": "tx_apply_exception",
        "what": "Transaction Apply Exception",
        "details": [
            {
                "message": "Unable to apply transaction (unknown key)",
                "file": "chain_controller.cpp",
                "line_number": 256,
                "method": "push_transaction"
            }
        ]
    }
}
```
from the previous
```
{
    "code": 400,
    "message": "Bad Request",
    "error": {
        "code": 3030025,
        "name": "tx_apply_exception",
        "message": "Transaction Apply Exception",
        "detail": "Unable to apply transaction (unknown key)"
        "stack_trace": [
            {
                "file": "chain_controller.cpp",
                "line_number": 256,
                "method": "push_transaction",
                ....
            }, ....
        ]
    }
}
```
To support multiple error message

andriantolie:
Merge conflict solved

libraries/chain/chain_controller.cpp
```diff
@@ -248,13 +249,13 @@ bool chain_controller::_push_block(const signed_block& new_block)
...
-      });
-   });
-} FC_CAPTURE_AND_RETHROW() }
+   { try {
+         return with_skip_flags(skip, [&]() {
+            return _db.with_write_lock([&]() {
+               return _push_transaction(trx);
+            });
+         });
+      } EOS_CAPTURE_AND_RETHROW( transaction_exception ) }
```
heifner: `FC_CAPTURE_AND_RETHROW` will catch `std::exception` and `...` and turn them into `fc::exception`. `EOS_CAPTURE_AND_RETHROW` only catches `fc::exception`. I think we should enhance `EOS_CAPTURE_AND_RETHROW` to also catch  `std::exception` and `...` and turn them into `fc::exception`.

heifner: ... which is exactly what you did. That is what I get for writing a comment before fully reviewing.

andriantolie:
Remove needs testing label from here and add it to the respected issue instead (#1860)

PR #1555, opened 2018-03-05
Add rend support in C DB API and reverse iteration in multi_index

arhag:
`multi_index` now makes it possible for contract developers to easily iterate through the objects in a table in reverse order of primary key or any of the secondary keys (using the `rbegin()` and `rend()` methods familiar from the C++ standard library containers).


arhag:
Turns out `rend()` is actually unnecessary in the C DB API. Convenient reverse iteration is instead implemented using C++ standard library's `reverse_iterator` in pull request #1566, which makes this pull request obsolete.

PR #1558, opened 2018-03-06
Fix docker-compose and documentation

jeremi:
The docker-compose config file was out of date for Dawn-2.0
I updated it, and updated the related documentation.


ericiles:
The dawn-2.x-docker-fix branch incorrectly renamed all the binaries from eos to eosio. Those changes were rejected, and the branch has reverted back to the original naming. As such, the changes in this PR are no longer applicable.

Issue #1567, opened 2018-03-06
DAWN-658 ⁃ Send on-block message that contains block header to system contract


blockone-syncclient:

bytemaster:
This is a dup #1758

PR #1568, opened 2018-03-06
GH-#1567 ⁃ Send on-block message that contains block header to system contract

zorba80:


zorba80:
I will close this pr. It's now part of pr #1699.

PR #1570, opened 2018-03-06
Added softfloat dependency and interface header

larryk85:
Added the softfloat library dependency to build scripts and a interface header file for it.  This is for arhag to start his work, this is also in preparation for softfloat injection.


spoonincode:
Should we be considering making this a submodule instead?

bytemaster:
Because Bucky had to customize softfloat (build files) we can include a copy in our repo.  We can refactor it into our own fork of softfloat later.

PR #1576, opened 2018-03-07
DAWN-639,DAWN-339,DAWN-380

dhaneshvb:
**Changes in DAWN-639:**
A new contract named testnet.sys has been introduced which is cloned fro test.sys and added currency transfer/issue support.

**Changes in DAWN-339:**
DAWN-380,383 are subtasks of it and these testcases porting doesnt require drastic changes


dhaneshvb:
no longer valid

PR #1586, opened 2018-03-07
Added cmake defaults for dependency root directories

larryk85:
Added these defaults so that the user can simply use 'cmake .' to build.


spoonincode:
Why not just load up the PATHS argument in the find_library calls with these paths?

larryk85:
We could do that, but for some of these the variables are used in areas other than just find_library.  So, we would still need to default the variables.

PR #1604, opened 2018-03-09
DAWN-551 Contract Debugging add support for native contract debugging

PaulCalabrese:
Allowing native contract debugging involved the following changes
* Development of debug_plugin that will enable debugging for natively-compiled contracts
* Addition of DEBUG_FLAG to the add_wast_executable so cmake can support compiling contracts into a shared library
* Changes to eosiolib to support compilation of code in both wasm and native environments. Most of these changes are controlled via preprocessor macro (EOSIO_NATIVE_CONTRACT_COMPILATION), but some changed the code for wasm contracts as well. The most visible of these was changing the typedef "time" to "eosio_time" to avoid collision with a posix-defined symbol.
* Adding a set of "native" intrinsic macros to libraries/chain/wasm_interface.cpp that allows for intrinsic calls to be made from natively-compiled contracts
* Documentation on how to debug in eos/plugins/debug_plugin/README.md


bytemaster:
Wow, this is amazing!  What can we do get get the sharedlib to use non-standard malloc/free so we can reset the heap between runs?  

Can we configure the library to use softfloat rather than real floats?

This is great work.

libraries/chain/wasm_interface.cpp
```diff
@@ -1615,5 +1621,265 @@ REGISTER_INTRINSICS(db_index_api_key64x64x64_value_index_by_scope_primary,   DB_
 REGISTER_INTRINSICS(db_index_api_key64x64x64_value_index_by_scope_secondary, DB_INDEX_METHOD_SEQ(secondary_i64i64i64));
 REGISTER_INTRINSICS(db_index_api_key64x64x64_value_index_by_scope_tertiary,  DB_INDEX_METHOD_SEQ(tertiary_i64i64i64));
 
+extern "C" {
```
wanderingbort: Is there a technical limitation that prevents this from using the same interface which registers intrinsics to the two non-debug backends?

Duplicating all of the registrations seems fragile and likely to end up with the debugger interface and the real interface not being in-sync

PaulCalabrese: I was looking at this for a while and you told me to hold off because of all the other work going on in this area. I could take another crack at it if things are more stable.

PaulCalabrese: What can we do get get the sharedlib to use non-standard malloc/free so we can reset the heap between runs?

I haven't looked too deeply into this, but it seems challenging in the existing approach because of the contract living in the same process space as the rest of the chain. 

Can we configure the library to use softfloat rather than real floats?

This is a follow-on task that Bart has mentioned, but I haven't looked into it yet.

spoonincode: >challenging in the existing approach because of the contract living in the same process space as the rest of the chain
Yes, this is why I was advocating for a sandboxed process. All these issues become a lot easier in that environment.

If you want to reset memory and "reset the heap" in the current design, one way of achieving this is through a hidden malloc/free in the .so & closing the .so after each contract invocation. closing/reopening the .so will cause all its initializers to refire so if someone has a static variable it will be reset. This would also allow you to reinitialize a heap that is visible only to the .so

PaulCalabrese:
I looked at dlclose() and it was my understanding that there was no
guarantee about static destructors/constructors running when
closing/reloading a shared library. I did not do any testing on our
platforms to see what the actual behavior was, but I am hesitant to depend
on non-standard behavior.

Issue #1489, opened 2018-02-27
DAWN-637 ⁃ Need to remove the remaining code for lazy table key validation


blockone-syncclient:
Need to remove the remaining code for lazy table key validation in contract_table_object.hpp (table_id_object::key_type).

ATC:
Run ctest and verify that all tests pass

brianjohnson5972:
Is this being assigned to Andrianto to validate it?

andriantolie:
I'm kinda confused with this ticket. I saw there are 3 PR referring to this ticket, will there any more in the future? Or is it ready for testing? If it's ready for testing, can we set the pipeline to be "In QA/ UAT" and add some ATC here? Thanks

andriantolie:
ATC Passes

PR #1639, opened 2018-03-13
DAWN-637 Remove Old Table Interfaces

brianjohnson5972:
#1489 
Removed commented out tests and un-compiled tests.
Removed un-used and un-compiled contracts.
Removed string key test for old table interface, since it will not be helpful for writing the new test when string keys are added to the new interface.
Removed db.hpp, table.hpp, and removed old interface declarations in db.h.

(Reviewer may want to use commits to review, since there is so much removed.)


brianjohnson5972:
#1650 moves tests/tests/block_tests.cpp, so one of these will have a merge conflict. probably want to merge the other one first and I'll fix the conflict.

PR #1641, opened 2018-03-13
Dawn 610 Context Free Actions

### LINKED ISSUE #1387: DAWN-610 ⁃ Context Free Actions
# Context Free Actions
Some computations necessary to build consensus are expensive from either a computational or block space perspective, or both.  For example, the proof for inter-blockchain communication (IBC) is large, and validating it can only pass or fail.  Once a block is irreversible, this code path and data is useless (unless someone is performing a verbose revalidation of the chain).

Witness Actions provide a means to perform intense checks in a way that the computation is cheaper during production/validation and both computation and block space are prunable.

This is conceptually equivalent to Segregated Witness for the EOS.IO blockchain.

### New Action Type
Witness Actions are a new type of action which may only read its own payload and a new section of transaction data referred to as "witness data".   They are otherwise identical to normal actions.

In the example of IBC this action's payload would be the provably retired action from the foreign chain and a block ID from the foreign chain whose state includes the transaction.

In an example of a document publication platform (PUB) this action's payload would be the hash of a document to publish.

### Witness Data
Transactions will contain a new section of data that provides witness data.  Witness data is anything needed to process a witness action that _doesn not_ need to persist if you can assume the action processed successfully.   This is a broader definition that most cryptographic witness data would include.

In the example of IBC, witness data would be the merkle proof that the retired action was present in a particular block on the foreign chain and the merkle proof that said block is included in the chain represented by the block ID listed in the action payload

In an example of PUB the witness data would be the plaintext document.

### Witness Data Access API
Witness actions can only read their payload and the associated witness data.  They cannot write anything.  They merely succeed in whole or fail.  This makes them stateless or context-free and allows trivial parallel processing.

In the example of IBC, the WebAssembly code will use this API and the action payload to `assert` that the proof is correct.

In an example of PUB, the WebAssembly code will use this API and the action payload to `assert` that the hash of the document matches the hash in the payload

### Sibling Action API
Non-witness actions will have access to a new API which allows them to `assert` that the transaction in which a current action is processing also includes another action.  Due to our processing model, the current action can then assume that the sibling action was also successful as any failure in any action is a failure of the transaction as a whole.

In the example of IBC, the normal action will assert that a sibling action existed routed to the proper action name, with a properly formed payload (the transaction and block ID of the foreign chain).  It can then proceed to do stateful processing based on the assumption that this sibling action succeeded and therefore the given action _was_ included in the foreign chain

In the example of PUB, the normal action will assert that a sibling action existed with the given hash as a payload and proceed to mark the document as published based on the assumption that this sibling action succeeded and therefore the given hash IS the hash of the desired document. 

### Resource Accounting
As witness action processing will happen substantially in parallel, it becomes harder to account for in resource controls.  However, as it depends on no state data it should also be possible to compute, ahead of time and deterministically, what resources this will consume.

A transaction will commit to this resource usage, the same way it will commit to packed size as part of #1386 

Note, this commitment will only cover the resources spent on retiring witness actions.   All normal actions are not deterministically billable ahead-of-time as they depend on state that is considered volatile until the precise moment a transaction is retired.   If a transaction fails, the resources billed to it ahead of time for witness data will be refunded. 

### Prunable Data
As part of this, the block structure (on disk and wire) will change such that all witness data and signatures are part of an explicit and well-contained section.  This section can be truncated off of blocks when they are considered irreversible and will remain replay/re-indexable as irreversibility proves that all dependent processing "passes" and can therefore be implied

heifner:
* Tests for context free actions
* Fixes for context free actions
* Additional APIs made available to context free actions
** checktime_api
** crypto_api
** system_api
** action_api
** console_api
** math_api
* nonce action changed to be a context free action in eosioc
* read_action changed to read_action_data to distinguish between action and action_data
* Also includes some of DAWN 617 work that was put on hold for this work
* Does not include pruning the context_free_data from the block log

Resolves #1387


heifner:
Closing this PR as another will be coming shortly with only CFA changes.

PR #1650, opened 2018-03-14
DAWN 379 - Restore block_tests

andriantolie:
Restore block_tests from 2.x and merge it (tests/tests/block_tests) with tests/chain_tests/block_tests


brianjohnson5972:
#1639 removes tests/tests/block_tests.cpp, so one of these will have a merge conflict. probably want to merge this one first.

PR #1653, opened 2018-03-14
Fix/softfloat submodule

larryk85:
Adding softfloat as a submodule for the project.


spoonincode:
I noticed on binaryen/musl/c++ we have an "eosio" branch and that's branch we intend to submodule in. I kind of feel like we should make all these submodules behave similarly. Is that easy to go back and do?

PR #1655, opened 2018-03-14
Added softfloat as a submodule from EOSIO fork

larryk85:
Added softfloat as a submodule.  Changed submodule to point to eosio fork of berkeley-softfloat-3 repo.


libraries/chain/include/eosio/chain/softfloat.hpp
```diff
@@ -1,186 +0,0 @@
-#pragma once
```
spoonincode: In your previous push, this file wasn't removed. Is this intended?

PR #1656, opened 2018-03-14
Building EOSIO Darwin with mongoDB plugin

pacificcode:
Added support for building EOSIO mongoDB plugin for darwin.
Added support for running make test after successful installation of EOSIO on darwin.
Added instructions for building MongoDB C++ driver in manual installation of dependencies for Darwin in README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -847,6 +847,7 @@ macOS additional Dependencies:
 
 * Brew
 * Newest XCode
+* MongoDB C++ driver
 
 Upgrade your XCode to the newest version:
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -864,9 +865,33 @@ Install the dependencies:
 
 ```bash
 brew update
-brew install git automake libtool boost openssl llvm@4 gmp ninja gettext
+brew install git automake libtool cmake boost openssl@1.0 llvm@4 gmp ninja gettext mongodb
```
heifner: Is it necessary to specify 1.0 of openssl?

pacificcode: Yes, there was an issue we ran into a while back that arose when building eosio-applesedemo against openssl 1.1. Matt can confirm.

spoonincode: applesedemo depends on it being 1.0 at the moment; can be fixed. But my original push back at the time was that it seemed we wanted a VERY explicit openssl version on Linux, MacOS, etc. That appears to no longer be the case since I think on all Linux OS we just link to system openssl anyways.

Do we know why we can't use system openssl on macos (technically libressl)? Is it strictly because our FindSSL cmake module can't find libressl? Or is there something known to be more problematic?

pacificcode: If I remember correctly, on darwin when compiling against openssl1.1 make would error out when building the fc module.

heifner: Can we give a try now and see how it goes?

spoonincode: I 100% guarantee you it will fail in applesedemo

pacificcode: agreed, I recommend we stay at 1.0 for now.

spoonincode: >Do we know why we can't use system openssl on macos (technically libressl)? Is it strictly because our FindSSL cmake module can't find libressl? Or is there something known to be more problematic?

I looked at this for a few minutes and it seems like there is no public openssl style interface for the included libressl. (there is no ssl.h and such)

PR #1657, opened 2018-03-14
Build script regex change to 1. extract decimal part and 2. global search.

jonnyhsy:
Example: df -h  /
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda1       9.7G  4.2G  5.5G  44% /
df -h / | grep /dev | tr -s ' ' | cut -d\  -f4 | sed 's/[^0-9]//'
55G
df -h / | grep /dev | tr -s ' ' | cut -d\  -f4 | sed 's/[^0-9.]//g'
5.5

### DIFF
--- a/scripts/eosio_build_amazon.sh
+++ b/scripts/eosio_build_amazon.sh
@@ -5,8 +5,8 @@
 	CPU_SPEED=$( lscpu | grep "MHz" | tr -s ' ' | cut -d\  -f3 | cut -d'.' -f1 )
 	CPU_CORE=$( lscpu | grep "^CPU(s)" | tr -s ' ' | cut -d\  -f2 )
 
-	DISK_TOTAL=`df -h / | grep /dev | tr -s ' ' | cut -d\  -f2 | sed 's/[^0-9]//'`
-	DISK_AVAIL=`df -h / | grep /dev | tr -s ' ' | cut -d\  -f4 | sed 's/[^0-9]//'`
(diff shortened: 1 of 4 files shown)


spoonincode:
I don't understand how allowing `.` works with the following comparison
`if [ $DISK_AVAIL -lt $DISK_MIN ]; then`

Because Bash doesn't support anything but integers.
```
$ if [ 5.5 -lt 8 ]; then echo it is less; fi
-bash: [: 5.5: integer expression expected
```

pacificcode:
You are correct, bash does not support float/double's and the comparison operator will fail. However, I did not take into consideration someone installing on a disk with that little space available. The correction should be:
DISK_AVAIL=9.7
$ if [ ${DISK_AVAIL%.*} -lt 10 ]; then echo it is less; fi
it is less

Ill add this fix as a commit to my latest PR #1656

jonnyhsy:
Ah you're right bash only recognize integers. I'm compiling on a vps machine with 8G ram and want to maintain as minimum disk as possible...  But the global replace flag 'g' is still needed, on my local laptop: 
 df -h /
Filesystem   Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk1  233Gi  172Gi   61Gi    74% 2622912 4292344367    0%   /

pacificcode:
This fix has been committed and is all set in PR #1656. Thank you jonnyhsy for bringing this to my attention.

PR #1664, opened 2018-03-15
DAWN-393 & DAWN-394 - Remove obsolete dawn 2.x test cases

andriantolie:
After discussed with Bart, we found out that table_abi_tests and slow_tests are obsolete and therefore they are not needed to be ported


brianjohnson5972:
All these changes are handled by #1639

heifner:
Closing as handled by #1639.

PR #1665, opened 2018-03-15
Fix for undef of N macro in abi_generator

heifner:


heifner:
Closing. Going with #1666 instead.

PR #1669, opened 2018-03-15
New injection and validation framework

larryk85:
A new unified framework for doing validations and injections at the module and instruction level.


libraries/wasm-jit/Include/IR/Operators.h
```diff
@@ -11,7 +11,8 @@ namespace IR
 
 	struct NoImm {};
 	struct MemoryImm {};
-
```
spoonincode: Can you clean up the whitespace differences in wasm-jit directory? These sort of things clutter diffs to upstream.

larryk85: Got it.

spoonincode:
One interesting thing in here is that I can't build with clang6&latest libc++. I hit these sort of errors
```
In file included from /home/spoon/eos-noon-net/libraries/wasm-jit/Source/Runtime/LLVMEmitIR.cpp:1:
In file included from /home/spoon/eos-noon-net/libraries/wasm-jit/Source/Runtime/LLVMJIT.h:35:
In file included from /home/spoon/llvm4libs/include/llvm/IR/IRBuilder.h:33:
/home/spoon/llvm4libs/include/llvm/IR/Instructions.h:4121:16: error: no template named 'pointer_to_unary_function' in namespace 'std'
  typedef std::pointer_to_unary_function<Value *, BasicBlock *> DerefFnTy;
```
This actually is expected because `pointer_to_unary_function` was _removed_ from c++17. Now, this should be fixable with proper cmake rules. Though I do wonder if we should do something else to avoid c++17 usage for now.

libraries/chain/include/eosio/chain/wasm_eosio_binary_ops.hpp
```diff
@@ -0,0 +1,699 @@
...
+template <typename Mutator>
+struct propagate_post_injection<Mutator> {
+   static constexpr bool value = Mutator::post;
+};
+
+template <bool Kills, bool Post>
+struct base_mutator {
+   static void accept( wasm_ops::instr* inst, wasm_ops::visitor_arg& arg ) {
+      std::vector<U8> self_code = inst->pack();
+      if constexpr ( Post )
```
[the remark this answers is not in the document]
wanderingbort: I ended up taking care of this

wanderingbort:
I do think we have some additional clean up to do on this task but they can be done after our next release as they are addressing future maintenance and tech debt.

For instance, I would prefer to convert some/all of the uses of BOOST_PP to type based template code.  I think overall it will be an equivalent amount of code and modern IDE/debuggers handle them better than the deep macros of BOOST_PP.  (this is my opinion but I had to give CLion an extra GB of ram to edit the main header file for this PR).

CMakeLists.txt
```diff
@@ -148,11 +148,11 @@ else( WIN32 ) # Apple AND Linux
         # Apple Specific Options Here
         message( STATUS "Configuring Eos on OS X" )
 #set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=c++14 -Wall -Wno-invalid-partial-specialization -Wno-deprecated-declarations" )
-        set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=c++14 -Wall -Wno-deprecated-declarations" )
+        set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=gnu++14 -Wall -Wno-deprecated-declarations" )
```
spoonincode: Cleaner way of doing this.. hold on for new commit to review

Issue #1671, opened 2018-03-15
Add Producer Scheduler Version to Block Header & Validate It


bytemaster:
Any time there is a new set of producers it is added to the block header along with an incrementing version number. This set of producers becomes effective once the block becomes irreversible. 

Light clients need to know when the block becomes irreversible so they can switch to the new schedule, but the current block headers requires light clients to process all headers (an indeterminate number) to find out when to apply the change.  

A sparse light client needs to know when the active set changes in a more efficient manner.  To do this each block header will include the producer schedule version that applies for a given block. Full nodes will validate this number, light nodes will trust this number.  It should only ever increment.

andriantolie:
Related PR is #1672

andriantolie:
ATC:
Add more unit tests to producer_schedule_tests where it verifies:
- Producer schedule version is updated (incrementally) inside the block header whenever a new set of producer schedule is applied
And ensure the unit test pass

PR #1672, opened 2018-03-15
Add Producer Scheduler Version to Block Header & Validate It #1671

bytemaster:


bytemaster:
You are right, I referenced the wrong field...

PR #1686, opened 2018-03-17
Change message to action

SheldonHH:
Tailor to Dawn 3.0x standard


elmato:
@SheldonHH I think those macros are not used anymore, they were replaced by eosio::testing::tester class.

heifner:
These testing macros are no longer used.

PR #1687, opened 2018-03-17
Correct CMAKE_BUILD_TYPE in eosio_build.sh

spoonincode:
Change RelWithDebugInfo to the correct RelWithDebInfo

(otherwise our builds via the script won't even have optimization 😳)


eosio_build.sh
```diff
@@ -133,7 +133,7 @@
 
 	COMPILE_EOS=1
 	COMPILE_CONTRACTS=1
-	CMAKE_BUILD_TYPE=RelWithDebugInfo
+	CMAKE_BUILD_TYPE=RelWithDebInfo
```
SheldonHH: `RelwithDebInfo` optimizes the code similarly to the behavior of Release mode. It produces fully optimized code, but also creates the symbol table and the debug metadata to give the debugger as much input as it is possible to map the execution back to the original code at any time.

PR #1690, opened 2018-03-17
Dawn 664 bios boot

pmesnier:
This patch is largely about the launcher and updating it to introduce the notion of a single bios node responsible for kicking off a new blockchain and then delegating block productions o a secondary group.  This launcher will stand up the group, providing each with a unique key. It also prepares a new script, currently called  boo.sh, which follows Bart's original blockchain ignition sequence. Note that testnet 2 is currently running the dawn-664-bios-boot branch, which is a few days behind he head of master.

The remaining changes in this PR are somewhat tangential but useful. In net_plugin, the reason a  go_away message is sent is now reported when i is actually sent. The producer_plugin change avoids reporting duplicated not-producing messages, substantially reducing log spam.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -965,7 +965,7 @@ namespace eosio {
    }
 
    void connection::enqueue( const net_message &m, bool trigger_send ) {
-      bool close_after_send = false;
+      go_away_reason close_after_send = no_reason;
```
SheldonHH: `go_away_reason` defined as `enum` in protocol.hpp

pmesnier: That's right. Semantically the "close_after_send" value's role is unchanged. Passing the actual reason value gives the added benefit of being able to report why a connection is being actively closed.

pmesnier:
This PR is obsolete, replaced by #1720

PR #1691, opened 2018-03-17
Some versions of git do not permit --recurse-submodules on the checko…

gleehokie:
…ut command so this switches to adding a git submodule update --recursive command instead.  Also cleans up some whitespace to be more consistent.


README.md
```diff
@@ -114,11 +113,10 @@ Now you can proceed to the next step - [Creating and launching a single-node tes
...
 ```bash
 git clone https://github.com/eosio/eos --recursive
-
 cd eos
-
-git checkout DAWN-2018-02-14 --recurse-submodules
-./eosio_build.sh
+git checkout DAWN-2018-02-14
+git submodule update --recursive
+./build.sh
```
heifner: This should be `./eosio_build.sh` on master.

README.md
```diff
@@ -162,11 +159,10 @@ Then clone the EOS repository recursively, checkout the branch that is compatibl
...
```
heifner: This should be `./eosio_build.sh` on master.

README.md
heifner: Sorry, missed that this is public testnet.

README.md
heifner: Missed that this is public testnet.

PR #1699, opened 2018-03-19
New system contract first revision

moskvanaft:
Includes:
staking for bandwith, storage and voting
registering producers and proxies
voting
proxy voting
applying parameters elected by producers
onblock message
inflation and payment to producers
tests


arhag:
I reviewed this PR (as of commit 5bb8f01) and made some changes. Jenkins seems to be having issues in general right now, so I don't necessarily think it has to do with this PR being bad. It seems we have non-deterministic failures in our tests, because sometimes when I ran `ctest` on my local machine all tests (including `nodeos_run_test-mongodb`) passed, and other times `nodeos_run_test` and/or `nodeos_run_test-mongodb` failed.

PR #1714, opened 2018-03-20
Generates Ricardian Contracts from a .abi file

joneric:
@jgiszczak 

This script generates Ricardian Contracts from a .abi file.

### DIFF
--- a/scripts/abi_to_rc/README.md
+++ b/scripts/abi_to_rc/README.md
@@ -2,13 +2,13 @@
 The `abi_to_rc.py` script processes a contract's .abi file in order to generate an overview Ricardian Contract and a Ricardian Contract for each action. The overview Ricardian Contract provides a description of the contract's purpose and also specifies the contract's action(s), input(s), and input type(s). The action Ricardian Contract provides a description of the action's purpose and also specifies the action's input(s), and input type(s).
 
 ## How to run
-`python abi_to_rc.py /path/to/smart-contract.abi`
+`$ python abi_to_rc.py /path/to/smart-contract.abi`
 
 ## Example
-`python abi_to_rc.py ../../contracts/currency/currency.abi`
+`$ python abi_to_rc.py ../../contracts/currency/currency.abi`
 
 ## Results
-For the example above, if all goes well, `abi_to_rc.py` should generate output files that have the following names: `currency-rc.md`, `currency-transfer`, `currency-issue-rc.md`, `currency-create-rc.md`.
+For the example above, `abi_to_rc.py` should generate output files that have the following names: `currency-rc.md`, `currency-transfer-rc.md`, `currency-issue-rc.md`, `currency-create-rc.md`.
 
 ## Notes
 Be sure to have `abi_to_rc.py`, `rc-overview-template.md`, and `rc-action-template.md` in the same folder.
--- a/scripts/abi_to_rc/README.md
+++ b/scripts/abi_to_rc/README.md
@@ -2,13 +2,13 @@
 The `abi_to_rc.py` script processes a contract's .abi file in order to generate an overview Ricardian Contract and a Ricardian Contract for each action. The overview Ricardian Contract provides a description of the contract's purpose and also specifies the contract's action(s), input(s), and input type(s). The action Ricardian Contract provides a description of the action's purpose and also specifies the action's input(s), and input type(s).
 
 ## How to run
-`python abi_to_rc.py ${{smart-contract.abi}} rc-overview-template.md rc-action-template.md`
+`python abi_to_rc.py /path/to/smart-contract.abi`
 
 ## Example
-`python abi_to_rc.py currency.abi rc-overview-template.md rc-action-template.md`
+`python abi_to_rc.py ../../contracts/currency/currency.abi`
 
 ## Results
-For the example above, if all goes well, `abi_to_rc.py` should generate output files that have the following names: `currency-rc.md`, `currency-issue-rc.md`, `currency-transfer-rc.md`.
+For the example above, if all goes well, `abi_to_rc.py` should generate output files that have the following names: `currency-rc.md`, `currency-transfer`, `currency-issue-rc.md`, `currency-create-rc.md`.
 
 ## Notes
-`abi_to_rc.py` was written in Python 3. Be sure to have `abi_to_rc.py`, `currency.abi` (or whichever `.abi` file that is being used), `rc-overview-template.md`, and `rc-action-template.md` all in the same folder.
+Be sure to have `abi_to_rc.py`, `rc-overview-template.md`, and `rc-action-template.md` in the same folder.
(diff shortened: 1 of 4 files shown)


scripts/abi_to_rc/README.md
```diff
@@ -0,0 +1,14 @@
+# Purpose
+The `abi_to_rc.py` script processes a contract's .abi file in order to generate an overview Ricardian Contract and a Ricardian Contract for each action. The overview Ricardian Contract provides a description of the contract's purpose and also specifies the contract's action(s), input(s), and input type(s). The action Ricardian Contract provides a description of the action's purpose and also specifies the action's input(s), and input type(s).
+
+## How to run
+`python abi_to_rc.py ${{smart-contract.abi}} rc-overview-template.md rc-action-template.md`
```
jgiszczak: If you use the `argparse` module as mentioned below, you can write this as `python abi_to_rc.py -h`.  That way if the arguments change, there's no need to update this document, and no chance it goes stale.

scripts/abi_to_rc/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Example
+`python abi_to_rc.py currency.abi rc-overview-template.md rc-action-template.md`
+
+## Results
+For the example above, if all goes well, `abi_to_rc.py` should generate output files that have the following names: `currency-rc.md`, `currency-issue-rc.md`, `currency-transfer-rc.md`.
+
+## Notes
+`abi_to_rc.py` was written in Python 3. Be sure to have `abi_to_rc.py`, `currency.abi` (or whichever `.abi` file that is being used), `rc-overview-template.md`, and `rc-action-template.md` all in the same folder.
```
jgiszczak: Requiring the abi file to be in the same directory as the script is needlessly unfriendly.  In addition to `os.path.splitext()` described below, first use `dirname, abi_filename = os.path.split()`, passing the appropriate identifier from `argparse`.  Then you can write to `os.path.join(dirname, rc_file_name)`. (And catch `IOError` in case that destination isn't writable and provide an English description of the problem.)  See [os.path](https://docs.python.org/3/library/os.path.html#module-os.path) for all your path manipulation needs (specific function suggestions both here and below).

scripts/abi_to_rc/abi_to_rc.py
```diff
@@ -0,0 +1,157 @@
...
+        else:
+            temp_row = temp_row.replace("${{ type }}",action['type'])
+
+        table_rows.append(temp_row)
+    return table_rows
+
+# generates an overview ricardian contract from the overview template
+def generate_rc_overview_file():
+    tr = build_table_rows()
+    rc_overview_template_file_name = sys.argv[2]
```
jgiszczak: To avoid the documented problem of calling the script while outside the script's own directory and being unable to find the template, derive a path: `os.path.join(os.path.dirname(sys.argv[0]), _RC_OVERVIEW)`.  (You did make a 'constant' of the template filename, right?)

scripts/abi_to_rc/abi_to_rc.py
```diff
@@ -0,0 +1,157 @@
...
+            print("$ python abi_to_rc.py ${{smart-contract.abi}} rc-overview-template.md rc-action-template.md")
+            sys.exit(0)
+        elif not os.path.isfile(sys.argv[3]):
+            print("ERROR: %s could not be found" % sys.argv[3])
+            sys.exit(0)
+    else:
+        print("ERROR: too many arguments")
+        print("-- Expected format --")
+        print("$ python abi_to_rc.py ${{smart-contract.abi}} rc-overview-template.md rc-action-template.md")
+        sys.exit(0)
```
jgiszczak: Error exit values should always be non-zero.  That way different non-zero exit values can mean different, specific error conditions, which makes automated scripting easier.  All Unix command line tools follow this convention, including all of the test code in the repository.  CTest looks for non-zero results and treats them as errors, while passing all zero return codes.

scripts/abi_to_rc/abi_to_rc.py
```diff
@@ -0,0 +1,155 @@
...
+                'input': 'inputs' if len(inputs) > 1 else 'input',
+                'type': 'types' if len(types) > 1 else 'type'      }
+    abi_file_name = os.path.split(args.abi_file)[1]
+    contract_name = os.path.splitext(abi_file_name)[0]
+    rc_file_name = contract_name + '-rc.md'
+    dirname = os.path.split(args.abi_file)[0]
+    rc_file = open(os.path.join(dirname, rc_file_name),"w+")
+    with open(os.path.join(os.path.dirname(sys.argv[0]), _RC_OVERVIEW)) as fp:
+        overview_template = Template(fp.read())
+        rc_file.write(overview_template.substitute( contract=contract_name,
```
jgiszczak: Rather than use the **kwds form of `substitute()`, it would be considerably less verbose to simply pass a dictionary.  You nearly have the one you need already.  Rename `plural` to something like `subs` and insert the other four things into it. `'contract': contract_name` can just be added to the initial declaration.  Then add the titlecase versions with a one-liner: `subs.update([(k+'_header',v.title()) for k,v in subs.copy().items()])`.  You'll end up with an extra `contract_header` entry, but Template will politely ignore it for you.  Then you can just call `overview_template.substitute(subs)`.

Admittedly, reading Python list comprehensions takes some practice.  Upside down `for` loops aren't exactly what you expect, but you get used to it.

scripts/abi_to_rc/abi_to_rc.py
```diff
@@ -0,0 +1,155 @@
...
+    dirname = os.path.split(args.abi_file)[0]
+    for action in actions:
+        plural = {  'input': 'inputs' if len(inputs[action['name']]) > 1 else 'input',
+                    'type': 'types' if len(types[action['name']]) > 1 else 'type'      }
+        rc_action_file_name = contract_name + "-" + action['name'] + '-rc.md'
+        rc_file = open(os.path.join(dirname, rc_action_file_name),"w+")
+        with open(os.path.join(os.path.dirname(sys.argv[0]), _RC_ACTION)) as fp:
+            action_template = Template(fp.read())
+            rc_file.write(action_template.substitute(   action=action['name'],
+                                                        input=plural['input'],
```
jgiszczak: Same as above.

scripts/abi_to_rc/abi_to_rc.py
```diff
@@ -0,0 +1,155 @@
...
+## how to run abi_to_rc.py
+how_to =    '''\n \nhow to run:\n  $ python abi_to_rc.py /path/to/smart-contract.abi'''
+
+## example of how to run abi_to_rc.py
+example =   '''\n \nexample:\n  $ python abi_to_rc.py ../../contracts/currency/currency.abi'''
+
+# argument parser
+parser = argparse.ArgumentParser(
+    prog="abi_to_rc.py",
+    formatter_class=argparse.RawDescriptionHelpFormatter,
```
jgiszczak: I don't entirely approve of the usage of the RawDescriptionHelpFormatter.  If you `chmod +x abi_to_rc.py` it can be run the way the default formatter says, since you have a shebang at the beginning.  (And git will preserve the executable permission when people clone/pull.)  Long descriptions beg for automated line break formatting.  You can add the example to the `epilog` argument of ArgumentParser's constructor.  I don't insist, but try it and see.

PR #1717, opened 2018-03-21
fix master Dockerfile

jeremi:
`/tmp/build`  and `/eos` should not be removed as they are needed in the second part of the Dockerfile. 
Copying files in /opt/eosio is also not needed as done in the second part of the build as well.


jcalfee:
confirmed

andriantolie:
The related issue (#1839) is marked as pass already

PR #1750, opened 2018-03-22
Add new WASM constraint on function stack usage

spoonincode:
Add a new constraint for each WASM function limiting the amount of stack it can use. This is  intended to bound the stack usage per function in a deterministic manner.

8192 chosen via fair dice roll. Opinions welcome.

To deterministically limit the entire stack size, another change will be required to track call depth.

### DIFF
--- a/libraries/chain/wasm_eosio_validation.cpp
+++ b/libraries/chain/wasm_eosio_validation.cpp
@@ -59,4 +59,19 @@ void globals_validation_visitor::validate( const Module& m ) {
             ("k", wasm_constraints::maximum_mutable_globals));
 }
 
+void maximum_function_stack_visitor::validate( const IR::Module& m ) {
+   for(const FunctionDef& func : m.functions.defs) {
+      unsigned function_stack_usage = 0;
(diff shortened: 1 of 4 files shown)


libraries/chain/wasm_eosio_validation.cpp
```diff
@@ -59,4 +59,19 @@ void globals_validation_visitor::validate( const Module& m ) {
             ("k", wasm_constraints::maximum_mutable_globals));
 }
 
+void maximum_function_stack_visitor::validate( const IR::Module& m ) {
```
bytemaster: This looks like it will not calculate the runtime stack size, just the "static size of all functions".  Many of these functions may never be called.

spoonincode: Yep, that's correct. I was trying to keep the PRs small and separate because big complex ones are harder to review and tend to languish.

This PR just makes sure that no single function consumes > 8192 bytes of stack size. A follow on PR will then add a runtime stack depth check. At that point maximum runtime usage will be bounded by maximum_func_local_bytes*maximum_stack_depth

PR #1798, opened 2018-03-23
Dawn 694 boot seq unit tests

dhaneshvb:
This unit test covers all the boot sequence steps
Not fully tested as eosio.toke, msig contracts are not ready yet


dhaneshvb:
Conflicts

PR #1849, opened 2018-03-26
Resurrect wasm unit test with new “OOB” wasm unit tests

### LINKED ISSUE #1823: Need to fix & reenable the table OOB unit test
This was `#if 0`ed because of weird SEGV race conditions in chain_test. Needs to be resolved and reenabled.

spoonincode:
We have a unit test that is #if 0ed (and some additional unit tests incoming) that attempt to access out of bounds memory. WAVM handles this cleanly by trapping the SEGV and then we retire the transaction as a failure.

However in chain_test the Boost Test framework by default tries to grab SEGV for itself. So depending on the tests that execute WAVM will be unable to trap the SEGV and instead it will be bubbled up to Boost Test as a test failure.

I tried a few approaches to resolve this cleanly but long story short nothing great came about. I need to set --catch_system_errors=no for operation in WAVM but I could not figure out an appropriate way of easily doing this at runtime via code. So.. adding it to ctest command.

But, I can’t have those troublesome tests run when someone just simply runs ‘chain_test’. Originally I tried to dynamically add/remove the oob tests based on if --catch_system_errors=no was set and/or we were running WAVM or not. But unfortunately disabled _at runtime_ tests are considered “skipped” by Boost Test which actually results in a failure return code.

So instead require these OOB tests to be opted in to manually… Added in the ctest file.

Fixes #1823

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -44,8 +44,8 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/consensus-validation-malicious-produc
 
 #Manually run chain_test for all supported runtimes
 #To run chain_test with all log from blockchain displayed, put --verbose after --, i.e. chain_test -- --verbose
-add_test(NAME chain_test_binaryen COMMAND chain_test --report_level=detailed --color_output -- --binaryen)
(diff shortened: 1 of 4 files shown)


spoonincode:
This PR is in limbo because it appears WAVM on Linux is not cleanly catching the SEGV.

spoonincode:
Superseded by #2420

Issue #1844, opened 2018-03-26
DAWN-690 ⁃ Enable integration tests 


blockone-syncclient:
Many of the tests were disabled temporarily due to bios related limitations. With Phil's launcher changes it should be possible to re-enable these tests.

ATC:
Run following tests successfully:

- [ ] tests/p2p_tests/sync/test.sh
- [ ] tests/p2p_tests/sync/test.sh ~~p 2~~ d 10
- [ ] tests/p2p_tests/sync/test.sh -m ~~p 21~~ n 21 ~~d 5~~ l
- [ ] tests/trans_sync_across_mixed_cluster_test.sh ~~p 1~~ n 2
- [ ] tests/distributed-transactions-test.py ~~p 1~~ n 4
- [ ] tests/distributed-transactions-remote-test.py --dump-error-detail
- [ ] tests/restart-scenarios-test.py ~~c resync ~~p3~~ -~~dump-error-details
- [ ] tests/restart-scenarios-test.py ~~c replay ~~p3~~ -~~dump-error-details
- [ ] tests/consensus-validation-malicious-producers.py ~~w 80~~ ~~dump~~error-details
- [X] tests/nodeos_run_test.py
- [ ] tests/nodeos_run_remote_test.py
- [X] tests/nodeos_run_test.py --mongodb

heifner:
The following tests re-enabled via #1852 

tests/nodeos_run_test.py
tests/nodeos_run_test.py --mongodb

brianjohnson5972:
All of the tests are now enabled.

PR #1852, opened 2018-03-26
Enable nodeos_run_test.py and mongo version as these tests now pass.

heifner:
This does not fully complete #1844.
Enables existing tests that pass.


heifner:
Runs fine on my machine and Ciju's. Fails on Jenkins. Investigating.

Issue #1022, opened 2018-01-02
DAWN-398 ⁃ User configurable generic security delay


wanderingbort:
# Delayed Signed Transactions
As a user, I may be willing to trade immediate access to certain features for increased security in the event of an account compromise.  #922 outlines the procedures by which I, as the user, will respond to such a compromise and regain secure control over my account however, there is a period of time where an attacker may be able to sign certain transactions on my behalf.  Depending on my own values, it may be favorable to impose a delay on sensitive transactions on myself in order to ensure that an attacker cannot issue them without recourse. 

## Feature Description
Each named `permission` on an account is given a required `delay` which defaults to "no delay". 
Updates to a `permission` that affect this delay must have been delayed by the existing delay.

When processing a newly learned transaction AFTER completely validating all necessary `Authority` 
IF a `permission` which has a non-defaulted `delay`, is used on the transaction:

* The transaction will be "boxed" into a deferred transaction, to be delivered no earlier than the current block time + maximum requested delay.
* Storage for this will count as usage from the account requesting the maximum delay, breaking ties based on explicit ordering in the `actions[]::authorizations[]` (first mentioned).
* The transaction's receipt in the block will have a new status code `delayed`
* The deferred transaction can be cancelled by sending a signed action to `config::system_account_name` with an action name of `canceldelay` and a payload of the transactions ID (BDJ -- assuming this should be sender_id.
  * this action only succeeds if there exists a transaction with that ID that is delayed and the `canceldelay` action has the `active` permission authorization of one of accounts mentioned in the original transaction's  in `actions[]::authorizations[]` 
* When/if the delayed transaction is finally included in a block it will carry the normal `excuted` status flag, there should be no contract-based error handling required.

## Example use cases
Alice has a "savings" account and wants to prevent the loss of any EOS, in the event of an account compromise assuming she notices the compromise within 30 days and is willing to delay withdraws from that account to guarantee this.   She sets up the sub-account with a permission named `withdraw` authorized by her primary account `alice@active` and having a delay of 30 days requested.   She uses `linkauth` to set the permission for any `eosio::transfer` action to `withdraw` (meaning that any transfer authorized by the savings account now has a 30 day delay).   She transfers her secure funds to the sub-account. 

[...]

PR #1859, opened 2018-03-26
Delayed Transactions

brianjohnson5972:
DAWN-398
#1022 
Added processing to box and delay transactions with actions that are either:
Linked to permissions with a non-zero delay
updateauth for a permission that has a delay
linkauth or unlinkauth to a previous permission that has a delay

Also added processing for canceldelay and mindelay actions.


brianjohnson5972:
Resolved conflicts with master.

Issue #1742, opened 2018-03-22
cleos get table always return no data


jeremi:
Using the latest version of `master` branch, I'm trying the currency  smart contract.
I'm able to:
- Create new currency
- Issue currency
- Transfer currency
- Cannot transfer more currency than what I have
So the currency contract seems to work as intended

But I'm not able to view table data. `cleos get table currency currency account` always return an empty array.

I attached the log to show how to reproduce the issue.
[log.txt](https://github.com/EOSIO/eos/files/1836780/-.txt)

jeremi:
I just noticed that all the tests that use the get table ( https://github.com/EOSIO/eos/blob/master/tests/nodeos_run_test.py#L453 ) are commented out. 
Any plan to fix the get table? I wasn't able to find an issue for it.

yanxi-me:
+1

navcsdev:
Now I can only query by API muti-index. Example: 
`void on( const balance& b, const account_name contract) {
            auto sym = b.symbol.symbol.name();
            stats statstable(contract, sym);
            const auto& st = statstable.get( sym );
            print(st.max_supply.amount);
        }`

arhag:
The issue is that the new currency contract ABI specifies that the accounts table name is "account" when the contract code uses the name "accounts". 

Even though that one small change should resolve the issue, it can still be difficult to use the `get table` commands in cleos to get the data because, for example, getting the appropriate "stat" table requires you to provide the correct scope, which in the case of the currency contract would be the `uint64_t` value representing the symbol name (without the least significant byte containing the precision) but passed into cleos as if it was an account name. For example the "CUR" symbol name would end up being represented by a name like "........edeo3".

To make it easier to use cleos `get table` in situations where more sophisticated scope names are needed (for example, a symbol name or an arbitrary `uint64_t` integer), I have added support (see PR #1874) in chain_plugin to interpret the `scope` string passed in (typically by cleos) to the `get_table_rows` endpoint as either a name, `uint64_t`, `symbol` (e.g. "4,CUR"), or a new built_in_type I added to the ABI serializer called `symbol_name` (e.g. "CUR") in that specific order of conversion attempts.

So, for example, after running `tests/nodeos_run_test.py --dont-kill`, it is possible (as of commit 238a141) to run cleos commands such as:
```
$ programs/cleos/cleos --port 8788 --wallet-port 8899 get table currency CUR stat
{
  "rows": [{
      "currency": "CUR",
      "supply": 1000000000
[...]

Issue #1832, opened 2018-03-26
multi-index bug?


zshchou:
## abi:
```json
"tables": [
    {
      "name": "account",
      "type": "account",
      "index_type": "i64",
      "key_names": ["name1"],
      "key_types": ["account_name"]
    }
  ]
```
## code:
```c++
struct account {
  account_name name1; // 
  account_name name2; //
  uint64_t count;     //
  auto primary_key() const { return name1; }
  EOSLIB_SERIALIZE(account, (name1)(name2)(count))
}

typedef multi_index<N(account), account> account_table;
```
## error:
use cleos get table ...
the row data is confusion.

## fix
```c++
EOSLIB_SERIALIZE(account, (name2)(count))
```
use cleos get table ...
the result is ok! why?

jeremi:
I think this is a duplicate of #1742

andriantolie:
Duplicate. Since #1742 pass, I will put the pass label here too

Issue #1827, opened 2018-03-26
Reading sample "currency" contract balance returns empty rows


CharlesJu1:
I built the master on Ubuntu 16.04. I was able to generate blocks, create currency account, create currency, issue currency, transfer currency  to eosio account. 

The issue is the account table in both eosio and currency has empty rows.

cju@ubuntu-16-04:~/eos-master/eos/build/programs/cleos$ ./cleos push action currency transfer '{"from":"currency","to":"eosio","quantity":"30.0000 CUR","memo":"my 2 transfer"}' --permission currency@active
3370617ms thread-0   main.cpp:935                  operator()           ] Converting argument to binary...
{
  "transaction_id": "efd7b38fdfdc5e55a28d011cc8c9051bfd607c190102351e754689e0ed59df46",
  "processed": {
    "status": "executed",
    "id": "efd7b38fdfdc5e55a28d011cc8c9051bfd607c190102351e754689e0ed59df46",
    "action_traces": [{
        "receiver": "currency",
        "act": {
          "account": "currency",
          "name": "transfer",
          "authorization": [{
              "actor": "currency",
              "permission": "active"
            }
          ],
          "data": {
            "from": "currency",
            "to": "eosio",
            "quantity": "30.0000 CUR",
            "memo": "my 2 transfer"
          },
          "hex_data": "0000001e4d75af460000000000ea3055e09304000000000004435552000000000d6d792032207472616e73666572"
        },
        "console": "transfer\n",
        "region_id": 0,
        "cycle_index": 0,
        "data_access": [{
            "type": "write",
            "code": "currency",
            "scope": "currency",
            "sequence": 3
          },{
            "type": "write",
            "code": "currency",
            "scope": "eosio",
            "sequence": 1
          },{
            "type": "read",
            "code": "currency",
            "scope": "........edeo3",
            "sequence": 2
          }
        ]
      },{
        "receiver": "eosio",
        "act": {
          "account": "currency",
          "name": "transfer",
          "authorization": [{
              "actor": "currency",
              "permission": "active"
            }
          ],
          "data": {
            "from": "currency",
            "to": "eosio",
            "quantity": "30.0000 CUR",
            "memo": "my 2 transfer"
          },
          "hex_data": "0000001e4d75af460000000000ea3055e09304000000000004435552000000000d6d792032207472616e73666572"
        },
        "console": "",
        "region_id": 0,
        "cycle_index": 0,
        "data_access": []
      }
    ],
[...]

PR #1874, opened 2018-03-27
Make get_table_rows via chain_plugin (and thus cleos) easier to use

arhag:
Added `symbol_name` built_in_type to abi_serializer, which is like `symbol` except it does not include the leading precision in the displayed string.

Now treating scope string as (in order of attempts): `name`, `uint64_t`, `symbol`, or `symbol_name`. Only throw error if the string cannot be interpreted as any of those four types. Ambiguity between a name consisting of only the digits 1 through 5 and an actual `uint64_t` number can be resolved by adding whitespace to the beginning or end of the string representation of the `uint64_t` number passed into cleos.

Switched order of scope and code arguments in cleos to better reflect the table structure.

Updated currency.abi to use the more descriptive types (such as `symbol` and `symbol_name`) and also to use the correct table name ("accounts") for the accounts tables.

This PR should fix issue #1742, #1827 and maybe #1832.

### DIFF
--- a/libraries/chain/contracts/abi_serializer.cpp
+++ b/libraries/chain/contracts/abi_serializer.cpp
@@ -59,7 +59,7 @@ namespace eosio { namespace chain { namespace contracts {
 
       //symbol.hpp
       built_in_types.emplace("symbol",                    pack_unpack<symbol>());
-      built_in_types.emplace("symbol_name",               pack_unpack<symbol_name>());
+      built_in_types.emplace("symbol_code",               pack_unpack<symbol_code>());
 
       //asset.hpp
       built_in_types.emplace("asset",                     pack_unpack<asset>());
--- a/libraries/chain/contracts/abi_serializer.cpp
+++ b/libraries/chain/contracts/abi_serializer.cpp
@@ -48,10 +48,6 @@ namespace eosio { namespace chain { namespace contracts {
       );
    }
 
-   struct symbol_name {
-      uint64_t value;
-   };
-
    abi_serializer::abi_serializer( const abi_def& abi ) {
       configure_built_in_types();
       set_abi(abi);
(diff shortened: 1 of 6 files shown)


libraries/chain/contracts/abi_serializer.cpp
```diff
@@ -48,6 +48,10 @@ namespace eosio { namespace chain { namespace contracts {
       );
    }
 
+   struct symbol_name {
+      uint64_t value;
+   };
```
heifner: Should `symbol_name` be a `name` and be defined in types.hpp along with action_name, scope_name, account_name, etc?

arhag: `symbol_name` cannot be a `name` because it must consist of uppercase letters. Perhaps the `_name` suffix is misleading and it should be renamed? But I can't think of a better name right now.

I think it would be more appropriate to put `symbol_name` into symbol.hpp rather than types.hpp. Expect that commit momentarily.

heifner: Yes, symbol.hpp for it and the variant code would be better. How about `symbol_code` I know code is very overloaded, but here at least those things are called currency codes. http://www.xe.com/symbols.php

I have also seen them called "Currency Abbreviation".

arhag: I like `symbol_code`. 

Also, I think the names and methods of the `symbol`-related classes in contracts/eosiolib/asset.hpp should be changed to mirror the symbol.hpp on the native side (and perhaps vice versa), but that is a task for another time when I do the token refactor work.

arhag:
Actually, don't merge this in just yet. I want to investigate something.

arhag:
There was another bug in the code to retrieve tables that was masking the fact that the ABI did not match the serialization used in the currency contract. The code in `chain_plugin` to retrieve table rows called `copy_row` (rather than `copy_inline_row`) which prepended the primary key to the binary data before converting to JSON, which through everything off in the deserialization.

With that bug fixed, the correct ABI for the currency contract could be used to get the appropriate balances and token symbols. And now the results will be pretty printed with the symbol code and the decimal in the right place.

Also the `copy_inline_row` fix should also fix unexpected results of other users of `cleos get table` (i.e. #1832).

Issue #1875, opened 2018-03-27
docker run fails


ekkis:
in https://github.com/EOSIO/eos/blob/master/Docker/README.md there is an instruction that fails:

`docker run --name nodeos -p 8888:8888 -p 9876:9876 -t eosio/eos nodeosd.sh arg1 arg2`

the error it generates is:

> docker: Error response from daemon: OCI runtime create failed: container_linux.go:348: starting container process caused "exec: \"nodeosd.sh\": executable file not found in $PATH": unknown.

arguably `arg1 arg2` are merely exemplary but without them it fails all the same since the script is not found

ekkis:
not surprisingly, the `docker-compose up` fails in the same manner

ekkis:
rather curiously, if I try to inspect the contents of the image:

`docker run --rm --name nodeos -t eosio/eos /bin/sh`

the command:

`# ls`

just hangs.  something else is going on here

ekkis:
however, this shows the file isn't there:

`docker run --rm --name nodeos -t eosio/eos ls /opt/eosio/bin`
> cleos	  eosio-abigen	  eosio-s2wasm	eosiowd  start_nodeos.sh
> data-dir  eosio-launcher  eosiocpp	nodeos

ekkis:
ah.  in comparing `start_nodeos.sh` with the `nodeosd.sh` found in the build directory, I see they're the same.  I don't understand how the file got renamed since the Dockerfile clearly doesn't rename it, but as a fix, we can just fix the documentation to use the correct file name

when I run:
`docker run --rm --name nodeos -t eosio/eos start_nodeos.sh`
I get something like:

> 328432ms thread-0   chain_plugin.cpp:93           plugin_initialize    ] initializing chain plugin
> 328432ms thread-0   http_plugin.cpp:141           plugin_initialize    ] host: 0.0.0.0 port: 8888 
> 328432ms thread-0   http_plugin.cpp:144           plugin_initialize    ] configured http to listen on 0.0.0.0:8888
> 328432ms thread-0   wallet_plugin.cpp:37          plugin_initialize    ] initializing wallet plugin
> 328432ms thread-0   wallet_api_plugin.cpp:112     plugin_initialize    ] 
> *************************************
> *                                   *
> *  --   Wallet NOT on localhost  -- *
> *  - Password and/or Private Keys - *
> *  - are transferred unencrypted. - *
> *                                   *
> *************************************

[...]

PR #1878, opened 2018-03-27
docker-compose fix

ekkis:
fixes bad reference to the startup command to the `nodeos` section

### DIFF
--- a/Docker/docker-compose.yml
+++ b/Docker/docker-compose.yml
@@ -5,7 +5,7 @@ services:
     build:
       context: .
     image: eosio/eos
-    command: /opt/eosio/bin/nodeosd.sh
+    command: /opt/eosio/bin/start_nodeos.sh
     ports:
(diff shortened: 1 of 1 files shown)


peterwillcn:
@ekkis Dockerfile has been renamed nodeosd in line with the UNIX naming rules, and your changes are out of date.

ekkis:
I just pulled from github so how my modification be out of date? the current code on github has the following line:

`COPY nodeosd.sh /opt/eosio/bin/nodeosd.sh`

but if you look in the image, the script is called `start_nodeos.sh`, ergo the fix.  see:

https://github.com/EOSIO/eos/issues/1875

peterwillcn:
@ekkis 
The following operations can be used to solve the problem:
docker rmi eosio/builder
cd eos/Docker/builder && docker build . -t eosio/builder
cd eos/Docker && docker-compose up -d

PR #1880, opened 2018-03-27
fix txn_test_gen_plugin for stress test(DAWN-562)

taokayan:


plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
```diff
@@ -243,13 +251,26 @@ struct txn_test_gen_plugin_impl {
...
+      }
+
+      act_a_to_b.data = currency_serializer.variant_to_binary("transfer", fc::json::from_string(fc::format_string("{\"from\":\"txn.test.a\",\"to\":\"txn.test.b\",\"quantity\":\"1.0000 CUR\",\"memo\":\"${l}\"}", fc::mutable_variant_object()("l", salt))));
+
+      act_b_to_a.data = currency_serializer.variant_to_binary("transfer", fc::json::from_string(fc::format_string("{\"from\":\"txn.test.b\",\"to\":\"txn.test.a\",\"quantity\":\"1.0000 CUR\",\"memo\":\"${l}\"}", fc::mutable_variant_object()("l", salt))));
+
       signed_transaction trx;
       trx.actions.push_back(act_a_to_b);
-      trx.actions.emplace_back(action({}, config::system_account_name, "nonce", fc::raw::pack(nonce)));
+      //trx.context_free_actions.emplace_back(action({}, config::system_account_name, "nonce", fc::raw::pack(nonce)));
```
pmesnier: Just delete the commented out code.

taokayan:
the reason of removing the nonce action is that: the nonce action is now considered as context_free_action and it is no longer able to make the transaction unique.
so my change is to make the "memo" field unique instead.

Issue #1735, opened 2018-03-21
DAWN-695 ⁃ Deferred transaction bugs


moskvanaft:
1. cancel_deferrerd is able to cancel transaction which will be created in the same action after the call to cancel_deferred
2. it's possible to create multiple transactions with the same sender_id
3. minor bug: print() doesn't work in deferred transactions (nothing appears in console)
4. apply_context::execute_deferred uses default_max_gen_trx_count instead of actual parameter

gleehokie:
Fixed by PR #1883

andriantolie:
Case 1 and 2 are verified by the existing api_tests/deferred_transaction_tests unit tests, while case 3 and 4 are verified by the additional unit test https://github.com/EOSIO/eos/pull/2173

PR #1883, opened 2018-03-27
Deferred transaction fixes

moskvanaft:
Fixes following bugs:
1. cancel_deferrerd was able to cancel transaction which will be created in the same action after the call to cancel_deferred
2. it was possible to create multiple transactions with the same sender_id
minor bug: print() doesn't work in deferred transactions (nothing appears in console)
3. apply_context::execute_deferred used default_max_gen_trx_count instead of actual parameter


gleehokie:
Fixes #1735

andriantolie:
The related issue #1735 passes the test

PR #1898, opened 2018-03-28
fix(exchange): include hpp instead of cpp

Akagi201:
In C++, It's bad to include cpp files. It's best practice to include header files.


bytemaster:
Changing the name doesn’t change the nature. The proper fix is to build it as multiple .o files.

PR #1917, opened 2018-03-29
Eosio.any

bytemaster:
Created tests on linking actions to eosio.any permission necessary to support contract-defined permission requirements such as those needed by eosio.msig


heifner:
Verified unittest is run as part of chain_test.

Issue #1938, opened 2018-03-29
Add handlers for native actions in system contract 


zorba80:

andriantolie:
Can we have ATC for this?

gleehokie:
@andriantolie 

Verify that native action handlers in system contract are being called properly.

zorba80:
For now, the handlers don't do anything. The way I verified that they are being called properly is by including print statements.

brianjohnson5972:
@gleehokie actually, (assuming that I am correct that this is referring to the apply method in eosio.system.hpp) it is not complete, it is not processing the currency::create action.  But shouldn't this task really be to come up with a way in code (either in the contract or with a test) to ensure that this remains to be the case?  We can verify right now and someone can commit a new message and it be insufficient immediately.

gleehokie:
@brianjohnson5972 @zorba80 I certainly like having automatic ways to handle the known case where the addition of new action/message will create untested conditions.  I think this is what @brianjohnson5972 is suggesting.

gleehokie:
@zorba80 and I talked, here's the ATC:

1. Verify that all native actions are routed to the system contract.  The current system contract has native action handlers in the system contract that are empty functions.  The system contract chooses what native actions to handle, if no handler is present the action is ignored.
2. A suggestion for how to do this is to modify eos/contracts/eosio.system/native.hpp and add a print statement to each of the handlers which are the on() functions, for example this line: https://github.com/EOSIO/eos/blob/2cb8030554d1e9145f44785b9a564cd6dd1e1cab/contracts/eosio.system/native.hpp#L98
This will verify that all of the native actions that have handlers in the system contract are being handled.
3. At a later date we can write a more complex test case that generates code to handle and verify **all** of the native actions by pulling the list from the code.

andriantolie:
ATC Passes.

I modified the handler inside native.hpp to print a statement. Also, for postrecovery/ passrecovery/ vetorecovery/ onerror, I modified recovery_tests to be able to reach the desired action handler. Running chain_test with --verbose, I can see the statement printed by native handler is called for every action.

PR #1942, opened 2018-03-29
GH#1938 - Add handlers for native actions in system contract

zorba80:


libraries/chain/apply_context.cpp
```diff
@@ -19,11 +19,13 @@ void apply_context::exec_one()
       if (native) {
          (*native)(*this);
       }
-      else if (a.code.size() > 0) {
+
+      if (a.code.size() > 0 && !(act.name == N(setcode) && config::system_account_name)) {
```
larryk85: This should be (receiver == config::system_account_name).

contracts/eosio.system/native.hpp
```diff
@@ -0,0 +1,83 @@
...
+#pragma once
+
+#include <eosiolib/types.hpp>
+
+namespace eosiosystem {
+   template <account_name SystemAccount>
+   class native {
+      public:
+         ACTION(SystemAccount, newaccount) {
+            EOSLIB_SERIALIZE(newaccount, BOOST_PP_SEQ_NIL)
```
larryk85: I think you forgot to add the fields for the actions.  These should be the same as the native actions.

brianjohnson5972:
#1938

Issue #1766, opened 2018-03-22
DAWN-671 ⁃ Stricter Authority Checks


blockone-syncclient:
There are a few pathological cases where we are not properly validating transactions and should be:

1. Now that CFAs exist all "normal" actions should require some sort of authorization.  This should be enforced and does not appear to be.
1. Enforcing the above will defeat the current "could be CFA" check.  That check should be changed to only respect the api usage.  In addition, this should be the **first** assertion checked so that any action that could be CFA is reported that way regardless of the other "normal" transaction rules. (see chain_controller __apply_transaction()).
1. We should enforce that the list of authorization is concise.  This should be done by collecting the authorizations actually needed during application (by tracking calls to `apply_context::require_authorization`) and verifying that the provided authorization exactly matches the used authorizations
1. We should enforce that all transactions have at least one valid "normal" action.  

ATC: demonstrable (and passing) unit test cases for all of the above.

bytemaster:
All transactions require at least one authorization, but not actions require an authorization. The only reason to require an authorization on a transaction is to know who to bill for bandwidth.

bytemaster:
Rather than enforcing authorizations to be concise, we could simply bill the user for computation based upon the number of authorizations specified.

andriantolie:
ATC Passes.
api_tests/cf_action_tests has verified the ATC and it passes

PR #1969, opened 2018-03-29
Feature/stricter auth checks

larryk85:
Added stricter checks for action/context-free action and transaction authorities.  
#1766


libraries/chain/include/eosio/chain/transaction_metadata.hpp
```diff
@@ -77,3 +79,4 @@ class transaction_metadata {
 
 } } // eosio::chain
 
+FC_REFLECT( eosio::chain::transaction_metadata, (raw_trx)(signing_keys)(id)(region_id)(cycle_index)(shard_index)(bandwidth_usage)(published)(sender)(sender_id))
```
brianjohnson5972: Should is_implicit be added to reflection?

larryk85: Yes, it should.

gleehokie:
Dan is not sure we want these checks, too restrictive.

gleehokie:
Dan and Bucky talked, we're going ahead and merging.

gleehokie:
For some reason Jenkins tried to mark the test as passed but GitHub rejected it.  Appears to be a GH issue.  Going ahead and merging.

andriantolie:
The related issue is #1766 and it passes

PR #1979, opened 2018-03-30
enforce resource limits

### LINKED ISSUE #353: DAWN-472 ⁃ Bandwidth Rate Limiting & Storage Usage Limits
## Types of Consensus
1. Soft Consensus - enforced only by block producers
2. Hard Consensus - enforced by all full nodes

## Resource Types
1. Bandwidth (rate limited, soft consensus)
2. Computation (rate limited, soft consensus)
3. Database  (locks tokens until released, hard consensus)
4. IPFS (locks tokens until released, hard consensus)
5. Print Log (per-contact stdout)
6. Event Log (per-contact structured output)

## Measuring Resources
1. Bandwidth - binary size of processed transaction as it exists in block
2. Computation - all transactions assumed to consume max quantum (1 ms) of CPU
     - each scope has its own measure of average use and all scopes (read or write) charge for 1 ms
3. Database - based on size of created / released records + estimated overhead
4. IPFS - based on the size of the declared file and only after producer confirms it as stored

## Who to Bill
In this case billing applies to which account has their resource usage adjusted.

1. Bandwidth - the user(s) who authorized the transaction (all billed equally even if it double bills)
2. Computation - the user(s) who authorized the transaction (all billed equally even if it double bills)
3. Database - the code account which executes the transaction
4. IPFS - the user who requests the file be stored in their account and can delete at will

## Challenges
The biggest challenge with all resource usage is to perform these calculations in a way that does not force everything into a single thread.  The second biggest challenge is how to build blocks incrementally while doing resource calculations in parallel. 

1. the authorizing accounts are independent of the scopes and we would like to avoid looping them into the scope.
2. the code account (used to bill storage) is often different from scope accounts (eg currency contact) 
3. storage and bandwidth use isn't known until after applying the transaction
4. The order in which transactions are applied impacts the global usage and therefore the fractional reserve status on the various resources. This means there must be a deterministic means of updating this data. 

## Potential Solutions
1. resource usage data is "constant" during a block and updated "between" blocks.  
     - this works well for rate limiting 
     - this does not work for storage (people can get double storage)
2. apply resource usage as a 2nd pass validation over the elements in a cycle
     - this allows us to pipeline resource usage updates with the next cycle
     - this may cause us to invalidate previously included transactions 
          - this is like a branch miss-prediction and forces us to reset and re-apply a cycle and every cycle after it.

## Goals
1. Enable Free Applications for Users
2. Enable Developers to cover User Costs
3. Don't make developers re-invent rate limiting 
4. Don't expose users to unexpected fees
5. Keep things as soft consensus and not part of validation consensus 

From a platform perspective these goals are conflicting. If applications cover the costs then applications must implement their own rate limiting algorithm to prevent abuse and that also means the rate limiting is part of the hard consensus.

## Assumptions

1. User's can control if they transact and thus to spam or not to spam
2. Application code controls allocation and release of storage
3. Most users can transact with a very small balance 
4. Storage locks tokens until it freed

## High Level Approach

1. Charge rate-limited resources to users
2. Charge database resources to application 
3. Charge IPFS to user 

## Principles
1. User's initiate consumption of resources and are 

Each write-scope implies a single-threaded throughput limit and maintains its own independent accounting of how many computation units an individual user has consumed in that scope.  

Note: this issue is still under development and will need to be defined in greater detail before work can begin.

wanderingbort:
This resolves #353 (eventually)

I wanted to start the PR process early while I finish up some of the interfacing and billing details

commits have more details about individual changes along the way

Still todo:
- [ ] full-integration testing (only basic functional tests are present)
- [x] add an objective limit set by the system contract on the max cpu usage any transaction can use
- [x] add network cost set by system contract per read/write lock in a transaction
- [x] add base network cost set by system contract per transaction
- [x] add pre-committed cpu usage from context free actions 
- [x] add discount set by system contract on all cpu usage associated with context free actions
- [x] add discount set by system contract on net usage from prunable data (signatures and context free data) enforced when checking the commitment to the packed size

@moskvanaft @zorba80 - I had to disable the call to `set_resource_limits` from the system contract because the values used in the test cases to exercise things like refunds and voting violate some basic principles of the resource system like:
 - its impossible to completely unstake all the ram for an account... where would we store _the account_ itself
 - Unstaking _all_ of the cpu_resources from an account leaves 0 capacity for delivering the `refund` action... maybe system floats some minimum stake to all accounts at all time to make stuff like this work out?
 - at all times, the account needs to maintain some stake, on occasion the voting tests would unstake too much and then not be able to proceed.


libraries/chain/chain_controller.cpp
```diff
@@ -300,12 +308,13 @@ transaction_trace chain_controller::_push_transaction(const packed_transaction&
 
          time_point_sec execute_after = head_block_time();
          execute_after += time_point_sec(delay);
-         deferred_transaction dtrx(context.get_next_sender_id(), config::system_account_name, execute_after, trx);
+         //TODO: !!! WHO GETS BILLED TO STORE THE DELAYED TX?
```
wanderingbort: who gets billed for storing auto-boxed transactions?   the authorizer with the highest delay?

libraries/chain/chain_controller.cpp
```diff
@@ -1449,6 +1474,28 @@ void chain_controller::update_signing_producer(const producer_object& signing_pr
    } );
 }
 
+void chain_controller::update_permission_usage( const transaction_metadata& meta ) {
```
wanderingbort: this was refactored out of the old `update_usage` method and made independent.

libraries/chain/include/eosio/chain/account_object.hpp
```diff
@@ -29,7 +29,6 @@ namespace eosio { namespace chain {
       shared_vector<char>  abi;
 
       void set_abi( const eosio::chain::contracts::abi_def& a ) {
-         // Added resize(0) here to avoid bug in boost vector container
```
wanderingbort: I should restore this comment... not sure why I took it out

libraries/chain/include/eosio/chain/contracts/contract_table_objects.hpp
```diff
@@ -94,7 +95,8 @@ namespace eosio { namespace chain { namespace contracts {
          table_id      t_id;
          uint64_t      primary_key;
          SecondaryKey  secondary_key;
-         account_name  payer;
+         account_name  payer = 0;
+         uint16_t      billed_overhead = 0;
```
wanderingbort: Billed overhead should be removed, it is not in use

libraries/chain/include/eosio/chain/account_object.hpp
spoonincode: I mean.. am I the only one who is seriously bothered by this? @PaulCalabrese did we ever ask boost folks about this?

wanderingbort: you are not the only one... but I can confirm that taking it out -OR- moving it all into the `create` call causes tests to fail in that same odd bug.   I had tried to at least consolidate the abi and the code "hack" into the same logical place but it seems there is some corruption that can occur between `create`/constructor and future calls.

libraries/chain/wasm_interface.cpp
```diff
@@ -113,34 +114,24 @@ class privileged_api : public context_aware_api {
...
+       * update the resource limits associated with an account.  Note these new values will not take effect until the
+       * next resource "tick" which is currently defined as a cycle boundary inside a block.
+       *
+       * @param account - the account whose limits are being modified
+       * @param ram_bytes - the limit for ram bytes
+       * @param net_weight - the weight for determining share of network capacity
+       * @param cpu_weight - the weight for determining share of compute capacity
+       */
+      void set_resource_limits( account_name account, int64_t ram_bytes, int64_t net_weight, int64_t cpu_weight) {
+         EOS_ASSERT(ram_bytes >= -1 && ram_bytes <= INT64_MAX, wasm_execution_error, "invalid value for ram resource limit expected [-1,INT64_MAX]");
```
wanderingbort: INT64_MAX is not an effective test on an int64_t

libraries/chain/wasm_interface.cpp
```diff
@@ -1030,7 +1021,15 @@ class transaction_api : public context_aware_api {
          context.execute_context_free_inline(std::move(act));
       }
 
-      void send_deferred( const unsigned __int128& val, const fc::time_point_sec& execute_after, array_ptr<char> data, size_t data_len ) {
+      void send_deferred( const unsigned __int128& val, account_name payer, const fc::time_point_sec& execute_after, array_ptr<char> data, size_t data_len ) {
+         account_name actual_payer = payer;
+         if (actual_payer != account_name(0)) {
+            const auto* paying_account = context.db.find<account_object, by_name>(payer);
+            EOS_ASSERT(paying_account, tx_unknown_argument, "The account for the payer: ${a}, does not exist!", ("a", payer));
```
wanderingbort: remove this in favor of failing when trying to bill

libraries/chain/chain_controller.cpp
wanderingbort: bill to highest delay

bytemaster:
I went over this with bart, looks good, outstanding issues can be addressed in future pull request.

taokayan:
Is there any ATC for this?

PR #1991, opened 2018-03-30
Re-enable mongo test and fix intermittent timing issues

PaulCalabrese:
The mongo test was failing because of conflicting mongodb plugins in node_00 and node_bios. As a quick fix, I changed the launcher code to not pass mongo-related options through to the bios node. I also removed some extra sleep time from the mongo test as I don't think it is needed any more.


PaulCalabrese:
This PR's changes were merged as part of PR2050.

PR #2010, opened 2018-03-31
Usability Enhancements of Cleos

bytemaster:


libraries/chain/include/eosio/chain/contracts/genesis_state.hpp
```diff
@@ -36,8 +36,8 @@ struct genesis_state_type {
       .max_generated_transaction_count = config::default_max_gen_trx_count
    };
 
-   time_point                               initial_timestamp;
-   public_key_type                          initial_key;
+   time_point                               initial_timestamp = fc::time_point::from_iso_string( "2018-03-01T12:00:00" );;
```
kesar: double semicolon. why is this hardcoded?

I guess temporal

Issue #1973, opened 2018-03-29
Run All Chain Tests work on both Producer and Validator Nodes


bytemaster:
Just because a transaction sequence works on a "sole producer" node, does not mean that the resulting block will validate if pushed to another node. A recent example of this discovered that scopes on the shard failed checks after-the-fact that passed during generation.

We have a tester framework, this tester framework needs to create "two chains".  Every time a block is produced on the main chain, it needs to be copied and pushed to the second chain. The two chains should be identical.

Unit testers should not see this behavior as it should all be behind the scenes of the tester.

gleehokie:
@andriantolie 

ATC: 

1) verify that both chains running properly and that blocks being produced on the main chain are pushed to the secondary chain
2) verify that blocks are accepted and are valid on the second chain
3) verify that the scopes on the shard (see description) are valid

brianjohnson5972:
@gleehokie the ATC seems to be more what the code review should have verified.  This issue was for a unit test to perform those things, so not sure why we are testing things that the unit test should perform.

gleehokie:
@brianjohnson5972 I struggled myself with an ATC for this one.  What I wrote essentially says "make sure it is doing what was intended", to verify the one problem case that was documented in the description no longer happens and to verify all of this happens correctly from a functional test point of view versus code review.  Tpfft.

andriantolie:
This issue is already validated by validating_tester, which is the default tester when NON_VALIDATING_TEST macro is not specified. Since running chain_test with no NON_VALIDATING_TEST macro specified passes, the solution to this issue is verified to be working.

PR #2018, opened 2018-04-01
Enhancement/dual node tester

larryk85:
Added a new tester type that validates the chain produced during unit tests.  This PR is for issue EOSIO/eos#1973 Run All Chain Tests work on both Producer and Validator Nodes.


libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -171,21 +176,85 @@ namespace eosio { namespace testing {
...
          chain_controller::controller_config           cfg;
-
          map<transaction_id_type, transaction_receipt> chain_transactions;
    };
 
    class tester : public base_tester {
-   public:
-      tester(chain_controller::runtime_limits limits = chain_controller::runtime_limits());
-      tester(chain_controller::controller_config config);
+      public:
```
brianjohnson5972: remove the indentation

libraries/testing/tester.cpp
```diff
@@ -28,8 +28,9 @@ namespace eosio { namespace testing {
       }
       return res;
    }
-
-   base_tester::base_tester(chain_controller::runtime_limits limits) {
+
```
brianjohnson5972: Can we get rid of some of this extra whitespace and extra newlines that was added? (would make the diff of actual code a lot easier to see)

larryk85: Sorry about that, will clean it up.

Issue #2020, opened 2018-04-01
Softfloat integer truncation overflow checks are wrong, and missing softfloat conversion tests.


larryk85:
For the integer truncation operations, the checks for overflow were incorrect.  They only check for the floating point number to equal the min or max for the integer type being truncated to.  To fix this, change the condition to check for greater than or less than max or min, this will also remove the need to check for infinity and negative infinity as they will be subsumed by the previously mentioned checks.

Also, the conversion spec tests for softfloat and corresponding f32_f64_conversion_tests need to be updated and fixed.

ATC:
Please verify that a floating point number 32 bit (float) and 64 bit (double) will cause an overflow assertion when trying to assign (truncate) this to an int32_t, uint32_t, int64_t, and uint64_t with values greater that int max and less that int min respectively.
Caveat:
The compiler could try to truncate these implicitly before wasm evaluation. One way around this is to write a contract that writes the floating point value to the database and then reads the data back and then assigns this to an int. The compiler cannot optimize anything away by doing this and you should get the wasm ops generated for the various truncations.

andriantolie:
Can we have ATC for this?

brianjohnson5972:
@larryk85 please add an ATC

larryk85:
Sorry, about that.  
Please verify that a floating point number 32 bit (float) and 64 bit (double) will cause an overflow assertion when trying to assign (truncate) this to an int32_t, uint32_t, int64_t, and uint64_t with values greater that int max and less that int min respectively.
Caveat:
The compiler could try to truncate these implicitly before wasm evaluation.  One way around this is to write a contract that writes the floating point value to the database and then reads the data back and then assigns this to an int.  The compiler cannot optimize anything away by doing this and you should get the wasm ops generated for the various truncations.

taokayan:
I've made a pull request to fix the truncation overflow check, and adding related unittest cases. 

details explanation:

-2147483649.0f is not representable in IEEE float32 format(only 24 bit precision in mantissa, see https://en.wikipedia.org/wiki/Single-precision_floating-point_format). Instead, it will be truncated into -2147483648.0f at compile time. Therefore,

_eosio_f32_le(af, -2147483649.0f))

should be changed into

_eosio_f32_lt(af, -2147483648.0f))


unitest results:
    Test case "wasm_tests/f32_f64_overflow_tests" has passed with:
      48 assertions out of 48 passed

PR #2021, opened 2018-04-01
Fixed truncation overflow checks

larryk85:
Fixed all truncation operations to check for overflow correctly.  In addition, I noticed that the conversion spec tests did not get merged.  I uncommented the spec tests in wasm_tests and fixed the conversion tests to now run.


brianjohnson5972:
#2020

Issue #1755, opened 2018-03-22
DAWN-606 ⁃ Signatures/Authorities are not checked when applying a block


blockone-syncclient:
The code path that it takes has this check commented out.

andriantolie:
Can we have ATC for this?

andriantolie:
This issue seems already covered by the unit test block_tests/irrelevant_sig_hard_check. Since the unit test passes, we can conclude that this issue is fixed.

Issue #1753, opened 2018-03-22
DAWN-693 ⁃ We are not currently validating: tapos, expiration, uniqueness for transactions


blockone-syncclient:
ATC: demonstrable test cases that show proper rejection of bad tapos, expiration and uniqueness

bytemaster:
We are checking for duplicate transactions.

andriantolie:
Related PR is https://github.com/EOSIO/eos/pull/1865

taokayan:
Is there any detail steps for QA testing?

taokayan:
new unitest cases for validating ta_pos, and transaction expiration are added. Please see pull request https://github.com/EOSIO/eos/pull/2055

PR #2035, opened 2018-04-02
Check signatures, transaction validation checks, and many other bug fixes

arhag:
Changes in this pull request:
 * Fixes to scoped_exit.
 * Check for redundant signatures due to multiple signatures from same key.
 * Fix bug in record_locks_for_data_access (did not account for intersection between read_locks of transaction_trace and write_locks of existing shard_trace).
 * Put limits on number of keys/accounts in authorities (but a high limit like 2^16) to make overflow of accumulated weights impossible.
 * Disallow empty parent for permissions other than owner.
 * Actually enforce the existence of permissions contained in an updated authority (changed `find` to `get`).
 * Reserve account names starting with "eosio." (previously was a much stricter check of whether the name contains "eosio.").
 * Reserve permission names starting with "eosio."
 * Create dummy `permission_object` as first object in `permission_index` during chain initialization so that a parent id of 0 can never refer to a legitimate object.
 * Use `fc::microseconds` for delay rather than `time_point`.
 * Fixed bug with how a permission visitor works with the `authority_checker`. It would visit permissions that did not end up contributing towards satisfying the authority. My quick fix (a more elegant fix is eventually desired) causes the permission visitor to visit unneeded permissions but now with extra surrounding information (`push_undo`, `pop_undo`, `squash_undo`) to help it discard the side effects of visits to permissions that ended up not being used. This bug had implications on the minimum delay calculated by check_authorization (they could have been higher than necessary).
 * Changes to validating signatures/authorities/TaPoS/expiration/uniqueness (affects #1753 and #1755): see details below.

There are two stages to validating a transaction. 

The first stage involves checks that can be made with minimal state information: just the information either included in the `packed_transaction` itself or the state available as of the end of the previous block (specifically head block timestamp and TaPoS block ids). The first stage of checks are mostly done in a function called `validate_transaction_with_minimal_state`. It checks that there is at least one action in the transaction, that is has not expired yet, that its TaPoS referenced block is valid, and that the network usage due to the `packed_transaction` alone is not too much to cause the transaction to be invalid based on its (soon to be) upper bound on net usage committed in the transaction header. There is also another check/processing that can be done in this first stage (except for generated transactions): recovering the public keys from the signatures and ensuring there are not multiple signatures signed by the same key.

The second stage involves the remaining checks that require the state information at the point at which the transaction is scheduled to execute (or be delayed). They include the checks on uniqueness of the transaction, that the expiration in the transaction header is not too far in the future relative to the appropriate reference time (head block time + any delay imposed) based on the limits in the chain parameters, and that the accounts referenced in the authorizations of the transaction exist. They also include checking the authorizations using: 
  * if an input transaction, the set of recovered public keys from the signatures;
  * or if a generated transaction, only the owner permission of the contract code that generated the transaction.

For delayed input transactions or generated transactions (whether delayed or not), checks occur at two different times. First, the delayed input or generated transaction is checked using the stage 1 and stage 2 checks at the time of dispatch. Then, at the time of scheduling/execution additional checks are performed: check that the transaction is not executing prematurely, check that the transaction has not expired yet again, and check uniqueness again.


libraries/chain/include/eosio/chain/authority_checker.hpp
```diff
@@ -109,6 +117,11 @@ namespace detail {
          {}
 
          bool satisfied(const permission_level& permission, uint16_t depth = 0) {
+            // QUESTION: Should we not catch the chain::permission_query_exception from permission_to_authority and try to
```
wanderingbort: I vote "yes" we treat failure from `permission_to_authority` callback as if `satisfied` returned `false`.

libraries/chain/chain_controller.cpp
```diff
@@ -344,32 +306,18 @@ transaction_trace chain_controller::_push_transaction(const packed_transaction&
...
 transaction_trace chain_controller::_push_transaction( transaction_metadata&& data )
 { try {
    auto process_apply_transaction = [this](transaction_metadata& meta) {
       auto cyclenum = _pending_block->regions.back().cycles_summary.size() - 1;
       //wdump((transaction_header(meta.trx())));
 
+      const auto& trx = meta.trx();
+
+      // Validate uniqueness and expiration again
+      validate_uniqueness(trx);
```
wanderingbort: is this only here to cover the execution path where we extract a `generated_transaction_object` into a `transaction_metadata` and call `_push_transaction` ?

If so, can we create a legit `_push_transaction(const generated_transaction_object& )` and put this logic in there as well as any other stuff like `transaction_metadata` creation?

arhag: Yes, that is exactly what it is for.

The way it is coded up now it does mean that a redundant check for uniqueness and  expiration will happen for undelayed input transactions. The `_push_transaction(const generated_transaction_object& )` overload you suggest is a way to get around that, but I think it is okay for the time being and we should delay that change until we eventually refactor the current code in chain_controller.cpp to be much easier to follow.

libraries/chain/include/eosio/chain/authority_checker.hpp
arhag: Done. Changes in commit 687e6ea.

PR #2042, opened 2018-04-02
fixes for review of PR 1979

### LINKED ISSUE #2031: Consolidated fixes for #1979 feedback
#1979 had a few issues come up during on-site review that were logged in the now merged PR by @wanderingbort on behalf of @bytemaster.  This is an issue to resolve these issues:

- [x] ~~bill delayed transactions to the authorizer who imposed the highest delay, break ties with order of declaration~~ Removed Req in favor of billing the first declared on first action
- [x] remove ability to specify `0` as the payer for a deferred transaction.  Require an explicit sender including when the sender is the contract account
- [x] allow privileged sender to bill _anyone_ for a deferred transaction's storage
- [x] allow privileged contract to bill _anyone_ for a database row
- [x] bill storage of new accounts to the account that creates them
- [x] remove use of `sizeof` when calculating ram billing (as it is potentially platform/compiler specific)
- [x] ~~change spec to not double bill accounts which appear multiple times in a transaction~~ Removed Req in favor of billing the first declared on first action
- [x] clean up use of IN64_MAX as a cap on an int64_t
- [x] remove pre-check for account existing when sending a deferred, fail on billing instead

wanderingbort:
This closes #2031 

That issue has the details.


libraries/chain/chain_controller.cpp
```diff
@@ -367,8 +367,12 @@ transaction_trace chain_controller::delayed_transaction_processing( const transa
 
    time_point_sec execute_after = head_block_time();
    execute_after += mtrx.delay;
-   //TODO: !!! WHO GETS BILLED TO STORE THE DELAYED TX?
-   deferred_transaction dtrx(context.get_next_sender_id(), config::system_account_name, config::system_account_name, execute_after, trx);
+
+   // TODO: update to better method post RC1?
+   FC_ASSERT(mtrx.trx().actions.size() > 0 && mtrx.trx().actions.at(0).authorization.size() > 0, "Delayed transactions must have at least one action and those must have at least one authorization");
```
arhag: This is imposing stricter requirements than our current requirement that at least one authorization must exist for some action in the transaction.

Instead, you could search through the actions for the first action with authorizations, and then choose the actor of the first authorizer of that action.

libraries/chain/include/eosio/chain/contracts/contract_table_objects.hpp
```diff
@@ -148,7 +147,42 @@ namespace eosio { namespace chain { namespace contracts {
    typedef secondary_index<uint64_t,index_double_object_type,soft_double_less>::index_object  index_double_object;
    typedef secondary_index<uint64_t,index_double_object_type,soft_double_less>::index_index   index_double_index;
 
-} } }  // namespace eosio::chain::contracts
+} // ::contracts
+
+namespace config {
+   template<>
+   struct billable_size<contracts::key_value_object> {
+      static const uint64_t overhead = overhead_per_row_per_index_ram_bytes * 2;  ///< overhead for 2x indices internal-key and primary key
```
arhag: We should probably also include the overhead of storing the `table_id_object`. This would overcharge for any table that isn't a singleton, but I don't know of any other simple solution to prevent abuse of memory by singleton tables.

Similar changes would be needed for the secondary index tables below.

arhag:
Another thing I noticed is that there is a potential for overflow in calculation of the total net usage in `chain_controller::calculate_transaction_net_usage`:

https://github.com/wanderingbort/eos/blob/af9a6d0ba6d7ba1b0d99a7e809a99f1dd74beaba/libraries/chain/chain_controller.cpp#L1226-L1234

A malicious user could take advantage of the overflow by setting a large value in `net_usage_words` in the transaction header such that the calculated transaction net usage ends up smaller than it should be.

This problem might go away when we change the meaning of the user supplied net and cpu usage fields in the transaction to be upper limits, but it is something to still keep in mind when working on the `calculate_transaction_net_usage` and `calculate_transaction_cpu_usage` code in the future. 

Furthermore, in `calculate_transaction_cpu_usage`, the returned value seems to be in CPU usage units rather than kilo CPU usage units. So there is bug with the last line in the code snippet below:

https://github.com/wanderingbort/eos/blob/af9a6d0ba6d7ba1b0d99a7e809a99f1dd74beaba/libraries/chain/chain_controller.cpp#L1218-L1223

It should instead return `meta.trx().kcpu_usage * 1000`. Technically overflow is also possibly here, but it wouldn't actually be advantageous to a user because the only way the code can get to the that line is if the overflowed value is greater or equal to `actual_usage`. Nevertheless, it would be best to check for overflows like that and return an appropriate error message.

libraries/chain/include/eosio/chain/contracts/contract_table_objects.hpp
wanderingbort: instead of including that overhead there (which became substantial) we are now billing the new table overhead to the first row's payer at the time it was created.  This does not track that rows payer as it changes.

When the last row of a table is deleted, we now reap the table and return the bytes to the original payer.

PR #2045, opened 2018-04-02
Multisig contract

moskvanaft:


contracts/eosiolib/dispatcher.hpp
```diff
@@ -39,10 +39,17 @@ namespace eosio {
 
    template<typename T, typename... Args>
    bool execute_action( T* obj, void (T::*func)(Args...)  ) {
-      char buffer[action_data_size()];
-      read_action_data( buffer, sizeof(buffer) );
+      size_t size = action_data_size();
+      char default_buffer[512];
```
wanderingbort: In order to not unnecessarily bloat the stack for actions below 512, can we refactor the unpack of args + call of function into another templated method and implement this method as a single branch on size that either reads into a malloc'd buffer OR a tightly sized stack buffer (as it did before) and then calls the newly refactored `execute_action_from_buffer(T*, void(T::*func)(Args...), buffer, size)` method?

Ideally, for non-large actions, the only difference between the new implementation and the former should be the cost of a branch and a trivially inline function call with no additional stack usage.

contracts/eosiolib/multi_index.hpp
```diff
@@ -439,10 +439,18 @@ class multi_index
 
          auto size = db_get_i64( itr, nullptr, 0 );
          eosio_assert( size >= 0, "error reading iterator" );
-         char tmp[size];
-         db_get_i64( itr, tmp, uint32_t(size) );
 
-         datastream<const char*> ds(tmp,uint32_t(size));
+         char default_buffer[512];
```
wanderingbort: while this method should be on top of the stack instead of buried, it is still probably worth seeing if we can refactor the such that the "usual" case of a small item is still doing a tight stack allocation

contracts/eosiolib/dispatcher.hpp
wanderingbort: It does all come from the same memory but because of some oddities we still need to treat it differently.  

There are 2 "types" of memory for WebAssembly (and I'm paraphrasing right now): Linear memory and the local variables on the stack.  And in C/C++ there are 2 types of memory stack and heap.  In WebAssembly, only linear memory is "addressable".  In C/C++ both are.

So, you cannot get a pointer to a variable held as a local variable.  To "fix" this, the tool-chain creates a simulated "stack" in linear memory and puts things that would normally be locals into that region of linear memory so that they can be addressable.  

This would suggest that "stack" is linear memory and its all the same thing.  Except it works differently because of when/how it is managed.  The "stack" becomes a compile time region of linear memory which is sized _at compile time_ and increases the base memory we allocate to a contract.  If this stack is exhausted, bad things happen.

We also treat the linear memory as the "heap" but it is managed at runtime.  We provide methods for expanding the size of that heap for incidental usage (actually its an instruction at the web assembly level these days).

So, the issue with a large "stack" allocation is that, even though its all linear memory, we have to size the base allocation for a contract (which we then have to pay to clean up) to account for the maximum virtual stack size.  Whereas, moving it to the "heap" means we can control the size of it at run-time allowing more efficient operation for the actions that do not require a large allocation.

dhaneshvb:
Marking it as passed as multisig_tests passed

Issue #2002, opened 2018-03-31
Canceling a delayed transaction should use the transaction id rather than a sender_id


arhag:
Related to #1022.

When a transaction needs to be delayed, the system creates a generated transaction with the appropriate `delay_until` time set. Like all generated transactions, it needs a `sender` (which is an account) and `sender_id` (a `uint128_t` number). In the case of delayed transactions, the `sender` is always the system account. The `sender_id` can be anything as long as it is unique for each unique delayed transaction. 

Currently an auto-incrementing number is used as the `sender_id`. So a user who signs a transaction which is delayed does not automatically know what `sender_id` to use to refer to that transaction they just signed and broadcast. Currently, the `canceldelay` action requires the `sender_id` of the delayed transaction to cancel. So a user would be forced to trust a full node to map the transaction id which they know to the `sender_id`. And our block structure does not provide any way for such a node to prove to a user (who trusts the block producers) that the mapping was correct.

Instead, the `canceldelay` action should be modified to take a transaction id. The `generated_transaction_index` already has a `by_trx_id` index to allow the native action handler to find the appropriate `generated_transaction_object` from which it can get its `sender_id` to use with `apply_context::cancel_deferred`.

ATC:
Unit test verify that canceldelay action can cancel a delayed transaction using the transaction_id.  Verify that all tests pass on all platforms.

brianjohnson5972:
All unit tests passed on all platforms

PR #2046, opened 2018-04-02
Canceling a delayed transaction with transaction id

larryk85:
Modified canceldelay to now take a transaction id instead of a sender id.  This allows the user to cancel a delayed transaction with the transaction id from the transaction they pushed and not the auto generated sender id.

### DIFF
--- a/contracts/eosio.system/native.hpp
+++ b/contracts/eosio.system/native.hpp
@@ -190,9 +190,9 @@ namespace eosiosystem {
          }
          
          ACTION( SystemAccount, canceldelay ) {
-            uint32_t   sender_id;
+            checksum256   trx_id;
             
-            EOSLIB_SERIALIZE( canceldelay, (sender_id) )
+            EOSLIB_SERIALIZE( canceldelay, (trx_id) )
          };
          
          static void on( const canceldelay& ) {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -260,8 +260,17 @@ void apply_context::execute_deferred( deferred_transaction&& trx ) {
    } FC_CAPTURE_AND_RETHROW((trx));
 }
 
-void apply_context::cancel_deferred( uint128_t sender_id ) {
+void apply_context::cancel_deferred( const uint128_t& sender_id ) {
    results.deferred_transaction_requests.push_back(deferred_reference(receiver, sender_id));
+}
+
+void apply_context::cancel_deferred( const transaction_id_type& trx_id ) {
+   const auto& generated_transaction = controller.get_database().get_index<generated_transaction_multi_index>();
+   const auto& generated_index       = generated_transaction.indices().get<by_trx_id>();
+   const auto& itr                   = generated_index.lower_bound(trx_id);
+   FC_ASSERT (itr != generated_index.end() && itr->sender == config::system_account_name && itr->trx_id == trx_id,
+              "cannot cancel trx_id=${tid}, there is no deferred transaction with that transaction id",("tid", trx_id));
+   results.deferred_transaction_requests.push_back(deferred_reference(receiver, itr->sender_id, trx_id));
 }
 
 const contracts::table_id_object* apply_context::find_table( name code, name scope, name table ) {
--- a/libraries/chain/contracts/chain_initializer.cpp
+++ b/libraries/chain/contracts/chain_initializer.cpp
@@ -163,7 +163,7 @@ abi_def chain_initializer::eos_contract_abi(const abi_def& eosio_system_abi)
 
    eos_abi.structs.emplace_back( struct_def {
       "canceldelay", "", {
-         {"sender_id", "uint32"},
+         {"trx_id", "checksum256"},
       }
    });
 
--- a/libraries/chain/contracts/eosio_contract.cpp
+++ b/libraries/chain/contracts/eosio_contract.cpp
@@ -552,12 +552,13 @@ void apply_eosio_vetorecovery(apply_context& context) {
 
 void apply_eosio_canceldelay(apply_context& context) {
    auto cancel = context.act.data_as<canceldelay>();
-   const auto sender_id = cancel.sender_id.convert_to<uint32_t>();
+   //const auto sender_id = cancel.sender_id.convert_to<uint32_t>();
+   const auto& trx_id = cancel.trx_id;
    const auto& generated_transaction_idx = context.controller.get_database().get_index<generated_transaction_multi_index>();
-   const auto& generated_index = generated_transaction_idx.indices().get<by_sender_id>();
-   const auto& itr = generated_index.lower_bound(boost::make_tuple(config::system_account_name, sender_id));
-   FC_ASSERT (itr != generated_index.end() && itr->sender == config::system_account_name && itr->sender_id == sender_id,
-              "cannot cancel sender_id=${sid}, there is no deferred transaction with that sender_id",("sid",sender_id));
+   const auto& generated_index = generated_transaction_idx.indices().get<by_trx_id>();
+   const auto& itr = generated_index.lower_bound(trx_id);
+   FC_ASSERT (itr != generated_index.end() && itr->sender == config::system_account_name && itr->trx_id == trx_id,
+              "cannot cancel trx_id=${tid}, there is no deferred transaction with that transaction id",("tid", trx_id));
 
    auto dtrx = fc::raw::unpack<deferred_transaction>(itr->packed_trx.data(), itr->packed_trx.size());
    set<account_name> accounts;
@@ -577,7 +578,7 @@ void apply_eosio_canceldelay(apply_context& context) {
 
    FC_ASSERT (found, "canceldelay action must be signed with the \"active\" permission for one of the actors"
                      " provided in the authorizations on the original transaction");
-   context.cancel_deferred(sender_id);
+   context.cancel_deferred(itr->sender_id);
 }
 
 void apply_eosio_mindelay(apply_context& context) {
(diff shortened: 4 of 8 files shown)


larryk85:
Adding changes to recovery to get rid of dependence of get_next_sender_id.

libraries/chain/contracts/eosio_contract.cpp
```diff
@@ -466,7 +466,7 @@ void apply_eosio_postrecovery(apply_context& context) {
       .data = recover_act.data
    }, update);
 
-   uint32_t request_id = context.get_next_sender_id();
+   auto& request_id = context.trx_meta.id;
```
brianjohnson5972: Make const auto&

libraries/chain/contracts/chain_initializer.cpp
```diff
@@ -150,19 +150,19 @@ abi_def chain_initializer::eos_contract_abi(const abi_def& eosio_system_abi)
 
    eos_abi.structs.emplace_back( struct_def {
       "passrecovery", "", {
-         {"account", "account_name"},
```
brianjohnson5972: Why are we changing the type for account to be "name"?  I know that technically it is but I think this definition should match the definition provided in types.hpp, to make it easier to manually maintain these definitions that need to match.

larryk85: That was an oversight, I had changed it's field before and changed it back.  Will change that back to just name.

libraries/chain/contracts/chain_initializer.cpp
```diff
@@ -177,7 +177,7 @@ abi_def chain_initializer::eos_contract_abi(const abi_def& eosio_system_abi)
    eos_abi.structs.emplace_back( struct_def {
       "pending_recovery", "", {
          {"account",    "name"},
-         {"request_id", "uint32"},
+         {"request_id", "checksum256"},
```
brianjohnson5972: I think this should be transaction_id_type and we should add the typedef in contracts/eoslib/types.h and the type def in chain_initializer.  The more the multiple definitions match, the easier it will be to keep them consistent, the more people have to look up to verify they are consistent will mean people will just assume they are correct.

libraries/chain/contracts/eosio_contract.cpp
```diff
@@ -484,19 +484,18 @@ void apply_eosio_postrecovery(apply_context& context) {
       .data = recover_act.data
    }, update);
 
-   uint32_t request_id = context.get_next_sender_id();
-
+   auto& request_id = *reinterpret_cast<const uint128_t*>(context.trx_meta.id._hash);
```
brianjohnson5972: Make "const auto&"

libraries/chain/apply_context.cpp
```diff
@@ -303,10 +303,19 @@ void apply_context::execute_deferred( deferred_transaction&& trx ) {
    } FC_CAPTURE_AND_RETHROW((trx));
 }
 
-void apply_context::cancel_deferred( uint128_t sender_id ) {
+void apply_context::cancel_deferred( const uint128_t& sender_id ) {
    results.deferred_transaction_requests.push_back(deferred_reference(receiver, sender_id));
 }
-
+/*
+void apply_context::cancel_deferred( const transaction_id_type& trx_id ) {
```
brianjohnson5972: If the transaction_id_type's _hash is going to be used as the sender_id, remove this commented out code.

libraries/chain/contracts/eosio_contract.cpp
brianjohnson5972: Sorry, missed the const uint128_t, ignore my previous comment

libraries/chain/contracts/eosio_contract.cpp
```diff
@@ -505,11 +504,11 @@ void apply_eosio_postrecovery(apply_context& context) {
 
    context.execute_deferred(std::move(dtrx));
 
-
    auto data = get_abi_serializer().variant_to_binary("pending_recovery", record_data);
-   const uint64_t id = account;
+   const uint128_t id = account;
```
brianjohnson5972: Why is an account_name type being converted to an uint128_t to pass into db_find_i64 as a uint64_t?

libraries/chain/contracts/eosio_contract.cpp
```diff
@@ -559,20 +558,21 @@ void apply_eosio_vetorecovery(apply_context& context) {
    FC_ASSERT(maybe_recovery, "No pending recovery found for account ${account}", ("account", account));
    auto recovery = *maybe_recovery;
 
-   context.cancel_deferred(recovery["request_id"].as<uint32_t>());
+   context.cancel_deferred(recovery["account"].as<name>());
```
brianjohnson5972: Why did you change this from taking the "request_id" to "account"?

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -467,7 +467,8 @@ class apply_context {
       void execute_inline( action &&a );
       void execute_context_free_inline( action &&a );
       void execute_deferred( deferred_transaction &&trx );
-      void cancel_deferred( uint128_t sender_id );
+      void cancel_deferred( const uint128_t& sender_id );
+      //void cancel_deferred( const transaction_id_type& trx_id );
```
brianjohnson5972: remove commented out code

libraries/chain/apply_context.cpp
```diff
@@ -303,7 +303,7 @@ void apply_context::execute_deferred( deferred_transaction&& trx ) {
    } FC_CAPTURE_AND_RETHROW((trx));
 }
 
-void apply_context::cancel_deferred( uint128_t sender_id ) {
+void apply_context::cancel_deferred( const uint128_t& sender_id ) {
```
brianjohnson5972: Sorry I didn't think of this earlier, but why don't we make deferred_reference (and everything else) take a transaction_id_type and then we don't need to cast the hash to an uint128_t?

contracts/eosio.system/native.hpp
```diff
@@ -190,9 +190,9 @@ namespace eosiosystem {
          }
          
          ACTION( SystemAccount, canceldelay ) {
-            uint32_t   sender_id;
+            checksum256   trx_id;
```
brianjohnson5972: we should add a transaction_id_type typedef and use that for trx_id

libraries/chain/apply_context.cpp
larryk85: For non-generated deferred transactions, the user knows the sender id as they are the ones that set it, so they can cancel them easily.  But, for their deferred transactions they would have to create table entries and track deferred transactions for each deferred they send.  With leaving it as sender id, they can use an account or code, account or receiver, account pair and not have to track them.

contracts/eosio.system/native.hpp
larryk85: Missed it, will do.

libraries/chain/chain_controller.cpp
```diff
@@ -368,7 +369,8 @@ transaction_trace chain_controller::delayed_transaction_processing( const transa
    time_point_sec execute_after = head_block_time();
    execute_after += mtrx.delay;
    //TODO: !!! WHO GETS BILLED TO STORE THE DELAYED TX?
-   deferred_transaction dtrx(context.get_next_sender_id(), config::system_account_name, config::system_account_name, execute_after, trx);
+   fc::uint128_t _id(trx.id()._hash[3], trx.id()._hash[2]);
```
brianjohnson5972: If the decision is to stay with uint128_t, then we should make this into a method, rather than maintaining this pattern in 3 or more places, or why don't we just have the user use the deferred_transaction::sender_id, and just do the conversion when we construct the deferred_transaction?

libraries/chain/contracts/chain_initializer.cpp
```diff
@@ -177,7 +177,7 @@ abi_def chain_initializer::eos_contract_abi(const abi_def& eosio_system_abi)
    eos_abi.structs.emplace_back( struct_def {
       "pending_recovery", "", {
          {"account",    "name"},
-         {"request_id", "uint32"},
+         {"request_id", "sender_id_type"},
```
brianjohnson5972: If it is decided to keep using uint128_t and sender_id_type, then change all of these related sender_id params to sender_id_type, so that we won't have errors like the:

context.cancel_deferred(record["request_id"].as<sender_id_type>())

you had to fix below, since the typedef would be what was changed.

libraries/chain/chain_controller.cpp
larryk85: So, the changes I have are for generated deferred transactions and not user produced deferred.  I like changing this just a function,  it just seems more natural for the user to just give a transaction id for the transaction they had originally sent.

libraries/chain/apply_context.cpp
brianjohnson5972: I'm saying, why have a deferred_tansaction have a sender and sender_id (which is a portion of the transaction_id), why not have sender and transaction_id, and then a generated_transaction_object just stores the sender and transaction_id (and not the sender_id, which is just redundant information from the transaction_id).  Basically everywhere we have just a sender_id, we instead have a transaction_id, and everywhere we have both, we just have the transaction_id.

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -4,7 +4,7 @@
  */
 #pragma once
 #include <eosio/chain/types.hpp>
-
+#include <eosio/chain/config.hpp>
```
brianjohnson5972: I don't see anything new in this file that would require a new include, is it needed, or was this accidentally brought in?

contracts/eosiolib/types.h
```diff
@@ -59,6 +59,8 @@ struct fixed_string16 {
    char str[16];
 };
 
+typedef struct checksum256 transaction_id;
```
brianjohnson5972: For consistency, this should match the type name in the non-contract code, so make this "transaction_id_type".

libraries/chain/apply_context.cpp
larryk85: I get that and agree with your statement, but we only have redundant information for a subset of all deferred transactions.  And I still think that allowing the contract writer the ability to write their own generator for sender id or use there account is more beneficial than any confusion it might raise.

brianjohnson5972:
#2002

taokayan:
This should be passed because #2002 is passed.

PR #2051, opened 2018-04-03
ABI Generator: Add detection of EOSIO_ABI macro

elmato:
Two more test cases (case25, case26) added to:
https://github.com/elmato/abi-generator-atc


tests/tests/abi_tests.cpp
```diff
@@ -390,15 +390,32 @@ struct abi_gen_helper {
...
+      include_param, boost_include_param, stdcpp_include_param,
+      stdc_include_param, pfr_include_param }
+    );
+    FC_ASSERT(res == true);
+
+    res = runToolOnCodeWithArgs(new generate_abi_action(false, opt_sfs, "", output, contract, actions), source,
+      {"-fparse-all-comments", "--std=c++14", "--target=wasm32", "-ffreestanding", "-nostdlib",
+      "-nostdlibinc", "-fno-threadsafe-statics", "-fno-rtti",  "-fno-exceptions",
+      include_param, boost_include_param, stdcpp_include_param,
+      stdc_include_param, pfr_include_param }
```
heifner: Since this vector of args is the same here and above, I would create the vector above this and pass it in so that it is easier to maintain.

PR #2037, opened 2018-04-02
Updated readme to reflect updated config folder locations

dskvr:
Resolves #2036

PR #2056, opened 2018-04-03
Made corrections for the new data and config folders.

rbw3821:
Also removed a lot of redundant text.  There were separate sections saying basically the same thing, that are no longer needed because the automated build script handles the differences within it.  There were still some old references to the public testnet that have been pulled out for now with reference to the dawn-2.x readme.  When the new details are ready, that section can be recreated.


dskvr:
Some of this is a duplicate of #2037

PR #2057, opened 2018-04-03
fix bug when run using docker-compose

noprom:
1. Specify the required docker-compose version;
2. Create volume before run docker-compose;
3. Add hostname to docker containers.


Docker/README.md
```diff
@@ -43,18 +44,20 @@ curl http://127.0.0.1:8888/v1/chain/get_info
 ## Start both nodeos and walletd containers
```
heifner: Please replace all `walletd` references with `keosd`.

noprom: OK, I'll do that.

Issue #2008, opened 2018-03-31
nodeos / cleos usability improvements


bytemaster:
The user experience for developers attempting to setup and run a new node and then interact with it from the command line is very poor, we need to improve it.

- [x] JSON Dumps of Transactions are not acceptable
- [x] Console output from running transactions should be visible
- [x] Account history should be readable 
- [x] hex blobs of action data are not readable 
- [x] configuration files should not need to be edited by default
- [x] eosio dev key should be available to all wallets without import step
- [x] producer plugin should be enabled by default
- [ ] cleos should start keos as a background daemon if it is not currently running so that users don't need to think about this background process. - will be completed by #2094
- [x] permission command line flags need to make "@active" optional 

```
./nodeos -e -p eosio --plugin eosio::wallet_api_plugin --plugin eosio::chain_api_plugin --plugin eosio::account_history_api_plugin
```

bytemaster:
Default system contract does not set an ABI and it needs to, setting bios fixes it.

zorba80:
One item is left:
_cleos should start keos as a background daemon if it is not currently running so that users don't need to think about this background process._
I will create a separate task for it.

zorba80:
ATC for item
_eosio dev key should be available to all wallets without import step_

- Unit tests check that the default eosio key is imported when a wallet is created.
- Run nodeos without --eosio-key option, create a wallet, check that the wallet includes the key 5KQwrPbwdL6PhXujxW37FSSQZ1JiwsST4cqQzDeyXtP79zkvFD3
- Rerun nodeos with the wallet_api_plugin option --eosio-key passing a different key as an argument, create a new wallet and verify that the wallet includes the passed key.

brianjohnson5972:
Performed tests with and without --eosio-key and tests passed.

PR #2064, opened 2018-04-03
Import eosio key automatically when a wallet is created - #2008

zorba80:


plugins/wallet_plugin/wallet_manager.cpp
```diff
@@ -40,16 +40,17 @@ std::string wallet_manager::create(const std::string& name) {
...
    }
-
+   
    wallet_data d;
    auto wallet = make_unique<wallet_api>(d);
    wallet->set_password(password);
    wallet->set_wallet_filename(wallet_filename.string());
    wallet->unlock(password);
-   wallet->save_wallet_file();
+   static const std::string eosio_wif_key("5KQwrPbwdL6PhXujxW37FSSQZ1JiwsST4cqQzDeyXtP79zkvFD3");
```
heifner: What do you think of making this a defaulted argument to the wallet plug so a user can specify a different key without re-compiling?

wanderingbort: I like this idea

zorba80: Added that

plugins/wallet_plugin/wallet_manager.cpp
```diff
@@ -40,16 +40,16 @@ std::string wallet_manager::create(const std::string& name) {
...
    }
-
+   
    wallet_data d;
    auto wallet = make_unique<wallet_api>(d);
    wallet->set_password(password);
    wallet->set_wallet_filename(wallet_filename.string());
    wallet->unlock(password);
-   wallet->save_wallet_file();
+   wallet->import_key(eosio_key);
```
wanderingbort: was `save_wallet_file()` no longer necessary here?

wanderingbort: nevermind I see that is handled in the wallet api for import_key

Issue #2001, opened 2018-03-31
Transaction delay should be based on a new delay field in transaction header


arhag:
Related to #1022.

Currently a transaction signer will not know with certainty whether the transaction they signed will (assuming it does not expire) be immediately executed in the block that it is included in or be delayed for some time before (maybe) being executed in a later block. It would be nice for the transaction signer(s) to know with certainty that a transaction they sign will not be delayed (more than the time until expiration of the transaction) unless it ends up never being included in a block because it expires.

More critically, the delay that will be used can only be determined after running through `check_authorization` which requires the set of signing public keys as an input (in addition to the state of the permission graph at the point in which the transaction was executed). This prevents the signatures from being pruned since they would still be needed by a validator (even though they trust the block) just to recover the delay value in order to achieve the same database state transitions as every other node. 

This particular requirement of being able to apply blocks one trusts without needing pruned data could be resolved (with respect to the new delayed transaction changes) by requiring the delay value to be placed into the `transaction_receipt` (at the cost of a few extra bytes per delayed transaction in the `signed_block_summary`), but both this requirement and the earlier one of providing some delay certainty to transaction signers could be achieved by requiring the delay value to be included in the transaction header. If a varint is used, then transactions which are not delayed (which we expect to be the vast majority of them) only require one extra byte in the header.

With this change, the `check_authorization` computation would still need to calculate a delay value. For the transaction to be valid, that calculated delay value would need to be no greater than the delay value in the transaction header. But the delayed transaction would use a delay based on the value in the transaction header. This also means that the current approach of increasing the delay of a transaction beyond the minimum required value via the special `mindelay` context-free action would no longer be necessary.

Replaying nodes can then simply use the delay value in transaction header without needing to run the `check_authorization` computation (which they could not do anyway if the signatures had been pruned by then).

With the original delay model, there was an open question regarding who to bill for the additional database memory used to store the delayed transaction until it was either scheduled and executed or dropped. The answer we were most likely going to go with was to bill the authorizer with the highest delay. But once again, finding out which account that actually is requires running through the `check_authorization` computation which nodes cannot be expected to do on replay of irreversible blocks since the necessary signatures may have already been pruned. Now, we can simply follow the similar model described in #1 and #2 (for figuring out which account to bill for CPU and network bandwidth respectively): by default bill the actor of the first authorization of the first context-aware action of the delayed transaction; but, allow this default account to bill to be overridden with a special context-aware action (that is a task left for another issue).

[...]

Issue #2000, opened 2018-03-31
Preferential billing for pruned data (and friendlier default network usage billing)


arhag:
Transaction headers include a field `net_usage_words` which is the number of 8-byte words to bill for network bandwidth (not counting the additional billing per read and write locks). Currently, there is no preferential rate for the data that can be pruned (signatures and context-free data). Similar to CPU billing, validation checks would require that the committed `net_usage_words` is greater than or equal to an effective calculated size which is `size(compressed unsigned transaction) + factor * (size(signatures) + size(compressed context-free data)` where the factor is a configurable chain parameter less than one.

(Question: Should the additional billing for read and write locks be added into the effective size so that the `net_usage_words` fields directly determines the total amount to bill for? How would that affect deferred transactions generated in contract code?)

Furthermore, similar to #1999, by default net usage would only be billed to the actor of the first authorization of the first context-aware action of the transaction (rather than all authorizing actors of the transaction as it is today). The default account to bill could be overridden with a special context-free action, but that is a task left for another issue.

SheldonHH:
net_usage_words > size(cut) + factor * (size(sig)+size(c-cfd))

wanderingbort:
We are going with transaction having an optionally declared "max"  in line with #1999

Preferential billing is still being debated.

arhag:
The `max_net_usage_words` has already been implemented.

What remains is to implement preferential billing of the prunable data in a similar way that we already have preferential billing of the CPU bandwidth of context-free action.

arhag:
Discount billing for prunable data is already in latest slim

Issue #1999, opened 2018-03-31
Transactions should commit to total CPU usage? (and friendlier default CPU usage billing)


arhag:
Currently, transaction headers include a field `context_free_kilo_cpu_usage` which holds the number of kilo CPU units to bill for processing signatures and context-free actions. The number of CPU units to bill for processing context-aware actions is currently not explicitly specified in the transaction and is instead calculated by running the context-aware actions with objective instruction counting.

This issue is to track the investigation and, if necessary, implementation of the following change in design:

* The field `context_free_kilo_cpu_usage` would be renamed to just `kilo_cpu_usage` and instead be the exact number of kilo CPU units to bill for the transaction. This value acts as the threshold value that the objectively calculated CPU usage of the transaction should not exceed. The objectively calculated CPU usage of the transaction would still have preferential rates for context-free computations (computing context-free actions, signature verification, and authorization checks).

* Furthermore, by default the cpu usage would only be billed to the actor of the first authorization of the first context-aware action of the transaction (rather than all authorizing actors of the transaction as it is today). The default account to bill could be overridden with a special context-free action, but that is a task left for another issue.

Requiring transactions to commit an upper bound on context free CPU usage allowed nodes to replay irreversible blocks while skipping over the computation of context-free actions and without needing pruned data (such as context-free data and signatures). The change described in this issue would allow further optimization of the replay performance of nodes by allowing even the overhead of objective instruction counting to be removed.

In addition, committing to an upper bound on total CPU usage within the transaction gives assurances to the transaction signers (in particular the account which will be billed for CPU usage) that they will definitely not be billed more than that committed amount if the transaction is successfully executed. They do not have to worry that the contract code that the transaction's actions will be delivered to (which can sometimes be hard to determine prior to execution because of `require_notify`, `send_inline`, and `send_context_free_inline`) will eat up a lot more of their CPU bandwidth than they expected. 

The downside of this approach is that the payer is billed for the full committed amount of CPU units even if their transaction ends up actually using less.  Their client side tools would also need to estimate this upper bound; while CPU usage could be calculated perfectly for signature verification and context-free actions (at least if you exclude those dispatched inline by context-aware actions), it can sometimes be difficult to estimate the CPU usage for the context-aware actions.

[...]

Issue #1972, opened 2018-03-29
Replace transaction_mroot with transaction_schedule_mroot


arhag:
Right now the block header (which is signed by producers) does not commit to the transaction schedule defined in in the vector of `region_summary` (see field `region` in `signed_block_summary`). 

We may be indirectly committing to the schedule (at least partially) via the `action_mroot`, but there are potential malleability vulnerabilities involved with that approach which are hard to analyze. Furthermore, a light client cannot validate whether the transaction schedule of the `signed_block` was actually signed by the producer because it is not possible to validate `action_mroot` without actually apply the block (which requires state).

The most straightforward way for a producer to commit to a particular schedule is to commit to the digest of the vector of `region_summary` in the block header. Instead of including another digest into the block header, we can instead concatenate that digest with the `transaction_mroot`, take a hash of that, and put that resulting digest in the block header in place of the existing `transaction_mroot`. This way, we do not increase the size of the block header and still provide small proofs (only one extra digest needed in the proof) to anyone who wants to verify that a particular transaction was included in the `input_transactions` of some block.

So to recap, we would have the following digests in the `block_header`:

- `previous` remains the same. It is still the block id of the previous block in the blockchain.
- `block_mroot` remains the same. This is something that can be validated without database state, but the entire blockchain (or at least block ids) up to the block that one is validating is necessary.
- `action_mroot` remains the same. However, perhaps the data we currently commit to in the leaf nodes of the merkle tree may need to be modified if they do not allow us to prove all the things we care about proving; but that is a task for another issue). Validating `action_mroot` require applying the entire block and therefore requires the entire database state.
- `transaction_schedule_mroot` which is Hash(Hash(`regions`) || `transaction_mroot`). `transaction_mroot` is the root of the Merkle tree over `input_transactions`.

A light client validator with access to just the chain of `signed_block_header`s would be able to validate all fields of the head `signed_block_header` of the chain except for `transaction_schedule_mroot` and `action_mroot`. A validator with access to the entire blockchain would be able to validate (without actually applying any blocks nor having access to database state) all fields of the `signed_block_header` of the head block of the chain except for `action_mroot`.

arhag:
There have been changes to the definition of `transaction_mroot` (or `transaction_schedule_mroot`) described at the top of this issue.

The objective of `transaction_mroot` (will not be renamed to `transaction_schedule_mroot` anymore) is still the same: 
* to commit to the set of input transactions included in the block;
[...]

PR #2084, opened 2018-04-04
Fixes to delayed input transaction processing, and transaction header usage fields now specify upper limits on cpu and net bandwidth

arhag:
Changes in this PR:
 * Now if `packed_transaction` is going to compress the unsigned transaction, it will also separately compress the `context_free_data` (if present) as well.
 * Change `kcpu_usage` field in transaction header to `max_kcpu_usage` which now specifies an upper limit (in kibi cpu instruction units) on the total cpu usage that can be billed for the transaction. A value of 0 for this field means no upper bound specified. See [this comment](https://github.com/EOSIO/eos/issues/1999#issuecomment-378439255) in #1999.
 * Change `net_usage_words` field in transaction header to `max_net_usage_words` which now specifies an upper limit (in 8-byte words) on the total net usage that can be billed for the transaction. A value of 0 for this field means no upper bound specified. See [this comment](https://github.com/EOSIO/eos/issues/2000#issuecomment-378444025) in #2000.
 * cleos by default sets both `max_kcpu_usage` and `max_net_usage_words` to 0. This can be changed with the `--max-cpu-usage` and `--max-net-usage` flags.
 * Fixed delayed input transaction processing to use `delay_sec` as the delay time and to enforce that `delay_sec` is not smaller than the delay time imposed by authorization checking. #2001 (Additional unit tests, not included in this PR, to test the new semantics of `delay_sec` are needed.)
 * Removed mindelay action since the `delay_sec` field makes it unnecessary.
 * apply_block and push_transaction now respect the skip_authority_check flag.

### DIFF
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -363,27 +363,12 @@ transaction_trace chain_controller::delayed_transaction_processing( const transa
 
    const auto& trx = mtrx.trx();
 
-   // add in the system account authorization
-   action for_deferred = trx.actions[0];
-   bool found = false;
-   for (const auto& auth : for_deferred.authorization) {
-      if (auth.actor == config::system_account_name &&
-          auth.permission == config::active_name) {
-         found = true;
-         break;
-      }
-   }
-   if (!found)
-      for_deferred.authorization.push_back(permission_level{config::system_account_name, config::active_name});
-
-   apply_context context(*this, _db, for_deferred, mtrx); // TODO: Better solution for getting next sender_id needed.
-
    time_point_sec execute_after = head_block_time();
    execute_after += mtrx.delay;
 
    // TODO: update to better method post RC1?
    account_name payer;
-   for(const auto& act : mtrx.trx().actions ) {
+   for(const auto& act : trx.actions ) {
       if (act.authorization.size() > 0) {
          payer = act.authorization.at(0).actor;
          break;
@@ -392,7 +377,15 @@ transaction_trace chain_controller::delayed_transaction_processing( const transa
 
    FC_ASSERT(!payer.empty(), "Failed to find a payer for delayed transaction!");
 
-   deferred_transaction dtrx(transaction_id_to_sender_id( trx.id() ), config::system_account_name, payer, execute_after, trx);
+   auto sender_id = transaction_id_to_sender_id( mtrx.id );
+
+   const auto& generated_index = _db.get_index<generated_transaction_multi_index, by_sender_id>();
+   auto colliding_trx = generated_index.find(boost::make_tuple(config::system_account_name, sender_id));
+   FC_ASSERT( colliding_trx == generated_index.end(),
+              "sender_id conflict between two delayed transactions: ${cur_trx_id} and ${prev_trx_id}",
+              ("cur_trx_id", mtrx.id)("prev_trx_id", colliding_trx->trx_id) );
+
+   deferred_transaction dtrx(sender_id, config::system_account_name, payer, execute_after, trx);
    FC_ASSERT( dtrx.execute_after < dtrx.expiration, "transaction expires before it can execute" );
 
    result.deferred_transaction_requests.push_back(std::move(dtrx));


plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
```diff
@@ -120,9 +120,9 @@ struct txn_test_gen_plugin_impl {
 
          trx.actions.emplace_back(vector<chain::permission_level>{{creator,"active"}}, contracts::newaccount{creator, newaccountC, owner_auth, active_auth, recovery_auth});
          }
-      
+
          trx.expiration = cc.head_block_time() + fc::seconds(30);
-         trx.net_usage_words = 500;
+         trx.max_net_usage_words = 500;
```
wanderingbort: I think we should just 0 these out, the txn_test_gen_plugin isn't concerned with abusive contracts.

programs/cleos/main.cpp
```diff
@@ -145,8 +145,8 @@ bool   tx_dont_broadcast = false;
 bool   tx_skip_sign = false;
 bool   tx_print_json = false;
 
-uint32_t tx_cf_cpu_usage = 0;
-uint32_t tx_net_usage = 0;
+uint32_t max_cpu_usage = 0;
```
wanderingbort: while the `tx_` prefix isn't pretty it is conventional with the parameters above like `tx_skip_sign` and `tx_print_json` we should maintain the convention or change it universally.

tests/chain_tests/block_tests.cpp
```diff
@@ -223,14 +223,14 @@ BOOST_AUTO_TEST_CASE(transaction_expiration) {
                               });
       trx.ref_block_num = static_cast<uint16_t>(chain.control->head_block_num());
       trx.ref_block_prefix = static_cast<uint32_t>(chain.control->head_block_id()._hash[1]);
-      trx.net_usage_words.value = 4096;
+      trx.max_net_usage_words.value = 4096;
```
wanderingbort: as this test is for expiration, we should zero out this max

tests/chain_tests/block_tests.cpp
```diff
@@ -78,7 +78,7 @@ BOOST_AUTO_TEST_CASE( push_invalid_block ) { try {
    trx.actions.emplace_back(std::move(on_block_act));
    trx.set_reference_block(chain.control->head_block_id());
    trx.expiration = chain.control->head_block_time() + fc::seconds(1);
-   trx.kcpu_usage = 2000; // 1 << 24;
+   trx.max_kcpu_usage = 2000; // 1 << 24;
```
wanderingbort: unless this is used as the failure case for the invalid block we should zero it

tests/chain_tests/block_tests.cpp
```diff
@@ -252,7 +252,7 @@ BOOST_AUTO_TEST_CASE(invalid_tapos) {
                            });
    trx.ref_block_num = static_cast<uint16_t>(chain.control->head_block_num() + 1);
    trx.ref_block_prefix = static_cast<uint32_t>(chain.control->head_block_id()._hash[1]);
-   trx.net_usage_words.value = 4096;
+   trx.max_net_usage_words.value = 4096;
```
wanderingbort: default it/zero it

arhag:
Yes, there is more work to do which I will get to in another PR and with my work on #1972. 
Specifically, I will be adding `kcpu_usage` and `net_usage_words` to the transaction receipt which will resolve the replay issue.

By the way, with my latest commits I also added the following additional fixes: 
 * `delay_sec` in generated transactions is now treated consistently with how it is treated for input transactions. And with that change, we no longer need the `delay_until`/`execute_after` parameter in the `send_deferred` intrinsic.
 * Check for conflicts in the `sender_id` derived from the transaction id of delayed transactions so that existing delayed transactions with colliding sender_id are not silently cancelled; instead the newly processed delayed transaction with conflicting `sender_id` is rejected. When we eventually go parallel, we will need to check for `sender_id` conflicts between shards.

taokayan:
Is the any ATC for this? And is the related test case available?

taokayan:
max_cpu_usage tests: pull request #2172 

max_net_usage tests: pull request #2268 

delayed input transaction tests:
  Test suite "delay_tests" has passed with:
    9 test cases out of 9 passed
    286 assertions out of 286 passed

PR #2090, opened 2018-04-04
adding asio

roelandp:
Please note the prev commit should also have added asio! 
(as far as I understand from the patch from websocketpp: https://github.com/zaphoyd/websocketpp/commit/0bb33e4bca4ccc42a36aa2321e4fb97f2562e519) 
`#include <websocketpp/common/asio.hpp>`


jgiszczak:
Redundant.

PR #2095, opened 2018-04-04
Feature/ricardian

larryk85:
Added support for ricardian contracts.  Stubbed out for contracts and clauses with our contracts to be filled in later. Removed setabi, supporting structs from eosio.system, we do not need to define this action and is wasteful.


gleehokie:
Reviewed with @bytemaster, looks good

PR #2106, opened 2018-04-04
Fix for #2105

### LINKED ISSUE #2105: Generate Block and Push Block desync on CPU usage
The root of the problem is that we are not propagating the number of signatures into the resource usage calculation during `transaction_trace chain_controller::push_transaction(const packed_transaction&);`

A fix is already in bound.

wanderingbort:
derive signature count from signatures not recovered keys.  Also do it the same whether in push_transaction or apply_block EOSIO/eos#2105

This resolves #2105


taokayan:
This pull request is verified in the source code and in the unitest case of pull request #2172

PR #2121, opened 2018-04-04
specialized operators << >> for vector<char> #2118

moskvanaft:
small improvement for serialization/deserialization

### DIFF
--- a/contracts/eosiolib/datastream.hpp
+++ b/contracts/eosiolib/datastream.hpp
@@ -254,11 +254,27 @@ DataStream& operator >> ( DataStream& ds, std::array<T,N>& v ) {
    return ds;
 }
 
+template<typename DataStream>
+DataStream& operator << ( DataStream& ds, const vector<char>& v ) {
+   ds << unsigned_int( v.size() );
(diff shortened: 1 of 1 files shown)


contracts/eosiolib/datastream.hpp
```diff
@@ -254,6 +254,13 @@ DataStream& operator >> ( DataStream& ds, std::array<T,N>& v ) {
    return ds;
 }
 
+template<typename DataStream>
```
brianjohnson5972: Could we use std::enable_if_t<std::is_fundamental<T>::value == true, int> = 0 and size_of(T) to make this optimization be for all primitives?

wanderingbort: https://github.com/EOSIO/eos/blob/c6311f255c56824149af203362d8624fc5ce934f/contracts/eosiolib/datastream.hpp#L365-L377

We already do

brianjohnson5972: I meant for vector<T>, to optimize more than just vector<T>

PR #2126, opened 2018-04-05
Changes to transaction receipt and transaction_mroot

arhag:
The CPU usage (`kcpu_usage`) and network usage (`net_usage_words`) that was billed for a transaction are now included in the `transaction_receipt`.

Delayed transactions are now also billed at the time they are included into the block. Previously they were only billed when they actually executed. That meant that their network billing only covered the cost of shard locks but not the much bigger cost of storing the `packed_transaction` in the blockchain. Also, the cost of the signature validation (which is billed at the time the input transaction is included in the block, even if it is delayed and executes later) was not being billed. This PR fixes those issues. 

Two things are worth mentioning about the current implementation of the billing delayed transactions:

The first is that the `max_net_usage_words` field of the transaction applies as an upper bound (assuming it is not 0) at both stages of processing. That means it needs to be large enough to cover the `packed_transaction` size at the time it is included in the block, but that same upper limit then applies for network billing at the time it actually executes (at which point network usage will not include the `packed_transaction` size component and will therefore just be a function of the fixed overhead and the per shard lock cost).

The second is that `max_kcpu_usage` also similarly applies at both stages of processing the delayed transaction. In practice, if it is set (assuming it is set to be non-zero at all) to be large enough to cover execution, it will likely be more than enough to cover the cost of signature verification + fixed overhead at the time of inclusion into the block. But it is important to note that it is not an upper bound on the total computation over the life of the delayed transaction as it would be for an immediately-executing input transaction.

Finally, this PR changes the way `transaction_mroot` is calculated and in doing so resolves issue #1972. You can read more about the new definition of `transaction_mroot` in [this comment](https://github.com/EOSIO/eos/issues/1972#issuecomment-378815285).


libraries/chain/block_trace.cpp
```diff
@@ -11,29 +11,53 @@
...
-      cpu_usage = 0;
-      action_roots.reserve(transaction_traces.size() * GUESS_ACTS_PER_TX);
-      for (const auto& tx :transaction_traces) {
-         for (const auto& at: tx.action_traces) {
-            digest_type::encoder enc;
+               fc::raw::pack(enc, at.receiver);
+               fc::raw::pack(enc, at.act.account);
+               fc::raw::pack(enc, at.act.name);
+               fc::raw::pack(enc, at.act.data);
+               fc::raw::pack(enc, at.region_id);
```
wanderingbort: Some of this data ( like `region_id` `cycle_index`) is now duplicated between `action_trace` and `transaction_trace`.  

It seems like we can pull that data from the `transaction_trace` here and remove it from `action_trace` so that the trace structure is more concise. 

We should probably consider making sure things like this that are 64bit in the transaction merkle are 64bit in the action merkle as well.

libraries/chain/block_trace.cpp
```diff
@@ -42,7 +66,7 @@ namespace eosio { namespace chain {
       for(const auto& rt: region_traces ) {
```
wanderingbort: Since we are here and doing this, do we want to add in an indirection for region parallelism?

PR #2135, opened 2018-04-05
Enhancement: Better error message for set contract in cleos

larryk85:
If you give cleos an incorrect folder for set contract, the error message states that it was unable to parse wasm.  Added a simple check for whether a file was found and assert with the proper error message if not.


programs/cleos/main.cpp
```diff
@@ -834,9 +835,10 @@ int main( int argc, char** argv ) {
       {
          abiPath = (cpath / (cpath.filename().generic_string()+".abi")).generic_string();
       }
-
+      
+      
       fc::read_file_contents(wastPath, wast);
-
+      FC_ASSERT( wast != string(), "no wast file found ${f}", ("f", wastPath) );
```
wanderingbort: change to `!was.empty() `?

wanderingbort: err `wast` not `was`

larryk85: Got it.

gleehokie:
Jenkins issues on the build, MacOS passed, merged

PR #2145, opened 2018-04-05
Fix for bad threshold on producers account

### LINKED ISSUE #2141: Producer Account Threshold is set wrong.
Bart Wyatt, [Apr 5, 2018 at 11:32:58 AM]:
...producers account properly lists the producer accounts BUT expects a threshold of 14

which cannot be achieved until you have at least 14 producers

wanderingbort:
this fixes #2141


andriantolie:
#2141 passes

Issue #2149, opened 2018-04-05
Read default eosio key from config.ini in addition to command line


zorba80:

andriantolie:
Verified.

Used --eosio-key parameter when starting the wallet and create a new wallet. The new wallet has the key specified after --eosio-key imported by default.

PR #2150, opened 2018-04-05
Read default eosio key from config.ini in addition to command line #2149

zorba80:


andriantolie:
The related issue #2149 passes the test

PR #2164, opened 2018-04-05
Fix #2140 - Check that producers are unique and corresponding account…

### LINKED ISSUE #2140: set_active_producers does not validate accounts exist, but it should.

zorba80:
…s exist


andriantolie:
The related unit test is "producer_schedule_tests/verify_producers", and all of the assertion pass.

PR #2276, opened 2018-04-09
Specify -DBOOST_ROOT=${BOOST_ROOT} in call to cmake

banshee:
Two other variables are passed explicitly to cmake (WASM_ROOT and OPENSSL_ROOT_DIR) in eosio_build.sh, and I'm wondering if passing BOOST_ROOT directly would solve some of the boost build issues I'm seeing on Telegram.  

Also, there's an old boost bug (see https://cmake.org/pipermail/cmake/2011-November/047533.html) that makes FindBoost ignore the environment variable, so I think if you have a version of boost installed with that bug the system you're building on will use the wrong boost libraries without this fix.


pacificcode:
After further investigation it looks as though this PR is not currently needed.

PR #2287, opened 2018-04-10
Reassign storm

pmesnier:
#2285, fix for newly uncovered synchronizer issues.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -263,9 +263,12 @@ namespace eosio {
        */
       chain::signature_type sign_compact(const chain::public_key_type& signer, const fc::sha256& digest) const;
 
-      static const fc::string logger_name;
-      static fc::logger logger;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -935,12 +941,13 @@ namespace eosio {
          break;
       }
       default:
+         fc_dlog(logger, "sending empty request but not calling sync wait on ${p}", ("p",peer_name()));
          enqueue( ( sync_request_message ) {0,0} );
       }
    }
 
    void connection::cancel_fetch() {
```
wanderingbort: It looks like the only call to `cancel_fetch` was removed in favor of an exact copy of this implementation (see line 1059) . Should that line call this method, or should this method be removed?

pmesnier: I removed it. I also removed the redundant logger instances that I commented out in this pr.

Issue #1824, opened 2018-03-25
Add unit tests for WASM memory OOB access


spoonincode:
There is (somewhat surprisingly) no unit tests that check for accessing WASM memory that is out of bounds. While there are ample unit tests that do make sure the invoker wrapper guards OOB memory, we really need to add some WASM only unit tests.

taokayan:
There's a memory wrapping bug which will probably crash the blockchain if triggered.
pull request #2311 created for the fix and adding unittest cases.

spoonincode:
That's a really good find. But I'd also like to see us make sure that pure WASM memory load/stores are also caught correctly before closing this issue. (this can be done in C++ by dereferencing an OOB pointer and load/storing it)

taokayan:
hi, I've committed another fix for the memory boundary check in the wasm case. also added unit test cases for load/store operations with different sizes.

This is my findings:
in the wasm-interpreter case: memory boundary checks for loads/stores are already performed in visitLoad() before calling the actual load() function:

`      Flow visitLoad(Load *curr) {
        NOTE_ENTER("Load");
        Flow flow = this->visit(curr->ptr);
        if (flow.breaking()) return flow;
        NOTE_EVAL1(flow);
        auto addr = instance.getFinalAddress(curr, flow.value);
        auto ret = instance.externalInterface->load(curr, addr);
        NOTE_EVAL1(addr);
        NOTE_EVAL1(ret);
        return ret;
      }

  template <class LS>
  Address getFinalAddress(LS* curr, Literal ptr) {
    auto trapIfGt = [this](uint64_t lhs, uint64_t rhs, const char* msg) {
      if (lhs > rhs) {
        std::stringstream ss;
        ss << msg << ": " << lhs << " > " << rhs;
        externalInterface->trap(ss.str().c_str());
      }
    };
    Address memorySizeBytes = memorySize * Memory::kPageSize;
    uint64_t addr = ptr.type == i32 ? ptr.geti32() : ptr.geti64();
    trapIfGt(curr->offset, memorySizeBytes, "offset > memory");
    trapIfGt(addr, memorySizeBytes - curr->offset, "final > memory");
    addr += curr->offset;
    trapIfGt(curr->bytes, memorySizeBytes, "bytes > memory");
    trapIfGt(addr, memorySizeBytes - curr->bytes, "highest > memory");
    return addr;
  }`

[...]

PR #2311, opened 2018-04-11
1824 fix memory wrapping bug and add unit test cases

taokayan:
fix memory wrapping case for #1824 and add the related test case.


taokayan:
fix the memory bound check in the wasm case. add more test cases for load & store.

wanderingbort:
api_tests/memory_test is still failing when using the WAVM backend.  You can test this locally by running chain_test with the --wavm flag

taokayan:
test passed:

    Test case "api_tests/memory_tests" has passed with:
      23 assertions out of 23 passed

PR #2318, opened 2018-04-11
feat(docker): add mongo_db_plugin support to docker

Akagi201:
Make Docker image support mongo_db_plugin.
Reference the options from the script <https://github.com/EOSIO/eos/blob/master/scripts/eosio_build_ubuntu.sh>

### DIFF
--- a/Docker/Dockerfile
+++ b/Docker/Dockerfile
@@ -3,7 +3,7 @@ FROM eosio/builder as builder
 RUN git clone -b master --depth 1 https://github.com/EOSIO/eos.git --recursive \
     && cd eos \
     && cmake -H. -B"/tmp/build" -GNinja -DCMAKE_BUILD_TYPE=Release -DWASM_ROOT=/opt/wasm -DCMAKE_CXX_COMPILER=clang++ \
-       -DCMAKE_C_COMPILER=clang -DCMAKE_INSTALL_PREFIX=/tmp/build  -DSecp256k1_ROOT_DIR=/usr/local \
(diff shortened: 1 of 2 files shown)


masato25:
HI, I got a problem when i build docker image. those error message are below:
```
CMake Error at plugins/mongo_db_plugin/CMakeLists.txt:22 (find_package):
  By not providing "Findlibbsoncxx-static.cmake" in CMAKE_MODULE_PATH this                                                                                                                  
  project has asked CMake to find a package configuration file provided by                                                                                                                  
  "libbsoncxx-static", but CMake did not find one.

  Could not find a package configuration file provided by "libbsoncxx-static"
  with any of the following names:

    libbsoncxx-staticConfig.cmake
    libbsoncxx-static-config.cmake

  Add the installation prefix of "libbsoncxx-static" to CMAKE_PREFIX_PATH or
  set "libbsoncxx-static_DIR" to a directory containing one of the above
  files.  If "libbsoncxx-static" provides a separate development package or
  SDK, be sure it has been installed.
```
I found the problem was the option -> "-DBUILD_MONGO_DB_PLUGIN=true". when I removed this, then image can built succeeded. I'm not sure do this only happen on my machine or it's a problem that need fix it.

Akagi201:
@masato25 hello, You should update/rebuild your eosio/builder, then build the docker.

masato25:
@Akagi201 thanks for your response.

yes, if i run `docker-compose build builder` first, then do `docker build . -t eosio/eos`. the problem is solved.

mosjin:
@masato25Your ideas also works for me, but whether there is any bad effect ...?!

Issue #2408, opened 2018-04-15
p2p improve handling of block/txn acceptance failures


pmesnier:
Blocks are now routinely shipped as summaries only. However there are some implicit transactions indicated by the summary that cannot be resolved by the receiver. This causes the receiver to enter a syncing mode in which it broadcasts an appeal for a full copy of the block in question. A better solution is to assume the sender is trying to send a valid block, and so in this case the receiver should treat this as a block notification, and request a full block right away. Even better would be to have the sender detect that the block contains such transactions and just sends the full block at that point 

Also adding a hook to retrieve the id of a packed transaction. This enables the p2p layer to see if a received transaction is a duplicate. This will cut down on the log noise and unnecessary error handling that would otherwise occur.

taokayan:
Would you give me an example of implicit transaction?

taokayan:
from the chain_controller::__apply_block, it looks like implicit_transactions will always be built before applying each transactions from the incoming block. I think it's pretty unlikely(or impossible) that it can't be resolved since the txnId for implication transaction should be identical among different nodes. 

Also I've made a debug printf locally in founding the implicit transaction:
```
diff --git a/libraries/chain/chain_controller.cpp b/libraries/chain/chain_controller.cpp
index 27ef47f..37a8724 100644
--- a/libraries/chain/chain_controller.cpp
+++ b/libraries/chain/chain_controller.cpp
@@ -890,6 +890,7 @@ void chain_controller::__apply_block(const signed_block& next_block)
                         return &*_temp;
                      } else {
                         //ilog( "implicit" );
+                        printf("got implicit txn in __apply_block()\n");
                         for ( size_t i=0; i < next_block_trace.implicit_transactions.size(); i++ ) {
                            if ( input_metas[i].id == receipt.id )
                               return &input_metas[i];
```

And here's what I got from stdout:
```
got implicit txn in __apply_block()
push block #2407 from initm 2018-04-23T02:47:56.500  00000967867a587f5a5a4d1159d045881db19e88c778f0cf58f94c14ab40d672 lib: 745 success
got implicit txn in __apply_block()
push block #2408 from initm 2018-04-23T02:47:57.000  00000968e7f53f7d671846d4ca2fb8800eba1a51d0a2fd35b3e653c9c70d3c4b lib: 745 success
got implicit txn in __apply_block()
```

PR #2410, opened 2018-04-15
P2p optimize retry 2408

pmesnier:
addresses issue #2408. 
Added transaction retrieval interface for fetching transaction id from the packed transaction.
Fix the handling of block summaries that refer to unknown implicit transactions. Now these blocks are handled the same as notices, causing an immediate request for the full block. Secondly, unlike ordinary notices the big message manager will forward on the requested full blocks.

### DIFF
--- a/libraries/chain/transaction.cpp
+++ b/libraries/chain/transaction.cpp
@@ -255,6 +255,13 @@ vector<bytes> packed_transaction::get_context_free_data()const
    } FC_CAPTURE_AND_RETHROW((compression)(packed_context_free_data))
 }
 
+transaction_id_type packed_transaction::id()const
+{
+   try {
+      return get_transaction().id();
+   } FC_CAPTURE_AND_RETHROW((compression)(packed_trx))
+}
+
 transaction packed_transaction::get_transaction()const
 {
    try {
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -555,6 +555,8 @@ namespace eosio {
       deque<block_id_type> _blocks;
       chain_plugin * chain_plug;
 
+      constexpr auto stage_str(stages s );
+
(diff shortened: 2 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1683,7 +1699,7 @@ namespace eosio {
             if (!b) {
                send_req = true;
                req.req_blocks.ids.push_back( blkid );
-               req_blks.push_back( blkid );
+               req_blks.push_back( (block_request){blkid, generated} );
```
PaulCalabrese: I think this would look cleaner without the typename: push_back( {blkid, generated} );

libraries/chain/transaction.cpp
```diff
@@ -255,6 +255,13 @@ vector<bytes> packed_transaction::get_context_free_data()const
    } FC_CAPTURE_AND_RETHROW((compression)(packed_context_free_data))
 }
 
+transaction_id_type packed_transaction::id()const
```
PaulCalabrese: I was looking into this function because I was concerned with potential performance impact (it unpacks the transaction just to get its id) and couldn't find any place it was used. Do we want to add it if we don't need it?

pmesnier: This is intended to be used inside the net plugin for handling packed transactions and weeding out duplicates before calling the chain controller. That way we avoid all the error handling logic. It seems that my git stashing, diffing, patching, and reverting managed to whack that code. I'm reapplying the change now.

pmesnier: Regarding the performance impact potential, that is why I added a function to the packed transaction type - this function can be refactored to, for example, if a separate unpacked transaction id is carried with the packed transaction. Or perhaps the intermediate unpacked transaction could be cached so it doesn't need to be redone.

Issue #2413, opened 2018-04-16
Error message for a refused connection to nodeos/ keosd is not displayed properly


andriantolie:
Previously, error for a refused connection to nodeos/ keosd is detected by parsing the error message thrown by httpc.cpp. The recent changes introduced to httpc.cpp has suppressed the error message thrown and hence the error message for the refused connection to nodeos/keosd is not displayed properly

jgiszczak:
Verified.

PR #2414, opened 2018-04-16
#2413 - Fix error message for refused connection not displayed properly

andriantolie:
Error message for refused connection not displayed properly because of the failure in processing the previously thrown error message

#2413

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -46,9 +46,12 @@ namespace eosio { namespace client { namespace http {
             endpoint_iterator = end;
          } catch( std::exception& e ) {
             ++endpoint_iterator;
-            if( endpoint_iterator != end )
+            if( endpoint_iterator != end ) {
                continue;
-            else throw;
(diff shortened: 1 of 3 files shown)


programs/cleos/httpc.cpp
```diff
@@ -46,9 +46,12 @@ namespace eosio { namespace client { namespace http {
...
          } catch( std::exception& e ) {
             ++endpoint_iterator;
-            if( endpoint_iterator != end )
+            if( endpoint_iterator != end ) {
                continue;
-            else throw;
+            } else {
+               FC_THROW_EXCEPTION(connection_exception, "Connection to ${server}:${port}${path} is refused",
+                                                         ("server", server)("port", port)("path", path));
+            }
```
heifner: Can we add the what() detail of the std::exception and include it when verbose output is requested? Seems like the that might have useful information in some cases.

andriantolie:
PR Updated to log std::exception what()

Issue #2195, opened 2018-04-06
Test Security Delays


bytemaster:
1. a great contract is published
2. people put money in, has an invest action.
3. the owner set huge delay for setcode to pass
4. community is reassured
5. rogue is tempted by the 750M accumulated
6. rogue deleteauth, unlinkauth then setcode with the invest method tweaked to transfer the funds to his account
7. .. profit ?!

The test:
1. unlink or relink of setcode to a new authority should have the delay of current setcode authority or active.
2. changing of the auth levels should require same permission level.

Unless owner or active has a delay on it, they can always instantly set permission levels. So for contract to be trusted it should:

1. Set permission levels with required delays on various actions
2. Set owner and active to NULL or something with an even longer delay.

brianjohnson5972:
As these manual (or integration script tests) are performed, also update the [delay unit tests](https://github.com/EOSIO/eos/blob/master/tests/chain_tests/delay_tests.cpp).

taokayan:
pull #2417 committed with extra test case to ensure min delay validation for updateauth, linkauth & unlinkauth

PR #2417, opened 2018-04-16
add testcase for #2195 (security delays)

taokayan:
#2195 add delay validation for linkauth/updateauth. add test cases for unlinkauth.


tests/chain_tests/delay_tests.cpp
```diff
@@ -923,6 +923,16 @@ BOOST_AUTO_TEST_CASE( link_delay_link_change_test ) { try {
...
 
+   try {
+      chain.push_action(config::system_account_name, contracts::linkauth::get_name(), tester_account, fc::mutable_variant_object()
+      ("account", "tester")
+      ("code", "currency")
+      ("type", "transfer")
+      ("requirement", "second"),
+      30, 3);
+      BOOST_FAIL("linkauth should have a min delay set by previous linkauth");
+   } catch (...) { }
```
brianjohnson5972: please add specific exception here and expected content (if it is used in more than one place).   We don't want to say this is passing if another exception is being thrown.

brianjohnson5972: Use  BOOST_REQUIRE_EXCEPTION (with a predicate that identifies the specific exception content, if needed)

tests/chain_tests/delay_tests.cpp
```diff
@@ -1024,6 +1034,202 @@ BOOST_AUTO_TEST_CASE( link_delay_link_change_test ) { try {
...
+   BOOST_REQUIRE_EQUAL(asset::from_string("0.0000 CUR"), liquid_balance);
+
+   try {
+      chain.push_action(config::system_account_name, contracts::unlinkauth::get_name(), tester_account, fc::mutable_variant_object()
+      ("account", "tester")
+      ("code", "currency")
+      ("type", "transfer"),
+      30, 7);
+      BOOST_FAIL("unlink should have a min delay set by previous linkauth");
+   } catch (...) { }
```
brianjohnson5972: Please us BOOST_REQUIRE_EXCEPTION with specific exception details.

tests/chain_tests/delay_tests.cpp
```diff
@@ -1444,6 +1650,18 @@ BOOST_AUTO_TEST_CASE( canceldelay_test ) { try {
...
+      chain.push_action(config::system_account_name, contracts::updateauth::get_name(), tester_account, fc::mutable_variant_object()
+            ("account", "tester")
+            ("permission", "first")
+            ("parent", "active")
+            ("data",  authority(chain.get_public_key(tester_account, "first")))
+            ("delay", 0),
+            30, 7
+      );
+      BOOST_FAIL("updateauth should have a min delay set by previous updateauth");
+   } catch (...) { }
```
brianjohnson5972: Please use BOOST_REQUIRE_EXCEPTION with specific exception predicate.

tests/chain_tests/delay_tests.cpp
```diff
@@ -1024,6 +1034,202 @@ BOOST_AUTO_TEST_CASE( link_delay_link_change_test ) { try {
...
+   BOOST_REQUIRE_EQUAL(asset::from_string("0.0000 CUR"), liquid_balance);
+
+   chain.produce_blocks(16);
+
+   liquid_balance = get_currency_balance(chain, N(tester));
+   BOOST_REQUIRE_EQUAL(asset::from_string("100.0000 CUR"), liquid_balance);
+   liquid_balance = get_currency_balance(chain, N(tester2));
+   BOOST_REQUIRE_EQUAL(asset::from_string("0.0000 CUR"), liquid_balance);
+
+   // this transaction will be delayed 20 blocks
```
brianjohnson5972: What does the second delayed transfer do for this test?  If it doesn't add to test coverage and just makes the test run longer, we should remove it.

brianjohnson5972: (I am basing my statement off of the exception you are expecting for the unlink above)

taokayan:
hi I've created another pull request replacing "try" into BOOST_REQUIRE_EXCEPTION. Please would you review it?

tests/chain_tests/delay_tests.cpp
```diff
@@ -923,15 +923,19 @@ BOOST_AUTO_TEST_CASE( link_delay_link_change_test ) { try {
...
       ("code", "currency")
       ("type", "transfer")
       ("requirement", "second"),
-      30, 3);
-      BOOST_FAIL("linkauth should have a min delay set by previous linkauth");
-   } catch (...) { }
+      30, 3),
+      transaction_exception,
+      [] (const transaction_exception &e)->bool {
+         BOOST_REQUIRE_EQUAL(std::string("transaction validation exception"), e.what());
```
brianjohnson5972: Please use  e.to_detail_string() and check for something like "greater than the delay specified" so we know that it is this exception, same for the other BOOST_REQUIRE_EXCEPTIONs that are similar to this.

Issue #2350, opened 2018-04-12
get table array values not being returned


brianjohnson5972:
Specific example is tic_tac_toe tutorial, "get table" returns:

{
  "rows": [{
      "challenger": "player2",
      "host": "player1",
      "turn": "player2",
      "winner": "none",
      "board": []
    }
  ],
  "more": false
}

even though board is a uint8_t[9] and has values

brianjohnson5972:
@wanderingbort                        
I think the issue is that in the abi the type is defined as "uint8_t[]" which maps to a vector<T> for pack/unpack, but in tic_tack_toe.hpp the parameter is:
uint8_t      board[9];
Which will NOT be mapped to operator<< with vector<T>, so the packing/unpacking don't match.

I think it is calling "DataStream& operator << ( DataStream& ds, const std::array<T,N>& v )" which is more efficient for packing/unpacking in contracts, but there is nothing on the abi that could handle this.

brianjohnson5972:
Handling this by implementing operator<< for std::array and adding one for "T (&v)[N}" and using a similar implementation to std::vector<T> so that there is a size for abi serializer to use to size the vector that it will allocate to handle all of these types.

andriantolie:
Played the tic_tac_toe contract and can verify for each step the board is printed properly by cleos get table command

PR #2424, opened 2018-04-16
contract c array packing

brianjohnson5972:
#2350 
Needed to change datastream.hpp and raw.hpp to provide packing/unpacking for c-arrays to add size to the packing so abi-serialization can be performed.

### DIFF
--- a/contracts/eosiolib/datastream.hpp
+++ b/contracts/eosiolib/datastream.hpp
@@ -225,18 +225,19 @@ inline datastream<Stream>& operator>>(datastream<Stream>& ds, checksum256& d) {
 template<typename DataStream>
 DataStream& operator << ( DataStream& ds, const std::string& v ) {
    ds << unsigned_int( v.size() );
-   for( const auto& i : v )
-      ds << i;
+   if (v.size())
+      ds.write(v.data(), v.size());
    return ds;
 }
 
 template<typename DataStream>
 DataStream& operator >> ( DataStream& ds, std::string& v ) {
-   unsigned_int s;
-   ds >> s;
-   v.resize(s.value);
-   for( auto& i : v )
-      ds >> i;
+   std::vector<char> tmp;
+   ds >> tmp;
+   if( tmp.size() )
+      v = std::string(tmp.data(),tmp.data()+tmp.size());
+   else
+      v = std::string();
    return ds;
 }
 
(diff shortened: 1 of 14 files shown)


libraries/fc/include/fc/io/raw.hpp
```diff
@@ -142,9 +142,26 @@ namespace fc {
 
     template<typename Stream, typename T, size_t N>
     inline void pack( Stream& s, const fc::array<T,N>& v) {
-      s.write((const char*)&v.data[0],N*sizeof(T));
+       for (uint64_t i = 0; i < N; ++i)
```
wanderingbort: We should probably not remove this optimization for lots of fundamental types.

Can we guard this something like `std::is_scalar && ! std::is_pointer` so that the slow path is only used when it will probably differ from the fast path?

contracts/eosiolib/datastream.hpp
```diff
@@ -254,6 +254,24 @@ DataStream& operator >> ( DataStream& ds, std::array<T,N>& v ) {
    return ds;
 }
 
+template<typename DataStream, typename T, std::size_t N>
```
wanderingbort: I would prefer at least a specialization for `char (&v)[N]` to match the `vector<char>` specializations below.  However I think even those could be expanded  to be `is_scalar && !is_pointer` style overloads

libraries/fc/include/fc/io/raw.hpp
brianjohnson5972: we use is_fundemental, isn't that the one I should use, since we don't support packing a pointer (at least not directly)

wanderingbort: `is_fundamental` won't catch pointers, enumerations pointers-to-members etc.  we have a change incoming to other uses of `is_fundamental` to change them to `is_scalar` and either `static_assert` if pointers are invalid.   It is more relevant when we are using "magic_get" because using the `for_each` on something like a pointer tends to cause an infinite recursion (if the call path into and out of for_each is recursive).

brianjohnson5972: done

contracts/eosiolib/datastream.hpp
brianjohnson5972: I just did the T (&v)[N} optimizations for is_scalar && !is_pointer, since the wasm compiler was exploding on the vector<T> optimization.

I also added an optimization for std::string.

brianjohnson5972:
Ready for re-review

PR #2427, opened 2018-04-16
Cleos multisig

moskvanaft:
cleos multisig propose
cleos multisig approve
cleos multisig unapprove
cleos multisig exec
cleos multisig cancel

+ README.md with actions description and cleos usage example


programs/cleos/main.cpp
```diff
@@ -1153,6 +1156,263 @@ int main( int argc, char** argv ) {
...
+            accountPermissions = vector<permission_level>{{proposer, config::active_name}};
+         } else {
+            EOS_THROW(tx_missing_auth, "Authority is not provided (either by multisig parameter <proposer> or -p)");
+         }
+      }
+      if (proposer.empty()) {
+         proposer = name(accountPermissions.at(0).actor).to_string();
+      }
+
+      arg = fc::mutable_variant_object()
```
wanderingbort: Is there a reason to construct this as a variant instead of a `transaction` ?  This seems like it would be pretty fragile if changes to the transaction structure occur.

programs/cleos/main.cpp
```diff
@@ -1153,6 +1156,263 @@ int main( int argc, char** argv ) {
...
+         ("code", "eosio.msig")
+         ("action", "propose")
+         ("args", fc::mutable_variant_object()
+          ("proposer", proposer )
+          ("proposal_name", proposal_name)
+          ("requested", requested_perm_var)
+          ("trx", fc::mutable_variant_object()
+           ("expiration", "2018-05-01T00:00:00")
```
wanderingbort: This seems suspect.  Are we not validating expiration for the proposed transaction or is this only succeeding because its in the future.

Perhaps we need to use the existing command line arguments for things like expiration, max_kcpu_usage etc?

programs/cleos/main.cpp
```diff
@@ -1153,6 +1156,263 @@ int main( int argc, char** argv ) {
...
+                 }
+           )
+          )
+         );
+      result = call(json_to_bin_func, arg);
+      send_actions({chain::action{accountPermissions, "eosio.msig", "propose", result.get_object()["binargs"].as<bytes>()}});
+   });
+
+   //resolver for ABI serializer to decode actions in proposed transaction in multisig contract
+   auto resolver = [](const name& code) -> optional<contracts::abi_serializer> {
```
wanderingbort: In an effort to keep cleos as "thin" as possible I think we should consider moving all of these uses of abi_serializer that depend on chain state (`get_code_func`) to the chain RPC with a call that mirrors and is the opposite of `abi_json_to_bin`.  It is a round-trip to the RPC either way but, if we legitimize that endpoint then other non-cleos HTTP services can benefit.

Later as an optimization if we want to introduce a client side cache for the ABI we can do so but, then the ABI's would come locally.

programs/cleos/main.cpp
moskvanaft: In order to construct transaction instead of json I would need to request ABI from nodeos and use ABI serializer.
In another comment to this PR you are arguing against using ABI serializer in cleos. So, I am confused .

programs/cleos/main.cpp
moskvanaft: I removed the hard-coded value. Now cleos multisig propose has additional parameter with 24 hours as default.

PR #2431, opened 2018-04-16
Modify net_plugin's logging initialization

spoonincode:
Change net_plugin to ask for its logger vs pushing a logger in to the logging framework. This allows net_plugin to honor the logging.json configuration easier. Non-contentious change: no longer can you log-level-net-plugin to change the debug level; use the common logging.json to set the log level. More contentious change: network logging is not added by default, it must be placed in to logging.json

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -265,7 +265,7 @@ namespace eosio {
    };
 
    const fc::string logger_name("net_plugin_impl");
-   fc::logger logger(logger_name);
+   fc::logger logger;
 
    template<class enum_type, class=typename std::enable_if<std::is_enum<enum_type>::value>::type>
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -265,7 +265,7 @@ namespace eosio {
    };
 
    const fc::string logger_name("net_plugin_impl");
-   fc::logger logger(logger_name);
+   fc::logger logger;
```
jgiszczak: So if a logger named "net_plugin_impl" doesn't appear in the config file, this nameless logger is the one that's used, and since it has no appenders, it acts as a silent sink for log messages?

spoonincode: Yes, that's correct. Not saying it's the best approach.

PR #2435, opened 2018-04-16
Remove log-level-net-plugin option completely

spoonincode:
In my haste to fix net_plugin's gelf logging, I didn't remove the now defunct log-level-net-plugin config item


spoonincode:
Not ready I need to figure out how to change the tests

Issue #2416, opened 2018-04-16
Contract Compilation Error using latest document of EOS dawn 3.0


forfrt:
I was following the steps of [document](https://github.com/EOSIO/eos/wiki/Smart%20Contract) in eos.io website to debug smart contract. And this error occured as follow:

    In file included from debug.cpp:2:
    ./debug.hpp:2:10: fatal error: 'eoslib/eos.hpp' file not found
    #include <eosiolib/eos.hpp>
             ^~~~~~~~~~~~~~~~~~
    1 error generated.

And I have investigated for a while and install the build to /usr/local/include by the command `make install` under build directory. But there is no file named eos.hpp and no directory called eoslib. Any suggestion or advices?

PR #2441, opened 2018-04-16
Fix for 2416 : discriminating prefixes 

wanderingbort:
resolve ambiguous prefixes on pub/priv keys and sigs

The prefix is now `PUB/PVT/SIG.K1/R1.`

The legacy rules are still in place so, a public key prefixed with `EOS` is still treated as an typeless-checksummed `PUB.K1.`  and a wif private key is imported as a `PVT.K1.`

fixes EOSIO/eos#2416


wanderingbort:
triple click seems to work 😈

spoonincode:
But that's only because triple click is whole line. I mean hey let's try, can you import my key PVT.R1.yXiskXjsTkQ5GxhQp1jJqPLfz2cZkah71ZZVwsAciAxsMbiDU in to your wallet?

heifner:
`_` works if you want to use that instead of `.`

Let's try underscore:, can you import my key PVT_R1_yXiskXjsTkQ5GxhQp1jJqPLfz2cZkah71ZZVwsAciAxsMbiDU in to your wallet?

danirabbani90:
Hi there , i have the question is that can i be able to customise the Prefix address symbol rather than EOS ?

PR #2448, opened 2018-04-17
Add example dice game session using cleos

elmato:
* Add example usage using cleos
* Remove generic_currency dependency
* Use eosio.token as the account holding the EOS native token
* Fix tests

### DIFF
--- a/contracts/dice/dice.cpp
+++ b/contracts/dice/dice.cpp
@@ -5,22 +5,20 @@
 
 #include <utility>
 #include <vector>
-#include <eosiolib/crypto.h>
-#include <eosiolib/types.hpp>
-#include <eosiolib/token.hpp>
-#include <eosiolib/print.hpp>
-#include <eosiolib/action.hpp>
-#include <eosiolib/multi_index.hpp>
+#include <string>
+#include <eosiolib/eosio.hpp>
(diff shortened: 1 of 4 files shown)


contracts/dice/dice.cpp
```diff
@@ -37,9 +35,10 @@ class dice : public eosio::contract {
       //@abi action
       void offerbet(const asset& bet, const account_name player, const checksum256& commitment) {
 
-         auto amount = eos_currency::token_type(bet);
+         eosio_assert( bet.symbol == S(4,EOS) , "only EOS token allowed" );
```
heifner: Is this check sufficient to guarantee THE EOS token? Currently, I believe, multiple accounts can have an EOS token. Are we planning on having a symbol lib and guarantee only one token of each unique symbol?

elmato: Yes, multiple accounts can create an _EOS_ token, but If **THE EOS** token will be hosted in the **eosio.token** contract (as I think it will be), the newest version of **eosio.token** can only host one token with _symbol_code == EOS_ (that hopefully will be **THE EOS** token).

Dice contract only interacts with **eosio.token**, so I think the check is sufficient.

Issue #2275, opened 2018-04-09
Block signing key fails with R1 keys


wanderingbort:
# Description

When a producer schedule includes a producer with an R1 key, blocks signed and pushed to the network the signature fails to verify:

```
3132994ms thread-0   chain_controller.cpp:254      _push_block          ] Failed to push new block:
3020000 block_validate_exception: block validation exception
Incorrect block producer key: expected EOSR14yFnKuAQyM9XziqJWhbr2nF2scfYESh9YmuBqiqtQtzAE2UdsA but got EOSR15i7ELS4HgwToNu7zYWDiNAKtq58aGGoBMon3tt7DwMFNDx6WBA
    {"e":"EOSR14yFnKuAQyM9XziqJWhbr2nF2scfYESh9YmuBqiqtQtzAE2UdsA","a":"EOSR15i7ELS4HgwToNu7zYWDiNAKtq58aGGoBMon3tt7DwMFNDx6WBA"}
    thread-0  chain_controller.cpp:1422 validate_block_header

    {"block_header(next_block)":{"previous":"00000537c187bc63de15f8ea2d8be695810d9e7d03c560a5d2f49e3cd7cc92dc","timestamp":"2018-04-09T14:52:13.000","transaction_mroot":"aa356815d0d1d8ce3a153d12ae8672a9f25eca05fb7f9328ed8a3a5d19a0e065","action_mroot":"94cb32371c4da45620baab5893b8bbf781c32fa32b459dd23eaa60762ca26b89","block_mroot":"294c5f8b6ce15d8b99002858b98b691aba3cab27947ef4a6a234a6613a6cf887","producer":"spoon.prod","schedule_version":2}}
    thread-0  chain_controller.cpp:1435 validate_block_header

    {"next_block.block_num()":1336}
    thread-0  chain_controller.cpp:957 __apply_block
3132994ms thread-0   net_plugin.cpp:2313           handle_message       ] block_validate_exception accept block #1336 syncing from matt-witherspoon-mbp.local:4444 - fb13e56
```

# Acceptance Test Criteria

- [ ] Block producer with R1 pubkey as block signing key can produce a block which is validated by the network

andriantolie:
Unit test https://github.com/EOSIO/eos/pull/2470 is added to proof that the issue is solved

PR #2470, opened 2018-04-18
Add unit test to verify producer can produce block with r1 signing key

andriantolie:
For testing #2275

### DIFF
--- a/tests/chain_tests/block_tests.cpp
+++ b/tests/chain_tests/block_tests.cpp
@@ -1187,4 +1187,25 @@ BOOST_AUTO_TEST_CASE(account_ram_limit) { try {
 
 } FC_LOG_AND_RETHROW() }
 
+BOOST_AUTO_TEST_CASE(producer_r1_key) { try {
+
+   // Use validating_tester to check that the block is synced properly
+   validating_tester chain;
+
+   // Set new producer
+   account_name tester_producer_name = "tester";
+   chain.create_account(tester_producer_name);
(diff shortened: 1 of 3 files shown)


tests/chain_tests/block_tests.cpp
```diff
@@ -1157,5 +1157,25 @@ BOOST_AUTO_TEST_CASE(transaction_mroot)
...
+      auto producer_r1_priv_key = chain.get_private_key<fc::crypto::r1::private_key_shim>( tester_producer_name, "active" );
+      auto producer_r1_pub_key = producer_r1_priv_key.get_public_key();
+      chain.push_action(N(eosio), N(setprods), N(eosio),
+                        fc::mutable_variant_object()("version", 1)("producers", vector<producer_key>{{ tester_producer_name, producer_r1_pub_key }}));
+
+      // Add signing key to the tester object, so it can sign with the correct key
+      chain.block_signing_private_keys[producer_r1_pub_key] = producer_r1_priv_key;
+
+      // Produce blocks, it should not throw any exception
+      BOOST_CHECK_NO_THROW(chain.produce_blocks(1000));
```
heifner: 1000 is excessive. Please reduce to minimize run time.
This should be sufficient:
```
chain.produce_blocks_until_end_of_round();
chain.produce_block();
```

andriantolie:
PR is updated. It now checks only until the new set of producers finishes its first round

PR #2473, opened 2018-04-18
P2p backcompat 2454

pmesnier:
mix in backwards compatibility by comparing the version number from the peer with list of old commits to determine if the old sync behavior should be used.


plugins/net_plugin/old_versions.hpp
```diff
@@ -0,0 +1,4806 @@
+/*
```
heifner: This seems very excessive to me. Can't we just add a version to handshake_message and check that version on what to do? You could distinguish old handhake_messages without version via a sizeof.

pmesnier: it would have to be a different message type. the p2p protocol sends serialized objects as binary blobs, keyed by message type. I think that a message of a given type that doesn't conform to the definition will be a problem. But I suppose adding a new message type is doable. However, we are already using git ids as "version" identifiers, and so my list is just a natural extension of that. And really either solution is temporary since the old versions will be obsolete soon anyway. I thought about trimming the list but what's the point. Even at 4k entries, it's only 16k so hardly worth the trouble.

heifner: Discussed this at length in SCRUM. Decided by the group to push this change in while a better more long lasting fix is put into place with a proper protocol version.

Issue #2253, opened 2018-04-09
keosd API to use a secure protocol (e.g. HTTPS)


gleehokie:

DenisCarriere:
This would be a great feature to include in RC2, right now we've got 13+ Block Producers all hosting over HTTPS (mostly LetsEncrypt + Nginx).

https://scholar.eosnation.io

Unfortunately, using `cleos` doesn't seem to support `HTTPS` domains at the moment.

Feel free to use the following BP's to test HTTPS `cleos` connections:

```
cleos -p 443 --host scholarnet.eosnation.io get block 1
cleos -p 443 --host testnet.eosocal.io get block 1
cleos -p 8911 --host scholartestnet.cryptolions.io get block 1
cleos -p 443 --host testnet.meet.one get block 1
cleos -p 443 --host testnet.eoscannon.io get block 1
```

CC: @GMory @wuyahuang @luckybean

PR #2490, opened 2018-04-18
HTTPS support for http_plugin

spoonincode:
Add HTTP support to the http_plugin. Users can configure an HTTPS listener via a new `https-server-address` config and optionally disable plain HTTP as well.

This could still use some testing with "real" certs (certs from trusted CAs).

Part of #2253

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -8,6 +8,7 @@
 #include <fc/log/logger_config.hpp>
 #include <fc/reflect/variant.hpp>
 #include <fc/io/json.hpp>
+#include <fc/crypto/openssl.hpp>
 
 #include <boost/asio.hpp>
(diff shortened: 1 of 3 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -102,15 +203,23 @@ namespace eosio {
...
+             "The local IP and port to listen for incoming http connections; set blank to disable.")
+
+            ("https-server-address", bpo::value<string>(),
+             "The local IP and port to listen for incoming https connections; leave blank to disable.")
+
+            ("https-certificate-chain-file", bpo::value<string>(),
+             "Filename with the certificate chain to present on https connections. PEM format. Required for https.")
+
+            ("https-private-key-file", bpo::value<string>(),
+             "Filename with https private key in PEM format. Requested for https")
```
jgiszczak: s/Requested/Required/

PR #2491, opened 2018-04-18
Disallow NaN floating point values in secondary tables

### LINKED ISSUE #2460: Disallow NaN floating point values in secondary tables
NaNs should not be allowed as a floating point secondary key value inside a secondary table since we cannot sort properly on them as part of a floating point secondary index.

arhag:
Fixes #2460. Also added in fixes to long double comparison intrinsics in compiler_builtins.

The transaction will fail if a contract tries to store/update a floating-point secondary table with a floating-point key value that is NaN.
The transaction will also fail if a contract tries to lookup (using find secondary, lower bound, or upper bound) a row in a floating-point secondary table with a floating-point key value that is NaN.

ATC:
* Using `db_idx_double_store` to store a NaN in a new row of a double secondary table should fail.
* Using `db_idx_double_update` to modify an existing row of a double secondary table to include a NaN value should fail.
* Using `db_idx_double_find_secondary` with a NaN lookup key should fail.
* Using `db_idx_double_lowerbound` with a NaN lookup key should fail.
* Using `db_idx_double_upperbound` with a NaN lookup key should fail.

This PR already implements tests that handle the above ATC.


contracts/test_api_db/test_db.cpp
```diff
@@ -495,3 +495,44 @@ void test_db::test_invalid_access(uint64_t receiver, uint64_t code, uint64_t act
...
    }
 }
+
+void test_db::idx_double_nan_create_fail(uint64_t receiver, uint64_t, uint64_t) {
+   double x = 0.0;
+   x = x / x; // create a NaN
+   db_idx_double_store( N(nan), N(nan), receiver, 0, &x); // should fail
+}
+
+void test_db::idx_double_nan_setup(uint64_t receiver, uint64_t, uint64_t) {
```
larryk85: Just move this store to each of the modify tests.

PR #2492, opened 2018-04-18
Cleos launches keos

moskvanaft:
cleos launches keos if wallet_port is not accepting connections
`cleos wallet stop` sends command that causes keosd to exit

keosd still uses tcp (not local socket as @spoonincode suggested)


programs/cleos/main.cpp
```diff
@@ -850,6 +895,10 @@ int main( int argc, char** argv ) {
    app.add_option( "--wallet-host", wallet_host, localized("the host where keosd is running"), true );
    app.add_option( "--wallet-port", wallet_port, localized("the port where keosd is running"), true );
 
+   if ( !port_busy( wallet_port ) ) {
```
spoonincode: What if wallet-host is not localhost?

moskvanaft: I added a simple check for localhost. It doesn't cover corner cases like wallet_host is external IP of localhost or hostname other than "localhost" (it's hard to make this check platform-independent). I guess in most cases wallet_host will be empty for local wallet. Worst case - we don't start keosd automatically.

PR #2493, opened 2018-04-18
P2p net vernum 2425

pmesnier:
revised network version computation to use a simple integer network version number. I found the longest gap between consecutive historical commit ids, and use that gap to count versions.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -272,7 +271,7 @@ namespace eosio {
        */
       chain::signature_type sign_compact(const chain::public_key_type& signer, const fc::sha256& digest) const;
```
jgiszczak: Also remove function starting at line 63.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2015,15 +2034,16 @@ namespace eosio {
             return;
          }
          if( msg.network_version != network_version) {
-            c->backwards_compatibility = is_old_version (msg.network_version);
+            int16_t mnv = to_net_version(msg.network_version);
+            c->backwards_compatibility = mnv < net_version;
```
heifner: Seems like you need to set net_version and check it as boolean backwards_compatibility will not be enough going forward.

PR #2517, opened 2018-04-20
Modify on_block_tx_mroot inside the block_tests/transaction_mroot

andriantolie:
Previously the on_block_tx_mroot inside the block_tests/transaction_mroot is hardcoded. With this modification, the on_block_tx_mroot is determined by producing an empty block and extracts its transaction_mroot instead of hardcoded.


arhag:
The definition of `transaction_mroot` is going to change with the upcoming slim branch, so there is no need to continue working on this particular test at the moment. In the the slim branch, the implicit on-block transaction is not committed to in the `transaction_mroot`.

When the slim branch is eventually merged, a new block_tests/transaction_mroot test will be needed to test the new definition of `transaction_mroot`.

PR #2526, opened 2018-04-20
Vote max transaction delay

moskvanaft:


moskvanaft:
Already merged in PR #2563

Issue #2475, opened 2018-04-18
refactor the delivery of large messages via notify/request pairs


pmesnier:
The implementation of this feature is unfinished. To ensure efficient testing, I propose doing he following:
1. add a new maximum "just sent it" message size that defaults to "infinite" meaning always just send the message, a size of 0 means always send a notice.
2. fix the missing functionality: 
  - ensure the peer connection's  block_state.is_known flag is set correctly in all cases. 
  - ensure that notice messages are propagated to subordinate peers once the node has received the original large message is retrieved
  - implement the big_msg_manager::retry_fetch function to retry large message retrieval in the case of a timeout or loss of connection while waiting for a requested block/transaction
3. Have the block producing node just aways send the block summary regardless of size since all of tis peers will ask it for he block anyway. Subsequent nodes should still send notices if necessary

PR #2527, opened 2018-04-20
p2p explicit notify/request feature - issue #2475

pmesnier:
adding in the rest of the notice/request code. Adding in some fixes for preceding prs as well. 
In this commit - I added code to handle a failed fetch, and also to ensure the is_known table is updated correctly for transactions and blocks.  I also returned the default implicit request ("just send it") size.

bonus patches include typo fixes in config option, and replacement of backwards_compatibility with a per-connection network protocol number that is checked as needed.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -269,7 +269,7 @@ namespace eosio {
        */
       chain::signature_type sign_compact(const chain::public_key_type& signer, const fc::sha256& digest) const;
 
-      int16_t to_protocol_version(int16_t v);
+      uint16_t to_protocol_version(uint16_t v);
    };
 
    const fc::string logger_name("net_plugin_impl");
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2356,11 +2430,18 @@ namespace eosio {
       } catch( ...) {
          elog( "handle sync block caught something else from ${p}",("num",blk_num)("p",c->peer_name()));
       }
+#if 0 // after reviewing my conversaion with Bart, I believe this to be improper.
+      // as he explained, if the the block is deemed invalid by the local chain, then this implies a
+      // fork by the sender.
       notice_message pending_notify;
```
PaulCalabrese: Why leave this dead code in?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2685,9 +2766,6 @@ namespace eosio {
          ( "p2p-server-address", bpo::value<string>(), "An externally accessible host:port for identifying this node. Defaults to p2p-listen-endpoint.")
          ( "p2p-peer-address", bpo::value< vector<string> >()->composing(), "The public endpoint of a peer node to connect to. Use multiple p2p-peer-address options as needed to compose a network.")
          ( "agent-name", bpo::value<string>()->default_value("\"EOS Test Agent\""), "The name supplied to identify this node amongst the peers.")
-#if 0 //disabling block summary support
```
PaulCalabrese: Why delete the send-whole-blocks code farther down, but leave this commented out?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1686,8 +1710,15 @@ namespace eosio {
       }
    }
 
-   void big_msg_manager::recv_transaction (connection_ptr c) {
+   void big_msg_manager::recv_transaction (connection_ptr c, const transaction_id_type& tid) {
       pending_txn_source = c;
+      if (c &&
```
PaulCalabrese: The reset logic contained in this if is repeated twice and should be pulled out into a member function of connection.

PR #2540, opened 2018-04-22
 Enforce max_push_transaction_us setting when a transaction is pushed to nodeos

elmato:


elmato:
@wanderingbort you are right.
I was hanging the local node with a simple while(1){} contract and I didn't see that **--max-pending-transaction-time** parameter is finally used in **wrap_transaction_processing** which is were all execution paths leads to.

Issue #2544, opened 2018-04-23
Missing option to configure shared_memory_size


andriantolie:
Since database_plugin is deprecated, there is no more option to configure the shared_memory_size and hence the shared_memory_size always defaults to 1GB. 1GB can quickly be filled (by just making 
 around 520,000 accounts), and when it's fully filled, any write action to the state will always throw a boost::interprocess::bad_alloc exception.

PR #2546, opened 2018-04-23
Add shared memory size option

andriantolie:
Related to #2544 

Previously, shared_memory_size is configurable via database_plugin config. But now, data_base_plugin is deprecated, so this PR add the shared_memory_size option to chain_plugin_config


spoonincode:
I think the consensus is that 1GB is too small, so do we want to consider other approaches before merging this?

heifner:
Phil's PR just changed the default to 32GB. So I think with his PR this gets us there.

PR #2548, opened 2018-04-23
feat(eosio.token): add inline_transfer static method

Akagi201:
Add `inline_transfer` static method to eosio.token.

This method is needed for other contracts to call tranfer action.

It is similar to what old [`currency::inline_transfer`](https://github.com/EOSIO/eos/blob/master/contracts/eosiolib/currency.hpp#L96) did.


Akagi201:
@zorba80 `transfer_args`  doesn't need a `EOSLIB_SERIALIZE` macro?
In my tests, structs in action params should have a `EOSLIB_SERIALIZE` macro.

Issue #2132, opened 2018-04-05
Make intrinsics that read data into a buffer consistent WRT undersized buffers


wanderingbort:
# Description

We have several intrinsics that will push data into buffers supplied by WebAssembly.  When the buffer provided is not properly sized to hold the data requested we do one of three things:
 
 - Copy nothing and return the size needed to copy the full data
 - Copy up to the buffer size and return the size copied
 - Copy up to the buffer size and return the size needed to copy the full data

We should pick one and make all of our intrinsics comply.

A non exhaustive list of such intrinsics is:
 - [x] get_context_free_data
 - [x] get_blockchain_parameters_packed
 - [x] get_active_producers
 - [x] read_action_data
 - [x] db_get_i64
 - [x] read_transaction

bytemaster:
A related issue is whether we want to support "streaming" reads or "random access" reads. 

This has huge implications for writing efficient contracts with minimal CPU and RAM usage.

brianjohnson5972:
We would need to add an offset to start from if we wanted to support
streaming/random access.

I think copying over a maximum of the passed in bytes and returning the
actual copied bytes, with the special case of size == 0 returning the
required size is the "easiest" to explain.

heifner:
Didn't find any more intrinsics than those listed that needed to be updated.

PR #2553, opened 2018-04-23
Intrinsic read normalization

heifner:
Issue #2132 

- Read intrinsic functions now support passing in size == 0 to return the required size.
- Read intrinsic functions copy up to the request size and returns amount copied.

### DIFF
--- a/contracts/eosiolib/action.h
+++ b/contracts/eosiolib/action.h
@@ -70,7 +70,7 @@ extern "C" {
     *  Copy up to @ref len bytes of current action data to the specified location
     *  @brief Copy current action data to the specified location
     *  @param msg - a pointer where up to @ref len bytes of the current action data will be copied
-    *  @param len - len of the current action data to be copied
+    *  @param len - len of the current action data to be copied, 0 to report required size
     *  @return the number of bytes copied to msg
     */
    uint32_t read_action_data( void* msg, uint32_t len );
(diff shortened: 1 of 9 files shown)


contracts/eosiolib/action.h
```diff
@@ -70,7 +70,7 @@ extern "C" {
     *  Copy up to @ref len bytes of current action data to the specified location
     *  @brief Copy current action data to the specified location
     *  @param msg - a pointer where up to @ref len bytes of the current action data will be copied
-    *  @param len - len of the current action data to be copied
+    *  @param len - len of the current action data to be copied, 0 to report required size
     *  @return the number of bytes copied to msg
```
brianjohnson5972: Need something like ", or number of bytes that can be copied if len==0 passed", similarly for other return statements below.

libraries/chain/wasm_interface.cpp
```diff
@@ -155,14 +155,18 @@ class privileged_api : public context_aware_api {
          });
       }
 
-      uint32_t get_blockchain_parameters_packed( array_ptr<char> packed_blockchain_parameters, size_t datalen) {
+      uint32_t get_blockchain_parameters_packed( array_ptr<char> packed_blockchain_parameters, size_t buffer_size) {
```
brianjohnson5972: Need a doxygen comment in privileged.h to indicate that the behavior for this function is to copy the buffer only if the buffer_size is sufficiently large.

heifner: Done.

PR #2554, opened 2018-04-23
Remove currency contract, generic_currency

zorba80:


spoonincode:
txn_test_gen_plugin changes look good

contracts/proxy/proxy.cpp
```diff
@@ -96,13 +94,10 @@ extern "C" {
...
              apply_onerror(receiver, deferred_transaction::from_current_action());
-          } if( action == N(transfer) ) {
-             // Comment this out for now so that the contract compiles, this will change with refactoring to use eosio.token
-             //        apply_transfer(receiver, code, unpack_action_data<eosiosystem::contract<N(eosio.system)>::currency::transfer_memo>());
           }
-       } else if ( code == N(currency) ) {
+       } else if ( code == N(eosio.token) ) {
           if( action == N(transfer) ) {
-             apply_transfer(receiver, code, unpack_action_data<eosio::currency::transfer>());
+             apply_transfer(receiver, code, unpack_action_data<transfer_args>());
```
brianjohnson5972: Why does this create its own version of an action structure that is owned by the eosio.token contract?

contracts/proxy/proxy.cpp
```diff
@@ -97,7 +97,7 @@ extern "C" {
           }
        } else if ( code == N(eosio.token) ) {
           if( action == N(transfer) ) {
-             apply_transfer(receiver, code, unpack_action_data<transfer_args>());
+             apply_transfer(receiver, code, unpack_action_data<eosio::token::transfer_args>());
```
brianjohnson5972: Shouldn't this be "transfer" like all the other action names that match the action structure? Also, can this use @bytemaster 's new approach that maps the action's data to a method's parameters?  Dan's new stuff looks to me like defining these actions would be going away and the method mapping would define an implicit structure.

zorba80: We need transfer_args as a struct that holds transfer parameters. We don't have a straightforward mapping between the two. In execute_action in dispatcher.hpp, parameters are unpacked into a tuple before being passed to the action itself. Here we unpack into a struct that is passed to proxy::apply_transfer. This dispatcher cannot be replaced with the new form as it involves more logic.

brianjohnson5972: I'll take the discussion to telegram, because I know this is because of the new design, but this sure seems like this won't help users to understand what is going on.

PR #2559, opened 2018-04-23
Centos: added install of lcov

pacificcode:
added installation of lcov to CentOS.


jgiszczak:
Graphviz is used as part of generating the doxygen documentation.

PR #2577, opened 2018-04-24
Add native contract debugging capability.

PaulCalabrese:
These changes allow contracts to be compiled natively into a shared library that can then be dynamically linked directly into nodeos. This allows contract code to be debugged using a native platform debugger.  In order to get this working a number of changes to code in eosiolib was necessary.  Most of these are marked with a conditional compilation flag that is tied to the compilation mode (native vs. wasm). The most intrusive of these changes is the renaming of the "time" type to "eosio_time" because of conflicts with a system type of "time". This affects the contract programming interface and all existing contracts that use this type.

Instructions for how to debug contracts can be found in plugins/debug_plugin/README.md.


PaulCalabrese:
I spent a lot of time trying to reuse the existing parameters. A number of
factors combined to make me abandon this approach:

   - The native versions of the intrinsics needed different types than the
   non-native versions. The current intrinsic definitions are fairly sloppy
   (use of int for pointers and other platform-dependent types) and don't
   always agree with the function declarations in eosiolib (the native
   intrinsic functions need a more exact match for the linker). A basic goal
   of mine was to avoid changing the existing intrinsic macro arguments which
   increased the difficulty.
   - I was unable to use templates in the generation of the native
   intrinsic functions because these are defined in an extern "C" context and
   templates are not allowed there. Macros are unable to pull apart parameters
   like void(int32_t, int32_t).
   - Because the native intrinsic functions are type-safe (unlike the wasm
   intrinsics) I am forced to do some type casting to get things to compile.

You might be able to combine things by completely redefining how the macro
arguments work (as opposed to just adding some new ones), but this seemed a
higher risk approach.

heifner:
Out of date and a new approach to contact debugging is being explored.

PR #2579, opened 2018-04-24
Revert "Allow to use higher version boost in MacOS"

heifner:
Reverts EOSIO/eos#2576

Boost 1.67 is causing test failures on debug build. Boost bug has been submitted: https://svn.boost.org/trac10/ticket/13533


heifner:
Reverting this doesn't help because 1.67 is still installed. @pacificcode is going to put in a better fix.

InfoHunter:
So @pacificcode , what's the current status of this ?

gleehokie:
@InfoHunter https://github.com/EOSIO/eos/pull/2599 switches back to Boost 1.66 while we wait for Boost to fix their bug.

PR #2593, opened 2018-04-25
Fix new logic for accepting newer boost versions

arielscarpinelli:


heifner:
We are now pulling in Boost 1.66 due to issues with 1.67.

PR #2602, opened 2018-04-26
Compatible with Ubuntu 17

wqxiang88:
Fixed the ubuntu17 some commands outputs.


wqxiang88:
@grigio 

Yes, it does work.  The title should be say,  "Supported ubuntu-17 and higher". ^_^

grigio:
@wqxiang88 ah ok, because I've seen that the eos docker images still rely on 16.04 LTS https://github.com/EOSIO/eos/blob/master/Docker/Dockerfile#L10

obnosis:
Verified by BlockSmith team on 18.04.

PR #2606, opened 2018-04-26
[in progress] generic SQL database plugin 

asiniscalchi:
GOAL: create a plugin that will use the following databases:

- DB2
- Firebird
- MySQL
- ODBC with specific database driver
- Oracle
- PostgreSQL
- SQLite 3


heifner:
I recommend you work of the `slim` branch.

noprom:
I think ElasticSearch plugin should also be added, how do tou think?

asiniscalchi:
I am thinking that it would be nice a way to load external plugins.

PR #2613, opened 2018-04-27
Add NODEOS_URL and KEOSD_URL environment variables to cleos

angelol:
This prevents a lot of repeated typing when using cleos with a host other than localhost


bytemaster:
The proper solution to this is to have keosd and nodeosd read a config file from a default location which should contain these options.

PR #2642, opened 2018-04-28
Not necessary to judge the conditions of inline action.

gguoss:
Inline. Inline is guaranteed to execute with the current transaction or unwind; no notification will be communicated regardless of success or failure.


bytemaster:
There is no need to inline transfer to yourself if you are the issuer.

PR #2669, opened 2018-04-30
Draft of eosio.unregd to keep unreg'd Ethereum addr + EOS balances on chain

abourget:
The goal would be to all the community to inject the unregistered snapshot, left on chain.. destroy the authority for the `eosio.unregd` account, and leave that for the future.

Maybe later, a worker proposal and some technical hiccups are alleviated and we can have the owners of the Ethereum address come prove their ownership and release their balance from the `eosio.unregd` contract (which would hold the funds in the interim).

### DIFF
--- a/contracts/eosio.unregd/eosio.unregd.cpp
+++ b/contracts/eosio.unregd/eosio.unregd.cpp
@@ -1,0 +1,46 @@
+#include <algorithm>
+
+#include "eosio.unregd.hpp"
+
+using eosio::unregd;
+
+EOSIO_ABI(eosio::unregd, (add))
+
+/**
+ * Add a mapping betwen an ethereum_account and an initial EOS token balance.
+ */
+void unregd::add(const ethereum_account& ethereum_account, const asset& balance) {
+  require_auth(_self);
+
+  auto symbol = balance.symbol;
+  eosio_assert(symbol.is_valid() && symbol == EOS_SYMBOL, "balance must be EOS token");
+
+  // TODO: What to do with existing mapping? Replace? Error?
+  auto itr = find_ethereum_account(ethereum_account);
+  if (itr == accounts.end()) {
+    accounts.emplace(_self, [&](auto& account) {
+      account.id = accounts.available_primary_key();
+      account.ethereum_account = ethereum_account;
+      account.balance = balance;
+    });
+  } else {
+    accounts.modify(itr, _self, [&](auto& account) {
+      account.ethereum_account = ethereum_account;
+      account.balance = balance;
+    });
+  }
+}
+
+/**
+ * For now, in `multi_index`, there is no possibility to create a secondary index
+ * based on a `std::string` or a fixed sized array (unless a key256). As such,
+ * for now, let's do a poor man search to find a matching account.
+ *
+ * TODO: Use a key256 type instead. The `ethereum_account` (40 hex characters,
+ * 80 bytes) would then be the 80 first bytes of the key and the rest would be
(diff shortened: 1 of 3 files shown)


contracts/eosio.unregd/eosio.unregd.cpp
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Add a mapping betwen an ethereum_account and an initial EOS token balance.
+ */
+void unregd::add(const ethereum_account& ethereum_account, const asset& balance) {
+  require_auth(_self);
+
+  auto symbol = balance.symbol;
+  eosio_assert(symbol.is_valid() && symbol == EOS_SYMBOL, "balance must be EOS token");
+
+  // TODO: What to do with existing mapping? Replace? Error?
```
dskvr: Snapshot generator returns distinct ethereum addresses.

contracts/eosio.unregd/eosio.unregd.cpp
```diff
@@ -0,0 +1,46 @@
...
+EOSIO_ABI(eosio::unregd, (add))
+
+/**
+ * Add a mapping betwen an ethereum_account and an initial EOS token balance.
+ */
+void unregd::add(const ethereum_account& ethereum_account, const asset& balance) {
+  require_auth(_self);
+
+  auto symbol = balance.symbol;
+  eosio_assert(symbol.is_valid() && symbol == EOS_SYMBOL, "balance must be EOS token");
```
kesar: CORE_SYMBOL

contracts/eosio.unregd/eosio.unregd.cpp
```diff
@@ -0,0 +1,46 @@
+#include <algorithm>
+
+#include "eosio.unregd.hpp"
+
+using eosio::unregd;
+
+EOSIO_ABI(eosio::unregd, (add))
+
+/**
+ * Add a mapping betwen an ethereum_account and an initial EOS token balance.
```
kesar: small typo: between

contracts/eosio.unregd/eosio.unregd.hpp
```diff
@@ -0,0 +1,51 @@
...
+ public:
+  unregd(account_name contract_account)
+      : eosio::contract(contract_account), accounts(contract_account, contract_account) {}
+
+  // Actions
+  void add(const ethereum_account& ethereum_account, const asset& balance);
+
+ private:
+  static const uint64_t EOS_PRECISION = 4;
+  static const asset_symbol EOS_SYMBOL = S(EOS_PRECISION, EOS);
```
kesar: SYS

abourget:
Hey folks, when I use this contract, each "add" calls slows down and down and down until I get some timeouts.

Is there a way to write something more performance? Something I didn't catch about augmenting a list?

In this state, we cannot use it for launch. 

Please advise.

abourget:
Right, we're doing a poor man's search here. What would be a better solution?

brianjohnson5972:
@abourget since there is no string secondary key lookup, I think your best bet would be to create a hash of the string and store it in one of the integer secondary key types. See contracts/multi_index_test for examples of using secondary indexes.

abourget:
Ok, here's what we've done: https://github.com/eoscanada/eos-bios/tree/master/eosio.unregd 

this one performs a lot better.. and all info is on chain for the future. The `eos-bios` loads it all up.. and transfers EOS to that account, for future use.

heifner:
This appears to have moved to its own repo: https://github.com/eoscanada/eos-bios
Thanks.

PR #2683, opened 2018-05-01
EOSIO build darwin, home-brew installing boost 1.66

pacificcode:
darwin: added homebrew installing boost back into darwin build.


scripts/eosio_build_darwin.sh
```diff
@@ -209,32 +212,31 @@
...
-		./b2 install
-		if [ $? -ne 0 ]; then
-			printf "\n\tInstallation of boost libraries failed. 1\n"
-			printf "\n\tExiting now.\n"
-			exit 1
+		if [ ! -d "/usr/local/Cellar/boost/1.66.0" ]; then
+			if ! brew install "https://raw.githubusercontent.com/Homebrew/homebrew-core/76a83d798d76aa43459ceb145b721ba4a16a54a6/Formula/boost.rb"
+			then
+				printf "\tUnable to install boost 1.66 libraries at this time. 0\n"
+				printf "\tExiting now. 0\n\n"
```
spoonincode: Are these debug numbers being left in deliberately?

pacificcode: Yes, but good catch. Only the first printf should have the 0. It's not an absolute requirement that  these are left in there, but it does save time when users open issues.

Issue #2628, opened 2018-04-27
Test scheduled trxn from non-priviledged accounts


gleehokie:
The unit tests currently do not test scheduled transactions from non-privileged accounts.

andriantolie:
Beside PR 2654, another PR 2700 is added here

heifner:
Test added via #2654

PR #2700, opened 2018-05-02
Additional unit test for #2628 non-priviledged deferred transaction

andriantolie:
This is another addition to the unit test for #2628. 

This unit test proofs that a deferred transaction which is authorized using a non-priviledged account's authority that has a delay, has its delay checked properly.


heifner:
Please fix conflicts.

andriantolie:
PR Updated

andriantolie:
PR is updated to fix buildkite issue with fedora, aws, and other linux distribution

Issue #2694, opened 2018-05-02
Updating new code to accounts that already have code will cause memory leak.


learnforpractice:
Possible solution:
 Check whether the code is only owned by one account, if it does, them after the account's code has been updated, the old code which cached in instantiation_cache should be delete.

heifner:
Addressed by recent pull requests.

PR #2701, opened 2018-05-02
Fix memory leak in instantiation_cache

learnforpractice:


learnforpractice:
https://github.com/EOSIO/eos/issues/2694

learnforpractice:
In order to break the module cache check, eosio.token need a small hack, placing the following code in eosio.token.cpp, testing code will replace "hello,world" with other characters in order to workaround the moduel cache check.

```
void sayHello() {
   prints("hello,world");
}
```

Running the following testing code under build/contracts
```
import os
import shlex
import subprocess

def toname(i):
    return chr(ord('A')+int(i%10)) + chr(ord('A')+int((i/10)%10)) + chr(ord('A')+int((i/100)%10))

cleos = './../programs/cleos/cleos'

create_account_cmd = '{0} create account eosio eosio.token EOS7ijWCBmoXBi3CgtK7DJxentZZeTkeUnaSDvyro9dq7Sd1C3dC4 EOS7ijWCBmoXBi3CgtK7DJxentZZeTkeUnaSDvyro9dq7Sd1C3dC4'.format(cleos)
create_account_cmd = shlex.split(create_account_cmd)
ret = subprocess.call(create_account_cmd)
'''
if ret != 0:
    raise Exception('Creating account failed')
'''

wast = 'eosio.token/eosio.token.wast'
if not os.path.exists(wast+'2'):
    with open(wast, 'rb') as f:
        with open(wast+'2', 'wb') as f2:
            data = f.read()
            f2.write(data)

for i in range(1, 1000):
    key_words = b"hello,world"
    with open(wast+'2', 'rb') as f:
        data = f.read()
        #data.find(key_words)
        replace_str = b"a%d"%(i,)
        replace_str.zfill(len(key_words))
        #replace key works with custom words to break the effect of code cache mechanism
        print(data.find(key_words))
        data = data.replace(key_words, replace_str)
        with open(wast, 'wb') as f:
            f.write(data)

    ret = subprocess.call(['./../programs/cleos/cleos', 'set', 'contract', 'eosio.token', 'eosio.token',  '-p', 'eosio.token'])
    if ret != 0:
        raise Exception('Setting contract failed')

    ret = subprocess.call(['./../programs/cleos/cleos', 'push', 'action', 'eosio.token', 'create','[ "eosio", "%d.0000 %s", 0, 0, 0]'%(i,toname(i)), '-p', 'eosio.token'])
```

Expecting to see no dramatically memory increase during the running of testing code.

spoonincode:
Keeping all the versions of code was an explicit choice at the time. Calling @wanderingbort what do you think?

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -19,6 +19,12 @@ using namespace Runtime;
 
 namespace eosio { namespace chain {
 
+   struct account_module {
+      digest_type digest;
+      std::shared_ptr<wasm_instantiated_module_interface> module_interface;
```
wanderingbort: This doesn't seem to be used anymore, meaning the `map<account_name, account_module> account_module_map` can probably become a `map<account_name, digest_type>`

That would remove the need to change the `std::unique_ptr` to a `std::shared_ptr` as well which is nicer.

tests/tests/misc_tests.cpp
```diff
@@ -447,6 +451,80 @@ BOOST_AUTO_TEST_CASE(alphabetic_sort)
...
+
+   std::string code = wast_code;
+   code += "(call $test (call $i64_trunc_u_f64 (f64.const 1)))\n";
+   code += "))";
+
+   wasm_interface_impl wasm_imp(wasm_interface::vm_type::binaryen);
+
+   auto wasm = ::eosio::chain::wast_to_wasm(code);
+
+   auto wasm_code = shared_vector<char>(alloc);
```
wanderingbort: shared_vector is just a vector that has a different allocator.  For the purposes of your test it could just be a `vector<char>` which would make the inclusion of `bip::managed_mapped_file seg(bip::open_or_create,"./test.db", 1024*1024);` and all the related code to derive `alloc` unnecessary

tests/tests/misc_tests.cpp
```diff
@@ -447,6 +451,80 @@ BOOST_AUTO_TEST_CASE(alphabetic_sort)
...
+   (func $i64_trunc_u_f64 (param $0 f64) (result i64) (i64.trunc_u/f64 (get_local $0)))
+   (func $test (param $0 i64))
+   (func $apply (param $0 i64)(param $1 i64)(param $2 i64)
+   )=====";
+
+   std::string code = wast_code;
+   code += "(call $test (call $i64_trunc_u_f64 (f64.const 1)))\n";
+   code += "))";
+
+   wasm_interface_impl wasm_imp(wasm_interface::vm_type::binaryen);
```
spoonincode: this needs to be the runtime that the unit tests are targeting

learnforpractice:
wasm_interface_impl::get_instantiated_module has shared_vector<char> as it’s parameter, 

there is no way you can convert vector<char> to shared_vector directly.

So managed_mapped_file is just for the purpose of keep the code in wasm_interface_private.hpp unchanged, but still it can be tested.

learnforpractice:
Or can someone tell me how to convert vector<char> to shared_vector<char> nice and easy, so I can remove the ugly code :).

wanderingbort:
> module_interface in account_module is used for reference counting, so when no account reference to the code anymore,

This is true however, you never access `account_module::module_instance` from an iterator or entry in the `account_module_map` as far as I can tell.  

I can only find an access for `account_module::digest` which is used to find the entry `instantiation_cache`, decrement the reference count and potentially erase the entry.   

Basically, the mechanism is sound but I don't see a reason to share ownership of `wasm_instantiated_module_interface` between the two maps.  Only one is ever used to access that data and therefore it seems like this mechanism would still work as designed with `instantiation_cache` having the only ownership and using `std::unique_ptr`

wanderingbort:
> wasm_interface_impl::get_instantiated_module has shared_vector<char> as it’s parameter,

Good catch, I did not realize that.  I think its cleaner to have the storage semantics defined elsewhere. In the past, we've other places we've passed `const char* buf, size_t buf_size` in place of a `shared_vector<char>` and then made the caller extract those values from their preferred storage class before calling the method.  

If you want to take that approach, I would approve of it.   It adds a small amount of complexity to the caller and removes a large amount of complexity from your test code and all the called code.

learnforpractice:
@wanderingbort, I think you understand why I use share_ptr, My purpose is to tracking the reference of code used by accounts since the same code may be used by more than one account and once no account reference to the code in instantiation_cache anymore, the code need to be release to prevent instantiation_cache from memory leak. Since accounts can update their code with no times limit, this bug may be used by malicious attack and paralyses the network in a few hours.

Can you show me a piece of code on how to tracking the reference of code with map<account_name, digest_type>?

I can only find on way:  iterate through the entire map every time updating the code in account, and I think that's inefficient.

I will accept your suggestion if I can understand it’s a better solution.

learnforpractice:
@wanderingbort Nice work, it’s better than I thought. Patch done. I got stuck in my own mind. Also return a  reference to shared_ptr is actually unsafe.

spoonincode:
I think after some internal discussion a couple days we decided this was not what we wanted to do. Two reasons: 1) hard forks could throb cache invalidation more than desired and 2) we actually need the cache to additionally be aware of the code's "injection version"

learnforpractice:
Hmm, it’s the core part, good to hear that you guys have a better thought about improvement.

gleehokie:
@spoonincode should we close this PR?

bytemaster:
I would like to revisit this issue.  Fork thrashing could be resolved by expiring cache on IRB rather than immediately upon changing of code.  Lets resolve the merge conflicts.

spoonincode:
I wonder if the garbage collector in WAVM would need to be fired for the WAVM modules to actually be freed.

learnforpractice:
Sorry, I deleted my eos fork accidentally, Hope this issue will be solved soon.

PR #2704, opened 2018-05-02
fix #2626(CFA) with additional unittests

### LINKED ISSUE #2626: Restore CFA tests
- [x] validate CFA are executed at time of deferred not when executed
- [x] validate CFA don't have access to stateful APIs

taokayan:
Fix deferred context free action scheduling.
Add unittest cases for context free action with stateful api, deferred context free actions.


libraries/chain/transaction_context.cpp
```diff
@@ -114,7 +114,7 @@ namespace eosio { namespace chain {
    {
       trace->scheduled = true;
       is_input = false;
-      apply_context_free = false;
+      //apply_context_free = false; // does CFAs always execute immediately?
```
heifner: CFA should execute immediately and not at time of delay. slim has been updated and should be correct.

PR #2710, opened 2018-05-02
Cleanup, bug fixes, discounted billing, and progress towards better handling of soft failures

arhag:
Changes in this PR:

- Sanity checks in resource_limits_manager to avoid potential divide by zero errors.
- Cleanup of exceptions (removed unused exceptions) and disambiguation of the various types of resource exhaustion exceptions (including soft versions of exceeding net or cpu usage because the block cannot support any more transactions).
- Cleanup of `transaction_context` to avoid using such a heavy constructor that can throw exceptions.
- Throw soft versions of cpu and net usage exceeded exceptions if it was because the block cannot support any more cpu or network usage (but not because the transactions limits are met or the payers could not afford to pay). This will later be used by the caller of `push_transaction` or `push_scheduled_transaction` to understand how to properly deal with a transaction that wasn't successfully included in the block.
- Discounted network usage billing of pruned data (#2000).
- Reimplement discounted cpu usage billing of context-free actions (which was initially dropped in the transition from master to slim).
- Fixed auth_tests/linkauth_special (which is now added back into ctest), wasm_tests/eosio_abi, and bootseq_tests.


libraries/chain/transaction_context.cpp
```diff
@@ -9,26 +9,25 @@
...
-   ,deadline(d)
-   ,published(p)
    ,net_usage(trace->net_usage)
    ,cpu_usage(trace->cpu_usage)
    {
       trace->id = id;
-      trace_ptr = trace;
       executed.reserve( trx.total_actions() );
+      is_input = false;
+      apply_context_free = false;
```
heifner: This line should be removed. You are setting apply_context_free to false here and nowhere does it become true.

libraries/chain/apply_context.cpp
```diff
@@ -352,7 +355,8 @@ void apply_context::reset_console() {
 
 void apply_context::checktime(uint32_t instruction_count) {
    cpu_usage += instruction_count;
-   trx_context.add_cpu_usage_and_check_time( instruction_count );
+   EOS_ASSERT( BOOST_LIKELY(cpu_usage <= cpu_usage_limit), tx_cpu_usage_exceeded, "action cpu usage exceeded" );
```
heifner: I would argue that BOOST_LIKELY (or the existing LIKELY of fc be part of EOS_ASSERT.

PR #2737, opened 2018-05-03
Change eosiocpp linker ordering

spoonincode:
Make dependent libraries go first.. as is tradition. Though I'm a little surprised llvm-link needed this. Resolves problem internally found by andrianto


lyjstudy:
#2564 fix eosiocpp link

PR #2740, opened 2018-05-03
fixes for dpos3

wanderingbort:


libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -22,10 +23,6 @@ struct block_header_state {
     public_key_type                   block_signing_key;
     vector<uint8_t>                   confirm_count;
```
heifner: Move confirmations up to here.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -60,7 +57,7 @@ struct block_header_state {
 } } /// namespace eosio::chain
 
 FC_REFLECT( eosio::chain::block_header_state,
-            (id)(block_num)(header)(dpos_last_irreversible_blocknum)
+            (id)(block_num)(header)(dpos_irreversible_blocknum)
             (pending_schedule_lib_num)(pending_schedule_hash)
             (pending_schedule)(active_schedule)(blockroot_merkle)
             (producer_to_last_produced)(block_signing_key)
```
heifner: Missing confirm_count. Also I would put these in the same order as declared.

libraries/chain/block_header_state.cpp
```diff
@@ -123,13 +106,28 @@ namespace eosio { namespace chain {
...
+          new_producer_to_last_produced[pro.producer_name] = result.dpos_irreversible_blocknum;
         }
       }
       result.producer_to_last_produced = move( new_producer_to_last_produced );
       result.producer_to_last_produced[prokey.producer_name] = result.block_num;
     }
 
+    /// grow the confirmed count
+    uint8_t required_confs = (result.active_schedule.producers.size() * 2 / 3) + 1;
+    if( confirm_count.size() < 1024 ) {
```
heifner: Why 1024, comment would be good.

libraries/chain/block_header_state.cpp
```diff
@@ -85,36 +85,19 @@ namespace eosio { namespace chain {
     result.block_num                             = block_num + 1;
     result.producer_to_last_produced             = producer_to_last_produced;
     result.producer_to_last_produced[prokey.producer_name] = result.block_num;
-//    result.dpos_last_irreversible_blocknum       = result.calc_dpos_last_irreversible();
+//    result.dpos_irreversible_blocknum       = result.calc_dpos_last_irreversible();
```
heifner: Can we remove this commented out code.

luomeiqin:
I have a question that why eos changes last_irreversible_block_num from bft_irreversible_blocknum to this ?
uint32_t controller::last_irreversible_block_num() const {
return std::max(my->head->bft_irreversible_blocknum, my->head->dpos_irreversible_blocknum);
 }
i was suppose to take bft_irreversible_blocknum is the LIB，after i go thought this issue # #3088

BottosGroup:
This is also a problem for us

raymondnuaa:
@luomeiqin 
Current EOS only updates dpos_irreversible_blocknum, the bft_irreversible_blocknum should be for realtime BFT. And please note there is another variable dpos_proposed_irreversible_blocknum, which is used to implemented the two rounds confirmation of BFT. (Related functions: set_confirmed and calc_dpos_last_irreversible()).

PR #2741, opened 2018-05-03
Fix eosio_token_tests/transfer_tests and auth_tests/link_then_update_auth

arhag:
Fix eosio_token_tests/transfer_tests.
Fix auth_tests/link_then_update_auth.
Remove unneeded test case from currency_tests/test_input_quantity.

### DIFF
--- a/unittests/eosio.token_tests.cpp
+++ b/unittests/eosio.token_tests.cpp
@@ -299,7 +299,7 @@ BOOST_FIXTURE_TEST_CASE( transfer_tests, eosio_token_tester ) try {
       transfer( N(alice), N(bob), asset::from_string("-1000 CERO"), "hola" )
    );
 
-   BOOST_REQUIRE_EQUAL( error( "condition: assertion failed: attempt to subtract asset with different symbol" ),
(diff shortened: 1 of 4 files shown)


unittests/eosio.token_tests.cpp
```diff
@@ -299,7 +299,7 @@ BOOST_FIXTURE_TEST_CASE( transfer_tests, eosio_token_tester ) try {
       transfer( N(alice), N(bob), asset::from_string("-1000 CERO"), "hola" )
    );
 
-   BOOST_REQUIRE_EQUAL( error( "condition: assertion failed: attempt to subtract asset with different symbol" ),
+   BOOST_REQUIRE_EQUAL( error( "condition: assertion failed: asset symbol has higher precision than expected" ),
```
brianjohnson5972: Why are we expecting the error message for too high of precision?  It sure seems like a mismatch with the data that is being passed.

arhag: With PR #2735, transfer should accept different precisions as long as the symbol name is the same and the precision of the input is no greater than the precision defined for the asset. You can see in currency_tests/test_input_quantity how lower precision assets with the same symbol name work just fine with transfer now.

PR #2750, opened 2018-05-04
non producing node support

wanderingbort:
- removed unused runtime_limits from controller 
 - added 3 new methods for chain::plugin_interface
    - sync incoming trx/block that should be used for legacy applications that need the exceptions/traces
    - a start_coordinator method
 - these new methods will only route to a single provider based on priority
 - migrated producer plugin to provide these 3 methods at a high priority
 - created validator plugin to provide these 3 methods at a low priority
 - validator plugin is default initialized, producer is no longer
 - validator plugin will start blocks and abort/restart as blocks come in

### DIFF
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -34,18 +34,12 @@ namespace eosio { namespace chain {
    class controller {
       public:
          struct config {
-            struct runtime_limits {
-               fc::microseconds     max_push_block_us = fc::microseconds(100000);
-               fc::microseconds     max_push_transaction_us = fc::microseconds(1000'000);
-            };
-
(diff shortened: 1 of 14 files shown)


wanderingbort:
I suppose I should explain how this works:

what used to be the producer_plugin's plugin_startup lifecycle hook is now the implementation that backs its provider for the `start_coordinator` appbase::method.  It registers this, as well as its providers for synchronous transaction and block injestion at priority 1 (low is higher).

a new validator plugin registers providers for all of these things at priority 100.

When the producer is present and initialized, the validators implementation is not invoked.  As a result, it is dormant.

However, without a higher priority implementation of these methods, it kicks in 

the `start_coordinator` appbase::method is invoked at the end of chain_plugin's plugin_startup effectively selecting the right plugin to start in a loosely coupled way

Issue #2630, opened 2018-04-27
Testing Upgrade Hooks Testing


gleehokie:
There are two intrensic apis:
```
activate_feature(name)
has_feature(name)
```

Has feature should only return true once the block that called activate_feature becomes irreversible.

Node should shutdown automatically if it attempts to switch to an irreversible block that it considers invalid.

heifner:
We still need to test an actual hard fork on a live network.

taokayan:
hi @heifner, I'm wondering do we also need to expose some API to get current version of the blockchain in the contract? Since a hard forked blockchain might support the same API but executes in a different way.

heifner:
@taokayan Probably not at this time.

PR #2756, opened 2018-05-04
add unitest cases for new api feature

taokayan:
Related to #2630


heifner:
This PR does not resolve #2630 .

Issue #2738, opened 2018-05-03
Add buyram to create account in cleos


heifner:
Ram is now required for account usage.
cleos needs to buyram for new account when new account is created.
Default to 8K ram.

heifner:
Resolved by #2771 and #2738.

Issue #2755, opened 2018-05-04
Can not create any account / set contract in slim


taokayan:
Create any account (even with creator eosio) in slim branch results in the following error:

```
58489ms thread-0   chain_api_plugin.cpp:79       operator()           ] Exception encountered while processing chain.push_transaction: 13 St9exception: unknown key
unknown key: 
    {"name":"","what":"unknown key"}
    thread-0  controller.cpp:1270 get_account
Failed to serialize type
    {"object":{"id":"363e6b09c9497b2a360672dfe4ce7f717f4e277d5e7a61d893cf1b19454ab1ec","receipt":{"status":"hard_fail","kcpu_usage":0,"net_usage_words":0},"elapsed":0,"net_usage":228,"cpu_usage":103232,"scheduled":false,"action_traces":[{"receipt":{"receiver":"","act_digest":"0000000000000000000000000000000000000000000000000000000000000000","global_sequence":0,"recv_sequence":0,"auth_sequence":[]},"act":{"account":"","name":"","authorization":[],"data":""},"elapsed":0,"cpu_usage":0,"console":"","total_cpu_usage":0,"trx_id":"0000000000000000000000000000000000000000000000000000000000000000","inline_traces":[]}],"failed_dtrx_trace":null,"hard_except":{"code":3080006,"name":"tx_deadline_exceeded","message":"**transaction took too long**","stack":[{"context":{"level":"error","file":"transaction_context.cpp","line":306,"method":"check_time","hostname":"","thread_name":"thread-0","timestamp":"2018-05-04T07:00:58.488"},"format":"deadline exceeded","data":{}}]}}}
    thread-0  abi_serializer.hpp:482 to_variant

```

I guess this is because checktime is being done for every action even for system accounts.

If "checktime( cfg.base_per_action_cpu_usage )" in apply_context::exec_one() is commented out, account creation can be done smoothly, but "setcode" still failed for the same reason as follows:

```
3556804ms thread-0   chain_api_plugin.cpp:79       operator()           ] Exception encountered while processing chain.push_transaction: 13 St9exception: unknown key
unknown key: 
    {"name":"","what":"unknown key"}
    thread-0  controller.cpp:1270 get_account
Failed to serialize type
[...]

PR #2757, opened 2018-05-04
Gh#2738 buyram newaccount

heifner:
Add buyram or buyrambytes to newaccount transaction in cleos.
Specifying a buyram of 0 will skip the addition of buyram to the newaccount transaction so that create account can be used without the eosio.system and eosio.token account being loaded. Otherwise eosio.system and eosio.token (created and issued) must be loaded.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 *.dot
 *.abi.hpp
 *.cmake
+*.ninja
 \#*
 \.#*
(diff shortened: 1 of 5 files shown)


gleehokie:
Related to #2755 @taokayan @ChristianDunst

heifner:
Resolves #2738

PR #2759, opened 2018-05-04
Changes to onerror and improvements to transaction_trace

### LINKED ISSUE #2671: New onerror handling is spoofable
Previously, the `onerror` action was explicitly part of the `eosio` contract space and was re-delivered to the original sender of a deferred transaction using the normal recipient actions.

In addition to having the "nice" quality of not requiring a naming convention on contracts for system reasons, this also meant that it was not possible for a user-signed transaction to publish something to the `onerror` handler for a contract without the explicit consent of the system contract (which would probably never be given in practice).

Currently, contract authors need to be aware that the contents of the `onerror` action are no longer trustworthy nor necessarily coming from the native system.  This will increase the cost associated with a safe `onerror` action handler and increase the likelihood that soft-errors become hard-errors due to objective resource limits.

arhag:
Changes in this PR:

* Fixed move constructor of fc::scoped_exit.
* Use `scoped_exit` to restore pending block state in case of failure during transaction execution.
* Removed obsolete `deferred_transaction` from eosiolib.
* `eosio::onerror` has gone back to using the `eosio` account scope which allows contracts to know it is a legitimate onerror and not an action sent by a user (this resolves #2671). However, the `eosio::onerror` is being dispatched directly to the sender account (rather than to `eosio` which then notifies the sender as it did it master).
* The `publication_time()` called in the error handler will now correctly provide the original publication time of the failed deferred transaction.
* Removed obsolete native handler for eosio::onerror. System contract WASM can still handle the onerror action for deferred transactions it sends (such as in eosio::postrecovery).
* Improvements to `transaction_trace` including: `receipt` in `transaction_trace` is now optional; the trace now has just an `except` and `except_ptr` without the unnecessary soft and hard distinctions; and the trace should include the latest `action_trace` including the action that caused transaction failure.
* Transaction validation changes:
  + Avoid unnecessary check of expiration for generated transactions (instead just set the expiration similar to how we set the TaPoS). This also means that `test_transaction::send_transaction_expiring_late` can be removed.
  + Avoid duplicate (redundant) validation of referenced accounts when executing a deferred transaction.
  + Avoid duplicate (and erroneous) TaPoS validation when executing a deferred transaction.
  + Avoid unnecessary validations of TaPoS, expiration, and referenced accounts for implicit transactions which can be trusted to be correctly constructed by the controller.


libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -65,6 +62,6 @@ FC_REFLECT( eosio::chain::base_action_trace,
 FC_REFLECT_DERIVED( eosio::chain::action_trace,
                     (eosio::chain::base_action_trace), (inline_traces) )
 
-FC_REFLECT( eosio::chain::transaction_trace, (id)(receipt)(elapsed)(net_usage)(cpu_usage)(scheduled)(action_traces)
-                                             (failed_dtrx_trace)(soft_except)(hard_except) )
+FC_REFLECT( eosio::chain::transaction_trace, (id)(receipt)(elapsed)(net_usage)(cpu_usage)(scheduled)
+                                             (action_traces)(failed_dtrx_trace)(except) )
```
heifner: Need to add except_ptr

PR #2777, opened 2018-05-04
net and cpu window size parameters 

moskvanaft:


bytemaster:
Looks like some of the unit tests broke on this one?

moskvanaft:
Unit-test fails because value we add to resource usage is divided by window_size.
```
value_ex += units * Precision / (uint64_t)window_size;
```
Longer window leads to smaller value added. In function `average`, `value_ex` gets divided by Precission, and the result is 0 because `value_ex` is too small.
I don't quite understand what is a reason to divide here.
If I replace it with just 
```
value_ex += units * Precision;
```
the test passes.
@wanderingbort please advise.

P.S. meanwhile I am committing my change to the pull request to see if anything else will fail.

wanderingbort:
See commit be5db3a

The short answer is that we were correctly representing that value in average-micro-units (in this case micro bytes/window) however, given the long window and the implied 0 in all empty window buckets the average was somewhere between 0 micro-units and 1,000,000 micro-units and was truncated to 0 on the call to `average()` despite the internal tracking being able to represent that non-zero value faithfully.

As a result, I have converted it over to conservative math.  When we accumulate in micro-units/time we will always round-up.  When we calculate an instantaneous value, we will also round-up (though the error this creates is not accumulated over time as that is a one-off "estimate" of the higher precision moving average).

Issue #2656, opened 2018-04-30
without "-j", it will report as the description.


yankunsam:
root@eosio:~/data# cleos get transactions eosio
813330ms thread-0   main.cpp:1764                 main                 ] Failed with error: Key Not Found (6)
Key data

Joelone:
Got the same error, here is more verbose output:

```
$ cleos -v get transactions user
3338195ms thread-0   main.cpp:1158                 main                 ] Failed with error: 6 key_not_found_exception: Key Not Found
Key data
    {"key":"data"}
    thread-0  variant_object.cpp:90 operator[]

```

Joelone:
It is duplicated of #2567 
https://github.com/EOSIO/eos/issues/2567

yankunsam:
Thanks

PR #2804, opened 2018-05-06
cleos: fix issue #2656 transactions error

Joelone:
- rename "data" to "transaction" based on "-j" output
- reuse get_transactions_func call result for output

Signed-off-by: Joelone <joeloneplusone@gmail.com>


bytemaster:
Not merging into master at this stage.

PR #2809, opened 2018-05-07
Major changes to check_authorization and permission_api (delay_tests now working)

arhag:
Changes in this PR:

* Redesign of `check_authorization`:
  - There are two variants of `check_authorization`: one checks if the provided authorizing factors satisfy the authorization requirements of a vector of actions; the other checks if the provided authorizing factors satisfy the authorization requirements of a permission.
  - Both variants take some combination of authorizing factors: public keys, permissions, and a satisfied delay. Note that the previous version of `check_authorization` had both a `provided_levels` and a `provided_accounts` (which satisfied any permission level of the provided accounts). These have now been replaced by a `provided_permissions` which is a `flat_set<permission_level>` very similar to the old `provided_levels`, except that the permission name in the `permission_level` can be empty to signify a wildcard permission, thus replicating the behavior of the old `provided_accounts`.
  - Both variants return the minimum delay that needed to be satisfied for the particular authorization check if the authorization check was successful. The functions will throw if the authorization check was not successful.
  - The redesign of `check_authorization` now permits a `checktime` to be passed into `authority_checker` so that the cost of the authorization checking algorithm can be properly billed objectively as CPU usage.
  - `check_authorization` internally calls `authority_checker::satisfied` on each permission that must be checked. Each top-level invocation of `authority_checker::satisfied` creates and manages (only for the lifetime of that `authority_checker::satisfied` call) a cache  of permissions it has already visited and whether they were satisfied or not. This optimizes the authorization checking algorithm. However, this cache is not kept across the multiple permission checks that may usually be done in the first variant of `check_authorization` which takes a vector of actions. The lack of this extra optimization was intentional in order to permit each permission check to be done in parallel while maintaining the goal of deterministic objective CPU billing for authorization checking.
  - The new authorization checking algorithm ignores permissions with delays higher than the provided delay. This fixes a quirk in the previous algorithm in which it _should_ have been possible to satisfy a permission with some particular delay, but the old algorithm would require a higher delay because of the way it visited permissions.
  - The new authorization checking algorithm does not bother going down the path of permissions which it is already in the process of evaluating. Such cyclic permission graphs should be avoided by the user anyway.
  - The new authorization checking algorithm does not consider the delay of the permissions encountered walking up the permission tree from the minimum linked permission of the action (inclusive) to declared permission in the authorizations of the action (exclusive). In other words, with the exception of the special case native actions that may modify the imposed delays, only the delays encountered in trying to satisfying the declared permissions of the transaction contribute towards the minimum imposed delay for the transaction.
* Redesign of permission_api:
  - Previously there were two intrinsics related to checking authorizations: `check_auth` and `check_authorization`. `check_auth` took a transaction and permissions and failed if the authorization was not checked. `check_authorization` took a permission and a set of keys and returned true or false depending on whether the authorization succeeded or failed. These intrinsics (which are now removed in this PR) had inconsistent behavior and were not as flexible as the two new intrinsics that now take their place in this PR: `check_transaction_authorization` and `check_permission_authorization`.
  - `check_transaction_authorization` takes a serialized transaction, a serialized set of public keys, and a serialized set of permission levels. It returns -1 if the authorization of the transaction was not satisfied, otherwise it returns the minimum delay (in microseconds) that needed to be satisfied for the particular authorization check (the same thing `check_authorization` returns). (Note that the authorization takes into account the `delay_sec` field of the transaction as a satisfied delay for the purposes of the authorization check.)
  - `check_permission_authorization` takes a permission level (in the form of two arguments specifying the account name and permission name), a serialized set of public keys, a serialized set of permission levels, and a provided delay in microseconds. It also returns -1 if the authorization of the permission was not satisfied, otherwise it returns the minimum delay (in microseconds) that needed to be satisfied for the particular authorization check.
  - For both intrinsics, either the serialized set of public keys or the serialized set of permission levels can be left empty (the corresponding size argument should be 0).
  - The eosio.msig contract has been updated to use the new `check_transaction_authorization` intrinsic. And with that change, it now respects the `delay_sec` field of the proposed transaction as part of the authorization check. (It also has been modified to disallow executing the proposed transaction if the expiration field is less than the current time.)
* delay_tests now passes and has been enabled in ctest.
* Checking expiration, TaPoS, and valid referenced accounts of an input transaction now occur prior to other `transaction_context` initializations (which mostly dealt with figuring out bandwidth billing details). This was primarily done to provide more meaningful error messages on an invalid transaction (particular when an authorizing actor account does not exist).


libraries/chain/authorization_manager.cpp
```diff
@@ -308,31 +316,40 @@ namespace eosio { namespace chain {
...
+   fc::microseconds
+   authorization_manager::check_authorization( const vector<action>&                actions,
+                                               const flat_set<public_key_type>&     provided_keys,
+                                               const flat_set<permission_level>&    provided_permissions,
+                                               fc::microseconds                     provided_delay,
+                                               const std::function<void(uint32_t)>& _checktime,
+                                               bool                                 allow_unused_keys
+                                             )const
    {
+      auto _noop_checktime = std::bind(&noop_checktime, std::placeholders::_1);
```
wanderingbort: Is it necessary for this to be a member function that is bound as opposed to a static/global `std::function<void(uint32_t)>` constructed with a simple lambda like `[](uint32_t){}`?

Issue #2803, opened 2018-05-06
Crazy Network Crash


bytemaster:
Started one node with:
```
./nodeos -e --producer eosio
...
1508503ms thread-0   producer_plugin.cpp:415       block_production_loo ] Produced block 00000180e0d4c715... #384 @ 2018-05-06T21:25:08.500 signed by eosio [trxs: 0, lib: 384, confirmed: 0]
1509003ms thread-0   producer_plugin.cpp:415       block_production_loo ] Produced block 00000181040d98a9... #385 @ 2018-05-06T21:25:09.000 signed by eosio [trxs: 0, lib: 385, confirmed: 0]
1509005ms thread-0   producer_plugin.cpp:124       on_incoming_block    ] Received block 000001821eca8981... #386 @ 2018-05-06T21:25:09.500 signed by eosio [trxs: 0, lib: 386, confirmed: 0]
1509005ms thread-0   producer_plugin.cpp:347       schedule_production_ ] 10 assert_exception: Assert Exception
when > header.timestamp: next block must be in the future
    {}
    thread-0  block_header_state.cpp:71 generate_next
1509005ms thread-0   producer_plugin.cpp:367       schedule_production_ ] Failed to start a pending block, will try again later
Segmentation fault: 11
```

And another with
```
./nodeos -d node2 --p2p-peer-address 127.0.0.1:9876 --p2p-listen-endpoint 127.0.0.1:6789
...
1508503ms thread-0   producer_plugin.cpp:124       on_incoming_block    ] Received block 00000180e0d4c715... #384 @ 2018-05-06T21:25:08.500 signed by eosio [trxs: 0, lib: 384, confirmed: 0]
1509004ms thread-0   producer_plugin.cpp:124       on_incoming_block    ] Received block 00000181040d98a9... #385 @ 2018-05-06T21:25:09.000 signed by eosio [trxs: 0, lib: 385, confirmed: 0]
1509004ms thread-0   producer_plugin.cpp:415       block_production_loo ] Produced block 000001821eca8981... #386 @ 2018-05-06T21:25:09.500 signed by eosio [trxs: 0, lib: 386, confirmed: 0]
1509005ms thread-0   producer_plugin.cpp:347       schedule_production_ ] 10 assert_exception: Assert Exception
when > header.timestamp: next block must be in the future
    {}
    thread-0  block_header_state.cpp:71 generate_next
[...]

PR #2822, opened 2018-05-07
Fixes EOSIO/eos#2803

wanderingbort:
- use the correct base time when calculating the minimum time to the next block
- prevent bad access of members of `pending` if `start_block` throws by cleaning up `pending` on the way out
- prevent bad access of vector when fulfilling a fork request results in no blocks to send

resolves EOSIO/eos#2803

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -301,7 +301,7 @@ bool producer_plugin_impl::start_block() {
    // If we would wait less than 50ms (1/10 of block_interval), wait for the whole block interval.
    fc::time_point now = fc::time_point::now();
    fc::time_point base = std::max<fc::time_point>(now, chain.head_block_time());
-   int64_t min_time_to_next_block = (config::block_interval_us) - (now.time_since_epoch().count() % (config::block_interval_us) );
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -301,7 +301,7 @@ bool producer_plugin_impl::start_block() {
    // If we would wait less than 50ms (1/10 of block_interval), wait for the whole block interval.
    fc::time_point now = fc::time_point::now();
    fc::time_point base = std::max<fc::time_point>(now, chain.head_block_time());
-   int64_t min_time_to_next_block = (config::block_interval_us) - (now.time_since_epoch().count() % (config::block_interval_us) );
+   int64_t min_time_to_next_block = (config::block_interval_us) - (base.time_since_epoch().count() % (config::block_interval_us) );
```
arhag: Although this isn't a problem with your change, this line made me realize that the code only works if our block timestamps are always divisible by by the block interval, which it currently does.

It would be nice if you added a static_assert to ensure that `config::block_timestamp_epoch` (which should probably be renamed to `config::block_timestamp_epoch_ms` to be clear that it is in milliseconds since the Unix epoch) is divisible by `config::block_interval_ms`.

Issue #2835, opened 2018-05-08
Shorter branch may take over much longer branch


taokayan:
An unexpected scenario in slim branch may happen in which the shorter branch takes over a (much) longer branch.

As illustrated below:
![lastirrnumber](https://user-images.githubusercontent.com/37097018/39747260-5a4a56d4-52df-11e8-8d23-46ad3d14259b.png)
4 producers A,B,C,D are producing in turns. After block 12, forks happen and result in 3 different branches: Branch A has producer A, branch B has producer B, branch C has producers C & D. From A's point of view, its last irreversible block number is 9. However, C&D's point of view, their last irreversible block number is 6. After then, branch A rejoined with branch C, C&D think that branch A has a bigger last irreversible number and switch into the shorter branch.

I'm wondering whether this is a design issue or something expected.

taokayan:
test case created in pull request #2839

wanderingbort:
This particular scenario's result is intended.  To understand why this is the case and, why it still holds that irreversibility is an optimization of pure DPOS (where longest chain wins) we have to play the scenario forward in time.  

What 2/3 + 1 producers on a given fork means is that, without detectable bad behavior, it should *eventually* produce the longest chain.  This is simple to understand because there can be at most 1/3 - 1 competitive producers on the minority fork and no matter what their lead is at some point in time, the 2/3 + 1 producers will eventually win.  Technically this is true with 1/2 + 1 vs 1/2 - 1 but for other reasons, the tolerance for DPOS is 2/3 + 1.

In your example we have 3 candidate forks: 
 - A in an isolated partition
 - B in an isolated partition
 - C/D in an isolated partition

None of these forks have 2/3 + 1 producers so, we don't have a clear winner for longest DPOS chain yet.  C/D is producing blocks faster _for now_ but it is not clear that another chain can not overtake it in the future.  

The rules are in place to make the best guess at what will be the longest chain in the end.  To review those rules are:

 - pick the fork with the most irreversible block 
 - break ties with the longest chain on top of that

So, when C/D reconnect to A they must decide which fork to support.  
 - Option 1) is their minority fork for which they can only prove that they are aware of (2 producers)
 - Option 2) is A's minority fork for which they can now prove that they AND A is aware of (3 producers)

[...]

PR #2839, opened 2018-05-08
create unitest to reproduce issue 2835

taokayan:
ref #2835 

Please uncomment the following line in prune_remove_branch test in forked_tests.cpp to reproduce the issue:

   //BOOST_REQUIRE_EQUAL(87, c.control->head_block_num());


bytemaster:
I'm going to merge this test except that I will change the expected head block number to the one it produces, 87 is not the proper behavior.

Issue #2625, opened 2018-04-27
Migrate net plugin to slim


gleehokie:

pmesnier:
Paul and I are working this.

pmesnier:
working branch for this issue is: slim_net_plugin

wanderingbort:
Once #2852 merges in there are a few remaining observational issues (not sure if they are slim specific or in master as well)

In a test last night with a single relay hub with 2 producing spokes and 1 validating spoke, syncing a 4th validating node by directly peering with either producer or the relay would cause blocks to be dropped.  This lead to the discovery of re-broadcasted transactions which should be fixed but suggests that the queues can get deep enough for low-priority messages (like syncing a new peer)                 to affect the deadlines of higher priority messages (relaying realtime blocks).

Another issue, which is unrelated to net_plugin but may affect testing, is that currently "subjective failures" in `controller` transaction reception can destroy some of the state that should be retained for the next block.  Until that is in place we cannot safely use the deadlines so, a sufficient transaction backlog can build up and cause timing issues.   This is being tracked in #2853

PR #2852, opened 2018-05-08
fixes for transaction re-broadcast and enable-stale-production timing

wanderingbort:
- added compat signal which mimics the on_pending_transaction signal from master
  - this signal is fired whenever an incoming signal is accepted or rejected and indicates the result
- moved check for stale_production to the block reception code since that now flows through producer_plugin


wanderingbort:
related to #2625

PR #2879, opened 2018-05-09
Eosio build darwin: extensive validation

pacificcode:
Added check for cmake cache file in eos/ doc root.
Added extensive command validation error checking in darwin build.

### DIFF
--- a/eosio_build.sh
+++ b/eosio_build.sh
@@ -35,6 +35,10 @@
 		printf "\\n\\tPlease cd into directory %s to run this script.\\n \\tExiting now.\\n\\n" "${CWD}"
 		exit 1
 	fi
+	if [ -f "${PWD}/CMakeCache.txt" ]; then
+		printf "\\n\\tPlease remove file %s/CMakeCache.txt before building EOS.IO.\\n \\tExiting now.\\n\\n" "${PWD}"
+		exit 1
+	fi
 
    	function usage()
    	{ 
--- a/scripts/eosio_build_darwin.sh
+++ b/scripts/eosio_build_darwin.sh
@@ -1,77 +1,81 @@
-	OS_VER=`sw_vers -productVersion`
-	OS_MAJ=`echo "${OS_VER}" | cut -d'.' -f1`
-	OS_MIN=`echo "${OS_VER}" | cut -d'.' -f2`
-	OS_PATCH=`echo "${OS_VER}" | cut -d'.' -f3`
+	OS_VER=$(sw_vers -productVersion)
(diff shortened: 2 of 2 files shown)


eosio_build.sh
```diff
@@ -35,6 +35,10 @@
 		printf "\\n\\tPlease cd into directory %s to run this script.\\n \\tExiting now.\\n\\n" "${CWD}"
 		exit 1
 	fi
+	if [ -f "${PWD}/CMakeCache.txt" ]; then
+		printf "\\n\\tPlease remove file %s/CMakeCache.txt before building EOS.IO.\\n \\tExiting now.\\n\\n" "${PWD}"
+		exit 1
+	fi
```
heifner: I thought people used the eosio_build.sh script to rebuild their eosio.io after an update. This prevents eosio_build.sh from working for an existing checkout with a previous build.

scripts/eosio_build_darwin.sh
```diff
@@ -211,170 +214,302 @@
...
+		printf "\\tBoost 1.66.0 found at /usr/local.\\n"
 	fi
 
-	printf "\n\tChecking MongoDB C++ driver installation.\n"
-    if [ ! -e /usr/local/lib/libmongocxx-static.a ]; then
-		cd ${TEMP_DIR}
-		brew install --force pkgconfig
-		brew unlink pkgconfig && brew link --force pkgconfig
+	printf "\\n\\tChecking MongoDB C++ driver installation.\\n"
+    if [ ! -e "/usr/local/lib/libmongocxx-static.a" ]; then
```
heifner: Note the slim branch does not have a mongodb plugin and will likely not have one for June.

eosio_build.sh
pacificcode: This file is required in directory /path/to/eos/build/
If this file is located in the eos/ directory CMake will write its build files to the eos/ directory and not the build directory which will cause:
cd build/
make make -j"${CPU_CORE}"
to error out i.e.
make: *** No targets specified and no makefile found.  Stop.

scripts/eosio_build_darwin.sh
pacificcode: I have $( git branch | grep \* ) writing the current branch to stdout.
I can add a check in to install mongo when necessary. We need to define what that condition/branch is.

heifner: It will be master branch soon.

eosio_build.sh
heifner: Hmm. I build in source, but I guess build script does out of source. ok.

PR #2881, opened 2018-05-09
Add wait weights to authority, simplify check authorization, and contracts are now only granted the "eosio.code" virtual permission

arhag:
Changes in this PR:

* The authorization checking algorithm for the special case actions (updateauth, deleteauth, linkauth, unlinkauth, and canceldelay) has been updated to be consistent with respect to delays as the other actions (i.e. only the delay requirements of the declared authorizations need to be considered, not those of the permissions encountered in walking up the permission tree from the minimum permission to the permission in the declared authorization).
* Delays are now included in the authority in a similar way as keys and permission_levels; they have associated weights (wait weights).
* The check authorization algorithm is greatly simplified because it is no longer necessary to track max delay encountered in satisfying a permission (permission visitors have been removed). If a sufficient delay is not provided, then permissions may not be satisfied.
* The any producer permission of the producer account has been removed.
* Contracts are now only granted a virtual permission of "eosio.code". A contract no longer satisfies all permissions of its account by default. If a contract needs to be able send an inline action or deferred transaction and requires adding some specific permission to the authorizations of the action, then the authority of that permission must be set up in such a way that it can be satisfied by just the "eosio.code" permission.


libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -119,31 +139,51 @@ inline bool validate( const Authority& auth ) {
...
                   std::is_same<weight_type, uint16_t>::value &&
                   std::is_same<typename decltype(auth.keys)::value_type, key_weight>::value &&
-                  std::is_same<typename decltype(auth.accounts)::value_type, permission_level_weight>::value,
+                  std::is_same<typename decltype(auth.accounts)::value_type, permission_level_weight>::value &&
+                  std::is_same<typename decltype(auth.waits)::value_type, wait_weight>::value,
                   "unexpected type for threshold and/or weight in authority" );
 
-   if( ( auth.keys.size() + auth.accounts.size() ) > (1 << 16) )
+   if( ( auth.keys.size() + auth.accounts.size() + auth.waits.size() ) > (1 << 16) )
       return false; // overflow protection (assumes weight_type is uint16_t and threshold is of type uint32_t)
```
heifner: The assumptions of this comment could be static_assert'ed

arhag: The assumptions are static_assert'ed. See the lines directly above it starting at line 139.

heifner: So it is.

PR #2884, opened 2018-05-09
Eosio.token better errors

heifner:
Provide a better error message than "Unknown key" if someone tries to `issue` before `create`.


capthienthang:
[eth.pdf](https://github.com/EOSIO/eos/files/2002876/eth.pdf)

Issue #2794, opened 2018-05-06
Update permission usage so get_permission_last_used API works properly


arhag:
In the transition from master to slim, controller left out `update_permission_usage`. This means that a `permission_usage_object` is never created in the current slim code and `postrecovery` just uses the account creation date as the last used time. Effectively the current state of the slim code would mean that after 30 days since the time an account was created, the only safeguard against lost password recovery is a 7 day delay.

The slim code should be fixed to track permission usage. Clearly, our unit tests on this are not very good, since recovery_tests are currently passing with the faulty code; so that needs to be improved as well. We should also just review the whole recovery mechanism altogether given all the changes we have recently made to deferred transactions, authorization checks and delays.

bytemaster:
This test needs to be updated to validate that the new intrinsics get_permission_last_used() work properly and that we properly set the permission used timestamps.  

We removed the native 'postrecovery' messages in favor of WASM based implementations.

gleehokie:
@arhag and @brianjohnson5972 please coordinate, I think you are both working on this

brianjohnson5972:
If @arhag is working this, then please remove me and assign yourself.  @heifner pointed me to this since it wasn't marked as assigned.

arhag:
PR #2893 fixes the issue of the last used time of permissions not being updated. With that fix and the `get_permission_last_used` intrinsic, contracts have enough tools to be able to implement account recovery. We still need to reimplement account recovery as a WASM contract and of course add proper tests for it, but that is a separate issue (#2894). So I am closing this one.

PR #2893, opened 2018-05-09
Update last used time of permissions used to authorize an input transaction

arhag:
All permissions of all accounts track a last used time. This timestamp is updated to the pending block time at the end of an input transaction (not deferred transactions nor implicit transactions like onblock or onerror) for all permissions that were used to authorize the input transaction.

So now the `get_permission_last_used` works properly. Unit tests were also added to test this intrinsic as well as the `get_account_creation_time` intrinsic.

The `get_permission_last_used` intrinsic is needed for a contract (such as eosio.recovery) to implement an account recovery feature (see #2794).


eosioAdminmanager (commented):
total?

Issue #2853, opened 2018-05-08
Handle Subjective failure cases in controller gracefully


wanderingbort:
Right now, when a processing deadline is exceeded while pushing a generated or unapplied transaction that transaction is incorrectly deleted from the corresponding retained state.  This means that transactions which straddle a subjective boundary (like how many deferred transactions to play) get dropped from the node.

This makes those deadlines unsafe to use in practice and we should fix that.

PR #2895, opened 2018-05-09
#2853 subjective failure handling

wanderingbort:
This refactors the expected usage of the `controller` WRT unapplied and deferred transactions.

  - Subjective failures : failures due to exceeding block level resource limits or caller specified wall-clock deadlines
    - transactions and scheduled transactions that encounter any subjective failure case should not record a failure to the blockchain, nor destroy any internal state of controller on the way out (existence in the unapplied transactions map OR generated transactions multi index)
 - callers of `push(_scheduled)?_transaction` are now responsible for deciding if the subjective failure caused by the deadline they imposed is actionable.
 - if a caller wants to impose a maximum transaction time, they may do so.  If that deadline is passed then the *caller* is responsible for:
    - calling `drop_unapplied_transaction` to remove the transaction from the internal cache on `controller`
    - maintaining a blacklist of scheduled transactions it doesn't want to replay (as the generated transaction table is part of chain state and not to be modified by outside systems)

Our 2 callers right now are:
  - tester : in the automated case, this will call all the unapplied and scheduled transactions it can, rethrowing the exceptions encountered 
  - producer plugin : this caller maintains a blacklist and per-transaction deadlines as well as soft deadlines for block limits
    - the soft deadline for block limits does not cause the failure of the transaction
    - transactions that straddle a block will cause a new block to be created and then the transaction will be retried.  This maintains the invariants expected by synchronous callers of `accept_transaction` while respecting subjective limits and block deadlines.

Turns out many tests were only succeeding because tester used to swallow problems with deferred and unapplied transactions.  Those are fixed too.


wanderingbort:
oh, I also removed the illusion of async caused by `sync_push` .  There were many places in and outside of controller that would not have been safe for async methods.  As such any plumbing for it now was getting in the way.

PR #2915, opened 2018-05-10
test case to reproduce issue 2914

taokayan:


bytemaster:
This is an invalid / malformed test.  See proof in this test: be020eb68265cd2cd452b7e361c984577c4cd5c1

PR #2923, opened 2018-05-10
DRY failure_is_subjective

heifner:
Remove duplicate code


wanderingbort:
I considered this but I didn't pull the trigger for the reason that they are already diverging.  

I don't think it will ever make sense for controller to consider the deadline subjective. And in the future, I could see producer plugin considering many things as "subjective"

heifner:
Fair enough.

PR #2938, opened 2018-05-10
Add Stack Exchange to Readme

jordigoyanes:
very useful resource


gleehokie:
Thanks for the suggestion @jordigoyanes!  I added it to the README and the Wiki.

PR #2956, opened 2018-05-11
Zero out keys of inactive producers

zorba80:
* Zero out keys of producers who haven't produced any blocks in 24 hours.
* Added testing for inactive producers being deactivated and replaced.
* Added require_auth(N(eosio)) to onblock.

### DIFF
--- a/contracts/eosio.system/voting.cpp
+++ b/contracts/eosio.system/voting.cpp
@@ -83,7 +83,7 @@ namespace eosiosystem {
             _producers.modify( *it, 0, [&](auto& p) {
                   p.time_became_active = block_time;
                });
-         } else if (  block_time > 2 * 21 * 12 + it->time_became_active &&
+         } else if ( block_time > 2 * 21 * 12 + it->time_became_active &&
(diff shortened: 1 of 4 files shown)


contracts/eosio.system/voting.cpp
```diff
@@ -78,8 +78,25 @@ namespace eosiosystem {
...
+         if ( it->time_became_active == 0 ) {
+            _producers.modify( *it, 0, [&](auto& p) {
+                  p.time_became_active = block_time;
+               });
+         } else if ( block_time > 2 * 21 * 12 + it->time_became_active &&
+                     block_time > it->last_produced_block_time + blocks_per_day ) {
+            _producers.modify( *it, 0, [&](auto& p) {
+                  p.producer_key = public_key();
+                  p.time_became_active = 0;
+                  p.last_produced_block_time = 0;
```
arhag: I think we can get rid of resetting `last_produced_block_time` without any problems right?

zorba80: Right. Setting last_produced_block_time to zero here is unnecessary.

PR #2979, opened 2018-05-12
Fixed dependency displaying code in ubuntu build script.

HomoEfficio:
Fixed broken message about dependency in ubuntu build script.

Before:
Imgur

After:
Imgur


pacificcode:
This issue was addressed by PR #2976.

PR #2987, opened 2018-05-12
Make the default key added to each wallet optional

spoonincode:
By default we add the "root" eosio key to each created wallet. This key is configurable, however it also is always required. Change to code to accept an empty key which is an indicator that no key shall be added to created wallets.

Orginally did this because it makes an incoming unit test a little easier; but I can also see why some users may not want a key placed in every single wallet they create too.

### DIFF
--- a/plugins/wallet_plugin/wallet_plugin.cpp
+++ b/plugins/wallet_plugin/wallet_plugin.cpp
@@ -54,7 +54,6 @@ void wallet_plugin::plugin_initialize(const variables_map& options) {
    }
    if (options.count("eosio-key")) {
       std::string eosio_wif_key = options.at("eosio-key").as<std::string>();
-      eosio_wif_key = fc::json::from_string(eosio_wif_key).as<std::string>();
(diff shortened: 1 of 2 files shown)


plugins/wallet_plugin/wallet_plugin.cpp
```diff
@@ -54,7 +54,6 @@ void wallet_plugin::plugin_initialize(const variables_map& options) {
    }
    if (options.count("eosio-key")) {
       std::string eosio_wif_key = options.at("eosio-key").as<std::string>();
-      eosio_wif_key = fc::json::from_string(eosio_wif_key).as<std::string>();
```
spoonincode: I could never figure out why this line was required. I had no problems setting a key once it was removed. It was a little annoying when present though because the empty key would always have to be specified with quotes. For example `eosio-key = ""` or `--eosio-key '""'`. Once removing this line I could simply leave `eosio-key = ` or `--eosio-key ""`. Setting the key still seemed to work fine.

PR #3037, opened 2018-05-14
nodeos SIGABRT "Pure virtual function called!" fix

moskvanaft:


moskvanaft:
@heifner,   I am not sure that it's correct to move the mutex to impl-class. Because each logger has it's own instance of impl. In my understanding the purpose of this mutex is to synchronize independent loggers using the same output (stdout).

heifner:
@moskvanaft Ah I see. Looking...

Issue #3067, opened 2018-05-15
Mismatch of cleos error advice


andriantolie:
eosio exceptions get refactored and some of the advice intended for the old eosio exception code now mismatchs

PR #3069, opened 2018-05-15
Fix mismatch error code with cleos error advice on DAWN 4.0

andriantolie:
Dawn 4.0 changes the code of the existing exceptions and this causes a mismatch with the error advice given by cleos. This causes confusion when understanding the given error advice. This PR rematch the error code with the error advice. 

At the same time, this PR also changes the reference from account_history_api_plugin in cleos to history_api_plugin since the former is obsolete.

#3067


bytemaster:
I'm going to merge this fix, but there should be an issue created to refactor error codes to enums.

andriantolie:
New issue is added to handle the refactor #3096

PR #3077, opened 2018-05-15
delete unneccesary assert

littlekfc:
The case `pivot != std::string::npos` will never happen in `else` block.


libraries/fc/src/crypto/private_key.cpp
```diff
@@ -98,7 +98,6 @@ namespace fc { namespace crypto {
          constexpr auto prefix = config::private_key_base_prefix;
 
          const auto pivot = base58str.find('_');
-         FC_ASSERT(pivot != std::string::npos, "No delimiter in string, cannot determine type: ${str}", ("str", base58str));
```
heifner: Can you also refactor so find is only called once.

Issue #2933, opened 2018-05-10
can't list more than 10 producers


ansigroup:
slim branch. It show only 10 producers, sorted by name in alphabetical order.  
commands used:
cleos system listproducers 
or (the same output)
cleos get table eosio eosio producers

So if you have 10 producer registered (all names starting for example with "k") and will add one more which start with "m" - it will not listed, but if you will add producer with  first letter in name "a" - it will add on beginning and remove last one. (removed producer keep producing, so probably its only affect output)

ansigroup:
upd. cleos system listproducers - sorts only by name and show first 10.

ansigroup:
I found the way to list more than 10: 
сleos get table eosio eosio producers -l 100

but there is no such key to extend limit for cleos system listproducers 
also as I see producers turns are in alphabetical order not by amount of staked tokens..

PR #3087, opened 2018-05-15
pagenate producer list in cleos (#2933)

tbfleming:


moskvanaft:
Is it working only for sorting by names ?
(which is not default, default is by votes)

tbfleming:
It doesn't depend on name sorting

tbfleming:
In-person discussion: since nodeos's get_table_rows always sorts by primary key then paginates the result, cloes's sort by vote is misleading. We're going to have to add a new API.

PR #3097, opened 2018-05-16
Do not bill for time needed to build WASM cache (and cleanup of CPU billing configuration parameters)

arhag:
Changes in this PR:
* Separated CPU time tracking into two timers: a regular timer tracking real time and a billing timer which can be paused/resumed.
* `deadline_exception` gets thrown on timeout of regular timer; other CPU exhaustion exceptions (including `leeway_deadline_exception`) get thrown on timeout of the billing timer.
* The billing timer is paused while creating a cache of `wasm_instantiated_module_interface`.
* `config::default_max_transaction_cpu_usage` parameter was changed to be half of `config::default_max_block_cpu_usage`.
* `min_transaction_cpu_usage` is now a consensus configuration parameter.
* Unneeded configuration parameters leftover from objective CPU billing were removed.


bytemaster:
This looks good, though we should in theory bill "extra time" on set code even though we don't hit the cache delay on set code.   This extra time should be equal to the actual "cache time" or "max allowed transaction time" and the user must have enough CPU bandwidth to cover this billing.

PR #3130, opened 2018-05-16
Fix for an infinite loop 

pmesnier:
triggered by the loss of all connections during resync for a node with multiple configured peers.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1281,35 +1281,34 @@ namespace eosio {
...
-         auto cptr = my_impl->connections.find(source);
-         if (cptr == my_impl->connections.end()) {
-            elog ("unable to find previous source connection in connections list");
-            source.reset();
-            cptr = my_impl->connections.begin();
-         } else {
-            ++cptr;
-         }
-         while (my_impl->connections.size() > 1) {
+         auto cptr = my_impl->connections.begin();
```
wanderingbort: this seems inconsistent with the comment above describing the selection process:

in the case where we have > 1 connections we:
 - start with a "range" of begin..end
 - if source is set (meaning we have a valid provider?) try to find that in the list setting the range to result..result
 - if source wasnt found (result === end) set the start of our range to begin.  OR re-open the range to begin..end which seems right
 - otherwise (result === source_itr), set the start of our range to the connection _after_ source and if source was the last connection, reset our start to begin.  
  - this gives us either cptr === source_itr + 1 and cend === source_itr (an inverted range) or cptr === begin and cend === source_itr

then we go through a while loop which has a check for cycling back  until cptr != end.   In the cases where cptr == begin and cend == end this is a trivial loop over all connections.  However, in our one edge case it seems like it would attempt to use any connection _except_ source?

wanderingbort: I guess, after trying all other options, if they fail `source` will still be set to whatever it was originally and we will use that.  So, its still inconsistent with the comment which implies that this passed in source has the highest priority when in fact it seems to have the lowest priority?

bytemaster: @wanderingbort is there a change request here or is this safe to merge?

wanderingbort: I'm withdrawing my immediate rejection because the code does work (i think).  We should address the inconsistency and depending on which way we go we should fix the comment or the code; or my read if I'm out of line

wanderingbort: Ok so, `source`, when entering this method, is either null or set to whichever connection we last requested from and we do intend to round-robin chunks to all our `current()` connections.  I suppose I got lost in the preconditions of `source` and the definition of "provider" in the comment which I assumed "provider" === `source` not "provider" === `conn`

PR #3148, opened 2018-05-17
Issue3140 - log and replay unconfirmed blocks

bytemaster:


libraries/chain/controller.cpp
```diff
@@ -174,8 +191,21 @@ struct controller_impl {
...
+         unconf_blocknum = objitr->blocknum;
+      }
+
       FC_ASSERT( db.revision() == head->block_num, "fork database is inconsistent with shared memory",
-                 ("db",db.revision())("head",head->block_num) );
+                 ("db",db.revision())("head",head->block_num)("unconfimed",unconf_blocknum) );
+
+      edump((unconf_blocknum));
+      /*
+      FC_ASSERT( head->block_num == unconf_blocknum, "unconfirmed block database out of sync",
```
arhag: If this assertion was failing, that is a sign that something is wrong with the code. Because as far as I see, the `unconf_blocknum` and `head->block_num` should never be out of sync.

libraries/chain/controller.cpp
```diff
@@ -354,14 +397,25 @@ struct controller_impl {
          emit( self.accepted_block_header, pending->_pending_block_state );
          head = fork_db.head();
          FC_ASSERT( new_bsp == head, "committed block did not become the new head in fork database" );
+
+      }
+
+      if( !replaying ) {
+         auto itr = unconfirmed_blocks.find<unconfirmed_block_object,by_num>(head->block_num );
+         if( !itr ) {
```
arhag: Can't we assert `!itr`? It seems like it would be a sign of a bug if we commit a block that already exists in the `unconfirmed_blocks`.

libraries/chain/controller.cpp
```diff
@@ -85,6 +92,9 @@ struct controller_impl {
     db( cfg.shared_memory_dir,
         cfg.read_only ? database::read_only : database::read_write,
         cfg.shared_memory_size ),
+    unconfirmed_blocks( cfg.block_log_dir/"unconfirmed",
+        cfg.read_only ? database::read_only : database::read_write,
+        std::min<uint64_t>(cfg.shared_memory_size,128*1024*1024ll) ), /// 1GB should store 1000 1MB blocks + overhead
```
arhag: The expected number of unconfirmed blocks in a health network of 21 producers is around 360. With 1 MB blocks, that would mean we would need at least 360 MB for the unconfirmed database, and this line appears to put a hard ceiling of 128 MB.

In an unhealthy network, the number of blocks required could easily exceed 360 and if the unconfirmed database size is too small, it will compound the problem by taking down more nodes (including potentially block producing nodes).

I think this should be a separate configuration parameter like `cfg.shared_memory_size` and the default value should be at least 512 MB.

bytemaster:
While cleaning this up I discovered a new issue, #3172 that can probably be resolved at a future point in time.

Issue #2404, opened 2018-04-15
test chain generator


pmesnier:
For at least issue #2390, we need a way to create chains of varying initial lengths. In some cases the chains need to be of a common source but we will need at least 3 source chains, two with the same genesis block but started by different bios producer instances, and another with a different genesis block. I'm not certain of the 3rd case.

Anyway, this variable length chain initializer could be built into nodeos as an extension to the hain controller, but I think it would be better as a separate utility, or maybe as a plugin module even.

Bonus idea - implement this as plugin, and allow it to accept source identifiers as some sort of URI format that could be local filesystem or network address, along with offset and count values, and now we have alternate synchronizer Dan mentioned the other day.  We could even throw in zlib based streaming compression. However integration with the existing p2p and chain controller code is left as an exercise for the reader.

taokayan:
hey I've pushed a enhancement to txn_test_gen_plugin which allows you to create different initial lengths for different nodes, while maintaining the ability to sync with other nodes. Please see https://github.com/EOSIO/eos/pull/3151

config changes in config.ini:

# param startup-produce-blocks: number of extra blocks be built at initial stage, can be different on different nodes.
# param startup-sign-key: the key used to sign the extra initial blocks, or the bios private key, should be same on different nodes.
# txn_test_gen_plugin must be enabled.

startup-produce-blocks = 100
startup-sign-key = ["EOS8L8R3azSpkC2Go1Gj62qnq7EiWsGus8fMfrar7BqcxiFF8TpxR","5KFxAo6AHDzUPTtYjPet5FHQttC6VqmKpJ6cGbEj5sLsankoGuK"]
plugin = eosio::txn_test_gen_plugin

brianjohnson5972:
@wanderingbort @heifner I am thinking this is OBE.  If I don't hear from you I will presume that is the case.

brianjohnson5972:
No longer part of the plan.

PR #3151, opened 2018-05-17
2404 enhance txn_test_gen_plugin to produce variable inital length

taokayan:
this enhancement is for #2404


plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
```diff
@@ -306,29 +306,84 @@ struct txn_test_gen_plugin_impl {
...
 void txn_test_gen_plugin::plugin_startup() {
+   ilog("starting txn_test_gen plugin");
    app().get_plugin<http_plugin>().add_api({
       CALL(txn_test_gen, my, create_test_accounts, INVOKE_V_R_R(my, create_test_accounts, std::string, std::string), 200),
       CALL(txn_test_gen, my, stop_generation, INVOKE_V_V(my, stop_generation), 200),
       CALL(txn_test_gen, my, start_generation, INVOKE_V_R_R_R(my, start_generation, std::string, uint64_t, uint64_t), 200)
    });
+
+   if (my->startup_produce_blocks > 0) {
+      ilog("ready to generate ${i} initial block(s)", ("i", my->startup_produce_blocks));
```
wanderingbort: This method of initializing a chain will be pretty cumbersome.
- because of the new explicit nature of block start/commit/abort we've centralized all control of blocks in the `producer_plugin` which must track additional state to properly operate.  This direct access of the `controller` will desync that state
- this does not seem to detect whether the chain is new, so on replay/restart it will create an additional `startup_produce_blocks` worth of blocks, i suspect on many chains this will assert out and cause the `nodeos` process to stop
- This does not guarantee that between the timestamp in genesis.json (assuming a new chain) and `fc::time_point::now()` there is room for `startup_produce_blocks` worth of block time.  As a result, it is entirely possible that the block that is started at the end is in the "past" 

the second points are solvable.  The first point is a deal breaker of sorts.  We don't want debugging features to leak into the `producer_plugin` and unfortunately it has invariants to maintain.  We can pass signed blocks into it the same way the `net_plugin` does but you would need a way of generating these signed blocks without corrupting the state of chain controller.  Perhaps for empty blocks this may be doable.

taokayan: In my understanding these changes are just for testing purpose for ticket #2404 in which we need to test the sync-up between nodes of different chain lengths. And they're all in txn_test_gen_plugin. For point 3, I guess I can figure out how many blocks at maximum we can insert between the genesis timestamp and time_point::now(). But for point 1, would you suggest some better idea?

wanderingbort: unfortunately it is not possible to run a node without the producer_plugin at this time.  Even though the changes are isolated they are violating some of the (undocumented) ownership model that the plugins need to respect for node stability.

wanderingbort: If by this you mean, this feature should never be used with an actual producer (it will produce blocks in this plugin and no where else for the life of the process), it may work.  In that case, I would suggest isolating it in its own plugin.  The txn_test_gen_plugin _is_ used in testnets which have producers and leaving an option that will silently try to corrupt things in a plugin that gets use is probably a bad idea.

taokayan: Got it. I'll isolate it into a new testing plugin then. Thanks!

heifner:
This has been overcome by events. Closing.

PR #3188, opened 2018-05-18
Fix #3187 - effecient detection of names with '.'

### LINKED ISSUE #3187: Efficient detection of '.' in names prohibited from eosio.system
This should be performed with bitmath vs string operations.

bytemaster:
- also exempt the eosio account from this restriction

### DIFF
--- a/contracts/eosio.system/delegate_bandwidth.cpp
+++ b/contracts/eosio.system/delegate_bandwidth.cpp
@@ -97,10 +97,14 @@ namespace eosiosystem {
                             /*  no need to parse authorites
                             const authority& owner,
                             const authority& active*/ ) {
-      auto name_str = eosio::name{newact}.to_string();
 
-      eosio_assert( name_str.size() == 12 || creator == N(eosio), "account names must be 12 chars long" );
-      eosio_assert( name_str.find_first_of('.') == std::string::npos  || creator == N(eosio), "account names cannot contain '.' character");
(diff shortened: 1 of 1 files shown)


contracts/eosio.system/delegate_bandwidth.cpp
```diff
@@ -97,10 +97,14 @@ namespace eosiosystem {
...
                             const authority& owner,
                             const authority& active*/ ) {
-      auto name_str = eosio::name{newact}.to_string();
 
-      eosio_assert( name_str.size() == 12 || creator == N(eosio), "account names must be 12 chars long" );
-      eosio_assert( name_str.find_first_of('.') == std::string::npos  || creator == N(eosio), "account names cannot contain '.' character");
+      if( creator != _self ) {
+         auto tmp = newact;
+         for( uint32_t i = 0; i < 12; ++i ) {
+           if( tmp & 0x1f ) eosio_assert( false, "name may not contain '.'");
```
brianjohnson5972: This will detect name's with less than 12 characters, but it will not give the users any idea what to fix since they will have no idea why their name was rejected for containing a '.' when they didn't put one in there.

bytemaster: it is actually wrong everything is shifted 4 bits in the wrong direction.

bytemaster:
This issue is resolved when pull #3206 is merged.

PR #3190, opened 2018-05-18
Add eosio_assert_code intrinsic

### LINKED ISSUE #2561: Error code assert
Assert should take an error code, this can be defined as a table in the abi for printing human readable messages.  
Two versions of assert will exist:
* ```assert(uint64_t)```
* ```asserts(const char*, size_t)```

The ```asserts``` intrinsic is the same as the old ```assert```, but will now also take a length.

arhag:
Resolves #2561 (with the exception of the logic to actually print the user-friendly error message on `eosio_assert_code` assertion failures).

Two new intrinsics have been added: `eosio_assert_message` (which is very similar to `eosio_assert` except it also takes the size of the message string) and `eosio_assert_code` (which takes a `uint64_t` error code rather than a message string).

A field `error_messages` has been added to the ABI which maps an `uint64_t` error code to a message string. 

The idea is that contract authors can include the error message strings in their ABI (which could theoretically be hosted off the chain) rather than bloating the WASM contract code with string literals. Then the code passes the appropriate error code to the `eosio_assert_code` function call which (assuming the assertion fails) supporting clients can use along with the ABI to map the error code to a user-friendly error message which is displayed. (Note: although the functions in `abi_serializer` to support the mapping are included in this PR, the logic to actually display the user-friendly error message has not yet been implemented.)

This PR also includes a couple of unrelated changes. 

First, the `eosio::print( const std::string& s)` function implementation in eosiolib (i.e. WASM side) has been changed to use the more efficient `prints_l` intrinsic rather than `prints`. 

Second, a bug in CPU and network usage limit tracking in `transaction_context` was fixed and the `wasm_tests/weighted_cpu_limit_tests` test was modified to pass given the relatively recent changes to the bandwidth resource limiting algorithm (it still has not been enabled in ctest since the test isn't super robust, i.e. depends on wall clock timing, although it is more robust than it was before).


libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -167,6 +167,10 @@ namespace eosio { namespace chain {
                                     3050001, "account name already exists" )
       FC_DECLARE_DERIVED_EXCEPTION( invalid_action_args_exception, action_validate_exception,
                                     3050002, "Invalid Action Arguments" )
+      FC_DECLARE_DERIVED_EXCEPTION( eosio_assert_message_exception, action_validate_exception,
+                                    3050003, "Invalid Action Arguments" )
+      FC_DECLARE_DERIVED_EXCEPTION( eosio_assert_code_exception, action_validate_exception,
+                                    3050004, "Invalid Action Arguments" )
```
heifner: Add diff string literals for these.

libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -84,19 +84,36 @@ struct clause_pair {
...
+
+   uint64_t error_code;
+   string   error_msg;
+};
+
 struct abi_def {
    abi_def() = default;
-   abi_def(const vector<type_def>& types, const vector<struct_def>& structs, const vector<action_def>& actions, const vector<table_def>& tables, const vector<clause_pair>& clauses)
-   :version("eosio::abi/1.0"), types(types), structs(structs), actions(actions), tables(tables), ricardian_clauses(clauses)
+   abi_def(const vector<type_def>& types, const vector<struct_def>& structs, const vector<action_def>& actions, const vector<table_def>& tables, const vector<clause_pair>& clauses, vector<error_message>& error_msgs)
```
heifner: `const vector<error_message>& error_msgs`

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -54,6 +55,9 @@ struct abi_serializer {
    type_name get_action_type(name action)const;
    type_name get_table_type(name action)const;
 
+   optional<map<uint64_t, string>::const_iterator>  find_error_message( uint64_t error_code )const;
```
heifner: Why an optional iterator? Doesn't an iterator act as an optional anyway?

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -54,6 +55,9 @@ struct abi_serializer {
...
+   const string&                                    get_error_message( uint64_t error_code )const;
```
heifner: I think returning a string would be safer. The abi_serializer objects can be short-lived.

libraries/chain/include/eosio/chain/abi_serializer.hpp
arhag: You would have to compare to the internal `error_messages` field, which I guess the caller can do since it is public, but I didn't want them to be concerned about the implementation details of the `abi_serializer` struct.

I didn't want to just return an `optional<string>` since that would create unnecessary copies, but maybe I should just replace `find_error_message` and `get_error_message` with one function that returns an `optional<string>` to address the safety issues at the cost of unnecessary copies.

heifner: Your right that is ugly. I do like the idea of the `optional<string>` return.

PR #3198, opened 2018-05-18
Fix ubuntu version check ( OS_MAJ -lt 16 )

rlaace423:
There's an if statement which checks ubuntu version is eosio_build_ubuntu.sh in ./scripts directory.

The minimum version which supports at EOS is ubuntu 16.04 or higher, so It should be changed like below.

BEFORE
if [ "${OS_MIN}" -lt 4 ]; then

AFTER 
if [ "${"OS_MAJ}" -lt 16 ]; then

### DIFF
--- a/scripts/eosio_build_ubuntu.sh
+++ b/scripts/eosio_build_ubuntu.sh
@@ -39,7 +39,7 @@
 		   fi
 		;;
 		"Ubuntu")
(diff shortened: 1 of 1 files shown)


rlaace423:
And I am curious that checking minor version number is intended (OS_MIN < 4). But in my thought, that if statement is not worthy because ubuntu doesn't distribute their os with x.04 or x.10 only.

PR #3207, opened 2018-05-18
Cleos system list bw

moskvanaft:
````
$ cleos system listbw anton
Receiver     Net bandwidth        CPU bandwidth        
anton        100000000.0000 SYS   100000000.0000 SYS   
user         1000.0000 SYS        2000.0000 SYS
````


pipilink:
Hello! 
It would not be bad to have a reverse command of this listbw, for example: relistbw
$ cleos system relistbw anton
Sender     Net bandwidth        CPU bandwidth        
user         1000.0000 SYS        2000.0000 SYS

How to do it?
Thanks!

PR #3210, opened 2018-05-18
get block display of old structs

### LINKED ISSUE #2099: Unable to retrieve block containing setprods data
Node containing setprods cannot be retrieved. If node 81 contains setprods cleos get block 81 command gets a serialization failure:

programs/cleos/cleos get block 81
840361ms thread-0   main.cpp:1186                 main                 ] Failed with error: Assert Exception (10)
itr != structs.end(): Unknown struct 
Failed to serialize type

This occurs specifically at bootstrap as setprods is in contract eosio.bios. The setprods action when invoked on the bios node is successful (new producers start producing blocks and call returns a success including a transaction id). The specific block containing the setprods is initially retrievable (within 1-5 seconds of block production), however after about 5 seconds becomes un-retrievable. That maybe when block becomes irreversible.

heifner:
Resolves #2099.

For get block to pretty print the action data struct in transactions it must have the abi for the struct. If a contract is updated with new abi, it must therefore keep the old abi around so that get block can display it. Otherwise get block will just print the hex version of the action data struct.

This PR adds a check for unknown type and reverts back to hexdata of the action data.
Also added the eosio.bios abi to eosio.system abi so that when eosio.system replaces eosio.bios users can still see the structs of blocks that contain eosio.bios abi.


abourget:
So in the end, are we keeping the old ABIs around, or falling back to simply sending the `hex_data` if off-sync ?

PR #3224, opened 2018-05-19
Allow inject CORE_SYMBOL_NAME from Env

kesar:
Fixes https://github.com/EOSIO/eos/issues/3225


doublesharp:
I see that you have chosen to use SYS as a symbol which might be confusing for your users as SYS has been the symbol for Syscoin for several years now - before EOS was founded actually.

sidhujag:
I can also confirm that @bytemaster knows of Syscoin because I had reached out to him on working on an interesting next-gen project that we are researching when he was still designing EOS. Not sure how he would forget it was only a few months back. Maybe he simply forgot?

bstr156:
SYS is the well-known trading symbol for Syscoin.  A bit concerning that EOS would pick that, especially considering some of the similar utility of the platforms.  Some clarification or response seems warranted.

dwasyluk:
SYS is already used by Syscoin which has been around since 2014, is in CMC top 100 and trades on several major exchanges (Polo, Binance, Bittex).

bytemaster:
The fact that sys conflicts with existing tokens is good. No one should use the default in production.

kesar:
@doublesharp @bstr156 @sidhujag great concerns. 👍 

SYS is just for internal development. The change is to set that variable using the environment to any other symbol (EOS, for example ;) ).

cydg:
Why not name it SOE or EOSH? *joking*
Using taken symbols is like begging for conflicts at some point down the road. 
Something like NULL or EMPTY should make it clearer than any already taken symbol.

doublesharp:
@bytemaster then why not use EOS? That conflicts with something no?

matthewdarwin:
./eosio_build.sh -DCORE_SYMBOL_NAME=EOS should work

matthewdarwin:
./eosio_build.sh -DCORE_SYMBOL_NAME=EOS should work

dwasyluk:
If the goal is to use a conflicting symbol why not use something painfully obvious as a conflicting symbol ...like BTC. The rationale for using SYS here doesn't completely add up; EOS and BTC would accomplish the same thing, no change needed.

PR #3234, opened 2018-05-21
Fix #3229: keosd does not respect the --wallet-dir option

### LINKED ISSUE #3229: keosd does not respect the --wallet-dir option
keosd does not respect the --wallet-dir option. Wallet files are always created in ~/eosio-wallet

test case:

```
docker run --rm --name keosd -v keosd-data-volume:/opt/eosio/bin/data-dir eosio/eos:dawn-v4.1.0 /opt/eosio/bin/keosd --wallet-dir=/opt/eosio/bin/data-dir

docker exec keosd cleos wallet create
"/opt/eosio/bin/keosd" launched
Creating wallet: default
Save password to use in the future to unlock this wallet.
Without password imported keys will not be retrievable.
"Password'
```

```
docker exec keosd ls -l /opt/eosio/bin/data-dir
[empty]
```

```
docker exec keosd ls -l /root/eosio-wallet
total 8
-rw-r--r-- 1 root root 1533 May 20 18:08 config.ini
-rw------- 1 root root  311 May 20 18:08 default.wallet
```

noprom:
Fix a docker-compose file issue: [#3229](https://github.com/EOSIO/eos/issues/3229)

### DIFF
--- a/Docker/docker-compose-dawn4.0.yaml
+++ b/Docker/docker-compose-dawn4.0.yaml
@@ -25,7 +25,7 @@ services:
 
   keosd:
     image: eosio/eos:latest
-    command: /opt/eosio/bin/keosd --wallet-dir /opt/eosio/bin/data-dir
+    command: /opt/eosio/bin/keosd --wallet-dir /opt/eosio/bin/data-dir --http-server-addres=127.0.0.1:8900
     hostname: keosd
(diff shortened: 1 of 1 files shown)


noprom:
@matthewdarwin  Thanks!

PR #3261, opened 2018-05-22
Beast P2P Network

bytemaster:
This pull request introduces an alternative P2P network protocol based upon boost::beast websockets. 

Major Features:
1.  multi-threaded - the networking code can use as many threads as necessary, number configured via command line
2. less than 1000 lines of code 
3. uses a simple yet reliable syncing algorithm 
4. auto reconnect 
5. graceful connection shutdown w/out delay 
6. prioritizes blocks over transactions without creating large queues

Tested at 1000 TPS with 3 nodes on localhost
Tested multiple syncing peers at same time as 1000 TPS 

Missing Features:
1. No RPC interface for dynamically adding/removing peers 
2. no attempts made to minimize bandwidth consumed via pre-acks

This net plugin could "in theory" work side-by-side with the existing net plugin and is entirely independent / optional.


abourget:
Will it negotiate a different network version than the current p2p protocol?

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -0,0 +1,1079 @@
...
+           if( ec ) {
+              return on_fail( ec, "accept" );
+           }
+
+           do_hello();
+           do_read();
+        }
+
+        void do_read() {
+           _ws->async_read( _in_buffer,
```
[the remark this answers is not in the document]
wanderingbort: for future reference, it looks like the default max is 16mb and is configurable via a call to `::read_message_max(size_t)`

https://www.boost.org/doc/libs/develop/libs/beast/doc/html/beast/ref/boost__beast__websocket__stream/read_message_max/overload1.html

spoonincode:
Boost 1.67 release notes mention
>This version fixes significant defects in websocket::stream which can lead to asserts or undefined behavior. Users are encouraged to update to the latest Boost release.

IMO we should move the build scripts to 1.67 before taking this

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -0,0 +1,1096 @@
...
+   }
+
+   bnet_plugin::~bnet_plugin() {
+   }
+
+   void bnet_plugin::set_program_options(options_description& cli, options_description& cfg) {
+      cfg.add_options()
+         ("bnet_endpoint", bpo::value<string>()->default_value("0.0.0.0:4321"), "the endpoint upon which to listen for incoming connections" )
+         ("bnet_threads", bpo::value<uint32_t>(), "the number of threads to use to process network messages" )
+         ("bnet_connect", bpo::value<vector<string>>()->composing(), "the endpoint upon which to listen for incoming connections" )
```
spoonincode: wrong help text here

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -0,0 +1,1096 @@
...
+   bnet_plugin::bnet_plugin()
+   :my(std::make_shared<bnet_plugin_impl>()) {
+   }
+
+   bnet_plugin::~bnet_plugin() {
+   }
+
+   void bnet_plugin::set_program_options(options_description& cli, options_description& cfg) {
+      cfg.add_options()
+         ("bnet_endpoint", bpo::value<string>()->default_value("0.0.0.0:4321"), "the endpoint upon which to listen for incoming connections" )
```
spoonincode: Other options use hyphen, not underscore

Issue #3017, opened 2018-05-14
Node doesn't try to reconnect with the valid peer after it receives connection from other peer that has unlinkable "second" block (i.e. it doesn't link with the node's genesis block)


andriantolie:
In this scenario, we have 3 nodes which are A1, A2, B. And they abide by the following rules:
- Node A1 and Node A2 share the same genesis
- Node B uses a different genesis (we can just modify the initial timestamp for this). So the genesis block of Node B has different id from A1 and A2
- Node A1 knows Node A2
- Node A2 knows Node B
- Node B knows Node A1
- Node A2 and Node B has enable-stale-production enabled, producer_plugin enabled, and the producer is eosio
- Node A1 has enable-stale-production disabled and producer_plugin disabled

And then we start the nodes in the following order: A1, B, A2. The reaction should be the following:
- Node A1 finds out that Node A2 is inactive and failed to connect with Node A2
- When Node B starts it tries to connect with Node A1. Node A1 will throw an exception `prior != by_id_idx.end(): unlinkable block`
- When Node A2 starts, it will try to connect to Node B, however, it will find out that B is forked so it rejects the connection with Node B
- **Normally**, after some time Node A1 will try to reconnect to A2 and be able to sync with A2. **However, this doesn't happen**

The above problem **only happens when Node A1 only have the genesis block** inside its chain. If Node A1 has another block on top of the genesis block, it will reject the connection with Node B and regard Node B as forked instead of throwing an exception. Then it will try to reconnect with Node A2 after some time

For convenience, I'm attaching the config and genesis for Node A1, A2, and B here
[3017-config-and-genesis.zip](https://github.com/EOSIO/eos/files/2032817/3017-config-and-genesis.zip)

Then they can be started by the following:
```
programs/nodeos/nodeos --data-dir data-dir-a1 --config-dir data-dir-a1 --resync
programs/nodeos/nodeos --data-dir data-dir-b --config-dir data-dir-b --resync
programs/nodeos/nodeos --data-dir data-dir-a2 --config-dir data-dir-a2 --resync
```

pmesnier:
Thanks for that detailed write up. I assume that is against the master branch?

andriantolie:
Yep

andriantolie:
Awesome. Tested and verified!

PR #3269, opened 2018-05-22
#3017 fix syncing with a forked peer

pmesnier:
addresses the reported failure by changing how the synchronizer handles block errors, getting rid of the "second try" logic in favor of a "one and done" approach instead. If one rejected block is received from a peer, that connection is closed and an updated curren state message is sent to all peers again.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -589,7 +589,6 @@ namespace eosio {
       uint32_t       sync_last_requested_num;
       uint32_t       sync_next_expected_num;
       uint32_t       sync_req_span;
-      uint32_t       last_repeated;
       connection_ptr source;
       stages         state;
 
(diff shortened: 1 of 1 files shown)


pmesnier:
Actually I think a better solution for finding the next chunk provider is to send pu another round of handshake messages and pick the first responder, since that will be the once with the least latency.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1256,58 +1254,66 @@ namespace eosio {
...
+            while (cptr != cend) {
+               //select the first one which is current and break out.
+               if ((*cptr)->current()) {
+                  source = *cptr;
+                  break;
+               }
+               else {
+                  // advance the iterator in a round robin fashion.
+                  if (++cptr == my_impl->connections.end()) {
+                     cptr = my_impl->connections.begin();
```
PaulCalabrese: What guarantees do we have that this loop will terminate? It seems if the source is not found (and cend is equal to the collection end()) and no connections are "current", then the loop will never end.

PR #3270, opened 2018-05-22
Added setprods to `eosio.system`.

abourget:
Solves the bootstrapping problem of a new network. Otherwise, there is
no way to delegate to the ABPs, and it makes the boot more brittle.

### DIFF
--- a/contracts/eosio.system/eosio.system.cpp
+++ b/contracts/eosio.system/eosio.system.cpp
@@ -77,8 +77,16 @@ namespace eosiosystem {
       set_privileged( account, ispriv );
    }
 
+   void system_contract::setprods( std::vector<eosio::producer_key> schedule ) {
+      (void)schedule; // schedule argument just forces the deserialization of the action data into vector<producer_key> (necessary check)
+      require_auth( _self );
+      char buffer[action_data_size()];
+      read_action_data( buffer, sizeof(buffer) ); // should be the same data as eosio::pack(schedule)
+      set_active_producers(buffer, sizeof(buffer));
+   }
+
 } /// eosio.system
- 
+
 
 EOSIO_ABI( eosiosystem::system_contract,
      (setram)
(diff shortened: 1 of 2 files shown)


wanderingbort:
I have concerns with this method.  While you can make the argument that 2/3+1 producers has the ability to affect election of new producers, this explicit endpoint suggests that the have the "charter" to unilaterally dictate a change in the producer set.   I think I would prefer a "mid-phase" contract that allows the ABP delegation which can be replaced with the "runtime" system contract which includes no such endpoint.

At the very least, I would expect we need some sort of coordination between this method and the election system so that they are mutually exclusive.  Otherwise, the expected state of the chain from the PoV of the election system may not reflect the actual state due to a call to this action.

abourget:
Right now, that endpoint lives in `eosio.bios`. Is that the mid-phase contract you mean ?

The problem I'm facing is that I need the system contract to buy ram, delegatebw, when injecting the snapshot. I could re-inject `eosio.bios` to call that one, and inject `eosio.system` yet another time.. because the `setprods` operation needs to be one of the last ones (before destroying the `eosio` permissions).

Could we have a quick chat abou that ? It's been moving quite a bit and most things in `eosio.bios` moved to `eosio.system` (like `setpriv`).

The tx needs to be signed by `eosio` (or the `msig`) dance.. I don't expect that would pass as a proposition, to play with that function.  But if we really wanted to call it, we could `setcode eosio.bios`, then `setprods`, then `setcode eosio.system` in the same Tx .. so I'm not sure not having it around is any real protection.

bytemaster:
The boot sequence I imagine is as follows:

1. deploy bios contract 
2. set prods to ABPs 
3. have initial owner of eosio (not ABPs) create eosio.token,msig, and create tokens, and issue them, etc...
4. deploy eosio.system contract
5. assign eosio owner/active to eosio.prods 
6. done

bytemaster:
If we did expose this the action would be undone at start of next round (in minutes) and returned back to the elected states. Therefore, the action makes no sense except before the 150M token activation threshold. 

That said, the community is free to use what ever system contract they want ;)

abourget:
The action *is* for before 15% unlock..

The thing is it breaks our launch design, where the BIOS Boot node fills in the chain with all of its transactions before even connecting to the world.. To fill in the chain, `eosio` needs to build all the blocks.. and then, he connects to the swarm, flooding the network with its block'd transactions.. and then, at the end, the ABPs discover they're on the hook to produce.

If we `setprods` early in the boot sequence, it means the swarm needs to be connected already, otherwise, the next transactions won't be processed.

PR #3299, opened 2018-05-22
producer plugin max-irreversible-block-age

### LINKED ISSUE #3279: Auto Shutdown when LIB stops advancing
Add a command line argument that will cause a node to automatically stop-production when the LIB block time is older than N seconds.

This will identify the following:
1. your local node is disconnected from all others
2. your local node is on a minority fork 
3. the network has caught fire and producers should probably manually intervene to resolve issue.

This solves the following:
1. stop digging a deeper hole and committing yourself to block numbers that you may want to sign on a real chain
2. minimizes difficulty in detecting and resuming real chain by minimizing the amount of competing blocks being considered / distributed.

wanderingbort:
this resolves #3279 

There is a new parameter for the producer plugin `max-irreversible-block-age` that creates another condition for speculating on a block rather than producing. 

If the irreversible-block is too old WRT to this parameter the node will stop producing so that it 
 * does not produce extra load on a network which seems unlikely to converge without operator intervention
 * does not put out block confirmations for more block numbers which may "lock out" a producer once the fork is resolved.

the default is set to 30 minutes.


bytemaster:
I think the default should be closer to 6 minutes (2 rounds), also it might be a good idea for the RPC command to "clear" this flag to enable production without a restart. 

We expect producers to be responsive within 15 minutes of an "issue" for manual intervention, the automated tools should be "faster" than the manual intervention or the point almost becomes moot.

PR #3323, opened 2018-05-23
Support two new options in 'cleos'

InfoHunter:
In some scenarios, 'nodeos' runs remotely from where 'cleos' locates, that implies the communication between the two parties should go across the network. So a new option '-r,--header' is proposed in this PR to allow user to customized the HTTP request headers sent by 'cleos' in case that some 'middle' devices may require for specific headers to contain some 'token'.

And also, if 'nodeos' is set up to provide HTTPS APIs, say by placing a reverse proxy in front, in many cases, especially for some testing environments, there might not be domains assigned with the 'nodeos' - so 'cleos' needs a way to 'tolerate' with the invalid certificates. To address this, a new option '-n,--no-verify' is proposed in this PR too, to disable the peer verification for current secure connection.

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -114,7 +114,13 @@ namespace eosio { namespace client { namespace http {
    request_stream << "Host: " << server << "\r\n";
    request_stream << "content-length: " << postjson.size() << "\r\n";
    request_stream << "Accept: */*\r\n";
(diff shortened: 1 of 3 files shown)


InfoHunter:
Seems I need to resolve some conflicts...

InfoHunter:
Conflicts are resolved...

PR #3334, opened 2018-05-23
producer api plugin

### LINKED ISSUE #3297: Producer API Plugin
MVP:
 - pause production
 - resume production

The goal is to allow producers to coordinate during bad situations to resume/pause the chain without having the variability of replay/resync time etc.

wanderingbort:
resolves #3297 

 - Add producer_api_plugin with endpoints to pause/resume and query the state of the producer plugin
 - Add the ability to pause production to the producer_plugin
 - Add a CLI/config option that controls whether the node starts in a paused state or not
 - cleanup to consolidate checks for production being disabled by policy rather than other things
 - fixed off-by-one error in calculation of a producers _next_ chance to produce which was not taking the pending block into account


wanderingbort:
for reference:
```
POST /v1/producer/pause - pause production
POST /v1/producer/resume - resume from paused production
POST /v1/producer/paused  - return true/false depending on whether the node is currently paused
```

we really need a better API framework as `/v1/producer/paused` should be a `GET`

DenisCarriere:
Great stuff! This will be excellent for quickly rolling over an active producer node with a new producing node. Will be trying out this feature once it's becomes available in the next release 👍

PR #3340, opened 2018-05-23
chain_plugin parameter --contracts-console added #3339

moskvanaft:


moskvanaft:
@sergmetelin , added, see #3385

PR #3342, opened 2018-05-23
Set Chain ID from CMake

larryk85:
Added support for setting the chain id with cmake.  Added the flag CHAIN_ID, this can be set to some hashed hex string.  If no flag is given then the default hash is from "eosio::chain version 1.0".


libraries/chain/include/eosio/chain/chain_id.hpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
```
bytemaster: This is a generated file, it shouldn't be checked in.

bytemaster:
This commit changes to much with the following problems

- [ ] config.hpp is included by everything, every time cmake runs we have to rebuild the world
- [ ] we should only change the implementation of the default constructor for chain_id_type and only in the chain_id.cpp file.  This will prevent us from rebuilding the world.

kesar:
Could we expose it also in get_info ? then clients / portals can identify a chain by that 🙏

PR #3379, opened 2018-05-24
Ubuntu: changed boost version from 1.66 to 1.67

pacificcode:
removed boost 1.66
added boost 1.67


scripts/eosio_build_ubuntu.sh
```diff
@@ -109,9 +109,9 @@
 	printf "\\n\\tChecking boost library installation.\\n"
```
heifner: eosio_build.sh has `export BOOST_ROOT="${HOME}/opt/boost_1_66_0"`

PR #3427, opened 2018-05-25
Add bnet support to launcher

PaulCalabrese:
Add an option --p2p-plugin that accepts values of "net" and "bnet" and defines the p2p plugin to use. At this point it is an either-or option and does not support mixed networks.

### DIFF
--- a/programs/eosio-launcher/main.cpp
+++ b/programs/eosio-launcher/main.cpp
@@ -322,6 +322,11 @@ struct last_run_def {
 };
 
 
+enum class p2p_plugin {
+   NET,
+   BNET
+};
+
 enum launch_modes {
   LM_NONE,
   LM_LOCAL,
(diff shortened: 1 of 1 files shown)


programs/eosio-launcher/main.cpp
```diff
@@ -989,10 +1016,18 @@ launcher_def::write_config_file (tn_node_def &node) {
 
   if(!is_bios) {
      auto &bios_node = network.nodes["bios"];
-     cfg << "p2p-peer-address = " << bios_node.instance->p2p_endpoint<< "\n";
+     if (p2p == p2p_plugin::NET) {
```
jgiszczak: In keeping with the net_plugin default, shouldn't this be reversed?  Specific to p2p_plugin::BNET, otherwise anything else gets the other setting.

gleehokie: IRL discussion occurred and this is okay.

PR #3467, opened 2018-05-28
Fix openssl elliptic compile issue

harrywong:


harrywong:
May be we also need to provide a test for openssl-elliptic and mixed-elliptic build.

spoonincode:
Can you provide a little more description on what this fixes? The expectation is that this file will not be used in our compilation due to secp256k1 library usage.

harrywong:
Yes, but the fc library also provide the option(aka. `ECC_IMPL` option) to be built with OpenSSL.
This fixes the compile issue when building fc library with OpenSSL-ECC implementation.

bytemaster:
It only fixes the compile issue, it doesn't actually make it canonical.  We should make sure the implementations are the same.

harrywong:
Hi @bytemaster, I've implemented canonical semantic, thx.

PR #3485, opened 2018-05-28
cleos skip-sign should not require wallet

### LINKED ISSUE #3449: We should not have to have a wallet open when calling with
Currently, in order to print an unsigned transaction, you have to have a wallet open.

It seems this should not be necessary. I re-wrote the client code locally so I wouldn't need a wallet open, and it seems to work.

I want to call this command:
./cleos --url=http://dev.cryptolions.io:38888 push action --skip-sign --json --dont-broadcast eosio delegatebw '{"from":"greginjungle", "receiver":"greginjungle", "stake_net_quantity": "100.0000 EOS", "stake_cpu_quantity": "100.0000 EOS", "transfer": true}'

At HEAD, I get this error:
Error 3120006: No available wallet
Ensure that you have created a wallet and have it open
Error Details:
You don't have any wallet!

I looked at the code and saw you are calling "determine_required_keys" even if you aren't going to need them:
https://github.com/EOSIO/eos/blob/master/programs/cleos/main.cpp#L284

I replaced that code with something like the following:
```
   if (!tx_skip_sign) {
      auto required_keys = determine_required_keys(trx);
      sign_transaction(trx, required_keys, info.chain_id);
   }

```
Now it goes through with output:
```
{
  "expiration": "2018-05-26T18:04:35",
  "ref_block_num": 52458,
  "ref_block_prefix": 261619818,
  "max_net_usage_words": 0,
  "max_cpu_usage_ms": 0,
  "delay_sec": 0,
  "context_free_actions": [],
  "actions": [{
      "account": "eosio",
      "name": "delegatebw",
      "authorization": [],
      "data": "a0229bfa4dc7d465a0229bfa4dc7d46540420f000000000004454f530000000040420f000000000004454f530000000001"
    }
  ],
  "transaction_extensions": [],
  "signatures": [],
  "context_free_data": []
}
```

heifner:
We should not have to have a wallet open when calling "push action" with --skip-sign

-Don't launch keosd if --skip-sign option passed.
-Fixed issue where wallet required with --skip-sign option (thanks @grock999)
-Fixed check for keosd specified as local before attempting to launch.

Resolves #3449

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -281,13 +281,11 @@ fc::variant push_transaction( signed_transaction& trx, int32_t extra_kcpu = 1000
       trx.context_free_actions.emplace_back( generate_nonce_action() );
    }
 
-   auto required_keys = determine_required_keys(trx);
-   size_t num_keys = required_keys.is_array() ? required_keys.get_array().size() : 1;
-
    trx.max_cpu_usage_ms = tx_max_net_usage;
    trx.max_net_usage_words = (tx_max_net_usage + 7)/8;
 
    if (!tx_skip_sign) {
+      auto required_keys = determine_required_keys(trx);
       sign_transaction(trx, required_keys, info.chain_id);
    }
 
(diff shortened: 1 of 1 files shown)


spoonincode:
>Don't launch keosd if --skip-sign option passed

This seems like a very targeted fix for a more broad problem with the auto-launch code. For example `cleos get info` also tries to launch keosd. Do you think there is a different/easy way to fix this to make the auto launch only engage when absolutely needed?

heifner:
I'll see about restricting auto launch in more places. Personally I dislike the auto launch and would prefer an opt-in instead of opt-out.

Issue #3314, opened 2018-05-23
More than two keosd processes can exist.


EugeneChung:
You can easily reproduce the problem.

1. Run keosd program twice.

$ eos/build/programs/keosd/keosd --http-server-address=127.0.0.1:8900 &
$ eos/build/programs/keosd/keosd --http-server-address=127.0.0.1:8900 &

2. You can see the logs below,

wallet_plugin.cpp:41          plugin_initialize    ] initializing wallet plugin
252121ms thread-0   http_plugin.cpp:247           plugin_initialize    ] configured http to listen on 127.0.0.1:8900
252121ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/keosd/stop
252121ms thread-0   http_plugin.cpp:285           plugin_startup       ] start listening for http requests
**252121ms thread-0   http_plugin.cpp:291           plugin_startup       ] http: Underlying Transport Error**
252121ms thread-0   wallet_api_plugin.cpp:74      plugin_startup       ] starting wallet_api_plugin
252121ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/create
252121ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/create_key
252121ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/get_public_keys
252121ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/import_key
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/list_keys
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/list_wallets
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/lock
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/lock_all
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/open
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/set_timeout
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/sign_transaction
252122ms thread-0   http_plugin.cpp:325           add_handler          ] add api url: /v1/wallet/unlock

3. $ ps -ef | grep keosd
[...]

PR #3489, opened 2018-05-28
keosd/nodeos process exits if failed to bind to http port #3314

moskvanaft:


spoonincode:
Be aware this causes a change in behavior to nodeos too -- it will shutdown on http/https start failure. This seems... okay enough to change at this point.

PR #3491, opened 2018-05-28
Show start_block exception

tbfleming:


libraries/chain/controller.cpp
```diff
@@ -750,8 +750,10 @@ struct controller_impl {
       try {
          auto onbtrx = std::make_shared<transaction_metadata>( get_on_block_transaction() );
          push_transaction( onbtrx, fc::time_point::maximum(), true, self.get_global_properties().configuration.min_transaction_cpu_usage );
+      } catch ( const fc::exception& e ) {
```
wanderingbort: consider using the `FC_LOG_AND_DROP` macro, it seems to be a superset of what you have here but also catches standard exceptions and makes an attempt at handling `...` 

Upon inspection, it seems like that macro is only using the passed in format args in the cases where it is _not_ an fc::exception which may not be ideal but is basically equivalent to this

PR #3493, opened 2018-05-28
Switch tests to use bnet plugin.

cj-oci:


tests/testUtils.py
```diff
@@ -1370,7 +1370,7 @@ def launch(self, pnodes=1, totalNodes=1, prodCount=1, topo="mesh", delay=1, only
         if len(self.nodes) > 0:
             raise RuntimeError("Cluster already running.")
 
-        cmd="%s -p %s -n %s -s %s -d %s -f" % (
+        cmd="%s -p %s -n %s -s %s -d %s -f --p2p-plugin bnet" % (
             Utils.EosLauncherPath, pnodes, totalNodes, topo, delay)
```
heifner: Ideally, we would like to run tests on both plugins. However, the tests already take a long time to run. Maybe we could switch the plugin used randomly?

Issue #3424, opened 2018-05-25
Difference between wavm and binaryen


tbfleming:
Noticed while developing cib:
* binaryen creates numbered names for any unnamed functions
* if these conflict with existing names in the wasm, then binaryen rejects the binary
* I suspect the wavm flow doesn't have this issue

This needs a test case to determine if there's a difference in behavior.

tbfleming:
```test.abi```:
```
{
  "structs": [{
      "name": "hi",
      "base": "",
      "fields": [{
          "name": "user",
          "type": "name"
        }
      ]
    }
  ],
  "actions": [{
      "name": "hi",
      "type": "hi"
    }
  ],
}
```

To reproduce:

* Create an account named ```test```
* Run the following:

```
[...]

PR #3508, opened 2018-05-28
Ignore user sections #3424

tbfleming:


spoonincode:
Can we add a unit test for the problematic wasm?

PR #3511, opened 2018-05-28
FIX Pass string with whitespaces to cleos in cleos.sh

sudoaza:
when passing string arguments to cleos.sh like --stake-net "100000.0000 EOS" it would fail with unrelated error ERROR: ExtrasError: [EOS] caused by the interpolation of the string argument.


bytemaster:
This could combine multiple args to one... can you post example that this fixed?

sudoaza:
```
./cleos.sh system newaccount eosio aza111111114 \
EOS6uoGxeNF69iPSsQHM8fvLGnEQ5BACSgFGS6WbE98DcVafmSgA9 \
EOS6uoGxeNF69iPSsQHM8fvLGnEQ5BACSgFGS6WbE98DcVafmSgA9 \
--buy-ram-EOS "100000.0000 EOS" --stake-net "100000.0000 EOS" --stake-cpu "100000.0000 EOS"
```

bug

sudoaza:
@bytemaster could you provide an input where this combines multiple arguments into one?

pdahlberg:
Thanks. This helped me past the step: 
$ cleos push action eosio.token create '[ "eosio", "1000000000.0000 SYS"]' -p eosio.token

... in this tutorial: https://github.com/EOSIO/eos/wiki/Tutorial-eosio-token-Contract

heifner:
@bytemaster I believe this is needed so that `"` are passed into the script.

wanderingbort:
This is the correct way to maintain the top-level tokenization of the input stream in a bash script.

Issue #3497, opened 2018-05-28
Limit P2P Connections from Same IP


bytemaster:
To prevent DOS attacks against public endpoints we need to limit the number of connections from the same IP.

pmesnier:
Right. In this case, the limit should be 1. Using multiple connections to avoid message backlogs or to have a priority band is useful and we may want to add that later, however in that case we should have a second listen socket bound to some ephemeral port number which is supplied to the peer during handshake, and maybe only if a handshake is received signed by a known peer. But I digress.

bytemaster:
for our testing we would like the option to have the same IP connect multiple times, but it should default to banning it.

pmesnier:
how about an option p2p-max-nodes-per-host. I was translating IP address to "node" but you mean from multiple nodes on a given host address. Should this also be enforced on localhost?

spoonincode:
Because the default max number of connections is 25 for net_plugin, it seems like this sort of change won't get you much for out of box protection?

bytemaster:
This can be addressed at proxy level.

PR #3517, opened 2018-05-28
fix for issue #3497

pmesnier:
add new config item, p2p-max-nodes-per-peer, to limit the number of client connections accepted from any one host IP address.


stvenyin:
收到！

PR #3550, opened 2018-05-29
Expose runtime options in the producer api plugin

### LINKED ISSUE #3487: --enable-stale-production should override --max-irreversible-block-age
when the user wants to reboot a blockchain they normally boot with -e, but unless --max-irreversible-block-age is set to something greater than the "down time" it will not start producing.

Because -e only applies to the first block a node probably doesn't want --max-irreversible-block-age to be changed from its default.

wanderingbort:
This resolves #3487 and restores functionality regarding #3299 that was removed ineb8f4c4a1e5a77ec38dd2dbb40ed4826631b0dd6

This adds 2 new HTTP endpoints:

`POST /v1/producers/get_runtime_options` - takes no payload and produces a JSON in the form:

```
{"max_transaction_time":400,"max_irreversible_block_age":1800}
```

which contains the values for the corresponding config.ini/command-line parameters in use.

`POST /v1/producers/update_runtime_options` - takes a JSON formed as above (each field is optional so you don't have to set _all_ options, you may use it sparsely.

These settings will take effect immediately.

### DIFF
--- a/plugins/producer_api_plugin/producer_api_plugin.cpp
+++ b/plugins/producer_api_plugin/producer_api_plugin.cpp
@@ -72,6 +72,10 @@ void producer_api_plugin::plugin_startup() {
             INVOKE_V_V(producer, resume), 201),
        CALL(producer, producer, paused,
             INVOKE_R_V(producer, paused), 201),
+       CALL(producer, producer, get_runtime_options,
+            INVOKE_R_V(producer, get_runtime_options), 201),
+       CALL(producer, producer, update_runtime_options,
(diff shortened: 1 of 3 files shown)


wanderingbort:
`--enable-stale-production` does not override this check.  Its purpose (allbeit poorly named) is to prevent a newly started node from producing before it hears a new-ish block.  

My guess is this is a genesis.json timestamp that is very old and we are bootstrapping into a new test chain.  In that case `--genesis-timestap now` will help start a chain where the genesis block is not older than 30 minutes.

PR #3554, opened 2018-05-29
Network buffers

heifner:
The default websocket asio http buffer is 32MB. Since block size is 1MB, reduced the max message size and http body size to 1M + 512K.

Also modified the bnet send/receive buffer down the the same value since a block could be larger than 1MB since that is a soft limit not a hard limit.


jeaimetu:
Hi,
Why don't you also change TCP buffer size. It's maximum is 8Mbyte. Even you change http buffer size from 32MByte to 1.5MByte, it will segmented in TCP layer. So 1.5Mbyte is safe figure to consider MIME encoding, then you'd better change TCP buffer size from 8KByte(default) to 1.5MByte + 4KByte(TCP header + some reservation). Then it reduced the load of BP server itself. Even MAC other routers can bother this efficiency.

But to change this buffer size need to be tested in Jungle test net or any staging servers to optimize the buffer Size.

spartucus:
> Since block size is 1MB

Hi Heifner, couldn't find any info about this, any tips?

PR #3570, opened 2018-05-29
 bios-boot-tutorial: update how accounts are staked and funded

tbfleming:


rbw3821:
The only question I have is whether the length of time that it might take someone to run this with a high number of users should be documented either in the script or in the documentation associated with the script.

Issue #3540, opened 2018-05-29
WAVM: Incorrect bounds check when translating a reference type can results in buffer overrun


guhe120:
When translating an offset to wasm memory for a reference, the bounds check is as following:

At wavm.hpp, line 585:

      if(!mem || ptr+sizeof(T) >= IR::numBytesPerPage*Runtime::getMemoryNumPages(mem))

         Runtime::causeException(Exception::Cause::accessViolation);

Here |ptr|'s type is I32, which means it can be a negative value, so in the below situation when
  -sizeof(T) < ptr < 0,   ptr+sizeof(T) will be a very small value which can bypass this check,

And later,

 T &base = *(T*)(getMemoryBaseAddress(mem)+ptr);

|base| will points beyond the wasm memory base, which means we can read/write beyond the wasm memory buffer.

This is a submission to the EOS bug bounty program.

This bug credits to:

Yuki Chen of Qihoo 360 Vulcan Team.

Thank you!

guhe120:
Hi,

Quick & Nice fix.
Now may I know if this submission is eligible for your bug bounty program? thank you!

viccon:
Congratulations @guhe120 !
We have decided to reward $300,000 for this report.

learnforpractice:
Hey, guys, should we check the overflow of (U32)ptr+sizeof(T) ?

learnforpractice:
Well, since the type of sizeof(T) is an unsigned long(8 bytes on x64 platform), so compiler will expand (U32)ptr(4 bytes) to unsigned long, there is no need to worry about overflow. Thank goodness.

Issue #3541, opened 2018-05-29
Binaryen: array_ptr_impl can cause buffer overflow in 32-bits build


guhe120:
Hi,

In function array_ptr_impl in Binaryen.hpp:

template<typename T>

inline array_ptr<T> array_ptr_impl (interpreter_interface* interface, uint32_t ptr, uint32_t length)

{
   return array_ptr<T>((T*)(interface->get_validated_pointer(ptr, length * (uint32_t)sizeof(T))));
}

Here we have full control of the |length| parameter, in 32-bits process, length * (uint32_t)sizeof(T) could overflow 32-bits integer range and results in a very small value, which then bypasses the boundary check in get_validated_pointer and returns an over-sized array_ptr.

Then we can read/write out of the bounds of the wasm memory buffer with this over-sized array_ptr.

This is a submission to EOS bug bounty program.

This bug credits to:

Yuki Chen of Qihoo 360 Vulcan Team.

Thank you!

PR #3572, opened 2018-05-29
Resolved issues #3540 and #3541

larryk85:
Various fixes from clang tidy and bug reports.


libraries/chain/include/eosio/chain/webassembly/binaryen.hpp
```diff
@@ -157,6 +157,7 @@ class binaryen_runtime : public eosio::chain::wasm_runtime_interface {
 template<typename T>
 inline array_ptr<T> array_ptr_impl (interpreter_interface* interface, uint32_t ptr, uint32_t length)
 {
+   FC_ASSERT( length < INT_MAX/(uint32_t)sizeof(T), "length will overflow" );
```
wanderingbort: Is there a reason to have `INT_MAX` vs unsigned int max?  Also can this be `<=` ?  the division will truncate down so if length == a/b then length * b should be <= a right?

bytemaster: in theory we should be nowhere near INT_MAX and bart's suggestion would only serve to increase the range to the maximum "safe" limit.  

Since this is suppose to be a sanity check, I think it is already conservative. No WASM module should address more than 32MB of RAM.

PR #3586, opened 2018-05-30
support passing CORE_SYMBOL_NAME parameter to eosio_build.sh

littlekfc:
The default CORE_SYMBOL_NAME is `SYS`. There is no way to pass `DCORE_SYMBOL_NAME` to cmake file. Need support a parameter to eosio_build.sh.
Usage:    `./eosio_build.sh -s EOS`

### DIFF
--- a/eosio_build.sh
+++ b/eosio_build.sh
@@ -61,7 +61,7 @@
 	txtrst=$(tput sgr0)
 
 	if [ $# -ne 0 ]; then
-		while getopts ":cdo:" opt; do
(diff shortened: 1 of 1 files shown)


littlekfc:
After merging, we should update the wiki [create-and-allocate-the-sys-currency](https://github.com/EOSIO/eos/wiki/Tutorial-Bios-Boot-Sequence#create-and-allocate-the-sys-currency). Since it needs to config the symbol before compiling.

arhag:
Replaced by #3691.

PR #3587, opened 2018-05-30
Add snapshot plugin

elmato:
This PR adds a new plugin that allows to take a snapshot of the blockchain state at a determined block.
The snapshot is taken after the block is applied and its saved when that block becomes irreversible.

The generated snapshot contains:
 - The chain id
 - The genesis state
 - All accounts (with permissions)
 - All the tables

### DIFF
--- a/libraries/chain/include/eosio/chain/account_object.hpp
+++ b/libraries/chain/include/eosio/chain/account_object.hpp
@@ -80,5 +80,4 @@ namespace eosio { namespace chain {
 CHAINBASE_SET_INDEX_TYPE(eosio::chain::account_object, eosio::chain::account_index)
 CHAINBASE_SET_INDEX_TYPE(eosio::chain::account_sequence_object, eosio::chain::account_sequence_index)
 
-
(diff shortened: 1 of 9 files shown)


abourget:
Just so you know.. I also have a solution to freeze a chain at a given block. It's in `eosc`.. a sort-of eos swiss-knife.

You can run:
```
$ eosc bps chain-freeze --help
Freeze the chain by proxying p2p blocks until a block including updateauth actions is passed through, then block/shutdown.

Usage:
  eosc bps chain-freeze [flags]

Flags:
      --listening-address string    return producers info in json (default ":19876")
  -t, --target-p2p-address string   return producers info in json (default "localhost:9876")
```

It listens and sniffs the p2p protocol, and stops propagation, leaving you with a blockchain frozen at a given state.  You can then normally query the contracts and all, without fearing it will continue.

abourget:
whoops, the link is here: https://github.com/eoscanada/eosc

elmato:
Replaced by #4837

PR #3592, opened 2018-05-30
Remove unused variable

spartucus:
Remove unused variable 'init', it looks like there's no need to defining a static variable.


libraries/fc/src/crypto/elliptic_r1.cpp
```diff
@@ -16,7 +16,7 @@ namespace fc { namespace crypto { namespace r1 {
           public_key_impl()
           :_key(nullptr)
           {
-          static int init = init_openssl();
+            init_openssl();
```
heifner: The static is used so that init_openssl() is only called once.

spartucus: Got it!

PR #3595, opened 2018-05-30
Gh#3567 bnet additional threading fixes

heifner:
Fix to prevent _sessions from being accessed from multiple threads.
Found by visual inspection and verified by testing.


plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -1085,21 +1081,24 @@ namespace eosio {
          }
 
          void on_session_close( const session* s ) {
+            if( !app().get_io_service().get_executor().running_in_this_thread() ) { elog( "wrong strand"); }
```
wanderingbort: I know we have this pattern in several handlers, but what is the "downside" of conditions where this message enters the log.  Does this imply that we have potential for concurrent access to `_session` which can cause undefined behavior/crashes?  

If so, we should probably have better handling than just a log message.   Halt-and-catch fire perhaps, or maybe optionally a locking strategy where this path can be expensive but safe?

heifner: A better solution then might be to log and raise a SIGTERM so that process is shutdown cleanly without corrupting the block log.

PR #3609, opened 2018-05-30
Build Amazon: moved directory opt/boost_1_67_0 to opt/boost 

pacificcode:
moved directory opt/boost_1_67_0 to opt/boost


pacificcode:
@heifner This was for ease of build kite builds (as per Eric's request) & for future boost upgrades.

Issue #3596, opened 2018-05-30
keosd private key reveal needs to require password


spoonincode:
It's possible to ask keosd to provide the private keys from unlocked wallets. This is deemed not sufficient -- the password should be given again to provide these private keys even if the wallet is already unlocked.

This may mean a change to how `list_keys` works, where now it would require two parameters: the wallet to list private keys from and the password to said wallet.

PR #3629, opened 2018-05-30
Require wallet password when revealing private keys

spoonincode:
In addition to being unlocked, wallets now require their password to reveal their private keys.

🚨🚨RPC & cleos changes in here

RPC endpoint changes:
`/v1/wallet/get_public_keys` -- No change. This endpoint prints all public keys from all unlocked wallets.
`/v1/wallet/list_keys` -- This endpoint now requires two strings: the name of an unlocked wallet and its password. It will then reveal the public and private key pairs of that wallet. Be aware this means a slight change in behavior in that it is now impossible to query private keys from all unlocked wallets.

cleos changes:
`wallet keys` -- This command has been changed to print all public keys from unlocked wallets (corresponds to the get_public_keys endpoint). Be aware that there is a change in behavior due to different behavior in how get_public_keys works vs previous behavior: if there is no open wallet or all wallets are locked, wallet keys will now return a "no available wallet" or "locked wallet" error. Previously an empty list would be returned.
`wallet private_keys` -- This is a new command that maps to the list_keys endpoint. It will take the wallet name and password as arguments before printing the public/private key pairs.

Issue #3596

### DIFF
--- a/plugins/wallet_plugin/wallet.cpp
+++ b/plugins/wallet_plugin/wallet.cpp
@@ -327,6 +327,16 @@ void wallet_api::unlock(string password)
 } EOS_RETHROW_EXCEPTIONS(chain::wallet_invalid_password_exception,
                           "Invalid password for wallet: \"${wallet_name}\"", ("wallet_name", get_wallet_filename())) }
 
+void wallet_api::check_password(string password)
+{ try {
+   FC_ASSERT(password.size() > 0);
+   auto pw = fc::sha512::hash(password.c_str(), password.size());
+   vector<char> decrypted = fc::aes_decrypt(pw, my->_wallet.cipher_keys);
+   auto pk = fc::raw::unpack<plain_keys>(decrypted);
+   FC_ASSERT(pk.checksum == pw);
+} EOS_RETHROW_EXCEPTIONS(chain::wallet_invalid_password_exception,
+                          "Invalid password for wallet: \"${wallet_name}\"", ("wallet_name", get_wallet_filename())) }
+
 void wallet_api::set_password( string password )
 {
    if( !is_new() )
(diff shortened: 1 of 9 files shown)


plugins/wallet_plugin/wallet.cpp
```diff
@@ -327,6 +327,16 @@ void wallet_api::unlock(string password)
...
+void wallet_api::check_password(string password)
+{ try {
+   FC_ASSERT(password.size() > 0);
+   auto pw = fc::sha512::hash(password.c_str(), password.size());
+   vector<char> decrypted = fc::aes_decrypt(pw, my->_wallet.cipher_keys);
+   auto pk = fc::raw::unpack<plain_keys>(decrypted);
+   FC_ASSERT(pk.checksum == pw);
+} EOS_RETHROW_EXCEPTIONS(chain::wallet_invalid_password_exception,
+                          "Invalid password for wallet: \"${wallet_name}\"", ("wallet_name", get_wallet_filename())) }
+
```
heifner: This does exactly what unlock does minus the assignment which would not hurt. Seems like you could just call `unlock`

spoonincode: I thought about that too but my concern is if I call `unlock()` on an already unlocked wallet. That will replace `_keys` with the decrypted keys. What if `_keys` has new keys that have not been written out to `cipher_keys` yet? Seems like I could potentially lose unsaved keys that way. I was a little surprised that `import_key()` for example doesn't save the wallet -- so `_keys` in that case will have contents that don't exist in `chiper_keys` until `encrypt_keys()` is called.

Basically I couldn't convince myself of the safety here.

spoonincode: Maybe one approach to reducing code duplication here is to simply have unlock() call the new check_password()

spoonincode: I'm a little nervous to mess with it...

spoonincode: Actually now that I look at it again I see that we do save right after import/create. Still, I'm a little nervous touching this further at this juncture

Issue #3581, opened 2018-05-30
blacklist producers


bytemaster:
A node should be able to blacklist producers which produce blocks which may contain an infinite loop.

In general DPOS trusts producers due to the election model, this means we trust them not to produce blocks with infinite loops and the consequences can be dealt with in arbitration / etc.  If a producer does "go bad" and attempt to bring down the network everyone else should be able to respond quickly to minimize damage by ignoring that producer.

abourget:
Is there a way to do that without hard-forking instantly ?

Tweicx:
WTF! a Epic bug!holy shit !!!

taokayan:
After discussing with @arhag , this turns out to be a tricky task. Making a blacklist on the producer plugin might introduce some synchronization issue between different producers. Also making it into a contract level might result in a big delay.

nehtechnine:
"load"reviewatom"

aclark-b1:
In order to focus our efforts on issues that are currently creating difficulty for the community we are closing tickets that were created prior to the EOSIO 2.0 release.  If you believe this issue is still relevant please feel free to reopen it or create a new one.  Thank you for your continued support of EOSIO!

PR #3632, opened 2018-05-30
Blacklist

taokayan:
#3581 allow blacklist producers

### DIFF
--- a/contracts/eosio.system/eosio.system.abi
+++ b/contracts/eosio.system/eosio.system.abi
@@ -386,6 +386,20 @@
       "fields": [
         {"name":"params", "type":"eosio_parameters"}
       ]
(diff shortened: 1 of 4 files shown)


taokayan:
I think making a black-list in the config might create some more problems. For example, if a malicious producer produced some blocks 1 year ago when he was not yet identified malicious, which makes recovery difficult to handle. And if each producer is going to modify their own black list separately, it's likely to be un-synced in such a way that some bad blocks will accepted by one producer but not accepted by another, creating forks, and may not possible be merged.

brianjohnson5972:
Can we close this PR, since it is so old?

taokayan:
Yes, already closed. Thanks.

PR #3643, opened 2018-05-31
Thomasbcox rc edits

thomasbcox:
removed deprecated arbitration clauses
renamed file to match naming standard


contracts/eosio.system/eosio.system-buyram-rc.md
```diff
@@ -1,19 +1,16 @@
...
-This Contract is legally binding and can be used in the event of a dispute. 
-
 ### Description
 
-The `{{ buyram }}` action... 
-
-### Inputs and Input Types
-
-The `{{ buyram }}` action requires the following `inputs` and `input types`:
+This action will attempt to reserve about {{quant}} SYS tokens worth of RAM on behalf of {{receiver}}.
```
gleehokie: @thomasbcox I think you could remove all of the "SYS" mentions from these documents without losing their meaning and then the docs would not need to be modified by each of the chains that rename the tokens.

thomasbcox: No actually you can't.

thomasbcox: OK, Dan says you can and indeed should, because {{quant}} contains the token ID (EOS, SYS, etc.) at least in THIS specific example.

PR #3670, opened 2018-05-31
Create missing system RCs

thomasbcox:


thomasbcox:
This creates stub RCs for newer System actions.

PR #3673, opened 2018-05-31
OOB block signing

### LINKED ISSUE #2586: Support block producer private key operations to be isolated in helper process
Block producer should be able to isolate its private key in separate process / separate host to protect from exfiltration if node is compromised.

Most flexible and simple option would be to delegate private key signing operation to a helper process, which may optionally delegate to HSM or other system.

wanderingbort:
This adds the ability to have an out of band block signer to the producer plugin

There are lots of things that were added/changed and each of the commits below has a bullet list of the changes.

A quick synopsis is:
 - added `signature-provider` as a configuration option
  - this takes <public key>=<provider> pairs (multiple if more than one public key is present)
  - <provider> is a <TYPE>:<DATA> pair where type can be
    - `KEY` for an in-process WIF encoded key where <DATA> is the WIF-encoded private key
    - `KEOSD` for an out-of-process keosd hosted key where <DATA> is the url to submit the signing request
      - HTTPS is supported
 - `private-key` has been DEPRECATED as a config option (but still works for now)
  - the canonical way of specifying in-process signing with a WIF encoded key is `signature-provider = <public key>=KEY:<WIF encoded private key>`

Other notable config options:
  - `keosd-provider-timeout` : the number of milliseconds the `producer_plugin` will wait for keosd to sign a block before giving up (this will abort the block)
  - `https-client-root-cert` : a PEM string (or path to file with a PEM in it) for a CA cert to trust when validating peers. 
    - this can be specified multiple times
    - sadly supporting the OS trust stores will come later
  - `https-client-validate-peers` : true/false (defaults to true).  Controls whether we allow connections to untrusted or invalid cert'd peers

this resolves #2586


libraries/fc/include/fc/static_variant.hpp
```diff
@@ -245,15 +245,25 @@ class static_variant {
...
 
-    template<typename X>
+    /*template<typename X>
     static_variant(const X& v) {
         static_assert(
             impl::position<X, Types...>::pos != -1,
             "Type not in static_variant."
         );
         init(v);
+    }*/
```
heifner: Might as well remove this and the one above.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+
+   template<typename SyncReadStream, typename Fn, typename CancelFn>
+   error_code sync_do_with_deadline( SyncReadStream& s, deadline_type deadline, Fn f, CancelFn cf ) {
+      optional<error_code> timer_result;
+      boost::asio::deadline_timer timer(s.get_io_service());
+
+      timer.expires_at(deadline);
+      bool timer_cancelled = false;
+      timer.async_wait([&timer_result, &timer_cancelled] (const error_code& error) {
+         if (!error && !timer_cancelled) {
```
heifner: Had to read that a couple of times and apply De Morgan's Law to understand. Would be easier to read as `if (error || timer_cancelled)`. Ah but maybe you were avoiding optional bool operator.  `if (!!error || timer_cancelled)` might still be better but then it is much more subjective.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+            timer_result.emplace(error);
+         }
+      });
+
+      optional<error_code> f_result;
+      f(f_result);
+
+      s.get_io_service().reset();
```
heifner: `reset` is deprecated. Use `restart`.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+      if (enabled) {
+         _sslc.set_verify_mode(ssl::verify_peer);
+      } else {
+         _sslc.set_verify_mode(ssl::verify_none);
+      }
+   }
+
+   template<typename SyncReadStream, typename Fn, typename CancelFn>
+   error_code sync_do_with_deadline( SyncReadStream& s, deadline_type deadline, Fn f, CancelFn cf ) {
+      optional<error_code> timer_result;
```
heifner: I think timer_error would be a better name.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+            timer.cancel();
+         } else if (timer_result) {
+            cf();
+         }
+      }
+
+      if (!timer_result) {
+         return *f_result;
+      } else {
+         return error_code(boost::system::errc::timed_out, boost::system::system_category());
```
heifner: Might be worthwhile preserving the error_code as it is not necessarily timed_out.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+      static const deadline_type epoch(boost::gregorian::date(1970, 1, 1));
+      auto deadline = epoch + boost::posix_time::microseconds(_deadline.time_since_epoch().count());
+
+      string path = dest.path() ? dest.path()->generic_string() : "/";
+      if (dest.query()) {
+         path = path + "?" + *dest.query();
+      }
+
+      http::request<http::string_body> req{http::verb::post, dest.path()->generic_string(), 11};
+      req.set(http::field::host, *dest.host());
```
heifner: optional host might not be set.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+   variant post_sync(const url& dest, const variant& payload, const fc::time_point& _deadline) {
+      static const deadline_type epoch(boost::gregorian::date(1970, 1, 1));
+      auto deadline = epoch + boost::posix_time::microseconds(_deadline.time_since_epoch().count());
+
+      string path = dest.path() ? dest.path()->generic_string() : "/";
+      if (dest.query()) {
+         path = path + "?" + *dest.query();
+      }
+
+      http::request<http::string_body> req{http::verb::post, dest.path()->generic_string(), 11};
```
heifner: use path since you calculated it above and optional path might not be set.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,361 @@
...
+            excp = std::make_shared<fc::exception>(err_var["code"].as_int64(), err_var["name"].as_string(), err_var["what"].as_string());
+
+            if (err_var.contains("details")) {
+               auto details = err_var["details"].get_array();
+               for (const auto dvar : details) {
+                  excp->append_log(FC_LOG_MESSAGE(error, dvar.get_object()["message"].as_string()));
+               }
+            }
+         } catch( ... ) {
+
```
heifner: A log statement would be good here. If "error" is ever renamed to something else this will silently fail.

heifner: Nevermind, I see you FC_THROW below if this fails.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -455,6 +466,31 @@ if( options.count(name) ) { \
    std::copy(ops.begin(), ops.end(), std::inserter(container, container.end())); \
 }
 
+static producer_plugin_impl::signature_provider_type
+make_key_signature_provider(const private_key_type& key) {
+   return [key]( const chain::digest_type& digest ) {
```
heifner: `key` can be `&key`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -455,6 +466,31 @@ if( options.count(name) ) { \
...
+      return key.sign(digest);
+   };
+}
+
+static producer_plugin_impl::signature_provider_type
+make_keosd_signature_provider(const std::shared_ptr<producer_plugin_impl>& impl, const string& url_str, const public_key_type pubkey) {
+   auto keosd_url = fc::url(url_str);
+   std::weak_ptr<producer_plugin_impl> weak_impl = impl;
+
+   return [weak_impl, keosd_url, pubkey]( const chain::digest_type& digest ) {
```
heifner: These can all be by ref as well.

libraries/fc/src/network/http/http_client.cpp
wanderingbort: So, it looks odd for sure.   I think I will change the way this works because there are only 2 options for the invocation of this according to boost:

call with a "success error" meaning the timer successfully expired OR `boost::asio::error::operation_aborted` meaning it was canceled when not in-flight.  the `timer_cancelled` bool handles canclations when the success is already in flight.  

All the code cares about is whether the timer successfully expired.  So, only when this is called with a "success error" (aka !error) and we havent set the cancelled flag, should we set the timer_result to something ...

abourget:
This will land in version 1.0 ?

heifner:
Yes.

libraries/fc/src/network/http/http_client.cpp
```diff
@@ -0,0 +1,365 @@
...
+      req.prepare_payload();
+
+      auto conn_iter = get_connection(dest, deadline);
+      auto eraser = make_scoped_exit([this, &conn_iter](){
+         _connections.erase(conn_iter);
+      });
+
+      // Send the HTTP request to the remote host
+      error_code ec = conn_iter->second.visit(write_request_visitor(this, req, deadline));
+      FC_ASSERT(!ec, "Failed to send request: ${message}", ("message",ec.message()));
```
ellipticasec: @wanderingbort won't this assert interrupt the program ? maybe it's a bit harsh if `signature_provider` is not yet ready to sign.

PR #3681, opened 2018-05-31
Disallow net_plugin empty messages; guard peeking at varint

spoonincode:
Add two hardening checks on initial message parsing in net_plugin:

Don't accept messages that indicate a length of 0 -- we would need at least the net_message's variant "what" varint to have a useful message.

Stop peeking at a net_message's what varint after 5 bytes; anything more is invalid.

This along with PR #3675 will complete issue #3662

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1188,7 +1188,7 @@ namespace eosio {
             pending_message_buffer.peek(&b, 1, index);
             which |= uint32_t(uint8_t(b) & 0x7f) << by;
             by += 7;
-         } while( uint8_t(b) & 0x80 );
+         } while( uint8_t(b) & 0x80 && by < 32);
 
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1188,7 +1188,7 @@ namespace eosio {
             pending_message_buffer.peek(&b, 1, index);
             which |= uint32_t(uint8_t(b) & 0x7f) << by;
             by += 7;
-         } while( uint8_t(b) & 0x80 );
+         } while( uint8_t(b) & 0x80 && by < 32);
```
heifner: Can this code be replaced with a call to `fc::io::raw::unpack(Stream&, unsigend int&)` ?

heifner: Sorry, no it can't. This is a peek, not a read.

PR #3685, opened 2018-05-31
cleos newaccount: require either --buyram-bytes or --byuram-eos, allow to delegate nothing

moskvanaft:


heifner:
I think we also need to rename `--buyram-eos` to just `--buyram` as symbol is not necessarily EOS. Also `--buyram-bytes` might be better named `--buyram-kbytes`

PR #3704, opened 2018-05-31
Got rid of most warnings, cleaned eosio_system_tests, added more rmvproducer checks

zorba80:


contracts/exchange/exchange.cpp
```diff
@@ -112,7 +112,7 @@ namespace eosio {
 
    void exchange::createx( account_name    creator,
                  asset           initial_supply,
-                 uint32_t        fee,
+                 uint32_t        /* fee */,
```
larryk85: can we just remove this

contracts/eosio.bios/eosio.bios.hpp
```diff
@@ -15,7 +15,7 @@ namespace eosio {
 
          void setalimits( account_name account, uint64_t ram_bytes, uint64_t net_weight, uint64_t cpu_weight ) {
             require_auth( _self );
-            set_resource_limits( account, ram_bytes, net_weight, cpu_weight );
+            set_resource_limits( account, (int64_t)ram_bytes, (int64_t)net_weight, (int64_t)cpu_weight );
```
larryk85: should we check for overflow

PR #3712, opened 2018-06-01
Update asset.hpp

OracleChain:
asset should be no negative number

for example:
```
         #define CORE_SYMBOL S(4,SYS)
         eosio::asset amount(0, CORE_SYMBOL);
         eosio::asset amounthalfmax(max_amount/2, CORE_SYMBOL);
         amount -= amounthalfmax;
```
It should not pass through


bytemaster:
It is possible to represent negative amounts.  This change would break things.

PR #3718, opened 2018-06-01
Gh#3715 stack alloc

### LINKED ISSUE #3715: Contract stack size is limited, minimize use
Contracts should avoid allocating more than 128 bytes on the stack since stack size is limited.

Update use of buffers in:
- action.hpp
- transaction.hpp
- eosio.msig

heifner:
Modified use of stack buffer to minimize stack buffer size to 128.

Resolves #3715


heifner:
Conversations with @bytemaster and @arhag decided to go with 512 to allow transfer with full 256 memo to fit.

PR #3724, opened 2018-06-01
regproducer.md

Commoneffort:
This is the latest version of the BP Agreement


gleehokie:
The BP agreement has been moved into community run repos and is no longer stored in this repo.

PR #3742, opened 2018-06-01
Relocate wiki images and remove unused Doxyfile.

jgiszczak:


jgiszczak:
The multinode private testnet wiki page has a new diagram, and no longer references the original three.  If that was the only place they were used, I can push a new commit to remove the images and their .dot files entirely.

gleehokie:
Hold off on that @jgiszczak, I'm not sure of the status.  @rbw3821 will follow up on it.

gleehokie:
@jgiszczak please delete circle.yml and Jenkinsfile as well (I confirmed with @ericiles).

PR #3784, opened 2018-06-03
wallet_api interface & soft_wallet refactor

spoonincode:
This is a non-functionally change that creates a wallet_api interface and renames the current wallet class to soft_wallet. wallet_manager will now operate strictly on wallet_api instances with the exception of when it is asked to create/open a wallet in which case it will create a soft_wallet instance.

The idea is that hardware wallet implementations will implement wallet_api. wallet_manager will add instances of these hardware wallet objects to its wallet_api list.

~~The biggest change is that a new method is added on sign_transaction so that when signing the transaction an external function is called to complete the signing. This seemed to be the best (least intrusive) way to modify signed_transaction to accomplish signing when it didn't have access to the private key. Open to alternate suggestions.~~

A couple other minor changes:

Addition of a list_public_keys() call and wallet_manager now using it for its get_public_keys() call. Hardware wallets won't have the ability to get private keys, so a function that didn't return a private_key_type was required.

~~Addition of a is_lockable() call. Some wallet_api implementations for hardware wallets may not have the ability to lock. At least in their initial implementation. I would like attempts to lock unlockable wallets to fail; so add a way for lock_all() to query if it should lock before attempting to lock said wallets. This limitation could be removed in the future if lockablity is better understood in all cases.~~

Work on #2252


bytemaster:
Lockable is a property of keos and simply means keos will not sign for a particular key and if possible will unload it form memory.  A hardware wallet keeps the key out of RAM by default, but still has the concept of being "locked" and "unlocked" based upon a wallet password.

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -90,6 +92,7 @@ namespace eosio { namespace chain {
       vector<bytes>             context_free_data; ///< for each context-free action, there is an entry here
 
       const signature_type&     sign(const private_key_type& key, const chain_id_type& chain_id);
+      const signature_type&     sign(sign_digest_functor signature_request, const chain_id_type& chain_id);
```
wanderingbort: This seems a little cumbersome.  I would naively suggest pulling a digest from the signed_transaction, passing it to another system and then adding a method to (or just directly) append a signature to a signed transaction.  If we want to be "safe" we could make this method validate that the signature is "correct" by recovering and comparing it to a provided public key

spoonincode: This can be done but it removes a fair bit of encapsulation of signed_transaction IMO

1) signed_transaction::sign knows that to compute the digest for signing it must take in to account the context_free_data of the signed_transaction
2) signed_transaction::sign will modify its signatures

granted _in both cases these are public members_. So maybe should not worry about it. The resulting code would look something like

```
signed_transaction stx;
chain_id_type cid;
digest_type dst = sig_digest(cid, stx.context_free_data);
stx.signatures.push_back(priv_key.sign(dst));
```

PR #3797, opened 2018-06-04
system contract bugfix: do not check if key is different

moskvanaft:
there can be other different parameters.


abourget:
I noticed you fixed the default `location`, but last time I checked, `location` wasn't used in the schedule preparation.. has that changed ?

moskvanaft:
@abourget 
No, we still don't user `location`.

PR #3800, opened 2018-06-04
Add new test for dirty db flag stickiness.

cj-oci:
Nodeos checks for db corruption upon startup. This test validates the corruption flag remains set upon repeat nodeos restart attempts.

### DIFF
--- a/tests/validate-dirty-db.py
+++ b/tests/validate-dirty-db.py
@@ -1,0 +1,130 @@
+#!/usr/bin/env python3
+
+import testUtils
+
+import argparse
+import random
+import subprocess
+import time
+import os
+import signal
+
+###############################################################
+# Test for validating the dirty db flag sticks repeated nodeos restart attempts
+###############################################################
+
+
+Print=testUtils.Utils.Print
+
+def errorExit(msg="", errorCode=1):
+    Print("ERROR:", msg)
+    exit(errorCode)
+
+parser = argparse.ArgumentParser()
+parser.add_argument("-v", help="verbose logging", action='store_true')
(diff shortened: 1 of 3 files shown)


tests/validate-dirty-db.py
```diff
@@ -0,0 +1,130 @@
...
+        """Startup nodeos, wait for waitTime (before forced shutdown) and collect output. Stdout, stderr and return code are returned in a dictionary.
+        Large waittimes have potential to deadlock as the popen PIPE buffer can get filled up."""
+        Print("Launching nodeos process id: %d" % (nodeId))
+        dataDir="var/lib/node_%02d" % (nodeId)
+        cmd="programs/nodeos/nodeos --config-dir etc/eosio/node_bios --data-dir var/lib/node_bios"
+        Print("cmd: %s" % (cmd))
+        proc=subprocess.Popen(cmd.split(), stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        try:
+            proc.wait(waitTime)
```
brianjohnson5972: Because of the Note about deadlocking on wait call, I think we should use proc.communicate(timeout=waitTime).

tests/validate-dirty-db.py
```diff
@@ -0,0 +1,130 @@
...
+
+        #Print(ret)
+
+        stderr=ret[1]["stderr"]
+        retCode=ret[1]["returncode"]
+        assert(retCode == 2)
+        assert("database dirty flag set" in stderr)
+
+    testSuccessful=True
+finally:
```
brianjohnson5972: This may not be something to do as part of this, but can we make this finally block a common method that we can call for all the tests?

PR #3808, opened 2018-06-04
Use https to fetch deps, remove --no-check-certificate

komachi:
Currently active man-in-the-middle can do anything with docker container in build time because deps fetched with ` --no-check-certificate`. Also we load GPG key with http and don't check fingerprint.

This is not an ideal fix, we should also check signatures and fingerprints, but it's at least something.


stvenyin:
I.SEE

PR #3821, opened 2018-06-05
[don't merge yet] Fix #3789. Allow --filter-on option to take the form of contract:acti…

### LINKED ISSUE #3789: Limitation of account history for token receiver of eosio.token transfer action
The filter-on option takes the form of receiver:action:actor, where actor can be left blank. The receiver part can not be left blank. 

For the eosio.token contract, with any given receiver, if I want to be able to check all the token transferred  to that receiver from all different actors, I could not do that with the current filter-on mechanism. 

Does it make sense to allow the receiver in the filter-on option to be blank as well? 

Note that a transfer action will trigger three action trace with eosio.token, user, tester as the receiver respectively.

```
easypayx@easypayxs-MacBook-Pro:~/eosio-commands $ cleos push action eosio.token transfer '[ "user", "tester", "25.0000 EOS", "m" ]' -p user
executed transaction: 00a9a3e1e109682496fe1c3730473f85be687cd2c362455c9ea72e2435993bbc  128 bytes  2177 us
#   eosio.token <= eosio.token::transfer        {"from":"user","to":"tester","quantity":"25.0000 EOS","memo":"m"}
#          user <= eosio.token::transfer        {"from":"user","to":"tester","quantity":"25.0000 EOS","memo":"m"}
#        tester <= eosio.token::transfer        {"from":"user","to":"tester","quantity":"25.0000 EOS","memo":"m"}
```

CharlesJu1:
--filter-on option can now take two forms.
1. receiver:action:actor where receiver and action may not be blank. actor can be blank. This is to maintain backwards compatibility.
2. contract:action:actor:receiver where contract and action may not be blank. Actor and receiver can be blank.

### DIFF
--- a/Docker/README.md
+++ b/Docker/README.md
@@ -81,7 +81,7 @@ cleos get account inita
 Upload sample exchange contract
 
 ```bash
-cleos set contract exchange contracts/exchange/exchange.wast contracts/exchange/exchange.abi
+cleos set contract exchange contracts/exchange/
 ```
 
 If you don't need keosd afterwards, you can stop the keosd service using


tbfleming:
Continuing conversation on #3789

PR #3832, opened 2018-06-05
Update eosio.token.hpp

filipniziol:
Move get_supply and get_balance methods to make them public


contracts/eosio.token/eosio.token.hpp
```diff
@@ -30,14 +30,16 @@ namespace eosio {
...
                         string       memo );
+      
+      
+         inline asset get_supply( symbol_name sym )const;
+         
+         inline asset get_balance( account_name owner, symbol_name sym )const;
 
       private:
 
          friend eosiosystem::system_contract;
```
arhag: Since you are making `get_supply` and `get_balance` public, please also get rid of the now unnecessary `friend eosiosystem::system_contract;`.

Issue #3849, opened 2018-06-05
cleos --print-request


moskvanaft:
It seems like some people are trying to use RPC interface and having problems.
It would be useful for them to being able to see what requests is cleos doing.

heifner:
Resolved by #3850

PR #3850, opened 2018-06-05
cleos --print-request #3849

moskvanaft:
See Issue #3849

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -123,12 +123,12 @@ namespace eosio { namespace client { namespace http {
    request_stream << "Accept: */*\r\n";
    request_stream << "Connection: close\r\n";
    request_stream << "\r\n";
-   request_stream << postjson;
    // append more customized headers
    std::vector<string>::iterator itr;
    for (itr = cp.headers.begin(); itr != cp.headers.end(); itr++) {
       request_stream << *itr << "\r\n";
    }
+   request_stream << postjson;
 
    if ( print_request ) {
       string s(request.size(), '\0');
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -104,8 +104,9 @@ namespace eosio { namespace client { namespace http {
                              const fc::variant& postdata,
                              bool print_request ) {
    std::string postjson;
-   if( !postdata.is_null() )
-      postjson = fc::json::to_string( postdata );
+   if( !postdata.is_null() ) {
+      postjson = print_request ? fc::json::to_pretty_string( postdata ) : fc::json::to_string( postdata );
+   }
 
    boost::asio::io_service io_service;
 
@@ -121,6 +122,8 @@ namespace eosio { namespace client { namespace http {
    request_stream << "content-length: " << postjson.size() << "\r\n";
    request_stream << "Accept: */*\r\n";
    request_stream << "Connection: close\r\n";
+   request_stream << "\r\n";
+   request_stream << postjson;
    // append more customized headers
    std::vector<string>::iterator itr;
    for (itr = cp.headers.begin(); itr != cp.headers.end(); itr++) {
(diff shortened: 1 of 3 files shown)


programs/cleos/httpc.cpp
```diff
@@ -128,6 +129,15 @@ namespace eosio { namespace client { namespace http {
    request_stream << "\r\n";
    request_stream << postjson;
 
+   if ( verbose ) {
+      string s(request.size(), '\0');
+      buffer_copy(boost::asio::buffer(s), request.data());
+      std::cerr << "REQUEST:" << std::endl
+                << "---------------------" << std::endl
+                << s << std::endl
```
heifner: We should print the `fc::json::to_pretty_string` instead of `fc::json::to_string`

programs/cleos/httpc.hpp
```diff
@@ -30,7 +30,7 @@ namespace eosio { namespace client { namespace http {
    parsed_url parse_url( const string& server_url );
 
    fc::variant do_http_call( const connection_param& cp,
-                             const fc::variant& postdata = fc::variant() );
+                             const fc::variant& postdata = fc::variant(), bool verbose = false );
```
heifner: verbose is too generic, I would use `print_request` here as well.

programs/cleos/main.cpp
```diff
@@ -1548,6 +1551,7 @@ int main( int argc, char** argv ) {
 
    bool verbose_errors = false;
    app.add_flag( "-v,--verbose", verbose_errors, localized("output verbose actions on error"));
+   app.add_flag("--print-request", print_request, localized("print HTTP request to STDERR"));
```
heifner: Why STDERR and not STDOUT?

programs/cleos/main.cpp
```diff
@@ -206,6 +207,8 @@ vector<chain::permission_level> get_account_permissions(const vector<string>& pe
    return accountPermissions;
 }
 
+static bool verbose = true;
```
heifner: Looks like this was added by mistake, remove.

programs/cleos/httpc.hpp
moskvanaft: No real reason for that. Just following a common practice - write real output to STDOUT and all diagnostic/debug message to STDERR. So, when you redirect output to file, you still see diagnostic messages in console and your file still contains valid output not mixed with the messages.

programs/cleos/main.cpp
moskvanaft: No real reason for that. Just following a common practice - write real output to STDOUT and all diagnostic/debug message to STDERR. So, when you redirect output to file, you still see diagnostic messages in console and your file still contains valid output not mixed with the messages.

heifner: ok

programs/cleos/httpc.cpp
```diff
@@ -125,9 +126,23 @@ namespace eosio { namespace client { namespace http {
...
+      std::cerr << "REQUEST:" << std::endl
+                << "---------------------" << std::endl
+                << s << std::endl;
+   }
+
    request_stream << "\r\n";
    request_stream << postjson;
 
+   if ( print_request ) {
+      std::cerr << fc::json::to_pretty_string( postdata ) << std::endl
```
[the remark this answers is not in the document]
moskvanaft: No. This if prints JSON. The if-block bellow prints HTTP-headers.
I checked, it looks as expected:
```
$ cleos --print-request get currency balance eosio.token eosio SYS
REQUEST:
---------------------
POST /v1/chain/get_currency_balance HTTP/1.0
Host: localhost
content-length: 55
Accept: */*
Connection: close

{
  "account": "eosio",
  "code": "eosio.token",
  "symbol": "SYS"
}
---------------------
500000000.0000 SYS
```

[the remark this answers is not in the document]
heifner: oh, right, I see you moved the `postjson`

PR #3893, opened 2018-06-07
mlock chainbase in memory #3891

moskvanaft:


spoonincode:
Can we ifndef win32 on this?

PR #3921, opened 2018-06-07
fixes to contract whitelist/blacklist

### LINKED ISSUE #3814: Special case onblock action from contract whitelist/blacklist to avoid misuse
As of v1.0.1, if a producer sets up their `contract-whitelist` to include accounts other than `eosio`, or alternatively has an empty `contract-whitelist` but includes `eosio` in the `contract-blacklist`, then that producer will reject the `onblock` transaction from executing. However, since we have no way of signaling whether the `onblock` transaction executed or not in the block structure, other nodes with a different whitelist/blacklist configuration may attempt to the run the `onblock` transaction. This leads to a discrepancy of the `action_mroot` and therefore the digest that the block producers signs compared to the digest other nodes expect the producer to have signed.

With the current behavior, it is inappropriate for a block producer to include `eosio` in the blacklist (or if a whitelist is set, to not include `eosio` in the whitelist). Either this needs to be made clear in the documentation, or better yet we should special case the `eosio::onblock` action so that it is not possible for a block producer to badly configure the whitelist/blacklist.

Thanks to EOS Canada for bringing this to our attention.

arhag:
Resolves #3814.

Whitelisting/blacklisting exceptions are subjective failures. They depend on the particular configuration of the block producer considering transactions for inclusion in a block. A subjective failure should never lead to a block being produced that relies on that subjective decision making (which is not made objective by including it in the block structure as an official part of the protocol, e.g. with the CPU billing amount) because the validators would not in general able to recreate that subjective decision making.

At the moment, there are a couple places where this rule is broken in the current code. 

First, since the onblock transaction is implicit which means it is not recorded as successful or not within the block, any subjective failure of the onblock action should not be allowed. However, the contract whitelist/blacklist was not respecting this rule. This means a producer who configured their whitelist/blacklists to effectively ban the `eosio` contract from executing would produce blocks that would be considered objectively invalid by all other validators. This was the issue described in #3814.

Second, the whitelist/blacklist exceptions were not treated as subjective failures in `controller`. This meant that whitelist/blacklist failures encountered while processing a deferred transaction would have incorrectly elevated the status of the retired transaction to `soft_fail` or `hard_fail`. A validator would have expected a different receipt status and again considered that block to be objectively invalid.

These bugs are fixed in this PR.

These bugs did not affect correct validation of a chain. All validators would still correctly accept or reject blocks according to the intended protocol on v1.0.1. The problem is with the block producers potentially producing objectively bad blocks due to these bugs. However, if block producers do not use the `contract-whitelist` and `contract-blacklist` options on v1.0.1, they are still safe from producing objectively bad blocks.

Since these bugs do not affect correct validation of a chain, the bug fixes in this PR (which do not modify validation logic†) are not consensus-breaking changes. 

With the changes in this PR (intended to be included in v1.0.2), the contract whitelist/blacklist features should be safe for block producers to use.

This PR also includes the following changes:
* Fail early on mismatching receipts with more meaningful error message (see footnote† below).
* Correct the C++ signature of `transaction::send` in eosiolib to accept a `uint128_t` sender ID rather than a `uint64_t` one. The signature of the C intrinsic (`send_deferred`) that `transaction::send` wraps was already correct, so once again this is not a consensus-breaking change.
* Added a helper script "scripts/abigen.sh" to make ABI generation from source code more convenient.
* Added two additional tests to whitelist_blacklist_tests related to the two bug fixes described earlier.
* Avoid producing the unnecessary additional block when deconstructing `validating_tester`. 

Footnotes:
† There is an immaterial change in the validation logic within this PR. The `transaction_receipt_header`s generated during the validation step are now compared against those in the signed block to ensure they are identical. This change is immaterial from the perspective of following the same protocol as before because any differences in the bits of these `transaction_receipt_header`s would have led to a very different block-signing digest used to recover the producer public key. So the block would still fail anyway due to a signature mismatch. The changes in this PR just make the validation fail earlier with a more meaningful error message.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -637,14 +637,14 @@ struct controller_impl {
          try {
             if( implicit ) {
                trx_context.init_for_implicit_trx();
-               trx_context.cannot_subjectively_fail = true;
+               trx_context.can_subjectively_fail = false;
             } else {
                trx_context.init_for_input_trx( trx->packed_trx.get_unprunable_size(),
                                                trx->packed_trx.get_prunable_size(),
                                                trx->trx.signatures.size() );
             }
 
-            if( !trx_context.cannot_subjectively_fail && pending->_block_status == controller::block_status::incomplete ) {
+            if( trx_context.can_subjectively_fail && pending->_block_status == controller::block_status::incomplete ) {
                check_actor_list( trx_context.bill_to_accounts ); // Assumes bill_to_accounts is the set of actors authorizing the transaction
             }
 
@@ -800,9 +800,9 @@ struct controller_impl {
                push_scheduled_transaction( receipt.trx.get<transaction_id_type>(), fc::time_point::maximum(), receipt.cpu_usage_us );
             }
             const transaction_receipt_header& r = pending->_pending_block_state->block->transactions.back();
-            FC_ASSERT( r == static_cast<const transaction_receipt_header&>(receipt),
-                       "receipt does not match",
-                       ("producer_receipt", receipt)("validator_receipt", pending->_pending_block_state->block->transactions.back()) );
+            EOS_ASSERT( r == static_cast<const transaction_receipt_header&>(receipt),
+                        block_validate_exception, "receipt does not match",
+                        ("producer_receipt", receipt)("validator_receipt", pending->_pending_block_state->block->transactions.back()) );
          }
 
          finalize_block();
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -492,10 +492,14 @@ struct controller_impl {
 
    bool failure_is_subjective( const fc::exception& e ) {
       auto code = e.code();
-      return (code == block_net_usage_exceeded::code_value) ||
-             (code == block_cpu_usage_exceeded::code_value) ||
-             (code == deadline_exception::code_value)       ||
-             (code == leeway_deadline_exception::code_value);
+      return    (code == block_net_usage_exceeded::code_value)
+             || (code == block_cpu_usage_exceeded::code_value)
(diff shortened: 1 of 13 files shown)


libraries/chain/controller.cpp
```diff
@@ -791,10 +796,13 @@ struct controller_impl {
                auto& pt = receipt.trx.get<packed_transaction>();
                auto mtrx = std::make_shared<transaction_metadata>(pt);
                push_transaction( mtrx, fc::time_point::maximum(), false, receipt.cpu_usage_us );
-            }
-            else if( receipt.trx.contains<transaction_id_type>() ) {
+            } else if( receipt.trx.contains<transaction_id_type>() ) {
                push_scheduled_transaction( receipt.trx.get<transaction_id_type>(), fc::time_point::maximum(), receipt.cpu_usage_us );
             }
+            const transaction_receipt_header& r = pending->_pending_block_state->block->transactions.back();
+            FC_ASSERT( r == static_cast<const transaction_receipt_header&>(receipt),
```
wanderingbort: Can this be made into an EOS_ASSERT with a proper type

libraries/chain/include/eosio/chain/transaction_context.hpp
```diff
@@ -75,6 +75,7 @@ namespace eosio { namespace chain {
          fc::microseconds              delay;
          bool                          is_input           = false;
          bool                          apply_context_free = true;
+         bool                          cannot_subjectively_fail = false;
```
wanderingbort: The double negative name is probably more confusing than an alternative like `can_subjectively_fail = true/false`.  I realize that will invert all the usages of it but there were enough "Not cannot" checks that I got concerned

PR #3935, opened 2018-06-07
Add action-blacklist

### LINKED ISSUE #3933: Add action_blacklist
Similar to `contract-blacklist` except it allows for fine-grained control over the actions to reject. It can also be used in conjunction with `contract-whitelist`/`contract-blacklist`

So for example if the `contract-whitelist` is set to `{"eosio", "eosio.msig", "eosio.token"}` and a problem is discovered with, for example, the `bidname` action of the system contract it shouldn't be necessary to block the entire system contract by removing `"eosio"` from `contract-whitelist`. Instead the producers could all also add `eosio::bidname` to `action-blacklist`.

arhag:
Resolves #3933.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -41,6 +41,7 @@ action_trace apply_context::exec_one()
       if( native ) {
          if( trx_context.can_subjectively_fail && control.is_producing_block() ) {
             control.check_contract_list( receiver );
+            control.check_action_list( act.account, act.name );
          }
          (*native)(*this);
       }
@@ -50,6 +51,7 @@ action_trace apply_context::exec_one()
       {
          if( trx_context.can_subjectively_fail && control.is_producing_block() ) {
             control.check_contract_list( receiver );
+            control.check_action_list( act.account, act.name );
          }
          try {
             control.get_wasm_interface().apply(a.code_version, a.code, *this);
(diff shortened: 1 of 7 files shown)


libraries/chain/apply_context.cpp
```diff
@@ -50,6 +51,7 @@ action_trace apply_context::exec_one()
       {
          if( trx_context.can_subjectively_fail && control.is_producing_block() ) {
             control.check_contract_list( receiver );
+            control.check_action_list( act.account, act.name );
```
wanderingbort: should this be `control.check_action_list( receiver, act.name )` ?

The receiver and act.account will only differ if the original action is processed and requires a different receiver.  This means the root of that action processing tree will be act.account/act.name.  If that was on the blacklist it would never get to a point where they differ.  However, by allowing it to be receiver you could avoid situations where the black list is subverted by "bouncing" it off an acceptable contract which is just a relay

arhag: If an acceptable contract action (e.g. `eosio.token::transfer`) is called a bad contract (for example the recipient `blackhat` of that transfer), then the `contract-blacklist` should be used to blacklist the `blackhat` account. The `action-blacklist` is for use when an otherwise good contract has a faulty action.

Also, if we want to prevent any `soft_fail`s from occurring, the producers would need to blacklist the `eosio::onerror` action. If we go with `receiver, act.name` rather than `act.account, act.name`, that would no longer work since the `eosio::onerror` action is first delivered immediately to the sending contract. It would not be feasible to add `receiver::onerror` to the action blacklist for every single account.

wanderingbort: ok, after discussion I agree this is fine.

PR #3949, opened 2018-06-08
Fix for large local sets in wasm

larryk85:


arhag:
CONSENSUS label was added because this change protects against transactions that would have stalled execution for too long, so if those transactions were included in a blockchain, nodeos with these changes would not be able to replay that blockchain.

PR #3951, opened 2018-06-08
resource manager audit

wanderingbort:
- doing a ratio<> multiply now checks for overflow conditions.  This is a safe assert to add to any chain with default parameters because:
    - ratio is used for elastic limits, the largest current use being max-block-cpu * max-virtual-cpu-mult * maximum-ratio aka 200,000 * 1000 * 1000 which is well under the 64bit limit
    - ratio is used to maintain an exponential moving average of usage, the largest current use being maximum-single-user-value * precision * window-size aka 200,000 * 1,000,000 * 172,800 which needs 55bits and therfore would not hit this assert
 - doing cast from uint128_t for math to uint64_t for results now asserts that the values are representable, for the same reasons as ratio<> this is safe with the defaults but added to protect against future deployments with significantly higher values
 - fixed a bug in the estimator of available cpu/net usage where our fixed point math ceiling divide was not taken into account.  This would lead to situations where our cpu timers were a few microseconds off but because the final accounting would catch anything there was no ability for this to affect concensus.  It is now just better at setting speculative limits

Also removed repetition between get_account* and get_account*_ex

### DIFF
--- a/libraries/chain/include/eosio/chain/resource_limits_private.hpp
+++ b/libraries/chain/include/eosio/chain/resource_limits_private.hpp
@@ -16,12 +16,21 @@ namespace eosio { namespace chain { namespace resource_limits {
 
       template<typename T>
       T operator* (T value, const ratio<T>& r) {
+         EOS_ASSERT(std::numeric_limits<T>::max() / r.numerator >= value, rate_limiting_state_inconsistent, "Usage exceeds maximum value representable after extending for precision");
          return (value * r.numerator) / r.denominator;
       }
 
-      constexpr uint64_t integer_divide_ceil(uint64_t num, uint64_t den ) {
+      template<typename UnsignedIntType>
+      constexpr UnsignedIntType integer_divide_ceil(UnsignedIntType num, UnsignedIntType den ) {
          return (num / den) + ((num % den) > 0 ? 1 : 0);
       }
+
+      template<typename LesserIntType, typename GreaterIntType>
+      constexpr LesserIntType downgrade_cast(GreaterIntType val) {
+         const auto max = std::numeric_limits<LesserIntType>::max();
+         EOS_ASSERT( val <= max, rate_limiting_state_inconsistent, "Casting a higher bit integer value ${v} to a lower bit integer value which cannot contain the value, max is ${m}", ("v", val)("m", max) );
+         return LesserIntType(val);
+      };
 
       /**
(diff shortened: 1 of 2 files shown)


libraries/chain/include/eosio/chain/resource_limits_private.hpp
```diff
@@ -16,13 +16,43 @@ namespace eosio { namespace chain { namespace resource_limits {
...
          return (num / den) + ((num % den) > 0 ? 1 : 0);
       }
 
+      /**
+       * Specialization for Signedness matching integer types
+       */
+      template<typename LesserIntType, typename GreaterIntType>
+      constexpr auto downgrade_cast(GreaterIntType val) ->
+         std::enable_if_t<std::is_signed<LesserIntType>::value == std::is_signed<GreaterIntType>::value, LesserIntType>
+      {
```
arhag: It would be even nicer if all of these templated functions had static asserts to check that the template type is an integer type.

libraries/chain/resource_limits.cpp
```diff
@@ -344,59 +349,6 @@ int64_t resource_limits_manager::get_account_cpu_limit( const account_name& name
    int64_t cpu_weight;
    get_account_limits( name, unused, unused, cpu_weight );
```
arhag: I would prefer if you used two different variables for the unused arguments (similar to the `x` and `y` used in `get_account_net_limit_ex`), so that we don't get confusing by spooky aliasing behavior while debugging `get_account_limits` in the future.

libraries/chain/include/eosio/chain/resource_limits_private.hpp
```diff
@@ -16,13 +16,43 @@ namespace eosio { namespace chain { namespace resource_limits {
 
       template<typename T>
       T operator* (T value, const ratio<T>& r) {
+         EOS_ASSERT(std::numeric_limits<T>::max() / r.numerator >= value, rate_limiting_state_inconsistent, "Usage exceeds maximum value representable after extending for precision");
```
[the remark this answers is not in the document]
wanderingbort: in current practice I dont think its ever zero but this is a fine additional check.

arhag:
We have determined that this change can actually prevent v1.0.2 from replay a blockchain produced with v1.0.1.

This was verified by replaying the Jungle testnet blockchain and observing a discrepancy at block 388461.

PR #3963, opened 2018-06-08
remove dir before cloning (if mongo build fails, dir will exists and clone...

barrage-ivan:
…clone would fail)


heifner:
Looks like we lost track of this PR. We have since upgraded the mongo driver. If this is still an issue can you submit a new PR against the release/1.1 branch.

PR #3981, opened 2018-06-09
finish the delegated_bandwidth->self_delegated_bandwidth

EOSBIXIN:
to fix the bug to get delegated account info:
```
cleos get account voter1111111
2735496ms thread-0   main.cpp:2659                 main                 ] Failed with error: Assert Exception (10)
itr != structs.end(): Unknown struct self_delegated_bandwidth
```


lhzw:
Good.

heifner:
The error should not have been with the name of the struct in the contract. The error is caused by a mismatch between versions of `cleos` and `nodeos`. This should no longer be an issue for you.

Also any changes to system contracts should now be submitted to the `eosio.contracts` repo without any Ricardian contracts.

PR #3987, opened 2018-06-09
update binaryen to include recursion limit fix

bytemaster:


spoonincode:
this sort of change ought to go in the common validation code. That way there is no divergence between runtimes (current and future) and it won't get lost when trying to upgrade binaryen or such.

wanderingbort:
Was this resolved by #4036 ?

Issue #1950, opened 2018-03-29
Document our "varint" encoding.


wanderingbort:
Previously, we exposed `signed_int` and `unsigned_int` as ABI types which were variable length encoded integers.   We are re-exposing these as part of #353 under the more conventional/descriptive names: `varint32` and `varuint32`.

However, the encoding of these is poorly documented. 

`varint32` is signed and uses [Zig-Zag encoding](https://developers.google.com/protocol-buffers/docs/encoding#signed-integers)

`varuint32` is unsigned and uses [VLQ or Base-128 encoding](https://en.wikipedia.org/wiki/Variable-length_quantity)

SheldonHH:
VLQ is a universal code that uses an arbitrary number of binary octets (8-bits bytes) to represent an arbitrarily large integer.

PR #3991, opened 2018-06-09
Doxygen Annotations for contracts/eosiolib interfaces

dskvr:
Documents `eosiolib` for better annotated documentation.


contracts/eosiolib/account.h
```diff
@@ -0,0 +1,34 @@
...
+    *  balance b;
+    *  b.account = N(myaccount);
+    *  balance(b, sizeof(balance));
+    *  @endcode
+    *
+    *  @deprecated This is deprecated in Dawn 3.0
+    */
+
+   bool account_balance_get( void* balance, uint32_t len );
+}
```
heifner: No longer supported, should be removed.

contracts/eosiolib/account.hpp
```diff
@@ -0,0 +1,82 @@
...
+ *  @return true if account's balance is found
+ * 
+ *  @deprecated This is deprecated in Dawn 3.0
+ */
+bool get(account_balance& acnt)
+{
+   return account_balance_get(&acnt, sizeof(account_balance));
+}
+
+} }
```
heifner: No longer supported, should be removed.

dskvr: @heifner removed.

contracts/eosiolib/account.h
dskvr: @heifner  removed.

contracts/eosiolib/privileged.h
```diff
@@ -29,19 +37,56 @@ extern "C" {
     */
    int64_t set_proposed_producers( char *producer_data, uint32_t producer_data_size );
 
+   /**
+    * @brief Set new active producers
+    * Set new active producers. Producers will only be activated once the block which starts the next round is irrreversible
```
wanderingbort: The conditions for transition are out of date, they are now:
calling this method "proposes" a schedule change
once the block that contains the proposal becomes irreversible, the schedule is promoted to "pending" automatically.  Once the block that promotes the schedule is irreversible, the schedule will become "active" 

It is no longer based on rounds etc.

contracts/eosiolib/transaction.h
```diff
@@ -37,67 +26,114 @@ extern "C" {
...
-    */
+    /**
+     *  Sends a deferred transaction.
+     *
+     *  @brief Sends a deferred transaction.
+     *  @param sender_id - ID of sender
+     *  @param payer - Account paying for RAM
+     *  @param serialized_transaction - Pointer of serialized transaction to be deferred
+     *  @param size - Size to reserve
+     */
```
wanderingbort: missing documentation for `replace_existing` if this is `0` then if the provided `sender_id` is already in use by an in-flight transaction from this contract, this will turn into a failing assert.  if it is `1` then this transaction will atomically _replace_ the inflight transaction (effectively cancelling it in the process)

contracts/eosiolib/varint.hpp
```diff
@@ -6,42 +6,207 @@
...
+ * @brief Defines variable length integer type which provides more efficient serialization
  * @ingroup types
- * @{
+ * @{/
  */
 
+/**
+ * Variable Length Unsigned Integer. This provides more efficient serialization of 32-bit unsigned int.
+ * It serialuzes a 32-bit unsigned integer in as few bytes as possible
+ *
```
wanderingbort: This is probably an appropriate place to define the encoding.

refer to https://github.com/EOSIO/eos/issues/1950

not sure why that issue was closed, maybe it was moved to another repo

Issue #3985, opened 2018-06-09
whitelist_blacklist_tests.cpp:10:10: fatal error: 'deferred_test/deferred_test.wast.hpp' file not found


ljbb:
There is an issue when building docker image, the detail information as below, could EOS support fix this issue? seems like the unit test cause the fatal error.

[1329/1345] Building CXX object unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o
FAILED: unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o 
/usr/bin/clang++   -Icontracts -I/eos/contracts -Iunittests/contracts -I/eos/unittests/contracts -I/eos/libraries/testing/include -Iunittests/tests/contracts -I/eos/unittests/wasm_tests -Iunittests/include -I/eos/libraries/chain/include -Ilibraries/chain/include -I/eos/libraries/chain/../wasm-jit/Include -I/eos/libraries/chain/../../externals/binaryen/src -I/eos/libraries/utilities/include -I/eos/libraries/utilities/../wasm-jit/Include -I/eos/libraries/fc/include -I/usr/local/include -I/usr/include/openssl/include -I/eos/libraries/fc/vendor/websocketpp -I/eos/libraries/chainbase/include -I/eos/libraries/wasm-jit/Source/Runtime/../../../chain/include -I/eos/libraries/softfloat/source/include -I/eos/libraries/softfloat/source/8086-SSE -I/eos/libraries/softfloat/build/Linux-x86_64-GCC -I/eos/libraries/builtins -I/eos/libraries/builtins../softfloat/source/include -Ilibraries/testing/include -I/eos/libraries/testing/../wasm-jit/Include -I/eos/libraries/abi_generator/include -I/usr/lib/llvm-4.0/include -Wall -Wno-invalid-partial-specialization -fcolor-diagnostics -O3 -DNDEBUG   -std=gnu++14 -MD -MT unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o -MF unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o.d -o unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o -c /eos/unittests/whitelist_blacklist_tests.cpp
/eos/unittests/whitelist_blacklist_tests.cpp:10:10: fatal error: 'deferred_test/deferred_test.wast.hpp' file not found
[...]

Issue #3969, opened 2018-06-09
Build error "'deferred_test/deferred_test.wast.hpp' file not found"


spartucus:
## Env:
OS: Ubuntu 16.04 x64
commit id: 26ec83de395ec9bd751bcd4308c751c441afa56a

## Reproduce:
`git pull`
`git submodule init`
`git submodule update`
`./eosio_build.sh`

```
[ 98%] Building CXX object unittests/CMakeFiles/unit_test.dir/database_tests.cpp.o
[ 98%] Building CXX object unittests/CMakeFiles/unit_test.dir/identity_tests.cpp.o
[ 98%] Building CXX object unittests/CMakeFiles/unit_test.dir/bootseq_tests.cpp.o
[ 98%] Building CXX object unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o
/home/niubob/eos/unittests/whitelist_blacklist_tests.cpp:10:10: fatal error: 'deferred_test/deferred_test.wast.hpp' file not found
#include <deferred_test/deferred_test.wast.hpp>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
unittests/CMakeFiles/unit_test.dir/build.make:542: recipe for target 'unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o' failed
make[2]: *** [unittests/CMakeFiles/unit_test.dir/whitelist_blacklist_tests.cpp.o] Error 1
CMakeFiles/Makefile2:6385: recipe for target 'unittests/CMakeFiles/unit_test.dir/all' failed
make[1]: *** [unittests/CMakeFiles/unit_test.dir/all] Error 2
Makefile:160: recipe for target 'all' failed
make: *** [all] Error 2
	>>>>>>>>>>>>>>>>>>>> MAKE building EOSIO has exited with the above error.

niubob@niubob-VirtualBox:~/eos$ 
```
Maybe should do a clean build?
`rm -rf build/`
`git pull`
`git submodule init`
`git submodule update`
`./eosio_build.sh`
```
[ 97%] Building CXX object unittests/CMakeFiles/unit_test.dir/producer_schedule_tests.cpp.o
[ 98%] Building CXX object unittests/CMakeFiles/unit_test.dir/resource_limits_test.cpp.o
[ 98%] Building CXX object unittests/CMakeFiles/unit_test.dir/block_timestamp_tests.cpp.o
[...]

PR #3992, opened 2018-06-09
Fix CMake include directories of unit_test build target

arhag:
I think these changes to CMakeLists.txt should resolve the build issues users are reporting in #3968, #3969, and #3985.

Although it is hard to say for sure since I couldn't replicate the build issues in the first place on my machines or on Buildkite.


samspk:
How to apply these changes from branch? Like this?

rm -rf build/
git pull
git submodule init
git submodule update
./eosio_build.sh

PR #4005, opened 2018-06-10
Update Docker Readme to match port

sgerov:
The actual wallet port exposed by `docker-compose` for `keosd` is 8900. If this alias is not set correctly `cleos` will spin up `keosd` each time and wallet won't be persisted.


gleehokie:
Thanks for your suggestion, this duplicates #3865 which has been merged now.

PR #4035, opened 2018-06-11
Better handling for non-local keosd in cleos

### LINKED ISSUE #3878: cleos --wallet-url does not work when keosd is hosted on another machine
It seems that current logic is hard coded to check that keosd is running on the local machine even if --wallet-url specifies a URL to a remote machine.

So something like this will fail
```
cleos --wallet-url http://{walletserver}:5555 wallet create
```
The below screenshot shows where it's hard coded.

wanderingbort:
- make an explicit type for resolved addresses
 - deduplicate logic to wait for a newly spawned keosd to be available
 - better checking for loopback addresses
 - proper support for hostnames that resolve to loopback addresses

resolves #3878

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -25,11 +25,30 @@
 
 using boost::asio::ip::tcp;
 namespace eosio { namespace client { namespace http {
-   void do_connect(tcp::socket& sock, const std::string& server, const std::string& port) {
+
+   namespace detail {
+      class http_context_impl {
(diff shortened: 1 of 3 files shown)


spoonincode:
I built this branch but now when I do
```
$ ./programs/cleos/cleos wallet create
Failed to connect to keosd at http://localhost:8900/; is keosd running?
$ 
```
🤔 why is it not launching?

I went back and tried some command on an older master and it did launch

PR #4036, opened 2018-06-11
Limit nested structures in wasm

larryk85:
This PR limits the nesting of block type structures (loop, block, if) to a depth of 1024.


abourget:
Can this fork if not applied by everyone? When applied locally with the 1024 limit, could it refuse blocks?

arhag:
@abourget:

No, the check only applies when producing a block not validating. So if a producer includes a transaction that violates this subjective nesting limit, the other nodes will be forced to accept it.

PR #4053, opened 2018-06-12
support secondary key in get table without the need of modifying abi file

taokayan:
Command example:

Get all the bids sorted by highbid:
```
./cleos get table eosio eosio namebids -k highbid -t i64
{
  "rows": [{
      "newname": "zoo",
      "high_bidder": "a123",
      "high_bid": 9990000,
      "last_bid_time": "1528797387500000"
    },{
      "newname": "ddd",
      "high_bidder": "a123",
      "high_bid": 120000,
      "last_bid_time": "1528797387500000"
    },{
      "newname": "abc",
      "high_bidder": "a123",
      "high_bid": 110000,
      "last_bid_time": "1528797387500000"
    },{
      "newname": "com",
      "high_bidder": "a123",
      "high_bid": 100000,
      "last_bid_time": "1528797387000000"
    }
  ],
  "more": false
}
```

Get all the producers which have votes equal to zero:
```
./cleos get table eosio eosio producers -k prototalvote -t float64 -L "0" -U "0.00001"
{
  "rows": [{
      "owner": "a123",
      "total_votes": "0.00000000000000000",
      "producer_key": "EOS8kE63z4NcZatvVWY4jxYdtLg6UEA123raMGwS6QDKwpQ69eGcP",
      "is_active": 1,
      "url": "",
      "unpaid_blocks": 0,
      "last_claim_time": 0,
      "location": 0
    }
  ],
  "more": false
}
```


programs/cleos/main.cpp
```diff
@@ -1735,6 +1737,8 @@ int main( int argc, char** argv ) {
    getTable->add_option( "-k,--key", table_key, localized("The name of the key to index by as defined by the abi, defaults to primary key") );
    getTable->add_option( "-L,--lower", lower, localized("JSON representation of lower bound value of key, defaults to first") );
    getTable->add_option( "-U,--upper", upper, localized("JSON representation of upper bound value value of key, defaults to last") );
+   getTable->add_option( "-t,--seckeytype", sec_key_type, localized("Secondary key type(i64,i128,i256,float64,float128)"));
+   getTable->add_option( "-p,--secindexposition", sec_index_position, localized("Secondary index position, default to 0(first secondary index)"));
```
heifner: Since we will support diff primary key types in the future, the abbreviated `-t -p` are too generic. Also too much implementation detail here. Seems odd to talk about first, second, third, etc secondary index. From the users perspective it should just be first (primary), second, third, etc.

How about:
```
"--index", "Index number, 1 - primary (first), 2 - secondary index (in order defined by mult_index), 3 - third index, etc."
"--key-type", "The key type of --index, primary currently only supports (i64), all others support (i64, i128, i256, hex256, float64, float128)"
```

heifner: "-k,--key" should either be removed or made to work. Currently, it is not used for anything.

heifner: @wanderingbort suggested also supporting "primary", "secondary", etc as args to --index so the user does not have to use literal numbers.

FortisFortuna:
This is direly needed

tbfleming:
This feels like a workaround for a mistake: the ABI serializer cleanly distinguishes among types, except for the indexes. Maybe we should revisit that.

"senary"? "denary"? Are these Portuguese words? Italian? Could we abandon using words as numbers?

heifner:
Might have gotten a little carried away. I'm ok with removing the word support or limiting it to primary, secondary, third.

wanderingbort:
@tbfleming I agree however, there will be some non-zero amount of time between pushing a new version of the ABI format which includes typed indices and when existing ABIs are migrated to it.  Even if this gets deprecated in the next release, we will have coverage for our users until then.   I do not think this is the last word on the usability of these indices.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -293,6 +295,116 @@ class read_only {
...
+            pos = 0;
+         } else {
+            pos -= 2;
+         }
+      }
+      index |= (pos & 0x000000000000000FULL);
+      return index;
+   }
+
+   template <typename IndexType, typename Scope, typename SecKeyType, typename ConvFn>
```
wanderingbort: It looks like `Scope` is no longer really needed as a template parameter and a relic of the days when we had multiple intrusive indices.  All of the callers use `by_secondary` and it is only used in a way where `by_secondary` is always the right choice.  We should probably simplify it out.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -293,6 +295,116 @@ class read_only {
...
    }
 
+   static uint64_t get_table_index_name(const read_only::get_table_rows_params& p) {
+      // see multi_index packing of index name
+      const uint64_t table = p.table;
+      uint64_t index = table & 0xFFFFFFFFFFFFFFF0ULL;
+      EOS_ASSERT( index == table, chain::contract_table_query_exception, "Unsupported table name: ${n}", ("n", p.table) );
+
+      uint64_t pos = 0;
+      if (p.index_position.empty() || p.index_position == "first" || p.index_position == "primary") {
```
wanderingbort: This is only called when we are explicitly not talking about the primary key.  However, the naming of the method and the operation would imply that there is no difference between the primary and secondary indices.  (They both return 0 for the index byte).  

Can we rename this such that it is obvious that this only works on the non-primary indices and let "first", "primary" etc fall through to the `throw` when it is misused?

shuaiLiWang:
not support search by string？

taokayan:
you can use ``` --key-type name``` if you want to search by name in string

shuaiLiWang:
this is my struct , if i want search by businessName，how to get it

shuaiLiWang:
this is my struct , if i want search by name or phone，how to get it

struct chaninfo {
         uint64_t channelId;
         string name;
         string phone;
         
       }
       uint64_t primary_key() const { return channelId; }
    
      EOSLIB_SERIALIZE(chaninfo, (channelId)(name)(phone); };

taokayan:
We don't support std::string as key of any indexes. However you can find a way to encode it to 64/128/256 bit integers.

shuaiLiWang:
image

image

i want to encode string into 64 bit integers by boost，but i get the error when i cleos set contract，so i change the contract，use stringstream and sprintf/sscanf ，also have the same problem，why?

FortisFortuna:
Does the abi have to be updated as well in the `key_names` and `key_types` fields?

PR #4063, opened 2018-06-12
Add unstaking info to get account of cleos and chain_plugin

### LINKED ISSUE #4059: add unstaking info to get account

arhag:
Resolves #4059.

Added `refund_request` to `get_account_results` of chain_plugin which is populated with the data of the corresponding table row from the system contract for the requested user if such row exists.

Cleos will now use the information in  `refund_request` of  `get_account_results` (if it exists) to display unstaking information for the `cleos get account` command.

Also, changed the display of various units for bytes (`Kb` -> `KiB`, `Mb` -> `MiB`, etc.) to clearly reflect that the values are multiples of 1024 rather than multiples of 1000.

Also, `head_block_time` returned from get info is now a `time_point` rather than `time_point_sec` (displays the millisecond resolution necessary with sub-second block times).

### DIFF
--- a/libraries/chain/include/eosio/chain/resource_limits_private.hpp
+++ b/libraries/chain/include/eosio/chain/resource_limits_private.hpp
@@ -54,6 +54,9 @@ namespace eosio { namespace chain { namespace resource_limits {
       {
          const GreaterIntType max = std::numeric_limits<LesserIntType>::max();
          const GreaterIntType min = 0;
+         if( !(val >= min && val <= max) ) {
+            wlog("downgrade_cast higher to lower");
+         }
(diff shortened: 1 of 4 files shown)


arhag:
v1.0.3 cleos should still be able to work as it used to with a chain_plugin version running with these changes. (It will however continue to truncate the millisecond precision of the actual head block time before displaying the results to the console.)

A cleos version with these changes should still be able to work (granted without displaying the unstaking info) with a v1.0.3 chain_api_plugin. (Furthermore, it may inaccurately display the head block time as 500 milliseconds older than the true head block time since older versions of chain_plugin would not provide the head block time in the JSON response at the necessary level of precision.)

However, these changes do technically change the JSON structure of `get_account_results` and `get_info_results`. In particular, a new `fc::variant` field called `refund_request` was added into `get_account_results` and the type of `head_block_time` in `get_info_results` was changed from `fc::time_point_sec` to `fc::time_point` (in the returned JSON, this just means that string representation of the time will include the millisecond precision now).

PR #4065, opened 2018-06-12
bnet node to follow irreversible block instead of head

moskvanaft:


plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -482,6 +485,12 @@ namespace eosio {
               idx.erase(itr);
               itr = idx.begin();
            }
+
+           auto bitr = _block_status.find(s->id);
```
wanderingbort: can we put this inside the branch on `_remote_request_irreversible_only` to avoid the cost of the lookup for all other types of connections?

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -117,16 +119,16 @@ using std::vector;
 struct hello {
    public_key_type               peer_id;
    string                        network_version;
-   string                        agent;
-   string                        protocol_version = "1.0.0";
+   string                        meta;
```
wanderingbort: This string-based extensions sets a bad precedent IMO.  I understand we are trying to rush a feature into an experimental plugin but we should probably address the underlying need for versioning in this protocol now instead of kicking the can down the road.  

I would suggest we deprecate `hello` by renaming it to `legacy_hello` and maintaining its location in the static_variant `hello` structure which is easier to version and support backwards/forwards compat on.

To this end, we should probably put statically sized version (whatever width int we think we need) as the first member so that peeking at a `hello` can tell you enough to know if you support the protocol version without unpacking the data.

We should also have a `vector<extensions>` as the last part of the basic hello, where an `extension` is a static_variant of 1 or more things.  We could move things like `pending_block_ids` and `user/password` into extensions so that the base `hello` is lean.  We can also take this chance to prune stuff that wasn't/isn't used

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -508,7 +517,7 @@ namespace eosio {
            if( fc::time_point::now() - s->block->timestamp  < fc::seconds(6) ) {
            //   ilog( "queue notice to peer that we have this block so hopefully they don't send it to us" );
               auto itr = _block_status.find(s->id);
-              if( itr == _block_status.end() || !itr->received_from_peer ) {
+              if( !_remote_request_irreversible_only && ( itr == _block_status.end() || !itr->received_from_peer ) ) {
```
wanderingbort: is this necessary?  is there a reason we wouldn't send block notices to another connection?  the semantics are "I know of this block, don't send it" and if the peer is getting reversible blocks from elsewhere and you don't offer notice then they may relay them to you even if they only want irreversible blocks from you

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -283,7 +283,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             _production_enabled = true;
 
 
-         if( fc::time_point::now() - block->timestamp < fc::seconds(5) || (block->block_num() % 1000 == 0) ) {
+         if( fc::time_point::now() - block->timestamp < fc::minutes(5) || (block->block_num() % 1000 == 0) ) {
```
wanderingbort: This doesn't really belong in this PR but if we are going to modify it we should probably make it configurable so that the _user_ can decide what is appropriate for logging.

moskvanaft: It belongs to this PR in a way that if you receive only irreversible blocks your head block time is always delayed for more then 5 seconds (in case of 21 block producers). So, it's a cheat suggested by Dan.

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -1474,9 +1497,28 @@ namespace eosio {
...
-         vector<string> params;
-         boost::split( params, hi.meta, boost::is_any_of("\n"));
-         _remote_request_irreversible_only = ( find( params.begin(), params.end(), "irreversible_only" ) != params.end() );
+         //optional extensions
+         while ( 0 < ds.remaining() ) {
+            unsigned_int size;
+            fc::raw::unpack( ds, size ); // next extension size
+            auto ex_start = ds.pos();
+            fc::datastream<const char*> dsw( ex_start, size );
+            unsigned_int wich;
```
wanderingbort: typo wich -> which

PR #4078, opened 2018-06-13
support for SO_RCVLOWAT

wanderingbort:
- add optional support for SO_RCVLOWAT on socket connections in an effort to minimize calls between kernel space and user space for large messages or low MTU links
  - this can be enabled in config or cli with the `use-socket-read-watermark=true` option

This is optional because platform support is something we need to investigate.  However, on platforms that support it properly it does seem to be a decent feature.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -194,6 +194,8 @@ namespace eosio {
 
       shared_ptr<tcp::resolver>     resolver;
 
+      bool                          use_socket_read_watermark = false;
+
       channels::transaction_ack::channel_type::handle  incoming_transaction_ack_subscription;
 
       void connect( connection_ptr c );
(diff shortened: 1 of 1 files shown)


spoonincode:
Does this even work on Linux? My recollection is that polling ignores SO_RCVLOWAT.

wanderingbort:
```
SO_RCVLOWAT
              is changeable only since Linux 2.4.  The select(2) and poll(2)
              system calls currently do not respect the SO_RCVLOWAT setting
              on Linux, and mark a socket readable when even a single byte
              of data is available.  A subsequent read from the socket will
              block until SO_RCVLOWAT bytes are available.
```
its questionable whether `epoll` === `poll` in this regard.  But its probably safe to say that its not different. 

It certainly behaves correctly in OSX and it defaults to disabled

spoonincode:
Yeah, I suspect it does not work in Linux (though it might be worth trying/reading code just to be sure). Really worth adding more complexity here for something that doesn't help for pretty much any real world use where this would be a problem?

PR #4104, opened 2018-06-14
Update mongo-c-driver 1.10.1 for MongoDB 3.3.0

eosbot:
MongoDB C++11 Driver 3.3.0 this version of mongocxx requires the MongoDB
C driver 1.10.1.


heifner:
Thanks for this, but we need to move to cmake build of mongo with the new version. @pacificcode is working on that now. Going to close this, merge #4100. And we should have a PR soon which will upgrade us back to stable mongo.

PR #4107, opened 2018-06-14
Remove default key from keosd

PaulCalabrese:
Removes the default key so the wallet plugin does not load it by default.
Add a remove_key interface to the wallet_api to allow users to remove keys from the wallet.
Add "wallet remove" command to cleos.
Add new unit tests to wallet unit tests for remove_key functionality.
Fix test breakage caused by removal of the default key.


plugins/wallet_plugin/include/eosio/wallet_plugin/wallet_manager.hpp
```diff
@@ -107,6 +107,13 @@ class wallet_manager {
    /// @throws fc::exception if wallet not found or locked.
    void import_key(const std::string& name, const std::string& wif_key);
 
+   /// Removes a key from the specified wallet.
+   /// Wallet must be opened and unlocked.
+   /// @param name the name of the wallet to import into.
```
heifner: ... to remove from

plugins/wallet_plugin/include/eosio/wallet_plugin/wallet_manager.hpp
```diff
@@ -107,6 +107,13 @@ class wallet_manager {
...
+   /// @param key the Public Key to remove, e.g. EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV
+   /// @throws fc::exception if wallet not found or locked or key is not imported.
+   void remove_key(const std::string& name, const std::string& key);
```
heifner: Needs to require password.

plugins/wallet_plugin/include/eosio/wallet_plugin/wallet_manager.hpp
```diff
@@ -126,7 +133,7 @@ class wallet_manager {
    std::chrono::seconds timeout = std::chrono::seconds::max(); ///< how long to wait before calling lock_all()
    mutable timepoint_t timeout_time = timepoint_t::max(); ///< when to call lock_all()
    boost::filesystem::path dir = ".";
-   std::string eosio_key = "5KQwrPbwdL6PhXujxW37FSSQZ1JiwsST4cqQzDeyXtP79zkvFD3";
+   std::string eosio_key;
```
heifner: This can be removed as not used now.

programs/cleos/main.cpp
```diff
@@ -2216,6 +2216,23 @@ int main( int argc, char** argv ) {
       std::cout << localized("imported private key for: ${pubkey}", ("pubkey", std::string(pubkey))) << std::endl;
    });
 
+   // remove keys from wallet
+   string wallet_rm_key_str;
+   auto removeKeyWallet = wallet->add_subcommand("remove", localized("Remove key from wallet"), false);
```
heifner: Lets name it "remove_key"

plugins/wallet_plugin/include/eosio/wallet_plugin/wallet_manager.hpp
PaulCalabrese: You can still pass it on the command line as an option.

heifner: Please remove that feature as well.

Issue #3965, opened 2018-06-08
Non-deterministic output from get table with large limits


igorls:
cleos get table outputs different number of rows when a large limit is applied, on that case the symptoms started around 1666 rows and at very large values it doesn't output the full data nor give any warning.

Strange behaviour:
![image](https://user-images.githubusercontent.com/4753812/41183412-ff8e5e86-6b50-11e8-8e6d-bcdd9e0130f6.png)

Correct behaviour:
![image](https://user-images.githubusercontent.com/4753812/41183589-d3cc479e-6b51-11e8-884c-728ed2ed2ca3.png)

cleos fetching a high row:
![image](https://user-images.githubusercontent.com/4753812/41183646-3a8337d6-6b52-11e8-8c90-bf6225b476a1.png)

cleos version: `1a27ee3a`
nodeos version: `438824506`
v1.0.1

aaroncox:
We are experiencing this as well, even when using limits of 300 or 400.

ScottSallinen:
At the end of the received json is the `"more":true` when not everything was retrieved. How to we enforce nodeos to get the whole table that we requested?

This is a big issue with listing producers, as it would only show ones high in alphabetical order.

perduta:
Tokenika has same problem. It makes hard to follow changes via API. :-1:

ScottSallinen:
https://github.com/EOSIO/eos/blob/c55061b52acda38b969c23e77ebc02d1d3de56be/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp#L312

This is the problem: there is an un-configurable 10ms timeout.

This should be configurable!

arhag:
If you increase the timeout too much you risk your API node being unresponsive. So there needs to be a sensible time limit. And yes perhaps that limit should be configurable in the config.ini.

But fundamentally, consumers of the RPC API will need to expect that some requests will take too long (or be too big) to fulfill all at once. This is the reason for the `more` boolean. The consumer can know that there are more rows in the table, so it knows to request more of them if it wants them all.

The way to do this with cleos is to use the `-L` option.

For example:
The initial request could be something like the following (but in practice with a larger limit than 5):
```
$ ./cleos get table -l 5 eosio.unregd eosio.unregd addresses 
{
  "rows": [{
      "id": 0,
      "ethereum_address": "0x89a61a6d6eb95f56da5bb2f5e879a569d85967ef",
      "balance": "1135.8899 EOS"
    },{
[...]

PR #4108, opened 2018-06-14
Raise chain_api get_table_rows_result time limit

perduta:
Hotfixes issue #3965. Careful, since this might increase DDoS possibilities.

Credits for @ScottSallinen who first commented on GH what might be the issue.


arhag:
> Careful, since this might increase DDoS possibilities.

Right. Which is why we don't want to merge this in.

I wrote a [comment](https://github.com/EOSIO/eos/issues/3965#issuecomment-397449923) addressing the actual problem discussed in issue #3965 and possible solutions.

Issue #4111, opened 2018-06-14
Invalid packed transaction pivot != std::string::npos:


LeifHorton:
I'm getting this error while deploying any contract on mainnet.
Invalid packed transaction pivot != std::string::npos: No delimiter in string, cannot determine type: EOSKasiRrNwYgrHjvYh9aBEhVMUfKi6cPcRs2eeUEwyA8T69nckD2d4knbhVL6A5LSiv9fKww72W9zTaALFeUNtosyyaDFBPg Failed to deserialize variant

PR #4112, opened 2018-06-14
Pretty print transactions

heifner:
Enhanced the abi_serializer to be able to deserialize transactions even if they have invalid abi set on the contract. So even with invalid abi the transaction can still be pretty-printed. In the case of invalid abi the action data is left in hex since there is no way to deserialize it. This allows indexers like the mongodb plugin to serialize the parts of transaction that are serializable without the user provided abi.

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -174,10 +174,11 @@ namespace eosio { namespace chain {
       }
    }
 
-   bool abi_serializer::is_type(const type_name& rtype)const {
+   bool abi_serializer::_is_type(const type_name& rtype, size_t recursion_depth)const {
+      if( ++recursion_depth > max_recursion_depth) return false;
       auto type = fundamental_type(rtype);
       if( built_in_types.find(type) != built_in_types.end() ) return true;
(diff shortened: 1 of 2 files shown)


LeifHorton:
@heifner My issue is still there. [issue#4111](https://github.com/EOSIO/eos/issues/4111)
This is the error I'm getting:
Invalid packed transaction
pivot != std::string::npos: No delimiter in string, cannot determine type: EOSK59ovuuemzcGX1QKATF4pEVg2AmUT2rjQAH8MrC52SpmBZ59YPVYMFYTBg7XNgCDQcX18vMyDsjEp4SbyDnkcUvruZ94BX
Failed to deserialize variant

LeifHorton:
This is the command I'm using and getting error:
cleos -H node1.eosphere.io --wallet-host localhost --wallet-port 8888 set contract eosuser build/contracts/eosio.token -p eosuser

Getting same error for every contract

cleos version client: 15953cc1

Please, help me to fix this.
Thanks

LeifHorton:
@heifner The problem is coming from here /eos/libraries/fc/src/crypto/signature.cpp:
 const auto pivot = base58str.find('_');
      FC_ASSERT(pivot != std::string::npos, "No delimiter in string, cannot determine type: ${str}", ("str", base58str));

Issue #3683, opened 2018-05-31
Producer configurable block lag compensation


wanderingbort:
# Description

Right now the "deadline" used to trigger sign/commit/send of a block on a producer is the block's timestamp.  However, in situations where the latency to the next producer/network is high OR if there is latency imposed by signature security (non local hardware module etc) it would be beneficial for the producers to set a value which will trigger sign/commit/send early.  This can compensate for signing and network lag when tolerances are tight. 

This should also be exposed through the API plugin so it can be adjusted at runtime.

taokayan:
https://github.com/EOSIO/eos/pull/4118

taokayan:
brief description:

This feature allows a producer to produce block sooner(negative number) or later(positive number), in terms of microseconds, in order to compensate the time different and network latencies between different producer nodes:

config/CLI parameters example:

```produce-time-offset-us = -5000``` This makes non-last block producing time to be scheduled 5ms earlier

```last-block-time-offset-us = -10000 ``` This makes last block producing time to be scheduled 10ms earlier.

to change at runtime, please use:
```
curl --data-binary '{"produce_time_offset_us":-3000}' http://localhost:8888/v1/producer/update_runtime_options
```
and
```
curl --data-binary '{"last_block_time_offset_us":-3000}' http://localhost:8888/v1/producer/update_runtime_options
```

abourget:
I'm not sure I immediately get how we'd use that.
Would it be in reaction to network lag between our slot and the previous / next producers? (as in `Received block...latency=560ms`) so we'd adjust our offset to fit within our peers' lag window ? This changes quite dynamically though... and I've seen some producers drift linearly (lack of NTP sync)..
This seems a little wonky.. but I'm sure there's a better explanation :)

wanderingbort:
If you find that your blocks are often microforked out of the chain because they arrive "late" from the point-of-view to the next producer, this allows you to ship some of the blocks off on the network early to try to compensate.  The idea is that a partial block in the chain is better than block out of tha chain.

Issue #3043, opened 2018-05-14
Producer Policy Suggestion: Send last block early


wanderingbort:
# Description

In globally diverse test networks, the hand-off between two producers has the greatest risk for benign forking due to latency.  The result is that there can be a lot of single block forks when producer N + 1  must choose whether to wait for Producer N's last block or build from their second to last block.

# Proposal

A producer who maintains the chains operation as their best interest should voluntarily sign and send out their last block _early_ to allow it to reach the next producer in time for them to produce the first block of their round.

This policy should be implemented by the producer plugin and driven by command line

taokayan:
Is that similar feature as #3683 ?

taokayan:
hi, would you review https://github.com/EOSIO/eos/pull/4118?

taokayan:
I think this ticket can be closed as #4118 has been merged

PR #4118, opened 2018-06-15
Producerlag

taokayan:
#3683 add "produce_time_offset_us" parameter to set the producer time compensation in us.

for example, set producing time to be 3ms eariler:
```
curl --data-binary '{"produce_time_offset_us":-3000}' http://localhost:8888/v1/producer/update_runtime_options
```


taokayan:
Hi @wanderingbort , I've updated the pull request. Would you check if it's ok to merge?

taokayan:
#3043 add last-block-time-offset-us

PR #4121, opened 2018-06-15
Bios tutorial docker

eossey:
Added a bios boot sequence tutorial created with docker-compose.

- This tutorial allows an easier understanding of the bios sequence run in the Docker environment.
- Multiple servers are created to simulate inter-communication between different servers.
- After the bios-boot, the tutorial has created a running network managed by block producers. This allows other tests such as claiming BP rewards etc..


jgiszczak:
Please resubmit against branch `EOSIO:release/1.1`.

heifner:
Thanks for your PR. This is not something we are looking to maintain.

PR #4139, opened 2018-06-15
Remove wallet unlock unlock-timeout option

heifner:
cleos wallet unlock has a --unlock-timeout option but this option only does something if the wallet is not already running. If you do anything else first, like say cleos wallet open this is silently ignored.

Remove this option until it can be supported via rpc interface on keosd. This could possibly confuse users into thinking their wallet will be locked quickly when it will not be.


coolmagic:
Is there any way I can keep wallet open all the time? Running local node

jgiszczak:
@coolmagic You may specify a very long `unlock-timeout=` configuration option.  The type is `std::chrono::seconds` which according to the specification should be able to hold a duration up to 292 years.

PR #4153, opened 2018-06-16
fix fee calculation in sellram and comments

trumae:


gleehokie:
@trumae what are you aiming to accomplish with this PR?  Also, it does not build.

trumae:
@gleehokie Sorry by build problem. There are two things in this PR. 1) The substituted old comment are wrong and turn the code  more hard to understand.  2) The calculation of fee was changed in buyram and not changed in sellram (see line 111).

arhag:
Closing in favor of #4293 which includes your commits as well as the unit test fixes. Thanks.

shuke0327:
Hi, @arhag  one question regarding the sellram and buyram fee mechanism adjustment, when using:

`(quant + 199)/200` instead of `quant/200` to calc the ramfee, the small quant buy/sell will be with quite high ram fee rate, as following:

image

ingore the quant with 0.5 and 1, as the code limited the buy or sell quant should be higher than 1 EOS.

for new account creating, the ram fee will be quite high, and the 1 EOS threshold will prevent quite a lot wallet or service providers for helping creating new account.

for example, before, I can just create an new account with about 0.1 EOS, and the ram fee will just be 0.0005 EOS. 

but now, after 1.0.7 released, I have to pay at least 1.1 EOS, with the ram fee as high as 1.0, and the ram is just 0.1 EOS!

It is really hard to understand this design.

will you please explain why use this pricing model regarding ram trading?

shuke0327:
If I misunderstand this issue, please fix me. thanks a lot.

Damon1002:
The fee is not flexible yet. In my practical buy action, the fee is still 0.5%

PR #4158, opened 2018-06-16
only thow unexpected fails in apply_block

wanderingbort:


wanderingbort:
Thanks to @pmesnier for pointing out the problem and @bytemaster for collaborating on the fix

libraries/chain/controller.cpp
```diff
@@ -804,7 +804,9 @@ struct controller_impl {
                EOS_ASSERT( false, block_validate_exception, "encountered unexpected receipt type" );
             }
 
-            if( trace && trace->except ) {
+            bool transaction_failed =  trace && trace->except;
+            bool transaction_can_fail = receipt.status == transaction_receipt_header::hard_fail && receipt.trx.contains<transaction_id_type>();
+            if( transaction_failed && !transaction_can_fail) {
```
heifner: Let's add a method to transaction_receipt for determination of delayed/deferred transactions and unit tests for next release.

testcode77: If the number of delayed transactions increases, this case is meaningless. My opinion is that the handling of delayed transactions has to change.

reymondtu: Just  PR Three lines of code can fix bug?

eosbot: super :)

enolife:
Big thanks to @pmesnier for discovering this and the people who contributed in fixing this

pereztechseattle:
"If the number of delayed transactions increases, this case is meaningless. My opinion is that the handling of delayed transactions has to change."

Doesn't sound very "fixed" to me.  Temporarily worked around, maybe?

libraries/chain/controller.cpp
rikelme: This reads funny (very unintuitive):
transaction_failed && !transaction_can_fail

PR #4202, opened 2018-06-18
Revert speculative state with read_mode = head

moskvanaft:


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -341,6 +377,10 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   _persistent_transactions.insert(transaction_id_with_expiry{trx->id(), trx->expiration()});
                }
                send_response(trace);
+               if ( _read_mode != read_mode::SPECULATIVE ) {
+                  chain.abort_block();
```
wanderingbort: This isn't a really safe way to accomplish this.  The producer plugin has several invariants it needs to maintain about the state of the block.  At the very least this should call `schedule_production_loop` however, that will be heavyweight as it will re-apply all of the deferred transactions, unapplied transactions etc.

As it is, before a transaction hits a speculative block it will be in a state where the time point is managed and the deferred/unapplied/persisted transactions are applied and after the first tx, you will have a clean block with a potentially different timestamp.   This doesn't really present a consistent view of HEAD, nor would naively calling `schedule_production_loop` (which would apply some transactions).

If we want to solve this in the producer plugin, we need to have a first class mode of operation and analyze all of the invariants the producer plugin holds for effects.

PR #4218, opened 2018-06-19
Print sys info

cj-oci:
Print basic system info at start of test.


cj-oci:
PR created accidentally. Closing.

PR #4253, opened 2018-06-20
backup reversible blocks into portable format and import reversible blocks from the portable format

arhag:
Output reversible blocks in portable format anytime the reversible block database is recovered.

Add tool to chain_plugin to create a reversible blocks database from the portable reversible blocks file.

This allows users to share their reversible blocks in a portable format so that devs can do forensic analysis.


arhag:
The documentation below assumes that PR #4680 has also been merged in:

------------------------

# Exporting and Importing the  Reversible Block Database using a Portable Format

The reversible block database, which is normally located in the "blocks/reversible" directory within the blockchain data directory, contains all the blocks in the current chain that are not yet irreversible. This database is stored as a shared memory file similar to the chain state database, which means it is not generally portable across machines.

If a node operator wants to move the reversible block database from one machine to another, they normally must use the `--export-reversible-blocks` and `--import-reversible-blocks` options on nodeos to, respectively, extract the reversible block database from the source machine into a portable file format and then import the blocks from that portable file into a newly constructed reversible block database on the target machine.

This feature is also incredibly useful in providing EOSIO blockchain developers with the necessary data that they need (in addition to the blocks.log file) to assist in debugging a problem that may have occurred on a live EOSIO blockchain.

## Exporting the Reversible Block Database

The reversible block database can be exported into a portable file using the `--export-reversible-blocks` option. For example:
```
nodeos --export-reversible-blocks path_to_exported_reversible_blocks_file
```
will save the reversible blocks into the file `path_to_exported_reversible_blocks_file`.

### Automatic Exportation of Reversible Block Database on Recovery

Whenever the reversible block database has to be recovered, either explicitly due to the `--fix-reversible-blocks` option or perhaps implicitly during a `--hard-replay-blockchain`, nodeos will automatically export the recovered reversible blocks into portable file format which is stored as the file `portable-reversible-blocks-<timestamp>` under the `blocks` directory where `<timestamp>` is the time at which the recovery occurred. This automatically generated file should be identical to the file that would have been generated with the explicit `--export-reversible-blocks` option.

## Importing the Reversible Block Database

Given a file containing the reversible block database in the portable format, it is possible to use the `--import-reversible-blocks` option to import those blocks into the reversible block database that nodeos can use. Using this option will first clear out any existing reversible block database before importing the blocks from the specified file. Here is an example of how nodeos can be used to import the reversible blocks that were previously exported from another node:
```
nodeos --import-reversible-blocks path_to_exported_reversible_blocks_file
```

PR #4270, opened 2018-06-21
update mongo-c-driver & mongo-cxx-driver  version

williamleecn:
must above  mongo-c-driver 1.10


pacificcode:
@williamleecn Updating the MongoDB drivers was addressed in PR #4204 and merged into branch release/1.1

Issue #4196, opened 2018-06-18
Read modes : speculative, head, LIB


moskvanaft:

moskvanaft:
For now it doesn't work with bnet plugin (disallowed because of suspicion that they would not get along). It's going to be addressed in #4468

abourget:
What's the rationale behind these modes?

Is your Jira not sync'd with the public anymore? :)

wanderingbort:
# Nodeos `read-mode` Settings

EOSIO provides a set of services and interfaces that enable contract developers to persist state across action, and consequently transaction, boundaries. (https://developers.eos.io/eosio-cpp/docs/db-api)
Contracts may use it for different purposes. For example, `eosio.token` contract keeps balances of all users in the database.

Each instance of `nodeos` keeps the database in memory, so contracts can read and write data and
`nodeos` provides access to this data over HTTP RPC API for reading the database.

However, at any given time there can be multiple correct ways to query that data: 
- `speculative` : this includes the side effects of unconfirmed transactions.
- `head` : this only includes the side effects of transactions which have been included in the produced and signed blocks that are part of the best chain `nodeos`
- `irreversible` : this only includes side effects of transactions which have been confirmed by the network as irreversible.

## speculative read-mode

speculative

Clients (`cleos` and RPC API users) will see database state as of current head block plus changes made by all transactions known to this node but potentially not included in the chain (unconfirmed). This mode is low latency but fragile, since there is no guarantee that the transactions reflected in the state will be included in the chain OR that they will reflected in the same order the state implies.  

This mode is the lowest latency mode, but also, the lowest consistency.

In this mode `nodeos` is able to execute transactions which have TaPOS pointing to any valid block in a fork considered to be the best fork by this node.

## head read-mode

Clients (`cleos` and RPC API users) will see database state as of current head block of the chain.  Since current head block is not yet irreversible and short-lived forks are possible, state read in this mode may become inaccurate  if `nodeos` switches to a better fork.  

This mode represents a good trade between highly consistent views of the data and latency.

In this mode `nodeos` is able to execute transactions which have TaPOS pointing to any valid block in a fork considered to be the best fork by this node.

## irreversible read-mode

Clients (`cleos` and RPC API users) will see database state as of latest irreversible block (LIB).  This state is guaranteed to be accurate as of the given LIB.  

This mode has the strongest guarantee of consistency but the highest latency.

[...]

PR #4291, opened 2018-06-22
Revert speculative state #4196

moskvanaft:


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -144,6 +182,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "Action (in the form code::action) added to action blacklist (may specify multiple times)")
          ("key-blacklist", boost::program_options::value<vector<string>>()->composing()->multitoken(),
           "Public key added to blacklist of keys that should not be included in authorities (may specify multiple times)")
+         ("read-mode", boost::program_options::value<eosio::chain::db_read_mode>()->default_value(eosio::chain::db_read_mode::SPECULATIVE),
+          "Database read mode (\"speculative\" or \"head\")")
```
wanderingbort: We should probably be overly verbose about what this means in the description but that can be done in an additional PR

PR #4295, opened 2018-06-23
Fixes to system contract delegatebw action

### LINKED ISSUE #4273: EOS 抵押 漏洞
image
攻击的命令示例：
cleos --wallet-url http://localhost:6666 --url http://mainnet.genereos.io:80 system delegatebw sujianzhong1 etokengogogo "0.0001 EOS" "0.0000 EOS" --transfer

攻击示意图：
image

image

image

arhag:
Resolves #4273.

Changes to system contract in this PR:
 * `transfer` no longer allowed to be true if delegating to self.
 * delegating to another account with or without transfer will always pull tokens from the source's liquid balance (fixes bug that forces receiver to restake from their unstaking balance)


alepacheco:
When this change gets merged you will be forced to wait 3 days when changing the account you are delegating tokens to?

PR #4339, opened 2018-06-26
Consolidated Security Fixes for 1.0.7

wanderingbort:
* Add option --abi-serializer-max-time-ms
* Make max_serialization_time configurable
* fix replay bug: force auth checks even on replay when in the context of executing a deferred trx or implicit trx
* Check that fetch_block_by_id() returned something valid
* Added stricter local set size for functions
* Add additional validation to abi_serializer unpacking of arrays

Co-authored-by: arhag <arhag@users.noreply.github.com>
Co-authored-by: Kevin Heifner <heifnerk@objectcomputing.com>
Co-authored-by: Bucky Kittinger <larrykittinger@gmail.com>
Co-authored-by: Bucky Kittinger <larrykittinger@gmail.com>
Co-authored-by: Matt Witherspoon <32485495+spoonincode@users.noreply.github.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ tests/plugin_test
 unittests/unit_test
 
 doxygen
+eos.doxygen
 
 wallet.json
 witness_node_data_dir
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -16,6 +16,9 @@ using namespace boost;
 
 namespace eosio { namespace chain {
 
+   const size_t abi_serializer::max_recursion_depth;
+   fc::microseconds abi_serializer::max_serialization_time = fc::microseconds(15*1000); // 15 ms
+
    using boost::algorithm::ends_with;
    using std::string;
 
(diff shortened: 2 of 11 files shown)


abourget:
What do we get if it crosses the deadline, and what does this affect ? Calls to `/v1/history/get_transaction`, `/v1/chain/abi_to_json` (and its flipside) and the reverse, `/v1/chain/get_block` ? 
Will we have a half-rendered ABI-mapped JSON, or will `data` just be `null` ?

Reading the diff didn't immediately show those side effects, so wanted to check with you guys.

thanks!

abourget:
On all of these calls?

Is `push_transaction` also affected? In that case, you wouldn't even receive the `transaction_id` but the transaction would be finding its way on chain. I remember seeing that issue when ABI serialization failed earlier on..

Small things like that can make working with the chain more brittle.. I'd be interested in timings for ABI serialization.. so we can find the sweet spot in terms of timing config. Perhaps with debug-mode instrumentation..

heifner:
@abourget I think that is a very good point about `push_transaction`. This potentially aggravates the issue of a `push_transaction` failing on `abi_serializer` of `transaction_trace` and reporting an exception to the user when the transaction has actually found its way on chain.

I'll create an issue for this. Anything other than `push_transaction` do you see also having a similar problem?

abourget:
push_trandactions :-)

PR #4346, opened 2018-06-26
More precise exception for "Cannot create account named..."

noisy:


wanderingbort:
@noisy Looks like there are some unit tests tied to the exception previously thrown, care to fix them up so we can approve this PR?

PR #4362, opened 2018-06-27
vote function in the boot-sequence shell is duplicated

evsward:
There are two vote function and they are same, reduce one of them.


evsward:
Sorry, but I think it's a useful step for testing          the performance of blockchain.

evsward:
I have tested the shell modified and it worked well, it's very convenient.

wanderingbort:
@tbfleming is in the right here.  we would gladly merge the duplicate removal and my create an alternate PR to do so but, in general committing unrelated changes to the branch makes it hard to accept a PR.  In the future you can avoid this by making your changes in branches (instead of directly in your fork of `master`) so that they may be independently accepted.

evsward:
It's a pity but also thank you for your kindly remind. @wanderingbort

PR #4382, opened 2018-06-28
Remove magic number

spartucus:


libraries/chain/abi_serializer.cpp
```diff
@@ -160,18 +164,18 @@ namespace eosio { namespace chain {
...
+      return ends_with(string(type), optional_mark);
    }
 
    type_name abi_serializer::fundamental_type(const type_name& type)const {
       if( is_array(type) ) {
-         return type_name(string(type).substr(0, type.size()-2));
+         return type_name(string(type).substr(type.size()-array_mark.size()));
       } else if ( is_optional(type) ) {
-         return type_name(string(type).substr(0, type.size()-1));
+         return type_name(string(type).substr(type.size()-optional_mark.size()));
```
heifner: This is not equivalent to the old code and is causing tests to fail. The first arg to `substr` is start position, second arg is length.

spartucus: yes, it is the wrong way.

PR #4384, opened 2018-06-28
Using unordered_map for received_txs and received_blocks

EugeneChung:
I'm suggesting to use unordered_map for received_transactions and received_blocks of net_plugin.

received_transactions and received_blocks are std::vector and the method for finding their item is just O(N) iteration(for loop with the begin iterator), which I think it's unnecessary overhead. std::unordered_map can remove the overhead.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -699,18 +699,8 @@ namespace eosio {
       vector<block_request> req_blks;
       vector<transaction_id_type> req_trx;
 
-      struct block_origin {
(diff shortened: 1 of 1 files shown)


wanderingbort:
Thanks for the suggestion, we will evaluate it over what we know considering the additional costs of insertion and growth for such a datastructure and the average usage statistics we have available. 

I've rebased your change to 1.1 as we would not make such a change as a patch release.

wanderingbort:
Thank you for the submission, it has been supplanted by #5202 which also handled the case of receiving the block/transaction from multiple connections.

PR #4409, opened 2018-06-29
Secure entry for cleos wallet import

andriantolie:
- Add secure entry for `cleos wallet import` if `--private-key` option is not used
- Add some additional error message to make it clear when key already exists in the wallet

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -2204,13 +2204,20 @@ int main( int argc, char** argv ) {
    string wallet_key_str;
    auto importWallet = wallet->add_subcommand("import", localized("Import private key into wallet"), false);
    importWallet->add_option("-n,--name", wallet_name, localized("The name of the wallet to import key into"));
-   importWallet->add_option("key", wallet_key_str, localized("Private key in WIF format to import"))->required();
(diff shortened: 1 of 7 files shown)


programs/cleos/main.cpp
```diff
@@ -2204,13 +2204,20 @@ int main( int argc, char** argv ) {
    string wallet_key_str;
    auto importWallet = wallet->add_subcommand("import", localized("Import private key into wallet"), false);
    importWallet->add_option("-n,--name", wallet_name, localized("The name of the wallet to import key into"));
-   importWallet->add_option("key", wallet_key_str, localized("Private key in WIF format to import"))->required();
+   importWallet->add_option("--private-key", wallet_key_str, localized("Private key in WIF format to import"));
    importWallet->set_callback([&wallet_name, &wallet_key_str] {
+      if( wallet_key_str.size() == 0 ) {
+         std::cout << localized("password: ");
```
heifner: Should prompt for "private key" not "password".

heifner: Also the python test scripts need to be updated to use the --private-key option.

andriantolie: Thanks for catching that. PR updated.

heifner:
@wanderingbort Should we add this to 1.0.8?

Issue #4368, opened 2018-06-27
access to extended CPU/Net virtual resources should be subjective


wanderingbort:
## Problem

As currently designed, when the chains ephemeral resources are below a consensus managed threshold, the resource system will "relax" and overcommit the actual hard resource limit creating a virtual resource pool much larger.  This allows accounts with relatively small amounts of stake to still use an idle chain even if they would be heavily throttled if it was purely stake-weighted. 

However, this also means that a malicious actor can impose a higher load on an idle network than their stake should allow.  Congestion mechanics will prevent this from being a sustainable situation but, accounts which repeatedly abuse this feature should lose access to it.  

## Proposal

Allow producers to subjectively deny access to the virtual pool of resource and instead impose a strict interpretation of stake-weighted resource allotment.  Note, this would not prevent an account from transacting on chain nor would it impede the burst mechanics built into the system.  It would only deny access to the additional resources that result from an idle chain in a "relaxed" state.  

- [ ] Allow a mode of transaction processing where the resource calculations are done with hard chain maximums for CPU and Net and not the virtual maximums
  * failure in this mode should not be considered "objective" so, it should not result in any slashing conditions being imposed
  * failure in this mode should not be considered as a failure which invokes the error handler or hard_fail mechanics of scheduled and implied transactions
  * failure in this mode should return an immediate error to the user if their RPC node rejects a transaction on the basis of it being over the hard limit. 
  * failure in this mode *should* be considered normal failure if the chain's virtual limit === the real limit (aka the chain is in a fully congested state)
- [ ] Allow specification in config.ini and CLI of "greylisted" accounts, where greylisting indicates that the account is denied access to the extended virtual pool of resources

taokayan:
Hi, would you review my pull request https://github.com/EOSIO/eos/pull/4410?

taokayan:
Hi, would you review my pull request #4410?

taokayan:
brief of changes:

Greylisted accounts are not allowed to access extended(elastic) cpu/net limit in a subjective way.

When a block is being produced, if a transaction violates the cpu/net subjective limit, it will be rejected by the node immediately and immediate error will be returned. This includes normal transactions and deferred transactions.

block replay & apply_block will always use the objective cpu/net limit for validation.

related commands:

```cloes get account``` will return the subjective cpu/net limit.

to start nodeos with greylist parameters:
```
[...]

PR #4410, opened 2018-06-29
Subjective extended cpu/net resource access

taokayan:
#4368 

greylisted accounts are not allowed to access extended cpu/net limit in a subjective way.
"cloes get account" will return the subjective cpu/net limit.
if a transaction violates the cpu/net subject limit, it will be rejected by the node immediately.
block replay & apply_block will always use the objective cpu/net limit for validation.

command line options for producer (can appear multiple times):
--greylist-account "a11111111111"

new api urls:
/v1/producer/add_greylist_accounts
/v1/producer/get_greylist
/v1/producer/remove_greylist_accounts


taokayan:
command example:

to start nodeos with greylist parameters:
```
./programs/nodeos/nodeos --data-dir=/Users/jiaendu/workspaces/eos/build/data-dir --config-dir=/Users/jiaendu/workspaces/eos/build/data-dir --greylist-account "a11111111111"
```

add greylist accounts in real-time:
```
curl --data-binary '{"accounts":["aaa","bbb"]}' http://localhost:8888/v1/producer/add_greylist_accounts
```

get current grey listed accounts:
```
curl --data-binary '' http://localhost:8888/v1/producer/get_greylis
```

remove greylist accounts:
```
curl --data-binary '{"accounts":["aaa","bbb"]}' http://localhost:8888/v1/producer/remove_greylist_accounts
```

heifner:
Closing as an attempt to rebase.

libraries/chain/controller.cpp
```diff
@@ -654,13 +655,17 @@ struct controller_impl {
    transaction_trace_ptr push_transaction( const transaction_metadata_ptr& trx,
                                            fc::time_point deadline,
                                            bool implicit,
-                                           uint32_t billed_cpu_time_us  )
+                                           uint32_t billed_cpu_time_us,
+                                           bool subjective = true)
```
wanderingbort: it looks like the "subjective" flag is just meant to account for whether we are in a state where our own subjective opinions about transactions apply.   If you look at how the white/black lists are implemented you will see that we have chosen to signal this based on `controller::block_status` where `controller::block_status::incomplete` implies that we are speculating OR producing a block and all other statuses imply that we are not in a mode where we can apply subjective measures.  

You should see the pattern in other parts of the code which will keep this from needing to be threaded down as part of the functional parameters.

libraries/chain/resource_limits.cpp
```diff
@@ -412,5 +412,16 @@ account_resource_limit resource_limits_manager::get_account_net_limit_ex( const
    return arl;
 }
 
+bool resource_limits_manager::is_greylisted(const account_name &name) const {
```
wanderingbort: Our other black/whitelists are managed and maintained in the controller, this should probably go in the same place. 

Each `resource_limits_manager` method can take a single bool that indicates whether the operation in question should allow access or deny access to the elastic limits as opposed to now where it takes a flag which indicates the greylist may apply and then checks the greylist.   That will lower the burden for future devs who will find all of the black/grey/whitelists in the same system and reduce the amount of "application" logic in the `resouce_limits_manager`

libraries/chain/resource_limits.cpp
```diff
@@ -356,7 +356,7 @@ account_resource_limit resource_limits_manager::get_account_cpu_limit_ex( const
 
    uint128_t window_size = config.account_cpu_usage_average_window;
 
-   uint128_t virtual_cpu_capacity_in_window = (uint128_t)state.virtual_cpu_limit * window_size;
+   uint128_t virtual_cpu_capacity_in_window = (uint128_t)(elastic ? state.virtual_cpu_limit : config.cpu_limit_parameters.max) * window_size;
```
wanderingbort: Sorry I didnt catch this on first review but the math here is now out of sync with the assertions in `::add_transaction_usage` ideally those would be the same.  I know controller is externally maintaining in variants about billing but internal to the system that is the method that enforces limits.  `::get_account_*_limit` are supposed to be for informational purposes and not enforcement

wanderingbort: In case it was not clear, these methods are correct, but `:add_transaction_usage` is now out of date.

PR #4424, opened 2018-06-29
Read mode = irreversible

moskvanaft:


unittests/forked_tests.cpp
```diff
@@ -446,4 +438,26 @@ BOOST_AUTO_TEST_CASE(confirmation) try {
...
+
+   tester head(true, db_read_mode::HEAD);
+   push_blocks(c, head);
+   BOOST_REQUIRE_EQUAL(head_block_num, head.control->fork_db_head_block_num());
+   BOOST_REQUIRE_EQUAL(head_block_num, head.control->head_block_num());
+
+   tester irreversible(true, db_read_mode::IRREVERSIBLE);
+   push_blocks(c, irreversible);
+   BOOST_REQUIRE_EQUAL(head_block_num, irreversible.control->fork_db_head_block_num());
+   BOOST_REQUIRE_EQUAL(head_block_num - 49, irreversible.control->head_block_num());
```
arhag: Instead of hard coding the offset of 49, this test should get the LIB from `c` right after setting `head_block_num` above and compare that to `irreversible.control->head_block_num()`.

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -573,7 +573,7 @@ namespace eosio {
            _app_ios.post( [self = shared_from_this(),callback]{
               auto& control = app().get_plugin<chain_plugin>().chain();
               auto lib = control.last_irreversible_block_num();
-              auto head = control.head_block_id();
+              auto head = control.fork_db_head_block_id();
```
arhag: Has irreversible mode been tested with the bnet plugin? I am less confident from just code analysis that the `head_block_*` to `fork_db_head_block_*` change is sufficient for the bnet_plugin as it was with the net_plugin. The bnet_plugin seems to maintain its own local head block numbers based on the accepted_block channel.

PR #4429, opened 2018-06-30
Add eosio.sudo contract

### LINKED ISSUE #4142: Add eosio.sudo contract
This contract would be very similar to the eosio.msig contract except it is designed to allow 2/3+1 of the active block producers to execute any transaction while bypassing the typical authorization checks.

arhag:
Resolves #4142.

Adds a new contract (eosio.sudo) that is meant to be deployed to a privileged account `eosio.sudo`. This contract provides just one action `exec` which takes a transaction and schedules it as a deferred transaction while bypassing normal authorization checks (which it can do since it is on a privileged account). 

It does however check for authorization of itself (as well as authorization of the `executer`). The idea is that the `eosio.sudo` account should have its permissions set up like all the other `eosio.*` accounts: satisfied entirely by the `active` permission of the `eosio` account, which in turn is fully satisfied by the `active` permission of the `eosio.prods` account. Effectively, it means that only a greater than two-thirds supermajority of the active block producers are able to use the `eosio.sudo::exec` action to execute any transaction while bypassing all regular authorization checks.

This PR also includes tests related to the eosio.sudo contract, including testing the ability to propose a transaction that uses `eosio.sudo::exec` via the eosio.msig contract.

Also, @moskvanaft's changes to cleos to add the `cleos multisig propose_trx` account were added into this PR as well (those changes were originally meant for the v1.1 release) because it would be infeasible to deploy and use the eosio.sudo contract without it.


contracts/eosio.sudo/eosio.sudo.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+   account_name executer;
+
+   datastream<const char*> ds( buffer, size );
+   ds >> executer;
+
+   require_auth( executer );
+
+   size_t trx_pos = ds.tellp();
+   send_deferred( (uint128_t(executer) << 64) | current_time(), executer, buffer+trx_pos, size );
```
moskvanaft: Shouldn't it be `size-trx_pos` ?

arhag: It should be. Thanks.

In practice, `send_deferred` is pretty flexible about the size passed in as long as it isn't smaller than the actual size of the serialized transaction data, but indeed `size-trx_pos` is the correct expression for the argument to pass into this `send_deferred` call.

PR #4448, opened 2018-07-02
File location change

blockchain-expert:
eosio-abigen is changed from bin/eosio-abigen to programs/eosio-abigen/eosio-abigen


wanderingbort:
It looks like `eosio-abigen` is still installed to `bin`,  are you seeing this behavior after a `make install`?  @jgiszczak did we change something about install paths?

blockchain-expert:
@wanderingbort  yes

PR #4465, opened 2018-07-03
Optimize compiled contract for size

alepacheco:
With the recent increase in ram usage, the use of `-Os` instead of `-O3` as the default optimization mode makes sense.

This way we will save on the consumption of the most scarce resource on the network at the moment and help the network as a whole.

- Some experiments show a reduction of the file size of 1/3. Effectively reducing the cost to deploy 33%

### DIFF
--- a/tools/eosiocpp.in
+++ b/tools/eosiocpp.in
@@ -39,7 +39,7 @@ function build_contract {
         name=`basename $file`
         filePath=`dirname $file`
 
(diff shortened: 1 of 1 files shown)


spoonincode:
IMO this should be a user option. Maybe someone wants to do `-Oz` even.

alepacheco:
Will a command option in eosiocpp be a better alternative?

heifner:
@alepacheco yes. A command option would be better.

heifner:
Any future work on this should be done here: https://github.com/EOSIO/eosio.wasmsdk

PR #4523, opened 2018-07-06
Fix validate_b1_vesting

spartucus:
Since "stake for voting cannot be negative", after 10 years later, `claimable` is bigger than `max_claimable`, `max_claimable - claimable` is negative which is <= `stake`, which means assert never crash, b1 can claim their tokens forever.

I do not know if I understand the right, please correct me if i am wrong.

### DIFF
--- a/contracts/eosio.system/delegate_bandwidth.cpp
+++ b/contracts/eosio.system/delegate_bandwidth.cpp
@@ -205,7 +205,7 @@ namespace eosiosystem {
       const int64_t max_claimable = 100'000'000'0000ll;
       const int64_t claimable = int64_t(max_claimable * double(now()-base_time) / (10*seconds_per_year) );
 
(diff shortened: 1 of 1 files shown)


wanderingbort:
a better way of saying this is reversing the conditional.  At all points in time, b1's `stake` (the tokens it has that it cannot liquidate) must be greater than or equal to the total granted tokens (`max_claimable`) less the tokens which have been released so far (`claimable`)

This assert restricts how fast b1 can unstake its tokens, it does not affect the quantity of tokens b1 was granted.

with your changes, assuming claimable is 0 (ie b1 cannot liquidate granted tokens at all), you end up with 1'000'000'000 >= stake.  In other words a restriction on the maximum stake can be.  This condition is only `false` when `stake` > 1'000'000'000 .  A stake of 0 (which would mean liquidating all b1's tokens) is valid at any time with that math.

PR #4526, opened 2018-07-07
YubiHSM 2 based hardware wallet for keosd

spoonincode:
Add a keosd wallet that supports hardware key storage and signing via the YubiHSM 2.

### Installation

You must download the YubiHSM2 libraries and tools from yubico's website. Depending on your OS the packaging is different. Ultimately you will need the shared libraries and yubihsm-connector. Usage of yubihsm-connector is required and outside the scope of these instructions.

### keosd Launch & Configuration

macOS users: There is something disruptive about the rpath of the yubi dylibs. You will need to set DYLD_LIBRARY_PATH to the path of the YubiHSM2 libraries until this is better handled. (Even just putting the dylibs in the same directory of keosd will not work)

Ubuntu 18.04 is known to work out of the box. Other Linux are untested and you may have .so name problems because the keosd implementation is not robust yet.

To enable the YubiHSM wallet, you must configure keosd with the YubiHSM Authkey to use for negotiating a session. An example keosd command line which uses the Authkey with key ID 1 might look something like this

`./keosd --http-server-address=localhost:8900 --yubihsm-authkey 1`

macOS users need the DYLD_LIBRARY_PATH path set, so it might look like
`DYLD_LIBRARY_PATH=/Users/spoonincode/Downloads/yubihsm2-sdk/lib ./keosd --http-server-address=localhost:8900 --yubihsm-authkey 1`

### Usage
Once yubihsm-authkey is configured, a YubiHSM wallet will be available in keosd. You can see this wallet via `cleos wallet list`. Unlocking the wallet requires the password to the Authkey given for yubihsm-authkey configuration. For the out-of-box (insecure, you should change) Authkey at key ID 1 you would unlock the wallet with the command

`cleos wallet unlock -n YubiHSM --password password`

This will connect to the yubihsm-connector and negotiate a session with the given password. If everything succeeds then the wallet will become unlocked and accessible. If you would like to create a key within the YubiHSM you can

`cleos wallet create_key -n YubiHSM`

If you use an Authkey that does not have the key creation capability ("permission"), the above command will fail.

If at any time the connection to the YubiHSM or yubihsm-connector fails, the wallet will become locked and you will need to unlock it again to use.

Key deletion and importation are not supported within keosd. These operations can be handled with yubihsm-shell. Be aware that keosd only populates its internal key list when the YubiHSM wallet is unlocked so changes made via yubihsm-shell will not be immediately visible if the wallet is already unlocked.


ellipticasec:
Would that work with Yubico Yubikey 4 (which is smart card PIV / PKCS11)

spoonincode:
No, this currently requires a YubiHSM 2

luoyhang003:
Is there any docs on how to setup a secure block producer node by using YubiHSM 2?

PR #4560, opened 2018-07-09
Vote back in Producer watermark fix

### LINKED ISSUE #3835: Incorrect (really big) number of confirmed blocks in header when producer get voted out then voted in again
We recently find a bug in producer plugin that cause the synchronization fails.

This is the short story:
1. producer kayan1111132 produced block at #145488
2. kayan1111132 no longer produce block for pretty long time because he was voted out.
3. kayan1111132 got voted in again and produced block at #228494, it tried to confirm the range from #145489 to #228493 by the producer plugin, but due to the type of uint16_t, the resulting number will be wrapped into 17469
4. all the validating nodes noticed that kayan1111132 is the new producer, and set his last produced block number to be the last irreversible number (#228000), which means that kayan1111132 should only confirm block range from #228001 to #228493. This results in double-confirm exception

In my understanding, we should fix this in the producer plugin, by not confirming the blocks that already became irreversible, instead of just checking the watermark (last produce block num).

related log:
```
2315501ms thread-0   producer_plugin.cpp:987       produce_block        ] Produced block 00023850866b7827... #145488 @ 2018-06-01T16:38:35.500 signed by kayan1111132 [trxs: 0, lib: 145128, confirmed: 0]
...
...
407502ms thread-0   controller.cpp:752            start_block          ] promoting proposed schedule (set in block 227656) to pending; current block: 227989 lib: 227664 schedule: {"version":17,"producers":[{"produce
r_name":"dhanesh","block_signing_key":"EOS5jKcVPZgsog1d2k4DFjay4Jdjr1rkpFbpBtKUShNTy3gQKuSJs"},{"producer_name":"kayan","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"
kayan1111111","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111112","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name
":"kayan1111113","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111114","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_n
ame":"kayan1111115","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111121","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"produce
r_name":"kayan1111122","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111123","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"prod
ucer_name":"kayan1111124","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111125","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"p
roducer_name":"kayan1111131","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111132","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},
{"producer_name":"kayan1111133","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111134","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w
"},{"producer_name":"kayan1111135","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111141","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkM
p1w"},{"producer_name":"kayan1111142","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJNkMp1w"},{"producer_name":"kayan1111143","block_signing_key":"EOS5d9ezziQCNU5cnrTPkLxCMfqmbSg1Y5ZfosKrrwAj1YJ
NkMp1w"},{"producer_name":"thezealot","block_signing_key":"EOS8dGGh6NXTwPm81dYH94hjW5QjQnEUaZ9uEFF83RErcUMvfCiqn"}]} 

...

671501ms thread-0   producer_plugin.cpp:987       produce_block        ] Produced block 00037c8d504d7213... #228493 @ 2018-06-02T04:11:11.500 signed by kayan1111131 [trxs: 0, lib: 228000, confirmed: 0]
672003ms thread-0   producer_plugin.cpp:987       produce_block        ] Produced block 00037c8ee816b731... #228494 @ 2018-06-02T04:11:12.000 signed by kayan1111132 [trxs: 0, lib: 228180, confirmed: 17469]
```

brianjohnson5972:
#3835
Moved watermark code to on_block and to check if the future block would result in the schedule changing to prevent causing a producer from dropping a block when it was put back in the production schedule after being removed a long time in the past.
Added initial long running test for voting that exposed this problem.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -226,7 +226,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             }
 
             for (const auto& new_producer: new_producers) {
-               _producer_watermarks[new_producer] = hbn + 1;
+               _producer_watermarks[new_producer] = hbn;
             }
          }
       }
@@ -1115,23 +1115,6 @@ void producer_plugin_impl::produce_block() {
    //idump((fc::time_point::now() - hbt));
 
    block_state_ptr new_bs = chain.head_block_state();
-   // for newly installed producers we can set their watermarks to the block they became
-   if (hbs->active_schedule.version != new_bs->active_schedule.version) {
-      flat_set<account_name> new_producers;
-      new_producers.reserve(new_bs->active_schedule.producers.size());
-      for( const auto& p: new_bs->active_schedule.producers) {
-         if (_producers.count(p.producer_name) > 0)
-            new_producers.insert(p.producer_name);
-      }
-
-      for( const auto& p: hbs->active_schedule.producers) {
-         new_producers.erase(p.producer_name);
-      }
-
-      for (const auto& new_producer: new_producers) {
-         _producer_watermarks[new_producer] = chain.head_block_num();
-      }
-   }
    _producer_watermarks[new_bs->header.producer] = chain.head_block_num();
 
    ilog("Produced block ${id}... #${n} @ ${t} signed by ${p} [trxs: ${count}, lib: ${lib}, confirmed: ${confs}]",
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -202,6 +202,33 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                }
             }
          } ) );
+
+         // since the watermark has to be set before a block is created, we are looking into the future to
+         // determine the new schedule to identify producers that have become active
+         chain::controller& chain = app().get_plugin<chain_plugin>().chain();
+         const auto hbn = bsp->block_num;
+         auto new_block_header = bsp->header;
+         new_block_header.timestamp = new_block_header.timestamp.next();
+         new_block_header.previous = bsp->id;
+         auto new_bs = bsp->generate_next(new_block_header.timestamp);
+
+         // for newly installed producers we can set their watermarks to the block they became active
+         if (new_bs.maybe_promote_pending() && bsp->active_schedule.version != new_bs.active_schedule.version) {
+            flat_set<account_name> new_producers;
+            new_producers.reserve(new_bs.active_schedule.producers.size());
+            for( const auto& p: new_bs.active_schedule.producers) {
+               if (_producers.count(p.producer_name) > 0)
+                  new_producers.insert(p.producer_name);
+            }
+
+            for( const auto& p: bsp->active_schedule.producers) {
+               new_producers.erase(p.producer_name);
+            }
+
+            for (const auto& new_producer: new_producers) {
+               _producer_watermarks[new_producer] = hbn + 1;
+            }
+         }
       }
 
       void on_irreversible_block( const signed_block_ptr& lib ) {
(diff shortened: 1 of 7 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -202,6 +202,33 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         // determine the new schedule to identify producers that have become active
+         chain::controller& chain = app().get_plugin<chain_plugin>().chain();
+         const auto hbn = bsp->block_num;
+         auto new_block_header = bsp->header;
+         new_block_header.timestamp = new_block_header.timestamp.next();
+         new_block_header.previous = bsp->id;
+         auto new_bs = bsp->generate_next(new_block_header.timestamp);
+
+         // for newly installed producers we can set their watermarks to the block they became active
+         if (new_bs.maybe_promote_pending() && bsp->active_schedule.version != new_bs.active_schedule.version) {
```
wanderingbort: it looks like this code is still part of `void producer_plugin_impl::produce_block()` below?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -202,6 +202,33 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               if (_producers.count(p.producer_name) > 0)
+                  new_producers.insert(p.producer_name);
+            }
+
+            for( const auto& p: bsp->active_schedule.producers) {
+               new_producers.erase(p.producer_name);
+            }
+
+            for (const auto& new_producer: new_producers) {
+               _producer_watermarks[new_producer] = hbn + 1;
```
wanderingbort: I think this should just be hbn?  

when we start a block we confirm `(uint16_t)(hbs->block_num - watermark)` in this case `hbs` should be the last block we received the `on_block` for and if our watermark is head-block-hum + 1 this works out to be hbn - (hbn + 1) or -1

plugins/producer_plugin/producer_plugin.cpp
brianjohnson5972: Yeah, missed pulling it out when I went to a clean branch.  Cleaning up now.

PR #4566, opened 2018-07-10
Fix confirmations check

spartucus:


spartucus:
`> *2 / 3` makes little difference to `>= *2/3 +  1`, it keeps consistence with others.
For example: https://github.com/EOSIO/eos/blob/master/libraries/chain/block_header_state.cpp#L72-L76

```
static_assert(std::numeric_limits<uint8_t>::max() >= (config::max_producers * 2 / 3) + 1, "8bit confirmations may not be able to hold all of the needed confirmations");

// This uses the previous block active_schedule because thats the "schedule" that signs and therefore confirms _this_ block
auto num_active_producers = active_schedule.producers.size();
uint32_t required_confs = (uint32_t)(num_active_producers * 2 / 3) + 1;
```
If you think there's no need or i understand this wrong, no hard feelings to close the PR. :)

PR #4578, opened 2018-07-10
Fix install

larryk85:
Some more changes to where and how we are installing.


eosio_build.sh
```diff
@@ -207,6 +207,11 @@
 		MONGOD_CONF=/usr/local/etc/mongod.conf
 		OPENSSL_ROOT_DIR=/usr/local/opt/openssl
 	fi
+   
+   ${PWD}/scripts/clean_old_install.sh
+   if [ $? -ne 0 ]; then
+      exit -1
```
jgiszczak: Need to print an error message, in case clean_old_install.sh doesn't during its own execution.

eosio_build.sh
```diff
@@ -236,7 +241,7 @@
 	if ! "${CMAKE}" -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_CXX_COMPILER="${CXX_COMPILER}" \
 		-DCMAKE_C_COMPILER="${C_COMPILER}" -DWASM_ROOT="${WASM_ROOT}" -DCORE_SYMBOL_NAME="${CORE_SYMBOL_NAME}" \
 		-DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=true \
-		-DENABLE_COVERAGE_TESTING="${ENABLE_COVERAGE_TESTING}" -DBUILD_DOXYGEN="${DOXYGEN}" ..
+		-DENABLE_COVERAGE_TESTING="${ENABLE_COVERAGE_TESTING}" -DBUILD_DOXYGEN="${DOXYGEN}" -DCMAKE_INSTALL_PREFIX="/usr/local/eosio" ..
```
jgiszczak: `CMAKE_INSTALL_PREFIX` should not be set, especially not at the top level.  Per @spoonincode we are taking special care to use the `CMAKE_INSTALL_FULL_<dir>` family of variables.  See the [CMake documentation](https://cmake.org/cmake/help/v3.0/module/GNUInstallDirs.html) for details.  Existing libraries and programs which are installed all have their own `install()` CMake directives.

larryk85: We want to set the install prefix to /usr/local/eosio for a default build and allow GNUInstallDirs to build out the correct file tree from there.  The GNUInstallDirs variables use CMAKE_INSTALL_PREFIX as their root, so I am not sure what this requested change is supposed to be.

PR #4586, opened 2018-07-11
Simplified Docker setup using eosio_build.sh script

liamcurry:
This PR simplifies the Docker setup by using the standard `eosio_build.sh` script. This makes the `builder` and `dev` images unnecessary so they have been removed.

This should help to avoid problems where the Dockerfiles become out of sync with the standard build scripts.

I've tested that this is working and starts a node that produces blocks with these commands:

```bash
cd eos/Docker
docker-compose build
docker-compose up
```


liamcurry:
Thanks @erichim for the detailed response, that makes a lot of sense. I know @heifner has already closed this PR, but I was already working on some changes based on your feedback so I thought I'd share what I have.

Here are some changes I made:

- Re-added the `dev` folder which has the `Dockerfile` that builds EOS
- Changed the primary `Dockerfile` to use the `dev` image as a builder
- Added an `ENTRYPOINT` and some default `CMD`s for the images
- Added an example `docker-compose-mongo.yml` file (should be working with 1.1)

The entrypoint changes are kind of nice since they let users pass arguments to `nodeos` really easily. For example:

```bash
# Quickly startup a node that produces blocks, no other configuration necessary
docker run -it eosio/eos
# Print nodeos help test
docker run -it eosio/eos --help
# Print default config
docker run -it eosio/eos --print-default-config
```

This PR shouldn't be merged, but I'm hoping it could help inform some changes to these Dockerfiles eventually. I'm still concerned about maintaining/syncronizing two different build scripts (`eosio_build.sh` and the Docker files). I've seen issues before the 1.0 release where the Dockerfiles became outdated -- it's an easy thing to miss after all.

I think a good solution might be to break up the `scripts/eosio_build_ubuntu.sh` script into two pieces, one for building dependencies and one for building EOS. Then the `eosio/builder` image could use the Ubuntu dependency builder script, and the `eosio/dev` and `eosio/eos` images could use the EOS builder script.

heifner:
@liamcurry We appreciate your work and would entertain a new PR.

PR #4618, opened 2018-07-12
relax restrictions on host name validation

wanderingbort:
- include the resolved and raw ports in the cleared host list
- allow an incoming connection to use the host of the IP:PORT it is actually connected to at a socket level

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -199,14 +199,11 @@ namespace eosio {
             try {
                bool is_secure = con->get_uri()->get_secure();
                const auto& local_endpoint = con->get_socket().lowest_layer().local_endpoint();
-               auto local_socket_host = local_endpoint.address().to_string();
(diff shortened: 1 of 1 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -195,9 +197,16 @@ namespace eosio {
          template<class T>
          void handle_http_request(typename websocketpp::server<detail::asio_with_stub_log<T>>::connection_ptr con) {
             try {
+               bool is_secure = con->get_uri()->get_secure();
+               const auto& local_endpoint = con->get_socket().lowest_layer().local_endpoint();
+               auto local_socket_host = local_endpoint.address().to_string();
+               auto port = local_endpoint.port();
+               if (!(is_secure && port == 443) || (!is_secure && port == 80)) {
```
heifner: Is there a reason to restrict this to 443 for https and 80 for http? Seems like you could enable no matter what local_endpoint port is used.

PR #4625, opened 2018-07-13
update the turorial of multi-signature

yankunsam:


contracts/eosio.msig/README.md
```diff
@@ -46,7 +46,7 @@ Cleos usage example.
 
 Prerequisites:
    - eosio.token contract installed to eosio.token account, eosio.msig contract installed on eosio.msig account which is a privileged account.
-   - account 'treasury' is the issuer of EOS token.
+   - account 'treasury' is the issuer of EOS token.Make sure that treasury has the "eosio.msig@eosio.code" permission.
```
wanderingbort: This should not be necessary if the prerequisite above "eosio.msig account which **is a privileged account**." is fulfilled (emphasis mine). 

If that is not the case this is a bug in the code and not a bug in the tutorial

wanderingbort:
I'm going to close this as a PR since the intent of the code is in opposition to the suggested changes.  If there is a situation where a privileged eosio.msig contract requires such a permission it should be re-opened as a BUG so it can be triaged.

yankunsam:
I suggested that you write a demo to verify the tutorial before you close any issue. Thanks.

tbfleming:
@yankunsam what do you mean by that? The msig contract doesn't require eosio.msig@eosio.code permission. If you had to add that permission then either there's a bug, or you didn't set the privileged flag on eosio.msig. See https://developers.eos.io/eosio-nodeos/docs/bios-boot-sequence#section-make-eosio-msig-a-privileged-account

Issue #4297, opened 2018-06-23
Account have sufficient balance, but can't buy ram or delegatebw cpu


itleaks:
Currently, many accounts created by 3rd party have no ram and cpu resource. So "buy ram" action will fail due to lack cpu bandwidth, also "delegatebw cpu" action will fail due to lack ram resource even though the account has sufficient balance. Then, these accounts can't do anything due to lack ram or cpu resource. 
    So, I propose to add a subcommand for cleos to pack  "buy ram" action and "delegatebw cpu" action in one transaction to fix above issue. However, in this case, system only allows 3000us  cpu bandwidth(leeway) to execute one transaction ,  while the two actions usually  cost 4000+ us. So system will throw below exception .
       "Error 3081001: transaction reached the deadline set due to leeway on account CPU limits"
  Therefore, EOS should make default leeway value large enough(may 6000us)  to execute "buy ram" action and "delegatebw cpu" action in one transaction.

itleaks:
<img width="666" alt="931fabab-f290-4f94-bd3f-56f9049a057b" src="https://user-images.githubusercontent.com/7782367/41805670-69bf437e-76e1-11e8-9fe0-91a8a3fe080c.png">
<img width="1066" alt="1cb21964-c3b8-496a-8aec-9a70620312b9" src="https://user-images.githubusercontent.com/7782367/41805681-861e5ad2-76e1-11e8-881c-b1da0abe20cb.png">
<img width="1059" alt="4ac224b7-eb80-4343-a646-ce951c0b74fa" src="https://user-images.githubusercontent.com/7782367/41805706-c99fea32-76e1-11e8-9563-b0f2a5937ef9.png">
<img width="769" alt="1ef1726c-c5af-4925-aaaa-bcaefeee2024" src="https://user-images.githubusercontent.com/7782367/41805708-cdd28588-76e1-11e8-967b-c50dbdf3393c.png">

noprom:
Hi @itleaks , try to let another account buyram for `testtest1233`

itleaks:
@noprom Thanks, your suggestion is one solution to solve this problem. But it's not easy to find another account currently. And I think eos should provide this basic ability.

boriskozak:
Yeah this is a big issue.   These accounts are effectively stuck and can't do anything.  Really poor design choice to not let accounts delegate bandwidth to themselves, because there is not enough bandwidth available to perform the actions.

taokayan:
[...]

PR #4635, opened 2018-07-13
4297 delegatebw with buyram

taokayan:
#4297 add --buyram parameter to delegatebw. This will helps the dead accounts(zero cpu/net stake, no available ram) to revive. But it needs to be executed in a fast producer node so that it can be finished within the cpu leeway limit(default 3ms).

command example:
```
./cleos system delegatebw a44444444444 a44444444444 "0.3000 EOS" "1.0000 EOS" --buyram "0.1000 EOS"
```


alepacheco:
As far as I know, you could do two separate actions, the first buying ram for the account and then delegate the resources with the delegatebw action

programs/cleos/main.cpp
```diff
@@ -1111,19 +1112,35 @@ struct delegate_bandwidth_subcommand {
...
-                  ("from", from_str)
-                  ("receiver", receiver_str)
-                  ("stake_net_quantity", to_asset(stake_net_amount))
-                  ("stake_cpu_quantity", to_asset(stake_cpu_amount))
-                  ("transfer", transfer)
-                  ;
-                  wdump((act_payload));
-         send_actions({create_action({permission_level{from_str,config::active_name}}, config::system_account_name, N(delegatebw), act_payload)});
+         if (buy_ram_amount.length()) {
+            fc::variant act_payload = fc::mutable_variant_object()
```
wanderingbort: Instead of duplicating this action in both sides of the branches can we assembly the vector of actions incrementally so that this action is always in it, the `buyram` is appended based on conditions and there is only a single call to `send_actions` with the result?

Issue #4555, opened 2018-07-09
RAM warning track + Graceful shutdown


wanderingbort:
Add a configurable warning track to nodeos which will attempt a clean shutdown when the shared memory file approaches a certain percentage used.  This will prevent nodes from running into a situation where they have to shutdown immediately and leave the database potentially dirty.

brianjohnson5972:
Resolved via #4666

PR #4666, opened 2018-07-16
add guard track to database

wanderingbort:
In addition to setting the size you can now set a minimum guard on the chain state db.  If there is a db-growing operation called on controller while this guard is violated it will throw and the node should shutdown gracefully.

the config is:
`chain-state-db-guard-size-mb` defaulted to 1MiB

related to EOSIO/eos#4555


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -341,6 +342,9 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
       if( options.count( "chain-state-db-size-mb" ))
          my->chain_config->state_size = options.at( "chain-state-db-size-mb" ).as<uint64_t>() * 1024 * 1024;
 
+      if( options.count( "chain-state-db-guard-size-mb" ))
```
brianjohnson5972: Forget this comment, missed it was a **minimum** size

wanderingbort: although this is fair, it looks like I was intending the track to be 512k... hrm

Issue #4676, opened 2018-07-16
Fix cmake instructions for long running voting test


brianjohnson5972:
Need to add in copying of the python script to build directory

PR #4678, opened 2018-07-16
CMake Copy Voting Script

brianjohnson5972:
#4676
Missed cmake directive to copy python voting script


brianjohnson5972:
setup wrong, closing

Issue #4702, opened 2018-07-17
replay fails in irreversible read mode


moskvanaft:

heifner:
Fixed by #6842

PR #4703, opened 2018-07-17
fixing replay in irreversible read mode #4702

moskvanaft:


arhag:
Out-of-date and also made unnecessary by #6624.

Issue #2032, opened 2018-04-02
new cleos set contract params don't seem to behave as documented


spoonincode:
```
$ ./cleos set contract --help
Create or update the contract on an account
Usage: ./cleos set contract [OPTIONS] account contract-dir [wast-file] [abi-file]

Positionals:
  account TEXT                The account to publish a contract for
  contract-dir TEXT           The the path containing the .wast and .abi
  wast-file TEXT              The file containing the contract WAST or WASM relative to contract-dir
  abi-file TEXT               The ABI for the contract relative to contract-dir
```

I would have expected this to work..
```
$ ls ~/eos/build/contracts/eosio.bios
CMakeFiles		Makefile		eosio.bios.abi		eosio.bios.bc		eosio.bios.s		eosio.bios.wast.hpp
CTestTestfile.cmake	cmake_install.cmake	eosio.bios.abi.hpp	eosio.bios.cpp.bc	eosio.bios.wast

$ ./cleos set contract eosio ~/eos/build/contracts/eosio.bios eosio.bios.wast eosio.bios.abi
Reading WAST...
Assembling WASM...
1941681ms thread-0   main.cpp:1186                 main                 ] Failed with error: Assert Exception (10)
!"error parsing wast": Error parsing WebAssembly text file:
:1:1: expected '('

^

```

drwaldo:
same error here on Mac / master using the "currency" contract
I'm trying to follow the tutorial steps: https://github.com/EOSIO/eos#singlenode

I also tried by using eosiocpp to generate the .wast file from the source code in /contracts/currency but it throughs exactly the same error.

$ cleos set contract currency currency.wast currency.abi
```
Reading WAST...
Assembling WASM...
2370762ms thread-0   main.cpp:1186                 main                 ] Failed with error: Assert Exception (10)
!"error parsing wast": Error parsing WebAssembly text file:
:1:1: expected '('
{
^
```

xnfy520:
cleos set contract currency ./currency -p currency
./currency is a contract directory.

drwaldo:
that worked like a charm! thanks!

realcom:
when I typed that commands, error has occured like

>File is actually a directory: ../../contracts/currency
ERROR: ValidationError: wast-file=../../contracts/currency
Create or update the contract on an account
Usage: cleos set contract [OPTIONS] account wast-file [abi-file]

[...]

PR #4721, opened 2018-07-18
modify wastPath and abiPath of cleos set contract to be relative to contract dir

andriantolie:
Based on the comment on github issue #2032, the documentation of `cleos set contract` expects wastPath and abiPath to be relative to contract directory. However, this is not the case and cause misunderstanding to developers who read the instruction. This PR makes wastPath and abiPath relative to contract directory.

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -2115,11 +2115,12 @@ int main( int argc, char** argv ) {
 
       if( cpath.filename().generic_string() == "." ) cpath = cpath.parent_path();
 
-      if( wastPath.empty() )
-      {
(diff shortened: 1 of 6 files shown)


andriantolie:
PR Updated to update the integration tests.

If we support both
```
cleos set contract contracts/eosio.token contracts/eosio.token.wast contracts/eosio.token.abi
cleos set contract contracts/eosio.token eosio.token.wast eosio.token.abi
```
I think the above command will be confusing because it can be interpreted as two things. In my opinion, it's better to either change the implementation (which this PR is doing) or change the cleos help documentation (close this PR and make another PR to update the cleos help documentation).

PR #4757, opened 2018-07-19
Cleos Tab auto completion support

rlaace423:
For me, the most inconvenient thing about using cleos was the lack of command auto-complete feature.

So, I added cleos auto completion feature at build script. To do that, I also added submodule inside "externals" directory.

autocompletion

If this is not the right position or you need a fork, you can do as you please.

Original repo is here : 
cleos-auto-completion (https://github.com/hexlant/cleos-auto-completion)


heifner:
Thanks for your contribution. At this time I don't see us including this into the main eos repo.

PR #4794, opened 2018-07-23
fix missing exception

Vasiliy-Bondarenko:


heifner:
See #5026 as the sql_plugin is being moved to its own repo.

Issue #4037, opened 2018-06-11
cleos/keosd enhancements to catch common user errors with updating permissions


arhag:
There are several ways one can lock themselves out of their own account. One user (#4033) unfortunately came across one of these ways. In that particular case, it is obvious that no reasonable user would have ever intended for that permission update to occur. So ideally cleos (or any other wallet app) should have detected it and rejected. In fact these class of mistakes can even be detected and rejected by producer / API nodes that are running the appropriate plugin.

However, there is a legitimate need for users to be able to intentionally lock themselves out of their account. One example is deploying an immutable smart contract. Cleos should still support a few standard ways to lock out access of an account; although it should print a warning message and ask for confirmation again from the user.

Then there are nearly limitless ways that a user can lock themselves out of their account that may be perfectly legitimate actions but that the wallet app cannot tell if it is legitimate or a user mistake. The simplest example of this class of mistakes is changing an owner key to a public key that the user does not have the private key for.

One way to help the user catch mistakes is for cleos to warn and require confirmation from the user when modifying a permission to something that keosd cannot satisfy. But it should not disallow the change entirely because there are legitimate use cases for such actions, for example:
 * Transferring ownership of the account to another user.
 * Transitioning from a single owner key to a 2-of-3 multisig where each wallet only has one of the keys. (In this case you aren't actually locked out if you have access to at least two of the three keys, but from the perspective of a cleos/keosd which only has access to one of the keys it cannot possibly know this at the time it is signing the transaction to change the owner permission.)

ValeryDubrava:
Sometimes such cases is not mistake. E.g. we want to do locked account for smart contracts, which cannot change code or get funds and so on. So, if you are going to add the protection, please, add a flag like --force or --unprotect also to disable protection.

wanderingbort:
@snstrand please take a look at this issue in a few stages:

1 . Determine classes of "bad" permissions.  Producing guidance on these outside of code is valuable.  For instance,
  * . #4033 had a user that set authorities such that a permission was only self-satisfiable (and therefore never satisfiable).  This could be broadened to a class of authorities which have critical cycles.  
  * There are probably other classifications that are detectable

2 . Determine a plan for addressing some/all of these classes in our tooling (cleos) that we can discuss within the team.

3 . Implement / Create tests.

wanderingbort:
[...]

Issue #4307, opened 2018-06-24
mac os 10.13.4 build failed with this issues. please help me!


theshdf:
when i exc ./eosio_build.sh.  issues  as  follow:
---
-- Using custom FindBoost.cmake
-- Boost version: 1.67.0
-- Found the following Boost libraries:
--   thread
--   date_time
--   filesystem
--   system
--   program_options
--   signals
--   serialization
--   chrono
--   unit_test_framework
--   context
--   locale
--   iostreams
-- Configuring EOSIO on OS X
-- binaryen building with -std=c++11
-- binaryen building for platform x86-64
-- binaryen building with -Wall
-- binaryen building with -Werror
-- binaryen building with -Wextra
-- binaryen building with -Wno-unused-parameter
-- binaryen building with -fno-omit-frame-pointer
-- binaryen building with -fPIC
-- binaryen building with -O2
-- binaryen building with -UNDEBUG
-- Using 'SYS' as CORE symbol name
-- Using 'EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV' as public key for 'eosio' account
-- Using WASM clang => /usr/local/wasm/bin/clang
-- Using WASM llc => /usr/local/wasm/bin/llc
-- Using WASM llvm-link => /usr/local/wasm/bin/llvm-link
-- Found Secp256k1: /usr/local/lib/libsecp256k1.a
-- Configuring fc to build on Unix/Apple
-- zlib found
-- Configuring Builtins
CMake Error at libraries/CMakeLists.txt:3 (add_subdirectory):
  The source directory

    /Users/zcm/eos/libraries/softfloat

  does not contain a CMakeLists.txt file.

-- Using custom FindBoost.cmake
-- Boost version: 1.67.0
-- Found the following Boost libraries:
--   thread
--   date_time
--   filesystem
--   system
--   chrono
--   unit_test_framework
--   locale
-- Configuring ChainBase on OS X
CMake Error at libraries/wasm-jit/Source/Runtime/CMakeLists.txt:26 (find_package):
  Could not find a package configuration file provided by "LLVM" (requested
  version 4.0) with any of the following names:

    LLVMConfig.cmake
    llvm-config.cmake

  Add the installation prefix of "LLVM" to CMAKE_PREFIX_PATH or set
  "LLVM_DIR" to a directory containing one of the above files.  If "LLVM"
  provides a separate development package or SDK, be sure it has been
  installed.

[...]

PR #4807, opened 2018-07-23
4037 prevent user lock the account out by setting invalid permission

taokayan:
#4037


heifner:
See comments in #4307

Issue #4554, opened 2018-07-09
Allow writing sensitive data to a file instead of stdout


wanderingbort:
related to #3452

`cleos wallet create ...` and `cleos create key` should allow for the outputs to be written to a file instead of stdout.  This will prevent sensitive information from ending up in a logged console sessions logging.

heifner:
Resolved by #4818

PR #4818, opened 2018-07-23
Password and keys to file instead of console

brianjohnson5972:
#4554
Changed "wallet create" and "create key" to take a "--file" parameter to pass a filename to write to or else to require "--to-console" to be passed to explicitly indicate sensitive information should be displayed on the console.


brianjohnson5972:
messed up branch again

PR #4837, opened 2018-07-24
Snapshot plugin

elmato:
This PR adds a new plugin that allows to take a snapshot of the blockchain state at a determined block.
The snapshot is taken and saved after the specified block is applied.

The generated snapshot contains:

- The chain id
- The genesis state
- The block_header_state object
- All accounts (with permissions)
- All the tables


heifner:
Superseded by #5956

PR #4844, opened 2018-07-24
installation changes

larryk85:
Added an install and uninstall script and a eosio.version.hpp.  In addition to these, this PR also adds some generated cmake files to be used by external projects like eosio.contracts to resolve dependencies more easily.  This also allows for an external project to build against an uninstalled eos.


lazaridiscom:
@larryk85 , does this fix the issue mention in https://github.com/EOSIO/eos/issues/4883#issuecomment-408205732 by @jgiszczak ?

larryk85:
@lazaridiscom and @jgiszczak yes this does fix that issue.

eosio_build.sh
```diff
@@ -52,7 +52,7 @@
 	DISK_MIN=20
 	DOXYGEN=false
 	ENABLE_COVERAGE_TESTING=false
-	CORE_SYMBOL_NAME="SYS"
+	CORE_SYMBOL_NAME="EOS"
```
heifner: Looks like this was changed by mistake.

larryk85: Didn't mean to push this here, reverting.

PR #4870, opened 2018-07-26
Enable multi-threading when building boost libraries

baegjae:
When I ran eosio_build.sh, it takes too long time.
The cause was that only one thread is used when building the boost libraries.
I modified the script to use multiple threads.
Our time is precious :)

### DIFF
--- a/scripts/eosio_build_amazon.sh
+++ b/scripts/eosio_build_amazon.sh
@@ -277,7 +277,7 @@
 			printf "\\n\\tExiting now.\\n"
 			exit 1
 		fi
-		if ! "${TEMP_DIR}"/boost_1_67_0/b2 install
+		if ! "${TEMP_DIR}"/boost_1_67_0/b2 -j"${CPU_CORE}" install
 		then
 			printf "\\n\\tInstallation of boost libraries failed. 1\\n"
 			printf "\\n\\tExiting now.\\n"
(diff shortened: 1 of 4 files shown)


heifner:
@jgiszczak What are your thoughts on this? Do you think it will cause more support requests than it is worth? Also it should be `JOBS` not `CPU_CORE`, I assume.

lazaridiscom:
@heifner, `CPU_CORE` is used already within the build-scripts, so I see no reason to not use it with boost, too.

baegjae:
@lazaridiscom, Thank you for your comment. When building the MongoDB driver,  `CPU_CORE` was used as an argument within the build-scripts. Likewise, I thought `CPU_CORE` should be used when building the boost libraries.

heifner:
See #4920. I think we should pass jobs # to the build script defaulted to 1. That would help with the build problems where people don't have enough ram. Let the user indicate how many cores they want to throw at it.

lazaridiscom:
@heifner , I understand, but the issues need to be decoupled:

* One issue is the false detection of `CPU_CORE` (#4920), which should be fixed in any way
* One issue (this here) is that `CPU_CORE` is not used for boost, which should be fixed in any way

Once the above issues are solved, **then** comes (independently)  your `JOBS` suggestion:

* Issue: Provide an option to manually set JOBS number
  * Default: JOBS uses CPU_CORE (which is determined and set automatically)
  * option: user can limit the JOBS number by providing a config/cmd-line option
    * validation: if JOBS > CPU_CORE then suggest user to get a new machine

PR #4891, opened 2018-07-27
Change http status code for transaction_exception and eof_exception 

andriantolie:
Change http status code for transaction_exception and eof_exception so its error message is not swallowed by cleos. Currently they are treated as 400, which by cleos will be displayed as:
```
Error 3200006: invalid http request
Error Details:
The server has rejected the request as invalid!
Please verify this url is valid: http://127.0.0.1:8888/v1/chain/push_transaction
If the condition persists, please contact the RPC server administrator for 127.0.0.1!
```
and is confusing for the developer when they are trying to investigate the problem.

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -460,12 +460,9 @@ namespace eosio {
          } catch (chain::tx_duplicate& e) {
             error_results results{409, "Conflict", error_results::error_info(e, verbose_http_errors)};
             cb( 409, fc::json::to_string( results ));
-         } catch (chain::transaction_exception& e) {
-            error_results results{400, "Bad Request", error_results::error_info(e, verbose_http_errors)};
-            cb( 400, fc::json::to_string( results ));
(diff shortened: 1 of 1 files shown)


yqzzj:
@andriantolie I seem to met this problem.

zzj@ubuntu:~/.local/share/eosio/nodeos/config$ cleos get transaction f04c38f5692143e360ecb28d1be41ec62a5bb4a405e5dfb6731317c79a975f64
Error 3200006: invalid http request
Error Details:
The server has rejected the request as invalid!
Please verify this url is valid: http://127.0.0.1:8888/v1/history/get_transaction
If the condition persists, please contact the RPC server administrator for 127.0.0.1!

I config http_plugin , http-validate-host and so on.

It should return "tx not found",right?
It really puzzled me for a long time.

andriantolie:
Even after the PR is merged?

yqzzj:
I haven't merged yet.
My code is still 1.1.1.
I think it will be ok after merged.
Thanks.

PR #4898, opened 2018-07-27
"cleos get table" support find by string 

shuaiLiWang:
i have the following information in my testtable.
cleos get table chan chan testtable --key-type i64 --index 1
image

i can Select the phone(string) info of "wls"
cleos get table chan chan testtable --key-type i64 --index 3 --findbystr wls
image

You must make some change like this when you create contract
image


jgiszczak:
@evsward It's not necessary to comment on an issue or pull request in order to track it.  Just click the 'Subscribe' button under 'Notifications' in the right side bar.

PR #4908, opened 2018-07-27
support sha256 in hex as secondary key when calling  get_table_rows 

venediktov:
This is related to pull request #4053 
We have a contract similar to `dice.cpp`  where table has non-unique index `checksum256 uuid` column and 
we need to retrieve all entries matching that fields as it serves  a **UUID** of a handshake session .

```
 struct message {
            uint64_t id{};
            account_name from{};
            account_name to{};
            checksum256 uuid{};
            std::string data;

            uint64_t primary_key() const { return id; }
            account_name by_account_from() const { return from; }
            account_name by_account_to() const { return to; }
            key256 by_uuid() const { return checksum256_to_key256(uuid); }
            EOSLIB_SERIALIZE(message, (id)(from)(to)(uuid)(data))
        };
```
@taokayan , @heifner - Please let me know if you think it's valid case to pass `sha256` hex string to `get_table_rows`

**Use case** 

```cleos get table account account messages --key-type sha256 --index 4 -L c01b39c7a35ccc3b081a3e83d2c71fa9a767ebfeb45c69f08e17dfe3ef375a7b```

### CODE REVIEW

File: plugins/chain_plugin/chain_plugin.cpp [hunk 1]
```diff
@@ -28,6 +28,7 @@
 
 #include <fc/io/json.hpp>
 #include <fc/variant.hpp>
+#include <fc/crypto/hex.hpp>
```
heifner: This include file not needed.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1029,6 +1030,14 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
             return f128;
          });
       }
+      else if (p.key_type == "sha256") {
+          return get_table_rows_by_seckey<index256_index, checksum256_type>(p, abi, [&p](const checksum256_type& v)->key256_t {
```
heifner: `&p` not used in lambda

venediktov: yes, left over from earlier code I used it for logging, will remove it.

`i256` assumes the string is in binary 32 bytes numeric string **not in hex**  , but also when I converted sha256 hex 64 bytes key to binary 32 and used it with `i256` it did not find my rows.

venediktov: I will test it again with `i256` and let you know @heifner , thank you!

plugins/chain_plugin/chain_plugin.cpp
venediktov: agree !  leftover from last use of `from_hex( const fc::string& hex_str, char* out_data, size_t out_data_len )`  , thank you!

plugins/chain_plugin/chain_plugin.cpp
venediktov: `python -c 'print str(int("c01b39c7a35ccc3b081a3e83d2c71fa9a767ebfeb45c69f08e17dfe3ef375a7b", 16))'
`
**86892170580511395102392104225515491945535356631148603271263139179137605261947**

venediktov: @heifner , I will use that decimal again to test if `i256` works .

gustavomick:
@heifner not sure if is a issue, or a miss config from my side, but if you have a sec, please check https://eosio.stackexchange.com/questions/3219/how-to-properly-use-cleos-get-table-with-key-type-sha256-secondary-index-fa

gustavomick:
@venediktov is this working for you on 1.4.+ ? could you provide me with an example using get table? thanks

venediktov:
@gustavomick I had multiple examples that worked unfortunately due to NDA with a client I can't share it here , we only used 1.3 though 

how do you initialize your `_hasher` in your class ```_hasher{_self,self}``` ? 
can you send more code that shows ```_hasher{code, scope}``` initialization ?

venediktov:
@gustavomick , I found your issue please see a change needed for your  checksum256_to_sha256
```
static key256 checksum256_to_sha256(const checksum256 &hash)
        {
            const uint128_t *p128 = reinterpret_cast<const uint128_t *>(&hash);
            //return key256::make_from_word_sequence<uint64_t>(p64[0], p64[1], p64[2], p64[3]);
            key256 k;
            k.data()[0] = p128[0];
            k.data()[1] = p128[1];
            return k;
        }
```

I justs tested it worked with this command 

```
cleos get table testchecksum testchecksum  hasher --index 2 --key-type sha256 -L45d944598dccadd8c5a56e29c9786f05fa4b23402b735de61ce2912c86bac4de  -U45d944598dccadd8c5a56e29c9786f05fa4b23402b735de61ce2912c86bac4df
```

PR #4924, opened 2018-07-29
improve `is_valid_movement` function in tic.tac.toe example

WLBF:
The original function will take `{ "row": 0, "column": 5 }` as a valid movement. Maybe the check should be more strict.


WLBF:
I accidentally push to my master branch, already remove the commit.

PR #4938, opened 2018-07-31
"cleos get table" support find by string 

shuaiLiWang:
We need to support string lookup in accordance with our business needs. I believe there are many applications that required. And how it works？
I will convert the string to uint64_t format in the second key function in our constract，and add --findbystr parameter to “cleos get table” command. when the parameter is add，the follow string info will be convert to uint64_t format by same algorithm.

Example:
I have the following information in my testtable.
cleos get table chan chan testtable
image

I can select the phone(string) info of "2linisyiengiis"
cleos get table chan chan testtable --key-type i64 --index 3 --findbystr 2linisyiengiis
image

You must make some change like this when you create contract
image

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -227,6 +227,7 @@ public:
       uint32_t    limit = 10;
       string      key_type;  // type of key specified by index_position
       string      index_position; // 1 - primary (first), 2 - secondary index (in order defined by multi_index), 3 - third index, etc
+      string      find_by_str;
     };
 
    struct get_table_rows_result {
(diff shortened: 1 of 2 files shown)


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -341,8 +342,28 @@ class read_only {
...
-         if (p.lower_bound.size()) {
+	
+	if(p.find_by_str.size())
+	{
+		 uint64_t hash = 0;
+		 char *str = const_cast<char*>(p.find_by_str.c_str());
+		 while (*str)
+		 {
+			hash = (*str++)+(hash<<6)+(hash<<16)-hash;
+		 }
```
heifner: This is too dependent on a particular hashing method.

heifner:
Thank you for your pull request. Please see the review comments.

PR #4951, opened 2018-07-31
Remove the redundant signature recovery and hash

wanderingbort:
Before applying blocks, we add them to fork-db.  Only if fork-db signals better header-validated fork do we bother to apply the block.  As part of header-validation we validate the signature therefore, the additional check in `::apply_block` is wasteful.


wanderingbort:
@heifner  I think those were legit regressions that should now be fixed (without changing the tests)

heifner:
@wanderingbort Tests doing their jobs, nice.

Issue #4930, opened 2018-07-30
unpack_action_data unresolvable exception


Hitmanv:
I trying to use the unpack_action_data function, it compiled, but when deploy to the node, 
there is an unresolvable exception. I found that when i comment out the memo field in the transfer struct, it can be deployed to the node. Did i miss something?

```cpp
#include <eosiolib/eosio.hpp>
#include <eosiolib/asset.hpp>
#include <eosio.token/eosio.token.hpp>

using namespace eosio;
using namespace std;

struct transfer {
    account_name from;
    account_name to;
    asset         quantity;
    string        memo; // when comment out, it works fine 
};

void deposit(uint64_t self, uint64_t action, transfer data){
    // print(self);
    // print(action);
    print_f("from: %\n", name{data.from});
    print_f("to: %\n", name{data.to});
    // print_f("memo: %\n", data.memo);
    print(data.quantity.symbol);
}


extern "C" {
    void apply(uint64_t receiver, uint64_t code, uint64_t action) {
        if(code == N(eosio.token) && action == N(transfer)) {
            transfer data = unpack_action_data<transfer>();
            deposit(receiver, action, data);
        }
    }
}
```

PR #4960, opened 2018-08-01
Free allocated memory in unpack_action_data

andriantolie:
Looking at issue https://github.com/EOSIO/eos/issues/4930 in these past days, I notice some people are still using `unpack_action_data` instead of `EOSIO_ABI.`

When using `unpack_action_data` and their action structure contains string, it will throw "env.free unresolveable" when uploading the contract. I guess it's not linked with the `free` defined in eosiolib.cpp?

### DIFF
--- a/contracts/eosiolib/action.hpp
+++ b/contracts/eosiolib/action.hpp
@@ -50,7 +50,11 @@ namespace eosio {
       size_t size = action_data_size();
       char* buffer = (char*)( max_stack_buffer_size < size ? malloc(size) : alloca(size) );
       read_action_data( buffer, size );
-      return unpack<T>( buffer, size );
+      auto res = unpack<T>( buffer, size );
+      if ( max_stack_buffer_size < size ) {
+         free(buffer);
+      }
+      return res;
    }
 
    using ::require_auth;
--- a/contracts/eosiolib/action.hpp
+++ b/contracts/eosiolib/action.hpp
@@ -48,10 +48,12 @@ namespace eosio {
    T unpack_action_data() {
       constexpr size_t max_stack_buffer_size = 512;
       size_t size = action_data_size();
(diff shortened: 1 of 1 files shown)


contracts/eosiolib/action.hpp
```diff
@@ -50,7 +50,11 @@ namespace eosio {
       size_t size = action_data_size();
       char* buffer = (char*)( max_stack_buffer_size < size ? malloc(size) : alloca(size) );
```
brianjohnson5972: Add:
const bool heap_allocation = max_stack_buffer_size < size;
and use heap_allocation in both places.

andriantolie: Thanks for the suggestion. It makes it clearer to read :)
I've updated the PR

heifner:
@andriantolie I fail to see how this would have anything to do with "env.free unresolveable". Where you able to reproduce the "env.free unresolveable" error? And if so, did this fix it?

andriantolie:
Yes, this actually doesn't really solve "env.free unresolvable" entirely. But this just solves the "env.free unresolvable" problem when people creates a simple contract with just unpack_action_data.

I couldn't figure out the reasoning behind this. But seems people get "env.free unresolvable" error when they include string object inside his action struct, but never called "free" function anywhere in the contract. That's why when people use EOSIO_ABI, or when they call "free" function at least once in the contract, they don't get the error. In the .wast file, the "free" function turns from "import" to "export" when it is being used in the contract. I don't have sufficient knowledge to understand the reasoning behind this though, so if somebody can explain I'm really happy to learn....

heifner:
@larryk85 @spoonincode Is there someway we can tell wast to export instead of import to solve this unresolvable error without having to actually call free? The memory will all be free'd at the end of the action, so seems like a waste to free when it is not necessary.

PR #4967, opened 2018-08-01
Fixed asset multiplication, added declaration to privileged.h and add…

larryk85:
…ed safeguard against /usr/local


CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ project( EOSIO )
 
 enable_testing()
 
+if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+   message(WARNING "CMAKE_INSTALL_PREFIX is set to default path of ${CMAKE_INSTALL_PREFIX}, resetting to /usr/local/eosio")
```
arhag: The message could use `${CMAKE_INSTALL_PREFIX}/eosio` rather than `/usr/local/eosio` in the string to be consistent with what `CMAKE_INSTALL_PREFIX` is actually set to below.

Similar change can be made in eosio.wasmsdk.

arhag: The message could use `${CMAKE_INSTALL_PREFIX}/eosio` rather than `/usr/local/eosio` in the string to be consistent with what `CMAKE_INSTALL_PREFIX` is actually set to below.

Similar change can be made in eosio.wasmsdk.

PR #4969, opened 2018-08-01
Renamed LICENSE.txt to LICENSE

larryk85:


heifner:
Needs to be updated:
- CMakeModules/installer.cmake
- Comment in eosio_build.sh
-

PR #4972, opened 2018-08-01
 Enable Stale Production Fix

### LINKED ISSUE #4971: Testnet won't restart when all nodes are shutdown and restarted
--enable-stale-production is needed on one production node to cause a testnet to restart after all production nodes were shutdown, but if the production node is not the currently scheduled producer when it enters producer_plugin_impl::start_block it would transition from producing to speculating to waiting (if head block age was more than 5 seconds old).
Need to create a window where node with --enable-stale-production continues to stay in speculating mode till the node is caught up to date.

brianjohnson5972:
#4971 
Created a window where node with --enable-stale-production continues to stay in speculating mode till the node is caught up to date.

@heifner  - review all
@wanderingbort  - review producer_plugin.cpp

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -418,7 +418,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       start_block_result start_block(bool &last_block);
 
       fc::time_point calculate_pending_block_time() const;
-      void schedule_timer(const std::weak_ptr<producer_plugin_impl>& weak_this, const block_timestamp_type& current_block_time);
+      void schedule_delayed_production_loop(const std::weak_ptr<producer_plugin_impl>& weak_this, const block_timestamp_type& current_block_time);
 };
 
 void new_chain_banner(const eosio::chain::controller& db)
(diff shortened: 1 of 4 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -903,8 +905,13 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block(bool
 
    if (_pending_block_mode == pending_block_mode::speculating) {
       auto head_block_age = now - chain.head_block_time();
-      if (head_block_age > fc::seconds(5))
-         return start_block_result::waiting;
+      if (head_block_age > fc::seconds(5)) {
+         if (!_stale_production_window_open)
```
wanderingbort: I think there are unintentional side effects of the change as written.  The intent is to make sure we don't enter the "waiting" state (ie no timer as we must wait for a new block before doing anything) when we want to enable stale production.  

This change will have the unwanted side effect of opening a speculative block as we wait for our turn to produce

The result will be a lot of wasted work when synching for the first time (creating and discarding a speculative block in between blocks) and potentially playing and rejecting transactions learned over RPC against an older state of the block chain.

perhaps we should make a change to the code that sets (or drops) the timer for wake-up based on this result so that we don't enter the waiting state if we are enabling stale production and instead calculate the appropriate next time to wake-up and produce said block.

the code that currently "sleeps" when waiting is here: https://github.com/EOSIO/eos/blob/21a4b0448e0e4fba4ecebad939ddb9cbadb07e56/plugins/producer_plugin/producer_plugin.cpp#L1103-L1106

down a few lines is code already made that would set the timer such that we wake up at our next chance to produce a block:
https://github.com/EOSIO/eos/blob/21a4b0448e0e4fba4ecebad939ddb9cbadb07e56/plugins/producer_plugin/producer_plugin.cpp#L1127-L1155

We should make sure to log a proper indication of what is about to happen in both of these cases (there probably should have been an `fc_dlog` message for the original waiting)

brianjohnson5972: I'm confused as to why we would worry about this special case of not wasting time for someone who restarted and passed in "--enable-stale-production", but I am guessing that is because I don't know all the use cases for "--enable-stale-production".  The one reason I know is for restarting production when all producing nodes have shutdown.  So if I have a producing node, that shutdown cleanly and restarted and needs to process blocks from other currently running producing nodes, what reason would I have for "--enable-stale-production" to be set?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1124,40 +1129,49 @@ void producer_plugin_impl::schedule_production_loop() {
...
-               wake_up_time = std::min<fc::time_point>(*wake_up_time, producer_wake_up_time);
-            } else {
-               wake_up_time = producer_wake_up_time;
-            }
+   } else {
+      fc_dlog(_log, "Speculative Block Created");
+   }
+}
+
+void producer_plugin_impl::schedule_timer(const std::weak_ptr<producer_plugin_impl>& weak_this, const block_timestamp_type& current_block_time) {
```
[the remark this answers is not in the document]
brianjohnson5972: schedule_delayed_production_loop?  Other suggestions?

Issue #4986, opened 2018-08-02
"unlinkable_block_exception: Unlinkable block"


spartucus:
It appears that my node forked on height `8974060` with output message below:

![image](https://user-images.githubusercontent.com/6071887/43561808-68d02a1c-964b-11e8-88b9-0c4bbf6c888b.png)

And restart didn't solve the problem, still with error:

```
2018-08-02T03:55:52.235 thread-0   producer_plugin.cpp:310       on_incoming_block    ] 3030001 unlinkable_block_exception: Unlinkable block
unlinkable block
    {"id":"0088f5616587f3554acc4c770b5be354f74f0d93cd837ab9d3f5e7abf5f61015","previous":"0088f560a402c3fe654984914abf5bf7b5e8ce31fe9f8467c90c2e335aa6da30"}
    thread-0  fork_database.cpp:151 add
rethrow
    {}
    thread-0  controller.cpp:924 push_block
2018-08-02T03:55:52.235 thread-0   controller.cpp:924            push_block           ] 3030001 unlinkable_block_exception: Unlinkable block
unlinkable block
    {"id":"0088f562f2c451b389e9f78ad163a5a7e0fead80c02a2f9b5f9548debafb5e44","previous":"0088f5616587f3554acc4c770b5be354f74f0d93cd837ab9d3f5e7abf5f61015"}
    thread-0  fork_database.cpp:151 add
2018-08-02T03:55:52.235 thread-0   producer_plugin.cpp:310       on_incoming_block    ] 3030001 unlinkable_block_exception: Unlinkable block
unlinkable block
    {"id":"0088f562f2c451b389e9f78ad163a5a7e0fead80c02a2f9b5f9548debafb5e44","previous":"0088f5616587f3554acc4c770b5be354f74f0d93cd837ab9d3f5e7abf5f61015"}
    thread-0  fork_database.cpp:151 add
rethrow
    {}
    thread-0  controller.cpp:924 push_block
2018-08-02T03:55:52.235 thread-0   controller.cpp:924            push_block           ] 3030001 unlinkable_block_exception: Unlinkable block
unlinkable block
    {"id":"0088f563fe0bbc6235c619aa6574e32932a4d9f48968e5e07784801794e3205c","previous":"0088f562f2c451b389e9f78ad163a5a7e0fead80c02a2f9b5f9548debafb5e44"}
    thread-0  fork_database.cpp:151 add

```

And here is my config.ini:
```
agent-name = AgentName

[...]

PR #4999, opened 2018-08-02
Mitigation for 4986

wanderingbort:
Currently, hard_fail scheduled transactions in blocks can be validly billed 0 cpu usage for the failure.  All releases have had a bug where this 0 value is ambiguously used as a totem that indicates that the execution engine should time the transaction subjectively.  As a result, accounts with low cpu availability could cause validators to subjectively reject a good block when subjectivity should have been disabled ( by the billed cpu time == 0 totem ).  Validators at v1.1.1 or earlier, were awarding those accounts some leeway cpu time (removed in 1.1.2) that made these validators slightly more tolerant when encountering transactions in this state.  However, the success or failure to validate the objectively good block would depend on whether the local validator could process the transaction, to its eventual objective failure, in the accounts available CPU + leeway time.

This has likely been the root cause of several intermittant failures to sync, particularly on slower hardware.

This fix will provide the same semantics for scheduled transactions as normal transactions, which is that the subjectivity is disabled when validating.

resolves EOSIO/eos#4986

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -70,7 +70,7 @@ namespace eosio { namespace chain {
          }
       }
 
-      if( billed_cpu_time_us > 0 )
+      if( explicit_billed_cpu_time || billed_cpu_time_us > 0 )
          validate_cpu_usage_to_bill( billed_cpu_time_us, false ); // Fail early if the amount to be billed is too high
 
       // Record accounts to be billed for network and CPU usage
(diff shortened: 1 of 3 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -70,7 +70,7 @@ namespace eosio { namespace chain {
          }
       }
 
-      if( billed_cpu_time_us > 0 )
+      if( explicit_billed_cpu_time || billed_cpu_time_us > 0 )
```
heifner: This is not needed as `validate_cpu_usage_to_bill` is just going to validate `billed_cpu_time_us` is <= than `objective_duration_limit`

arhag:
Replaced by #5001.

PR #5001, opened 2018-08-02
Fix for #4986 and proper CPU billing for soft_fail and hard_fail

arhag:
This PR makes the following changes:
- Fix for #4986
- `transaction_context` contains new boolean field `explicit_billed_cpu_time` which is true iff the `billed_cpu_time_us` was explicitly set (always happens while validating a block). `transaction_context` will never change the value of `billed_cpu_time_us` field if `explicit_billed_cpu_time` is true.
- In the case of soft_fail, the node will now always try to correctly bill for the billable CPU time of the error handler's execution on top of the billable CPU time of the execution of the original deferred transaction which had failed and caused the error handler to run in the first place. 
- In the case of hard_fail, the node will always try to bill the appropriate amount (billable CPU time expended so far) but correctly limited by the maximum of what the protocol rules allow it to charge. This prevents controller from calling `add_transaction_usage` with an inappropriately high amount which would cause it to throw an exception that propagates up until it causes the production of the block to fail.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -468,7 +468,7 @@ struct controller_impl {
                                         fc::time_point deadline,
                                         fc::time_point start,
                                         uint32_t billed_cpu_time_us,
-                                        bool from_block                    ) {
+                                        bool explicit_billed_cpu_time = false ) {
       signed_transaction etrx;
       // Deliver onerror action containing the failed deferred transaction directly back to the sender.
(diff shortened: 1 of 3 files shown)


libraries/chain/controller.cpp
```diff
@@ -624,11 +630,22 @@ struct controller_impl {
 
       trace->elapsed = fc::time_point::now() - trx_context.start;
```
heifner: This is already done above. Also the comment is wrong as only hard failure logic is below.

PR #5026, opened 2018-08-03
Dev additional plugins

asiniscalchi:
just run cmake with 
$ cmake .. -DEOSIO_ADDITIONAL_PLUGINS="<path_to_plugin_0>[;<path_to_plugin_1] ..."

have a look to https://github.com/asiniscalchi/eosio_sql_plugin 
plugin side you have to add:
eosio_additional_plugin(sql_db_plugin)

with the plugins targets you want to link with.

Cheers


programs/nodeos/CMakeLists.txt
```diff
@@ -1,3 +1,10 @@
+include(additionalPlugin)
+
+foreach(ADDITIONAL_PLUGIN_SOURCE_DIR ${EOSIO_ADDITIONAL_PLUGINS})
+    message(STATUS "[Additional Plugin] source dir: ${ADDITIONAL_PLUGIN_SOURCE_DIR}")
+    add_subdirectory(${ADDITIONAL_PLUGIN_SOURCE_DIR} ${CMAKE_BINARY_DIR}/additional_plugins/${ADDITIONAL_PLUGIN_SOURCE_DIR})
```
wanderingbort: this would appear to concatenate the whole path to the additional plugin to the directory.  Will this cause problems if the path is relative aka "../../../my_plugin" ? 

When I was investigating something similar I used
```
get_filename_component(plugin_name "${external_plugin_path}" NAME)
```
to get just the last directory name in the path (aka "my_plugin" in the example) and concatenated that to the CMAKE_BINARY_DIR.  It carries the problem of all plugins needing unique directory names.

asiniscalchi: Ah! You right.
I see 2 solution :
1. use a random string 
2. use fixed path + incremental integer (ex plugin_1, plugin_2)

asiniscalchi: 3! With a bit of magic I could use the plugin first target name (ex build/additional_plugins/sql_db_plugin)

Targets have to be unique in cmake ... so it will handle the error cases for us.

asiniscalchi: @wanderingbort done

programs/nodeos/CMakeLists.txt
```diff
@@ -1,3 +1,11 @@
+include(additionalPlugin)
+
+foreach(ADDITIONAL_PLUGIN_SOURCE_DIR ${EOSIO_ADDITIONAL_PLUGINS})
+    string(UUID ADDITIONAL_PLUGIN_SOURCE_DIR_MD5 NAMESPACE "00000000-0000-0000-0000-000000000000" NAME ${ADDITIONAL_PLUGIN_SOURCE_DIR} TYPE MD5)
```
wanderingbort: I like where you went with this, should normalize the input first via `get_filename_component( ... ABSOLUTE ... )` ?

 I assume that ".././my_plugin", "../my_plugin", "../my_plugin/../my_plugin", "..//my_plugin" etc are all going to have different UUIDs if we don't normalize them.

asiniscalchi: I don't think it's necessary: cmake will fail because you got targets with same name.

scripts/eosio_build_ubuntu.sh
```diff
@@ -62,7 +62,7 @@
 
 	DEP_ARRAY=(clang-4.0 lldb-4.0 libclang-4.0-dev cmake make automake libbz2-dev libssl-dev \
 	libgmp3-dev autotools-dev build-essential libicu-dev python2.7-dev python3-dev \
-    autoconf libtool curl zlib1g-dev doxygen graphviz)
+    autoconf libtool curl zlib1g-dev doxygen graphviz libsoci-dev)
```
jgiszczak: This dependency should be installed by the plugin build process, not by the eosio build process.

gleehokie:
@wanderingbort is this stable now and should it be used by the community?

cc32d9:
@gleehokie better after https://github.com/EOSIO/eos/issues/5229  is resolved

PR #5043, opened 2018-08-05
MongoDB Filtering and Optimizations

ScottSallinen:
This PR adds several new configuration options to the MongoDB plugin, including some optimizations.

- Adds `filter-on` functionality, similar to the history plugin.
- Adds `filter-out` functionality, to help avoid a history of spam.
- Adds toggle for storing raw blocks. Setting this off decently improves performance.
- Adds toggle for storing block state (headers). Setting this off slightly improves performance.
- Adds toggle for storing raw tx traces. Setting this off slightly improves performance. (This may be undesired, as this is kinda the whole point of history plugins -- but maybe you just want block headers or something.)
- Add toggle to swap action history being determined via `accepted` txs and instead by `applied` txs. This drastically improves performance due to removing multiple lookups and writes on every irreversible block to all TX's that are to be confirmed.
- Add option to caches ABI serializations. Setting this on greatly improves performance (at the cost of memory).

Example settings:
```
plugin = eosio::mongo_db_plugin
mongodb-uri = mongodb://localhost:27017
mongodb-raw-blocks = false
mongodb-block-state = false
mongodb-store-act-via-txtraces = true
mongodb-cache-abis = true
mongodb-filter-on  = *
mongodb-filter-out = spammer::
mongodb-queue-size = 2048
abi-serializer-max-time-ms = 5000
```

The result is a mongodb synchronization time that is competitive with the regular history plugin.

### DIFF
--- a/plugins/mongo_db_plugin/mongo_db_plugin.cpp
+++ b/plugins/mongo_db_plugin/mongo_db_plugin.cpp
@@ -18,6 +18,7 @@
 #include <boost/thread/thread.hpp>
 #include <boost/thread/mutex.hpp>
 #include <boost/thread/condition_variable.hpp>
+#include <boost/algorithm/string.hpp>
 
 #include <queue>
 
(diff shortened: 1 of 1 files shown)


ScottSallinen:
Aha - the problem with decentralized development, not everyone is in sync :smile: 
Would you like me to do the integration against the develop branch, or would you like to do that? 

We'd need to check which optimizations we can keep based on the changes you made in your latest PR -- for example, removing certain collections allow you to totally remove their hooks/queues (e.g. on accepted tx), which is a huge performance boost.

heifner:
Actually, `develop` branch is also out of date, I plan on adding your changes to the `gh#4968-inline-actions` branch which is currently sitting in a PR.

ScottSallinen:
Ok, good to know! Once this all gets merged up theres more optimizations I'd like to look at as well (such as inserting queued block states in bulk rather than one by one).
Is there a better way we can sync up with planned changes to this plugin?

heifner:
Coordinate with me as I'm actively working mongo changes. Feel free to ping me on telegram.

PR #5066, opened 2018-08-06
Gh#4968 inline actions

### LINKED ISSUE #4968: MongoDB Plugin accounts Collection Missing Smart Contract Created Accounts
It appears that the mongoDB plugin is missing accounts that have been created by a smart contract. For example the account [accountcreat](https://bloks.io/account/accountcreat) has a [contract](https://github.com/angelol/smart_account_creator/blob/master/smart_account_creator.cpp) that on line 80 calls `INLINE_ACTION_SENDER` for `newaccount`. Any of the accounts seen on a block explorer for [accountcreat](https://bloks.io/account/accountcreat) do not appear to show up.

heifner:
Resolves #4968 

- Action traces now stored instead of actions
  -- New action_traces collection replaces actions collection
  -- action_traces include all actions including inline, implicit, and scheduled
- irreversible attribute on transactions and blocks is now only added with a value of true when irreversible to prevent a race condition of an update of transaction overwriting the irreversible true with a false. This is now needed since upsert is used for insert and strict ordering is not guaranteed. 
- Simplified transaction serialization and now store all eosio abi as abi_def instead of bytes. 
- Added filter-on and filter-out similar to history_plugin.
  -- Thanks to Scott Sallinen (TeamGreymass) for implementation.
- Added options to not store blocks, block-states, transactions, transaction-traces, and action-traces.

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -55,6 +55,11 @@ namespace eosio { namespace chain {
       set_abi(abi, max_serialization_time);
    }
 
+   void abi_serializer::add_specialized_unpack_pack( const string& name,
+                                                     std::pair<abi_serializer::unpack_function, abi_serializer::pack_function> unpack_pack ) {
+      built_in_types[name] = std::move( unpack_pack );
+   }
+
    void abi_serializer::configure_built_in_types() {
 
       built_in_types.emplace("bool",                      pack_unpack<uint8_t>());
--- a/libraries/chain/include/eosio/chain/abi_serializer.hpp
+++ b/libraries/chain/include/eosio/chain/abi_serializer.hpp
@@ -86,10 +86,12 @@ struct abi_serializer {
       return false;
    }
 
-   static const size_t max_recursion_depth = 32; // arbitrary depth to prevent infinite recursion
-
    typedef std::function<fc::variant(fc::datastream<const char*>&, bool, bool)>  unpack_function;
    typedef std::function<void(const fc::variant&, fc::datastream<char*>&, bool, bool)>  pack_function;
+
+   void add_specialized_unpack_pack( const string& name, std::pair<abi_serializer::unpack_function, abi_serializer::pack_function> unpack_pack );
+
+   static const size_t max_recursion_depth = 32; // arbitrary depth to prevent infinite recursion
 
 private:
 
--- a/libraries/chain/include/eosio/chain/abi_serializer.hpp
+++ b/libraries/chain/include/eosio/chain/abi_serializer.hpp
@@ -134,6 +134,7 @@ namespace impl {
              std::is_same<T, packed_transaction>::value ||
              std::is_same<T, transaction_trace>::value ||
              std::is_same<T, transaction_receipt>::value ||
+             std::is_same<T, base_action_trace>::value ||
              std::is_same<T, action_trace>::value ||
              std::is_same<T, signed_transaction>::value ||
              std::is_same<T, signed_block>::value ||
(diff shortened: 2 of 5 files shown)


heifner:
brief of changes: (draft - in progress)

The optional `eosio::mongo_db_plugin` provides archiving of blockchain data into a MongoDB. It is recommended that the plugin be added to a non-producing node as it is designed to shut down on any failed insert into the MongoDB and is resource intensive. For best results dedicate a nodeos instance to running this one plugin. The rationale behind this shutdown on error is so that any issues with the mongo database or connectivity can be fixed and the nodeos can be restarted without having to resync or replay.

The `mongo_db_plugin` does slow down replay/resync as the conversion of block data to JSON and insertion into MongoDB is resource intensive. The plugin does use a worker thread for processing the block data, but this does not help much when replaying/resyncing. 

It is recommended that a large `--abi-serializer-max-time-ms` value be passed into the nodeos running the `mongo_db_plugin` as the default abi serializer time limit is not large enough to serialize large blocks. 

To avoid speculate transaction_trace data in MongoDB enable `read-mode = head`. See read-mode documentation for more information. 

`mongo_db_plugin` options:
```
Config Options for eosio::mongo_db_plugin:
  -q [ --mongodb-queue-size ] arg (=256)
                                        The target queue size between nodeos 
                                        and MongoDB plugin thread.
  --mongodb-wipe                        Required with --replay-blockchain, 
                                        --hard-replay-blockchain, or 
                                        --delete-all-blocks to wipe mongo 
                                        db.This option required to prevent 
                                        accidental wipe of mongo db.
  --mongodb-block-start arg (=0)        If specified then only abi data pushed 
                                        to mongodb until specified block is 
                                        reached.
  -m [ --mongodb-uri ] arg              MongoDB URI connection string, see: 
                                        https://docs.mongodb.com/master/referen
                                        ce/connection-string/. If not specified
                                        then plugin is disabled. Default 
                                        database 'EOS' is used if not specified
                                        in URI. Example: mongodb://127.0.0.1:27
                                        017/EOS
```

The following mongo collections are created:
* accounts - created on accepted transaction
  * Currently limited to just name and ABI if contract on account
  * Mostly for internal use as the stored ABI is used to convert action data into JSON for storage as associated actions on contract are processed.
  * Invalid ABI on account will prevent conversion of action data into JSON for storage resulting in just the action data being stored as hex. For example, the original eosio.system contract did not provide ABI for the `onblock` action and therefore all 'onblock` action data is stored as hex until the time `onblock` ABI is added to the eosio.system contract.

* actions - created on accepted transaction
  * action_num - the # of the action in its transaction
  * trx_id
  * account
  * name
  * authorization
    * actor
    * permission
  * data - Available if valid ABI stored on account and the action data can be converted to mongo json for storage.
  * hex_data

* block_states - created on accepted block
  * block_num
  * block_id
  * block_header_state
  * validated
  * in_current_chain

* blocks - created on accepted block
  * block_num
  * block_id
  * irreversible - updated to true on irreversible block
  * block

* transaction_traces - created on applied transaction
  * transaction_trace

* transactions - created on accepted transaction
  * trx_id
  * irreversible - updated to true on irreversible block
  * transaction_header
  * signing_keys
  * actions
  * context_free_actions
  * transaction_extensions
  * signatures
  * context_free_data

DenisCarriere:
Looking forward to see this merged in the future releases 😄

heifner:
@DenisCarriere Sorry about the actions->action_traces and changes to transactions. Should be more stable going forward. Hopefully, it will not cost you too much rework on: https://github.com/EOS-BP-Developers/eosio-mongodb-queries

DenisCarriere:
@heifner No worries, I'll adjust accordingly to these changes.

I'll rework MongoDB Queries once this is released 👍

DenisCarriere:
🎉 Woot

DenisCarriere:
@heifner & @ScottSallinen

I'm trying to only use `mongodb-filter-on` however it still returns all `action_traces` as if it would be `mongodb-filter-on=*` (the default).

**`config.ini` example**

```
mongodb-filter-on = eosio:delegatebw:
mongodb-filter-on = eosio:undelegatebw:
```

At the moment, this config still returns `eosio:onblock` & all others

Maybe I'm missing something?

heifner:
Looks like you are correct. I'll get a fix put in.

jafri:
Perhaps a timestamp on both transaction_trace and action_trace would be useful

PR #5072, opened 2018-08-06
Include cmake in ubuntu build dependencies

jpopesculian:
cmake is required for the ubuntu build but not included in the list of
dependencies which can cause a failed build


jpopesculian:
Sorry didn't see develop branch will make a new one

PR #5096, opened 2018-08-07
Use eosio::chain::config::system_account_name constant 

EugeneChung:
Use eosio::chain::config::system_account_name constant instead of N(eosio)


EugeneChung:
Aligned with #5106

PR #5097, opened 2018-08-08
Log which peer gives unexpected length messages

ScottSallinen:
This allows a quicker understanding of which peer is misbehaving, allowing faster turnaround time to blocking bad peers.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2122,7 +2122,7 @@ namespace eosio {
                            auto index = conn->pending_message_buffer.read_index();
                            conn->pending_message_buffer.peek(&message_length, sizeof(message_length), index);
                            if(message_length > def_send_buffer_size*2 || message_length == 0) {
-                              elog("incoming message length unexpected (${i})", ("i", message_length));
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2122,7 +2122,7 @@ namespace eosio {
                            auto index = conn->pending_message_buffer.read_index();
                            conn->pending_message_buffer.peek(&message_length, sizeof(message_length), index);
                            if(message_length > def_send_buffer_size*2 || message_length == 0) {
-                              elog("incoming message length unexpected (${i})", ("i", message_length));
+                              elog("incoming message length unexpected (${i}), from ${p}", ("i", message_length)("p",boost::lexical_cast<std::string>(conn->socket->remote_endpoint())));
```
heifner: This can throw an exception which will be caught below and report std::exception below. Pass `boost::system::error_code ec;` to `remote_endpoint(ec);` to avoid exception.

ScottSallinen:
Updated -- I'm not sure that it can cause an exception there without excepting on the connection close, but good to squash it anyways I suppose.

PR #5102, opened 2018-08-08
Define N(eosio) as system_account_name at the eosiolib level

EugeneChung:
I wanted to use eosio::config namespace but it showed the symbol collision with proxy::config class. So I decided to use eosio::cfg namespace.


EugeneChung:
I will check build again.

PR #5106, opened 2018-08-08
Define N(eosio) as system_account_name at the eosiolib level

EugeneChung:
I wanted to use eosio::config namespace but it showed the symbol collision with proxy::config class. So I had to use eosio::cfg namespace.


wanderingbort:
Thanks for the submission.  We are no longer maintaining the eosio.system contracts or the eosiolib from this repository.  We port over eosiolib changes to maintain parity while it is in deprecation but the official repository is : https://github.com/EOSIO/eosio.wasmsdk/

Once the base feature is added and released there, then the contracts repo can be upgraded to take advantage of it: https://github.com/EOSIO/eosio.contracts/

PR #5117, opened 2018-08-08
Add the git "version string" to get_info

spoonincode:


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -90,6 +90,7 @@ class read_only {
       uint64_t                block_net_limit = 0;
       //string                  recent_slots;
       //double                  participation_rate = 0;
+      optional<string>        server_version_string;
```
spoonincode: The optional here is not strictly required. But it protects a new cleos from displaying an empty `server_version_string` when talking to an old nodeos.

PR #5147, opened 2018-08-10
http Connection:Close without content-length

jjnetcn:
some bp or node RPC with large response data, 
it maybe set connection:close, and not set conntent-length, in this case,
we should read all data until EOF( recv()==0 )

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -81,22 +81,36 @@ namespace eosio { namespace client { namespace http {
       // Process the response headers.
       std::string header;
       int response_content_length = -1;
+      int isclose = 0;
       std::regex clregex(R"xx(^content-length:\s+(\d+))xx", std::regex_constants::icase);
+      std::regex coregex(R"xx(^connection:\s+close)xx", std::regex_constants::icase);
       while (std::getline(response_stream, header) && header != "\r") {
          std::smatch match;
          if(std::regex_search(header, match, clregex))
             response_content_length = std::stoi(match[1]);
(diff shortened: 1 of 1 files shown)


jjnetcn:
u can check and tcpdump with this example:  
./cleos -u http://api-direct.eosasia.one get block 10377000

spoonincode:
That is actually how it use to work. But I changed it in this commit
https://github.com/EOSIO/eos/commit/cbac9107740c60a987de73de1a4d60eec259c3b1#diff-eac8b8c3999103c2ad988a230978627c
The reason is when I was testing HTTPS I was encountering HTTPS servers/proxies where the TCP connection is closed before the TLS close notification is received. I couldn't just wait until EOF via a while loop because I'd actually get an exception thrown before EOF.

Unfortunately the servers I was testing against the exhibited this problem are no longer accessible to me and I'm entirely sure how to easily reproduce said configuration. This change will probably be tripped up on those types of servers though.

EugeneChung:
https://github.com/EOSIO/fc/issues/10 This is due to fc/vendor/websocketpp.

spoonincode:
@jjnetcn I can't reproduce the original issue with
`./cleos -u http://api-direct.eosasia.one get block 10377000`
Nor can I reproduce the issue with a handful of other API nodes including ones behind cloudflare. Do you think this change is still needed?

jjnetcn:
@spoonincode  i thank this is an tempory solution and not prefect， 
many http server under proxy ex cloudflare are not according to RFC2616 strictly，
and be used more and more tricks to compatible it.  i thank here should be closed this issue

PR #5160, opened 2018-08-10
 Remove WAST support from cleos set code/contract

spoonincode:
The textual format of WebAssembly continues to be somewhat fluid. For example, the renaming of `grow_memory` with `memory.grow`. The binary format has been stable for some time. Only consume the binary format in cleos now -- the eosio tools (like eosiocpp) have produced binary .wasm files since the 1.0 release by default as do most other toolchains. This protects us from further deviations of the textual format

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -114,7 +114,6 @@ Options:
 #include <Inline/BasicTypes.h>
 #include <IR/Module.h>
 #include <IR/Validate.h>
-#include <WAST/WAST.h>
 #include <WASM/WASM.h>
 #include <Runtime/Runtime.h>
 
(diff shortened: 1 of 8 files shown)


programs/cleos/main.cpp
```diff
@@ -2161,27 +2160,15 @@ int main( int argc, char** argv ) {
...
-         wasm = vector<uint8_t>(wast.begin(), wast.end());
-      }
-      else {
-         std::cerr << localized("Assembling WASM...") << std::endl;
-         wasm = wast_to_wasm(wast);
-      }
+      std::cerr << localized(("Reading WASM from " + wasmPath + "...").c_str()) << std::endl;
+      fc::read_file_contents(wasmPath, wast);
+      EOS_ASSERT( !wast.empty(), wast_file_not_found, "no wasm file found ${f}", ("f", wasmPath) );
+      vector<uint8_t> wasm = vector<uint8_t>(wast.begin(), wast.end());
```
heifner: Somewhat minor, but `wast` variable should be `wasm`

PR #5198, opened 2018-08-13
Read only mode

heifner:
- Create a new read-only db mode where incoming blocks are processed but with no speculative processing of blocks. This is similar to head mode except for no speculative execution or relaying of transactions are performed.
- In read-only mode no p2p connections are allowed since transactions are not speculatively executed or relayed.
- This mode is useful for plugins such as the mongo_db_plugin where consistent state is the most important characteristic.
- Note that non-irreversible blocks are still processed so forks are recorded when they happen.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1423,6 +1423,7 @@ void controller::push_confirmation( const header_confirmation& c ) {
 
 transaction_trace_ptr controller::push_transaction( const transaction_metadata_ptr& trx, fc::time_point deadline, uint32_t billed_cpu_time_us ) {
    validate_db_available_size();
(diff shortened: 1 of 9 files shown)


abourget:
> In read-only mode no p2p connections are allowed since transactions are not speculatively executed or relayed.

Curious.. how can _incoming blocks_ be applied if no p2p connections are allowed?

abourget:
Do you mean we refuse incoming connections, but outgoing are okay?

heifner:
@abourget Correct incoming connections are not allowed since transactions would be dropped on the floor and not speculatively executed.

ScottSallinen:
This is /almost/ perfect for a use case I have in mind: block-only relays, to speed up block propagation. I imagine them being run alongside other seed nodes, so producers have routes for TX's, but also a fast route for blocks.

With this in mind, would it be possible to optionally enable accepting incoming connections? This way, a block relay could be made public for other producer's own block relays to connect to.

Issue #5199, opened 2018-08-13
Prevent Spurious Long Running Test failures


brianjohnson5972:
As long as there is more than one node producing blocks it may take several blocks for transactions to make it into a block.  Adding logic to ensure valid transactions do not get reported.

PR #5230, opened 2018-08-14
Prevent ABI Timeout

brianjohnson5972:
#5199
Increased timeout for ABI in launcher

### DIFF
--- a/tests/Cluster.py
+++ b/tests/Cluster.py
@@ -129,7 +129,7 @@ class Cluster(object):
         if self.staging:
             cmdArr.append("--nogen")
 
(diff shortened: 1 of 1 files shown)


tests/Cluster.py
```diff
@@ -129,7 +129,7 @@ def launch(self, pnodes=1, totalNodes=1, prodCount=1, topo="mesh", p2pPlugin="ne
         if self.staging:
             cmdArr.append("--nogen")
 
-        nodeosArgs="--max-transaction-time 50000 --abi-serializer-max-time-ms 50000 --filter-on * --p2p-max-nodes-per-host %d" % (totalNodes)
+        nodeosArgs="--max-transaction-time 50000 --abi-serializer-max-time-ms 65000 --filter-on * --p2p-max-nodes-per-host %d" % (totalNodes)
```
heifner: I think we could have gone with even larger. Nothing in these tests require a small value.

PR #5257, opened 2018-08-15
Warn when trying to set code with something non-wasm

spoonincode:
Add a warning when set code/contract detects an apparent non-wasm being used (like wast)


spoonincode:
Replaced with #5259

PR #5286, opened 2018-08-16
Producer logging fix

heifner:
- Fixed access to pending_block_state when not available. 
- Cleanup some logging.
- Added some sanity checks for pending_block_state.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -931,10 +931,8 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block(bool 
 
    if (_pending_block_mode == pending_block_mode::speculating) {
       auto head_block_age = now - chain.head_block_time();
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -931,8 +931,10 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block(bool
 
    if (_pending_block_mode == pending_block_mode::speculating) {
       auto head_block_age = now - chain.head_block_time();
-      if (head_block_age > fc::seconds(5))
+      if (head_block_age > fc::seconds(5)) {
+         fc_dlog(_log, "Greater than 5 secs behind, waiting");
```
brianjohnson5972: Will this give us any more information than the log statement "Waiting till another block is ..." which will be printed out in producer_plugin_impl::schedule_production_loop?

heifner: Good point, I'll remove it.

PR #5291, opened 2018-08-16
Use transaction delay_sec in get_required_keys determination

### LINKED ISSUE #5281: authorization_manager::get_required_keys use provided_delay wrong way
clieos use RPC "/v1/chain/get_required_keys" to detect which user support pubkey  can hold  the trx's actions authority.  it wrong when an account's authority has delay weight.
at current version,  authorization_manager::get_required_keys not set and use the default provided_delay=0 params, and failed with Error 3090003 in checker.satisfied.
we should use  seconds(trx.delay_sec) to replace provided_delay in make_auth_checker to fix it.

for example:
we set account ACCOUNT with authority(pubkey+delay5), then transfer with delay trans.
it failed with 3090003 
./cleos set account permission ACCOUNT active '{"keys":[{"key":"PUBKEY", "weight":1}], "threshold":2, "waits":[{"wait_sec":5, "weight":1}] owner 
./cleos transfer ACCOUNT ANOTHERACCOUNT "0.0001 EOS" -n 5  
( i apply an patch #5280 let cleos support deferred action with -n option)

heifner:
Resolves #5281 

- Use the transaction delay_sec in the determination of required keys.-

### DIFF
--- a/contracts/eosiolib/types.h
+++ b/contracts/eosiolib/types.h
@@ -27,13 +27,13 @@ typedef uint64_t account_name;
 
 /**
  * @brief Name of a permission
- * @details Name of an account
+ * @details Name of a permission
  */
 typedef uint64_t permission_name;
 
 /**
  * @brief Name of a table
- * @details Name of atable
+ * @details Name of a table
  */
 typedef uint64_t table_name;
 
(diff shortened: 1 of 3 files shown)


jjnetcn:
pretty_input.delay_sec is unsigned_int transacton::delay_sec;
get_required_keys param provided_delay is fc::microseconds.
it will cast thought
template<typename T>
unsigned_int::operator T() const { return static_cast<T>(value); } with T = fc::microseconds, 
and static_cast call explicit fc::microseconds::microseconds(int64_t c) to set member variable _count to c.

so, call inline function fc::seconds(pretty_input.delay_sec) to replace pretty_input.delay_sec, 
set _count=delay_sec*1000000

heifner:
Although not strictly needed for the this change the fc submodule update was intentional for the changes to varint.hpp.

PR #5316, opened 2018-08-18
Added check for INSTALL_PREFIX path existence

NorseGaud:
When installing into Mac OSX, I get:

<pre>[100%] Built target print_floats
Install the project...
-- Install configuration: ""
-- Installing: /usr/local/eosio/etc/eosio
CMake Error at cmake_install.cmake:44 (file):
  file INSTALL cannot make directory "/usr/local/eosio/etc/eosio": No such
  file or directory

make: *** [install] Error 1

 >>>>>>>>>>>>>>>>>>>> MAKE installing EOSIO has exited with the above error.</pre>

To fix this, I had to execute `sudo mkdir -p /usr/local/eosio && sudo chown -R user:staff /usr/local/eosio`, but this should be clearly indicated in the script with the failure.


NorseGaud:
I'll document --full if this is accepted and merged in

larryk85:
The install and uninstall script should require sudo to install.  I agree that the error is a bad error message, as it doesn't help the user to resolve the problem, but I think a better error message is needed, as in use sudo to run the script (shouldn't need to manually construct the directory, cmake should handle it).  In addition, nodeos should manage the creation of its default locations, I am not aware of nodeos failing if that directory does not exist, it should create that directory and continue forward.  Adding the nodeos default directories for all the supported architectures to the uninstall.sh is okay.

NorseGaud:
I will do the following:
1. See why nodeos isn't creating /usr/local/eosio
2. Default directory cleanup should definitely happen
3. Have the installation scripts throw a warning if not being run with sudo

NorseGaud:
Follow up to this.
1. Running with sudo creates the directory. 
2. Opened https://github.com/EOSIO/eos/pull/5383 for --full uninstall argument
3. Sudo check inside pull/5383

PR #5339, opened 2018-08-20
Add block info & ram delta to action_trace

### LINKED ISSUE #5149: mongodb plugin :: Why no block_time on the action object returned from actions collection
Right now we are on mainnet-1.1.0

Using the history plugin, actions returned by get_actions have block_time included. We are starting to use mongo plugin and have noticed that there is no block_time property included in the action object returned from the actions collection. Right now the only way we've been able to determine block_time is by:

1. get transaction id from action 
2. find the transaction in transactions collection 
3. get block_id from transaction 
4. find the block in blocks collection 
5. get the timestamp of block

Is there a reason why block_time is not included in the action object returned from actions collection? I understand that the mongo id for that action has an embedded timestamp, but that is pretty much useless that record was inserted at the same time it was created.

Any help much appreciated!

heifner:
Resolves #5149 

action_trace & transaction_trace changes:
- Add block_num, block_time, producer_block_id, context_free to action_trace
- Add block_num, block_time, producer_block_id to transaction_trace
- Add ram_delta per account to action_trace

mongo_db_plugin chnages:
- Update block_state with irreversible=true on irreversible signal
- Add transaction_trace status to each action_trace
- Insert transaction_trace before its action_traces


libraries/chain/include/eosio/chain/transaction_context.hpp
```diff
@@ -74,6 +74,7 @@ namespace eosio { namespace chain {
          vector<action_receipt>        executed;
          flat_set<account_name>        bill_to_accounts;
          flat_set<account_name>        validate_ram_usage;
+         flat_map<account_name, int64_t> account_ram_delta; // reset for each action
```
arhag: Instead of explicitly resetting for each action, move this field to `apply_context`.

libraries/chain/controller.cpp
```diff
@@ -1388,7 +1395,7 @@ fork_database& controller::fork_db()const { return my->fork_db; }
 
 void controller::start_block( block_timestamp_type when, uint16_t confirm_block_count) {
    validate_db_available_size();
-   my->start_block(when, confirm_block_count, block_status::incomplete );
+   my->start_block(when, confirm_block_count, block_status::incomplete, block_id_type() );
```
arhag: With the change to `optional<block_id_type>`, the last argument of this function call would be `optional<block_id_type>()` instead of `block_id_type()`.

libraries/chain/transaction_context.cpp
```diff
@@ -365,6 +368,7 @@ namespace eosio { namespace chain {
       if( ram_delta > 0 ) {
          validate_ram_usage.insert( account );
       }
+      account_ram_delta[account] += ram_delta;
```
arhag: Instead of adding this line here, you can create a new wrapper function in `apply_context`:

```
void apply_context::add_ram_usage( account_name account, int64_t ram_delta ) {
   trx_context.add_ram_usage( account, ram_delta );
   account_ram_delta[account] += ram_delta;
}
```

Then all calls to `transaction_context::add_ram_usage` in eosio_contract.cpp and apply_context.cpp could be replaced with calls to `apply_context::add_ram_usage`.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -266,6 +279,28 @@ void mongo_db_plugin_impl::accepted_transaction( const chain::transaction_metada
...
+      // It is better to avoid this potential confusion by not logging traces from speculative execution, i.e. emitted
+      // from an incomplete block. This means that traces will not be recorded in speculative read-mode, but
+      // users should not be using the mongo_db_plugin in that mode anyway.
+      //
+      // Allow logging traces if node is a producer for testing purposes, so a single nodeos can do both for testing.
+      //
+      // It is recommended to run mongo_db_plugin in read-mode = read-only.
+      //
+      if( !is_producer && !t->producer_block_id.valid() )
+         return;
```
heifner: This allows our nodeos_run_test.py --mongodb test to pass since it has the same nodeos as both a producer and running the mongo_db_plugin. I figured people running a single test node might find that useful.

sim31:
I think `account_ram_delta` will be a very useful field. So I'm looking forward for this getting merged. But I see currently each element in account_ram_delta is an array which represents a pair (account, delta). I noticed, it could make it kind of complicated to make queries for this kind of schema. Would it be possible to make it an object?

sim31:
Say, I want to get all actions where account A payed for RAM. Now I have to query array of arrays for that. I've tried in ways, which seemed straightforward to me first - they didn't work. I had to dig around on the internet to find out [how to do that](https://stackoverflow.com/questions/12629692/querying-an-array-of-arrays-in-mongodb), but that's not the problem. The problem is that I've found in multiple places where people say that that's a bad document structure and there will be problems when trying to do aggregation for example. [Here for example](https://stackoverflow.com/questions/41938095/is-there-a-way-to-query-for-dict-of-arrays-of-array-in-mongodb).

Now, let's say I want to create an index on the payer field, to make the query I mentioned above faster. Didn't have time to think about it much yet, but I have no idea how to do it for the current document structure. Would be straightforward if `account_ram_delta` would contain objects. Now I could only find [this](https://stackoverflow.com/questions/16363980/mongodb-array-of-arrays-index) so far, which again suggests that it's not the best document structure.

I'm under impression mongodb expects objects to be used these kind of scenarios and it's probably optimized better if objects are used as well.

heifner:
@sim31 

```
	"account_ram_deltas" : [
		{
			"account" : "eosio",
			"delta" : -128
		},
		{
			"account" : "gyydsmbxguge",
			"delta" : 128
		}
	],
```

Does this work for you?

heifner:
@arhag Ready for your review of changes.

PR #5358, opened 2018-08-22
Idle HTTP connection timeout of eosio::http_plugin

EugeneChung:
In consistent with keep-alive feature of websocketpp, this option should be needed.
https://github.com/EOSIO/fc/pull/15


b1bart:

Issue #5350, opened 2018-08-21
Notify amount of RAM bought in buyram action


chengevo:
Currently there's no way to tell how much RAM bought in a `buyram` transaction, except for comparing the RAM change before and after buying.

The reason is, unlike any token transfer in EOSIO system, is handled by specific action in that token contract, by tracking that action, you can tell much token an account received/sent. However, when a account "received" certain amount of RAM, the `eosio.system` contract simply changes account's RAM amount, and didn't reveal how much RAM you bought.

Event though there's a `buyrambytes` function provided, while upon buying ram, user can specify the amount of RAM want to buy instead of the token to spend. Underneath it firstly calculate the amount of EOS going to cost given how much RAM you want to buy.

 But I think this approach might not that accurate, given that the actual RAM market price can differ that the price doing the calculation. So user cannot 100 percent trust that he/she can get the RAM claims to buy.

It would be nice to include a receipt indicating how much RAM bought in a transaction. And it can be achieved by setup an action(function), which did nothing just accepting the amount of RAM bought in a transaction. Every time when the RAM market response with filling result, call this function as a inline action, so that buyer can be notified how much RAM he/she got.

The implementation might looks like this:

```c++
void system_contract::buyram( account_name payer, account_name receiver, asset quant )
{
    // ....
     int64_t bytes_out;

      const auto& market = _rammarket.get(S(4,RAMCORE), "ram market does not exist");
      _rammarket.modify( market, 0, [&]( auto& es ) {
          bytes_out = es.convert( quant_after_fee,  S(0,RAM) ).amount;
      });
      
      SEND_INLINE_ACTION( *this, notifyRAM, {from, N(active)}, {payer, receiver, bytes_out, "RAM bought"} );
    
    // ....
}
```

wanderingbort:
Thank you for the submission, we are tracking features/additions to those contracts out of the [eosio.contracts repository](https://github.com/EOSIO/eosio.contracts/).  I have copied this issue there and will close this one as a result.

PR #5368, opened 2018-08-22
Notify RAM bought in buyram transaction

chengevo:
This PR addresses the issue #5350 .

Use inline action to 'notify' payer and receiver in an `buyram` transaction exactly how much RAM the receiver got.

Following is the outcome using cleos tool to buy RAM:

image


tbfleming:
This is a nice addition, but we’d need to train users that the output is preliminary. To get the final value you have to get the inline action from the execution which happened on the block that was included in the chain. cleos shows a speculative execution.

chengevo:
Exactly

chengevo:
@wanderingbort hi, I've moved this PR to eosio.contracts repo, [#52](https://github.com/EOSIO/eosio.contracts/pull/52)

PR #5378, opened 2018-08-22
Added has_auth2 Contract API

xJonathanLEI:
Added a new contract interface `has_auth2` which mimics the same pattern of `require_auth` and `require_auth2`.

The reason for adding the interface is for contracts to be able to check a specific permission without first calling `require_auth2` and then cathing the exception thrown.

PR #5385, opened 2018-08-23
Added has_auth2 Contract API

xJonathanLEI:
Added a new contract interface `has_auth2` which mimics the same pattern of `require_auth` and `require_auth2`. The reason for adding the interface is for contracts to be able to check a specific permission.

I thought it was ok even without the addition of this new interface, as I could first call `require_auth2` and then catch the exception thrown, until I found that **exceptions are not allowed in contracts**. There's **no way** to check a specific permission.

This is a duplicate PR as #5378 as I submitted the previous PR to the `master` branch incorrectly. Please merge this one instead.

### DIFF
--- a/contracts/eosiolib/action.h
+++ b/contracts/eosiolib/action.h
@@ -119,6 +119,15 @@ extern "C" {
     */
    void require_auth2( account_name name, permission_name permission );
 
+   /**
+    *  Verifies that @ref name has auth.
+    * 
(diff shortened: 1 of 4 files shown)


xJonathanLEI:
This will result in a hard fork. It probably should be added to the Hard Fork milestone.

b1bart:

PR #5416, opened 2018-08-24
Add a new WASM backend: wabt

spoonincode:
Add a new WASM backend using wabt’s interpreter. This interpreter is considerably faster than the current default of binaryen — a replay on the mainnet typically completes in less than half the time compared to binaryen.

For now this will be considered experimental.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -27,3 +27,6 @@
 [submodule "libraries/fc"]
 	path = libraries/fc
 	url = https://github.com/EOSIO/fc
+[submodule "libraries/wabt"]
+	path = libraries/wabt
+	url = http://github.com/EOSIO/wabt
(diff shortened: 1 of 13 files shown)


abourget:
When do you expect this to graduate from its `experimental` status?

maoueh:
Maybe this issue might shed some lights: https://github.com/EOSIO/eos/pull/5754

spoonincode:
> When do you expect this to graduate from its experimental status?

ah, yeah, that comment was written when the PR was fresh over 25 days ago. Since then it has been tested more. By all objective measures it is better than binaryen. No known problems with it either. You will have to use your own judgement for when it is safe to enable on your own network & nodes. But as pointed out it will be the default in 1.4.

PR #5425, opened 2018-08-26
Add UNIX socket HTTP server to keosd

spoonincode:
The http_plugin now has the capability to serve its HTTP RPC on a UNIX domain socket in addition to the TCP and TLS over TCP sockets already supported. This option is currently only present in keosd even though nodeos also uses the http_plugin.

### User facing changes for keosd
* A new option unix-socket-path configures where to create the unix socket. By default this will be “keosd.sock” and created within keosd's data-dir which by default is ~/eosio.wallet/ . So the socket would reside at ~/eosio.wallet/keosd.sock.
You can also use an absolute path for unix-socket-path if you want to place it more globally accessible. This may be useful for more advanced deployments where, for example, nodeos and keosd run as different users but are members of the same group. The socket has mode 0660 so only the user who executed keosd as well as members of the group of the user who executed keosd can open it.
If you wish to disable the unix socket, you can use either `unix-socket-path =` (in the config file) or `--unix-socket-path ''` (on the command line) to set an empty path which disables it. If you desire to only use the unix socket, you can similarly set the http-server-address to empty which will disable it. 
* The default HTTP port for keosd has changed to 8900. This resolves an inconstancy where when you launched keosd manually with no options it would listen on 8888, but if you had cleos launch keosd for you it would listen on 8900. Be aware that this default port change won’t take affect for current users — your config.ini will still contain the original 8888 default which will be treated as user specified and thus overrule the new default.

### Using the socket with cleos
By default cleos will still try to access the wallet on http://127.0.0.1:8900. To have it access via UNIX socket pass something like `--wallet-url unix://${HOME}/eosio-wallet/keosd.sock`

### Using the socket with nodeos producer_plugin
If you would like to use keosd via UNIX socket as a signature provider the syntax is of the form
`signature-provider = EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV=KEOSD:unix:///home/username/eosio-wallet/keosd.sock/v1/wallet/sign_digest`
Nodeos will discover that /home/username/eosio-wallet/keosd.sock is the path to the socket, and /v1/wallet/sign_digest is the URL path.

### Implementation/PR notes
I had to pretty much copy pasta the asio_endpoint.hpp code from websocketpp in to a new local_endpoint.hpp and change a few items. It seemed like tcp was baked deep in to how the original code worked.

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -3,6 +3,7 @@
  *  @copyright defined in eos/LICENSE.txt
  */
 #include <eosio/http_plugin/http_plugin.hpp>
+#include <eosio/http_plugin/local_endpoint.hpp>
 #include <eosio/chain/exceptions.hpp>
 
 #include <fc/network/ip.hpp>
@@ -79,9 +80,42 @@ namespace eosio {
 
           static const long timeout_open_handshake = 0;
       };
+
+      struct asio_local_with_stub_log : public websocketpp::config::asio {
+          typedef asio_local_with_stub_log type;
+          typedef asio base;
+
+          typedef base::concurrency_type concurrency_type;
+
+          typedef base::request_type request_type;
+          typedef base::response_type response_type;
+
+          typedef base::message_type message_type;
+          typedef base::con_msg_manager_type con_msg_manager_type;
+          typedef base::endpoint_msg_manager_type endpoint_msg_manager_type;
+
+          typedef websocketpp::log::stub elog_type;
+          typedef websocketpp::log::stub alog_type;
+
+          typedef base::rng_type rng_type;
+
+          struct transport_config : public base::transport_config {
+              typedef type::concurrency_type concurrency_type;
+              typedef type::alog_type alog_type;
+              typedef type::elog_type elog_type;
+              typedef type::request_type request_type;
+              typedef type::response_type response_type;
+              typedef websocketpp::transport::asio::basic_socket::local_endpoint socket_type;
+          };
+
+          typedef websocketpp::transport::asio::local_endpoint<transport_config> transport_type;
+
+          static const long timeout_open_handshake = 0;
+      };
    }
 
    using websocket_server_type = websocketpp::server<detail::asio_with_stub_log<websocketpp::transport::asio::basic_socket::endpoint>>;
+   using websocket_local_server_type = websocketpp::server<detail::asio_local_with_stub_log>;
    using websocket_server_tls_type =  websocketpp::server<detail::asio_with_stub_log<websocketpp::transport::asio::tls_socket::endpoint>>;
    using ssl_context_ptr =  websocketpp::lib::shared_ptr<websocketpp::lib::asio::ssl::context>;
 
@@ -98,6 +132,7 @@ namespace eosio {
          size_t                   max_body_size;
 
          websocket_server_type    server;
+         websocket_local_server_type local_server;
 
          optional<tcp::endpoint>  https_listen_endpoint;
          string                   https_cert_chain;
@@ -163,7 +198,7 @@ namespace eosio {
          }
 
          template<class T>
-         static void handle_exception(typename websocketpp::server<detail::asio_with_stub_log<T>>::connection_ptr con) {
+         static void handle_exception(typename websocketpp::server<T>::connection_ptr con) {
             string err = "Internal Service error, http: ";
             try {
                con->set_status( websocketpp::http::status_code::internal_server_error );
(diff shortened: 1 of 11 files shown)


programs/cleos/main.cpp
```diff
@@ -148,6 +149,22 @@ FC_DECLARE_EXCEPTION( localized_exception, 10000000, "an error occured" );
     FC_MULTILINE_MACRO_END \
   )
 
+//copy pasta from keosd's main.cpp
```
spoonincode: the changes in this file are left over cruft from when unix socket was the default; probably could remove all this

plugins/http_plugin/http_plugin.cpp
```diff
@@ -275,17 +329,41 @@ namespace eosio {
...
    http_plugin::~http_plugin(){}
 
    void http_plugin::set_program_options(options_description&, options_description& cfg) {
-      cfg.add_options()
-            ("http-server-address", bpo::value<string>()->default_value("127.0.0.1:8888"),
-             "The local IP and port to listen for incoming http connections; set blank to disable.")
+      my->mangle_option_names();
+      if(current_http_plugin_defaults.default_unix_socket_path.length())
+         cfg.add_options()
+            (my->unix_socket_path_option_name.c_str(), bpo::value<boost::filesystem::path>()->default_value(current_http_plugin_defaults.default_unix_socket_path),
```
jgiszczak: Needs to be std::string with manual parsing into bfs path, or paths with spaces are spuriously rejected.  (Ancient boost defect.)

plugins/http_plugin/http_plugin.cpp
```diff
@@ -353,7 +431,14 @@ namespace eosio {
             }
          }
 
-         if( options.count( "https-server-address" ) && options.at( "https-server-address" ).as<string>().length()) {
+         if( options.count( my->unix_socket_path_option_name ) && !options.at( my->unix_socket_path_option_name ).as<boost::filesystem::path>().empty()) {
+            boost::filesystem::path sock_path = options.at(my->unix_socket_path_option_name).as<boost::filesystem::path>();
+            if (sock_path.is_relative())
+               sock_path = app().data_dir() / sock_path;
+            my->unix_endpoint = asio::local::stream_protocol::endpoint(sock_path.string());
```
jgiszczak: Does boost do something sane with permissions on the socket file?

plugins/http_plugin/http_plugin.cpp
```diff
@@ -275,17 +329,41 @@ namespace eosio {
...
+      my->mangle_option_names();
+      if(current_http_plugin_defaults.default_unix_socket_path.length())
+         cfg.add_options()
+            (my->unix_socket_path_option_name.c_str(), bpo::value<boost::filesystem::path>()->default_value(current_http_plugin_defaults.default_unix_socket_path),
+             "The filename (relative to data-dir) to create a unix socket for HTTP RPC; set blank to disable.");
+
+      if(current_http_plugin_defaults.default_http_port)
+         cfg.add_options()
+            (my->http_server_address_option_name.c_str(), bpo::value<string>()->default_value("127.0.0.1:" + std::to_string(current_http_plugin_defaults.default_http_port)),
+             "The local IP and port to listen for incoming http connections; set blank to disable.");
```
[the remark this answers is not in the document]
spoonincode: That works for me.
```
$ programs/nodeos/nodeos 2>&1 | grep listen
2018-08-29T23:42:50.438 thread-0   http_plugin.cpp:344           plugin_initialize    ] configured http to listen on 127.0.0.1:8888
2018-08-29T23:42:50.440 thread-0   http_plugin.cpp:401           plugin_startup       ] start listening for http requests
2018-08-29T23:42:50.440 thread-0   net_plugin.cpp:3010           plugin_startup       ] starting listener, max clients is 25
```

```
$ programs/nodeos/nodeos --http-server-address '' 2>&1 | grep listen
2018-08-29T23:43:46.528 thread-0   net_plugin.cpp:3010           plugin_startup       ] starting listener, max clients is 25
```
I also double checked and the config.ini works being blank too.

plugins/http_plugin/http_plugin.cpp
spoonincode: It just will get the default via whatever current umask is. You'll find in the local_endpoint.hpp file I monkey with the umask to ensure the file is created 0660

plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp
```diff
@@ -0,0 +1,790 @@
...
+    lib::asio::local::stream_protocol::socket & get_next_layer() {
+        return *m_socket;
+    }
+
+    lib::asio::local::stream_protocol::socket & get_raw_socket() {
+        return *m_socket;
+    }
+
+    std::string get_remote_endpoint(lib::error_code & ec) const {
+        return "fixme";
```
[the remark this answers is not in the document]
spoonincode: The iostream transport implementation just returns a static "iostream transport" response w/o setting error. I'll change this to something other than fixme,

plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp
```diff
@@ -0,0 +1,790 @@
...
+               ::unlink(ep.path().c_str());
+            else if(test_ec != boost::system::errc::no_such_file_or_directory)
+               bec = test_ec;
+        }
+
+        if (!bec) {
+            m_acceptor->open(ep.protocol(),bec);
+        }
+        if (!bec) {
+            mode_t old_mask = umask(S_IXUSR|S_IXGRP|S_IRWXO);
```
jgiszczak: Shouldn't this honor an existing more restrictive umask?  In particular 077.

spoonincode: that's a valid point

spoonincode: I am thinking now, maybe should not be touching this at all

plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp
```diff
@@ -0,0 +1,788 @@
...
+        }
+        if (!bec) {
+            m_acceptor->listen(boost::asio::socket_base::max_listen_connections,bec);
+        }
+        if (bec) {
+            if (m_acceptor->is_open()) {
+                m_acceptor->close();
+            }
+            log_err(log::elevel::info,"asio listen",bec);
+            ec = bec;//make_error_code(error::pass_through);
```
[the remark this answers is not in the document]
spoonincode: thanks; bad copy pasta is bad

PR #5438, opened 2018-08-27
Simplify Wallet Tools EOSIO Blockchain Detection for End Users

marlonwilliams:
A simple addition of core_symbol to get_info results will go a long way in making it easier for users to interact with all EOSIO chains, by simply entering an endpoint. The get_info results from this endpoint should return not only the chain info as is today, but also the CORE_SYMBOL, making EOSIO extremely easy to interact with/configure for multi-chain use from an end user's perspective. Both eosjs and eosjs-ecc supports different prefixes for public keys (e.g EOSxxx, TLOSxxx, etc), so core support for returning the core symbol can help make multichain support faster and out of the box! 

Please consider this non-breaking, simple change.

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -947,6 +947,7 @@ read_only::get_info_results read_only::get_info(const read_only::get_info_params
       //std::bitset<64>(db.get_dynamic_global_properties().recent_slots_filled).to_string(),
       //__builtin_popcountll(db.get_dynamic_global_properties().recent_slots_filled) / 64.0,
       app().version_string(),
+      symbol().name(),
    };
 }
 
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -97,6 +97,7 @@ public:
       //string                  recent_slots;
       //double                  participation_rate = 0;
       optional<string>        server_version_string;
+      string                  core_symbol;
    };
    get_info_results get_info(const get_info_params&) const;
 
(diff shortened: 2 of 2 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -947,6 +947,7 @@ read_only::get_info_results read_only::get_info(const read_only::get_info_params
       //std::bitset<64>(db.get_dynamic_global_properties().recent_slots_filled).to_string(),
       //__builtin_popcountll(db.get_dynamic_global_properties().recent_slots_filled) / 64.0,
       app().version_string(),
+      symbol().name(),
```
jgiszczak: To successfully interact with the chain, you also need to know the precision of that symbol, which `symbol().name()` does not provide.  I'm not certain what the best format of that string should be.  Perhaps @jcalfee can help.  At the very least, it should be `precision,name`.  It may be even better to provide precision in another field.

jgiszczak:
New features need to be submitted for merge against the `develop` branch rather than `master`.

plugins/chain_plugin/chain_plugin.cpp
marlonwilliams: Thank you so much for your prompt review of this. The use case that this is valuable for is to support EOSIO-based blockchains that has non EOS prefixes. The precision is irrelevant in these cases, as the prefix is the only factor when generating the public key. Please see the following in eosjs-ecc for reference. https://github.com/EOSIO/eosjs-ecc/pull/30

The general idea is this:

1) User downloads an EOSIO tool (e.g. wallet, toolkit) that involves some sort of key-creation/key lookup process.

2) Tool asks user for an API endpoint

3) With a simple call to get_info for the specified API endpoint, the tool gets the chainId, and the CORE_SYMBOL associated with said chain

4) For the tool's "Generate Key"/"Validate Key"/"Lookup Account" features, it knows which prefix is being used by that chain based on the results from get_info. As a result, it's able to call the functions in eosjs and eosjs-ecc [such as eosjs-ecc. privateToPublic("key", "key_prefix")] automatically w/o any additional interfaces requesting the prefix from the user.

Scatter recently implemented this feature, but currently requires users to know the prefix of the blockchain they're interacting with (e.g. https://github.com/GetScatter/ScatterDesktop/pull/19)

Looking forward to your feedback.

Thanks!

marlonwilliams:
Sorry I was unaware of which branch to submitted this request against. I'll await your feedback on my comment and re-issue on the developer branch as needed. Thanks again

plugins/chain_plugin/chain_plugin.cpp
jgiszczak: Yes, that is your use case.  However it's not the only one.  If the intent is to provide an automated means to interact with an EOS.IO chain with an arbitrary CORE_SYMBOL, you also need to know the precision to perform transactions using the symbol.  Your code may not do that, but someone else may wish to write code that does.  Automating the full discovery required to successfully do that is perfectly reasonable.

marlonwilliams: Good point. Will take a look and issue a pull request to the ``develop`` branch with the changes.

marlonwilliams: New pull request with incorporated feedback submitted in https://github.com/EOSIO/eos/pull/5463

Regards

jgiszczak:
Closing in favor of #5463.

PR #5439, opened 2018-08-27
suppress greylist related exception loggings

taokayan:
This pull request will suppress exception loggings on transaction failure due to lack of resources of greylisted account.


arhag:
Replaced by #5462.

PR #5444, opened 2018-08-27
stop creating anonymous volumes during image build

tmclane:
Best practices dictate that you do not define VOLUMEs in the DockerFile.

Instead the volume should be defined in the docker-compose.yml file such that you can name them so they are not 'lost' and can be identified later and managed. 
Creating them during the build means they are only used during the image creation process and then are never utilized again.   This is a wasted effort plus a waste of disk space.


tmclane:
@ericiles done.  I didn't run into that problem because in my docker-compose.yml file I had mounted a volume at that location so it existed.

PR #5446, opened 2018-08-27
New test to ensure require_recipient to generator is ignored

heifner:
- New unittest to ensure require_recipient to the generator of notify is ignored.


unittests/api_tests.cpp
```diff
@@ -437,6 +437,36 @@ BOOST_FIXTURE_TEST_CASE(action_tests, TESTER) { try {
...
+      produce_blocks(2);
+      create_account( N(testapi) );
+      create_account( N(acc5) );
+      produce_blocks(1);
+      set_code( N(testapi), test_api_wast );
+      set_code( N(acc5), test_api_wast );
+      produce_blocks(1);
+
+      // test require_notice
+      auto scope = std::vector<account_name>{N(testapi)};
```
brianjohnson5972: Is this actually used?

unittests/api_tests.cpp
```diff
@@ -437,6 +437,36 @@ BOOST_FIXTURE_TEST_CASE(action_tests, TESTER) { try {
...
+      auto test_require_notice = [this](auto& test, std::vector<account_name>& scope){
```
brianjohnson5972: Does the lambda actually get you anything here?  If not simplify it and just make it part of the method.

PR #5450, opened 2018-08-28
Better Dockerfile

NorseGaud:
Better Dockerfile to check if user has set args or not, then output what we're using for verification.


NorseGaud:
Ignore, I screwed something up.

PR #5452, opened 2018-08-28
Docker Improvements

NorseGaud:
Various Docker improvements:

- Better Readme, in rational order
- New docker-compose.yml with container-name
- New docker-compose-testnet.yml to replace -latest.yml, with container-names and volume naming changes
- Dockerfile improvements with $user so forks work
- User home for data-volume

### DIFF
--- a/Docker/latest/docker-compose.yml
+++ b/Docker/latest/docker-compose.yml
@@ -1,12 +1,6 @@
 version: "3"
 
 services:
-  builder:
-    container_name: "${COMPOSE_PROJECT_NAME}_builder"
-    build:
-      context: ../builder
-    image: eosio/builder
-
   nodeos:
     env_file:
       .env


Docker/docker-compose-testnet.yml
```diff
@@ -0,0 +1,35 @@
+version: "3"
+
+services:
+  nodeosd_test:
+    container_name: nodeosd_test
+    image: eosio/eos:v1.1.0
```
ericiles: It is probably best to either use the ```latest``` tag or actually specify the latest version (```v1.2.3``` as of right now). This would also apply to the keosd_test container as well.

If the ```latest``` tag is used, the documentation should direct the user to use ```docker-compose pull``` to get the latest version of the image before running ```docker-compose up```

NorseGaud: Yep, understood. Curious though, could we backport the older 1.1.x branch with v1.1.0? A lot of testnet docs say you need to start with 1.1.0, then upgrade, so I'm just looking to make it as easy as possible for folks to get up and running with as little changes as possible.

NorseGaud: Or, actually, how about `image: eosio/eos:$EOS_TESTNET_VERSION` and then telling the user to set an ENV? That would put the responsibility on the user and not on the eos devs to rememeber to update the value in the file. Life would be much easier :)

ericiles: I would feel comfortable using an environmental variable for the testnet docker-compose file. It might also be a good idea to create a .env file with the default value of ```EOS_TESTNET_VERSION``` being set to ```latest```. This would allow a testnet to be spun up without having to define any environmental variables, while still allowing the tag to be overidden by the environmental variable if it is present.

heifner: @NorseGaud the Jungle testnet has special requirements for v1.1.0 because of an issue they ran into. The mainnet does not have that issue.

NorseGaud: Thanks guys, I'll get the changes into the PR. I believe using .env file is a perfect solution to allow users greater control in an easy way.

NorseGaud:
Ok, I've created and tested all of the scripts and docker files. I've added a few extras to help users get started quickly. Let me know if anything is off in the README and I can address. I really like the .env route. Thanks for the reminder about that.

NorseGaud:
Ok, the final version is committed. I've changed the entire design so that users can just create a copy of the "latest" directory to run multiple eos instances. I've tested "latest" and "dev" setup.sh scripts to confirm they're working. 

Let me know what you think.

heifner:
@ericiles @NorseGaud where are we with this? Looks like there are conflicts now.

NorseGaud:
@heifner Just waiting on Eric I believe.

jacktang:
Hello, can I issue EOS token in eos-dev docker now?

NorseGaud:
@jacktang, yep, modify the .env under the Dev directory to EOS

PR #5462, opened 2018-08-28
CPU greylist fixes

arhag:
Replaces #5439 rebased to release/1.2.x.

When a transaction exhausts CPU resources possibly due to greylisting, it will throw a subjective `greylist_cpu_usage_exceeded` rather than an objective `tx_cpu_usage_exceeded`. Beyond being more accurate in its description, this has important implications when it comes to CPU resource exhaustion in deferred transactions. A producer with greylisting enabled could fail a deferred transaction in a manner that is not reproducible during validation (since greylisting does not apply during validation), which would lead to generating blocks that are rejected by all other nodes.

This PR also suppresses logging greylist related exceptions within the http plugin.

Note that the current greylisting code (both before and after the changes in this PR) modifies the values of `net_limit` and `objective_duration_limit` in `transaction_context::finalize` based on subjective data (the non-elastic limits used only for those accounts in the local greylist) thus making those variables no longer objective generally. This doesn't break the greylisting feature, but I think it isn't appropriate use of those variables and could potentially lead to future problems if those variables are later used in ways that assume they must be objective. In short, this part of the code is in need of a refactor to clean up a lot of the complicated and hard-to-follow logic, but this patch release is not the time for it.

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -92,9 +92,10 @@ namespace eosio { namespace chain {
       // Calculate the highest network usage and CPU time that all of the billed accounts can afford to be billed
       int64_t account_net_limit = 0;
       int64_t account_cpu_limit = 0;
-      bool greylisted = false;
-      std::tie( account_net_limit, account_cpu_limit, greylisted ) = max_bandwidth_billed_accounts_can_pay();
+      bool greylisted = false, greylistedCPU = false;
+      std::tie( account_net_limit, account_cpu_limit, greylisted, greylistedCPU) = max_bandwidth_billed_accounts_can_pay();
       net_limit_due_to_greylist |= greylisted;
+      cpu_limit_due_to_greylist |= greylistedCPU;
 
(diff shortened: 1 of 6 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -296,9 +300,15 @@ namespace eosio { namespace chain {
...
-               EOS_THROW( tx_cpu_usage_exceeded,
-                          "transaction was executing for too long",
-                          ("now", now)("deadline", _deadline)("start", start)("billing_timer", now - pseudo_start) );
+               if (cpu_limit_due_to_greylist) {
+                  EOS_THROW( greylist_cpu_usage_exceeded,
+                           "transaction was executing for too long",
+                           ("now", now)("deadline", _deadline)("start", start)("billing_timer", now - pseudo_start) );
+               } else {
+                  EOS_THROW( tx_cpu_usage_exceeded,
+                           "transaction was executing for too long",
```
heifner: Would like to see a diff string for these two cases, to make it more obvious if grepping.

PR #5463, opened 2018-08-28
Simplify Wallet Tools EOSIO Blockchain Detection for End Users

marlonwilliams:
Follow-up to thread: https://github.com/EOSIO/eos/pull/5438

A simple addition of core_symbol to get_info results will go a long way in making it easier for users to interact with all EOSIO chains, by simply entering an endpoint. The get_info results from this endpoint should return not only the chain info as is today, but also the CORE_SYMBOL, making EOSIO extremely easy to interact with/configure for multi-chain use from an end user's perspective. Both eosjs and eosjs-ecc supports different prefixes for public keys (e.g EOSxxx, TLOSxxx, etc), so core support for returning the core symbol can help make multichain support faster and out of the box!

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -98,7 +98,7 @@ public:
       //double                  participation_rate = 0;
       optional<string>        server_version_string;
       string                  core_symbol;
-      uint64_t                precision;
+      uint64_t                core_symbol_precision;
    };
    get_info_results get_info(const get_info_params&) const;
 
@@ -617,7 +617,7 @@ private:
 FC_REFLECT( eosio::chain_apis::permission, (perm_name)(parent)(required_auth) )
 FC_REFLECT(eosio::chain_apis::empty, )
 FC_REFLECT(eosio::chain_apis::read_only::get_info_results,
-(server_version)(chain_id)(head_block_num)(last_irreversible_block_num)(last_irreversible_block_id)(head_block_id)(head_block_time)(head_block_producer)(virtual_block_cpu_limit)(virtual_block_net_limit)(block_cpu_limit)(block_net_limit)(server_version_string)(core_symbol)(precision) )
+(server_version)(chain_id)(head_block_num)(last_irreversible_block_num)(last_irreversible_block_id)(head_block_id)(head_block_time)(head_block_producer)(virtual_block_cpu_limit)(virtual_block_net_limit)(block_cpu_limit)(block_net_limit)(server_version_string)(core_symbol)(core_symbol_precision) )
 FC_REFLECT(eosio::chain_apis::read_only::get_block_params, (block_num_or_id))
 FC_REFLECT(eosio::chain_apis::read_only::get_block_header_state_params, (block_num_or_id))
 
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -97,6 +97,8 @@ public:
       //string                  recent_slots;
       //double                  participation_rate = 0;
       optional<string>        server_version_string;
+      string                  core_symbol;
+      uint64_t                precision;
    };
    get_info_results get_info(const get_info_params&) const;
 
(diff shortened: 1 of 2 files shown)


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -97,6 +97,8 @@ class read_only {
       //string                  recent_slots;
       //double                  participation_rate = 0;
       optional<string>        server_version_string;
+      string                  core_symbol;
+      uint64_t                precision;
```
jgiszczak: For clarity in JSON, this should be core_symbol_precision.

marlonwilliams: Great idea. This has been updated. Thanks

tbfleming:
> Both eosjs and eosjs-ecc supports different prefixes for public keys (e.g EOSxxx, TLOSxxx, etc), so core support for returning the core symbol can help make multichain support faster and out of the box!

Returning CORE_SYMBOL information in `get_info` is a nice improvement. It doesn't, however, solve the key prefix problem. e.g. most test chains have a system token of either `SYS` or `EOS`, and keep the key prefix as is.

Chains modifying the key prefix is a bug. It creates a gratuitous incompatibility to support a "brand". See #5467 for how we plan to address that bug.

marlonwilliams:
#5467 will never be adopted for users. How keys are handled internally in the EOSIO system is irrelevant to the average user. Every one of us have been issued an "EOSxxx" public key - most will not go through the hassle of replacing this with a PUBxxx key. As a result, the user-facing method of key support must always be supported for backwards compatibility purposes. Most new chains are also maintaining this pattern (PREFIXxxx) public key, so this is a welcoming addition that'll make things simpler for users.

I've been writing software for 20 years - guess what? I personally won't go through the hassle of changing my EOSxxx public keys simply because it pleases developers. Will I be forced to convert to a PUBxxx format or be at risk of losing my EOS?

If it ain't broke, don't fix it.

tbfleming:
No conversion necessary for chains which didn’t mess with the prefix. It will always accept the EOSxxx format as input. On output it will use PUB_K1_xxxx. 1 key, 1 binary format, 2 currently-existing ascii formats.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -97,6 +97,8 @@ class read_only {
       //string                  recent_slots;
       //double                  participation_rate = 0;
       optional<string>        server_version_string;
+      string                  core_symbol;
+      uint64_t                core_symbol_precision;
    };
```
heifner: These need to be optional to support working with diff versions of nodeos and cleos.

heifner:
Please test with different combinations of versions of `cleos` and `nodeos`.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
tbfleming: These add additional fields to JSON output. AFAIK this struct doesn't cross the wire in binary form, so adding fields shouldn't be a problem.

marlonwilliams: Got it. Also, they've been updated to optional. Thanks

marlonwilliams:
Have tested with 1.2.1. Will test with previous minor revisions. Thanks

heifner:
Specifically test with different versions of cleos and nodeos. old nodeos & new cleos. new nodeos & old cleos.

tbfleming:
I'm going to have to apologize on this one. We recognize that wallets and apps need a reliable way to retrieve core symbol info, and this appeared to be the right way to do it. Unfortunately having the API expose these fields creates a problem for a couple of our long-term development thrusts. We want future API versions to not depend on certain contracts being on the chain, such as the system contract or the token contract. We see the need for apps and wallets to conveniently access contract-specific information, for example the core token symbol, so we intend to address this in a future version of the API, but in a more general way. In the mean time, we don't want to continue adding contract-specific information to the existing API.

We're going to have to undo this PR. For the near term, there is a workaround, though it's not ideal. If you call `get_account` on a well-known account that's likely to have a balance, e.g. `eosio` or `eosio.stake`, then `core_liquid_balance` will have the correct symbol. We're striving for a much better solution for this long term.

PR #5476, opened 2018-08-29
Launcher tests

PaulCalabrese:
Add use of the launcher-generated bios_boot.sh as the default for existing tests
Minor modifications to some tests to work with bios_boot.sh (usually buying more RAM)
Modify some tests to use the existing bootstrap method (they require partial bootstrapping)
Add a new launcher test that tests launcher options such as --bounce and --down


programs/eosio-launcher/main.cpp
```diff
@@ -1650,14 +1651,30 @@ launcher_def::bounce (const string& node_numbers) {
...
+      }
+      else {
+         string cmd = "./scripts/eosio-tn_bounce.sh " + eosd_extra_args;
+         bp::child c(cmd,
+                     bp::env["EOSIO_HOME"] = host.eosio_home,
+                     bp::env["EOSIO_NODE"] = node_num );
+
+         if(!c.running()) {
+           cerr << "child not running after spawn " << cmd << endl;
+           for (int i = 0; i > 0; i++) {
```
brianjohnson5972: We really want this tight loop, with no delay, running till i rolls over to negative?

programs/eosio-launcher/main.cpp
```diff
@@ -1650,14 +1651,30 @@ launcher_def::bounce (const string& node_numbers) {
...
+            exit (-1);
+         }
+      }
+      else {
+         string cmd = "./scripts/eosio-tn_bounce.sh " + eosd_extra_args;
+         bp::child c(cmd,
+                     bp::env["EOSIO_HOME"] = host.eosio_home,
+                     bp::env["EOSIO_NODE"] = node_num );
+
+         if(!c.running()) {
```
brianjohnson5972: Can we make the following lines a method to use here and below?

scripts/eosio-tn_up.sh
```diff
@@ -33,8 +35,8 @@ rm $datadir/stderr.txt
 ln -s $log $datadir/stderr.txt
 
 relaunch() {
-    echo "$rundir/$prog $* --data-dir $datadir --config-dir etc/eosio/node_$EOSIO_NODE > $datadir/stdout.txt  2>> $datadir/$log "
-    nohup $rundir/$prog $* --data-dir $datadir --config-dir etc/eosio/node_$EOSIO_NODE > $datadir/stdout.txt  2>> $datadir/$log &
+    echo "$rundir/$prog $qargs $* --data-dir $datadir --config-dir etc/eosio/node_$EOSIO_NODE > $datadir/stdout.txt  2>> $datadir/$log "
```
brianjohnson5972: Does this not run into the same issues with  the * exploding to all the files in the current directory?

PaulCalabrese: No because I have added quotes around the * argument. This is not a very general solution but works for --filter-on *.

programs/eosio-launcher/main.cpp
PaulCalabrese: A lot of this code is copy-pasted from elsewhere in the launcher. I will simplify it and pull it out into a function.

PR #5479, opened 2018-08-29
Trivial wallet directory locking for keosd

spoonincode:
Really Bad Things can happen if multiple keosd are using the same wallet directory simultaneously. Like, loss of keys. Make a trivial locking mechanism so that this is prevented.

Of course, this doesn't protect against pre 1.3 keosd, but so far not really encountered any great ways to overcome that limitation.

Yes, I'm aware of boost:ipc:scoped_lock but I had trouble with it. This is simple enough.


spoonincode:
Replaced with #5495

Issue #5380, opened 2018-08-22
Feature request: Provide table scope index for smart-contracts


jcalfee:
In some cases developers are choosing not to use the table scope because it lacks an iterator or a lookup function.  Seems like that is not a security concern.  If not, if that were added I think we would get contracts that are a little easier to query (get_table_rows contract >>scope<<) and make use of that built-in index.

Kevin
>It acts kind of like a sub-table name we have no way to iterate over tables either. Something to consider.

James
:+1:

gravitt8460:
This would be greatly beneficial. I very rarely use the scope feature because it prevents me from getting all records. Additional filtering capabilities would also be highly beneficial.

jcalfee:
I guess there could be a scalability or performance concerns.  No guarantee that you can do everything within the CPU limit for the block so that may need in-contract paging.  Still probably better to have it than to find out after a contract deployed that this is not possible.

tbfleming:
Internally, `table_id_multi_index` has the `by_code_scope_table` index. We could expose `lower_bound`, `find`, and iteration on this index to contracts in a future hard fork.

tbfleming:
`by_code_table_scope` would probably be more convenient, but since it doesn't already exist it would increase row ram costs.

taokayan:
scope can be useful when you need to store something between 2 different accounts. For example, delegatebw uses scope as delegator and key as the account receiving the resources. But I agree there's no way to iterate over a scope.

taokayan:
Please follow https://github.com/EOSIO/eos/pull/5486

spartucus:
Any progress on this?

PR #5486, opened 2018-08-30
5380 support iteration over scopes & tables in cleos

taokayan:
#5380  This allow user to off-chain iterate over scopes and tables for a given contract. 

Adds new `/v1/chain/get_table_by_scope` rpc.

Usage: ./cleos get scope [OPTIONS] contract

Positionals:
  contract TEXT               The contract who owns the table (required)

Options:
  -t,--table TEXT             The name of the table as filter
  -l,--limit UINT             The maximum number of rows to return
  -L,--lower TEXT             lower bound of scope
  -U,--upper TEXT             upper bound of scope

for example: to iterate over all accounts that has EOS balance table:
```
./cleos get scope eosio.token -t accounts -l 4
{
  "rows": [{
      "code": "eosio.token",
      "scope": "a11111111111",
      "table": "accounts",
      "payer": "eosio",
      "count": 1
    },{
      "code": "eosio.token",
      "scope": "a123",
      "table": "accounts",
      "payer": "eosio",
      "count": 1
    },{
      "code": "eosio.token",
      "scope": "a22222222222",
      "table": "accounts",
      "payer": "eosio",
      "count": 1
    },{
      "code": "eosio.token",
      "scope": "a33333333333",
      "table": "accounts",
      "payer": "eosio",
      "count": 1
    }
  ],
  "more": true
}
```


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -250,6 +250,27 @@ class read_only {
...
+   };
+   struct get_table_by_scope_result_row {
+      name        code;
+      name        scope;
+      name        table;
+      name        payer;
+      uint32_t    count;
+   };
+   struct get_table_by_scope_result {
+      vector<fc::variant> rows; ///< one row per item, either encoded as hex String or JSON object
```
heifner: Why is this a variant? Also the comment seems wrong.

programs/cleos/main.cpp
```diff
@@ -2011,6 +2011,23 @@ int main( int argc, char** argv ) {
...
+   getScope->add_option( "-l,--limit", limit, localized("The maximum number of rows to return") );
+   getScope->add_option( "-L,--lower", lower, localized("lower bound of scope") );
+   getScope->add_option( "-U,--upper", upper, localized("upper bound of scope") );
+   getScope->set_callback([&] {
+      auto result = call(get_table_by_scope_func, fc::mutable_variant_object("code",code)
+                         ("table",table)
+                         ("lower_bound",lower)
+                         ("upper_bound",upper)
+                         ("limit",limit)
+                         );
```
heifner: It would be nice to have some tests for this and the other get table functionality. Not required for this PR, but something to do when you have some time.

heifner:
New features should target develop branch not release/1.2.x branch

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
taokayan: Will fix this accordingly.

programs/cleos/main.cpp
taokayan: agree. let me make some tests for "get table" related functionality.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -262,6 +262,27 @@ class read_only {
...
+   struct get_table_by_scope_result_row {
+      name        code;
+      name        scope;
+      name        table;
+      name        payer;
+      uint32_t    count;
+   };
+   struct get_table_by_scope_result {
+      vector<get_table_by_scope_result_row> rows;
+      bool                more = false; ///< true if last element in data is not the end and sizeof data() < limit
```
heifner: We want to move away from true/false for more. Instead fill it with the next value if there is more. See for example, get_producers or get_scheduled_transactions. Would be nice to also make that change for get table row as well.

kasperfish: why not {next : "nextitem", more: 12345} or is there an other way to get the length of a table? I got referenced to this issue from here https://github.com/EOSIO/eos/issues/5543

taokayan: @kasperfish there's no way to know the remaining number unless you iterate all of them.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1106,6 +1106,48 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
 }
 
+read_only::get_table_by_scope_result read_only::get_table_by_scope( const read_only::get_table_by_scope_params& p )const {
+   const auto& d = db.db();
+   const auto& idx = d.get_index<chain::table_id_multi_index, chain::by_code_scope_table>();
+   decltype(idx.lower_bound(boost::make_tuple(0, 0, 0))) lower;
+   decltype(idx.upper_bound(boost::make_tuple(0, 0, 0))) upper;
+
+   if (p.lower_bound.size()) {
+      uint64_t scope = convert_to_type<uint64_t>(p.lower_bound, "lower_bound scope");
```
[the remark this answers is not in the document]
taokayan: This is not a bug. The value of scope will be read as uint64_t first, then read as account_name if failed to read as uint64_t, as for different contracts the meaning of scopes may defer. If you want to read it as account_name always, you need to use ```cleos -u https://mainnet.meet.one --print-request get scope eosio.token -t accounts -l 1 -L " 313313313313" ```

noprom:
Is it possible to get a contract scope in contract?

PR #5487, opened 2018-08-30
mysql_db_plugin added.

eosBLACK:
mysql_db_plugin is a plugin that plays the same role as mongo_db_plugin and is a plugin for RDB users of MySQL series.
This plugin helps you to verify the desired data using simple SQL to query the block history.
Also, since I used libzdb, it can be applied to SQLite, Oracle, and PostgreSQL, but this plugin focused only on MySQL.
mysql_db_plugin was created with reference to mongo_db_plugin and sql_db_plugin, which was removed in 1.2.x.
Please reply us after reveiwing the above requests.


heifner:
Thank you for your submission. We are now asking plugins like this to be hosted in your own repo. See https://github.com/EOSIO/eos/pull/5026 for details on building your plugin externally.

PR #5517, opened 2018-09-03
Add account_name related utility functions

xJonathanLEI:
Added three utility functions to both `eosiolib` and the `chain` library:

    // Gets a length * 5 bit mask
    static constexpr uint64_t name_mask(size_t length);
    // Gets part of a name
    static constexpr uint64_t name_mask(size_t length);
    // Gets the length of name (normalized)
    static constexpr size_t name_length( uint64_t n );

Unit tests have been written for these functions.


heifner:
Thanks for your pull request. This has moved to eosio.cdt repo and is no longer needed.

PR #5530, opened 2018-09-04
Make bios-boot-tutorial.py compatible with Python 3.5

10gic:
Underscores in Numeric Literals (PEP 515) is introduced in Python 3.6
However, Python 3.5 is widely used in some platforms. So, remove this
usage.


tbfleming:
I'm uneasy making numbers less readable to support older tools.

10gic:
Script bios-boot-tutorial.py is a convenient tool, it's more convenient if make it compatible with Python 3.5

PR #5524, opened 2018-09-04
5523 fix abi serialization exception in get block

taokayan:
solves #5523 #5519

Issue #5519, opened 2018-09-03
Get Block Info - Error 3015004: The type defined in the ABI is invalid


mawenpeng:
Got similar issues to issue #4915.
EOS v1.2.2, CentOS Linux 7.5.
Updated to EOS v1.2.3, issues were the same.

[root@bc-nodes ~]# cleos --url https://api.eosnewyork.io get block 13990004
Error 3015004: The type defined in the ABI is invalid
Error Details:
Failed to serialize type
[root@bc-nodes ~]# cleos --url https://api.eosnewyork.io get block 13990026
Error 3015004: The type defined in the ABI is invalid
Error Details:
Failed to serialize type
[root@bc-nodes ~]# cleos -u http://127.0.0.1:8888 get block 13990004
Error 3015004: The type defined in the ABI is invalid
[root@bc-nodes ~]# cleos -u http://127.0.0.1:8888 get block 13990026
Error 3015004: The type defined in the ABI is invalid

config.ini plugins setup:
plugin = eosio::chain_api_plugin
plugin = eosio::net_api_plugin
plugin = eosio::history_plugin
plugin = eosio::chain_plugin
plugin = eosio::http_plugin
plugin = eosio::history_api_plugin
plugin = eosio::mongo_db_plugin

Removed history_plugin, history_api_plugin, mongo_db_plugin, issues were the same.

xJonathanLEI:
I believe I've found the cause. Although the exact reason is still unknown.

This guy [danakilblock](https://eosflare.io/account/danakilblock) is putting files onto the blockchain, and  `nodeos` obvisouly has trouble deserializing them:

https://eosflare.io/tx/e9d2410c58362cc7ecced45b6c52dc878f44efe83feef4cac7a95800ceefbdf7

https://eosflare.io/tx/eb2b41dac816f64cd6fd0699ab14e17ac69b2dcd4f118bd5a3341e40e2200bcc

firesWu:
I think the reason is danakilblock update contract and abi haven't include previous action.

taokayan:
Looks like some bug in the ABI serilazation, or timeout happened.

eluzgin:
Reproduced on our API nodes as well. 
Retrieved actual error and block from the log - attached here!
[bad_block.txt](https://github.com/EOSIO/eos/files/2347209/bad_block.txt)

PR #5548, opened 2018-09-05
fix abi serialization exception in get block

### LINKED ISSUE #5523: invalid_type_inside_abi on block 14363575
It's impossible to read block 14363575 over HTTP client, all eos nodes fail on that request. I have tested this on local EOSIO 1.2.3 node and also on several public nodes.

curl -X POST -d '{"block_num_or_id":14363575}' https://api.eosnewyork.io/v1/chain/get_block

heifner:
Resolves #5523 #5519
Replaces #5524


firesWu:
#5648

PR #5602, opened 2018-09-11
Correct assert message by including delay_max_limit_ms

baegjae:
In authorization manager, `delay_max_limit_ms` is defined to print `delay_max_limit`. 
(#5563 Thanks for explanation)
However, `delay_max_limit_ms` was not included in the assert message.
As intended, `delay_max_limit_ms` is included in the assert message.


heifner:
Replaced by #5676 rebased to develop.

PR #5603, opened 2018-09-11
Remove doxygen warning

EugeneChung:


wanderingbort:
@larryk85  @dskvr are these already represented in the CDT?

Issue #5589, opened 2018-09-10
cleos get transaction <txhash A>, but the result is  the content of txhash B?


guaiguaihw:
![image](https://user-images.githubusercontent.com/11972366/45285291-6d731780-b515-11e8-9c2b-8d3304aa38b9.png)
Our fullnode is connected to mainnet.  Why could this happen ?

arhag:
Resolved by #5723.

taokayan:
Thanks @arhag

PR #5607, opened 2018-09-11
fix prefix comparison in get_transaction in history_plugin

taokayan:
this fixes issue #5589 
compare hex transaction id with adaptive prefix length ranging from 8 bytes(short id) to 64 bytes(full id)


arhag:
Replaced by #5723.

PR #5617, opened 2018-09-11
Fix unchecked unapplied transaction growth on relays

wanderingbort:
Unapplied transactions were only being pruned in a few instances:
- when an HTTP RPC request was made to the node, the node would persistently attempt to apply that transaction until it was done OR expired, cleaning it up appropriately
- when a Producer attempts to produce a block, it would apply as many unapplied transactions as it could, cleaning them up appropriately

This means that relay nodes which never produce would also never prune the unapplied list.

We now have 4 phases:

1. [If not a producing node] prune all unapplied transactions that are not "persisted" due to being directly received over HTTP RPC
1. [If potentially producing] prune all unapplied transactions that expired
1. [If not "read only"] attempt application of all unapplied transactions that are "persisted" due to being directly received over HTTP RPC
1. [If producing the next block] attempt application of all remaining unapplied transactions

Each of these phases potentially prunes.  Phases 1 & 2 are the new phases that cover P2P relays

Additionally, Read-only nodes were never pruning the unapplied transactions.  As they will never have need of them, nodes configured in this mode will unilaterally drop all unapplied transactions

However, a p2p relay node neither producers no accepts the transactions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -80,7 +80,22 @@ using transaction_id_with_expiry_index = multi_index_container<
    >
 >;
 
+struct lazy_unapplied_transactions {
+   lazy_unapplied_transactions (const controller& _controller)
+   : _controller(_controller)
+   {
+   }
 
+   vector<transaction_metadata_ptr>& get() {
+      if (!_data) {
+         _data.emplace(_controller.get_unapplied_transactions());
+      }
+      return *_data;
+   }
+
+   optional<vector<transaction_metadata_ptr>>   _data;
+   const  controller&                           _controller;
+};
 
 enum class pending_block_mode {
    producing,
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -995,30 +1029,27 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block(bool
...
+                     std::swap(unapplied_trxs.at(idx), unapplied_trxs.back());
+                     unapplied_trxs.pop_back();
+                  } else {
+                     ++idx;
                   }
                }
             }
 
+            // FINAL STEP FOUR if we are producing, attempt to apply unapplied transactions
+            // if this becomes the none final step it will need to prune the local cache of unapplied transactions
```
gleehokie: What does “if this becomes the none final step” mean?

wanderingbort:
During review and refactor a lot of things changed.  @arhag correctly identified that the "read-only" optimization was redundant and we ended up with (hopefully) more readable and understandable code a result. 

the code combines a category on the transaction with the current state of the node/block in order to determine the final operation: apply, drop, no-op.

winlin:
is there any article talk about the `read-only` mode except for the release onte @wanderingbort

Issue #5268, opened 2018-08-15
Allow for trusted producers for nodeos


wanderingbort:
Propagation delay has become a problem on networks and with modern producers running entire networks of nodes to provide resilience and high availability the number of hops a block must traverse to even leave the trusted network may be > 1.  

As a result, we fielded a request from BPs to add the ability to implicitly trust blocks from a given producer.  

This would degrade validation for those blocks to `light` for any block signed by an implicitly trusted producer.

The intended use case is that a logical producer may want to implicitly trust their own blocks throughout their whole network of nodes in order to facilitate faster propagation/validation.

brianjohnson5972:
We want to have a configuration parameter to indicate a trusted-producer, and once we validate that producer's signature for a block, then we will behave exactly like my->conf.block_validation_mode == validation_mode::LIGHT (until we are done with that block).

arhag:
Resolved by #5631.

PR #5631, opened 2018-09-12
Implement Trusted Producer Light Validation

brianjohnson5972:
#5268 
Added configuration options to set trusted producers in chain_plugin.
In controller_impl::push_block added setting (and resetting on exit) a flag that provides the same behavior as light validation.
Made changes to validating_tester to allow calling the main chain and then calling the validating_node separately.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1020,8 +1020,8 @@ struct controller_impl {
    void push_block( const signed_block_ptr& b, controller::block_status s ) {
       EOS_ASSERT(!pending, block_validate_exception, "it is not valid to push a block when there is a pending block");
 
-      auto reset_prod_light_validation = fc::make_scoped_exit([this]() {
-         conf.trusted_producer_light_validation = false;
+      auto reset_prod_light_validation = fc::make_scoped_exit([old_value=conf.trusted_producer_light_validation, this]() {
+         conf.trusted_producer_light_validation = old_value;
       });
       try {
          EOS_ASSERT( b, block_validate_exception, "trying to push empty block" );
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1018,14 +1018,20 @@ struct controller_impl {
 
 
    void push_block( const signed_block_ptr& b, controller::block_status s ) {
-    //  idump((fc::json::to_pretty_string(*b)));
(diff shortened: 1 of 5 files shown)


libraries/chain/controller.cpp
```diff
@@ -1018,14 +1018,20 @@ struct controller_impl {
...
+      });
       try {
          EOS_ASSERT( b, block_validate_exception, "trying to push empty block" );
          EOS_ASSERT( s != controller::block_status::incomplete, block_validate_exception, "invalid block status for a completed block" );
          emit( self.pre_accepted_block, b );
          bool trust = !conf.force_all_checks && (s == controller::block_status::irreversible || s == controller::block_status::validated);
          auto new_header_state = fork_db.add( b, trust );
+         if (conf.trusted_producers.count(b->producer)) {
+            conf.trusted_producer_light_validation = true;
+         };
```
heifner: Instead of setting a flag, I wonder if we could just add this check in `light_validation_allowed` then the extra flag and reset would not be needed.

brianjohnson5972: I think that would be much more complicated (handling identifying if we are speculating or applying a block) and this gets called multiple times in a transaction pass, versus being called once for the block.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -74,6 +74,8 @@ namespace eosio { namespace chain {
             validation_mode          block_validation_mode  = validation_mode::FULL;
 
             flat_set<account_name>   resource_greylist;
+            flat_set<account_name>   trusted_producers;
+            bool                     trusted_producer_light_validation = false;
```
wanderingbort: I dont really like this "implementation detail" exposed in the `config` structure (The `bool`,  the `flat_set` is appropriate here).  There doesn't seem to be a reason why an external system would ever need to know about this.  I think I would prefer that be part of the `controller_impl` class instead alongside `in_trx_requiring_checks`

wanderingbort: also `trusted_producers` should be added to the reflection of this structure: https://github.com/EOSIO/eos/blob/c7a5a9e9e8c775cb125571db484aef8db9b71aba/libraries/chain/include/eosio/chain/controller.hpp#L294

Issue #5600, opened 2018-09-10
ABI Format: binary extensions


tbfleming:
This supports extending existing binary formats; e.g. adding new fields to tables.

Caution: binary extensions only operate correctly in certain locations.

* safe: a non-embedded struct stored in a row may have binary extensions at its end
* safe: an action may have use binary extensions to add additional arguments to its end
* unsafe: a struct with binary extensions may not be used inside an array or another struct
* unsafe: a struct with binary extensions may not be used as a base of another struct
* unsafe: fields with types which don't end in `$` following fields with types which do
* unsafe: `$` used anywhere except in struct field types

Note that there are, in fact, safe special cases within some of the unsafe cases above. The initial release treats most of them as unsafe.

## ABI version string

`eosio::abi/1.1`

## ABI Text format

Types may have a `$` suffix. During binary-to-json conversion, fields with a `$` type don't error out when end-of-data has been reached; instead they're omitted. During json-to-binary conversion, missing fields don't error out as long as no non-missing fields follow in the ABI. This omits the bytes from the output stream.

e.g.

```json
    {
      "name": "my_table_struct",
      "base": "",
      "fields": [
        {
          "name": "required_field_1",
          "type": "string"
        },
        {
          "name": "required_field_2",
          "type": "float32[]"
        },
        {
          "name": "optional_field_3",
          "type": "float32[]$"
        },
        {
          "name": "optional_field_4",
          "type": "string$"
        },
      ]
    },
```

## JSON representation

Missing fields aren't included; null isn't used. E.g. all of these are valid JSON representations of `my_table_struct`:

```json
{
    "required_field_1": "foo",
    "required_field_2": [1,2,3,4]
}
```

```json
{
    "required_field_1": "foo",
    "required_field_2": [1,2,3,4],
    "optional_field_3": [5,6,7,8]
}
```

```json
{
    "required_field_1": "foo",
    "required_field_2": [1,2,3,4],
    "optional_field_3": [5,6,7,8],
    "optional_field_4": "bar"
}
```

## ABI Binary format

`$` can be included in type strings. No other changes.

## eosiolib

A new type tag (TBD) will identify these types. When an extension field is missing during deserialization, it will be left default-constructed.

Issue #5599, opened 2018-09-10
ABI Format: variants


tbfleming:
This is for serializing `std::variant` or `fc::static_variant`.

## JSON representation

```
[type, value]
```

`type` is a string identifying the type. `value` is the content.

## Binary representation

```
varuint
value
```

The varuint identifies type. 0 is the first type in the `variant_def`.

## ABI version string

`eosio::abi/1.1`

## ABI Text format

New section at the same level as `types`, `structs`, etc. named `variants`. Example:

```json
   "variants": [
      {
         "name": "operation_amount",
         "types": [
            "extended_asset",
            "float64",
            "time_point_sec"
         ]
      },
      {
         "name": "operation_subaccount",
         "types": [
            "name",
            "extended_symbol",
            "symbol_code"
         ]
      },
      {
         "name": "subop",
         "types": [
            "event",
            "balance_change",
            "set_value",
            "error"
         ]
      }
   ],
```

## ABI Binary format

The new definitions will be after `abi_extensions`. Including it within `abi_extensions` would break the abi's ability to describe itself. See https://github.com/EOSIO/eosjs2/blob/master/src/abi.abi.json for the current abi abi (doesn't support variants). The new part of the abi abi will be:

```

    {
      "name": "abi_def",
      "base": "",
      "fields": [
        ...
        {
          "name": "abi_extensions",
          "type": "extensions_entry[]"
        },
        {
          "name": "variants",
          "type": "variant_def[]$"
        }
      ]
    },
    {
      "name": "variant_def",
      "base": "",
      "fields": [
        {
          "name": "name",
          "type": "string"
        },
        {
          "name": "types",
          "type": "string[]"
        }
      ]
    }
```

The `$` suffix is a strawman for a new type of optional that supports format extensions. See #5600

## Prototype

A partial implementation lives in https://github.com/EOSIO/eosjs2/tree/exchange . It supports `variant_def` in json-format ABI files and can do binary-to-json conversions using it. json-to-binary doesn't support it yet and eosjs2 doesn't support it in binary-format ABI files yet.

PR #5652, opened 2018-09-13
ABI 1.1

tbfleming:
ABI 1.1

The ABI serializer supports 2 new features plus a bug fix:
* Variants: These are values which support multiple types. See #5599 for details.
* Binary extensions: These support extending existing rows and actions with new fields. See #5600 for details.
* It now checks the version string to verify compatibility.

Watch for ABI 1.1 support in upcoming releases of `eosio.cdt`, `abieos`, and `eosjs2`.


fproulx-eoscanada:
@tbfleming am I missing something, or this change in `1.3.0` (which I can now test on Kylin) is a breaking change that makes, for instance, `cleos get table` not work anymore for existing ABIs on chain that omitted specifying the version (`version: ""`). There are tons of tons on the mainnet right now. It looks like I have no way to `get table` on those accounts if I talk to the API of a full node is `1.3.0`

Is there a way for `cleos` to be made backward compatible or you need to release `1.3.1` to relax this?

tbfleming:
We scanned for accounts like that and only found a couple.

PR #5670, opened 2018-09-14
Add filter-on ability to accept blank action as wildcard

heifner:
- Update mongo_db_plugin `mongodb-filter-on` & `mongodb-filter-out` with ability to wildcard action & receiver.
  -- Now receiver, action, and actor can be wildcarded.
- This updates mongo_db_plugin with latest filter capability of history plugin
  -- See #4739 
- Includes changes from #5691. Thanks go to @venediktov for implementation.

### DIFF
--- a/plugins/mongo_db_plugin/mongo_db_plugin.cpp
+++ b/plugins/mongo_db_plugin/mongo_db_plugin.cpp
@@ -214,11 +214,14 @@ const std::string mongo_db_plugin_impl::account_controls_col = "account_controls
 
 bool mongo_db_plugin_impl::filter_include( const chain::action_trace& action_trace ) const {
    bool include = false;
-   if( filter_on_star || filter_on.find( {action_trace.receipt.receiver, action_trace.act.name, 0} ) != filter_on.end() ) {
+   if( filter_on_star ||
+       filter_on.find( {action_trace.receipt.receiver, 0, 0} ) != filter_on.end() ||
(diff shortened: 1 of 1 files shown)


plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -1421,9 +1423,9 @@ void mongo_db_plugin::set_program_options(options_description& cli, options_desc
          ("mongodb-store-action-traces", bpo::value<bool>()->default_value(true),
           "Enables storing action traces in mongodb.")
          ("mongodb-filter-on", bpo::value<vector<string>>()->composing(),
-          "Mongodb: Track actions which match receiver:action:actor. Actor may be blank to include all. Receiver and Action may not be blank. Default is * include everything.")
+          "Mongodb: Track actions which match receiver:action:actor. Actor may be blank to include all. Action and Actor both blank allows all from Receiver. Receiver may not be blank.")
```
wanderingbort: is my read of the code wrong or does this not also allow `reciever:0:actor` which is not covered in the output?

heifner: How about:

"Mongodb: Track actions which match receiver:action:actor. Action and/or Actor may be blank to include all. Action and Actor both blank allows all from Receiver. Receiver may not be blank."

wanderingbort: I'll buy it

PR #5691, opened 2018-09-17
fix mongo_db_plugin filter deficiency to allow any combination of receiver,action,actor

venediktov:
@heifner , @arhag in the current  implementation of mongo_db_filters only subset of permutations of `{receiver, action,actor}` is allowed which makes it difficult to fine tune what gets written to backend store .

In our case we were looking to do `--mongodb-filter-on ":issue:" --mongodb-filter-on ":issue4snap:" --mongodb-filter-on ":transfer:`

The format of the filter is `receiver:action:actor` and current implementation only allows `receiver:action:`  but does not allow omitting  `receiver` like `:action:actor` or `:action:` ( for all actors)

### DIFF
--- a/plugins/mongo_db_plugin/mongo_db_plugin.cpp
+++ b/plugins/mongo_db_plugin/mongo_db_plugin.cpp
@@ -53,11 +53,8 @@ struct filter_entry {
    name receiver;
    name action;
    name actor;
-   std::tuple<name, name, name> key() const {
-      return std::make_tuple(receiver, action, actor);
-   }
    friend bool operator<( const filter_entry& a, const filter_entry& b ) {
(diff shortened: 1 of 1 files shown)


venediktov:
Thank you  @heifner ! It looks like if we merge with PR #5670 we need to change `--mongodb-filter-on` option description as to something like
> Any combination of `"receiver:0:actor"` , `"0:action:actor"` , `"0:0:actor"` , `"0:action:actor"` , `"0:action:0"` including `"0:0:0"` which can also be denoted as simply passing `"*"` to this option

wanderingbort:
Closing in favor of consolidating work into #5670

PR #5703, opened 2018-09-17
add installs for libtester

larryk85:
Needed to add some libs for libtester to compile correctly with eosio.contracts.


spoonincode:
Instead of the changes to wabt submodule, how about disabling the tools like is done here for tests
https://github.com/EOSIO/eos/blob/45e609fe73050d9c648ff44962c3178458e817f0/libraries/CMakeLists.txt#L13-L15

PR #5710, opened 2018-09-18
Final keosd unix socket changes

spoonincode:
This PR consists of the 3 final keosd unix socket changes.

1. keosd no longer listens on 127.0.0.1:8900 by default
1. cleos' default wallet_url is the keosd socket created at ~/eosio-wallet/keosd.sock
1. cleos' keosd auto launch has been modified to strictly use the new unix socket at the above path. Additionally, keosd will be launched such that its http and https server are disabled.

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -166,7 +166,7 @@ bfs::path determine_home_directory()
 }
 
 string url = "http://127.0.0.1:8888/";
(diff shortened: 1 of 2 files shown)


spoonincode:
FWIW, I did verify that the "docker quick start instructions" still work. Those instructions launch keosd with --http-server-address so not really affected by change at all.

PR #5725, opened 2018-09-18
Transaction trace logging for visibility of transactions on P2P

wanderingbort:
In conversations with various block producers, it became apparent that there was no good visibility on how `nodeos` was propagating transactions (or not).  This PR implements some basic transaction tracing intended on a single node in a separate log with the identifier "transaction_tracing" as well as some other basic statistics in the producer plugin specific log.

The goal is that collaborative efforts to aggregate and analyze logs can trace a transaction as it is relayed through the P2P network and eventually finds a place in a block or determine why/when it is unable to achieve that end.


abourget:
Will that slow down normal operations, or are the serialization of all those logs disabled when the option isn't enabled?

PR #5727, opened 2018-09-18
binary releases of cleos and keosd

larryk85:
Please note this will generate a dmg with cleos and keosd on Mac OSX, and a full install for DEB and RPM packages.  Without changing all of the install cmake commands to disable during installation when using CPack generation, the work around is that it installs cleos and keosd to what CPack uses for it's default bin path (/usr/bin for both it seems) and everything else to /usr/local/eosio like normal.  This at least doesn't expose the other binaries unless someone adds /usr/local/eosio to their path.  sudo apt install <deb name> and sudo apt remove eos.io work cleanly.  And yum installs and uninstalls cleanly too.


debian/postinst
```diff
@@ -36,6 +36,10 @@ case "$1" in
       adduser --no-create-home --group --system ${USER}
     fi
     set -e
+    touch /var/log/${PACKAGE}
```
jgiszczak: All references to `/var/log`, `/var/lib`, and `/etc` should be eliminated.  None of those directories are used by the software by default or explicitly by the documentation.  Instead, that information is all kept in `~/.local/share/eosio`.  The old packages needed to install a `genesis.json` file, which is also no longer required.

With the elimination of installation to both `/var` and `/etc`, it's likely that usage of `CMAKE_INSTALL_FULL_*` is no longer required for any component.

spoonincode: Not sure I follow what you expect.
```
message(WARNING ${CMAKE_INSTALL_FULL_BINDIR})
message(WARNING ${CMAKE_INSTALL_BINDIR})
```

This gives
```
CMake Warning at CMakeLists.txt:12 (message):
  /usr/local/bin
CMake Warning at CMakeLists.txt:13 (message):
  bin
```

larryk85:
Superseded by #6003

Issue #5674, opened 2018-09-14
Improve Integration Test Diagnostics


brianjohnson5972:
There is a fairly consistent error where a transaction cannot be retrieved from a node.  Adding diagnostics to retrieve blocks and print out to better understand what is happening.

PR #5752, opened 2018-09-19
Report Transaction Information in Integration Tests

brianjohnson5972:
#5674 
Added reporting transaction information when a transaction is created in the integration tests.


brianjohnson5972:
Had wrong commit branch

PR #5754, opened 2018-09-19
Make wabt the default wasm runtime

spoonincode:
As mentioned in the 1.3 release notes, wabt will be the default wasm runtime in 1.4+


wanderingbort:
@dskvr @sergmetelin this will be something we want to make sure we communicate very well in the next release

PR #5763, opened 2018-09-20
add duplicate check for set code/abi/contract

### LINKED ISSUE #5748: cleos set contract does not check if ABI has been changed
There an issue with cleos regrading to deploy contracts.

The issue is if you changed contract code, you cannot deploy it with cleos set contract in some cases. 

What actually happened is cleos set contract will check if the contract is same. If so it will preventing deploy the contract again. But it _does not check if the ABI of contract has been updated_. 

The problem is if I’m developing a contract and I will modify interface along the way. it will be very confused to be told that the contract code is same, but actually I modified the interface has been change. A different interface opens up all kind of problems with eosjs integration as well.

I am aware that people can work around this issue by adding few dummy code in their contract or use cleos set abi.  But in general a two steps to deploy is very annoying.

But since cleos set contract is suppose to do what both set abi and set code does, I would say it make sense that cleos set contract also check if ABI has changed.

taokayan:
fix #5748 
this will prevent uploading the same code/abi into nodeos. This is good for "set contract" as users don't need to check whether wasm or abi has actually changed or not.

for example, if only abi is changed but not wasm:

```
./cleos set contract a123 ./contracts/hello/
Reading WASM from ./contracts/hello/hello.wasm...
Skipping set code because the new code is the same as the existing code
Publishing contract...
executed transaction: cb68c2ff6b3327c46859b200dc41722478cf017b32c58ccd8f2c90e33bf0ee97  2296 bytes  771 us
#         eosio <= eosio::setabi                {"account":"a123","abi":"0e656f73696f3a3a6162692f312e30000204646174610003036b65790675696e743634046b6...
warning: transaction executed locally, but may not be confirmed by the network yet    ] 
```

or if wasm & abi are unchanged:
```
./cleos set contract a123 ./contracts/hello/
Reading WASM from ./contracts/hello/hello.wasm...
Skipping set code because the new code is the same as the existing code
Skipping set abi because the new abi is the same as the existing abi
```

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -2333,12 +2333,14 @@ int main( int argc, char** argv ) {
 
       std::vector<char> old_wasm;
       bool duplicate = false;
+      fc::sha256 old_hash, new_hash;
       if (!suppress_duplicate_check) {
          try {
-            const auto result = call(get_raw_code_and_abi_func, fc::mutable_variant_object("account_name", account));
-            old_wasm = result["wasm"].as_blob().data;
+            const auto result = call(get_code_hash_func, fc::mutable_variant_object("account_name", account));
+            old_hash = fc::sha256(result["code_hash"].as_string());
          } catch (...) {
-            std::cout << "Failed to get old contract, continue without duplicate check..." << std::endl;
+            std::cerr << "Failed to get existing code hash, continue without duplicate check..." << std::endl;
+            suppress_duplicate_check = true;
          }
       }
 
@@ -2362,12 +2364,15 @@ int main( int argc, char** argv ) {
         if(wasm.compare(0, 8, binary_wasm_header))
            std::cerr << localized("WARNING: ") << wasmPath << localized(" doesn't look like a binary WASM file. Is it something else, like WAST? Trying anyways...") << std::endl;
         code_bytes = bytes(wasm.begin(), wasm.end());
-
-        if (code_bytes.size() > 0 && code_bytes.size() == old_wasm.size()) {
-           duplicate = (memcmp(&(code_bytes[0]), &(old_wasm[0]), code_bytes.size()) == 0);
-        }
       } else {
         code_bytes = bytes();
+      }
+
+      if (!suppress_duplicate_check) {
+         if (code_bytes.size()) {
+            new_hash = fc::sha256::hash(&(code_bytes[0]), code_bytes.size());
+         }
+         duplicate = (old_hash == new_hash);
       }
 
       if (!duplicate) {
@@ -2387,14 +2392,11 @@ int main( int argc, char** argv ) {
       bool duplicate = false;
       if (!suppress_duplicate_check) {
          try {
-            const auto result = call(get_raw_code_and_abi_func, fc::mutable_variant_object("account_name", account));
-            const std::vector<char> abi_v = result["abi"].as_blob().data;
-            abi_def abi_d;
-            if (abi_serializer::to_abi(abi_v, abi_d)) {
-               old_abi = fc::raw::pack(abi_d);
-            }
+            const auto result = call(get_raw_abi_func, fc::mutable_variant_object("account_name", account));
+            old_abi = result["abi"].as_blob().data;
          } catch (...) {
-            std::cout << "Failed to get old contract, continue without duplicate check..." << std::endl;
+            std::cerr << "Failed to get existing raw abi, continue without duplicate check..." << std::endl;
+            suppress_duplicate_check = true;
          }
       }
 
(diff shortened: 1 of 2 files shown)


spoonincode:
You could reduce the amount of network chatter by
1. Using get_code_hash for checking code uniqueness
1. Using get_raw_abi to just get the abi to check for uniqueness

programs/cleos/main.cpp
```diff
@@ -2347,19 +2363,41 @@ int main( int argc, char** argv ) {
            std::cerr << localized("WARNING: ") << wasmPath << localized(" doesn't look like a binary WASM file. Is it something else, like WAST? Trying anyways...") << std::endl;
         code_bytes = bytes(wasm.begin(), wasm.end());
 
+        if (code_bytes.size() > 0 && code_bytes.size() == old_wasm.size()) {
+           duplicate = (memcmp(&(code_bytes[0]), &(old_wasm[0]), code_bytes.size()) == 0);
```
wanderingbort: I think I'd prefer to move us away from lexicographical comparison for equality.   what do you think about using `duplicate = std::equal(code_bytes.begin(), code_bytes.end(), old_wasm.begin())`

programs/cleos/main.cpp
```diff
@@ -2347,19 +2363,41 @@ int main( int argc, char** argv ) {
...
+        }
       } else {
         code_bytes = bytes();
```
wanderingbort: should we not also check to make sure that when clearing we are not "duplicating" an already empty wasm?

programs/cleos/main.cpp
```diff
@@ -2347,19 +2363,41 @@ int main( int argc, char** argv ) {
...
    auto set_abi_callback = [&]() {
+
+      bytes old_abi;
+      bool duplicate = false;
+      if (!suppress_duplicate_check) {
+         try {
+            const auto result = call(get_raw_code_and_abi_func, fc::mutable_variant_object("account_name", account));
+            const std::vector<char> abi_v = result["abi"].as_blob().data;
+            abi_def abi_d;
+            if (abi_serializer::to_abi(abi_v, abi_d)) {
```
wanderingbort: Is there a reason to unpack and instantiate the ABI?  If it is a compliant ABI with our standard then the `abi_v` should === `fc::raw::pack(abi_d)` I would think this is a redundant operation considering the duplicate check below will only match if my assertion holds true which is the only time we are concerned with the contents of the old abi.  

Can we just use `abi_v` as the `old_abi`?

programs/cleos/main.cpp
```diff
@@ -2374,17 +2412,23 @@ int main( int argc, char** argv ) {
         EOS_ASSERT( fc::exists( abiPath ), abi_file_not_found, "no abi file found ${f}", ("f", abiPath)  );
 
         abi_bytes = fc::raw::pack(fc::json::from_file(abiPath).as<abi_def>());
-
+        if (abi_bytes.size() > 0 && abi_bytes.size() == old_abi.size()) {
+           duplicate = (memcmp(&(abi_bytes[0]), &(old_abi[0]), abi_bytes.size()) == 0);
```
wanderingbort: same comment about `std::equal(...)`

programs/cleos/main.cpp
```diff
@@ -2374,17 +2412,23 @@ int main( int argc, char** argv ) {
...
+        }
       } else {
         abi_bytes = bytes();
```
wanderingbort: same comment about checking for "duplicate" clears

programs/cleos/main.cpp
```diff
@@ -2396,8 +2440,10 @@ int main( int argc, char** argv ) {
       shouldSend = false;
       set_code_callback();
       set_abi_callback();
-      std::cerr << localized("Publishing contract...") << std::endl;
-      send_actions(std::move(actions), 10000, packed_transaction::zlib);
+      if (actions.size()) {
+         std::cerr << localized("Publishing contract...") << std::endl;
+         send_actions(std::move(actions), 10000, packed_transaction::zlib);
+      }
```
wanderingbort: I think we should have a final message here if `actions.size()` is 0.  Sure, you should see two "Skipping..." messages but may as well also be explicit about the fact that no transaction is sent if everything is skipped

heifner:
@taokayan Can you address peer review comments please.

programs/cleos/main.cpp
taokayan: sure will fix that accordingly

programs/cleos/main.cpp
taokayan: sure will fix that.

programs/cleos/main.cpp
taokayan: sure will fix that.

wanderingbort:
@dskvr Please note the change in behavior that may be relevant to documentation: now `cleos set contract` create valid transactions if code or abi would be unchanged by the transaction (previously, it would construct a transaction that the chain would reject).

PR #5775, opened 2018-09-21
Hide unaligned warnings unless console prints on

spoonincode:
Hide the contract unaligned parameter warning unless contracts-console is enabled. These warnings are not useful except in contract development.


spoonincode:
> What about the ones in binaryen.hpp?

Will be done in 1.4 not worth messing with it

PR #5798, opened 2018-09-24
Update README for txn_test_gen_plugin 

warrick-eosny:
--http-validate-host=0 is required when launching the non-producer or curl requests will fail. 
keosd no longer adds the default private key to the wallet on creation. Manually add. 
Added link to the demo video. Also removed eosio::wallet_api_plugin as it is no longer used.


spoonincode:
I would like to avoid `http-validate-host=0` in any demos as I kind of see it as weakening security. Instead can you either a) change the curl requests to be http://127.0.0.1, or b) change the http-server-address to be localhost:8888? Either of those should alleviate the need for http-validate-host. Probably option "a" is cleanest in this case but either is fine with me.

warrick-eosny:
Good point, a) seems like the best option.

PR #5800, opened 2018-09-24
 Fix Block Walker For Transactions

brianjohnson5972:
#5674 
Fixed paths where original transaction was not passed through by instead caching the original transaction and looking it up by transaction id.

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -327,14 +327,10 @@ class Node(object):
         return self.isBlockPresent(blockNum, blockType=BlockType.lib)
 
     class BlockWalker:
-        def __init__(self, node, transOrTransId, startBlockNum=None, endBlockNum=None):
-            assert(isinstance(transOrTransId, (str,dict)))
-            if isinstance(transOrTransId, str):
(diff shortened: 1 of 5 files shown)


tests/Node.py
```diff
@@ -326,21 +327,19 @@ def isBlockFinalized(self, blockNum):
...
+            self.transId=transId
             self.node=node
             self.startBlockNum=startBlockNum
             self.endBlockNum=endBlockNum
 
         def walkBlocks(self):
             start=None
             end=None
+            if self.trans is None and self.transId in self.transCache:
+                self.trans=self.transCache[self.transId]
```
jgiszczak: This will throw a `KeyError` exception if `self.transId` is not in `self.transCache`.  It will not return `None`.

brianjohnson5972: I need self.transId in self.transCache.keys(), right?

jgiszczak: Use `self.transCache.get(self.transId)`.  That returns `None` if the key is not present in the dictionary.

PR #5816, opened 2018-09-26
remove binaryen runtime from nodeos

spoonincode:
Other binaryen components (like eosio-s2wasm) need to remain around until eosiocpp is removed


libraries/testing/tester.cpp
```diff
@@ -96,14 +96,10 @@ namespace eosio { namespace testing {
...
 
       for(int i = 0; i < boost::unit_test::framework::master_test_suite().argc; ++i) {
-         if(boost::unit_test::framework::master_test_suite().argv[i] == std::string("--binaryen"))
-            cfg.wasm_runtime = chain::wasm_interface::vm_type::binaryen;
-         else if(boost::unit_test::framework::master_test_suite().argv[i] == std::string("--wavm"))
+         if(boost::unit_test::framework::master_test_suite().argv[i] == std::string("--wavm"))
             cfg.wasm_runtime = chain::wasm_interface::vm_type::wavm;
          else if(boost::unit_test::framework::master_test_suite().argv[i] == std::string("--wabt"))
             cfg.wasm_runtime = chain::wasm_interface::vm_type::wabt;
-         else
```
jgiszczak: Does it make sense to pick one of the remaining options as default?

spoonincode: Not sure the origin or reason for that line. A `controller::config` instance gets the default automatically
https://github.com/EOSIO/eos/blob/0f6695cb8d31e3d2f318d1720e9d9ad23e1ce80d/libraries/chain/include/eosio/chain/controller.hpp#L71

Issue #5825, opened 2018-09-26
--disable-replay-opts leads to undo stack error


spoonincode:
When using `--disable-replay-opts` at the end of replay I get and error `cannot set revision while there is an existing undo stack`.

According to @wanderingbort:
>thats probably a goof I introduced, there is code at the end of replay that sets the revision of the database which should probably only happen when replay opts are enabled.  When they are disabled, it should probably revert to the previous behavior which was to assert consistency of that version (which is only possible when replay opts are disabled) (one of the replay opts is not to do undo states which means we have nothing to validate at the end)

arhag:
Resolved by #5827.

PR #5827, opened 2018-09-26
fix replay with --disable-replay-opts

arhag:
Fixes the bug in which a replay with `--disable-replay-opts` ends with the `cannot set revision while there is an existing undo stack` exception.


spoonincode:
fixes #5825

PR #5829, opened 2018-09-26
Added a pull request template to the repository.  

zreyn:
Combined with processing a webhook event, the contents of the PR will begin showing up in a release notes draft.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
+**Change Description**
```
wanderingbort: can we put check lists in these PR templates like:

- [ ] BUG
- [ ] User Experience
- [x] RPC

alternatively we can use tags, but it would be nice to have the template provide some very basic set of things a developer could click to indicate the areas-affected/scope of the change

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,11 @@
...
+
+<!-- E.g. a bugfix, feature, refactoring, build related change, etc… What is the change and the motivation for it.  Reference issues where possible.-->
+
+**Implications and Backward Compatibility**
```
wanderingbort: I think we need to break this section into 2, the more likely section is what APIs does this "break" if any.  Right now we consider the HTTP-RPC API as our primary API.  These would largely be the implications and backwards compatibility notes I think this section is meant to address.

the second class which is rare but severe is concensus changes.  This includes changes to the WebAssembly ABI or any of the processing that affects validation of blocks in the chain.  99% of the time, this section would be blank but, if it isn't then it needs to be called out separately and handled.

.github/PULL_REQUEST_TEMPLATE.md
zreyn: Sure, the template is just markdown, so we can put checkboxes wherever.  Do you have a preference between keeping the template small vs letting it get larger?  For example, we could add just a few checkboxes at the top or we can make this thing more like a form, where we provide checkboxes for what aspects are impacted (like documentation, APIs, consensus, specific plugins, etc.) and then ask for comments about each box checked.

.github/PULL_REQUEST_TEMPLATE.md
zreyn: Good idea.  Basically the intent is to ferret out:
1) things that developers should be aware of before they upgrade
2) documentation that needs to be updated to reflect the changes

.github/PULL_REQUEST_TEMPLATE.md
wanderingbort: willing to entertain opinions on it.  My gut is telling me that its more likely to include information if inputing said information is trivial.  Maybe some amount of "form" like stuff is warranted?

zreyn:
It may be hard to do a normal code-review style conversation for this kind of change.  Let's try a gist:
https://gist.github.com/zreyn/6cc2d15254c52abd41d2ba8e1cd978fc

PR #5841, opened 2018-09-27
Fix bug that creates a gap in the syncing blocks

wanderingbort:
When transitioning from the LIB sync mode to the mode where net_plugin catches up to the head of a remote peers best known branch, we calculate the branch to send based on the peers local LIB which could have shifted since the time the request was made due to latency.  If this it the first time we are synchronizing with that branch this means that there could be a gap in delivered blocks resulting in the blocks sent being unlinkable.   The more latency on a connection the more likely it is that the LIB of your block provider moved forward between the packets.

As a fix, we now send, at most, the branch from the LIB we last sent to that peer instead of the current LIB.  In addition, if that peer has reported its head id and we are building on top of that branch, we also squelch the delivery of blocks they already know about.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -853,7 +853,6 @@ namespace eosio {
       block_id_type head_id;
       block_id_type lib_id;
       block_id_type remote_head_id;
-      uint32_t lib_num = 0;
       uint32_t remote_head_num = 0;
       try {
          if (last_handshake_recv.generation >= 1) {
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -872,6 +887,13 @@ namespace eosio {
       block_id_type null_id;
       for (auto bid = head_id; bid != null_id && bid != lib_id; ) {
          try {
+
+            // if the last handshake received indicates that we are catching up on a fork
+            // that the peer is already partially aware of, no need to resend blocks
+            if (block_header::num_from_id(bid) == remote_head_num && remote_head_id == bid) {
```
heifner: Is there a reason this is not just:
`if( remote_head_id == bid ) {`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -852,10 +852,25 @@ namespace eosio {
       }
       block_id_type head_id;
       block_id_type lib_id;
-      uint32_t lib_num;
+      block_id_type remote_head_id;
+      uint32_t lib_num = 0;
```
heifner: `lib_num` is not used.

plugins/net_plugin/net_plugin.cpp
wanderingbort: I suppose not given the recent updates to sha256 equality comparison, its probably more efficient to just do the `==`

PR #5842, opened 2018-09-27
Increase transaction allowed duration and try other wallet ports

brianjohnson5972:
#5674
Increased the max_block_cpu_usage in the genesis.json file
Changed WalletMgr to use the passed in port as a suggested port and to increment the port and try again if the port was not available


programs/eosio-launcher/main.cpp
```diff
@@ -1157,27 +1164,18 @@ launcher_def::write_logging_config_file(tn_node_def &node) {
 
 void
 launcher_def::init_genesis () {
-  bfs::path genesis_path = bfs::current_path() / "genesis.json";
-   bfs::ifstream src(genesis_path);
-   if (!src.good()) {
+   const bfs::path genesis_path = genesis.is_complete() ? genesis : bfs::current_path() / genesis;
```
jgiszczak: While we're at it, let's work around the boost defect that prevents `("genesis,g",bpo::value<bfs::path>(&genesis)` from working if the argument has spaces in the name.

Issue #5840, opened 2018-09-27
v1.3.0-dirty ?


AllenYaoya:
Hi
 I manually downloaded and compiled the source code.  But when i done it, i get the nodeos with version : v1.3.0-dirty.  I just follow the official instrument.
```
git clone https://github.com/EOSIO/eos --recursive

cd eos
./eosio_build.sh

./eosio_install.sh
```
could someone help me to figure out why this happen?

thanks,

spoonincode:
What does a `git status` show?

taokayan:
can you try to checkout branch ```release/1.3.x```?

AllenYaoya:
hi 

i work on the aws, and cat /etc/os-release shows below,
```
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
```
But the auto-build script need the NAME="Amazon Linux AMI" to run. So i manually revised the eos/eosio_build.sh and eos/scripts/eosio_build_amazon.sh to make the script run for NAME = "Amazon Linux". 

Seems due to the difference exists in the git status. Then i get a dirty version after compiling. So in my case, what should i do to make the auto-build script run successful?

thanks!

AllenYaoya:
> What does a `git status` show?
hi,
i manually revised a couple of build scripts. git diff shows the difference.

spoonincode:
Yes, if you edit any files that git is aware of the build will be marked dirty.

spoonincode:
Amazon Linux 2 added to build script; closing

PR #5851, opened 2018-09-28
Fix build script for Amazon Linux 2

spoonincode:
A user implicitly reported via #5840 that Amazon Linux 2 does not build out of the box. Fix the build script to handle it.

```
$ cat /etc/os-release 
NAME="Amazon Linux"
VERSION="2"
ID="amzn"
ID_LIKE="centos rhel fedora"
VERSION_ID="2"
PRETTY_NAME="Amazon Linux 2"
ANSI_COLOR="0;33"
CPE_NAME="cpe:2.3:o:amazon:amazon_linux:2"
HOME_URL="https://amazonlinux.com/"
```

### DIFF
--- a/eosio_build.sh
+++ b/eosio_build.sh
@@ -153,7 +153,7 @@
       OS_NAME=$( cat /etc/os-release | grep ^NAME | cut -d'=' -f2 | sed 's/\"//gI' )
 
       case "$OS_NAME" in
-         "Amazon Linux AMI")
+         "Amazon Linux AMI"|"Amazon Linux")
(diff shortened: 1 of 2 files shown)


wanderingbort:
I would prefer an actual review from someone who knows the build scripts but should this be backported on to 1.3.x as well for a patch release?

spoonincode:
ah yeah should probably be targeted to 1.3. FWIW I did test building on both after the change

heifner:
Automatic rebase via github didn't work. Need to manually rebase to release branch.

spoonincode:
replacing with #5872 so that it hits 1.3.x first

PR #5852, opened 2018-09-28
Add options for buying ram in bytes to newaccount, delegatebw and buyram

### LINKED ISSUE #5783: cleos: allow users to buy bytes
`cleos system create` supports `--buy-ram` (tokens) and `--buy-ram-kbytes`, but not `--buy-ram-bytes`.

`cleos system delegatebw` supports `--buyram`, but not the bytes or kbytes form.

`cleos system buyram` supports `--buy-ram` (tokens) and `--buy-ram-kbytes`, but not `--buy-ram-bytes`. 

If a user wants to buy ram and specify the number of bytes this is not currently possible since the `--buy-ram-kbytes` option take *KB* instead of bytes.

Please add a "--buy-ram-bytes" option to all three cleos commands.

conr2d:
Fixes #5783

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -890,6 +890,7 @@ struct create_account_subcommand {
    string stake_net;
    string stake_cpu;
    uint32_t buy_ram_bytes_in_kbytes = 0;
+   uint32_t buy_ram_bytes = 0;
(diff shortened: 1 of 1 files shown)


heifner:
@dskvr Doc change for 1.4

Issue #4573, opened 2018-07-10
build failed on Centos7.5


leofantast:
OS Version: CentOS Linux release 7.5.1804 (Core)
CPU: 2C Mem: 4G

I have modified line 25 content `if [ "${MEM_MEG}" -lt 7000 ]; then` to `if [ "${MEM_MEG}" -lt 3000 ]; then` in file `scripts/eosio_build_centos.sh`.

building error fragment  as below: 

```
            if (i == confirm_count.size() - 1) {
                ~~^~~~~~~~~~~~~~~~~~~~~~~~~~~
[ 93%] Building CXX object libraries/chain/CMakeFiles/eosio_chain.dir/fork_database.cpp.o
[ 93%] Building CXX object libraries/chain/CMakeFiles/eosio_chain.dir/controller.cpp.o
^@[ 93%] Building CXX object libraries/chain/CMakeFiles/eosio_chain.dir/authorization_manager.cpp.o
^@g++: internal compiler error: 已杀死 (program cc1plus)
Please submit a full bug report,
with preprocessed source if appropriate.
See <http://bugzilla.redhat.com/bugzilla> for instructions.
make[2]: *** [libraries/chain/CMakeFiles/eosio_chain.dir/build.make:231：libraries/chain/CMakeFiles/eosio_chain.dir/controller.cpp.o] 错误 4
make[2]: *** 正在等待未完成的任务....
^@^@make[1]: *** [CMakeFiles/Makefile2:1873：libraries/chain/CMakeFiles/eosio_chain.dir/all] 错误 2
make: *** [Makefile:163：all] 错误 2

	>>>>>>>>>>>>>>>>>>>> MAKE building EOSIO has exited with the above error.
```

any other encounter this?

heifner:
You likely ran out of memory causing an internal compiler error. If you reduce the suggested memory, you should also reduce the number of compiler jobs.

leofantast:
@heifner Thanks! I will try it at other server.

PR #5864, opened 2018-09-30
issues/4573 : Use of $JOBS instead of $CPU_CORE for make -j

NorseGaud:
I noticed that $CPU_CORE is being used in eosio_build.sh, causing my 24 core server to run out of memory (it only has 16GB). Looks like others have been having this issue and my proposed solution should solve it for everyone.

https://github.com/EOSIO/eos/issues/4573

### DIFF
--- a/eosio_build.sh
+++ b/eosio_build.sh
@@ -276,7 +276,7 @@
       printf "\\n\\t>>>>>>>>>>>>>>>>>>>> EOSIO has been successfully configured but not yet built.\\n\\n"
       exit 0
    fi
-   if [ $CPU_CORE -gt 6 ]; then CPU_CORE=$((CPU_CORE / 3)); fi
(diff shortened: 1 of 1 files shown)


eosio_build.sh
```diff
@@ -49,6 +49,14 @@
...
+   # Use current directory's tmp directory if noexec is enabled for /tmp
+   if (mount | grep "/tmp " | grep --quiet noexec); then
+        mkdir -p $SOURCE_DIR/tmp
+        TEMP_DIR="${SOURCE_DIR}/tmp"
+        rm -rf $SOURCE_DIR/tmp/*
+   else # noexec wasn't found
+        TEMP_DIR="/tmp"
+   fi
    START_MAKE=true
    TEMP_DIR="/tmp"
```
heifner: `TEMP_DIR` being reset here.

NorseGaud: Whoops! Thanks for catching that.

PR #5869, opened 2018-09-30
modify the sync numbers of 'start_sync' function

RootkitKiller:
if the 'known_trx.mode' of the 'notice_message' is 'last_irr_catch_up', should use  'known_trx.pending'  instead of 'known_blocks_pending'. because the blocks of 'known_blocks_pending' needs to be verified.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1537,7 +1537,7 @@ namespace eosio {
       else {
          c->last_handshake_recv.last_irreversible_block_num = msg.known_trx.pending;
          reset_lib_num (c);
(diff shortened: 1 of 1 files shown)


RootkitKiller:
What is the cause of the problem that 'Some checks were not successful
2 failing and 6 successful checks'?

spoonincode:
1.3.x?

heifner:
@spoonincode @wanderingbort and I discussed 1.3.x for this. It is a bit obscure. I'm good with either 1.3.x or develop. Thoughts?

spoonincode:
I didn't actually study the change closely. If you & bart already think it's obscure enough to wait for 1.4 I've got no beef -- just wanted to make sure we considered it

PR #5871, opened 2018-09-30
Centos build improvements

NorseGaud:
- if;then cleanup
- Better formatting for stdout
- `yum` package installation separation from SCL + use of `rpm -qa` to check the existence
- Moved `yum` update before `yum` installs
- Using "Don't Repeat Yourself" (DRY) for installations throughout, making it easier to change version

### DIFF
--- a/scripts/eosio_build_centos.sh
+++ b/scripts/eosio_build_centos.sh
@@ -21,6 +21,7 @@
 	printf "\\tDisk install: %s\\n" "${DISK_INSTALL}"
 	printf "\\tDisk space total: %sG\\n" "${DISK_TOTAL%.*}"
 	printf "\\tDisk space available: %sG\\n" "${DISK_AVAIL%.*}"
+	printf "\\tConcurrent Jobs (make -j): ${JOBS}\\n"
 
 	if [ "${MEM_MEG}" -lt 7000 ]; then
 		printf "\\n\\tYour system must have 7 or more Gigabytes of physical memory installed.\\n"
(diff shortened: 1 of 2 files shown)


scripts/eosio_build_centos.sh
```diff
@@ -493,224 +495,204 @@ mongodconf
...
-			printf "\\tExiting now.\\n\\n"
+			printf "\\t!! Unable to download MongoDB C driver at this time !!\\n"
+			printf "\\tExiting now.\\n"
 			exit 1;
 		fi
-		if ! tar xf mongo-c-driver-1.10.2.tar.gz
-		then
-			printf "\\tUnable to unarchive file %s/mongo-c-driver-1.10.2.tar.gz.\\n" "${TEMP_DIR}"
-			printf "\\tExiting now.\\n\\n"
+		if ! tar xf "${MONGODRIVERTGZ}"; then
```
wanderingbort: it seems highly likely that this line has been bad for sometime and is now causing CI to reject your changes.  Should it be `tar xf "${TEMP_DIR}/${MONGODRIVERTGZ}"` ?

NorseGaud: Ah, good catch. I'll pwd in there and see what directory I'm in when I extract the tar. I'll confirm later tonight

NorseGaud: `pwd` shows I'm in the tmp directory when it gets to tar xf, so we're good there. Further investigation reveals MONGODRIVEURL was used for STATUS=$() and not MONGODRIVERURL, with an R.

PR #5877, opened 2018-10-02
Fix return value when throws connection_exception in cleos

### LINKED ISSUE #5809: cleos Returns Bad Return Code
If --url or --wallet-url are bad addresses, cleos gives an error message on stderr, but has a return code of 0

conr2d:
Fixes #5809


heifner:
Replaced by #5881 rebased to release/1.3.x branch

PR #5893, opened 2018-10-03
Add mongodb-update-via-block-num option

heifner:
- Currently, as mini-forks happen the forked blocks are orphaned in MongoDB
- Enable mongodb-update-via-block-num to overwrite blocks by block number instead of by block id
- Option affects both blocks and block_states collections
- Default is false, add `mongodb-update-via-block-num = true` to config.ini to enable

### DIFF
--- a/plugins/mongo_db_plugin/mongo_db_plugin.cpp
+++ b/plugins/mongo_db_plugin/mongo_db_plugin.cpp
@@ -127,6 +127,7 @@ public:
    bool filter_on_star = true;
    std::set<filter_entry> filter_on;
    std::set<filter_entry> filter_out;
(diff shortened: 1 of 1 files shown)


heifner:
Check for duplicates in mongo block_states collection:

`> db.block_states.aggregate([ {$group:{"_id":"$block_num","name":{$first:"$block_num"},"count":{$sum:1}}}, {$match:{"count":{$gt:1}}}, {$project:{"name":1,"_id":0}}, {$group:{"_id":null,"duplicates":{$push:"$name"}}}, {$project:{"_id":0,"duplicates":1}}], {"allowDiskUse":true} )`

PR #5902, opened 2018-10-04
fix nodeos_under_min_avail_ram_lr_test

### LINKED ISSUE #5891: Eosio v1.3.1 test fails.
The last test nodeos_under_min_avail_ram_lr_test fails.

taokayan:
fix #5891, break the loop if push transaction failed consecutively couple times.

### DIFF
--- a/tests/nodeos_under_min_avail_ram.py
+++ b/tests/nodeos_under_min_avail_ram.py
@@ -48,6 +48,7 @@ class NamedAccounts:
         Print("NamedAccounts Name for %d is %s" % (temp, retStr))
         return retStr
 
+
 ###############################################################
 # nodeos_voting_test
 # --dump-error-details <Upon error print etc/eosio/node_*/config.ini and var/lib/node_*/stderr.log to stdout>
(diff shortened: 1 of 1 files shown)


tests/nodeos_under_min_avail_ram.py
```diff
@@ -167,8 +169,15 @@ def setName(self, num):
             try:
                 trans=nodes[0].pushMessage(contract, action, data, opts)
                 if trans is None or not trans[0]:
+                    timeOutCount+=1
+                    if timeOutCount>=3:
+                       Print("Failed to push create action to eosio contract for %d consecutive times, looks like nodeos already exited." % (timeOutCount))
```
brianjohnson5972: This should be a errorExit(...) call so that we report that we are failing because we failed pushing the message <timeOutCount> times.

taokayan: In my understanding this test script does expect nodeos exits due to no available memory, then the script will continue to check how many rounds of transactions have been pushed before nodeos exits. am I correct?

brianjohnson5972: Yes, so  if it gets a None transaction, nodeos hasn't shutdown, so it is something else going wrong, and we don't want to evaluate anything more. and should stop.

brianjohnson5972: Ok, just ran this and not sure why the failure is occurring different than when I first wrote it, but @taokayan has it right.

PR #5919, opened 2018-10-05
Add missing tags to exception text for arg processing.

### LINKED ISSUE #5909: Error 3015004: The type defined in the ABI is invalid
Please include the name of the type that is invalid in the error reporting.

I am working on a pretty simple contract here. 

I had the wrong type defined as uint64_t instead of uint64. Got this error - Error 3015004 - when I tried to run an action on that contract.

The type defined in the ABI is invalid - but it would be VERY helpful to show the NAME of the TYPE  that was invalid.

jgiszczak:
Resolves #5909.


heifner:
Not needed. Merged in from release/1.3.2.

PR #5920, opened 2018-10-05
Add missing tags to exception text for arg processing.

### LINKED ISSUE #5909: Error 3015004: The type defined in the ABI is invalid

jgiszczak:
Resolves #5909 for next patch release.

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -118,7 +118,7 @@ namespace eosio { namespace chain {
          structs[st.name] = st;
 
       for( const auto& td : abi.types ) {
-         EOS_ASSERT(_is_type(td.type, 0, deadline, max_serialization_time), invalid_type_inside_abi, "invalid type {type}", ("type",td.type));
+         EOS_ASSERT(_is_type(td.type, 0, deadline, max_serialization_time), invalid_type_inside_abi, "invalid type ${type}", ("type",td.type));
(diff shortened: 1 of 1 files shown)


libraries/chain/abi_serializer.cpp
```diff
@@ -118,7 +118,7 @@ namespace eosio { namespace chain {
          structs[st.name] = st;
 
       for( const auto& td : abi.types ) {
-         EOS_ASSERT(_is_type(td.type, 0, deadline, max_serialization_time), invalid_type_inside_abi, "invalid type", ("type",td.type));
+         EOS_ASSERT(_is_type(td.type, 0, deadline, max_serialization_time), invalid_type_inside_abi, "invalid type {type}", ("type",td.type));
```
heifner: These need to be protected by `maybe_shorten`. However, since that context is not available, might be easiest to just create a `shorten`  static method that `maybe_shorten` can call that can also be called from these locations.

jgiszczak: `maybe_shorten` is conditional on whether or not the `short_path` flag has been set in the context.  However, these are error messages, so they should not be shortened (and with the context not available, the flag can't be checked anyway).

Added missing $.

PR #5933, opened 2018-10-09
Fixes for remote tests

cj-oci:
The remote tests are broken on account of being inactive. This path will fix the tests.


tests/nodeos_run_test.py
```diff
@@ -202,14 +206,14 @@
     cluster.validateAccounts(None)
 
     # create accounts via eosio as otherwise a bid is needed 
-    Print("Create new account %s via %s" % (testeraAccount.name, cluster.eosioAccount.name))
-    transId=node.createInitializeAccount(testeraAccount, cluster.eosioAccount, stakedDeposit=0, waitForTransBlock=False, exitOnError=True)
+    Print("Create new account %s via %s" % (testeraAccount.name, cluster.defproduceraAccount.name))
```
brianjohnson5972: Does the comment above this still apply?

PR #5938, opened 2018-10-09
refactor eosio_exit for WAVM; improve long running performance

spoonincode:
WAVM becomes increasingly slow as more and more contracts are activated. The ultimate reason is due to high exception overhead from eosio_exit. I believe this stems from the huge number of memory mappings and interaction with the unwinder. Refactor eosio_exit implementation on WAVM to not use exceptions and instead longjmp out of the running wasm code.

A popular EOSIO based blockchain with approximately 20MM blocks replayed in half the time for me.

I’m aware of the wonkiness with the implementation here — the “get out of a running instance” method is on the global wasm runtime class and not on an instance of a wasm module. This was the easiest way to add the change and touch the least amount of code. Plus, at any point this isn’t sufficient due to having more than one contract running simultaneously, we will need to refactor much of the wasm execution code anyways.

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface.hpp
@@ -65,6 +65,9 @@ namespace eosio { namespace chain {
          //Calls apply or error on a given code
          void apply(const digest_type& code_id, const shared_string& code, apply_context& context);
 
+         //Immediately exits currently running wasm. UB is called when no wasm running
+         void exit();
+
       private:
          unique_ptr<struct wasm_interface_impl> my;
          friend class eosio::chain::webassembly::common::intrinsics_accessor;
--- a/libraries/chain/include/eosio/chain/webassembly/runtime_interface.hpp
+++ b/libraries/chain/include/eosio/chain/webassembly/runtime_interface.hpp
@@ -17,6 +17,9 @@ class wasm_runtime_interface {
    public:
       virtual std::unique_ptr<wasm_instantiated_module_interface> instantiate_module(const char* code_bytes, size_t code_size, std::vector<uint8_t> initial_memory) = 0;
 
+      //immediately exit the currently running wasm_instantiated_module_interface. Yep, this assumes only one can possibly run at a time.
+      virtual void immediately_exit_currently_running_module() = 0;
+
       virtual ~wasm_runtime_interface();
 };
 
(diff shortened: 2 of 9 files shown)


taokayan:
hi, just want to understand is there any potential risk to use "siglongjmp"? I found out that it will skip destructors of local objects, for example:
```
#include <stdio.h>
#include <stdlib.h>
#include <unistd.h>
#include <setjmp.h>

jmp_buf env;
struct A {
  A() { printf("A::A() "); }
  ~A() { printf("A::~A() "); }
};
void g() {
  A a;
  siglongjmp(env, 1);
}

int main(int argc, char **argv) {
  bool isRet = sigsetjmp(env, 1);
  if (!isRet) {
    g();
  }
  printf("exit...");
  return 0;
}
```
this will output ```A::A() exit...``` without calling A::~A()

spoonincode:
It's dangerous to do blindly, of course. But let's look what the call stack looks like here.

`setjmp` is called here, right before `thunk()` is called
https://github.com/EOSIO/eos/blob/13d0a6c2cff74f8d9e37495b3cea6102fb43c31d/libraries/wasm-jit/Source/Platform/POSIX.cpp#L257
`thunk()` in this case is the code here
https://github.com/EOSIO/eos/blob/905e7c85714aee4286fa180ce946f15ceb4ce73c/libraries/wasm-jit/Source/Runtime/Runtime.cpp#L115-L127
This is basically just calling in to the x86_64 assembly of the apply() call for a contract.

As the wasm executes it can call what we refer to as "intrinsics" (imports from the env module that call c++ code in nodeos). One of these intrinsics is eosio_exit. Also keep in mind that no other intrinsics call eosio_exit. So when wasm calls eosio_exit (and thus siglongjmp) the call graph from the wasm code is

https://github.com/EOSIO/eos/blob/13d0a6c2cff74f8d9e37495b3cea6102fb43c31d/libraries/chain/include/eosio/chain/webassembly/wavm.hpp#L332
https://github.com/EOSIO/eos/blob/13d0a6c2cff74f8d9e37495b3cea6102fb43c31d/libraries/chain/include/eosio/chain/webassembly/wavm.hpp#L357
https://github.com/EOSIO/eos/blob/13d0a6c2cff74f8d9e37495b3cea6102fb43c31d/libraries/chain/include/eosio/chain/webassembly/wavm.hpp#L661
https://github.com/EOSIO/eos/blob/64574226b357d36b99f07c3aba3236eac5680d79/libraries/chain/wasm_interface.cpp#L954

At no point in this graph is heap allocated nor are other c++ resources allocated. So jumping back to before the `thunk()` is safe -- anything the x86_64 wasm code left on the stack is discarded by the stack pointer being "rewound".

Keep in mind too that this is not the first time siglongjmp is used in WAVM -- it is used the exact same way to recover from a SEGV within a contract.

PR #5940, opened 2018-10-09
Added Diagnostics for Missing Transactions in integration tests

brianjohnson5972:
#5674
Added using the block log tool to write out the block logs for all nodes if a node indicates that there was a missing transaction (a request that created a transaction, but a subsequent search for the transaction could not find the transaction)


brianjohnson5972:
Forgot to change merge branch

PR #5952, opened 2018-10-10
 Turning Off Buildkite Automatic retry

brianjohnson5972:
#5674 
Since the EOSIO and EOSIO - LRT suites are stable now, turning off the automatic retry to see if we can identify the cause of the occasional spurious errors that have occurred in the past.


brianjohnson5972:
Just have to review commit 3fc49c9506f397de3b282336d6b5861162ed1bf4, since the rest are from #5941

PR #5957, opened 2018-10-10
Create action_trace & action_receipt on hard_fail

heifner:
- `hard_fail` was creating default constructed `action_trace` to be added to `transaction_trace`
- Fill out `action_trace` and `action_receipt` even when exception while processing action
- Add optional `except` to `action_trace` to store exception that is also stored on `transaction_trace`
- **Note:** Includes change to `action_trace` which will likely cause problems with `history_plugin` unless full replay

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -92,11 +92,11 @@ void apply_context::exec_one( action_trace& trace )
 
    trx_context.executed.emplace_back( move(r) );
 
+   finalize_trace( trace, start );
+
    if ( control.contracts_console() ) {
       print_debug(receiver, trace);
    }
-
-   finalize_trace( trace, start );
 }
 
 void apply_context::finalize_trace( action_trace& trace, const fc::time_point& start )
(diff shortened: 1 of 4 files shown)


libraries/chain/apply_context.cpp
```diff
@@ -74,36 +47,75 @@ action_trace apply_context::exec_one()
...
+      trace.except = e;
+      finalize_trace( trace, start );
+      throw;
+   }
 
    trx_context.executed.emplace_back( move(r) );
 
    if ( control.contracts_console() ) {
-      print_debug(receiver, t);
+      print_debug(receiver, trace);
```
arhag: `print_debug` will be passed a trace with an empty `console` since it hasn't been set yet with `finalize_trace`. It should be moved to after the `finalize_trace` call below.

heifner: Moved.

heifner:
Note I had to move the setting of `r.code_sequence` and `r.abi_sequence` to after action execution because they can change. I figured since I moved them I might as well move the other receipt settings that modify db after the execution as well.

PR #5976, opened 2018-10-12
Fix documentation of current_time

yukozh:
#4652 is a wrong fix. current_time() in EOS is returning seconds, not milli seconds


heifner:
- Changes to eosiolib/* should be pull requests to https://github.com/EOSIO/eosio.cdt
- current_time is in microseconds

PR #5987, opened 2018-10-12
Remove config.ini options that are no longer supported by nodeos

wpwood:
The default eosio/eos-dev docker image will not run nodeos without replacing the config.ini with a working version. It will give an error message about invalid options in the config.ini.


spoonincode:
any chance you can go ahead and remove everything other then the ones absolutely needed?

wpwood:
@spoonincode I would if I had any idea what was or was not needed. I trimmed just enough to fix the error. If you know, you're more than welcome to make the change.

wpwood:
Need to reopen as a PR to develop.

PR #6003, opened 2018-10-15
Binary Releases

larryk85:
This is supplanting #5727, currently with this set up it will only create a package for [`cleos`, `keosd`, and `nodeos` (if we think it is okay)].  Currently, I have this set up to install the binaries to /usr/opt/eosio/v1.x.x/bin and install symlinks to /usr/bin.  I used this model for `eosio.cdt`'s binary releases as it gives us clean versioning, but with a few simple modifications we can change this to install the binaries to /usr/bin.  For this incarnation of the binary releases it doesn't make much difference in terms of versioning, but when we cleanup and fix `EosioTester` to be included with a binary release, proper versioning becomes a very real issue, so I suggest the /usr/opt/eosio/v1.x.x approach.  Or, /usr/share for `eos`, as everything should be native and not mostly wasm specific like `eosio.cdt`.

#### nodeos
Currently, `nodeos` has a dependence on `CORE_SYMBOL` through `chain_plugin::get_account`.  This is only used if the `core symbol` cannot be determined (no `rammarket`), and is used to return the accounts liquid balance.  I have not been able to come up with a reasonable solution to remove this dependence entirely, but the worst that happens is the query fails to find the accounts liquid balance and is displayed in `cleos` as such.


scripts/generate_tarball.sh
```diff
@@ -0,0 +1,40 @@
...
+#sed "s/_PREFIX_/\/${SPREFIX}/g" ${BUILD_DIR}/modules/EosioTesterPackage.cmake &> ${EOS_PREFIX}/lib/cmake/${PROJECT}/EosioTester.cmake
+#sed "s/_PREFIX_/\/${SPREFIX}\/${SSUBPREFIX}/g" ${BUILD_DIR}/modules/${PROJECT}-config.cmake.package &> ${EOS_PREFIX}/lib/cmake/${PROJECT}/${PROJECT}-config.cmake
+
+# install includes
+#cp -R ${BUILD_DIR}/include/* ${EOS_PREFIX}/include
+
+# make symlinks
+#pushd ${PREFIX}/lib/cmake/${PROJECT} &> /dev/null
+#ln -sf ../../../${SUBPREFIX}/lib/cmake/${PROJECT}/${PROJECT}-config.cmake ${PROJECT}-config.cmake
+#ln -sf ../../../${SUBPREFIX}/lib/cmake/${PROJECT}/EosioTester.cmake EosioTester.cmake
```
spoonincode: All these commented out lines in this file. Is it supposed to be that way?

larryk85: Yeah, when we create full binary releases with libeosio and tester hopefully all that needs to be done in the generator scripts is uncomment these lines.

PR #6021, opened 2018-10-16
Remove wallet plugin settings from Docker config.ini.

jgiszczak:
One setting removed from develop branch already, but not all.


spoonincode:
I really think we should consider removing the config.ini altogether

PR #6033, opened 2018-10-17
trx catch_up never processed

fingera:
**trx catch_up never processed**
catch_up send:
https://github.com/EOSIO/eos/blob/2232802dee4be1a3f6ca436edad64af3c45d7029/plugins/net_plugin/net_plugin.cpp#L1462-L1466

It looks like it has not been processed and returned.
https://github.com/EOSIO/eos/blob/2232802dee4be1a3f6ca436edad64af3c45d7029/plugins/net_plugin/net_plugin.cpp#L2416-L2418


heifner:
Thanks @fingera. It looks to me like `normal` and `catch_up` are both possible and should be allowed. Also those seem to be the only two, so I just removed the `if`. Thoughts?

Replaced by #6037

fingera:
OK
Only catch_up has set send_req=true
Protocol changes may require a full test

PR #6051, opened 2018-10-18
Remove eos_utilities lib

heifner:
**Change Description**

- Removed the eos_utilities library as it contained little of value
- Moved the few utils used to their use or to chain library


programs/cleos/main.cpp
```diff
@@ -2732,7 +2730,7 @@ int main( int argc, char** argv ) {
          fc::set_console_echo(true);
       }
 
-      auto priv_key = fc::crypto::private_key::regenerate(*utilities::wif_to_key(str_private_key));
+      auto priv_key = private_key_type(str_private_key);
       trx.sign(priv_key, *chain_id);
```
heifner: @spoonincode Please review the above.

PR #6065, opened 2018-10-19
No need to prompt the user to set bnet-no-trx

fingera:
**Change Description**
No need to prompt the user to set bnet-no-trx to true？

### DIFF
--- a/plugins/bnet_plugin/bnet_plugin.cpp
+++ b/plugins/bnet_plugin/bnet_plugin.cpp
@@ -1383,7 +1383,10 @@ namespace eosio {
 
 
       if( app().get_plugin<chain_plugin>().chain().get_read_mode() == chain::db_read_mode::READ_ONLY ) {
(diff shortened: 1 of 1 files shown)


plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -1403,7 +1384,6 @@ namespace eosio {
 
       if( app().get_plugin<chain_plugin>().chain().get_read_mode() == chain::db_read_mode::READ_ONLY ) {
          my->_request_trx = false;
-         ilog( "setting bnet-no-trx to true since in read-only mode" );
```
heifner: I would prefer to keep the ilog so it is clear the side effect. You can ilog only if my->_request_trx is true.

fingera: ok, added ilog when bnet-no-trx=false

Issue #6067, opened 2018-10-19
bidnameinfo returning bidname even though it doesn't exist


NorseGaud:
<pre>cleos -uhttp://api.eosnewyork.io system bidnameinfo notcreated
bidname:                              note
highest bidder:               chaoxiaojuan
highest bid:                            12
last bid time:     2018-06-21T02:32:34.500</pre>

This can be replicated using anything that doesn't exist in the bidname system. It looks to be matching the first few characters:

<pre>cleos -uhttp://api.eosnewyork.io system bidnameinfo t3st
bidname:                                t4
highest bidder:               eoszzzeoszzz
highest bid:                             1
last bid time:     2018-06-29T16:09:29.000</pre>

I believe it should return "No bidname record found" (cleos/main.cpp):
<pre>         if ( result.rows.empty() ) {
            std::cout << "No bidname record found" << std::endl;
            return;
         }</pre>

conr2d:
@brianjohnson5972 
Can you review #6070 ? That will solve more issues like this.

taokayan:
Looks like this issue is solved:)

PR #6068, opened 2018-10-19
Fix bidnameinfo to just return actual matches

brianjohnson5972:
**Change Description**

#6067 
nodeos's table api returns row entries from lower_bound to upper_bound, so needed to prevent showing bidname rows that were returned with the next available bidname after the requested bidname.

**Consensus Changes**

N/A (This only effects read only cleos functionality)

**API Changes**

This affects cloes to only return a matching bidname when "cleos system bidnameinfo" is called.

**Documentation Additions**

This would be expected behavior (previous behavior was a bug).

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -1296,7 +1296,8 @@ struct bidname_info_subcommand {
             return;
          }
          auto result = rawResult.as<eosio::chain_apis::read_only::get_table_rows_result>();
-         if ( result.rows.empty() ) {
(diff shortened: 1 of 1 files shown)


programs/cleos/main.cpp
```diff
@@ -1296,7 +1296,8 @@ struct bidname_info_subcommand {
             return;
```
heifner: Should the raw also be checked and return `{}` or something?

brianjohnson5972:
#6070 was a better solution, closing this

PR #6070, opened 2018-10-19
Fix cleos to get table rows only when satisfying condition

### LINKED ISSUE #6067: bidnameinfo returning bidname even though it doesn't exist

conr2d:
**Change Description**

Fixes #6067
When cleos gets table rows to print info, the condition is given only to lower bound, so not exact matched rows are included to the result. This patch adds upper bound, so the rows which not matching exactly will not included.

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -1033,10 +1033,11 @@ struct approve_producer_subcommand {
                                ("table", "voters")
                                ("table_key", "owner")
                                ("lower_bound", voter.value)
+                               ("upper_bound", voter.value + 1)
                                ("limit", 1)
             );
             auto res = result.as<eosio::chain_apis::read_only::get_table_rows_result>();
-            if ( res.rows.empty() || res.rows[0]["owner"].as_string() != name(voter).to_string() ) {
+            if ( res.rows.empty() ) {
                std::cerr << "Voter info not found for account " << voter << std::endl;
                return;
             }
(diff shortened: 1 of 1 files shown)


conr2d:
Yes, I tested following commands that can be affected by this PR:
- cleos system voteproducer approve
- cleos system voteproducer unapprove
- cleos system bidnameinfo
- cleos multisig review

arhag:
The actual problem with #6067 is that there is a bug in how chain_plugin handles `upper_bound`. It does the same thing as `lower_bound`.

The meaning of `upper_bound` in `get_table_rows` is _supposed_ to be the same as `std::upper_bound`: the first element in the range `[first, last)` that is greater than the supplied `value` or `last` if no such element is found.

This means that one can use the same `value` for the `lower_bound` and `upper_bound` and the sub-range `[lower_bound, upper_bound)` would include the set of all items that match that `value`. With the current implementation of chain_plugin's `get_table_rows`, passing in the same `value` for both `lower_bound` and `upper_bound` will always give you no results.

This PR assumes a different definition of `upper_bound` (the one that is unfortunately currently implemented in chain_plugin). And with that assumed definition it does an effective job fixing #6067. But I think the correct bug fix is to first fix chain_plugin and then change the cleos code touched in this PR to pass in the same value for both the `lower_bound` and `upper_bound` field.

PR #6100, opened 2018-10-22
Remove llvm@4 force link for Mac OS X builds

larryk85:
**Change Description**
This removes brew --force link for the llvm@4 package, this force link had unfortunate side effects of overwriting the system clang++/clang with clang 4.0 and libc++ which was causing issues with binary builds.

**Consensus Changes**

**API Changes**

**Documentation Additions**

Before running the fixed `eosio_build.sh` it is recommended to run `brew remove llvm@4` to allow the script to do a fresh install or at the very least `brew unlink llvm@4`.


CMakeLists.txt
```diff
@@ -14,6 +14,14 @@ endif()
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/libraries/fc/CMakeModules")
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 
+if (UNIX)
+   if (APPLE)
+      if (LLVM_DIR STREQUAL "" OR NOT LLVM_DIR)
+         set(LLVM_DIR "/usr/local/Cellar/llvm@4/4.0.1/lib/cmake/llvm")
```
spoonincode: Where is LLVM_DIR used? I can't seem to find any other reference to it

larryk85: It is used in FindLLVM.cmake.

spoonincode: okay yes I see, LLVM_DIR is used as an input hint to find_package(LLVM)

PR #6135, opened 2018-10-24
Update Sanitizers Pipeline ctest command

prattj:
**Change Description**

Updated the ctest command to always return true for the undefined sanitizer and address sanitizer command steps in the buildkite sanitizers pipeline.  Failure of this step is not critical.  The pertinent sanitizer error information is being captured as a build artifact sanitizer log.

**Consensus Changes**

None.

**API Changes**

None.

**Documentation Additions**

None.


zreyn:
Yup, this is good.  The outcome of the tests don't matter much for this.  We care about the tests executing the code paths so that the sanitizers can do the runtime analysis.  The tests that assert time will frequently fail because the sanitizers incur a performance penalty, but we don't want the pipeline to appear broken when it's doing what we want.

PR #6149, opened 2018-10-25
Multi-thread transaction key recovery

heifner:
**Change Description**

- Transaction key recovery is now done on separate threads via a thread pool in the controller when applying a signed block.

**Consensus Changes**

None

**API Changes**

- chain_plugin has new configuration option:
  -- chain-threads - number of threads for the controller thread_pool. Currently used for transaction key recovery, but could be used for other operations in the future.

**Documentation Additions**

- New chain-threads nodeos configuration option `chain-threads` - number of threads to use for transaction key recovery when applying a signed block.


libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -25,6 +26,7 @@ struct block_header_state {
     public_key_type                   block_signing_key;
     vector<uint8_t>                   confirm_count;
     vector<header_confirmation>       confirmations;
+    std::shared_future<public_key_type> block_signing_key_future;
```
wanderingbort: This doesn't seem to be used?  was there some files left out of this PR (despite it operating fine)?  Or is this future expansion

heifner: I was experimenting with block signing future. I'll remove this.

Issue #6056, opened 2018-10-18
Nodeos takes very long time to start


lhs168:
I upgraded the nodeos to 1.4.1 and ran it. But it took about 30 minutes for startup. Why so long? My system is CentOS 7.4, CPUx2, RAM 8G.

These are logs:
---------------------
info  2018-10-18T17:33:30.807 thread-0  chain_plugin.cpp:333          plugin_initialize    ] initializing chain plugin
info  2018-10-18T17:33:30.812 thread-0  block_log.cpp:134             open                 ] Log is nonempty
info  2018-10-18T17:33:30.812 thread-0  block_log.cpp:161             open                 ] Index is nonempty
info  2018-10-18T17:33:30.836 thread-0  net_plugin.cpp:2901           plugin_initialize    ] Initialize net plugin
info  2018-10-18T17:33:30.836 thread-0  net_plugin.cpp:2927           plugin_initialize    ] host: 0.0.0.0 port: 6688
info  2018-10-18T17:33:30.837 thread-0  net_plugin.cpp:2998           plugin_initialize    ] my node_id is d91cae0be414716bb94bb858caf7a056bd822f9bb7ffe869c0af92f2daa9f65a
info  2018-10-18T17:33:30.837 thread-0  http_plugin.cpp:422           plugin_initialize    ] configured http to listen on 0.0.0.0:80
info  2018-10-18T17:33:30.837 thread-0  http_plugin.cpp:460           plugin_initialize    ] configured https to listen on 0.0.0.0:443 (TLS configuration will be validated momentarily)
warn  2018-10-18T17:33:30.837 thread-0  net_api_plugin.cpp:96         plugin_initialize    ] 

info  2018-10-18T17:33:30.837 thread-0  http_plugin.cpp:377           operator()           ] configured http with Access-Control-Allow-Origin: *
info  2018-10-18T17:33:30.837 thread-0  main.cpp:110                  main                 ] nodeos version v1.4.1
info  2018-10-18T17:33:30.837 thread-0  main.cpp:111                  main                 ] eosio root is /root/.local/share

*************************************************
[ Nodeos was stuck here about 30 minutes ]
*************************************************

info  2018-10-18T17:33:30.837 thread-0  controller.cpp:389            init                 ] database initialized with hash: b0a8a50183a121074a9506423313ba5ac243baeb93b1e5dcb982c2c08ba76b6d
[...]

PR #6164, opened 2018-10-26
Gh#6056 long startup

heifner:
**Change Description**

As documented by #6056 some people are seeing slow startup. This is caused by `calculate_integrity_hash()` taking a long time to run. 

Only call the startup `calculate_integrity_hash` when starting from a snapshot to avoid delay.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -387,7 +387,8 @@ struct controller_impl {
       }
 
       if( snapshot ) {
-         ilog( "database initialized with hash: ${hash}", ("hash", calculate_integrity_hash()) );
+         const auto hash = calculate_integrity_hash();
+         ilog( "database initialized with hash: ${hash}", ("hash", hash) );
       }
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -386,7 +386,10 @@ struct controller_impl {
          db.undo();
       }
 
-      ilog( "database initialized with hash: ${hash}", ("hash", calculate_integrity_hash()));
+      if( snapshot ) {
+         const auto hash = calculate_integrity_hash();
```
brianjohnson5972: Are there side-effects in calculate_integrity_hash that we want to occur even if the information logging is not turned on?

heifner: No. Should have no side-effects.

brianjohnson5972: Nevermind, just read Bart's telegram and I see why we want to do this outside the ilog.

PR #6167, opened 2018-10-26
Start block state creation early

heifner:
**Change Description**

- `controller::push_block()` now takes a future to a block state that can be created by `controller::create_state_block_future()`. This allows the creation of the block state to begin earlier (before abort block) including the verfication of block signee.
- Added a `controller::replay_push_block` for use during replay since futures would only slow that process down (block signee is not verified during normal replay).
- Capture weak_ptr for futures so they do not keep shared_ptr alive

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1178,13 +1178,18 @@ struct controller_impl {
          start_block( b->timestamp, b->confirmed, s , producer_block_id);
 
          std::vector<transaction_metadata_ptr> packed_transactions;
+         packed_transactions.reserve( b->transactions.size() );
          for( const auto& receipt : b->transactions ) {
             if( receipt.trx.contains<packed_transaction>()) {
                auto& pt = receipt.trx.get<packed_transaction>();
                auto mtrx = std::make_shared<transaction_metadata>( pt );
                if( !self.skip_auth_check() ) {
-                  mtrx->signing_keys_future = async_thread_pool( [this, mtrx]() {
-                     return std::make_pair( this->chain_id, mtrx->trx.get_signature_keys( this->chain_id ) );
+                  std::weak_ptr<transaction_metadata> mtrx_wp = mtrx;
+                  mtrx->signing_keys_future = async_thread_pool( [chain_id = this->chain_id, mtrx_wp]() {
+                     auto mtrx = mtrx_wp.lock();
+                     return mtrx ?
+                            std::make_pair( chain_id, mtrx->trx.get_signature_keys( chain_id ) ) :
+                            std::make_pair( chain_id, decltype( mtrx->trx.get_signature_keys( chain_id ) ){} );
                   } );
                }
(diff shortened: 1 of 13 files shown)


heifner:
Tieing block signee delayed verification to light validation is the opposite of what this should be tied to. Light validation implies you are doing the same validation as a light client. Block signature verification is not something you want to postpone in that mode.

libraries/chain/controller.cpp
```diff
@@ -1251,19 +1256,38 @@ struct controller_impl {
...
       try {
-         EOS_ASSERT( b, block_validate_exception, "trying to push empty block" );
-         EOS_ASSERT( s != controller::block_status::incomplete, block_validate_exception, "invalid block status for a completed block" );
+         block_state_ptr new_header_state = block_state_future.get();
+         auto& b = new_header_state->block;
          emit( self.pre_accepted_block, b );
-         bool trust = !conf.force_all_checks && (s == controller::block_status::irreversible || s == controller::block_status::validated);
-         auto new_header_state = fork_db.add( b, trust );
+
+         fork_db.add( new_header_state );
```
arhag: Calling `fork_database::add( block_state_ptr )` here instead of the original `fork_database::add( signed_block_ptr bool )` skips some checks that were done in the original version.

`create_block_state_future` replicates those `fork_database::add( signed_block_ptr, bool)` checks, but it does so at a different time (and therefore potentially with different state of the fork database) than the call to `fork_db.add` here.

One of the two necessary checks, checking for duplicate blocks, is already handled in `fork_database::add( block_state_ptr )` with the assertion that `inserted.second` is true.

The other check that is necessary to again here (so it could be done in `fork_database::add( block_state_ptr )`) is to ensure that the block still links to an existing block in the fork database. Without that check, it is possible to add a block into the fork database that does not link. For example, a user of `controller` could in theory: 
1. Call `create_block_state_future` on two different blocks A and B where both currently link to existing blocks in the fork database and where B builds off of the current head.
2. Then calls `push_block` with the future of block A which causes: a fork switch, LIB advancement, and purging of orphaned branches which may include the block that the block B builds off of.
3. Then calls `push_block` with the future of block B.

heifner: The current code does not have more than one `create_block_state_future` in play at a time. It is single threaded from producer plugin `on_incoming_block` to `push_block`. The `create_block_state_future` creation of block_state which includes digest, merkle, and signee verification can run at the same time as `abort_block` but the rest of the order is all single threaded.

Now if we updated the code so that more than one `create_block_state_future` could be in flight at a time then that could certainly be a problem. That would take some work as it does verify previous is in the fork_db. Could create a version of `create_block_state_future` that takes the previous block so that a whole chain of blocks could be processed at once and then verify in fork_db add that the previous is there. However, that would require changing net_plugin to receive more than one block at a time or queue up blocks which would increase latency which is what this is attempting to reduce.

That said the extra check doesn't hurt and should be quick, so I'll go ahead and add it.

Issue #6177, opened 2018-10-29
Fix set_code and set_abi to report only json to stdout


brianjohnson5972:
When --json flag is passed, only json should be returned on stdout.

andriantolie:
PR 6178 is merged. I'm closing this issue

PR #6178, opened 2018-10-29
Fix set_code and set_abi to report only json to stdout

brianjohnson5972:
**Change Description**

#6177 
set_code and set_abi were reporting error information on stdout, which is expected to be json output if json flag is set.

**Consensus Changes**

N/A

**API Changes**

N/A

**Documentation Additions**

N/A

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -2388,7 +2388,7 @@ int main( int argc, char** argv ) {
             send_actions(std::move(actions), 10000, packed_transaction::zlib);
          }
       } else {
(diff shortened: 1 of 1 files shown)


programs/cleos/main.cpp
```diff
@@ -2437,7 +2437,7 @@ int main( int argc, char** argv ) {
             send_actions(std::move(actions), 10000, packed_transaction::zlib);
          }
       } else {
-         std::cout << "Skipping set abi because the new abi is the same as the existing abi" << std::endl;
+         std::cerr << "Skipping set abi because the new abi is the same as the existing abi" << std::endl;
       }
```
heifner: Wrap in `localized`

Issue #6190, opened 2018-10-31
eos --hard-replay is so slow


bitcodernull:
if i don't save old eos chain data, how can i speed up the replay

it is so slowly now.

jgiszczak:
The major EOS.IO-based networks generally provide a block log downloadable from some service using traditional protocols.  This can be faster than live syncing.

If you still have the problem and wish to discuss further, please post it in [https://eosio.stackexchange.com](https://eosio.stackexchange.com). To keep things organized, Github issues are intended for bugs reporting and features tracking, and Stack Exchange is the right place for any technical support discussions. Thanks!

matthewdarwin:
If you're on mainnet, make sure you have lots of RAM (256GB at least), or something near RAM (optane)

PR #6197, opened 2018-10-31
Replay complete, delete the unuseful backup dir.

evsward:
## delete backup dir after replay

**change**
Add oneline code after replay complete.
```
fc::remove_all(backup_dir)
```

**two reasons for change:**
- The old 'backup_dir' data is wasting of space. 
- The operator just want recover the 'blocks.log' as the same with the old one.  It's puzzled about the complex directory structure. 

Related the space issue, maybe lead to [GH-6190](https://github.com/EOSIO/eos/issues/6190)

**potential impact**
If this PR was merged, I think there's no impact to the current blockchain.  Because the  old 'backup_dir' data is recovered into the real 'blocks.log'.

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -495,6 +495,7 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
                fc::copy( backup_dir / config::reversible_blocks_dir_name / "shared_memory.meta",
                          my->chain_config->blocks_dir / config::reversible_blocks_dir_name / "shared_memory.meta" );
             }
+          fc::remove_all( backup_dir ); // Replay complete, delete the unuseful backup_dir.
          }
       } else if( options.at( "replay-blockchain" ).as<bool>()) {
          ilog( "Replay requested: deleting state database" );


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -495,6 +495,7 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
                fc::copy( backup_dir / config::reversible_blocks_dir_name / "shared_memory.meta",
                          my->chain_config->blocks_dir / config::reversible_blocks_dir_name / "shared_memory.meta" );
             }
+          fc::remove_all( backup_dir ); // Replay complete, delete the unuseful backup_dir.
```
heifner: - Replay is not complete at this point, so the comment is wrong.
- If `truncate-at-block` is specified that backup can be useful.
- If backups need to be deleted that can be done via scripts, not sure we want to remove backup_dir unless we know for sure it is no longer needed.

evsward: - It is complete because functions are synchronous.
- ```truncate-at-block``` is only useful with --fix-reversible-blocks without a replay(**no ```backup_dir```**) or with --hard-replay-blockchain.
- Once the data is recovered, the old backup is useless.  Lots of backups lead to a complex data structure and space wasting.

wanderingbort: The actual replay doesn't happen until later.  What has happened by this point is `block_log::repair_log` and `recover_reversible_blocks`.  The comment is erroneous. 

I agree with @heifner  that the backups have value and can be cleaned up out-of-process.  That said, I would entertain a CLI option that prevents the backup from persisting after the process.

evsward: I see.  But the CLI option is debatable because we can delete the backup through the OS.  Unless some validations need to be done before.

heifner:
I take @evsward comment to indicate no intention to add CLI, therefore I'm going to close this issue. Please forgive me if I'm reading this wrong. Thank you @evsward for your pull request. We will happily entertain another with a CLI option that prevents the backup from persisting.

PR #6208, opened 2018-11-01
Build eos image for eos

benbenben1010:
**Change Description**

Buildkite was building the eos-dev image and pushing it to gcr as eos.
Instead, build eos and push it as eos.

**Consensus Changes**

None

**API Changes**

None

**Documentation Additions**

None


zreyn:
Grabbing this review from Eric, since I know about this change.  This should have no effect on anything currently running; it just basically makes sure we don't have unnecessarily bloated images when in the non-dev environment.

Issue #3537, opened 2018-05-29
Several problems with the script "bios-boot-tutorial.py"


trungtt1981:
I'd like to inform several problems with the script `bios-boot-tutorial.py` in `$EOS_ROOT_DIR/tutorials/bios-boot-tutorial`.

1. `unlockTimeout = 99999999999`

This doesn't work out and lead to the following error. When changing to another smaller value such as `unlockTimeout = 1000`, then it works fine.

```
Creating wallet: default
Save password to use in the future to unlock this wallet.
Without password imported keys will not be retrievable.
"PW5Kbrn46xYb5U46Jq3u7FTeBbkbudFCRiKF4WbtH7H79QRe9FFRv"
1673153ms thread-0   http_plugin.cpp:369           handle_exception     ] FC Exception encountered while processing wallet.import_key: 3120003 wallet_locked_exception: Locked wallet
Wallet is locked: default
    {"w":"default"}
    thread-0  wallet_manager.cpp:172 import_key
Error 3120003: Locked wallet
Ensure that your wallet is unlocked before using it!
Error Details:
Wallet is locked: default
```

2. 
In the below function

```
def getJsonOutput(args):
    print('bios-boot-tutorial.py:', args)
    logFile.write(args + '\n')
    proc = subprocess.Popen(args, shell=True, stdout=subprocess.PIPE)
    return json.loads(proc.communicate()[0])
```
The code line `return json.loads(proc.communicate()[0])` has caused the following error for python 3.6.3 with GCC 7.1.0 on Ubuntu 16.04.

`the JSON object must be str, not 'bytes'`

A work-around is to change that code line as below:

`return json.loads(proc.communicate()[0].decode('utf-8'))`

Please be noted that on Ubuntu 16.04, it's impossible to upgrade to GCC 7.2.0

3.
Compiler error with this code line `factor = 1_000_000_000 / sum(dist)` for python 3.6.3 with GCC 7.1.0 on Ubuntu 16.04.

Work-around is to change to `factor = 1000000000 / sum(dist)`

taokayan:
We haven't received any update for this. Please test with the latest version as this seems to be too old. If you still have the problem and wish to discuss further, please post it in https://eosio.stackexchange.com

PR #6215, opened 2018-11-02
fixed external node can't access the node service, and the json parse…

littleLj:
**Change Description**
1、fixed external node can't access the  service
2、fixed err  [https://github.com/EOSIO/eos/issues/3537](url)


littleLj:
it's more easy  for user to build a  testing net  quickly.

Issue #6214, opened 2018-11-01
A transaction that will normally fail to execute due to a uniqueness violation, if submitted as a deferred transaction will partially execute instead of failing entirely as expected


acoutts:
When submitting a transaction that will cause a uniqueness constraint violation, it will fail immediately when submitting the transaction through cleos as expected. But submitting the exact same transaction as a deferred transaction causes some of the actions to succeed according to the history plugin and signals from the chain plugin while throwing the uniqueness constraint violation in nodeos:

```
#   19   2018-11-01T19:47:06.000     chintailease::prepare => chintailease  cc941cbb... {"memo":"useraccount2|1|30|0.1000|0|0.5000|cqfcymm6p2"}...
#   20   2018-11-01T19:47:06.000     eosio.token::transfer => chintailease  cc941cbb... {"from":"useraccount2","to":"chintailease","quantity":"3.000...
#   21   2018-11-01T19:47:06.000    chintailease::activate => chintailease  cc941cbb... {"memo":"useraccount2|1|30|0.1000|0|0.5000|cqfcymm6p2"}...
#   22   2018-11-01T19:47:42.000  chintailease::processpool => chintailease  555148f7... "00"...
#   23   2018-11-01T19:48:57.000  chintailease::processpool => chintailease  9828235f... "00"...
#   24   2018-11-01T19:50:12.000  chintailease::processpool => chintailease  d4e08b44... "00"...
#   25   2018-11-01T19:51:27.000  chintailease::processpool => chintailease  1467e696... "00"...
#   26   2018-11-01T19:52:42.000  chintailease::processpool => chintailease  1c51d943... "00"...
#   27   2018-11-01T19:53:25.000     chintailease::prepare => chintailease  606788ba... {"memo":"useraccount2|1|30|0.1000|0|0.5000|cqfcymm6p2"}...
#   28   2018-11-01T19:53:57.000  chintailease::processpool => chintailease  0c79cf6a... "00"...
```

Line 27 is an attempt to submit a duplicate transaction already done in line 19.

I haven't confirmed yet if any contract tables were actually changed or if it's just a bug in the history plugin where it shouldn't have recorded the second transaction attempt at line 27 due to it failing for uniqueness constraint violation.

Here is the nodeos output including some tags from the different signals in the chain plugin showing that the action trace was visible in accepted/applied transaction even though the transaction should have failed:
```
[...]

PR #6220, opened 2018-11-02
failed deferred transactions were in history (6214)

tbfleming:
**Change Description**

The history plugin sometimes included failed deferred transactions in its history. Since it doesn't store receipt status, there wasn't a reliable way for users of the history API to know which actions were executed. #6214 

As a reminder, the history plugin is deprecated. This fix is included because the bug opens an attack vector against apps which rely on data from the history plugin to make decisions.

**Documentation Additions**

This is a behavior change. Previously `get_transaction` could retrieve failed or expired deferred transactions.

### DIFF
--- a/plugins/history_plugin/history_plugin.cpp
+++ b/plugins/history_plugin/history_plugin.cpp
@@ -282,6 +282,8 @@ namespace eosio {
          }
 
          void on_applied_transaction( const transaction_trace_ptr& trace ) {
+            if( !trace->receipt || trace->receipt->status != transaction_receipt_header::executed )
+               return;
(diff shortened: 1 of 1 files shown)


plugins/history_plugin/history_plugin.cpp
```diff
@@ -282,6 +282,8 @@ namespace eosio {
          }
 
          void on_applied_transaction( const transaction_trace_ptr& trace ) {
+            if( !trace->receipt || trace->receipt->status != transaction_receipt_header::executed )
```
arhag: It should not reject `soft_fail` either since that causes state changes.

PR #6234, opened 2018-11-05
Fix bug:default producer vote num is bigger than actual producers num.

evsward:
## Bug description
Executing command:
> ./bios-boot-tutorial.py -a --user-limit 30 --producer-limit 3

At the stage of voting, program will call function vote within:
> random.sample(population, k)：

- The first param 'population' is ```range(firstProducer, firstProducer + numProducers)```, equals (30,31,32).
- The second param 'k' is ```args.num_producers_vote```, use the default value which is 20, because we haven't specified in the command.

There is an issue here:
## Changes
I add a variate named k in the function vote.  According the producers number, the k will self-adaption.


tbfleming:
Thanks!

PR #6249, opened 2018-11-06
Reduce noise

heifner:
**Change Description**

- Remove debug log messages seen during replay and normal processing that only confuse


spoonincode:
>Has there been any thought put into actually having debug statements in the code and not turning on debug output in general, so that someone could start up a node with debug turned on to then get more info, rather then turning on everything and removing any extra info from the code?

That has somewhat been accomplished by having more than one logger -- for example the producer plugin logger and the net plugin logger. Anything other than the default logger is disabled by default.

PR #6262, opened 2018-11-07
Add get_table_rows key

fingera:
**Add get_table_rows key**
--key means exactly match

ex1:
I want to get which BP votes are 0
`cleos get table eosio eosio producers --index 2 --key-type float64 -k 0`

ex2:
I want to  get which BP names 111112222244
`cleos get table eosio eosio producers -k 111112222244`

OR

[lower_bound, upper_bound) => [lower_bound, upper_bound]

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -405,24 +405,37 @@ public:
          auto lower = secidx.lower_bound(boost::make_tuple(low_tid));
          auto upper = secidx.lower_bound(boost::make_tuple(next_tid));
 
-         if (p.lower_bound.size()) {
+         auto lower_bound = p.lower_bound;
+         auto upper_bound = p.upper_bound;
+         if (p.table_key.size()) {
+            lower_bound = p.table_key;
+            upper_bound = p.table_key;
+         }
+
+         if (lower_bound.size()) {
             if (p.key_type == "name") {
-               name s(p.lower_bound);
+               name s(lower_bound);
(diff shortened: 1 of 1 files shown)


UMU618:
Why -k,--key TEXT is Deprecated?
I barely understand.

fingera:
ex1:
I want to get which BP votes are 0
cleos get table eosio eosio producers --index 2 --key-type float64 -k 0
no --key:
cleos get table eosio eosio producers --index 2 --key-type float64 -L 0 -U 0.000000001
If the second index is name and the results more than one:
cleos get table NAME SCOPE TABLE --index 2 --key-type name -L aaaaaaaaaaa -U aaaaaaaaaaa+1?

ex2:
I want to get which BP names 111112222244
cleos get table eosio eosio producers -k 111112222244
no --key:
cleos get table eosio eosio producers -L 111112222244 --limit 1
If there is no 111112222244 , we will still get 111112222245 or something else.

I means --key equal to [key, key], but -L -U is [key, key)
--key does not mean that there is only one data.

I don't want to change the semantics of API.

arhag:
> I means --key equal to [key, key], but -L -U is [key, key)
--key does not mean that there is only one data.
>
> I don't want to change the semantics of API.

I consider the current semantics of `upper_bound` to be a bug. PR #6285 changes the semantics of `upper_bound` to be what they always should have been (same semantics as [`std::upper_bound`](https://en.cppreference.com/w/cpp/algorithm/upper_bound)).

If PR #6285 is merged in, it will make this issue unnecessary since it will allow you to do what you want to do by just passing in the same key value for both the `lower_bound` and `upper_bound` fields of the RPC.

arhag:
Made unnecessary due to #6285.

fingera:
Good.
Maybe remove cleos --key Deprecated because is not implement.
Or let --key override -L -U in server?

PR #6280, opened 2018-11-09
remove repeat register of history_plugin

Smalinuxer:
Remove repeat register of history_plugin from nodeos.

### DIFF
--- a/programs/nodeos/main.cpp
+++ b/programs/nodeos/main.cpp
@@ -6,7 +6,6 @@
 
 #include <eosio/chain_plugin/chain_plugin.hpp>
 #include <eosio/http_plugin/http_plugin.hpp>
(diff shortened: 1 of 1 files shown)


conr2d:
If the repeated line you mentioned is https://github.com/EOSIO/eos/blob/master/programs/nodeos/CMakeLists.txt#L55, I think it's not for loading history_plugin in runtime, but it tries link `nodeos` to `history_plugin` during compilation. In that case, `nodeos` doesn't load `history_plugin` repeatedly.

heifner:
It is actually registered via:
https://github.com/EOSIO/eos/blob/master/plugins/history_plugin/history_plugin.cpp#L17

You are correct that explicit register in main is not needed.

Smalinuxer:
okay , misunderstand about that . but same result . 

hah ! very nice .

PR #6285, opened 2018-11-09
Fix upper_bound bug in chain_plugin

### LINKED ISSUE #6273: fix upper_bound bug in chain_plugin; update cleos to use fixed upper_bound correctly
See the comment here: https://github.com/EOSIO/eos/pull/6070#issuecomment-436807613

~~Change `get_table_rows_ex` to set `upper` (within the if statement checking if `upper_bound` was specified by the caller) using the `upper_bound` function rather than the `lower_bound` function.~~

~~Change `get_table_rows_by_seckey` to set `upper` (within the if statement checking if `upper_bound` was specified by the caller) using the `upper_bound` function rather than the `lower_bound` function. However the `upper_bound` function should be passed a tuple that includes a third argument `std::numeric_limits<uint64_t>::max()`.~~

~~Change `get_table_by_scope` to set `upper` (within the if statement branch that asserts that `upper_bound` was specified by the caller) using the `upper_bound` function rather than the `lower_bound` function and with a tuple `boost::make_tuple(p.code, scope, std::numeric_limits<uint64_t>::max())` passed to the `upper_bound` function.~~

Set `lower_bound_lookup_tuple` and `upper_bound_lookup_tuple` appropriately based on passed in RPC arguments. Ensure `lower_bound_lookup_tuple <= upper_bound_lookup_tuple`. Then the `lower` and `upper` iterators in all cases should just be:
```
auto lower = idx.lower_bound( lower_bound_lookup_tuple );
auto upper = idx.upper_bound( upper_bound_lookup_tuple );
```

Audit all uses of `lower_bound` and `upper_bound` in cleos and update them as appropriate to work with the corrected `upper_bound` implementation in chain_plugin.

arhag:
**Change Description**

Resolves #6273.

PR #6070 revealed an existing bug in the implementation of `chain_plugin` with regards to how it handles the `upper_bound` field in the `get_table_rows` and `get_table_by_scope` RPC calls.

The intended definition for this `upper_bound` field is similar to how it is defined in the [`std::upper_bound` function in the C++ standard library](https://en.cppreference.com/w/cpp/algorithm/upper_bound): the first element within the specified ordered range with a value _greater than_ the given comparison value.

Instead, the implementation in `chain_plugin` handled `upper_bound` as if the definition was the same as that of `lower_bound`: the first element within the specified ordered range with a value _greater than or equal to_ the given comparison value.

The intended definition has the advantage of being consistent with the standard definition of `upper_bound` well known to C++ programmers and it also enables callers to request an [`equal_range`](https://en.cppreference.com/w/cpp/algorithm/equal_range) by simply passing in the same comparison value for both the `lower_bound` and `upper_bound` fields. Attempting to do this with the old definition required knowing and calculating the value immediately succeeding the `lower_bound` value and passing that in as the `upper_bound` field (as PR #6070 did within cleos). Passing in the same comparison value for both the `lower_bound` and `upper_bound` using the old definition would result in no ranges ever returned.

The existing implementation of `chain_plugin` has a further bug (which this PR also fixes) in the case where `upper_bound` is less than `lower_bound`. The correct behavior in such a situation is to never return any results.

This PR fixes the bug in the case where `upper_bound` is less than `lower_bound` and it also changes the implementation of `chain_plugin` to handle the `upper_bound` field according to the intended definition rather than the existing definition. The implications of these changes to users of the chain API are described in detail under the "API Changes" section.

Furthermore, this PR adds back the additional checks in cleos for the `system voteproducer approve`, `system voteproducer unapprove`, `multisig review` sub-commands that were removed in #6070 and ensure the results returned from the chain API RPC are exactly what was requested. It also adds a similar check for the `system bidnameinfo` sub-command (similar to what was in #6068) to fix the bug described in #6067. These additional checks are necessary (even though they were not necessary in PR #6070) because this PR keeps the changes to the `upper_bound` value specified by cleos made in PR #6070 in which the `upper_bound` value is the value immediately succeeding the `lower_bound` value; that `upper_bound` workaround was kept just so that the version of cleos incorporating the changes in this PR can still behave correctly when talking to an old version of nodeos that has not incorporated the changes in this PR. (Also note that an old cleos at version v1.4.x or older would still behave correctly with the `system voteproducer approve`, `system voteproducer unapprove`, and `multisig review` sub-commands even when talking to a new version of nodeos that incorporates the changes in this PR. However, in such a situation, the bug described in #6067 would still of course apply when using `system bidnameinfo`.)

**Consensus Changes**

None

**API Changes**

Users of the `get_table_rows` and `get_table_by_scope` RPCs of the chain API are unaffected by these changes if they are not already using the `upper_bound` field. Existing users of the `upper_bound` field of those two RPCs should read on to understand the changes introduced by this PR.

**`get_table_rows`**

*First some background on `get_table_rows`:*

A particular index type (either the primary key type or one of the five secondary index types) is selected (whether implicitly or explicitly using the `key_type` field) in the `get_table_rows` call. The index type defines the type of the index key within table records to use to sort the table rows in the table for the purposes of this RPC.

A particular index ordinal is selected (whether implicitly or explicitly using the `index_position` field). This index ordinal must correspond to the selected index type (based on how the tables are structured in the persistent state on the blockchain), and the selected index ordinal along with the selected index type (together can be referred to as the selected index) determine how the tables rows in the table are ordered. For selected indices that are secondary indices, the sort order is based first on the index key (in ascending order) and then ties are broken by the primary key (in ascending order).

A particular table is specified by: the index type, index ordinal, and the tuple `(code, scope, table)`. The rows in the table have an index key (of index type) which is the value used to sort the order of the table rows within the selected index.

If `lower_bound` is specified, let `candidate_rows` be the set of table rows in the specified table which have an index key greater than or equal to the value specified in `lower_bound`. Otherwise, let `candidate_rows` be the set of all table rows in the specified table.

*Behavior prior to this PR:*

If `upper_bound` is specified, let `excluded_rows` be the set of table rows in the specified table which have an index key greater than or equal to the value specified in `upper_bound`. Otherwise, let `excluded_rows` be the empty set.

If the value specified in `upper_bound` is less than the value specified in `lower_bound`, the behavior of `get_tables_rows` is undefined. Otherwise, `get_table_rows` returns the set of tables rows (in sorted order) that are in `candidate_rows` but not in `excluded_rows`.

*Behavior as of this PR:*

If `upper_bound` is specified, let `excluded_rows` be the set of table rows in the specified table which have an index key greater than the value specified in `upper_bound`. Otherwise, let `excluded_rows` be the empty set.

If the value specified in `upper_bound` is less than the value specified in `lower_bound`, `get_table_rows` returns no rows. Otherwise, `get_table_rows` returns the set of tables rows (in sorted order) that are in `candidate_rows` but not in `excluded_rows`.

*Implications to existing users of `upper_bound`:*

- Users should not be calling `get_table_rows` with an `upper_bound` less than `lower_bound`. If they currently are doing so and using the returned results, they are relying on undefined behavior. This PR fixes that undefined behavior so that a logically consistent result is returned should users continue to make this inappropriate call.
- Users of `get_table_rows` that call it with an `upper_bound` value greater than or equal to the `lower_bound` value may get back less rows with the new behavior than they received with the old behavior. If users adjust their call by setting `upper_bound` to the value immediately succeeding the value they provided originally, they will receive no less rows with the new behavior than they did before with the old behavior (they can then further filter on the client side).

**`get_table_by_scope`**

*First some background on `get_table_by_scope`:*

The tables within a given `code` account are sorted (for the purposes of `get_table_by_scope`) first by `scope` (in ascending order) and then ties are broken by `table` name (in ascending order).

If `lower_bound` is specified, let `candidate_tables` be the set of tables in the specified `code` account which have `scope` greater than or equal to the value specified in `lower_bound`. Otherwise, let `candidate_tables` be the set of all tables in the specified `code` account.

If the `table` is not specified in the RPC (or its value is 0), then let `excluded_tables1` be the empty set. Otherwise, let `excluded_tables1` be the set of tables in the specified `code` account which have a `table` name equal to the specified `table` value in the RPC.

*Behavior prior to this PR:*

If `upper_bound` is specified, let `excluded_tables2` be the set of tables in the specified `code` account which have a `scope` greater than or equal to the value specified in `upper_bound`. Otherwise, let `excluded_tables2` be the empty set.

If the value specified in `upper_bound` is less than the value specified in `lower_bound`, the behavior of `get_table_by_scope` is undefined. Otherwise, `get_table_by_scope` returns the set of tables (in sorted order) that are in `candidate_tables` but not in the union of `excluded_tables1` and `excluded_tables2`.

*Behavior as of this PR:*

If `upper_bound` is specified, let `excluded_tables2` be the set of tables in the specified `code` account which have a `scope` greater than the value specified in `upper_bound`. Otherwise, let `excluded_tables2` be the empty set.

If the value specified in `upper_bound` is less than the value specified in `lower_bound`, `get_table_by_scope` returns no tables. Otherwise, `get_table_by_scope` returns the set of tables (in sorted order) that are in `candidate_tables` but not in the union of `excluded_tables1` and `excluded_tables2`.

*Implications to existing users of `upper_bound`:*

- Users should not be calling `get_table_by_scope` with an `upper_bound` less than `lower_bound`. If they currently are doing so and using the returned results, they are relying on undefined behavior. This PR fixes that undefined behavior so that a logically consistent result is returned should users continue to make this inappropriate call.
- Users of `get_table_by_scope` that call it with an `upper_bound` value greater than or equal to the `lower_bound` value may get back less tables with the new behavior than they received with the old behavior. If users adjust their call by setting `upper_bound` to the value immediately succeeding the value they provided originally, they will receive no less tables with the new behavior than they did before with the old behavior (they can then further filter on the client side).

**Documentation Additions**

Documentation about the `get_table_rows` and `get_table_by_scope` RPCs of the chain API may need to be updated to reflect the new definition of `upper_bound` and the implications of that as discussed in the "API Changes" section above.

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -43,6 +43,75 @@ namespace eosio {
    using chain::abi_def;
    using chain::abi_serializer;
 
+   template<typename T>
+   struct secondary_key_traits {
+      using value_type = std::enable_if_t<std::is_integral<T>::value, T>;
+
+      static_assert( std::numeric_limits<uint128_t>::is_specialized, "value_type does not have specialized numeric_limits" );
+
+      static constexpr value_type true_lowest() { return std::numeric_limits<value_type>::lowest(); }
+      static constexpr value_type true_highest() { return std::numeric_limits<value_type>::max(); }
+   };
+
+   template<size_t N>
+   struct secondary_key_traits<std::array<uint128_t, N>> {
+   private:
+      static constexpr uint128_t max_uint128 = uint128_t(std::numeric_limits<uint64_t>::max()) << 64 | std::numeric_limits<uint64_t>::max();
+      static_assert( std::numeric_limits<uint128_t>::max() == max_uint128, "numeric_limits for uint128_t is not properly defined" );
+
+   public:
+      using value_type = std::array<uint128_t, N>;
+
+      static value_type true_lowest() {
+         value_type arr;
+         return arr;
+      }
+
+      static value_type true_highest() {
+         value_type arr;
+         for( auto& v : arr ) {
+            v = std::numeric_limits<uint128_t>::max();
+         }
+         return arr;
+      }
+   };
+
+   template<>
+   struct secondary_key_traits<float64_t> {
+      using value_type = float64_t;
+
+      static value_type true_lowest() {
+         float64_t f;
+         f.v = 0xff00000000000000ull;
+         return f; // -infinity?
+      }
+
+      static value_type true_highest() {
+         float64_t f;
+         f.v = 0x7f00000000000000ull;
+         return f; // +infinity?
+      }
+   };
+
+   template<>
+   struct secondary_key_traits<float128_t> {
+      using value_type = float128_t;
+
+      static value_type true_lowest() {
+         float128_t f;
+         f.v[0] = 0x0ull;
+         f.v[1] = 0xffff000000000000ull;
+         return f; // -infinity?
+      }
+
+      static value_type true_highest() {
+         float128_t f;
+         f.v[0] = 0x0ull;
+         f.v[1] = 0x7fff000000000000ull;
+         return f; // +infinity?
+      }
+   };
+
 namespace chain_apis {
 struct empty{};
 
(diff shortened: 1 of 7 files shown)


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -398,34 +401,46 @@ class read_only {
...
       const auto* index_t_id = d.find<chain::table_id_object, chain::by_code_scope_table>(boost::make_tuple(p.code, scope, table_with_index));
-      if (t_id != nullptr && index_t_id != nullptr) {
-         const auto& secidx = d.get_index<IndexType, chain::by_secondary>();
-         decltype(index_t_id->id) low_tid(index_t_id->id._id);
-         decltype(index_t_id->id) next_tid(index_t_id->id._id + 1);
-         auto lower = secidx.lower_bound(boost::make_tuple(low_tid));
-         auto upper = secidx.lower_bound(boost::make_tuple(next_tid));
+      if( t_id != nullptr && index_t_id != nullptr ) {
+         using secondary_key_type = std::result_of_t<decltype(conv)(SecKeyType)>;
+         static_assert( std::is_same<decltype(IndexType::value_type::secondary_key), secondary_key_type>::value, "Return type of conv does not match type of secondary key for IndexType" );
```
wanderingbort: isn't `decltype(IndexType::value_type::secondary_key)` == `IndexType::value_type::secondary_key_type` ? or was this just less characters because of the need for `typename`?

arhag: Yeah, I guess I didn't notice the `secondary_key_type` typedef. Either way works.

znebby:
@arhag I am confused about this quote

> Users of get_table_by_scope that call it with an upper_bound value greater than or equal to the lower_bound value may get back less tables with the new behavior than they received with the old behavior. If users adjust their call by setting upper_bound to the value immediately succeeding the value they provided originally, they will receive no less tables with the new behavior than they did before with the old behavior (they can then further filter on the client side).

Is it not the opposite? Since upper_bound is now exclusive, rather than inclusive. For example, if I was searching on a uint64 index and let's say I wanted all rows where the index = 0. Previously, I would set lower_bound to 0 and upper_bound to 1. With this new behaviour, if I understood correctly, I can keep both lower_bound and upper_bound to 0.

So, if I didn't change my code and I left lower to 0 and upper to 1, I would now receive all rows with the index value being 0 or 1. That is _more_ rows, not less. Is there a misunderstanding somewhere?

arhag:
@znebby:

> Is it not the opposite? Since upper_bound is now exclusive, rather than inclusive. For example, if I was searching on a uint64 index and let's say I wanted all rows where the index = 0. Previously, I would set lower_bound to 0 and upper_bound to 1. With this new behaviour, if I understood correctly, I can keep both lower_bound and upper_bound to 0.
> 
> So, if I didn't change my code and I left lower to 0 and upper to 1, I would now receive all rows with the index value being 0 or 1. That is _more_ rows, not less. Is there a misunderstanding somewhere?

You are understanding correctly. Perhaps my wording was just unclear.

What I was trying to say, continuing with your example, was that if your client was updated to now set both `lower_bound` and `upper_bound` to 0 in order to get all table rows with a key of 0, it would work fine if talking to the new nodeos, but would return back less results (in this case no results) if the same request was made to the old nodeos.

znebby:
Ah when talking to an old nodeos. Got it.

PR #6287, opened 2018-11-09
sudo removed from yum info (manually tested on centos 7 VM)

NorseGaud:
(Second attempt at branch of develop)
Removing threat stack warnings about sudo being used for yum info when it's not necessary. Per Brian M.


heifner:
@zreyn @ericiles I'm assuming this is good. Can I get a review?

PR #6293, opened 2018-11-11
Add ability to sign arbitrary messages to cleos

k26dr:
Right now cleos can only sign transaction messages with a private key. I've modified cleos so it can sign arbitrary messages as well with the `cleos sign message` command. 

There are 2 changes to the cleos API: 

1. `cleos sign` is now `cleos sign transaction`
2. `cleos sign message` is a new command 

Documentation will have to be changed to reflect these changes.

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -2872,7 +2872,7 @@ int main( int argc, char** argv ) {
          std::getline( std::cin, str_private_key, '\n' );
          fc::set_console_echo(true);
       }
-      auto priv_key = fc::crypto::private_key::regenerate(*utilities::wif_to_key(str_private_key));
+      auto priv_key = private_key_type(str_private_key);
       fc::sha256 digest = fc::sha256::hash(trx_json_to_sign.c_str(), trx_json_to_sign.length());
       fc::crypto::signature sig = priv_key.sign( digest );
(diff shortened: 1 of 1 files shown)


k26dr:
We're building out the UTXO on EOS model that Dan brought up a while ago, and it requires us to sign arbitrary messages so we built this feature out. It's more for our personal use than anything, I don't see who else would find this useful at the moment. We're probably better off maintaining a separate branch with the feature for those who want to use it.

wanderingbort:
> We're probably better off maintaining a separate branch with the feature for those who want to use it.

I see.  Thanks again for responding.  I agree with the above statement.  This seems like a domain specific feature that is more appropriate in the UI/UX for that domain.  If you are in an environment where javascript is usable,  [eosjs](https://github.com/EOSIO/eosjs) and its dependencies provide signing facilities that can be used to construct such a UI, you can also investigate using @eoscanada 's [Go Library](https://github.com/eoscanada/eos-go) .

Right now **this** repository does not provide a clean abstraction for using its internals as libraries.  Perhaps we can one day offer it in the future but I cannot, in good conscious, suggest that you attempt to build the UI/UX/Tool to support this use case in this repo today.

Good luck, feel free to reach out on telegram if we can help you with your project!

PR #6308, opened 2018-11-13
Explicitly throw away get_table_type return value for code clarity

### LINKED ISSUE #6276: Two pieces of unused code
In the get_currency_balance function on line 1209 of the chain_plugin.cpp file
```
vector<asset> read_only::get_currency_balance( const read_only::get_currency_balance_params& p )const {

   const abi_def abi = eosio::chain_apis::get_abi( db, p.code );
   auto table_type = get_table_type( abi, "accounts" );
```
I didn't find the table_type variable used in this function.
Are these two pieces of code redundant?

jgiszczak:
**Change Description**

Resolves #6276.

get_table_type() is called in order to generate more readable errors in the case the 'accounts' table does not exist when querying currency balances.

**Consensus Changes**

N/A

**API Changes**

N/A

**Documentation Additions**

N/A


conr2d:
`chain_plugin.cpp:1278` has same line for 'stat'.

PR #6362, opened 2018-11-20
Pull in changes to build scripts that were merged to develop

emorybarlow:
**Change Description**

Pulls in changes to build scripts that were merged into develop

**Consensus Changes**

**API Changes**

**Documentation Additions**

### DIFF
--- a/scripts/generate_bottle.sh
+++ b/scripts/generate_bottle.sh
@@ -3,7 +3,7 @@
 VERS=`sw_vers -productVersion | awk '/10\.13\..*/{print $0}'`
 if [[ -z "$VERS" ]];
 then
-   VERS=`sw_vers -productVersion | awk '/10\.14\..*/{print $0}'`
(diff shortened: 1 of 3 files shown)


scripts/generate_bottle.sh
```diff
@@ -3,7 +3,7 @@
 VERS=`sw_vers -productVersion | awk '/10\.13\..*/{print $0}'`
 if [[ -z "$VERS" ]];
 then
-   VERS=`sw_vers -productVersion | awk '/10\.14\..*/{print $0}'`
+   VERS=`sw_vers -productVersion | awk '/10\.14.*/{print $0}'`
```
wanderingbort: Are there version strings that omit the dot after `10.14`, if so, is that the same for line 3 where we have `10.13.` as the literal match?

emorybarlow: For some reason the mojave version string is just `10.14` where as high sierra is `10.13.xx`. So in the case of high sierra we need that extra dot.

PR #6365, opened 2018-11-20
net_plugin should use plugin-local logger instead of default logger

v-gu:
**Change Description**

In case some user want to exclude those logs, there is no means to disable it except to turn 'off' all EOS logging output of the 'default' logger.

**Consensus Changes**

None.

**API Changes**

None.

**Documentation Additions**

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -704,7 +704,7 @@ namespace eosio {
         fork_head_num(0),
         last_req()
    {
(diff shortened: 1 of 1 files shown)


wanderingbort:
the `peer_?log` pattern is only valid when the message is coming from the context of a connected peer.  We should not remove it.  But we can add an additional version of the macro that does not require a peer and still wraps the logger for cases where the log messages are produced without an associated peer if there are such cases.

heifner:
Replaced by #6803

Issue #6305, opened 2018-11-13
Need a function to check if it is a contract account


jin10086:
So I can refuse the contract attack

jin10086:
Now basically all the random numbers are broken, which are all implemented by the contract account.
If I can in smart contract check if it is a contract account  , I can reject the request.

evsward:
- RPC请求get_account -> last_code_update.  如果最后code修改时间不是genesis_timestamp，那这个账户就是合约账户。  
- 或者使用cleos get code {account_name}，查看某账户的code值，如果不是0则证明该账户是合约账户。

jin10086:
@evsward  谢谢回复，外部的话，我当然可以拿到...
我想要的是 在合约内 判断...

evsward:
没有文档，兄弟，需要自己尝试调用一下，你可以参考read_only::get_code源码。在合约中通过状态库查询到账户的信息，然后判断返回对象的code.size()是否不为空，是的话则证明是合约账户。

jin10086:
好的，谢谢啦.我尝试一下，然后成功的话，把代码贴出来，方便后面的兄弟...

flyq:
我也尝试了。
参考：
https://github.com/EOSIO/eos/blob/master/plugins/chain_plugin/chain_plugin.cpp#L1578

```cpp
        const auto& d = db.db();
        const auto& accnt  = d.get<account_object,by_name>( params.account_name );        
    
        if(accnt.code.size()) {
          print("Msg from a account with contract, you may be a hacker.", name{user});
        } else {
          print("Msg from a account without contract, you are a good man.", name{user});
        }
```
   
但是因为引入了```eosio/chain/*```里面的头文件，编译不通过

evsward:
因为合约的执行是与区块链分隔开的，因此合约不能直接使用节点的环境变量。所以你要尝试在合约中自己构建db对象，请参考

https://github.com/EOSIO/eos/blob/master/libraries/chain/controller.cpp#L173

taokayan:
> Now basically all the random numbers are broken, which are all implemented by the contract account.
> If I can in smart contract check if it is a contract account , I can reject the request.

This is unlikely to solve your fundamental problem, as a hacker can run a modified local node not to broadcast his transactions that will potentially lose the games.

fingera:
Even if send_defered twice, BP can still control HASH, but maybe safe for small amounts.

jin10086:
Someone helped create a contract database.
[...]

PR #6379, opened 2018-11-23
9 add is contract api

fingera:
**Change Description**
solve  #6305 

**Consensus Changes**
new contract api hardfork

**API Changes**

add new api is_contract

**Documentation Additions**


arhag:
Also, and I'm addressing this to any EOSIO contract developers who may be reading this, if `get_code_version` is eventually added in a future consensus protocol upgrade, please don't try to use it to "fix" issue #6305. Finding out whether an account has a contract deployed on it or not is not going to fix a broken random number generator. 

There are appropriate and inappropriate ways to get a proper source of entropy for random number generation in a smart contract. The appropriate ways require a commit-and-reveal process by enough non-colluding parties.

PR #6397, opened 2018-11-27
net_plugin reduce signed_block copies

heifner:
**Change Description**

- Made signed_block copy constructor private so the compiler will prevent accidental copies of signed_block
- Updated net_plugin to avoid unneeded signed_block copies

**Consensus Changes**

None

**API Changes**

net_plugin message format did NOT change.

**Documentation Additions**

None

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1528,7 +1528,7 @@ fc::variant read_only::get_block_header_state(const get_block_header_state_param
 
 void read_write::push_block(read_write::push_block_params&& params, next_function<read_write::push_block_results> next) {
    try {
-      app().get_method<incoming::methods::block_sync>()(std::make_shared<signed_block>(std::forward<signed_block>(params)));
+      app().get_method<incoming::methods::block_sync>()(std::make_shared<signed_block>(std::move(params)));
       next(read_write::push_block_results{});
    } catch ( boost::interprocess::bad_alloc& ) {
       chain_plugin::handle_db_exhaustion();


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1084,22 +1090,47 @@ namespace eosio {
...
+      size_t header_size = sizeof(payload_size);
+      size_t buffer_size = header_size + payload_size;
+
+      auto send_buffer = std::make_shared<vector<char>>(buffer_size);
+      fc::datastream<char*> ds( send_buffer->data(), buffer_size);
+      ds.write( header, header_size );
+      fc::raw::pack( ds, unsigned_int( which ));
+      fc::raw::pack( ds, *sb );
+
+      enqueue_buffer( send_buffer, true, no_reason );
```
arhag: Should be `enqueue_buffer( send_buffer, trigger_send, no_reason );`

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -569,7 +569,7 @@ class read_write {
 
    using push_block_params = chain::signed_block;
    using push_block_results = empty;
-   void push_block(const push_block_params& params, chain::plugin_interface::next_function<push_block_results> next);
+   void push_block(push_block_params params, chain::plugin_interface::next_function<push_block_results> next);
```
arhag: This could be `void push_block(push_block_params&& params, chain::plugin_interface::next_function<push_block_results> next);`

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1526,9 +1526,9 @@ fc::variant read_only::get_block_header_state(const get_block_header_state_param
    return vo;
 }
 
-void read_write::push_block(const read_write::push_block_params& params, next_function<read_write::push_block_results> next) {
+void read_write::push_block(read_write::push_block_params params, next_function<read_write::push_block_results> next) {
```
arhag: `read_write::push_block_params params` could be `read_write::push_block_params&& params` to avoid a copy.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -612,13 +613,18 @@ namespace eosio {
...
    };
 
-   struct msgHandler : public fc::visitor<void> {
+   struct msg_handler : public fc::visitor<void> {
       net_plugin_impl &impl;
       connection_ptr c;
-      msgHandler( net_plugin_impl &imp, connection_ptr conn) : impl(imp), c(conn) {}
+      msg_handler( net_plugin_impl &imp, connection_ptr conn) : impl(imp), c(conn) {}
+
+      void operator()( signed_block& msg ) const
```
arhag: Moving a passed in lvalue reference could break the expectations of the caller. If this was `void operator()( signed_block&& msg ) const` it could be called directly with a moved `signed_block`. Since `fc::static_variant` does not currently support visiting an rvalue reference of a `static_variant`, it does mean the visiting logic at the caller side has to be more complicated.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1197,7 +1228,7 @@ namespace eosio {
          auto ds = pending_message_buffer.create_datastream();
          net_message msg;
          fc::raw::unpack(ds, msg);
-         msgHandler m(impl, shared_from_this() );
+         msg_handler m(impl, shared_from_this() );
          msg.visit(m);
```
arhag: If `msg` contains `signed_block`, this will require special case logic to move it and pass that rvalue reference to the function call operator of the `m` visitor. Otherwise, it could use the regular `visit` method of `msg`.

Issue #6393, opened 2018-11-27
Why don't we stop unix_server from listening in shutdown stage as well?


evsward:
https://github.com/EOSIO/eos/blob/59626f1e6361df3b715e926ee13a9a8e84d177af/plugins/http_plugin/http_plugin.cpp#L550
add
```

      if(my->unix_server.is_listening())
         my->unix_server.stop_listening();
```

jgiszczak:
Which should also unlink the socket file.  Pull requests gratefully accepted.  Please branch from the `develop` branch.

spoonincode:
>We should also unlink the socket file

Already does that

spoonincode:
https://github.com/EOSIO/eos/blob/59626f1e6361df3b715e926ee13a9a8e84d177af/plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp#L205

spoonincode:
oh and here too, so it looks like `stop_listenting()` should remove it
https://github.com/EOSIO/eos/blob/59626f1e6361df3b715e926ee13a9a8e84d177af/plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp#L435

evsward:
I think the destructor is not enough.  The shutdown stage need explicit call.

spoonincode:
Fix merged

PR #6401, opened 2018-11-28
Stop the unix_server from listening in shutdown stage as well.

evsward:
**Change Description**
We should also unlink the socket file in shutdown stage.
#6393 

**API Changes**
none
**Documentation Additions**
none


evsward:
It's not  an observed issue.  It's a code cleanup I think.

PR #6430, opened 2018-12-04
add a test that reduces the size of the producer set

wanderingbort:
**Change Description**

Add an additional test to the producer_schedule_tests that decreases the size of the producer set from 3 to 2 

Add an additional test to the producer_schedule_tests that verifies that a proposed empty producer schedule cannot become active and will not cause gaps to form in the schedule versions of active producer schedules.

**Consensus Changes**

**API Changes**

**Documentation Additions**

### DIFF
--- a/unittests/producer_schedule_tests.cpp
+++ b/unittests/producer_schedule_tests.cpp
@@ -260,4 +260,61 @@ BOOST_FIXTURE_TEST_CASE( producer_schedule_promotion_test, TESTER ) try {
    BOOST_REQUIRE_EQUAL( validate(), true );
 } FC_LOG_AND_RETHROW()
 
+BOOST_FIXTURE_TEST_CASE( producer_schedule_reduction, tester ) try {
+   create_accounts( {N(alice),N(bob),N(carol)} );
+   produce_block();
+
+   auto compare_schedules = [&]( const vector<producer_key>& a, const producer_schedule_type& b ) {
+      return std::equal( a.begin(), a.end(), b.producers.begin(), b.producers.end() );
+   };
+
+   auto res = set_producers( {N(alice),N(bob),N(carol)} );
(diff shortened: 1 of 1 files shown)


unittests/producer_schedule_tests.cpp
```diff
@@ -260,4 +260,61 @@ BOOST_FIXTURE_TEST_CASE( producer_schedule_promotion_test, TESTER ) try {
...
+      return std::equal( a.begin(), a.end(), b.producers.begin(), b.producers.end() );
+   };
+
+   auto res = set_producers( {N(alice),N(bob),N(carol)} );
+   vector<producer_key> sch1 = {
+                                 {N(alice), get_public_key(N(alice), "active")},
+                                 {N(bob),   get_public_key(N(bob),   "active")},
+                                 {N(carol),   get_public_key(N(carol),   "active")}
+                               };
+   //wlog("set producer schedule to [alice,bob]");
```
heifner: remove

unittests/producer_schedule_tests.cpp
```diff
@@ -260,4 +260,61 @@ BOOST_FIXTURE_TEST_CASE( producer_schedule_promotion_test, TESTER ) try {
...
+   BOOST_CHECK_EQUAL( control->active_producers().version, 1 );
+   BOOST_CHECK_EQUAL( true, compare_schedules( sch1, control->active_producers() ) );
+   produce_blocks(7);
+
+   res = set_producers( {N(alice),N(bob)} );
+   vector<producer_key> sch2 = {
+                                 {N(alice), get_public_key(N(alice), "active")},
+                                 {N(bob),   get_public_key(N(bob),   "active")}
+                               };
+   wlog("set producer schedule to [alice,bob,carol]");
```
heifner: This should be [alice,bob], right?

Issue #5088, opened 2018-08-07
Remove "carriage return" updating when replaying


wanderingbort:
Docker does not handle the way we "dynamically" update our replay status well.  Furthermore, any logging solution that is _not_ an active user at a terminal gets less information from this mode of display than a regular interval full log message.  

We should remove this anti-feature or at least gate it by some semi-reliably `is_tty` style detection of a real console vs piped/redirected stdout/stderr.

PR #6445, opened 2018-12-06
Regularize replay logging

jgiszczak:
## Change Description

When stderr is redirected to a file, log replay progress messages normally rather than the terminal-friendly status message. #5088

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

Replay logging will appear different when redirected to a file.


libraries/chain/include/eosio/chain/fileno.hpp
```diff
@@ -0,0 +1,192 @@
+#pragma once
```
heifner: This should go in fc.
Please add a comment giving credit and a pointer to where the code came from. If you made any modifications to the source please highlight those with comments.

libraries/chain/controller.cpp
```diff
@@ -317,17 +318,24 @@ struct controller_impl {
             ("s", start_block_num)("n", blog_head->block_num()) );
 
       auto start = fc::time_point::now();
+      bool cerr_is_tty = isatty(fileno_hack(std::cerr));
```
spoonincode: I'm a little confused by the need for fileno_hack here. stderr is, by definition, always fd 2. Why can't this code just be `bool cerr_is_tty = isatty(STDERR_FILENO);` and call it a day.

jgiszczak: The constant stderr is indeed always fd 2, but the actual stream may no longer be fd 2 if it has been manipulated.  Specifically, in some versions of macOS, `fclose()` of stdin or stdout followed by an `freopen()` of stderr will change the file descriptor to the first available one, which could be 0 or 1.  This behavior is strictly POSIX compliant because of the wording of the description of `freopen()`, which reads "The original stream shall be closed regardless of whether the **subsequent** open succeeds."  [emphasis added]  Various standard libraries, including glibc, are intentionally noncompliant, but macOS strictly complied in at least 10.9.4 and possibly other versions.

spoonincode: okay, fair enough, but that's something our process would have to explicitly do. And we know it doesn't do that. So is it worth dipping our hand in to the private implementation of the c++ library for something we'll never do?

jgiszczak: Per opinion of attorneys, CC-BY-SA licensed code can not be used without changing the current license.

PR #6448, opened 2018-12-07
cleos print contract log only one line

firesWu:
## Change Description

when use `cleos push action` or  `get actions account --console`, console only print one line.

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A


programs/cleos/main.cpp
```diff
@@ -2354,8 +2355,9 @@ int main( int argc, char** argv ) {
                     stringstream out;
                     std::stringstream ss(console);
                     string line;
-                    std::getline( ss, line );
-                    out << ">> " << line << "\n";
+                    while(std::getline( ss, line )){
+                       out << ">> " << line << "\n";
+                    }
                     cerr << out.str(); //ilog( "\r${m}                                   ", ("m",out.str()) );
```
heifner: The printing of one line here and above was on purpose. If you want to add a command line option to enable full output that would be acceptable.

heifner:
Thanks. Replaced by #6806

PR #6470, opened 2018-12-12
Update the bios-boot-tutorial.py script and readme.md

iamveritas:
## Change Description

1. System contract has to be initialized before it can be used (e.g. creating staked acconts).
2. Commands list updated to include the new addition stepInitSystemContract.
3. When args.contracts_dir doesn't end in a forward slash, when it is concatenated with a subpath, the subpath has to start with a forward slash; just a safety measure and more forgiving for the end user.
4. README.md updated as well to reflect the current status of the script and requirements.

### DIFF
--- a/tutorials/bios-boot-tutorial/bios-boot-tutorial.py
+++ b/tutorials/bios-boot-tutorial/bios-boot-tutorial.py
@@ -284,17 +284,20 @@ def stepStartBoot():
     startNode(0, {'name': 'eosio', 'pvt': args.private_key, 'pub': args.public_key})
     sleep(1.5)
 def stepInstallSystemContracts():
-    run(args.cleos + 'set contract eosio.token ' + args.contracts_dir + 'eosio.token/')
-    run(args.cleos + 'set contract eosio.msig ' + args.contracts_dir + 'eosio.msig/')
+    run(args.cleos + 'set contract eosio.token ' + args.contracts_dir + '/eosio.token/')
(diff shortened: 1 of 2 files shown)


tutorials/bios-boot-tutorial/bios-boot-tutorial.py
```diff
@@ -326,23 +329,24 @@ def stepLog():
...
-    ('X', 'xfer',           stepTransfer,               False,   "Random transfer tokens (infinite loop)"),
-    ('l', 'log',            stepLog,                    True,    "Show tail of node's log"),
+    ('k', 'kill',               stepKillAll,                True,    "Kill all nodeos and keosd processes"),
+    ('w', 'wallet',             stepStartWallet,            True,    "Start keosd, create wallet, fill with keys"),
+    ('b', 'boot',               stepStartBoot,              True,    "Start boot node"),
+    ('s', 'sys',                createSystemAccounts,       True,    "Create system accounts (eosio.*)"),
+    ('c', 'contracts',          stepInstallSystemContracts, True,    "Install system contracts (token, msig)"),
+    ('t', 'tokens',             stepCreateTokens,           True,    "Create tokens"),
+    ('S', 'sys-contract',       stepSetSystemContract,      True,    "Set system contract"),
+    ('I', 'init-sys-contract',  stepInitSystemContract,     True,    "Initialiaze system contract"),
```
iamveritas: this is the only actual change in this big chunk; the rest is re-alignment.

PR #6471, opened 2018-12-12
Optimize transaction signature recovery

heifner:
## Change Description

- Multi-thread transaction signature recovery for incoming transactions

## Consensus Changes

None

## API Changes

- New `producer_plugin` option:
  --   `--producer-threads arg (=2)           Number of worker threads in producer thread pool`

- New `chain_plugin` option:
  --   `--signature-cpu-billable-pct arg (=50)
                                        Percentage of actual signature recovery
                                        cpu to bill. Whole number percentages, 
                                        e.g. 50 for 50%`

## Documentation Additions

See API Changes above

### DIFF
--- a/libraries/chain/include/eosio/chain/abi_serializer.hpp
+++ b/libraries/chain/include/eosio/chain/abi_serializer.hpp
@@ -418,11 +418,11 @@ namespace impl {
          mutable_variant_object mvo;
          auto trx = ptrx.get_transaction();
          mvo("id", trx.id());
-         mvo("signatures", ptrx.signatures);
-         mvo("compression", ptrx.compression);
-         mvo("packed_context_free_data", ptrx.packed_context_free_data);
+         mvo("signatures", ptrx.get_signatures());
+         mvo("compression", ptrx.get_compression());
+         mvo("packed_context_free_data", ptrx.get_packed_context_free_data());
          mvo("context_free_data", ptrx.get_context_free_data());
-         mvo("packed_trx", ptrx.packed_trx);
+         mvo("packed_trx", ptrx.get_packed_transaction());
          add(mvo, "transaction", trx, resolver, ctx);
 
          out(name, std::move(mvo));
@@ -577,32 +577,33 @@ namespace impl {
          const variant_object& vo = v.get_object();
          EOS_ASSERT(vo.contains("signatures"), packed_transaction_type_exception, "Missing signatures");
          EOS_ASSERT(vo.contains("compression"), packed_transaction_type_exception, "Missing compression");
-         from_variant(vo["signatures"], ptrx.signatures);
-         from_variant(vo["compression"], ptrx.compression);
+         std::vector<signature_type> signatures;
+         packed_transaction::compression_type compression;
+         from_variant(vo["signatures"], signatures);
+         from_variant(vo["compression"], compression);
 
-         // TODO: Make this nicer eventually. But for now, if it works... good enough.
+         bytes packed_cfd;
          if( vo.contains("packed_trx") && vo["packed_trx"].is_string() && !vo["packed_trx"].as_string().empty() ) {
-            from_variant(vo["packed_trx"], ptrx.packed_trx);
-            auto trx = ptrx.get_transaction(); // Validates transaction data provided.
+            bytes packed_trx;
+            std::vector<bytes> cfd;
+            from_variant(vo["packed_trx"], packed_trx);
             if( vo.contains("packed_context_free_data") && vo["packed_context_free_data"].is_string() && !vo["packed_context_free_data"].as_string().empty() ) {
-               from_variant(vo["packed_context_free_data"], ptrx.packed_context_free_data );
+               from_variant(vo["packed_context_free_data"], packed_cfd );
             } else if( vo.contains("context_free_data") ) {
-               vector<bytes> context_free_data;
-               from_variant(vo["context_free_data"], context_free_data);
-               ptrx.set_transaction(trx, context_free_data, ptrx.compression);
+               from_variant(vo["context_free_data"], cfd);
             }
+            ptrx = packed_transaction( std::move(packed_trx), std::move(signatures), std::move(packed_cfd), std::move(cfd), compression );
          } else {
             EOS_ASSERT(vo.contains("transaction"), packed_transaction_type_exception, "Missing transaction");
-            transaction trx;
-            vector<bytes> context_free_data;
+            signed_transaction trx;
+            trx.signatures = std::move(signatures);
             extract(vo["transaction"], trx, resolver, ctx);
             if( vo.contains("packed_context_free_data") && vo["packed_context_free_data"].is_string() && !vo["packed_context_free_data"].as_string().empty() ) {
-               from_variant(vo["packed_context_free_data"], ptrx.packed_context_free_data );
-               context_free_data = ptrx.get_context_free_data();
+               from_variant(vo["packed_context_free_data"], packed_cfd );
             } else if( vo.contains("context_free_data") ) {
-               from_variant(vo["context_free_data"], context_free_data);
+               from_variant(vo["context_free_data"], trx.context_free_data );
             }
-            ptrx.set_transaction(trx, context_free_data, ptrx.compression);
+            ptrx = packed_transaction( std::move(trx), std::move(packed_cfd), compression );
          }
       }
    };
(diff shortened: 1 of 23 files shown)


libraries/chain/transaction.cpp
```diff
@@ -343,10 +344,35 @@ signed_transaction packed_transaction::get_signed_transaction() const
 
 }
 
-void packed_transaction::set_transaction(const transaction& t, packed_transaction::compression_type _compression)
+packed_transaction::packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs,
```
arhag: I think it would be clearer if this constructor was split into two:

- `packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs,  bytes&& packed_cfd, compression_type _compression )`
- `packed_transaction( bytes&& packed_txn, vector<signature_type>&& sigs,  vector<bytes>&& cfd, compression_type _compression )`

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -577,32 +577,33 @@ namespace impl {
...
          EOS_ASSERT(vo.contains("compression"), packed_transaction_type_exception, "Missing compression");
-         from_variant(vo["signatures"], ptrx.signatures);
-         from_variant(vo["compression"], ptrx.compression);
+         std::vector<signature_type> signatures;
+         packed_transaction::compression_type compression;
+         from_variant(vo["signatures"], signatures);
+         from_variant(vo["compression"], compression);
 
-         // TODO: Make this nicer eventually. But for now, if it works... good enough.
+         bytes packed_cfd;
```
arhag: The `std::vector<bytes> cfd;` could also be moved up here, and a `bool use_packed_cfd = true;` also added here.

The code that checks for `packed_context_free_data` or `context_free_data` fields and extracts the appropriate one from the variant is repeated twice below. It could also be pulled out and just included once here. If there is no valid `packed_context_free_data` to extract, but there is a `context_free_data` field, it would set the `use_packed_cfd` to `false`.

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -577,32 +577,33 @@ namespace impl {
...
             if( vo.contains("packed_context_free_data") && vo["packed_context_free_data"].is_string() && !vo["packed_context_free_data"].as_string().empty() ) {
-               from_variant(vo["packed_context_free_data"], ptrx.packed_context_free_data );
+               from_variant(vo["packed_context_free_data"], packed_cfd );
             } else if( vo.contains("context_free_data") ) {
-               vector<bytes> context_free_data;
-               from_variant(vo["context_free_data"], context_free_data);
-               ptrx.set_transaction(trx, context_free_data, ptrx.compression);
+               from_variant(vo["context_free_data"], cfd);
             }
+            ptrx = packed_transaction( std::move(packed_trx), std::move(signatures), std::move(packed_cfd), std::move(cfd), compression );
```
arhag: This line could then be replaced with:

```
if( use_packed_cfd ) {
   ptrx = packed_transaction( std::move(packed_trx), std::move(signatures), std::move(packed_cfd), compression );
} else {
   ptrx = packed_transaction( std::move(packed_trx), std::move(signatures), std::move(cfd), compression );
}
```

(Assumes splitting of `packed_transaction` constructor mentioned elsewhere.)

libraries/chain/transaction.cpp
```diff
@@ -343,10 +344,35 @@ signed_transaction packed_transaction::get_signed_transaction() const
...
+,packed_context_free_data(std::move(packed_cfd))
+,packed_trx(std::move(packed_txn))
+{
+   EOS_ASSERT(packed_cfd.empty() || cfd.empty(), tx_decompression_error, "Invalid packed_transaction");
+   if( !cfd.empty() ) {
+      set_context_free_data(cfd);
+   }
+}
+
+packed_transaction::packed_transaction( signed_transaction&& t, bytes&& packed_cfd, compression_type _compression )
```
arhag: I think it would be clearer if this constructor was split into two:

- `packed_transaction( signed_transaction&& t, compression_type _compression )`
- `packed_transaction( transaction&& t, vector<signature_type>&& sigs, bytes&& packed_cfd, compression_type _compression )`

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -577,32 +577,33 @@ namespace impl {
...
          } else {
             EOS_ASSERT(vo.contains("transaction"), packed_transaction_type_exception, "Missing transaction");
-            transaction trx;
-            vector<bytes> context_free_data;
+            signed_transaction trx;
```
arhag: This whole else clause could then be replaced with:
```
} else {
   EOS_ASSERT(vo.contains("transaction"), packed_transaction_type_exception, "Missing transaction");
   if( use_packed_cfd ) {
      transaction trx;
      extract(vo["transaction"], trx, resolver, ctx);
      ptrx = packed_transaction( std::move(trx), std::move(signatures), std::move(packed_cfd), compression );
   } else {
      signed_transaction trx;
      trx.signatures = std::move(signatures);
      trx.context_free_data = std::move(cfd);
      extract(vo["transaction"], trx, resolver, ctx);
      ptrx = packed_transaction( std::move(trx), compression );
   }
}
```

(Assumes splitting of `packed_transaction` constructor mentioned elsewhere.)

heifner:
@taokayan Thanks for testing on your machine.

I was keeping it simple with one thread_pool since it is configurable and a user could just double the thread_pool size if this is an issue. But I can see adding another thread_pool to producer_plugin for these waits. I could also see using different thread_pools for apply_block sig recovery so that incoming transactions can't hold up apply_block.

libraries/chain/transaction_metadata.cpp
```diff
@@ -0,0 +1,40 @@
...
+         }
+      }
+      signing_keys = std::make_pair( chain_id, trx.get_signature_keys( chain_id ));
+   }
+   return signing_keys->second;
+}
+
+void transaction_metadata::create_signing_keys_future( const transaction_metadata_ptr& mtrx,
+                                                       boost::asio::thread_pool& thread_pool, const chain_id_type& chain_id ) {
+   if( mtrx->signing_keys && mtrx->signing_keys->first == chain_id )
```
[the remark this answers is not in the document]
heifner: Yes. Removed.

libraries/chain/transaction.cpp
heifner: Fixed in #6494

libraries/chain/include/eosio/chain/abi_serializer.hpp
heifner: #6494

libraries/chain/include/eosio/chain/abi_serializer.hpp
heifner: Fixed in #6494

libraries/chain/transaction.cpp
heifner: Fixed in #6494

libraries/chain/include/eosio/chain/abi_serializer.hpp
heifner: Fixed in #6494

libraries/chain/transaction.cpp
```diff
@@ -82,28 +82,28 @@ digest_type transaction::sig_digest( const chain_id_type& chain_id, const vector
...
 
    constexpr size_t recovery_cache_size = 1000;
    static thread_local recovery_cache_type recovery_cache;
+   fc::time_point start = fc::time_point::now();
    const digest_type digest = sig_digest(chain_id, cfd);
 
    flat_set<public_key_type> recovered_pub_keys;
    for(const signature_type& sig : signatures) {
+      auto now = fc::time_point::now();
+      EOS_ASSERT( start + now <= deadline, tx_cpu_usage_exceeded, "transaction signature verification executed for too long",
```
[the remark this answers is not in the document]
heifner: Thanks.

libraries/chain/transaction.cpp
```diff
@@ -88,14 +88,12 @@ flat_set<public_key_type> transaction::get_signature_keys( const vector<signatur
...
    static thread_local recovery_cache_type recovery_cache;
-   fc::time_point start = fc::time_point::now();
    const digest_type digest = sig_digest(chain_id, cfd);
 
    flat_set<public_key_type> recovered_pub_keys;
    for(const signature_type& sig : signatures) {
-      auto now = fc::time_point::now();
-      EOS_ASSERT( start + now <= deadline, tx_cpu_usage_exceeded, "transaction signature verification executed for too long",
-                  ("now", now)("deadline", deadline)("start", start) );
+      EOS_ASSERT( fc::time_point::now() < deadline, tx_cpu_usage_exceeded, "transaction signature verification executed for too long",
```
arhag: Better to use `auto now = fc::time_point::now();` so that the exception message reflects the actual condition that was checked in the assertion message.

PR #6490, opened 2018-12-17
Return 400 on get_block with nonexist block number [Issue 6374]

### LINKED ISSUE #6374: get_block api call returns incorrect http code for unknown block
Calling get_block for an unknown block - typically one with a block number greater than the head block number.

`curl --request POST --url https://node2.eosphere.io/v1/chain/get_block --data '{"block_num_or_id":"100000000"}'`

Returns http result code 500, the correct http result code to return in this situation is 404.

500 indicates an Internal Server Error, 404 "indicates that the origin server did not find a current representation for the target resource" 

404 seems to more accurately reflect the reality of asking for a block that does not exist.

necokeine:
Return 404 on get_block with uonexist block number.

Resolves #6374

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -561,6 +561,9 @@ namespace eosio {
       try {
          try {
             throw;
+         } catch (chain::unknown_block_exception& e) {
+            error_results results{400, "Unknown Block", error_results::error_info(e, verbose_http_errors)};
+            cb( 400, fc::json::to_string( results ));
          } catch (chain::unsatisfied_authorization& e) {
             error_results results{401, "UnAuthorized", error_results::error_info(e, verbose_http_errors)};
             cb( 401, fc::json::to_string( results ));


plugins/http_plugin/http_plugin.cpp
```diff
@@ -561,6 +561,9 @@ namespace eosio {
       try {
          try {
             throw;
+         } catch (chain::unknown_block_exception& e) {
+            error_results results{404, "Unknown Block", error_results::error_info(e, verbose_http_errors)};
+            cb( 404, fc::json::to_string( results ));
```
jgiszczak: Should be code 400.  See discussion on #6374.

necokeine:
Thank you for review this code, it is my first time to send a merge request on Github. I have several questions here.
1. Shall I develop on master thread directly? or something else.
2. Shall I follow some code style here?

heifner:
@necokeine Please develop on `develop` unless it is a bug fix for a specific version.
Please follow the code style you find in the files you are modifying.
Thanks for your contributions.

PR #6501, opened 2018-12-18
optimization when writing shared_blob data

wanderingbort:
## Change Description

@taokayan analyzed the cost associated with the `resize` + `memcpy` approach to updating these buffers and discovered that `resize` was very slow.  In addition, the default copy constructor for `shared_blob` had the same issues.

using `assign` directly in both cases was a clear win. 

I've ported these changes from another branch @taokayan shared with so that they can be adopted for the 1.6.0-rc1 pre-release.

## Consensus Changes

none

## API Changes

none

### DIFF
--- a/libraries/chain/include/eosio/chain/types.hpp
+++ b/libraries/chain/include/eosio/chain/types.hpp
@@ -112,10 +112,13 @@ namespace eosio { namespace chain {
             assign(s.c_str(), s.size());
          }
 
-         shared_blob &operator=(const shared_blob &s) {
+
+         shared_blob& operator=(const shared_blob& s) {
             assign(s.c_str(), s.size());
             return *this;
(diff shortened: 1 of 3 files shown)


libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -103,7 +103,19 @@ namespace eosio { namespace chain {
...
+         shared_blob(const shared_blob& s)
+         :shared_string(s.get_allocator())
+         {
+            assign(s.c_str(), s.size());
+         }
+
+         shared_blob &operator=(const shared_blob &s) {
+            assign(s.c_str(), s.size());
+            return *this;
+         }
```
heifner: Seems we should have a move operator=

PR #6557, opened 2019-01-04
Remove binaryen & other vesigigial pieces

spoonincode:
## Change Description

After the change to (optionally) use cdt this stuff can be removed.

Binaryen finally gone 🎉 🎉


.gitmodules
```diff
@@ -14,16 +14,9 @@
 	path = contracts/libc++/upstream
 	url = https://github.com/EOSIO/libcxx.git
 	branch = eosio
-[submodule "externals/binaryen"]
```
heifner: `contracts/musl/upstream` and `contracts/libc++/upstream` above can go as well

spoonincode: that's interesting `git submodule` didn't show those

heifner: Maybe because parent directory does not exist, but who knows.

PR #6569, opened 2019-01-08
Remove LLVM4 wasm & clang from build scripts

spoonincode:
## Change Description
amazon, centos, and fedora still need a manually built LLVM4 for wavm. However, they no longer need clang & wasm support. Don't compile clang & wasm -- hopefully speeding up this step for users

### DIFF
--- a/scripts/eosio_build_amazon.sh
+++ b/scripts/eosio_build_amazon.sh
@@ -534,9 +534,9 @@ fi
 		printf "\\tMongo C++ driver found at /usr/local/lib64/libmongocxx-static.a.\\n"
 	fi
 
-	printf "\\n\\tChecking LLVM with WASM support.\\n"
+	printf "\\n\\tChecking LLVM4.\\n"
 	if [ ! -d "${HOME}/opt/wasm/bin" ]; then
-		printf "\\tInstalling LLVM & WASM.\\n"
+		printf "\\tInstalling LLVM4.\\n"
 		if ! cd "${TEMP_DIR}"
 		then
 			printf "\\n\\tUnable to cd into directory %s.\\n" "${TEMP_DIR}"
@@ -567,12 +567,6 @@ fi
 			printf "\\n\\tExiting now.\\n"
 			exit 1;
 		fi
-		if ! git clone --depth 1 --single-branch --branch release_40 https://github.com/llvm-mirror/clang.git
-		then
-			printf "\\tUnable to clone clang repo @ https://github.com/llvm-mirror/clang.git.\\n"
-			printf "\\tExiting now.\\n\\n"
-			exit 1;
-		fi
 		if ! cd "${TEMP_DIR}/llvm-compiler/llvm"
 		then
 			printf "\\n\\tUnable to change directory into %s/llvm-compiler/llvm.\\n" "${TEMP_DIR}"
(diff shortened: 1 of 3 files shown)


scripts/eosio_build_amazon.sh
```diff
@@ -592,22 +586,22 @@ fi
 			exit 1;
 		fi
 		if ! "$CMAKE" -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX="${HOME}/opt/wasm" \
```
jgiszczak: Did you consider changing the install prefix to `${HOME}/opt/llvm4`?

spoonincode: yeah but that means either 1) the script has to rebuild llvm4 when the script doesn't find ~/opt/llvm4 or 2) the script needs to move opt/wasm to opt/llvm4. And then would have to worry about what was in cmake's variable cache that may need updating. Just seemed like a lot of potential issues

jgiszczak: With the change in build defines, isn't a rebuild going to be triggered anyway?  Quite aside from the confusion of an llvm4 named 'wasm' that has no WASM build features.  And also there's now clang cruft in any previous build.  The uninstall script already takes a fairly heavy handed approach to removing anything that was previously installed with a script.  This seems like more of the same.

spoonincode: >With the change in build defines, isn't a rebuild going to be triggered anyway?

A rebuild of LLVM is what I'm getting at. That's expensive (from a time standpoint)

> The uninstall script already takes a fairly heavy handed approach to removing anything that was previously installed with a script. 

AFAICT the uninstall script only uninstalls stuff built from cmake. Once the eosio_build.sh script drops llvm4, mongo c libs, boost, etc all over your box there is no script to undo that.

I left like we had 3 options:
1) Continue to use ~/opt/wasm for both current and new users. New users will get an llvm4 built in there, current users will continue to use the llvm4+clang4 cruft. (current approach) The name of course is a misnomer (especially for new users).
2) Use ~/opt/llvm4 where by both new and current users need to build llvm4. For current users this is lousy because an llvm4 build can take 15, 30, or more minutes.
3) Use ~/opt/llvm4 but for current users try to migrate their ~/opt/wasm here. That still leaves clang4 cruft in ~/opt/llvm4 but at least the name is more representative of what we're actually using that directory for. My biggest fear of this approach is dealing with cmake cache issues.

However there is a potential compromise -- current users continue to use ~/opt/wasm (so that we don't need to have them spend 15+ minutes rebuilding llvm4). New users will get a build of llvm4 in ~/opt/llvm4 (so that the name is accurate). I _think_ we should be able to set up `LLVM_DIR=$HOME/opt/llvm4:$HOME/opt/wasm` to search in both locations.

NorseGaud: New build scripts I've written (to support the new ci images) symlink LLVM, regardless of OS, into ~/opt/llvm. Then, each OS build script prints instructions to modify PATH and LD_LIBRARY_PATH to ~/opt/llvm:
`PATH=\$HOME/bin:\$PATH:$MONGODB_LINK_LOCATION/bin:\$HOME/opt/llvm/bin`
`LD_LIBRARY_PATH=\$HOME/opt/llvm/lib:\$LD_LIBRARY_PATH`

Due to the recent change of LLVM arguments, I think a rebuild of LLVM, except for the brew install on darwin and the llvm-4.0 install in ubuntu, shouldn't be too painful for the users.

NorseGaud: Also, everything else is installed into ~/bin ~/lib, etc. I'm trying to move closer to a rootless install.

spoonincode:
Holding this PR up for now as it may be a moot point with NorseGaud's changes coming

scripts/eosio_build_amazon.sh
jgiszczak: @NorseGaud 
> Also, I've got a new cleanup script that does go across the machine to clean up older installed libraries out of /usr/local. I was thinking about also scripting something that would read an installation manifest file that's created on install and just delete all of the files it gets out of the file.

CMake has a philosophical disagreement with this feature and refuses to implement it in CMake-generated Makefiles, or we would have used the feature ourselves.  Their argument is that arbitrary scripting can result in installed files anywhere at any time, untracked by automatically generated manifests, and so they shouldn't try to generate any manifest at all.  Personally I think that's ridiculous, and the API should simply include the ability to manually add to the manifest, but CMake's authors don't listen to me.

@spoonincode 
> A rebuild of LLVM is what I'm getting at. That's expensive (from a time standpoint)

So was I.  Since the LLVM build settings themselves have changed, even previous users will be rebuilding LLVM.  The `-DLLVM_TARGETS_TO_BUILD="host" -DLLVM_BUILD_TOOLD=false` changes should trigger a rebuild.  It may be relatively cheap, but it may not.  I don't know how deeply those defines are used in the code.  If they're present in one major header, a rebuild could get expensive.

I'm almost completely unconcerned about rebuild time because almost all of the people who are doing their own builds are accustomed to a long fire-and-forget build process anyway.  For those few who sit and wait it out the optimization is helpful, but I question how many people do that.  I do like the idea to make LLVM_DIR multiple directories.

> AFAICT the uninstall script only uninstalls stuff built from cmake.

I was referring more to `scripts/clean_old_install.sh` which is invoked by the installer script.

PR #6572, opened 2019-01-08
Add get_supported_apis RPC API

conr2d:
## Change Description

Each API endpoint doesn't support every APIs because of different sets of activated plugins. Generally, wallet applications can change API endpoint, but they are not sure what features can be provided with that endpoint. (Eg. without history_api_plugin, wallets cannot provide transaction list) By calling this API, they can deactivate UI related the features which cannot be provided with given endpoint.

## Consensus Changes

None

## API Changes

`/v1/node/get_supported_apis` RPC API is added.

## Documentation Additions

Need to add RPC API docmentation

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -632,7 +632,8 @@ namespace eosio {
       get_supported_apis_result result;
 
       for (const auto& handler : my->url_handlers) {
-         result.apis.emplace_back(handler.first);
+         if (handler.first != "/v1/node/get_supported_apis")
+            result.apis.emplace_back(handler.first);
       }
 
       return result;
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -542,21 +542,18 @@ namespace eosio {
          }
       }
 
-#define CALL(api_name, call_name, http_response_code) \
-{std::string("/v1/" #api_name "/" #call_name), \
-   [&](string, string body, url_response_callback cb) mutable { \
-          try { \
-             if (body.empty()) body = "{}"; \
-             auto result = (*this).call_name(); \
-             cb(http_response_code, fc::json::to_string(result)); \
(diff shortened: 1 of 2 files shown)


conr2d:
I failed to build with `develop` branch, so wrote this on `master` branch. Let me know if this PR needs to be rebased on `develop`.

I have a plan add one more API like `/v1/node/get_activated_plugins`, but this requires changes on `appbase`, because appbase doesn't expose a method returning a list of plugins or activated plugins. Feel free to give me suggestions about it.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -541,6 +541,22 @@ namespace eosio {
             throw;
          }
       }
+
+#define CALL(api_name, call_name, http_response_code) \
```
wanderingbort: As we only have 1 endpoint to add, and a `#define` in a poor place that would imply some amount of scoping to the method body (which doesn't exist), can we just eliminate the `#define` by interpolating it out ?

plugins/http_plugin/http_plugin.cpp
```diff
@@ -615,4 +631,13 @@ namespace eosio {
       return verbose_http_errors;
    }
 
+   http_plugin::get_supported_apis_result http_plugin::get_supported_apis()const {
+      get_supported_apis_result result;
```
wanderingbort: Is it recursive that the `get_supported_apis` endpoint will list itself?

plugins/http_plugin/http_plugin.cpp
conr2d: I added this macro in preparation for `/v1/node/get_activated_plugins` in following patch, but it would be all right to add it in that.

plugins/http_plugin/http_plugin.cpp
conr2d: I also considered this, but it seems better to exclude `get_supported_apis` from the result on second thoughts.

PR #6577, opened 2019-01-09
Priority queue

heifner:
## Change Description

- Add priority queue for main `application` thread
- Use low priority for transaction processing
- Use high priority for block production and block propagation

All posted jobs now require a priority that determines the order of execution. If any jobs are posted to the main application thread without using the provided priority wrapper they are executed as if they have the highest priority. Therefore, the priority wrapper should be used for all jobs posted either via `post`, `async_read`, `async_write`, `async_wait`, etc.

## Consensus Changes

None

## API Changes

Any plugins posting to the main `application` thread should consider using the new priority wrapper.

## Documentation Additions

See API changes.

### DIFF
--- a/plugins/bnet_plugin/bnet_plugin.cpp
+++ b/plugins/bnet_plugin/bnet_plugin.cpp
@@ -1251,10 +1251,10 @@ namespace eosio {
             /// add some random delay so that all my peers don't attempt to reconnect to me
             /// at the same time after shutting down..
             _timer->expires_from_now( boost::posix_time::microseconds( 1000000*(10+rand()%5) ) );
-            _timer->async_wait([=](const boost::system::error_code& ec) {
+            _timer->async_wait(app().get_priority_queue().wrap(priority::low, [=](const boost::system::error_code& ec) {
                 if( ec ) { return; }
                 on_reconnect_peers();
-            });
+            }));
          }
    };
 
--- a/plugins/bnet_plugin/bnet_plugin.cpp
+++ b/plugins/bnet_plugin/bnet_plugin.cpp
@@ -917,7 +917,7 @@ namespace eosio {
          */
         void wait_on_app() {
            app().post( priority::low, [self = shared_from_this()]() {
-              boost::asio::bind_executor( self->_strand, [self] { self->do_read(); } );
+              app().get_io_service().post( boost::asio::bind_executor( self->_strand, [self] { self->do_read(); } ) );
            } );
         }
 
--- a/plugins/bnet_plugin/bnet_plugin.cpp
+++ b/plugins/bnet_plugin/bnet_plugin.cpp
@@ -1252,7 +1252,7 @@ namespace eosio {
             _timer->async_wait(app().get_priority_queue().wrap(priority::low, [=](const boost::system::error_code& ec) {
                 if( ec ) { return; }
                 on_reconnect_peers();
-            }, __FILE__, __LINE__, __func__));
+            }));
          }
    };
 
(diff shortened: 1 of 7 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1363,12 +1366,13 @@ void producer_plugin_impl::schedule_production_loop() {
       _timer.expires_from_now( boost::posix_time::microseconds( config::block_interval_us  / 10 ));
 
       // we failed to start a block, so try again later?
-      _timer.async_wait([weak_this,cid=++_timer_corelation_id](const boost::system::error_code& ec) {
-         auto self = weak_this.lock();
-         if (self && ec != boost::asio::error::operation_aborted && cid == self->_timer_corelation_id) {
-            self->schedule_production_loop();
-         }
-      });
+      _timer.async_wait( app().get_priority_queue().wrap( priority::medium,
```
brianjohnson5972: Should this be high?  If not add comment why medium.

plugins/bnet_plugin/bnet_plugin.cpp
```diff
@@ -919,9 +916,9 @@ namespace eosio {
          * the connection from being closed.
          */
         void wait_on_app() {
-            app().get_io_service().post( 
-                boost::asio::bind_executor( _strand, [self=shared_from_this()]{ self->do_read(); } )
-            );
+           app().post( priority::low, [self = shared_from_this()]() {
```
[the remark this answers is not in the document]
heifner: changed

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1978,7 +1997,7 @@ namespace eosio {
                            }
                         }
                      }
-                     start_read_message(conn);
+                     app().post(priority::medium, [this, conn]() { start_read_message(conn); });
```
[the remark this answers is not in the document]
heifner: removed

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2027,7 +2046,7 @@ namespace eosio {
    void net_plugin_impl::send_all(const std::shared_ptr<std::vector<char>>& send_buffer, VerifierFunc verify) {
       for( auto &c : connections) {
          if( c->current() && verify( c )) {
-            c->enqueue_buffer( send_buffer, true, no_reason );
+            c->enqueue_buffer( send_buffer, true, priority::low, no_reason );
```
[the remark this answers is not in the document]
heifner: renamed

plugins/producer_plugin/producer_plugin.cpp
heifner: changed to high

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2425,33 +2443,33 @@ namespace eosio {
...
                start_conn_timer( connector_period, std::weak_ptr<connection>());
             }
-         });
+         }));
    }
 
    void net_plugin_impl::start_txn_timer() {
       transaction_check->expires_from_now( txn_exp_period);
-      transaction_check->async_wait( [this](boost::system::error_code ec) {
+      transaction_check->async_wait(app().get_priority_queue().wrap(priority::low-1, [this](boost::system::error_code ec) {
```
wanderingbort: I don't really like the mixed metaphor of priority is a category and priority is a number.   I'm fine with having named points on the range, but then perhaps we can name this one as well (even if only locally).   

I guess, in a single logical bit of code it should consistently treat priority as a category or a number but not both?

I have some comments on the appbase changes too that regard this duality

heifner: Used a variable to name it locally. appbase was updated to just always execute the highest so there are no more ranges only named points on the range.

PR #6582, opened 2019-01-11
Update txn_test_gen_plugin to print the actual TPS

baegjae:
## Change Description

When you stop the test, "us / transaction" shows the average value of `cpu_usage_us` (= billed CPU usage) of transactions.
Based on this value, you can not estimate how many transactions per second (TPS) an actual node generates.

We recorded the time at which the test was performed and print the actual TPS.

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

### DIFF
--- a/plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
+++ b/plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
@@ -91,6 +91,7 @@ struct txn_test_gen_plugin_impl {
 
    uint64_t _total_us = 0;
    uint64_t _txcount = 0;
+   time_point _start_time;
 
    int _remain = 0;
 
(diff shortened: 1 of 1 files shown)


plugins/txn_test_gen_plugin/txn_test_gen_plugin.cpp
```diff
@@ -368,10 +370,12 @@ struct txn_test_gen_plugin_impl {
          throw fc::exception(fc::invalid_operation_exception_code);
       timer.cancel();
       running = false;
+      auto elapsed_ms = (fc::time_point::now() - _start_time).count()/1000;
       ilog("Stopping transaction generation test");
 
       if (_txcount) {
-         ilog("${d} transactions executed, ${t}us / transaction", ("d", _txcount)("t", _total_us / (double)_txcount));
+         ilog("${d} transactions executed for ${s} ms (${tps} TPS), ${t} us / transaction",
+            ("d", _txcount) ("s", elapsed_ms) ("tps", (uint16_t)(((double)_txcount/elapsed_ms)*1000)) ("t", _total_us / (double)_txcount));
```
heifner: "tps" here is number transactions submitted per second not executed. This is misleading and not what is typically reported as tps.

PR #6583, opened 2019-01-11
Add support for account creation with permission level to cleos

conr2d:
## Change Description

This PR allows user to create account with permission level instead of public key. For example,
```
cleos create account eosio eosio.token eosio@active eosio@active
```

## Consensus Changes

 None

## API Changes

None

## Documentation Additions

Need to add description for cleos command `cleos create account` and `cleos system newaccount`


programs/cleos/main.cpp
```diff
@@ -512,14 +514,20 @@ void send_transaction( signed_transaction& trx, int32_t extra_kcpu, packed_trans
...
+
+chain::action create_newaccount(const name& creator, const name& newaccount, auth_type owner, auth_type active) {
    return action {
       get_account_permissions(tx_permission, {creator,config::active_name}),
       eosio::chain::newaccount{
          .creator      = creator,
          .name         = newaccount,
-         .owner        = eosio::chain::authority{1, {{owner, 1}}, {}},
-         .active       = eosio::chain::authority{1, {{active, 1}}, {}}
+         .owner        = !owner.which() ? authority(owner.get<public_key_type>()) : authority(owner.get<permission_level>()),
```
wanderingbort: please use `owner.contains<public_key_type>()` so that its explicit and durable to reordering instead of implicitly depending on the falsey nature of the `0` value returned from `which()`

programs/cleos/main.cpp
```diff
@@ -1010,14 +1018,23 @@ struct create_account_subcommand {
...
-            public_key_type owner_key, active_key;
+            auth_type owner, active;
             try {
-               owner_key = public_key_type(owner_key_str);
-            } EOS_RETHROW_EXCEPTIONS(public_key_type_exception, "Invalid owner public key: ${public_key}", ("public_key", owner_key_str));
+               owner = public_key_type(owner_key_str);
+            } catch (...) {
+               try {
+                  owner = get_permission_level(owner_key_str);
+               } EOS_RETHROW_EXCEPTIONS(public_key_type_exception, "Invalid owner public key: ${public_key}", ("public_key", owner_key_str));
```
wanderingbort: I think we need to clean this error message up to include permissions, also it seems like `get_permission_level` will throw `name_type_exception` if the names are bad or generically assert if there is no separator.  Both of which will not get caught as a `public_key_type_exception`

programs/cleos/main.cpp
wanderingbort: please use `owner.contains<public_key_type>()` so that its explicit and durable to reordering instead of implicitly depending on the falsey nature of the `0` value returned from `which()`

PR #6588, opened 2019-01-11
Refactor block_header_state

arhag:
This is a refactor of how `block_header_state`s are updated as new blocks come in. It conceptually simplifies the process by introducing a `pending_block_header_state` for the new block which is then finalized to the `block_header_state` when the necessary information to do so becomes available. This two stage process allows the class instances to remain immutable after instantiation and thereby get type safety benefits and lower developer cognitive overhead with the previous approach of mutating the `block_header_state` as it moves along through this process.

This refactor helps prepare for the later changes that will be made to `block_header_state` to introduce the field tracking the currently activated protocol features (which is not yet included in this change).

I am aware that this change will break both snapshots and the serialization format of `forkdb.dat`. This is the first part of a larger sequence of other refactors and further changes towards building the foundations that enable consensus protocol upgrade features. With those later changes, the `forkdb.dat` format will officially change and the new format will be stored with proper versioning in a file `fork_db.dat`. Furthermore, the later changes would introduce a second version of the chain snapshot and nodeos will not support reading from the first version of chain snapshots. These eventual changes will force a replay from genesis, which is desired behavior to support one of the protocol features planned.

### DIFF
--- a/libraries/chain/include/eosio/chain/block_header_state.hpp
+++ b/libraries/chain/include/eosio/chain/block_header_state.hpp
@@ -5,6 +5,42 @@
 
 namespace eosio { namespace chain {
 
+struct block_header_state;
+
+struct pending_block_header_state {
+   uint32_t                          block_num = 0;
+   block_id_type                     previous;
+   block_timestamp_type              timestamp;
+   account_name                      producer;
+   uint16_t                          confirmed = 1;
+   uint32_t                          dpos_proposed_irreversible_blocknum = 0;
+   uint32_t                          dpos_irreversible_blocknum = 0;
+   uint32_t                          active_schedule_version = 0;
+   uint32_t                          prev_pending_schedule_lib_num = 0; /// last irr block num
+   digest_type                       prev_pending_schedule_hash;
+   producer_schedule_type            prev_pending_schedule;
+   producer_schedule_type            active_schedule;
+   incremental_merkle                blockroot_merkle;
+   flat_map<account_name,uint32_t>   producer_to_last_produced;
+   flat_map<account_name,uint32_t>   producer_to_last_implied_irb;
+   public_key_type                   block_signing_key;
+   vector<uint8_t>                   confirm_count;
+   bool                              was_pending_promoted = false;
+
+   signed_block_header make_block_header( const checksum256_type& transaction_mroot,
+                                          const checksum256_type& action_mroot,
+                                          optional<producer_schedule_type>&& new_producers )const;
+
+   block_header_state  finish_next( const signed_block_header& h, bool skip_validate_signee = false )&&;
+
+   block_header_state  finish_next( signed_block_header& h,
+                                   const std::function<signature_type(const digest_type&)>& signer )&&;
+
+private:
+   block_header_state  _finish_next( const signed_block_header& h )&&;
+};
+
+
 /**
  *  @struct block_header_state
  *  @brief defines the minimum state necessary to validate transaction headers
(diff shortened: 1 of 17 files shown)


libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -27,17 +63,18 @@ struct block_header_state {
...
+   block_header_state   next( const signed_block_header& h, bool skip_validate_signee = false )const;
 
-    void set_new_producers( producer_schedule_type next_pending );
-    void set_confirmed( uint16_t num_prev_blocks );
-    void add_confirmation( const header_confirmation& c );
-    bool maybe_promote_pending();
+    //void set_new_producers( producer_schedule_type next_pending );
+    //void set_confirmed( uint16_t num_prev_blocks );
+    //void add_confirmation( const header_confirmation& c );
+    //bool maybe_promote_pending();
```
heifner: Remove commented out code. There is also some in the source file that could also be removed.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -493,15 +493,16 @@ namespace eosio {
...
             auto blk = chain.fetch_block_by_number( result.block_num );
-            if( blk == nullptr ) { // still in pending
-                auto blk_state = chain.pending_block_state();
-                if( blk_state != nullptr ) {
-                    blk = blk_state->block;
-                }
+            if( blk ) {
+               receipts = &blk->transactions;
+            } else if( chain.is_building_block() ) { // still in pending
+               receipts = &chain.get_pending_trx_receipts();
```
heifner: This seems rather fragile to me. If `get_pending_trx_receipts()` changes this could break easily. How about this:
```
if( blk || chain.is_building_block() ) {
   const vector<transaction_receipt>& receipts = blk ? blk->transactions : chain.get_pending_trx_receipts();
...
```

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -45,46 +54,57 @@ struct pending_block_header_state {
...
+struct block_header_state : public detail::block_header_state_common {
+   block_id_type                     id;
+   signed_block_header               header;
+   detail::schedule_info             pending_schedule;
+
+   block_header_state() = default;
+
+   block_header_state( detail::block_header_state_common&& base )
+   :detail::block_header_state_common( std::move(base) )
+   {}
```
heifner: explicit constructor here would provide guard against accident creation from block_header_state_common.

arhag: Resolved in https://github.com/EOSIO/eos/pull/6624/files#diff-3701e88f728b2bd0a37f5d347a042124R64.

libraries/chain/include/eosio/chain/block_header_state.hpp
arhag: Resolved in https://github.com/EOSIO/eos/pull/6588/commits/2cb4dbc8825af74f45c18f5b35d030d2888bfd3c.

plugins/history_plugin/history_plugin.cpp
arhag: Resolved in https://github.com/EOSIO/eos/pull/6588/commits/5961a38c7abb418eae4342c1f8a6d5e92eb61170.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -5,6 +5,42 @@
 
 namespace eosio { namespace chain {
 
+struct block_header_state;
+
+struct pending_block_header_state {
```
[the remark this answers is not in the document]
arhag: Resolved in https://github.com/EOSIO/eos/pull/6588/commits/be795a8fad651857a9b467eff2695fd1cac3195f.

PR #6589, opened 2019-01-12
Use https url for wabt submodule

coderobe:
## Change Description

A user agent that doesn't respect hsts could potentially be coerced into downloading malicious sources for wabt via a mitm attack. Prevent this by using a https upstream like the other submodules already do

## Consensus Changes

n/a

## API Changes

n/a

## Documentation Additions

n/a

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -14,4 +14,4 @@
 	url = https://github.com/EOSIO/fc
 [submodule "libraries/wabt"]
 	path = libraries/wabt
-	url = http://github.com/EOSIO/wabt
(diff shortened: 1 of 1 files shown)


spoonincode:
This is good but I assume it only works for fresh clones? Even after checking out your branch, doing a submodule deinit+init+update, my submodule still seems to be using http. Do you know of a good & proper way to update this in place too?

coderobe:
You can probably change this in-place using `git config`, the config key on the submodule is `url`.

coderobe:
@spoonincode done

PR #6591, opened 2019-01-12
Fix incorrect query with 256-bit integer key

conr2d:
## Change Description

There is an issue when making a query with 256-bit integer key by `get_table_rows` API, because `::sha256()` puts binary from the lowest address, but `eosio::checksum256` (`eosio::fixed_bytes<32>`) rearranges binary by `uint128_t` unit.

```
struct [[eosio::table]] row {
   uint64_t    id;
   std::string nickname;
   uint64_t primary_key()const { return id; }
   checksum256 secondary_key()const { return sha256(nickname.data(), nickname.size()); }
}

typedef eosio::multi_index<"table"_n, row,
   indexed_by<"secondary"_n, const_mem_fun<row, checksum256, &row::secondary_key>
> table_index;
```

For example, if user add a row into multi_index like above with `nickname = "eosio"`, the hashed value with `::sha256()` is `efbb5450f4528e0dd33ff1873f52a175` `e8f280eb591c6725a1227b9aec8eac0e`. However, `eosio::checksum256` will be `75a1523f87f13fd30d8e52f45054bbef` `0eac8eec9a7b22a125671c59eb80f2e8`, so currently `get_table_rows` with 256-bit integer doesn't work in intuitive way. User cannot use hashed value directly, but need to change the sequence of binary.

This patch converts the passed hash value to fit the type [`eosio::chain::key256_t`](https://github.com/EOSIO/eos/blob/master/libraries/chain/include/eosio/chain/contract_table_objects.hpp#L129) when calling `get_table_rows`. 

## Consensus Changes

None.

## API Changes

None.

## Documentation Additions

None.

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -597,6 +597,37 @@ public:
  constexpr const char dec[]       = "dec";
  constexpr const char hex[]       = "hex";
 
+ template<typename Word>
+ static void set_from_word_sequence(const Word* arr_begin, const Word* arr_end, chain::key256_t& key) {
+    using word_t = uint128_t;
+
+    auto itr = key.begin();
+    word_t temp_word = 0;
+    const size_t sub_word_shift = 8 * sizeof(Word);
+    const size_t num_sub_words = sizeof(word_t) / sizeof(Word);
+    auto sub_words_left = num_sub_words;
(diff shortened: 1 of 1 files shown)


conr2d:
@arhag 

Do you think this patch breaks existing behavior? If then, I would add another option like `--bypass` not to change the sequence of binary instead of modifying original code.

wangyilin1987:
Will this problem be solved in the future?

conr2d:
The changes in this PR is applied by #7530.

Issue #6409, opened 2018-11-28
About irreversible block do not increase. Very incredibling!


anlebcoder:
Hi,EOS TEAM:

When BP died(<2/3), the irreversible height did not increase for a period of time. For example, stop for 10 minutes

Ten minutes later, we revived BP（>2/3） and the irreversible height began to grow.

You'll find that we're here(on_applied_irreversible_block) to find that the irreversible heights are about 10 minutes slower to receive messages.

 my->irreversible_block_connection.emplace(
               chain.irreversible_block.connect( [&]( const chain::block_state_ptr& bs ) {
                  my->applied_irreversible_block( bs );
               } ));
why? It's incredible.

jgiszczak:
You seem to be asserting that one node of your testnet is increasing the last irreversible block, but 10 minutes after the other nodes increase it to the same height.  I can assure you, whatever you are seeing, that's not what's happening.  The software does not work that way.  A producer network with 10 minutes of latency is not functional and will not advance the last irreversible block at all.

Please attempt a different description of what you are seeing, as I am sure my interpretation must not be correct.

anlebcoder:
Hello, maybe I didn't express it clearly.

I think the main network has this problem now

When the whole network survives less than two-thirds of the nodes, you will find that the irreversible height will stop growing, but the reversible height will increase.


After this state lasts for 10 minutes, if you restore the node to more than two-thirds, then you will find that the irreversible height starts to increase again.

Then there is the phenomenon I explained above, you will find

The log output on the screen is similar to this:

```
2018-11-29T03:52:14.031 thread-1 producer_plugin.cpp:332 on_incoming_block ] Received block a6c6c7803f6cbfb3... #15918315 @ 2018-11-29T03:52:14.000 signed by eosteaeostea [trxs: 0, lib: 15917986, conf : 0, latency: 31 ms]
```

#15918315 is Reversible height,
Lib:15917986 is irreversible height.

They differ by more than 300 pieces, why are there more than 300 pieces, I am very clear about the principle.

(2 * 21 * (2/3) * 12 = 360 )


Then the problem comes, we see that the output in the log is 15917986 - 15918315 = 329, this is the correct expectation.

In addition `get_info` is also the correct result, but you will find out if the code outputs irreversible height information:

```
 My->irreversible_block_connection.emplace(
               Chain.irreversible_block.connect( [&]( const chain::block_state_ptr& bs ) {
                  My->applied_irreversible_block( bs );
               } ));
```

[...]

PR #6624, opened 2019-01-17
Refactor fork database

arhag:
Builds on top of PR #6588.

This PR allows irreversible mode to be enabled. It also fixes some bugs that were only triggered during irreversible mode.

This PR also refactors the fork database code to ensure that blocks that become irreversible are immediately appended to the block log, which resolves issues such as #6409. 

This refactor changes the format of the portable serialized fork database file that is stored on exit of nodeos (in addition to the existing changes due to #6588). First, the file is saved as "fork_db.dat" rather than "forkdb.dat" to signal that the format has changed in an important way and also to force a replay which is going to be required anyway for other reasons. This new format has a magic number at the start to help distinguish the file as a serialized fork database file, and more importantly it now has a version number to enable future upgrades to the format without necessarily having to abandon support for reading from the old versions going forward.

The new fork database code distinguishes the best head in the fork tree when only considering fully validated blocks it knows about from the best head in the fork tree when considering all blocks it knows about that have at least passed block header validation; this latter type of head is called the pending head in the fork database code.

In modes other than irreversible mode, pending head and the regular head should be identical except for the brief period of time between adding the block to the fork database and applying it in `maybe_switch_forks`. If the block added to the fork database does not cause any new block to be applied, then the pending head and regular head still remain the same because the pending head would not have changed due to the addition of the block to the fork database (again this is in modes other than irreversible mode).

In irreversible mode, the pending head becomes far more relevant. In that mode, the regular head and pending head are expected to be different. In fact, the regular head block should be the irreversible block from the perspective of the pending head, and so it will be an ancestor of the pending head block. 

Two new fields are added to the results returned by `get_info` in the chain plugin: `fork_db_head_block_num` and `fork_db_head_block_id`. These represent the best known head in the fork database tree. The idea is that `get_info` now provides information for three blocks in the blockchain, where two of the three blocks should always be identical in the `get_info` response. The three blocks are: the last irreversible block (or `last_irreversible_block`); the best known block in the fork database (known as the `fork_db_head_block`); and the block that is tracked by the current consensus state of node (known simply as the `head_block`). In irreversible mode, the `head_block` is equivalent to the `last_irreversible_block`. In all other modes, the `head_block` is equivalent to the `fork_db_head_block`. 

This PR also adds a unit test(`forked_tests/irreversible_mode`) which checks that a node in irreversible mode will properly be following the irreversible block even as two other nodes feed it conflicting blocks which cause it to do fork switches. 

I have identified further cleanup/refactoring of the code that try to find the appropriate block given a block number, but I am going to hold off on that for the time being so that I can continue making progress with the other features needed for protocol feature foundations.


arhag:
Since this PR also merges in the latest changes from develop but is targeting a branch that is not up to date with develop, it brings with it a lot more code changes than the changes the PR is primarily meant to make as discussed in the description of the PR.

The secondary goal of this PR is to bring the `refactors-for-protocol-feature-foundations` branch up to date with the latest develop (commit 22fa35a7cc3ca9105c186cedd767db55d40337fa at the time of this writing) in order to reduce merge conflicts later on when the relevant protocol features branch will be merged into develop.

Even though it brings in changes from #6588 which may have already been reviewed, I recommend referencing the following diff to assist with the review:

https://github.com/EOSIO/eos/compare/22fa35a7cc3ca9105c186cedd767db55d40337fa...7d06096382eeec3743f7893bd8987b022e0ef4f0

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1382,39 +1409,44 @@ void producer_plugin_impl::schedule_production_loop() {
...
+                  fc_dlog( _log, "Producing Block #${num} returned: ${res}", ("num", block_num)( "res", res ) );
+               }
+            } ) );
    } else if (_pending_block_mode == pending_block_mode::speculating && !_producers.empty() && !production_disabled_by_policy()){
       fc_dlog(_log, "Specualtive Block Created; Scheduling Speculative/Production Change");
-      EOS_ASSERT( chain.pending_block_state(), missing_pending_block_state, "speculating without pending_block_state" );
-      const auto& pbs = chain.pending_block_state();
-      schedule_delayed_production_loop(weak_this, pbs->header.timestamp);
+      EOS_ASSERT( chain.is_building_block(), missing_pending_block_state, "speculating without pending_block_state" );
+      schedule_delayed_production_loop(weak_this, chain.pending_block_time());
```
heifner: For consistence use `pending_block_time` from above.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -104,6 +104,8 @@ class read_only {
       //string                  recent_slots;
       //double                  participation_rate = 0;
       optional<string>        server_version_string;
+      uint32_t                fork_db_head_block_num = 0;
+      chain::block_id_type    fork_db_head_block_id;
    };
```
heifner: Need to be optional to allow new cleos to work with old nodeos as cleos main.cpp uses `call(url, get_info_func).as<eosio::chain_apis::read_only::get_info_results>()`

libraries/chain/fork_database.cpp
```diff
@@ -23,27 +32,31 @@ namespace eosio { namespace chain {
...
+               member<detail::block_header_state_common, uint32_t,      &detail::block_header_state_common::block_num>,
+               member<block_header_state,                block_id_type, &block_header_state::id>
             >,
-            composite_key_compare< std::greater<uint32_t>, std::greater<uint32_t>, std::greater<uint32_t> >
+            composite_key_compare<
+               std::greater<bool>,
+               std::greater<uint32_t>,
+               std::greater<uint32_t>,
+               std::less<block_id_type>
+            >
```
heifner: Since replay is needed anyway, this is a good time to use the faster less for sha256:
```
   struct sha256_less {
      bool operator()( const sha256& lhs, const sha256& rhs ) const {
       return
             std::tie(lhs._hash[0], lhs._hash[1], lhs._hash[2], lhs._hash[3]) <
             std::tie(rhs._hash[0], rhs._hash[1], rhs._hash[2], rhs._hash[3]);
      }
   };
```

libraries/chain/fork_database.cpp
```diff
@@ -57,48 +70,135 @@ namespace eosio { namespace chain {
...
+
+            fc::datastream<const char*> ds( content.data(), content.size() );
+
+            // validate totem
+            uint32_t totem = 0;
+            fc::raw::unpack( ds, totem );
+            EOS_ASSERT( totem == magic_number, fork_database_exception,
+                        "Fork database file '${filename}' has unexpected magic number!",
+                        ("filename", fork_db_dat.generic_string())
+            );
```
heifner: Report totem so user knows what was read.

libraries/chain/fork_database.cpp
```diff
@@ -57,48 +70,135 @@ namespace eosio { namespace chain {
...
+            fc::raw::unpack( ds, bhs );
+            reset( bhs );
+
+            unsigned_int size; fc::raw::unpack( ds, size );
+            for( uint32_t i = 0, n = size.value; i < n; ++i ) {
+               block_state s;
+               fc::raw::unpack( ds, s );
+               for( const auto& receipt : s.block->transactions ) {
+                  if( receipt.trx.contains<packed_transaction>() ) {
+                     auto& pt = receipt.trx.get<packed_transaction>();
```
heifner: `const auto&` would make me feel better.

libraries/chain/fork_database.cpp
```diff
@@ -209,71 +357,47 @@ namespace eosio { namespace chain {
       return result;
    } /// fetch_branch_from
 
-   /// remove all of the invalid forks built of this id including this id
+   /// remove all of the invalid forks built off of this id including this id
    void fork_database::remove( const block_id_type& id ) {
       vector<block_id_type> remove_queue{id};
+      auto& previdx = my->index.get<by_prev>();
+      const auto head_id = my->head->id;
```
heifner: This can be `const auto&`

libraries/chain/fork_database.cpp
```diff
@@ -108,64 +208,112 @@ namespace eosio { namespace chain {
...
+      if( itr != my->index.end() )
+         return *itr;
+
+      return block_header_state_ptr();
    }
 
-   block_state_ptr fork_database::add( signed_block_ptr b, bool skip_validate_signee ) {
-      EOS_ASSERT( b, fork_database_exception, "attempt to add null block" );
-      EOS_ASSERT( my->head, fork_db_block_not_found, "no head block set" );
+   void fork_database::add( block_state_ptr n ) {
```
heifner: `n` is not moved from so can be passed as `const&`

libraries/chain/fork_database.cpp
```diff
@@ -108,64 +208,112 @@ namespace eosio { namespace chain {
...
+      EOS_ASSERT( my->root, fork_database_exception, "root not yet set" );
+      EOS_ASSERT( n, fork_database_exception, "attempt to add null block state" );
 
-      const auto& by_id_idx = my->index.get<by_block_id>();
-      auto existing = by_id_idx.find( b->id() );
-      EOS_ASSERT( existing == by_id_idx.end(), fork_database_exception, "we already know about this block" );
+      EOS_ASSERT( get_block_header( n->header.previous ), unlinkable_block_exception,
+                  "unlinkable block", ("id", string(n->id))("previous", string(n->header.previous)) );
```
heifner: Not that it matters, but the `string()` wrappers are not needed.

libraries/chain/fork_database.cpp
```diff
@@ -108,64 +208,112 @@ namespace eosio { namespace chain {
...
 
-      auto prior = by_id_idx.find( b->previous );
-      EOS_ASSERT( prior != by_id_idx.end(), unlinkable_block_exception, "unlinkable block", ("id", string(b->id()))("previous", string(b->previous)) );
+      auto inserted = my->index.insert(n);
+      EOS_ASSERT( inserted.second, fork_database_exception, "duplicate block added" );
```
heifner: Maybe add the block id.

libraries/chain/fork_database.cpp
```diff
@@ -57,48 +70,135 @@ namespace eosio { namespace chain {
...
+            reset( bhs );
+
+            unsigned_int size; fc::raw::unpack( ds, size );
+            for( uint32_t i = 0, n = size.value; i < n; ++i ) {
+               block_state s;
+               fc::raw::unpack( ds, s );
+               for( const auto& receipt : s.block->transactions ) {
+                  if( receipt.trx.contains<packed_transaction>() ) {
+                     auto& pt = receipt.trx.get<packed_transaction>();
+                     s.trxs.push_back( std::make_shared<transaction_metadata>( std::make_shared<packed_transaction>(pt) ) );
```
heifner: Would be nice to have a test that failed before that now works for transactions in fork_db. Maybe this is something @brianjohnson5972 or @taokayan could create.

libraries/chain/controller.cpp
```diff
@@ -1226,36 +1374,37 @@ struct controller_impl {
...
-         EOS_ASSERT(producer_block_id == pending->_pending_block_state->header.id(),
-                   block_validate_exception, "Block ID does not match",
-                   ("producer_block_id",producer_block_id)("validator_block_id",pending->_pending_block_state->header.id()));
+         auto id = block_ptr->id();
+         EOS_ASSERT( producer_block_id == id, block_validate_exception, "Block ID does not match",
+                     ("producer_block_id",producer_block_id)("validator_block_id",id) );
 
-         // We need to fill out the pending block state's block because that gets serialized in the reversible block log
-         // in the future we can optimize this by serializing the original and not the copy
+         auto&& ab = pending->_block_stage.get<assembled_block>();
```
heifner: Here and in `controller::finalize_block` I don't like the inconsistency of returning signed_block from finalize_block but also reaching into `pending->_block_stage.get<assembled_block>()`. Seems you could either not return anything from `finalize_block` or return the block_stage. In either case if id was added to assembled_block then it would not have to be recalculated via `id()`.

libraries/chain/controller.cpp
```diff
@@ -2003,6 +2227,8 @@ int64_t controller::set_proposed_producers( vector<producer_key> producers ) {
 
    int64_t version = sch.version;
 
+   wlog( "proposed producer schedule with version ${v}", ("v", version) );
+
```
heifner: This is not really a warning right? `ilog` instead?

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -5,63 +5,106 @@
...
-    void                 verify_signee(const public_key_type& signee)const;
-};
+struct block_header_state : public detail::block_header_state_common {
+   block_id_type                     id;
+   signed_block_header               header;
+   detail::schedule_info             pending_schedule;
+
+   block_header_state() = default;
+
+   block_header_state( detail::block_header_state_common&& base )
```
heifner: Make explicit

plugins/producer_plugin/producer_plugin.cpp
arhag: I'm not following. Which `pending_block_time` are you referring to?

heifner: That was me looking at the diff wrong. Ignore.

arhag:
With the changes here and in #6842, nodeos now supports an irreversible read-mode with the ability to easily switch back and forth between irreversible mode and the other read modes.

### Switching from speculative mode to irreversible mode

Shut down nodeos and then restart it with the `--read-mode=irreversible` command line option (the configuration can also be added to the config.ini file).

At startup the nodeos logs should print a warning that looks something like the following:
```
warn  2019-04-22T21:51:31.318 thread-0  controller.cpp:576            init                 ] database revision (1799) is greater than head block number (1798), attempting to undo pending changes
warn  2019-04-22T21:51:31.320 thread-0  controller.cpp:591            init                 ] read_mode has changed to irreversible: erasing reversible blocks
```

These warnings will only appear on the first restart that switches from `speculative` mode (or `head` or `read-only` modes) to `irreversible` mode. Further restarts of nodeos that keep it in `irreversible` mode will not display these messages.

As the warning states, the reversible blocks database is erased. However, those blocks should still exist in the fork database (which is stored in the `fork_db.dat` file in the `state` directory on nodeos exit). This allows nodeos to recover the reversible blocks if it switches back to speculative mode.

While in irreversible mode, the current state always reflects the state as of the end of the last irreversible block. This can be confirmed by comparing the `head_block_num` with the `last_irreversible_block_num` in the result of the `get_info` chain API RPC; the two block numbers should be the same which will likely be less than `fork_db_head_block_num`.

###  Switching from irreversible mode to speculative mode

Shut down nodeos while it is configured to be in irreversible mode. Then restart nodeos with the `--read-mode=speculative` command line option (this is currently the default mode if not `read-mode` configuration is specified).

At startup the nodeos logs should print a warning that looks something like the following:
```
warn  2019-04-22T21:59:30.276 thread-0  controller.cpp:657            init                 ] read_mode has changed from irreversible: applying best branch from fork database
warn  2019-04-22T21:59:30.276 thread-0  controller.cpp:663            init                 ] applying branch from fork database ending with block: 00000707f8606d21d07f4342975e677f41369f60514581a81693b3f5c9887ee2
```
Nodeos will recover the best available branch from the fork database and reapply those blocks. Doing so will add those blocks back into the reversible blocks database. It will also bring the current state back into sync with that of the best head block. This can be confirmed by comparing the `head_block_num` with the `fork_db_head_block_num` in the result of the `get_info` chain API RPC; the two block numbers should be the same which will likely be greater than `last_irreversible_block_num`.

PR #6625, opened 2019-01-17
get_code_id intrinsic API

conr2d:
## Change Description

`hardfork` is required. User can check whether code is changed before sending an inline action.

## Consensus Changes

None.

## API Changes

`void get_code_id(capi_name code, capi_checksum256* hash);` would be added into `eosiolib` in CDT.

## Documentation Additions

Description for new contract API `get_code_id` would be added.


b1bart:

Issue #4033, opened 2018-06-11
Setting Permission Active&Owner using SAME account LOCKS YOU OUT


mikefisher2k:
Given the account "mikefisher2k"

If you run the following commands,

cleos set account permission mikefisher2k active '{"threshold":1,"keys":[],"accounts":[{"permission":{"actor":"mikefisher2k","permission":"active"},"weight":1}],"waits":[]}' owner -p mikefisher2k@owner

cleos set account permission mikefisher2k owner '{"threshold":1,"keys":[],"accounts":[{"permission":{"actor":"mikefisher2k","permission":"owner"},"weight":1}],"waits":[]}' -p mikefisher2k@owner

You will be setting the ACTIVE and OWNER of mikefisher2k to

mikefisher2k@active
mikefisher2k@owner

THIS SHOULD NOT BE ALLOWED

Setting the same account as active and owner essentially locks you out of the account

This is technically a bug...and should NOT be allowed to happen....

Is there anyway to recover from such a situation?!?

PR #6631, opened 2019-01-17
 Add support for setting authority with permission level 

conr2d:
## Change Description

Following patch of #6583. This PR allows users to set authority with permission level during not only account creation (`cleos create account`) but setting account permission (`cleos set account permission`).
However, by allowing this, the possibility of locking up account by mistake can increase like #4033. It seems that there's already progress about this issue (#4037), so feel free to reject this if this PR conflicts with someone's work.

## Consensus Changes

None.

## API Changes

None.

## Documentation Additions

Both `cleos create account` and `cleos set account permission` can set authority by public key, permission level, JSON string or file. This needs to be added to documentation.


b1bart:

PR #6633, opened 2019-01-18
Set minimum dependencies for Debian packages.

### LINKED ISSUE #6618: nodeos 1.6.0 deb package should not depend on development packages
Why does a pre-compiled binary package depend on development packages?  (libbz2-dev, libssl-dev, libgmp3, build-essential, libicu-dev, zlib1g-dev)

This came up during the discussion on the PR of the fix for https://github.com/EOSIO/eos/issues/6590

dpkg --info eosio_1.6.0-1-ubuntu-18.04_amd64.deb
```
 new Debian package, version 2.0.
 size 15149956 bytes: control archive=432 bytes.
     315 bytes,     9 lines      control              
 Package: eosio
 Version: 1.6.0-1
 Section: devel
 Priority: optional
 Depends: libbz2-dev (>= 1.0), libssl-dev (>= 1.0), libgmp3-dev, build-essential, libicu-dev, zlib1g-dev, libtinfo5
 Architecture: amd64
 Homepage: https://github.com/eosio/eos
 Maintainer: support@block.one
 Description: Software for the EOS.IO network
```

jgiszczak:
## Change Description

Set minimum dependencies for Debian packages. Also ensure Debian subdirectory has correct permissions regardless of umask while packaging.  Resolves #6618.

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

### DIFF
--- a/scripts/generate_deb.sh
+++ b/scripts/generate_deb.sh
@@ -33,14 +33,14 @@ fi
 
 mkdir -p ${PROJECT}/DEBIAN
 chmod 0755 ${PROJECT}/DEBIAN
-echo "Package: ${PROJECT} 
+echo "Package: ${PROJECT}
 Version: ${VERSION_NO_SUFFIX}-${RELEASE}
 Section: devel
 Priority: optional
 Depends: libc6, libgcc1, ${LIBSSL}, libstdc++6, libtinfo5, zlib1g
 Architecture: amd64
-Homepage: ${URL} 
-Maintainer: ${EMAIL} 
+Homepage: ${URL}
+Maintainer: ${EMAIL}
 Description: ${DESC}" &> ${PROJECT}/DEBIAN/control
 
 export PREFIX
@@ -50,8 +50,8 @@ export SSUBPREFIX
 
 bash generate_tarball.sh ${NAME}.tar.gz
 
-tar -xvzf ${NAME}.tar.gz -C ${PROJECT} 
-dpkg-deb --build ${PROJECT} 
+tar -xvzf ${NAME}.tar.gz -C ${PROJECT}
+dpkg-deb --build ${PROJECT}
 BUILDSTATUS=$?
 mv ${PROJECT}.deb ${NAME}.deb
 rm -r ${PROJECT}
--- a/scripts/generate_deb.sh
+++ b/scripts/generate_deb.sh
@@ -15,14 +15,14 @@ NAME="${PROJECT}_${VERSION_NO_SUFFIX}-${RELEASE}_amd64"
 
 mkdir -p ${PROJECT}/DEBIAN
 chmod 0755 ${PROJECT}/DEBIAN
-echo "Package: ${PROJECT}
+echo "Package: ${PROJECT} 
 Version: ${VERSION_NO_SUFFIX}-${RELEASE}
 Section: devel
 Priority: optional
-Depends: libbz2-dev (>= 1.0), libssl-dev (>= 1.0), libgmp3-dev, build-essential, libicu-dev, zlib1g-dev, libtinfo5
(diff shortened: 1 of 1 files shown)


scripts/generate_deb.sh
```diff
@@ -13,16 +13,13 @@ fi
...
-   DEPS_STR="${DEPS_STR} Depends: ${dep}"
-done
 mkdir -p ${PROJECT}/DEBIAN
+chmod 0755 ${PROJECT}/DEBIAN
 echo "Package: ${PROJECT} 
 Version: ${VERSION_NO_SUFFIX}-${RELEASE}
 Section: devel
 Priority: optional
-Depends: libbz2-dev (>= 1.0), libssl-dev (>= 1.0), libgmp3-dev, build-essential, libicu-dev, zlib1g-dev
+Depends: libc6, libgcc1, libssl1.0.0, libstdc++6, libtinfo5, zlib1g
```
spoonincode: > libssl1.0.0

This would seem to imply that the .deb *must* be built on 16.04. Is that intended?

jgiszczak: As far as I can see it's absolutely required because of libc6 and libstdc++.  In my testing, if I built on a later Ubuntu then installed on 16, the resulting binaries fail to run, complaining that version CXXABI_1.3.11 is not found.  On Ubuntu 16, the out-of-the-box libstdc++ stops with CXXABI_1.3.9.

While it is apparently possible to get CXXABI_1.3.11 on Ubuntu 16, it requires installing gcc, and all that brings with it.  It was my intent to avoid such dramatic alteration of the system if it could be avoided.

spoonincode: We currently provide two .deb files though, one for 16.04 and one for 18.04. These are clearly built on the respective ubuntu versions.

The problem with putting `libssl1.0.0` here is that a deb built on 18.04 won't have the right dependencies because 1.1 is what is used on 18.04. (1.1 & 1.0 are not ABI compatible)

jgiszczak: That's a problem then, because there is no `libssl` meta package.  Its name on Ubuntu 16 is `libssl1.0.0` and there is no way to get only the library package by another name.  Debian dependency parsing does not support wildcards, either.  Since we're eschewing debhelper, our alternative is to hack in Ubuntu version recognition and substitute that package name depending on the result.

spoonincode:
Would it be a better approach to derive the value from something like
`ldd nodeos | grep libssl.so.1.1`
So we didn't have to baby the script on new releases? Hard to say... new releases may do something wildly different anyways. I think I'm mostly indifferent, at least this gets us the correct values for the two versions we support.

jgiszczak:
I used
```bash
$ ldd ./nodeos | awk '/=>/{print $(NF-1)}' | while read n; do dpkg-query -S $n; done | sed 's/^\([^:]\+\):.*$/\1/' | sort | uniq
```
(and again for keosd and cleos) to compose the dependency list and even that results in a spurious error from dpkg-query, as the '=>' from ldd isn't successfully filtered out.  That mess comes very close, but I hesitate to call it robust enough to use it as a fire and forget package builder.  That's what debhelper is for.

scripts/generate_deb.sh
```diff
@@ -32,8 +50,8 @@ export SSUBPREFIX
 
 bash generate_tarball.sh ${NAME}.tar.gz
 
-tar -xvzf ${NAME}.tar.gz -C ${PROJECT}
-dpkg-deb --build ${PROJECT}
+tar -xvzf ${NAME}.tar.gz -C ${PROJECT}
```
spoonincode: can you fix the trailing whitespace in this file (several places)

PR #6638, opened 2019-01-18
new build scripts, cmakelists, buildkite pipeline, and cleanup script

NorseGaud:
### Build Script changes:
1. Variables used throughout are now located in eosio_build.sh (D.R.Y)
2. BOOST, Mongo+mongolibs, CMAKE, and others are now installed into and using ~/lib,bin,etc,data (Eliminates need for root as much as possible)
3. MongoDB launch command uses ~/ paths for configs, logs, etc (CI/CD pipeline as well)
4. Installation from source is closer to what the developer portal suggests
5. Ubuntu build script MongoDB installation accounts for 16.04 or 18.04 versions (18 gets v4, vs v3; v3 doesn't exist for 18)
6. Wrapped apt/yum/brew update with warning
7. Using custom homebrew formula, which forces us to use the manually built cmake version under ~/bin
8. Better cleanup script to handle old installs or new
9. Fix for gettext being keg-only

### CMakeModules changes:
1. LLVM/WASM path is now ~/opt/llvm + FindWasm using ~/opt/llvm for both darwin and linux

### Test changes
1. tests/nodeos_run_test.py -> Setting currentBlockNum later in code so that a slow machine that generates too many blocks won't fail the tests

### Notes:
- We can now copy and paste almost everything between OS build scripts. (Time savings for maintenance)
- We have consistency between Linux and Mac paths.
- Build script are used to create images, "shifting left" the responsibility of maintaining them to DevOps, instead of both Dockerfiles AND build scripts.
- Output of build scripts are much cleaner and easier to follow.

### VM Image Speed Improvements:
- mojave build from 8m to 5m
- centos build from 15m to 8m

### DIFF
--- a/scripts/eosio_build_centos.sh
+++ b/scripts/eosio_build_centos.sh
@@ -302,7 +302,7 @@ function print_instructions()
 	printf "export LD_LIBRARY_PATH=\$HOME/lib:\$HOME/lib64:\$HOME/opt/llvm/lib:\$LD_LIBRARY_PATH\\n" # libmongoc is installed into $HOME/lib64
 	printf "export CMAKE_MODULE_PATH=\$HOME/lib/cmake\\n"
 	printf "export CPATH=\$HOME/include:\$CPLUS_INCLUDE_PATH\\n" # libmongoc is installed into $HOME/include
-	printf "Then, source the file (or restart the terminal), and run:"
+	printf "Then, source the file (or restart the terminal) and run:"
 	printf "$( command -v mongod ) --dbpath ${MONGODB_DATA_LOCATION} -f ${MONGODB_CONF} --logpath ${MONGODB_LOG_LOCATION}/mongod.log &\\n"
 	printf "cd ${BUILD_DIR} && make test\\n"
 	return 0
--- a/scripts/eosio_build_centos.sh
+++ b/scripts/eosio_build_centos.sh
@@ -302,6 +302,7 @@ function print_instructions()
 	# llvm/bin last to prevent llvm/bin/clang from being used over /usr/bin/clang (We don't symlink into $HOME/bin)
 	printf "export PATH=\$HOME/bin:\$PATH:$MONGODB_LINK_LOCATION/bin:\$HOME/opt/llvm/bin\\n"
 	printf "export LD_LIBRARY_PATH=\$HOME/lib:\$HOME/lib64:\$HOME/opt/llvm/lib:\$LD_LIBRARY_PATH\\n" # libmongoc is installed into $HOME/lib64
+	printf "export CMAKE_MODULE_PATH=\$HOME/lib/cmake\\n"
 	printf "export CPATH=\$HOME/include:\$CPLUS_INCLUDE_PATH\\n" # libmongoc is installed into $HOME/include
 	printf "$( command -v mongod ) --dbpath ${MONGODB_DATA_LOCATION} -f ${MONGODB_CONF} --logpath ${MONGODB_LOG_LOCATION}/mongod.log &\\n"
 	printf "cd ${BUILD_DIR} && make test\\n"
--- a/scripts/eosio_build_centos.sh
+++ b/scripts/eosio_build_centos.sh
@@ -51,46 +51,40 @@ if ! YUM=$( command -v yum 2>/dev/null ); then
 fi
 printf " - Yum installation found at %s.\\n" "${YUM}"
 
-printf "\\nDo you wish to update YUM repositories?\\n\\n"
-select yn in "Yes" "No"; do
-	case $yn in
-		[Yy]* ) 
-			printf "\\n\\nUpdating...\\n\\n"
-			if ! "${YUM}" -y update; then
-				printf "\\nYUM update failed.\\n"
-				printf "\\nExiting now.\\n\\n"
-				exit 1;
-			else
-				printf "\\nYUM update complete.\\n"
-			fi
-		break;;
-		[Nn]* ) 
-			echo "Proceeding without update!"
-		break;;
-		* ) echo "Please type 1 for yes or 2 for no.";;
-	esac
-done
+if [ $1 == 0 ]; then read -p "Do you wish to update YUM repositories? (y/n)? " answer; fi
+case ${answer} in
+	1 | [Yy]* ) 
+		printf "\\n\\nUpdating...\\n\\n"
+		if ! "${YUM}" -y update; then
+			printf "\\nYUM update failed.\\n"
+			printf "\\nExiting now.\\n\\n"
+			exit 1;
+		else
+			printf "\\nYUM update complete.\\n"
+		fi
+	;;
+	[Nn]* ) echo "Proceeding without update!";
+	* ) echo "Please type 1 for yes or 2 for no.";;
+esac
 
 printf "Checking installation of Centos Software Collections Repository...\\n"
 SCL=$( rpm -qa | grep -E 'centos-release-scl-[0-9].*' )
 if [ -z "${SCL}" ]; then
-	printf " - Do you wish to install and enable this repository?\\n"
-	select yn in "Yes" "No"; do
-		case $yn in
-			[Yy]* )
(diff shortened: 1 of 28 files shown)


scripts/eosio_build_centos.sh
```diff
@@ -1,698 +1,312 @@
...
+	printf " - MongoDB C++ driver successfully installed @ ${MONGO_CXX_DRIVER_ROOT}.\\n"
+else
+	printf " - MongoDB C++ driver found with correct version @ ${MONGO_CXX_DRIVER_ROOT}.\\n"
+fi
+
+
+printf "\\n"
+
+
+printf "Checking LLVM with WASM support...\\n"
```
spoonincode: The wording here (and a handful of other places) needs to be changed. This is only building LLVM4 libraries. Nothing to do with wasm.

CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 if (UNIX)
    if (APPLE)
       if (LLVM_DIR STREQUAL "" OR NOT LLVM_DIR)
-         set(LLVM_DIR "/usr/local/Cellar/llvm@4/4.0.1/lib/cmake/llvm")
+         set(LLVM_DIR "$ENV{HOME}/opt/llvm/lib/cmake/llvm")
```
spoonincode: LLVM_DIR is a cached variable does this change work correctly on a preexisting checkout?

scripts/eosio_build_darwin.sh
```diff
@@ -1,406 +1,271 @@
...
 	fi
+	DEPS=$DEPS"${brewname},"
+	DISPLAY="${DISPLAY}${COUNT}. ${name}\\n"
+	printf " - %s ${bldred}NOT${txtrst} found.\\n" "${name}"
+	(( COUNT++ ))
+done < "${CURRENT_DIR}/scripts/eosio_build_darwin_deps"
+IFS="${var_ifs}"
 
+if [ ! -d /usr/local/Frameworks ]; then
+	printf "\\n${bldred}/usr/local/Framworks is necessary to brew install python@3. Run the following commands as sudo and try again:${txtrst}\\n"
```
spoonincode: typo

scripts/eosio_build_darwin_deps
```diff
@@ -0,0 +1,10 @@
+automake,-x,/usr/local/bin/automake,automake,http://ftp.gnu.org/gnu/automake/automake-1.15.tar.gz
+Libtool,-x,/usr/local/bin/glibtool,libtool,http://gnu.askapache.com/libtool/libtool-2.4.6.tar.gz
+OpenSSL,-f,/usr/local/opt/openssl/lib/libssl.a,openssl,https://www.openssl.org/source/openssl-1.0.2n.tar.gz
+wget,-x,/usr/local/bin/wget,wget,https://ftp.gnu.org/gnu/wget/wget-1.19.2.tar.gz
+GMP,-f,/usr/local/opt/gmp/include/gmpxx.h,gmp,https://ftp.gnu.org/gnu/gmp/gmp-6.1.2.tar.bz2
+gettext,-x,/usr/local/opt/gettext/bin/gettext,gettext,https://ftp.gnu.org/pub/gnu/gettext/gettext-latest.tar.gz
+llvm,-x,/usr/local/opt/llvm@4/bin/clang-4.0,llvm@4,http://releases.llvm.org/5.0.1/llvm-5.0.1.src.tar.xz
```
spoonincode: I don't think the download link is used here but still pretty wonky there is a 5.0.1 tar ball for a 4.0 dependency

scripts/eosio_build_darwin.sh
```diff
@@ -1,406 +1,271 @@
...
-    if [ -e "/usr/local/lib/libmongocxx-static.a" ]; then
-		MONGO_INSTALL=false
-		if ! version=$( grep "Version:" /usr/local/lib/pkgconfig/libmongocxx-static.pc | tr -s ' ' | awk '{print $2}' )
-		then
-			printf "\\tUnable to determine mongodb-cxx-driver version.\\n"
-			printf "\\tExiting now.\\n\\n"
-			exit 1;
-		fi
+# Failed to find Gettext libintl (missing: Intl_INCLUDE_DIR); known bug
+brew unlink gettext 2> /dev/null && brew link --force gettext
```
spoonincode: I recommend not force linking anything.

scripts/eosio_build_fedora.sh
```diff
@@ -1,500 +1,252 @@
...
+	done
+else
+	printf " - No required YUM dependencies to install.\\n"
+fi
+
+
+printf "\\n"
+
+
+printf "Checking CMAKE installation...\\n"
```
spoonincode: Is there a reason we are building cmake on platforms we haven't needed to before?

CMakeLists.txt
NorseGaud: Previous checkouts didn't have the ~/opt symlink

scripts/eosio_build_darwin.sh
NorseGaud: I agree, but the problem requires the --force flag to fix. I wish we didn't need the workaround at all really.

CMakeLists.txt
spoonincode: Does it work though? You've tried it? LLVM_DIR is a cached variable which typically means you can't change it through a simple `set()`

scripts/eosio_build_fedora.sh
NorseGaud: Versions of cmake for fedora differ greatly. Consistency and copy and paste ability between build files is why I chose to do this; we shouldn't need to worry about keep track of differences. Also, I had a lot of problems which are fixed by using the same cmake version we use for all other distros.

spoonincode: I commented on the fedora one but we're also now building cmake on ubuntu & macos -- both platforms we did not do this in the past either.

As it is we only support 3.5 and newer. Can you elaborate on what the issues are? We know 3.5, 3.10, 3.12, and 3.13 work today on the develop branch. I do not think we should be supplanting software that the OS or brew provides unless we need to.

scripts/eosio_build_darwin.sh
spoonincode: What is the problem? Actually where is gettext used? AFAIK not for nodeos/cleos/keosd

CMakeLists.txt
NorseGaud: Great question: My goal was to have mac as close as possible to how Linux; We can avoid having if/else in the CMake files looking for llvm by using a symlink from ~/opt/llvm. I also believe that using llvm instead of wasm for the folder name is easier for developers to understand. Especially now that we don't install it using experimental arguments on cmake.

scripts/eosio_build_darwin.sh
NorseGaud: Looks like gettext is a dependency of libidn2, which is required by wget. I believe I removed the majority of reliance on wget to download source packages (using Curl instead), so maybe we can remove it?

scripts/eosio_build_fedora.sh
NorseGaud: Unfortunately the cmake which brew provides is not guaranteed to stay at 3.13 (which is the current version). There is also no cmake@3.10 or anything like that. I've had the brew folks upgrade a version of something we were dependent on at my last job and refuse to create the @ older version formula or accept the one I submitted for them. I'm looking to avoid our reliance on brew as much as possible. Not only does this make it easier to maintain long term, but we can be assured that things won't randomly stop working due to a homebrew formula upgrade we aren't notified of.

scripts/eosio_build_darwin.sh
spoonincode: Even if it is a dependency of wget, it doesn't need to be force linked to use wget.

spoonincode: Looks like gettext is needed for cleos, actually. Surprised me

PR #6653, opened 2019-01-23
Refactor  filter_include function of mongo_db_plugin

xinyijun:
## Change Description
Refactor filter_include function of mongo_db_plugin(remove redundancy code):
filter.match( receiver, act_name, 0 ) return true if we leave actor field in config.ini as blank, but if we do this, filter.match( receiver, act_name, a.actor ) will also return true for any actor.


heifner:
Can you explain more why you would like this change? It looks to me like this does the same thing only slower.

xinyijun:
It removes redundancy and make code more clear and concise. It does do the same thing, but for performance, the final assembly code should be similar, it shouldn't cause slower.

PR #6678, opened 2019-01-29
Enable websocketpp part threading

bspark8:
## Change Description

Nodeos is a CPU intensive task and previously one main thread handled all of the network, transaction processing, block creation until the 1.6 version has separated transaction verification jobs through producer and chain threads.
This PR allows separation of the web socket server(websocketpp) thread as well.
A simple evaluation showed that performance can be increased in case of nodeos which use http plugin.

Environment
- Nodes: 2 BPs, 8 Clients
- HW: Intel(R) Xeon(R) Silver 4114 CPU @ 2.20GHz x2 (10 core), 256GB
- Options: net-plugin, 10 producer threads, 10 chain threads, wabt
- Benchmark: cleos based token transfer dapp, each clients 8 works (https://github.com/CryptoLions/TX-test-sender)

Current
- About 935 TPS
- the main thread CPU usage: 98%

After patch
- About 1178 TPS (26% performance improvement)
- the main thread / websocketpp thread CPU usage: 92% / 52%

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

PR #6687, opened 2019-01-30
Http response thread pool

heifner:
## Change Description

- Thanks go to @bspark8 #6678 for testing and verifying the improvement of processing/sending http request on a different thread.
- This PR adds a thread pool to the http_plugin used for processing/sending http requests.
- Remove auto load of http_plugin in nodeos main.cpp since not all nodeos need the http_plugin and the extra threads it spawns.
- Application logic is still processed on the application thread.
- See #6678 for additional info

## Consensus Changes

None

## API Changes

None

## Documentation Additions

- http_plugin has new option:
  --http-threads arg (=2)               Number of worker threads in http thread pool

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -123,6 +123,7 @@ namespace eosio {
    using websocket_local_server_type = websocketpp::server<detail::asio_local_with_stub_log>;
    using websocket_server_tls_type =  websocketpp::server<detail::asio_with_stub_log<websocketpp::transport::asio::tls_socket::endpoint>>;
    using ssl_context_ptr =  websocketpp::lib::shared_ptr<websocketpp::lib::asio::ssl::context>;
+   using io_work_t = boost::asio::executor_work_guard<boost::asio::io_context::executor_type>;
 
    static bool verbose_http_errors = false;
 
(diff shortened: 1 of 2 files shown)


heifner:
I missed that you are passing the new io_context to init. Thanks for pointing that out.

heifner:
@bspark8 I've added a new io_context for the websocketpp server as you suggested.

bspark8:
@heifner Thank you for additional threading for the part that receives the HTTP requests.
In addition, I did the same test that I did in the previous PR(#6678) in the current patch. Below is the result.

Current patch
- About 1137 TPS (22% performance improvement)
- Each thread CPU usage
main: 88%
receiving HTTP requests: 52%
responding to the HTTP requests: 14%

plugins/http_plugin/http_plugin.cpp
```diff
@@ -138,6 +139,13 @@ namespace eosio {
 
          websocket_server_type    server;
 
+         uint16_t                                    thread_pool_size = 2;
```
brianjohnson5972: So we still want to default to 2 threads, even though we can have 1?

heifner: Since we now only load http_plugin if we need it, sure.

PR #6691, opened 2019-02-02
Monitor wallet lock file and shut down if it's removed.

### LINKED ISSUE #5998: New keosd process is created even though there is an existing keosd process (develop branch)
I was investigating issue #4064, where `cleos wallet create` fails if "eosio-wallet" directory doesn't exist and keosd is already running.

This issue still persists in the master branch but has disappeared in the develop branch. 

However, I noticce the following thing in develop branch:
1. I call `cleos wallet create`, a keosd process is spawned
2. I delete the "~/eosio-wallet" folder
3. I call `cleos wallet create` again, another keosd process is spawned. So I have two keosd processes now running, which I think should be prevented since I already have a keosd process running?

Although it's not harmful or critical, I think it will be tidier if it can be cleaned up.

jgiszczak:
## Change Description

Monitor wallet lock file and shut down keosd if it's removed. Resolves #5998.

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

### DIFF
--- a/plugins/wallet_plugin/wallet_manager.cpp
+++ b/plugins/wallet_plugin/wallet_manager.cpp
@@ -2,6 +2,7 @@
  *  @file
  *  @copyright defined in eos/LICENSE
  */
+#include <appbase/application.hpp>
 #include <eosio/wallet_plugin/wallet_manager.hpp>
 #include <eosio/wallet_plugin/wallet.hpp>
 #include <eosio/wallet_plugin/se_wallet.hpp>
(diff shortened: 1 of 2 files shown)


plugins/wallet_plugin/wallet_manager.cpp
```diff
@@ -275,6 +276,22 @@ void wallet_manager::own_and_use_wallet(const string& name, std::unique_ptr<wall
    wallets.emplace(name, std::move(wallet));
 }
 
+void wallet_manager::start_lock_watch(std::shared_ptr<boost::asio::deadline_timer> t)
+{
+   t->async_wait([t, this](const boost::system::error_code& /*ec*/)
+   {
+      struct stat statbuf;
+      int rc = stat(lock_path.string().c_str(), &statbuf);
```
spoonincode: Can this be made in to a boost filesystem call instead?

spoonincode:
Polling is pretty lousy but I realize there is no quick way to integrate inotify/kevent/whatever-win32-uses

jgiszczak:
I did try inotify first and unfortunately it's completely unreliable.  On Ubuntu with encrypted home directories, encryptfs completely mangles inotify events, to the point of unuseability.  Most events are entirely lost.  Except when they're not.  Then they are again.  With both false positives and false negatives rampant, I gave up on it entirely.

plugins/wallet_plugin/wallet_manager.cpp
jgiszczak: Converted to boost::filesystem.

PR #6715, opened 2019-02-08
V2 build scripts for ci

NorseGaud:
Updates since review:
1. build/install/uninstall scripts moved into scripts directory (out of root) + paths in scripts changed to accommodate dependency installation scripts for each distro
2. Symlinking in install script has been removed. Output after install has been modified

### Build Script changes:
1. Variables used throughout are now located in eosio_build.sh (D.R.Y)
2. BOOST, Mongo+mongolibs, CMAKE, and others are now installed into and using ~/lib,bin,etc,data (Eliminates need for root as much as possible)
3. MongoDB launch command uses ~/ paths for configs, logs, etc (CI/CD pipeline as well)
4. Installation from source is closer to what the developer portal suggests
5. Ubuntu build script MongoDB installation accounts for 16.04 or 18.04 versions (18 gets v4, vs v3; v3 doesn't exist for 18)
6. Wrapped apt/yum/brew update with warning
8. Better cleanup script to handle old installs or new
9. Fix for gettext being keg-only

### CMakeModules changes:
1. LLVM/WASM path is now ~/opt/llvm + FindWasm using ~/opt/llvm for both darwin and linux

### Test changes
1. tests/nodeos_run_test.py -> Setting currentBlockNum later in code so that a slow machine that generates too many blocks won't fail the tests

### Notes:
- We can now copy and paste almost everything between OS build scripts. (Time savings for maintenance)
- We have consistency between Linux and Mac paths.
- Build script are used to create images, "shifting left" the responsibility of maintaining them to DevOps, instead of both Dockerfiles AND build scripts.
- Output of build scripts are much cleaner and easier to follow.

### VM Image Speed Improvements:
- mojave build from 8m to 5m
- centos build from 15m to 8m


CMakeLists.txt
```diff
@@ -19,12 +19,11 @@ if (UNIX)
       execute_process(COMMAND xcrun --show-sdk-path
                       OUTPUT_VARIABLE CMAKE_OSX_SYSROOT
                       OUTPUT_STRIP_TRAILING_WHITESPACE)
-      if (LLVM_DIR STREQUAL "" OR NOT LLVM_DIR)
-         set(LLVM_DIR "/usr/local/Cellar/llvm@4/4.0.1/lib/cmake/llvm")
-      endif()
    endif()
 endif()
 
+set(LLVM_DIR "$ENV{HOME}/opt/llvm/lib/cmake/llvm")
```
spoonincode: Does this break users who do not use the script? If a path is unique to how our build script sets things up it should probably be set from the build script that sets things up

scripts/eosio_build_darwin.sh
```diff
@@ -1,406 +1,311 @@
...
+			IFS="$OIFS"
+		;;
+		[Nn]* ) echo "User aborting installation of required dependencies, Exiting now."; exit;;
+		* ) echo "Please type 'y' for yes or 'n' for no."; exit;;
+	esac
+else
+	printf "\\n - No required Home Brew dependencies to install.\\n"
+fi
 
+printf "Checking CMAKE installation...\\n"
```
spoonincode: Why is cmake being checked for twice in this script?

scripts/eosio_build_darwin.sh
```diff
@@ -1,406 +1,311 @@
...
-					BREW=$( command -v brew )
-				fi
-				break;;
-				[Nn]* ) echo "User aborted homebrew installation. Exiting now.";
-						exit 1;;
-				* ) echo "Please enter 1 for yes or 2 for no.";;
-			esac
-		done
-	fi
+printf "Checking CMAKE installation...\\n"
```
spoonincode: No change in my opinion from last PR. Don't believe we need to build cmake from source on platforms that give us a new enough cmake

spoonincode:
>* We can now copy and paste almost everything between OS build scripts. (Time savings for maintenance)

If parts to the scripts are common we should place them in a function.

scripts/eosio_build_darwin.sh
NorseGaud: Whoops, must have copied and pasted improperly

scripts/eosio_build_darwin.sh
NorseGaud: We could just remove this entirely, right? It was only for Darwin originally, and never set for Linux, but from what I can tell if LLVM_DIR isn't set it looks under what I've specified in the ENV vars we have users set.

CMakeLists.txt
NorseGaud: We could just remove this entirely, right? It was only for Darwin originally, and never set for Linux, but from what I can tell if LLVM_DIR isn't set it looks under what I've specified in the ENV vars we have users set.

spoonincode: Yeah it looks like you can remove it since you export `LLVM_DIR` from the script it will be picked up that way

scripts/eosio_build_ubuntu.sh
```diff
@@ -1,430 +1,262 @@
...
+
+cd ..
+printf "\\n"
 
-	function print_instructions()
-	{
-		printf '\n\texport PATH=${HOME}/opt/mongodb/bin:$PATH\n'
-		printf "\\t%s -f %s &\\n" "$( command -v mongod )" "${MONGOD_CONF}"
-		printf "\\tcd %s; make test\\n\\n" "${BUILD_DIR}"
+function print_instructions()
```
spoonincode: IMO asking users to do these instructions is a support nightmare. I don't know about the mongo stuff, but for cmake, cmake options, and ld_library_path there are probably better ways of going about this

CMakeLists.txt
NorseGaud: Ah, actually, we need to keep it. Most of the folks here don't use the build scripts to build, so they won't get the ENV I set. I'll expand this for BOOST_ROOT as well and interpolate in the EOSTester files

spoonincode: But does it matter? At least on macos and ubuntu I don't think you need anything at all here right?

scripts/eosio_build_ubuntu.sh
NorseGaud: To make mac and linux more uniform (for supportability) I had to use $HOME for the build/link locations. Unfortunately, this required manipulating ENV (LD/CPATH, etc). This sets us up to support situations like the one with fedora27 where the include directory is non-standard. I'm open to suggestions of course.

spoonincode: If a user is using the build script, the build script is the sole method by which cmake is called and cmake is configured. There shouldn't be any need to have things like CMAKE_MODULE_PATH or CPATH in the environment or have cmake in path because any sort of configuration of that nature can be done entirely in the script.

What is LD_LIBRARY_PATH used/needed for? It sounds like we're building an executable that pulls in shared libraries that we also build but I'm not sure what that would be, mongo maybe?

NorseGaud: It seemed as if `make test` after build was failing without having the llvm/lib stuff loaded in the user's environment. This was due to the $HOME/opt/llvm path.

spoonincode: What was the error? I don't believe anything in ~/opt/llvm would be used post build as long as everything was still being linked statically

scripts/eosio_build_darwin.sh
NorseGaud: Totally understand the concern. Unfortunately brew's cmake (or any brew packages) should not be relied upon for build/testing, as it is subject to change without warning and differ slightly (arguments they use to install) from the linux packages. In order to ensure confidence, reliability, and consistency between distros, we should be source installing as much as we can. 

My big concerns are around: What if brew upgrades the version of cmake that differs from the linux versions? If that breaks the mac build process for users of the scripts, and we don't notice it, do we just rely on them to notify us (create a bad user experience)?

It's possible that we can just continue to use the homebrew-eosio repo for our own tap, but then we (likely the blockchain team) have to maintain the formulas, which costs money and time for engineers/developers. Through uniformity (building cmake from source for all distros), we can eliminate many types of waste for everyone involved.

This also applies to llvm and other packages we're not source installing, but I'm told we're moving away from needing them in the future anyway.

spoonincode: This is a slippery slope. Are you going to start compiling openssl and gmp too? Are you going to lock the build down to a specific Xcode version? The line has to be drawn somewhere and historically it has been drawn at the clear point where version differences impact nodeos data file compatibility.

We’re not building an embedded or proprietary software system, we’re building open source software that uses open source tools and components. Platform and tool diversity is a part of life and ignoring it reflects poorly — much more complex open source software doesn’t need to lock it self down to a specific cmake version for example.

CMakeLists.txt
NorseGaud: I'm not sure actually. Bucky was saying they should both be there. I've wrapped them with if statements to avoid setting them if they already exist.

scripts/eosio_build_darwin.sh
NorseGaud: Do you think the current if statement to check if cmake already exists solves most of your concerns?
```CMAKE=$(command -v cmake 2>/dev/null)
if [ -z $CMAKE ]; then```

File: CMakeLists.txt (hunk 1 above)
Comment (spoonincode) [reply CMakeLists.txt]: AFAIK llvm4 on brew & ubuntu should be installed in such a way that this isn't needed since the cmake module is installed in a system location. haven't tried myself, but we should be more sure of our changes

File: scripts/eosio_build_ubuntu.sh (hunk 4 above)
Comment (NorseGaud) [reply scripts/eosio_build_ubuntu.sh]: I'm pulling the llvm/lib right now and running some tests. I'll let you know how it goes

File: CMakeLists.txt (hunk 1 above)
Comment (spoonincode) [reply CMakeLists.txt]: ah okay for macos we do need to specify where LLVM is, it won't get added to system cmake directory

File: scripts/eosio_build_darwin.sh [hunk 5]
```diff
@@ -1,406 +1,304 @@
...
+
+printf "\\n"
+
+
+# We install llvm into /usr/local/opt using brew install llvm@4
+printf "Checking LLVM 4 support...\\n"
+if [ ! -d $LLVM_ROOT ]; then
+	ln -s /usr/local/opt/llvm@4 $LLVM_ROOT \
+	|| exit 1
+	printf " - LLVM (WASM compiler) successfully linked from /usr/local/opt/llvm@4 to ${LLVM_ROOT}\\n"
```
spoonincode: Need to remove "WASM compiler" this isn't a wasm compiler any more

spoonincode:
Note #6733 adds some new dependencies, we will need to somehow rectify that in whatever order the PRs are accepted

scripts/eosio_build_darwin.sh
```diff
@@ -1,406 +1,304 @@
...
-		done
-	fi
+printf "Checking CMAKE installation...\\n"
+CMAKE=$(command -v cmake 2>/dev/null)
+if [ -z $CMAKE ]; then
+	printf "Installing CMAKE...\\n"
+	curl -LO https://cmake.org/files/v$CMAKE_VERSION_MAJOR.$CMAKE_VERSION_MINOR/cmake-$CMAKE_VERSION.tar.gz \
+	&& tar xf cmake-$CMAKE_VERSION.tar.gz \
+	&& cd cmake-$CMAKE_VERSION \
+	&& ./bootstrap --prefix=$HOME \y
```
spoonincode: stray `y`

NorseGaud: Good catch!

scripts/eosio_build_darwin.sh
NorseGaud: Believe this was addressed after discussions about homebrew in Dawn channel today.

scripts/eosio_build_ubuntu.sh
```diff
@@ -1,430 +1,269 @@
...
+		continue
+	fi
+done
+if [ "${COUNT}" -gt 1 ]; then
+	printf "\\nThe following dependencies are required to install EOSIO:\\n"
+	printf "${DISPLAY}\\n\\n" 
+	if [ $1 == 0 ]; then read -p "Do you wish to install these packages? (y/n) " answer; fi
+	case ${answer} in
+		1 | [Yy]* )
+			if ! apt-get -y install ${DEP}; then
```
spoonincode: The removal of sudo here means that as a normal user (i.e. not root) I can't just `./eosio_build.sh` if I'm missing a dependency. I guarantee this indirectly encourages users to run the script as root (bad).

If we,
* Don't want sudo in scripts (good practice IMO but I'm not staunchly against it)
* Don't want users running and building as root (never good to encourage this, very dangerous)

Then I think we might need to change this to instructions instead, like print out something:
> You need to install some additional dependencies to build eosio, run sudo apt-get install blah blah

I'm also fine with adding back the sudo if we prefer that approach. But as-is this is troublesome.

NorseGaud: Sudo added back

scripts/eosio_build_ubuntu.sh
```diff
@@ -1,430 +1,259 @@
-	OS_VER=$( grep VERSION_ID /etc/os-release | cut -d'=' -f2 | sed 's/[^0-9\.]//gI' )
-	OS_MAJ=$(echo "${OS_VER}" | cut -d'.' -f1)
-	OS_MIN=$(echo "${OS_VER}" | cut -d'.' -f2)
+if [ $1 == 1 ]; then answer=1; fi # NONINTERACTIVE
```
spoonincode: How is this designed to work? Won't arbitrary options to the build script cause noninteractive mode to be enabled?

NorseGaud: eosio_build.sh has `. "$FILE" $NONINTERACTIVE # Execute OS specific build file` and then above that earlier in the script it will default to NONINTTERACTIVE=0 if arg0 is not "noninteractive". We can always change this (using --noninteractive,etc), but it's useful for automated pipeline/image creations purposes.

spoonincode: okay I see what you did there. We should probably integrate the option into the rest of the getopts, and no need to pass it as a parameter since the eosio_build_* scripts are sourced. Can hit this after merge.

NorseGaud: Sounds good

PR #6716, opened 2019-02-08
Unit Tests Now Run In Parallel

kj4ezj:
## Change Description
unit_test_wavm was our longest-running, non-LRT test at 228 seconds. unit_test_wavm is a set of test suites and, to enable these individual suites to be run in parallel, has been broken down by test suite now. unit_test_wabt is actually the same set of test cases called with one different command-line parameter, so that has been parallelized as well. These tests now run over twice as fast in buildkite and almost six times faster on my local machine.

## Consensus Changes
None.

## API Changes
None.

## Documentation Additions
None?

### DIFF
--- a/unittests/CMakeLists.txt
+++ b/unittests/CMakeLists.txt
@@ -55,7 +55,7 @@ target_include_directories( unit_test PUBLIC
 
 ### MARK TEST SUITES FOR EXECUTION ###
 foreach(TEST_SUITE ${UNIT_TESTS}) # create an independent target for each test suite
-  execute_process(COMMAND bash -c "grep -E 'BOOST_AUTO_TEST_SUITE\\s*[(]' ${TEST_SUITE} | cut -d ')' -f 1 | cut -d '(' -f 2" OUTPUT_VARIABLE SUITE_NAME OUTPUT_STRIP_TRAILING_WHITESPACE) # get the test suite name from the *.cpp file
+  execute_process(COMMAND bash -c "grep -E 'BOOST_AUTO_TEST_SUITE\\s*[(]' ${TEST_SUITE} | grep -v '//' | cut -d ')' -f 1 | cut -d '(' -f 2" OUTPUT_VARIABLE SUITE_NAME OUTPUT_STRIP_TRAILING_WHITESPACE) # get the test suite name from the *.cpp file
   if (NOT "" STREQUAL "${SUITE_NAME}") # ignore empty lines
     execute_process(COMMAND bash -c "echo ${SUITE_NAME} | sed -e 's/s$//' | sed -e 's/_test$//'" OUTPUT_VARIABLE TRIMMED_SUITE_NAME OUTPUT_STRIP_TRAILING_WHITESPACE) # trim "_test" or "_tests" from the end of ${SUITE_NAME}
     # to run unit_test with all log from blockchain displayed, put "--verbose" after "--", i.e. "unit_test -- --verbose"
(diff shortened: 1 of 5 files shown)


unittests/CMakeLists.txt
```diff
@@ -55,59 +53,60 @@ target_include_directories( unit_test PUBLIC
...
- -t \!wasm_tests/weighted_cpu_limit_tests
- --report_level=detailed --color_output -- --wabt)
-
+### MARK TEST SUITES FOR EXECUTION ###
+foreach(TEST_SUITE ${UNIT_TESTS}) # create an independent target for each test suite
+  execute_process(COMMAND bash -c "grep 'BOOST_AUTO_TEST_SUITE(' ${TEST_SUITE} | cut -d ')' -f 1 | cut -d '(' -f 2" OUTPUT_VARIABLE SUITE_NAME OUTPUT_STRIP_TRAILING_WHITESPACE) # get the test suite name from the *.cpp file
+  if (NOT "" STREQUAL "${SUITE_NAME}") # ignore empty lines
+    execute_process(COMMAND bash -c "echo ${SUITE_NAME} | sed -e 's/s$//' | sed -e 's/_test$//'" OUTPUT_VARIABLE TRIMMED_SUITE_NAME OUTPUT_STRIP_TRAILING_WHITESPACE) # trim "_test" or "_tests" from the end of ${SUITE_NAME}
+    # to run unit_test with all log from blockchain displayed, put "--verbose" after "--", i.e. "unit_test -- --verbose"
+    if("wasm" STREQUAL "${TRIMMED_SUITE_NAME}") # edge case: exclude some test cases from suite
```
spoonincode: Instead of special casing wasm_tests should probably turn off wasm_tests/weighted_cpu_limit_tests by default

unittests/CMakeLists.txt
```diff
@@ -55,59 +53,60 @@ target_include_directories( unit_test PUBLIC
...
-add_test(NAME unit_test_wavm COMMAND unit_test
- -t \!wasm_tests/weighted_cpu_limit_tests
- --report_level=detailed --color_output --catch_system_errors=no -- --wavm)
- add_test(NAME unit_test_wabt COMMAND unit_test
- -t \!wasm_tests/weighted_cpu_limit_tests
- --report_level=detailed --color_output -- --wabt)
-
+### MARK TEST SUITES FOR EXECUTION ###
+foreach(TEST_SUITE ${UNIT_TESTS}) # create an independent target for each test suite
+  execute_process(COMMAND bash -c "grep 'BOOST_AUTO_TEST_SUITE(' ${TEST_SUITE} | cut -d ')' -f 1 | cut -d '(' -f 2" OUTPUT_VARIABLE SUITE_NAME OUTPUT_STRIP_TRAILING_WHITESPACE) # get the test suite name from the *.cpp file
```
spoonincode: This seems super fragile to me. If someone types `BOOST_AUTO_TEST_SUITE (` instead of `BOOST_AUTO_TEST_SUITE(` the tests will be silently dropped?

unittests/CMakeLists.txt
kj4ezj: I removed this special case and deleted the problematic test case (weighted_cpu_limit_tests) from our source code. I confirmed with Bucky and Areg that this test is not reliable and, since unreliable tests are useless, should be deleted. It has been excluded using the add_testI() command for a long time, and I just copied that behavior. I have written about this in my notes in case we want to rewrite a test for this in the future.

unittests/CMakeLists.txt
kj4ezj: I have written a regular expression that ignores whitespace. Great idea! I hadn't even considered that. Tested locally.

PR #6720, opened 2019-02-11
Develop

learnforpractice:
## Change Description

 1. vm api for EOS
 2. Accelerating action execution of token transfer


heifner:
Thank you for your contribution, however, this is the opposite direction eos.io is headed. We are working to move more capabilities into WASM/contracts not move contract behavior into native code.

If you look far enough back in the history of the project you will find that transfer was at one point a native action. It was moved to be non-native so it can be upgraded without requiring a hard-fork.

Also note that eosiolib has moved to https://github.com/EOSIO/eosio.cdt/tree/master/libraries/eosiolib

negtise:
Well, I think it’s a compatible solution. That means we can reuse source code of smart contract. And I don’t understand what hard-fork will that cause since the code behavior is the same between the eosin.token code compile to native code and compile to wasm bytecode. 

If you are talking about platform compatibilities, that’s a problem but not for eosio.token code. 

We should compile compatible code to native code to improve performance.

Having token transfer execution faster about 3x is a good news for Eos users, don’t you think?

learnforpractice:
Well, I think it’s a compatible solution. That means we can reuse source code of smart contract. And I don’t understand what hard-fork will that cause since the code behavior is the same between the eosin.token code compile to native code and compile to wasm bytecode. 

If you are talking about platform compatibilities, that’s a problem but not for eosio.token code. 

We should compile compatible code to native code to improve performance.

Having token transfer execution faster about 3x is a good news for Eos users, don’t you think?

heifner:
I agree this can be done without requiring a hard-fork. However, it is a bit dangerous as the wasm and native have to be kept in-sync otherwise it is a fork.

WASM also protects the execution from various attacks that are possible in native code. e.g. memory boundary protection.

learnforpractice:
I agree with you too.

Keeping code in wasm and native in sync do require good programming experience and additional testing work to ensure the code behavior is the same.

But performance is also an important thing for Eos.

PR #6721, opened 2019-02-11
Fix boost 1.67 brew install - 1.6.x

spoonincode:
## Change Description
The latest homebrew code balks at something in the old 1.67 package file. Fix the package file and store it locally in our repo for now.

We try and pin the boost version because boost upgrades invalidate nodeos data files.


carlosvaztec:
Worked fine here. Thanks.

NorseGaud:
This will be fixed in https://github.com/EOSIO/eos/pull/6715

PR #6724, opened 2019-02-11
Set proper directory for baked in macOS LLVM_DIR - 1.6.x

spoonincode:
## Change Description
Some users (including myself) were seeing llvm@4 unpacked to 4.0.1_1 instead of 4.0.1. Stuff unpacked to the Cellar directory appears to be a kind of implementation detail — /usr/local/opt is the proper place to reference here.


NorseGaud:
This will be fixed in https://github.com/EOSIO/eos/pull/6715

Issue #6696, opened 2019-02-04
Optimize net_plugin


heifner:
- [x] Unpack only block header to determine if block needed before unpacking the rest
- [x] Use thread_pool for socket read and unpacking of message

PR #6725, opened 2019-02-11
Net plugin opt

heifner:
## Change Description

- Setup a thread pool for net_plugin sockets
- Currently only used for server io_context

## Documentation Additions
--net-threads arg (=1)                Number of worker threads in net_plugin 
                                        thread pool


heifner:
Resolves part of #6696

PR #6728, opened 2019-02-12
Not building `test-contracts` by default

johndebord:
## Change Description

By default, the contracts in directory `eos/unittests/test-contracts/` shall not be built, but have their respective `.wasm`'s and `.abi`'s copied over to the build directory. This is to prevent CLion users and the like from running into headaches when getting up and running. But, if a user so chooses, the flag `EOSIO_COMPILE_TEST_CONTRACTS` can be explicitly set to tell CMake to build the contracts using the machine's currently installed [eosio.cdt](https://github.com/EOSIO/eosio.cdt).

For example:
`~/eos/build$ cmake -DEOSIO_COMPILE_TEST_CONTRACTS=TRUE ..`

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A


unittests/test-contracts/test_ram_limit/CMakeLists.txt
```diff
@@ -1,5 +1,4 @@
-if( ${eosio.cdt_FOUND} )
-   message(STATUS "Not building test_ram_limit, read README.txt in eos/unittests/test-contracts/test_ram_limit")
```
arhag: Keep this message.

PR #6733, opened 2019-02-13
libyubihsm2 upgrade & YubiHSM 2.1 support

spoonincode:
## Change Description
Upgrade the yubihsm wallet to use libyubihsm2. This brings about two user facing features:

* Support for newly shipping YubiHSM 2.1 firmware
* Support for talking to a YubiHSM directly through USB instead of through the connector (use `yubihsm-url = yhusb://`)

libyubihsm2 is open source so now we build and statically link to the library internal to the eosio project. This makes using YubiHSM for users much easier as they don't have to chase down Yubico's installer and dump the .sos in the correct directory (correct can be tricky here). However, this also makes eosio have three new dependencies: libusb, libcurl, and pkg-config. All three of these are pretty standard fare so I don't feel too guilty adding them as dependencies.

CI is currently failing because of the inability to have libusb available in the test instance (I link to it dynamically since I consider it much like other system level packages like openssl)


scripts/eosio_build_darwin.sh
```diff
@@ -159,8 +159,11 @@
...
 						exit 1;
 					fi
-                                        if [[ "$DEP" == "llvm@4" ]]; then
-                                                "${BREW}" unlink ${DEP}
+					DEP=${DEP//libusb /}
+					if [[ "$DEP" == "llvm@4" ]]; then
+						"${BREW}" unlink ${DEP}
+					elif [[ -z "$DEP" ]]; then
+						true
 					elif ! "${BREW}" unlink ${DEP} && "${BREW}" link --force ${DEP}
```
spoonincode: This here is all sorts of WTFBBQ. The script seems written such that for any dependencies we're installing unlink them, _but if we fail to unlink force link them_. That just really boggles my mind. Also bizarre is that `DEP` appears to be a _list_ of items. So the `if [[ "$DEP" == "llvm@4" ]];` checks seems... not right. How could DEP ever be "llvm@4" when the line to change DEP is `DEP=$DEP"${brewname} "` -- it has a space after every DEP.

I need libusb to remain linked, which is its default, but that `unlink ${DEP}` burns me. So add some stuff here to remove libusb from the DEP list and skip the unlink/force link if DEP is then empty because of that removal.

scripts/eosio_build_amazon.sh
```diff
@@ -56,11 +56,12 @@
 	if [[ "${OS_NAME}" == "Amazon Linux AMI" ]]; then
 		DEP_ARRAY=( git gcc72.x86_64 gcc72-c++.x86_64 autoconf automake libtool make bzip2 \
 		bzip2-devel.x86_64 openssl-devel.x86_64 gmp-devel.x86_64 libstdc++72.x86_64 \
-		python27.x86_64 python36-devel.x86_64 libedit-devel.x86_64 doxygen.x86_64 graphviz.x86_64)
+		python27.x86_64 python36-devel.x86_64 libedit-devel.x86_64 doxygen.x86_64 graphviz.x86_64
+      libusb1-devel libcurl-devel)
 	else
```
heifner: tab vs spaces?

spoonincode: good call thanks, we're usually so consistent I failed to notice it here

heifner:
Looks like you need to rebase to develop.

NorseGaud:
mac builders -
cicdm012-b1-v2-1
cicdm011-b1-v2-1
mac testers - 
cicdm003-t1-v2-1
cicdm008-t1-v2-1

PR #6765, opened 2019-02-18
Accelerating with jit

learnforpractice:
## Change Description
Hi, folks
This pull use JIT to accelerating privileged smart contract execution.
I'm not expecting this pull request will merge to develop branch.
I think It may give some developers some inspiration, so I submitted this pull request.
Have fun, guys.


heifner:
Thanks for your PR. As you said, this will not be merged. 

- Web assembly is purposely used to protect nodeos from the executing contracts.
- transfer used to be part of native code, but it was moved like many other non-essential constructs to web assembly
- eosio can run with any contract for token transfer, this allows for various chains to have their own token implementation or no token contract at all if a chain doesn't need a token

learnforpractice:
The most important part of this pull request is accelerating privileged contract execution such as eosio.system and eosio.token, eosio.msig  with JIT,

that can save lots of CPU time, especially for normal Eos users.

tbfleming:
Token isn't privileged. Our goal over time is to increase flexibility and reduce maintenance burden by moving functionality out of nodeos and into contracts instead. We're also working to improve the performance of all contracts.

learnforpractice:
Please at least take at look at the following source code:

https://github.com/learnforpractice/eos-1/blob/1c5b00d7e86fbdd210a360af60d1b1dc9602c39f/libraries/vm/vm_manager/vm_manager.cpp#L138 <https://github.com/learnforpractice/eos-1/blob/1c5b00d7e86fbdd210a360af60d1b1dc9602c39f/libraries/vm/vm_manager/vm_manager.cpp#L138>
https://github.com/learnforpractice/eos-1/blob/1c5b00d7e86fbdd210a360af60d1b1dc9602c39f/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L261 <https://github.com/learnforpractice/eos-1/blob/1c5b00d7e86fbdd210a360af60d1b1dc9602c39f/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L261>

If there’s a better solution, I’d be happy to hear that.

tbfleming:
We don’t recommend wavm on prducer and api nodes because of the compilation delay. We also don’t want to treat contracts differently from each other, except where absolutely necessary. Our optimization plans are a long-term effort that I can’t do justice by trying to summarize here.

We’re moving the code in a certain direction; these patches take it in a different direction. Even patches which go in the same direction, but make significant changes to this area of code, are likely to conflict.

learnforpractice:
This pull request already has a solution to solve the compilation delay. It’s a solution that can take effect immediately, and yes it also takes a lot of changes.
Block.one has a fantastic team, and obviously Eos has its own plan on its track.
As I said, it’s only a pull request that in the hope that it will give some developers some thought. 

Nice talking with you.

PR #6776, opened 2019-02-19
Add warning to --state-history-endpoint

tbfleming:
## Change Description

The help message for `--state-history-endpoint` now includes a warning to only expose this port to private networks.

The default endpoint for the state history plugin has changed to `127.0.0.1:8080`.


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -512,7 +512,8 @@ void state_history_plugin::set_program_options(options_description& cli, options
    options("trace-history", bpo::bool_switch()->default_value(false), "enable trace history");
    options("chain-state-history", bpo::bool_switch()->default_value(false), "enable chain state history");
    options("state-history-endpoint", bpo::value<string>()->default_value("0.0.0.0:8080"),
-           "the endpoint upon which to listen for incoming connections");
+           "the endpoint upon which to listen for incoming connections. Caution: only expose this port to "
+           "your internal network.");
```
heifner: Should the provided default be `127.0.0.1:8080` ?

tbfleming: It depends on if users normally run the fillers on the same or different machines

heifner: Should you change the default value on line 514 to `127.0.0.1:8080` since that will be the normal case?

PR #6783, opened 2019-02-19
limit assert 1.6.x

heifner:
## Change Description

- Limit contract assert messages to 1024 charactors

## Consensus Changes

None

## API Changes

None

## Documentation Additions

None


heifner:
Already approved.

PR #6791, opened 2019-02-20
Reinitialize net plugin logger on HUP

### LINKED ISSUE #2432: logging SIGHUP doesn't reinitialize net_plugin's logger
There are a couple of obvious ways to fix this:

* Make logging reinitaliztion be a plugin lifecycle action. When logging config is updated, call out to each plugin. This would give net_plugin an opportunity to get the new net_plugin_impl logger.
* Make net_plugin just get the net_plugin_impl logger on each log event. This is no worse than what `ilog()` etc already do (but with the 'default' logger in that case). One argument against this is that the net plugin hits its debug logger harder than the default logger is typically used.

I suppose another option is to drop HUP support.

jgiszczak:
## Change Description

Reinitialize net_plugin's logger on receipt of SIGHUP.  Resolves #2432.

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

### DIFF
--- a/plugins/bnet_plugin/bnet_plugin.cpp
+++ b/plugins/bnet_plugin/bnet_plugin.cpp
@@ -1342,8 +1342,7 @@ namespace eosio {
    }
 
    void bnet_plugin::plugin_startup() {
-      if(fc::get_logger_map().find(logger_name) != fc::get_logger_map().end())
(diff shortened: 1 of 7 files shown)


spoonincode:
When that issue was originally written net_plugin was the only other logger. But be aware now there are a few others (in producer plugin & bnet plugin).

PR #6801, opened 2019-02-21
uninstaller for brew packages, doxygen/graphviz as brew packages

NorseGaud:


scripts/eosio_build_darwin_deps
```diff
@@ -6,7 +7,7 @@ GMP,-f,/usr/local/opt/gmp/include/gmpxx.h,gmp,https://ftp.gnu.org/gnu/gmp/gmp-6.
 llvm,-x,/usr/local/opt/llvm@4/bin/clang-4.0,llvm@4,http://releases.llvm.org/4.0.1/llvm-4.0.1.src.tar.xz
 python,-d,/usr/local/Cellar/python/3.7.2_1,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
 python@2,-d,/usr/local/Cellar/python@2/2.7.15_2,python@2,https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tgz
-doxygen,-x,/usr/local/bin/doxygen,eosio/eosio/doxygen --with-graphviz --build-from-source,http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.14.src.tar.gz
-cmake,-f,/usr/local/bin/cmake,cmake,https://github.com/Kitware/CMake/releases/download/v3.13.4/cmake-3.13.4.tar.gz
+doxygen,-f,/usr/local/bin/doxygen,doxygen,http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.14.src.tar.gz
+graphviz,-d,/usr/local/opt/graphviz,graphviz,https://fossies.org/linux/misc/graphviz-2.40.1.tar.gz
 libusb,-f,/usr/local/lib/libusb-1.0.0.dylib,libusb,https://github.com/libusb/libusb/releases/download/v1.0.22/libusb-1.0.22.tar.bz2
-pkgconfig,-x,/usr/local/bin/pkg-config,pkgconfig,https://pkg-config.freedesktop.org/releases/pkg-config-0.29.2.tar.gz
+pkgconfig,-x,/usr/local/bin/pkg-config,pkgconfig,https://pkg-config.freedesktop.org/releases/pkg-config-0.29.2.tar.gz
```
spoonincode: I'm reasonably certain you need the newline here for the parser to work and see pkgconfig. I think it doesn't show up as an issue now because pkgconfig is a dependency of something else earlier

scripts/dep_uninstaller.sh
```diff
@@ -10,6 +10,9 @@ if [ -d "/usr/local/include/eosio" ] || [ -d "$HOME/opt/eosio" ] || [[ $1 =~ for
                if [ $( uname ) == "Darwin" ]; then
                   # gettext and other brew packages are not modified as they can be dependencies for things other than eosio
                   brew uninstall llvm@4 --force
+                  brew uninstall doxygen --force
```
spoonincode: These additions seem somewhat against the original comment. Force removing could break other packages the user has installed right?

scripts/eosio_build_darwin_deps
NorseGaud: Good catch!

scripts/dep_uninstaller.sh
NorseGaud: I can remove: does that mean we should avoid it for llvm@4 as well?

spoonincode: the risk with llvm@4 seems very low since I assume very few things in brew depend on llvm@4. But the new additions are pretty easy to imagine being something the user needs directly or indirectly. We have no idea if eosio installed those or not, nor any idea if the user is using them for something since building eosio. I think we should remove at least the new ones for now

NorseGaud: Added noninteractive support to the uninstaller script as well prompts to uninstall them. I personally save a lot of time having them available while I'm testing changes locally on my mac.

PR #6832, opened 2019-02-25
Add descriptions to priority queue

heifner:
## Change Description

- Add description to the priority queue for debugging and profiling.
- The overhead when the description is not used should be minimal.
- Includes change to txn_text_gen_plugin to create two new accounts and to log private keys.
- Depends on https://github.com/EOSIO/appbase/pull/50

## Consensus Changes

None

## API Changes

None

## Documentation Additions

None

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -300,7 +300,7 @@ namespace eosio {
                if( handler_itr != url_handlers.end()) {
                   con->defer_http_response();
                   bytes_in_flight += body.size();
-                  app().post( appbase::priority::low,
+                  app().post( appbase::priority::low, resource,
                               [ioc = this->server_ioc, &bytes_in_flight = this->bytes_in_flight, handler_itr,
                                resource{std::move( resource )}, body{std::move( body )}, con]() {
                      try {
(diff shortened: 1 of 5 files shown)


spoonincode:
>Includes change to txn_text_gen_plugin to create two new accounts and to log private keys.

Why?

heifner:
> Why?

As just a simple way to have two accounts with the same key. Log the keys so it is easy to grab the keys for use by other manual tests.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -300,7 +300,7 @@ namespace eosio {
                if( handler_itr != url_handlers.end()) {
                   con->defer_http_response();
                   bytes_in_flight += body.size();
-                  app().post( appbase::priority::low,
+                  app().post( appbase::priority::low, resource,
                               [ioc = this->server_ioc, &bytes_in_flight = this->bytes_in_flight, handler_itr,
                                resource{std::move( resource )}, body{std::move( body )}, con]() {
```
heifner: `resource` is moved into lambda. Need to copy so `resource` is available for description.

heifner:
Going to close this for now. It was more for testing/debugging purposes.

PR #6833, opened 2019-02-25
GCC defect fixes.

jgiszczak:
## Change Description

This PR works around two gcc defects, one of which was fixed in gcc 7.1, the other of which was fixed in gcc 7.2.  Supported platforms building with gcc will encounter the problems.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66297
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=67274

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -206,7 +206,7 @@ struct controller_impl {
    SET_APP_HANDLER( eosio, eosio, canceldelay );
 
    fork_db.irreversible.connect( [&]( auto b ) {
-                                 on_irreversible(b);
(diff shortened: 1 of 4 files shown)


spoonincode:
It seems like an alternative fix to 67274 is to remove the `auto`. Is that a better stylistic approach?

jgiszczak:
It could obviate the need for a comment explaining the use of explicit `this` pointers but risks being 'streamlined' back to `auto` in some later commit.  If the risk is deemed low enough, I can make that change.

PR #6834, opened 2019-02-25
Remove Docker directory

NorseGaud:
This is used to start a conversation around the removal. Approval of ALL stakeholders will be required before it happens.

**Problem:** Docker files within the eosio repo add a maintenance burden for the Blockchain team.

**Goal:** Remove Docker folder from eosio, place it into separate repo.

**Reasons:**

1. Docker scripts and documentation is not tested with newer versions. Community members may try to use them, experiencing problems, and give up trying to develop with the software.
2. Software maintainers (the blockchain team) have to spend time making changes to docker files/docs instead of focusing solely on eosio. Mono repos can be harder to maintain.

* I have had discussions with Developer Relations about this and they seem to agree with the removal, but would like to maintain their own file using the nodeos binary.
* Discussions indicate they are using some of the files for internal pipelines/images. Need to collect a list and decide on the best course of action (likely move into eosio.dev or team specific repo).


NorseGaud:
Adding more stakeholders to this since it's now getting public attention.

JohnnyZhao:
So basically, BlockOne has no extra human power to just maintain some docker scripts?

Seems like we have to maintain it ourselves from now on.

PR #6836, opened 2019-02-25
Remove *.cmake from .gitignore

larryk85:
I'm not sure how this crept in.  This could've caused issues with lost changes to .cmake files under the CMakeModules folder if the developer wasn't careful.

## Change Description
Remove *.cmake from .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,8 @@
 *.s
 *.dot
 *.abi.hpp
+*.cmake
+!CMakeModules/*.cmake
 *.ninja
 \#*
 \.#*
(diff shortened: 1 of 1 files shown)


spoonincode:
This was there for people doing in-tree builds. The file contains a boat load of other paths for in-tree builds, should those be removed too?

heifner:
https://stackoverflow.com/questions/3637660/how-to-exclude-file-only-from-root-folder-in-git

Maybe something like:
*.cmake
!CMakeModules/*.cmake

PR #6839, opened 2019-02-25
new utility to truncate blocks.log

snstrand:
## Change Description
New program blockLogTrim will truncate blocks.log and blocks.index after a given block number.

To trouble shoot a problem at block 'n' can truncate blocks.log to before that, then run nodeos, then snapshot the state. Then can debug nodeos as it runs forward from the snapshot.

First command line arg should be the directory where blocks.log and blocks.index are located and second arg is -b flag to give last block kept. Example:

blockLogTrim ../eos_data/blocks -b33000111

### DIFF
--- a/programs/blockLogTrim/CMakeLists.txt
+++ b/programs/blockLogTrim/CMakeLists.txt
@@ -1,0 +1,7 @@
+cmake_minimum_required(VERSION 3.13)
+project(eos)
+
+set(CMAKE_CXX_STANDARD 14)
+
+add_executable(eos
+		blockLogTrim.cpp)
--- a/programs/blockLogTrim/blockLogTrim.cpp
+++ b/programs/blockLogTrim/blockLogTrim.cpp
@@ -1,0 +1,120 @@
+#include <iostream>
+#include <iomanip>
+#include <fstream>
+#include <unistd.h>         //truncate(char* path, u64 newLen)
+using namespace std;
+
+uint32_t reverseBytes(uint32_t x) {      //convert between big and little endian
+    uint32_t rval;
+    char* src= (char*)&x;
+    char* dst= (char*)&rval;
+    dst[0]= src[3];
+    dst[1]= src[2];
+    dst[2]= src[1];
+    dst[3]= src[0];
+    return rval;
+}
+
+char const* help=
+    "This utility truncates blocks.log and blocks.index after a given block number.\n"
+    "First arg is directory containing the two files then -b arg for last block to keep.\n"
+    "Example: blockLogTrim eos_data/blocks -b30111222\n";
+
+struct __attribute__((packed)) BlockHead {    //first 46 bytes of each block
+  uint32_t timestamp;
+  uint64_t prodname;
+  uint16_t confirmed;
+  uint32_t blknum;           //low 32 bits of previous blockid, is block number, is big endian
+  uint32_t previous;         //high 32 bits
+  uint64_t blkid[3];         //rest of 256 bit blockid for previous block
+} blkHead;
+
+char path[1024];
+
+int main(int argc, char** argv) {
+    uint32_t n;                     //last block to keep
+    if (argc < 3 || argv[2][0]!='-' || argv[2][1]!='b' || !(n = atoi(argv[2] + 2))) {
+        cout << help;
+        return 1;
+    }
+    strncpy(path, argv[1], 1010);     //leave enough room for "/blocks.index"
+    char *file = path + strlen(path);
+    if (file > path && (file[-1]!='/'))   //add '/' if needed
+        *file++ = '/';
+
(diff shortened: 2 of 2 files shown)


snstrand:
I have code for reconstructing blocks.index from blocks.log that is much faster than nodeos (in my test case 2 minutes vs. 90 minutes). I also have code for exploring blocks.log. This code should be added to blockLogTrim so that it becomes a more general utility for dealing with blocks.log and blocks.index.

heifner:
There is also this: https://github.com/EOSIO/eos/blob/master/programs/eosio-blocklog/main.cpp
Seems like it should all go into one util.

programs/blockLogTrim/blockLogTrim.cpp
```diff
@@ -0,0 +1,120 @@
+#include <iostream>
+#include <iomanip>
+#include <fstream>
+#include <unistd.h>         //truncate(char* path, u64 newLen)
+using namespace std;
+
+uint32_t reverseBytes(uint32_t x) {      //convert between big and little endian
```
heifner: Please follow existing code standards. `reverse_bytes`

programs/blockLogTrim/blockLogTrim.cpp
```diff
@@ -0,0 +1,120 @@
...
+    dst[3]= src[0];
+    return rval;
+}
+
+char const* help=
+    "This utility truncates blocks.log and blocks.index after a given block number.\n"
+    "First arg is directory containing the two files then -b arg for last block to keep.\n"
+    "Example: blockLogTrim eos_data/blocks -b30111222\n";
+
+struct __attribute__((packed)) BlockHead {    //first 46 bytes of each block
```
heifner: I would rather see `eosio/chain/block_header.hpp` `block_header` be used.

programs/blockLogTrim/blockLogTrim.cpp
```diff
@@ -0,0 +1,120 @@
...
+  uint32_t timestamp;
+  uint64_t prodname;
+  uint16_t confirmed;
+  uint32_t blknum;           //low 32 bits of previous blockid, is block number, is big endian
+  uint32_t previous;         //high 32 bits
+  uint64_t blkid[3];         //rest of 256 bit blockid for previous block
```
heifner: If this has to be used the attribute names should match those in block_header.hpp

programs/blockLogTrim/blockLogTrim.cpp
```diff
@@ -0,0 +1,120 @@
...
+} blkHead;
+
+char path[1024];
```
heifner: I see no reason for this `C` like code. Use `std::string`

programs/blockLogTrim/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.13)
+project(eos)
+
+set(CMAKE_CXX_STANDARD 14)
+
+add_executable(eos
```
spoonincode: Probably need a better name for the executable vs just `eos`

spoonincode:
Should this be included in binary releases?

Should this have unit tests?

jgiszczak:
Note that nodeos supports `--truncate-at-block` which will play back a block log to the indicated block number then stop.  It produces a blocks.log, blocks.index, and matching state to the given point, which may then be snapshoted.  This tool appears to be redundant for the stated purpose.

tbfleming:
@jgiszczak I requested this because I'm tired of waiting long periods of time on `--truncate-at-block`. Future functionality is at https://github.com/EOSIO/devdocs/wiki/nodeos-utils

programs/blockLogTrim/blockLogTrim.cpp
```diff
@@ -0,0 +1,120 @@
...
+    fin0.close();
+    if (bnum!=n) {
+        cout << "\nblocks.index does not agree with blocks.log so abort\n";
+        return 7;
+    } else {
+        cout << " as expected.\n";
+    }
+
+    strcpy(file,"blocks.log");
+    if (truncate(path,fpos1)) {
```
spoonincode: I don't think we should add POSIXisms unless we have to. Can boost filesystem's `resize_file` be used in place here?

programs/blockLogTrim/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+cmake_minimum_required(VERSION 3.13)
```
spoonincode: eosio project still targets cmake 3.5+, since this cmakelists.txt is part of a larger project lines 1-2 should be removed and line 4 probably should be removed.

jgiszczak:
#6884

heifner:
Subsumed by #6884

PR #6848, opened 2019-02-27
Validate producer signature providers

### LINKED ISSUE #6768: signature-provider is not required when specifying producer-name
Starting nodeos with a config.ini that contains "producer-name=" and but no "signature-provider" results nodeos starting, but it will never sign any blocks.

Should it be fatal error if no signature provider is given?

```
producer plugin:  plugin_startup() begin
producer_plugin.cpp:772       plugin_startup       ] Launching block production for 1 producers at 2019-02-18T17:00:03.808.
producer_plugin.cpp:784       plugin_startup       ] producer plugin:  plugin_startup() end
```

jgiszczak:
## Change Description

This change ensures that any configuration specifying `producer-name` also specifies a corresponding `signature-provider` (or `private-key`).  Any permission with a matching key is accepted.  No attempt is made to validate the permission weight configuration.  Resolves #6768.

## Consensus Changes

N/A

## API Changes

N/A

## Documentation Additions

N/A

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -672,17 +672,17 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
    }
 
    for (const auto& producer : my->_producers) {
+      bool signature_available = false;
       chain::controller& chain = my->chain_plug->chain();
       const auto& permissions = chain.db().get_index<permission_index,by_owner>();
       auto perm = permissions.lower_bound( boost::make_tuple( producer ) );
       while( perm != permissions.end() && perm->owner == producer ) {
-         if(perm->name == N(active)) {
-            for(const auto& keyweight : perm->auth.keys) {
-               EOS_ASSERT(is_producer_key(keyweight.key), plugin_config_exception, "Missing signature-provider for producer ${prod}", ("prod", producer));
-            }
+         for(const auto& keyweight : perm->auth.keys) {
+            signature_available |= is_producer_key(keyweight.key);
          }
          ++perm;
       }
+      EOS_ASSERT(signature_available, plugin_config_exception, "Missing signature-provider for producer ${prod}", ("prod", producer));
    }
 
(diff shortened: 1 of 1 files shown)


matthewdarwin:
Just reading the description (not the code), I'm not sure that we want to validate that the key is is valid.  Many BPs switch their active keys around to activate one computer or another.  If they do this, then it sounds like this test will fail.

My ask for this issue was to simply check there is a line in the config file... not to try to validate that the key is valid.

jgiszczak:
@matthewdarwin If checking for the presence of any signature-provider is all anyone wants, it's trivial to write a script to validate a config file at least that much.  Checking the return codes of a couple of grep commands is perfectly adequate.

It seems to me that checking for any random signature-provider at all is worse than useless.  It would lead a node operator to believe their node is correctly configured even if the signature-provider does not match the producer account's key.

Actually validating the signature-provider means a node operator can be assured their node is correctly configured the moment it launches.  Their node is certain not to miss producing any blocks if and when their producer account appears in the schedule.  This is especially useful for standby producers who may not produce any blocks for months at a time, until voting changes promote them to active.

My initial pass at this was indeed to see if any signature-provider is configured at all.  Unfortunately this doesn't work well as there is a default key, so there is always at least one signature-provider even if a node has not been explicitly configured with one.

In addition, there is an existing error message if a producer's turn arrives and no valid signature-provider is available.  It looks like this:
```
error 2019-02-28T18:29:45.500 thread-0  producer_plugin.cpp:1095      start_block          ] Not producing block because I don't have the private key for EOS...
```
I am not convinced the habit people have of changing keys to activate a different producer node is a best practice.  It requires the network to be functioning well enough to accept a transaction changing the key and to propagate that transaction to the specific node in question.  A failover node should not depend on receiving a block incorporating the transaction that will activate it.  That isn't failover.  That's hand-off.  The pause/resume functionality provided by producer_api_plugin exists for the purpose of facilitating failover.  If something about that functionality is inadequate, that's what should be fixed.

jgiszczak:
Withdrawn.  See related issue.

PR #6852, opened 2019-02-28
Fix testee tester brew

NorseGaud:
Upgrades to python, which change the version minor or patch, cause the build script to think it's missing (we look for the specific version folder existence).

I've modified the build script to support wildcard matching for folders.

### DIFF
--- a/scripts/eosio_build_darwin_deps
+++ b/scripts/eosio_build_darwin_deps
@@ -5,8 +5,8 @@ OpenSSL,-f,/usr/local/opt/openssl/lib/libssl.a,openssl,https://www.openssl.org/s
 wget,-x,/usr/local/bin/wget,wget,https://ftp.gnu.org/gnu/wget/wget-1.19.2.tar.gz
 GMP,-f,/usr/local/opt/gmp/include/gmpxx.h,gmp,https://ftp.gnu.org/gnu/gmp/gmp-6.1.2.tar.bz2
 llvm,-d,/usr/local/opt/llvm@4,llvm@4,http://releases.llvm.org/4.0.1/llvm-4.0.1.src.tar.xz
-python,-d,/usr/local/Cellar/python/3\*,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
-python@2,-d,/usr/local/Cellar/python@2/2\*,python@2,https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tgz
+python,-d,/usr/local/Cellar/python,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
+python@2,-d,/usr/local/Cellar/python@2,python@2,https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tgz
 doxygen,-f,/usr/local/bin/doxygen,doxygen,http://ftp.stack.nl/pub/users/dimitri/doxygen-1.8.14.src.tar.gz
 graphviz,-d,/usr/local/opt/graphviz,graphviz,https://fossies.org/linux/misc/graphviz-2.40.1.tar.gz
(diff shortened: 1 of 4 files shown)


scripts/eosio_build_darwin_deps
```diff
@@ -4,9 +4,9 @@ Libtool,-x,/usr/local/bin/glibtool,libtool,http://gnu.askapache.com/libtool/libt
 OpenSSL,-f,/usr/local/opt/openssl/lib/libssl.a,openssl,https://www.openssl.org/source/openssl-1.0.2n.tar.gz
 wget,-x,/usr/local/bin/wget,wget,https://ftp.gnu.org/gnu/wget/wget-1.19.2.tar.gz
 GMP,-f,/usr/local/opt/gmp/include/gmpxx.h,gmp,https://ftp.gnu.org/gnu/gmp/gmp-6.1.2.tar.bz2
-llvm,-x,/usr/local/opt/llvm@4/bin/clang-4.0,llvm@4,http://releases.llvm.org/4.0.1/llvm-4.0.1.src.tar.xz
-python,-d,/usr/local/Cellar/python/3.7.2_1,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
-python@2,-d,/usr/local/Cellar/python@2/2.7.15_2,python@2,https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tgz
+llvm,-d,/usr/local/opt/llvm@4,llvm@4,http://releases.llvm.org/4.0.1/llvm-4.0.1.src.tar.xz
+python,-d,/usr/local/Cellar/python/3\*,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
```
spoonincode: Looking in the Cellar is a bad idea IMO -- it's an implementation detail of sorts. Can't we look at `/usr/local/opt/python@2/` and `/usr/local/opt/python` as a more appropriate place?

NorseGaud: I was thinking about that too, it would simplify things up until the point python 4 is released and uses `/usr/local/opt/python`. Do we want to accommodate for that future situation?

spoonincode: That may not end up being a problem: is python 4 going to be a breaking change like 2->3 was?

NorseGaud: I think you're right actually. Probably not a big deal for now

scripts/eosio_build_darwin_deps
```diff
@@ -4,9 +4,9 @@ Libtool,-x,/usr/local/bin/glibtool,libtool,http://gnu.askapache.com/libtool/libt
 OpenSSL,-f,/usr/local/opt/openssl/lib/libssl.a,openssl,https://www.openssl.org/source/openssl-1.0.2n.tar.gz
 wget,-x,/usr/local/bin/wget,wget,https://ftp.gnu.org/gnu/wget/wget-1.19.2.tar.gz
 GMP,-f,/usr/local/opt/gmp/include/gmpxx.h,gmp,https://ftp.gnu.org/gnu/gmp/gmp-6.1.2.tar.bz2
-llvm,-x,/usr/local/opt/llvm@4/bin/clang-4.0,llvm@4,http://releases.llvm.org/4.0.1/llvm-4.0.1.src.tar.xz
-python,-d,/usr/local/Cellar/python/3.7.2_1,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
-python@2,-d,/usr/local/Cellar/python@2/2.7.15_2,python@2,https://www.python.org/ftp/python/2.7.13/Python-2.7.13.tgz
+llvm,-d,/usr/local/opt/llvm@4,llvm@4,http://releases.llvm.org/4.0.1/llvm-4.0.1.src.tar.xz
+python,-d,/usr/local/Cellar/python,python,https://www.python.org/ftp/python/3.7.2/Python-3.7.2.tgz
```
spoonincode: I thought we were going to get rid of Cellar path check?

NorseGaud: Apologies, you're right. I just realized that /usr/local/opt has python2 and python3 folders. That's way better.

PR #6867, opened 2019-03-05
Snap build

krrishnarraj:
## Change Description

Snap is a modern way of distributing applications on linux based systems. The project is maintained by Ubuntu and a wide array of OSes are supported including RPi https://docs.snapcraft.io/installing-snapd/6735

To accomplish snap build, a new cmake provision is added to build dependencies inline in a more portable way. All the dependencies required for eos are manually built inside build/ directory using CMakeModules/BuildDepsInline.cmake. The exact versions of each dependency & build options can be specified in one file. The cmake option 'BUILD_DEPS_INLINE' to enable this is turned off by default

### Snapcraft

Snapcraft is a store to publish and find snap apps https://snapcraft.io/store. The command to build a snap is 'snapcraft' which is available from snapcraft store. This can be installed as `sudo snap install --classic snapcraft`

TODO:

* Due to security policy of snap apps, eos snap binaries are available in system path as 'eos.nodeos', 'eos.cleos' & 'eos.keosd'. This can be aliased to nodeos, cleos & keosd by the maintainer in snapcraft store as per https://docs.snapcraft.io/deprecation-notice-5/8405
* To further add to security confinement of snap apps, they are given a different $HOME directory like "\~/snap/eos/". They are not allowed to write to "\~" by default. A separate plug has been added to access "\~". At the moment, nodeos stores data in "\~/snap/eos/common/", but keosd stores in "\~/eosio-wallet/" due to a hard-coded path in https://github.com/EOSIO/eos/blob/master/programs/keosd/main.cpp#L27 . The final $HOME directory strategy needs to be decided

### Windows cmake step

Since building dependencies is portable, we have a successfull cmake step on windows too. This is just a cmake step, the build fails with a big list of errors. Prerequisites for windows cmake step:

* Python installed
* OpenSSL dev package(non-light) from https://slproweb.com/products/Win32OpenSSL.html installed at C:\OpenSSL
* Delete outdated FindBoost.cmake in libraries/fc/CMakeModules or https://github.com/EOSIO/fc/commit/a41ae0edbc8f1e8ca0c0a3393d158481d8fca0ab which is yet to be merged
* set BUILD_DEPS_INLINE to ON while invoking cmake

## Consensus Changes

NA

## API Changes

NA

## Documentation Additions

* Will have to change documents related to manual build of dependencies if BUILD_DEPS_INLINE is to be used
* New documents for snap build might have to be added
* Initial document for windows cmake step


CMakeModules/BuildDepsInline.cmake
```diff
@@ -0,0 +1,173 @@
...
+    PREFIX              llvm
+    SOURCE_DIR          llvm/src/llvm
+    DOWNLOAD_COMMAND    \"\"
+    UPDATE_COMMAND      \"\"
+    BUILD_COMMAND       \"${CMAKE_COMMAND}\" --build . -- ${CMAKE_PARALLEL_OPT}
+    CMAKE_ARGS
+      -DCMAKE_INSTALL_PREFIX=${WASM_ROOT}
+      -DCMAKE_BUILD_TYPE=Release
+      -DLLVM_TARGETS_TO_BUILD=
+      -DLLVM_EXPERIMENTAL_TARGETS_TO_BUILD=WebAssembly
```
spoonincode: WebAssembly not needed

CMakeModules/BuildDepsInline.cmake
```diff
@@ -0,0 +1,173 @@
...
+    GIT_REPOSITORY      https://github.com/llvm-mirror/llvm
+    GIT_TAG             release_40
+    GIT_SHALLOW         1
+    GIT_PROGRESS        1
+    CONFIGURE_COMMAND   \"\"
+    BUILD_COMMAND       \"\"
+    INSTALL_COMMAND     \"\"
+  )
+
+  ExternalProject_add(  clang_clone
```
spoonincode: clang not needed

CMakeModules/BuildDepsInline.cmake
krrishnarraj: Done

CMakeModules/BuildDepsInline.cmake
krrishnarraj: Done

krrishnarraj:
> Unless you've got a working WIN32 build IMO we should not be adding WIN32 commands. The last thing our cmake files need is more inactive cruft.

Agreed on that. But this is the first step towards windows build. Currently the build is failing with a huge list of errors and a separate effort is required to fix them, which can be taken-up by someone. Are you sure about this?

spoonincode:
The road to proper WIN32 support can be travelled via bite sized PRs targeting the components that need WIN32 support. Off hand I can think of
* De-POSIXing the checktime timer
* Proper TLS CA handling
* Named pipe support for keosd
* Removal/refactoring of WAVM's SEH for mingw, if mingw is the intended build target

Those can be dealt with one by one in the future. Adding in cmake rules now that may not even be the proper solution (because they literally 'fail with a huge list of errors' at the moment) doesn't provide utility -- we won't even have a way to test that they're right.

krrishnarraj:
Windows specific options are gone now. I'll maintain them in a separate branch

CMakeModules/BuildDepsInline.cmake
```diff
@@ -0,0 +1,97 @@
+
+include(ProcessorCount)
+ProcessorCount(NUM_CORES)
+
+set(BOOST_ROOT "${CMAKE_CURRENT_BINARY_DIR}/boost_install")
+set(MONGO_DRIVER_ROOT "${CMAKE_CURRENT_BINARY_DIR}/mongo_driver_install")
+set(WASM_ROOT "${CMAKE_CURRENT_BINARY_DIR}/wasm_install")
```
spoonincode: WASM_ROOT is a misnomer -- it's just a build of llvm4 being placed in here

krrishnarraj: Done

krrishnarraj:
One check has failed. Can we see the log?

spoonincode:
>One check has failed. Can we see the log?

transient failure unrelated to your change

krrishnarraj:
What happens next?

PR #6893, opened 2019-03-07
Move json::to_string response processing to http thread pool

### LINKED ISSUE #6891: Add json::to_string to http thread pool
- Currently `json::to_string` of `variant` for http responses is on the main application thread. Move this processing to the http thread pool.

heifner:
## Change Description

- Resolves #6891 
- Move the conversion from `fc::variant` to JSON string reponse into the http thread pool to reduce the amount of work done on the application thread.

## API Changes

- The http_plugin `url_response_callback` changed from `std::function<void(int,std::string)>` to `std::function<void(int,fc::variant)>`
-- Code calling the url_response_callback is now expected to provide a `fc::variant` instead of a string. This means the the type must be setup with `FC_REFLECT` so it can be converted to variant. This was already the case for all plugins in `eos` repo, but if external plugins were doing their own json conversion then that will no longer work.

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -304,18 +304,20 @@ namespace eosio {
                               [ioc = this->server_ioc, &bytes_in_flight = this->bytes_in_flight, handler_itr,
                                resource{std::move( resource )}, body{std::move( body )}, con]() {
                      try {
-                        bytes_in_flight -= body.size();
                         handler_itr->second( resource, body,
-                              [ioc{std::move(ioc)}, &bytes_in_flight, con]( int code, std::string response_body ) {
-                           bytes_in_flight += response_body.size();
-                           boost::asio::post( *ioc, [ioc, response_body{std::move( response_body )}, &bytes_in_flight, con, code]() {
-                              size_t body_size = response_body.size();
-                              con->set_body( std::move( response_body ) );
+                              [ioc{std::move(ioc)}, &bytes_in_flight, con]( int code, fc::variant response_body ) {
+                           boost::asio::post( *ioc, [ioc, response_body{std::move( response_body )}, &bytes_in_flight, con, code]() mutable {
+                              std::string json = fc::json::to_string( response_body );
+                              response_body.clear();
+                              const size_t json_size = json.size();
+                              bytes_in_flight += json_size;
+                              con->set_body( std::move( json ) );
                               con->set_status( websocketpp::http::status_code::value( code ) );
                               con->send_http_response();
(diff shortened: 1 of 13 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -304,18 +304,20 @@ namespace eosio {
...
-                        bytes_in_flight -= body.size();
                         handler_itr->second( resource, body,
-                              [ioc{std::move(ioc)}, &bytes_in_flight, con]( int code, std::string response_body ) {
-                           bytes_in_flight += response_body.size();
-                           boost::asio::post( *ioc, [ioc, response_body{std::move( response_body )}, &bytes_in_flight, con, code]() {
-                              size_t body_size = response_body.size();
-                              con->set_body( std::move( response_body ) );
+                              [ioc{std::move(ioc)}, &bytes_in_flight, con]( int code, fc::variant response_body ) {
+                           boost::asio::post( *ioc, [ioc, response_body{std::move( response_body )}, &bytes_in_flight, con, code]() mutable {
+                              std::string json = fc::json::to_string( response_body );
```
brianjohnson5972: Can this be const? (or does move prevent that?)

plugins/http_plugin/http_plugin.cpp
```diff
@@ -304,18 +304,20 @@ namespace eosio {
...
                      try {
-                        bytes_in_flight -= body.size();
                         handler_itr->second( resource, body,
-                              [ioc{std::move(ioc)}, &bytes_in_flight, con]( int code, std::string response_body ) {
-                           bytes_in_flight += response_body.size();
-                           boost::asio::post( *ioc, [ioc, response_body{std::move( response_body )}, &bytes_in_flight, con, code]() {
-                              size_t body_size = response_body.size();
-                              con->set_body( std::move( response_body ) );
+                              [ioc{std::move(ioc)}, &bytes_in_flight, con]( int code, fc::variant response_body ) {
+                           boost::asio::post( *ioc, [ioc, response_body{std::move( response_body )}, &bytes_in_flight, con, code]() mutable {
```
brianjohnson5972: I don't see anything new that would require the mutable key word

plugins/http_plugin/http_plugin.cpp
heifner: The move prevents it.

plugins/http_plugin/http_plugin.cpp
heifner: I think you are correct. Originally it did need the mutable because originally the move into set_body was being ignored. I'm inclined to leave it mutable to avoid that kind of issue.

PR #6896, opened 2019-03-07
Update to C++ 17

larryk85:
## Change Description
Update libraries and CMake to build with C++ 17 instead of C++ 14.


CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ include( InstallDirectoryPermissions )
 include( MASSigning )
 
 set( BLOCKCHAIN_NAME "EOSIO" )
-set( CMAKE_CXX_STANDARD 14 )
+set( CMAKE_CXX_STANDARD 17 )
 set( CMAKE_CXX_EXTENSIONS ON )
```
spoonincode: Is it worth turning this off? Seems like it compiles okay with it off

spoonincode:
`set( CMAKE_CXX_STANDARD 17 )` requires a cmake minimum bump to 3.6 I think. Also I wonder if we could just completely remove the compiler version check due to `CXX_STANDARD_REQUIRED` being on

larryk85:
Closing. Superceded by PR #7107

Issue #6429, opened 2018-12-04
Consensus protocol upgrade activation mechanism


arhag:
This issue tracks the code changes needed to facilitate consensus protocol upgrades.

## Signaling consensus protocol upgrade features

Consensus protocol upgrade changes are associated with a feature (called a "consensus protocol upgrade feature") recognized by nodeos. Each feature will have an associated 256-bit digest that uniquely identifies it. There are two classes of consensus protocol upgrade features: those that require pre-activation via an intrinsic and those that do not.

A consensus protocol upgrade feature is activated by signaling its activation in the block header through a new block header extension type called the "protocol feature activation" extension. This extension type (which will be discussed in more detail below), includes the 256-bit digests of consensus protocol upgrade features to be activated at the start of that block.

There are constraints on whether a consensus protocol upgrade feature (of either class) can be activated:

- the nodeos software must be configured to recognize and accept the feature;
- and, any dependencies of the feature must have already been activated.

Furthermore, a consensus protocol upgrade feature of the class that requires pre-activation has further requirements:

- activation of the feature, signaled through the block header extension, is disallowed if the same feature has not already been pre-activated via an intrinsic;
- any block building on a block that included pre-activation of a feature must signal activation of that feature in its block header extension otherwise it is an invalid block.

Most consensus protocol upgrade features will likely be of the type that require pre-activation. Consensus protocol upgrade features that do not require pre-activation have the characteristic that any one of the active block producers can initiate the activation of the feature. The other block producers could still ignore that block if they were not ready to accept activation of the feature, but that is a subjective producer policy decision that would have to be coded into the nodeos software that the block producer was running. One type of subjective policy decision that will be formalized as part of this mechanism, however, is to optionally reject activation of features if they occur before a locally specified time known as the "earliest allowed activation time".

Having a type of consensus protocol upgrade feature that requires pre-activation allows the subjective policy decision by each producer to be replaced by an objective policy that runs on the blockchain. A system contract action with strict authorization requirements can use a special intrinsic to pre-activate the feature; and block producers can use, for example, the `eosio.msig` contract to satisfy the authority of that system contract action.

[...]

PR #6949, opened 2019-03-15
fix single producer node case for preactivation & add python test

taokayan:
## Change Description
Fix single producer node case for feature pre-activation for #6429 , added python test cases prod_preactivation_test.py

## Consensus Changes
No.

## API Changes
No.

## Documentation Additions
No.

### DIFF
--- a/tests/prod_preactivation_test.py
+++ b/tests/prod_preactivation_test.py
@@ -67,7 +67,7 @@ try:
         cluster.killall(allInstances=killAll)
         cluster.cleanup()
         Print("Stand up cluster")
-        if cluster.launch(pnodes=prodCount, totalNodes=prodCount, prodCount=prodCount, onlyBios=onlyBios, dontBootstrap=dontBootstrap, p2pPlugin=p2pPlugin, useBiosBootFile=False, extraNodeosArgs=" --plugin eosio::producer_api_plugin") is False:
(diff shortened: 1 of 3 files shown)


tests/prod_preactivation_test.py
```diff
@@ -0,0 +1,176 @@
...
+    node0 = cluster.getNode(0)
+    contract="eosio.bios"
+    contractDir="unittests/contracts/%s" % (contract)
+    wasmFile="%s.wasm" % (contract)
+    abiFile="%s.abi" % (contract)
+
+    Print("publish a new bios contract %s should fails because env.is_feature_activated unresolveable" % (contractDir))
+    retMap = node0.publishContract("eosio", contractDir, wasmFile, abiFile, True, shouldFail=True)
+
+    #Print(retMap)
```
brianjohnson5972: Same as above

tests/prod_preactivation_test.py
```diff
@@ -0,0 +1,176 @@
...
+    Print("PORT: %d" % (port))
+
+    if enableMongo and not cluster.isMongodDbRunning():
+        errorExit("MongoDb doesn't seem to be running.")
+
+    if localTest and not dontLaunch:
+        cluster.killall(allInstances=killAll)
+        cluster.cleanup()
+        Print("Stand up cluster")
+        if cluster.launch(pnodes=prodCount, totalNodes=prodCount, prodCount=prodCount, onlyBios=onlyBios, dontBootstrap=dontBootstrap, p2pPlugin=p2pPlugin, extraNodeosArgs=" --plugin eosio::producer_api_plugin") is False:
```
brianjohnson5972: Comments sounds like this test only expects one producer, but you allow prodCount to be set on the command line, and use it to set the number of producer nodes? Is that what you want to do?

PR #6952, opened 2019-03-15
state-history doc

tbfleming:
Temporary home for state-history plugin doc


heifner:
@tbfleming @lparisc anything here we don't already have documented?

brianjohnson5972:
@lparisc and @tbfleming does this need to be kept open?  Do we know what needs to happen?

jeffreyssmith2nd:
Closing due to lack of response/activity. Can be reopened if this is still needed.

PR #6958, opened 2019-03-18
* improve for sharding

UMU618:
## Change Description

improve for mongodb sharding


plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -1427,39 +1435,44 @@ void mongo_db_plugin_impl::init() {
          }
 
          try {
+            // Due to the vast amounts of data, we suggest:
+            //   1. enableSharding database (default to EOS)
+            //   2. shardCollection: blocks, action_traces, transaction_traces, especially action_traces
+            //   3. Use compound index with shard key (default to _id), to improve query performance.
+
```
heifner: Are you saying you would like something added to the mongo_db_plugin for 1 and 2 OR that you merely suggest users should enableSharding on there database and collections?

plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -1427,39 +1435,44 @@ void mongo_db_plugin_impl::init() {
...
             // blocks indexes
             auto blocks = mongo_conn[db_name][blocks_col];
-            blocks.create_index( bsoncxx::from_json( R"xxx({ "block_num" : 1 })xxx" ));
-            blocks.create_index( bsoncxx::from_json( R"xxx({ "block_id" : 1 })xxx" ));
+            blocks.create_index( bsoncxx::from_json( R"xxx({ "block_num" : 1, "_id" : 1 })xxx" ));
+            blocks.create_index( bsoncxx::from_json( R"xxx({ "block_id" : 1, "_id" : 1 })xxx" ));
```
heifner: Do you have any feel for how much larger the indexes are when _id is added?

plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -310,10 +310,12 @@ void mongo_db_plugin_impl::queue( Queue& queue, const Entry& e ) {
 }
 
 void mongo_db_plugin_impl::accepted_transaction( const chain::transaction_metadata_ptr& t ) {
+   if( !store_transactions ) {
+      return;
+   }
+
    try {
```
heifner: Why move these statements out of try/catch? Do you see faster performance?

plugins/mongo_db_plugin/mongo_db_plugin.cpp
UMU618: I suggest users should do 1 and 2 on production environment.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
UMU618: This plugin itself is CPU-bound.  if-return statements are never throw, so it's not necessary in try-block which may have performance costs. (It depends on the compiler)

plugins/mongo_db_plugin/mongo_db_plugin.cpp
UMU618:
```
use EOS
db.action_traces.stats()
```
shows:
```
"indexSizes" : {
				"_id_" : 18581565440,
				"trx_id_1" : 16938258432,
				"block_num_1" : 4512980992,
				"block_num_1__id_1" : 21147549696
```

`block_num_1__id_1` is 16,634,568,704 bytes larger than `block_num_1`, close to "_id_" : 18,581,565,440.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
heifner: Can you update the comment to make that clearer.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
heifner: I would be surprised if that made any difference but would be interested if you could point me to anything that demonstrates that.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
heifner: I wonder if the plugin should even create any indexes and let that be left to the user. However, I'm inclined to accept this PR if you rebase to `develop`, remove the try-catch changes (or provide some article/test/proof it makes a difference), and update the comment.

Thanks.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
UMU618: The difference is tiny. It's more a coding style issue. I'll remove the try-catch changes later.

PR #6963, opened 2019-03-18
Created test scripts with xUnit, auto-scaling parallelism, and empty test detection; and added them to pipeline

kj4ezj:
## Change Description

I moved the test steps for serial, parallel, and long-running tests into bash scripts. This enables us to use features not supported by raw Buildkite commands, which run in shell.

Tests now:
+ Fail if there are zero tests to run
+ Automatically detect the number of CPU cores and scale across all of them for parallelizable tests (previously we assumed 8 cores)
+ Produces an xUnit-formatted XML file containing test results
+ Uploads the XML file as a test artifact

Also increased LRT time limit from 60 to 90 minutes, tested in [build 2218](https://buildkite.com/EOSIO/eosio-lrt/builds/2218).

Zane asked me to delete the following files, which have already been migrated to our central Buildkite repo:
/.buildkite/coverage.yml
/.buildkite/debug.yml
/.buildkite/docker.yml
/.buildkite/sanitizers.yml

## Consensus Changes
- [ ] Consensus Changes

None.

## API Changes
- [ ] API Changes

None.

## Documentation Additions
- [ ] Documentation Additions

None.


scripts/parallel-test.sh
```diff
@@ -0,0 +1,28 @@
...
+echo "Starting MongoDB..."
+~/bin/mongod --fork --dbpath ~/data/mongodb -f ~/etc/mongod.conf --logpath "$(pwd)"/mongod.log
+cd /data/job/build
+# run tests
+echo "Running tests..."
+CPU_CORES=$(getconf _NPROCESSORS_ONLN)
+echo "$CPU_CORES cpu cores detected."
+TEST_COUNT=$(ctest -N -LE _tests | grep -i 'Total Tests: ' | cut -d ':' -f 2 | awk '{print $1}')
+[[ $TEST_COUNT > 0 ]] && echo "$TEST_COUNT tests found." || (echo "ERROR: No tests registered with ctest! Exiting..." && exit 1)
+echo "$ ctest -j $CPU_CORES -LE _tests --output-on-failure -T Test"
```
spoonincode: If you want to print the command you're running and avoid duplication (which is at risk of diverging over time), one approach is

`(set -x; echo hello)`

kj4ezj: Unfortunately I did not see this comment before I merged, I will be more cautious in the future!

I agree with you that this would be a better approach in these test scripts. I ran this on my local machine in both iterm2 and terminal, and it didn't have the expected behavior that you and others online described. Both "set -x" and "set -v" print a bunch of junk, looks like a stack trace, on my local machine. This undesired behavior must be a mac thing because both worked as expected in a linux docker image.

Either way, we will be doing more work on this pipeline soon and I have added this to my notes. I will implement a platform-independent method of echoing desired commands without duplication next time I work on this pipeline.

PR #6982, opened 2019-03-21
#6980: fix memory leak when --trace-history not used

### LINKED ISSUE #6980: Memory leak in state-history plugin
If the state-history plugin is run without the `--trace-history` option, then it will collect traces in `cached_traces` but never flush the cache.

Possibly related to #6377.

tbfleming:
## Change Description
Fix #6980. There was a memory leak when `--trace-history` not used

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
@tbfleming do you need a `develop` PR? This will not be automatically merged to develop.

tbfleming:
yes

Issue #6909, opened 2019-03-09
Replay slow down around block number 32M running against Mainnet data


nksanthosh:
Nodeos replay with Mainnet data slows down as shown in the attached graph

1) Restart nodeos @ 32-33 million blocks:

Safely restarting nodeos while replay is between 32 & 33 million blocks seems to optimize total replay time, across all 3 tests.  Later than this means the replay stalls significantly, earlier and a second restart is needed or replay may stall prior to completion.

2) Restart nodeos after replay completes:

Post-replay sync performance (catch up with blocks generated after replay start) begins around 500-1000 blocks per minute.  A 2nd restart during this sync increases performance by a factor of about 5 (again, consistent across all 3 tests), and decreases total sync time from ~24 hours to ~4-5 hours.

nksanthosh:
![replayissues](https://user-images.githubusercontent.com/8515898/54076280-e8747b00-4277-11e9-9fd4-2a849ecff022.png)

wanderingbort:
@nksanthosh @dconry Please specify the configuration and hardware used to derive this graph so that future readers of this issue can properly understand.  

My assumption is that this graph looks very different with `wabt` as a web assembly back-end even if it is slower in an absolute sense.

dconry:
@wanderingbort @nksanthosh Here are some hardware/runtime details for tests above, numbered left to right in graph:

Test 1 (aws m4.4xlarge: 8x2.5GHz CPU/64GB RAM): *--replay --wasm-runtime wavm* on nodeos 1.6.0rc2
Test 2 (aws m4.4xlarge: 8x2.5GHz/64GB): *--replay --wasm-runtime wavm --disable-replay-opts --trace-history --chain-state-history --plugin eosio::state_history_plugin* on nodeos 1.6.1
Test 3 (aws z1d.6xlarge: 12x3.4-4.0Ghz/192GB): *--replay --wasm-runtime wavm* on nodeos 1.6.1

Also, tests 1 & 2 were allowed to process slowly for several days before a nodeos restart unstalled them.  Test 3 was restarted promptly (prior to stall) in attempt to avoid stall period.  This (plus faster CPU freq) contributed to the faster completion seen in graph for Test 3.

Please let me know if further details would be useful...

matthewdarwin:
Might be related to https://github.com/EOSIO/eos/issues/6533?

dconry:
[...]

PR #6983, opened 2019-03-22
 Clean up wasm cache entries based on irreversibility & fix wavm module cleanup 

spoonincode:
## Change Description
wasm cache entries have previously remained indefinitely creating a sort of memory leak. If using the wavm runtime this also creates a performance drag. This change is a naive fix to clean the wasm cache. It improves upon #2701 in that it waits for irreversibility before cleaning the cache (reducing cache thrashing around forks), and it also implements the garbage collection tracking for wavm so that wavm will actually free its instances.

Unfortunately there are still problems with this naive approach that makes this change not acceptable. For one, code instantiated on a fork that is thrown out is never pruned. More severely, code that is shared by multiple accounts tends to be overly pruned because getting a reference count of code used upon nodeos startup is required to avoid that.

This is mainly be put up as draft at this time to judge improvement with #6909. Unfortunately we still leave a _lot_ of performance on the table with wavm (nearly 50%) even with this fix but it should at least not grind to a halt any longer.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -295,6 +295,8 @@ struct controller_impl {
          }
          emit(self.irreversible_block, s);
       }
+
+      wasmif.prune_wasm_cache(s->block_num);
    }
 
    void replay(std::function<bool()> shutdown) {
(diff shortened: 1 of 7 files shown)


spoonincode:
superseded by other PRs

PR #6987, opened 2019-03-22
Vote Test => Schedule Change Test: Fix Intermittent Failures, Simplify to Reflect Purpose

kj4ezj:
## Change Description

Previously I diagnosed the intermittent long-running vote test failures. I identified next steps to make the test faster and minimize the test to reflect its purpose, which I expect to eliminate the intermittent failures as a by-product.

The purpose of this test is to verify that only the correct producers are participating after the schedule has been changed and to verify that their participation advances the last irreversible block without any forks.

This is an integration test focusing on how the net plugin plays with controller and the rest of nodeos. We already have unit tests which test the block producer schedule in controller.cpp, and we already have unit tests which test the voting contracts.

**Story 1** — Vote Test Change Schedule Using BIOS
2019/03/20-22 (first four commits)
- Rename test to reflect purpose
   - "nodeos_voting_test.py" => "nodeos_schedule_change_test.py"
   - "nodeos_voting_lr_test" => "nodeos_schedule_change_lr_test"
   - "nodeos_voting_bnet_lr_test" => "nodeos_schedule_change_bnet_lr_test"
- Change block producer schedule using eosio.bios contract instead of voting on eosio.system contract
   - Copied cluster.boostrap() to create cluster.minBootstrap(), which does not initialize system contract
   - Added a boolean flag to cluster.launch() called minimize which causes launch() to call minBootstrap() instead of legacy bootstrap()
   - Deleted code dependent on eosio.system from nodeos_schedule_change_test.py
This has broken verifyProductionRounds(), the test validation code. This has not been fixed as the follow-up story involves rewriting this function, so investing time in fixing this now would be out-of-scope. As a result, these changes cannot be merged at this time.

**Story 2** — Add Checks to Vote Test
Date TBD
- Remove schedule checking
- Update delay
- Verify block producer participation
- Verify last irreversible block advances
- Minimize code in nodeos_schedule_change_test.py
- Minimize code in cluster.minBootstrap()

## Consensus Changes
- [ ] Consensus Changes

None.

## API Changes
- [ ] API Changes

None.

## Documentation Additions
- [ ] Documentation Additions

None.


kj4ezj:
I think with the work Hong Kong is doing, this is no longer relevant. There is no sense implementing the second half of these changes when Hong Kong is writing a new testing framework and we will eventually throw out the Python one.   
I think it would be better to wait until they are done and invest time in migrating this test over to the new framework, or reimplementing it on the new framework altogether.

PR #7005, opened 2019-03-28
Python 36 for centos7 and amazonlinux1

NorseGaud:
scandir was introduced in python 3.5


NorseGaud:
https://buildkite.com/EOSIO/eosio-nightly-builds/builds/239

PR #7017, opened 2019-03-29
[XC 10.2] Remove boost::thread usage from mongo plugin

spoonincode:
## Change Description
boost::thread is problematic on some new compiler + old boost combos

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/mongo_db_plugin/mongo_db_plugin.cpp
+++ b/plugins/mongo_db_plugin/mongo_db_plugin.cpp
@@ -16,11 +16,10 @@
 #include <boost/algorithm/string.hpp>
 #include <boost/chrono.hpp>
 #include <boost/signals2/connection.hpp>
-#include <boost/thread/thread.hpp>
-#include <boost/thread/mutex.hpp>
-#include <boost/thread/condition_variable.hpp>
 
 #include <queue>
+#include <thread>
+#include <mutex>
 
 #include <bsoncxx/builder/basic/kvp.hpp>
 #include <bsoncxx/builder/basic/document.hpp>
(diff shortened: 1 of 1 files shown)


plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -292,15 +291,15 @@ bool mongo_db_plugin_impl::filter_include( const transaction& trx ) const
 
 template<typename Queue, typename Entry>
 void mongo_db_plugin_impl::queue( Queue& queue, const Entry& e ) {
-   boost::mutex::scoped_lock lock( mtx );
+   std::unique_lock<std::mutex> lock( mtx );
```
spoonincode: This could use a double check

heifner: `std::lock_guard` would be slightly better. But probably not worth pushing a new commit.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -408,7 +407,7 @@ void mongo_db_plugin_impl::consume_blocks() {
       _account_controls = mongo_conn[db_name][account_controls_col];
 
       while (true) {
-         boost::mutex::scoped_lock lock(mtx);
+         std::unique_lock<std::mutex> lock(mtx);
          while ( transaction_metadata_queue.empty() &&
```
heifner: This could also be `std::lock_guard`, but again not worth changing unless you change the one above.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
spoonincode: I don't think `std::lock_guard` works here because the code then does things like `lock.unlock();` -- `unlock()/lock()` isn't exposed by lock_guard AFAICT

heifner: You are correct. I missed the `unlock()/lock()`

PR #7031, opened 2019-03-29
Use shared_future - v1.7.x

heifner:
## Change Description

- Since `transaction_metadata` future is accessed from multiple threads, use a `shared_future` which is thread safe.
- Simplified `recover_key` usage and removed storage of `sig_cpu_usage` and `signing_keys` since they are stored in the `shared_future`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -999,9 +999,10 @@ struct controller_impl {
          auto start = fc::time_point::now();
          const bool check_auth = !self.skip_auth_check() && !trx->implicit;
          // call recover keys so that trx->sig_cpu_usage is set correctly
-         const flat_set<public_key_type>& recovered_keys = check_auth ? trx->recover_keys( chain_id ) : flat_set<public_key_type>();
+         const fc::microseconds sig_cpu_usage = check_auth ? std::get<0>( trx->recover_keys( chain_id ) ) : fc::microseconds();
+         const flat_set<public_key_type>& recovered_keys = check_auth ? std::get<1>( trx->recover_keys( chain_id ) ) : flat_set<public_key_type>();
          if( !explicit_billed_cpu_time ) {
-            fc::microseconds already_consumed_time( EOS_PERCENT(trx->sig_cpu_usage.count(), conf.sig_cpu_bill_pct) );
+            fc::microseconds already_consumed_time( EOS_PERCENT(sig_cpu_usage.count(), conf.sig_cpu_bill_pct) );
 
             if( start.time_since_epoch() <  already_consumed_time ) {
                start = fc::time_point();
@@ -1199,7 +1200,7 @@ struct controller_impl {
                auto& pt = receipt.trx.get<packed_transaction>();
                auto mtrx = std::make_shared<transaction_metadata>( std::make_shared<packed_transaction>( pt ) );
                if( !self.skip_auth_check() ) {
-                  transaction_metadata::create_signing_keys_future( mtrx, thread_pool, chain_id, microseconds::maximum() );
+                  transaction_metadata::start_recover_keys( mtrx, thread_pool, chain_id, microseconds::maximum() );
                }
                packed_transactions.emplace_back( std::move( mtrx ) );
             }
--- a/libraries/chain/include/eosio/chain/transaction_metadata.hpp
+++ b/libraries/chain/include/eosio/chain/transaction_metadata.hpp
@@ -15,7 +15,8 @@ namespace eosio { namespace chain {
 
 class transaction_metadata;
 using transaction_metadata_ptr = std::shared_ptr<transaction_metadata>;
-using signing_keys_future_type = std::shared_future<std::tuple<chain_id_type, fc::microseconds, flat_set<public_key_type>>>;
+using signing_keys_future_value_type = std::tuple<chain_id_type, fc::microseconds, flat_set<public_key_type>>;
+using signing_keys_future_type = std::shared_future<signing_keys_future_value_type>;
 using recovery_keys_type = std::pair<fc::microseconds, const flat_set<public_key_type>&>;
 
 /**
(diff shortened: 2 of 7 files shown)


heifner:
@arhag To make `recover_keys` so that it could work without first calling `start_recover_keys`, `recover_keys` would need to take the `thread_pool` and `time_limit`. I just went with asserts for bad usage. I do think it is cleaner.

arhag:
@heifner That's okay, but `recover_keys` signature would not need to change. If the future was not set up or did not match the requested chain ID, it would not try to do the recovery asynchronously. It would just compute the results right there with no time limit. Thus no need for `thread_pool` or `time_limit`.

libraries/chain/controller.cpp
```diff
@@ -999,9 +999,10 @@ struct controller_impl {
          auto start = fc::time_point::now();
          const bool check_auth = !self.skip_auth_check() && !trx->implicit;
          // call recover keys so that trx->sig_cpu_usage is set correctly
-         const flat_set<public_key_type>& recovered_keys = check_auth ? trx->recover_keys( chain_id ) : flat_set<public_key_type>();
+         const fc::microseconds sig_cpu_usage = check_auth ? std::get<0>( trx->recover_keys( chain_id ) ) : fc::microseconds();
```
arhag: Too bad we can't use C++17 structured bindings yet.

libraries/chain/include/eosio/chain/transaction_metadata.hpp
```diff
@@ -50,13 +50,14 @@ class transaction_metadata {
...
       }
 
-      const flat_set<public_key_type>& recover_keys( const chain_id_type& chain_id );
-
       // must be called from main application thread
-      // signing_keys_future should only be accessed by main application thread
       static signing_keys_future_type
-      create_signing_keys_future( const transaction_metadata_ptr& mtrx, boost::asio::thread_pool& thread_pool,
-                                  const chain_id_type& chain_id, fc::microseconds time_limit );
+      start_recover_keys( const transaction_metadata_ptr& mtrx, boost::asio::thread_pool& thread_pool,
```
arhag: I think this should be a member function rather than a static function. And it should not return anything. The `signing_keys_future` can be made protected.

plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -779,8 +779,8 @@ void mongo_db_plugin_impl::_process_accepted_transaction( const chain::transacti
    }
 
    string signing_keys_json;
-   if( t->signing_keys.valid() ) {
-      signing_keys_json = fc::json::to_string( t->signing_keys->second );
+   if( t->signing_keys_future.valid() ) {
```
arhag: This logic would be simplified if `recover_keys` synchronously recovered the keys with no time limit.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -351,9 +351,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       void on_incoming_transaction_async(const transaction_metadata_ptr& trx, bool persist_until_expired, next_function<transaction_trace_ptr> next) {
          chain::controller& chain = chain_plug->chain();
          const auto& cfg = chain.get_global_properties().configuration;
-         signing_keys_future_type future =
-               transaction_metadata::create_signing_keys_future( trx, *_thread_pool, chain.get_chain_id(),
-                     fc::microseconds( cfg.max_transaction_cpu_usage ) );
+         signing_keys_future_type future = transaction_metadata::start_recover_keys( trx, *_thread_pool,
+               chain.get_chain_id(), fc::microseconds( cfg.max_transaction_cpu_usage ) );
          boost::asio::post( *_thread_pool, [self = this, future, trx, persist_until_expired, next]() {
             if( future.valid() )
```
arhag: Instead of waiting on the future. You could simply call `recover_keys` to ensure it was computed and cached.

libraries/chain/transaction_metadata.cpp
```diff
@@ -4,35 +4,32 @@
...
-         }
+   if( signing_keys_future.valid() ) {
+      const std::tuple<chain_id_type, fc::microseconds, flat_set<public_key_type>>& sig_keys = signing_keys_future.get();
+      if( std::get<0>( sig_keys ) == chain_id ) {
+         return std::make_pair( std::get<1>( sig_keys ), std::cref( std::get<2>( sig_keys ) ) );
       }
-      flat_set<public_key_type> recovered_pub_keys;
-      sig_cpu_usage = packed_trx->get_signed_transaction().get_signature_keys( chain_id, fc::time_point::maximum(), recovered_pub_keys );
-      signing_keys.emplace( chain_id, std::move( recovered_pub_keys ));
+      EOS_ASSERT( false, chain_id_type_exception, "chain id ${cid} does not match start_recover_keys ${sid}",
```
arhag: Now that I think about it more, I strongly prefer to maintain the old behavior of always returning a valid result rather than throwing an exception. That way even if it was used incorrectly somehow in the code (i.e. not calling `start_recover_keys` before), we know it wouldn't throw an exception when before it didn't (and thus be a consensus break).

So, if the future is not valid or it doesn't match the provided `chain_id`, it should replace the future with a new one for the requested `chain_id` and then immediately synchronously recover the keys with no deadline and set the results into the future, before then grabbing it out from the future to return.

libraries/chain/controller.cpp
heifner: I know. This is a bit messy.

plugins/producer_plugin/producer_plugin.cpp
heifner: Would require a mutex.

libraries/chain/include/eosio/chain/transaction_metadata.hpp
heifner: Can't be a member function unless we derive transaction_metdata from `enable_shared_from_this`. I'm ok with deriving from `enabled_shared_from_this` but that is the reason it is static.

arhag: No, it's okay. We can just stick with minimal changes for this patch release.

plugins/producer_plugin/producer_plugin.cpp
arhag: We can wait for the minor release for the larger changes needed to actually make `transaction_metadata` thread-safe.

PR #7036, opened 2019-03-31
[XC 10.2] Remove final boost thread references in cmake & change macos boost build

spoonincode:
## Change Description
Because this could be the last boost thread reference, we need to tell cmake to still pass thread compiler flags. There is also an appbase sync in here that removes boost thread there and picks up another unsynced change.

Also, when building boost on macos, only build the libraries needed by eosio

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


NorseGaud:
Hey @spoonincode, just wanted to mention that we can use $CPU_CORE instead of the `$(sysctl -in machdep.cpu.core_count)`.

PR #7044, opened 2019-04-01
Restructure transaction traces

### LINKED ISSUE #6897: Restructure transaction traces to disambiguate and provide more information
The current transaction trace is structured in a manner that prevents determining some potentially useful information about transaction execution due to ambiguities in the structure. For example, if an action notifies another contract, and there is another inline action in the `inline_traces` of the original action, it is not clear whether that inline action was sent by the original action or the notified contract. This information can be useful if an application wishes to react to the event of the inline action but only if sent by the appropriate contract. Adding authorizations can help, but it does not fully resolve the potential ambiguities (also authorizations are not available for context-free actions).

Consider a transaction with a single action `alice::act1` which initiates the following sequence of operations:

1.  `alice::act1` executes in the `alice` contract which:
    * calls `require_recipient` with `bob`;
    * sends two inline actions `alice::act2` and then `alice::act3`;
    * calls `require_recipient` with `charlie`.
2.  `alice::act1` executes in the `bob` contract which:
    * sends an inline action `bob::foo`;
    * calls `require_recipient` with `david`.
3.  `alice::act1` executes in the `charlie` contract which:
    * sends an inline action `charlie::bar`.
4.  `alice::act1` executes in the `david` contract which does nothing.
5.  `alice::act2` executes in the `alice` contract which:
    * calls `require_recipient` with `david` and then `erin`;
    * sends an inline action `alice::act4`.
6.  `alice::act2` executes in the `david` contract which does nothing.
7.  `alice::act2` executes in the `erin` contract which does nothing.
8.  `alice::act4` executes in the `alice` contract which does nothing.
9.  `alice::act3` executes in the `alice` contract which does nothing.
10.  `bob::foo` executes in the `bob` contract which does nothing.
11. `charlie::bar` executes in the `charlie` contract which does nothing.

The current implementation of nodeos will generate a transaction trace with the following structure:

```
(alice::act1 -> alice) [1]
   ||
   ||===> (alice::act1 -> bob) [2]
   ||
   ||===> (alice::act1 -> charlie) [3]
   ||
   ||===> (alice::act1 -> david) [4]
   ||
   ||===> (alice::act2 -> alice) [5]
   ||        ||
   ||        ||===> (alice::act2 -> david) [6]
   ||        ||
   ||        ||===> (alice::act2 -> erin) [7]
   ||        ||
   ||        ||===> (alice::act4 -> alice) [8]
   ||
   ||===> (alice::act3 -> alice) [9]
   ||
   ||===> (bob::foo -> bob) [10]
   ||
   ||===> (charlie::bar -> charlie) [11]
```

The numbers in the bracket act as convenient labels to refer to the individual actions but also show the order of execution of the actions. This would correspond to the order of the global sequence numbers included in each action receipt. 

One problem with the above structure is that action traces 9 and 10 both appear as children of action trace 1, however inline action 9 was sent by the execution of action 1 while inline action 10 was sent by the execution of action 2.

Another example of the problem of the above structure is that action trace 4 appears as a notification of action 1 to `david` in a similar manner as the notifications show in action traces 2 and 3; however, action traces 2 and 3 are there because of a `require_recipient` calls made during the execution of action 1, while action trace 4 is there because of a `require_recipient` call made during the execution of action 2.

Finally consider what the returned trace would look like if a failure was encountered during execution of action 5 in between the `require_recipient` calls that notified `david` and `erin`:

```
(alice::act1 -> alice) [1]
   ||
   ||===> (alice::act1 -> bob) [2]
   ||
   ||===> (alice::act1 -> charlie) [3]
   ||
   ||===> (alice::act1 -> david) [4]
   ||
   ||===> (alice::act2 -> alice) [5] (action trace will include error encountered)
```

However, up to the point of encountering the error, more information was known that was simply discarded because of the way transaction traces are structured. For example, it was known that action 2 sent inline action 10 and action 3 sent inline action 11, even though actions 10 and 11 did not have a chance to execute yet. It was also known that action 5 has notified `david`, however the notification to `erin` and inline action 8 being sent would not have been known.

The transaction trace is structured in a way that makes it easy to follow the order of execution. However, enough information already exists in the action receipts of the traces (specifically the `global_sequence`) to determine the exact order in which the actions executed. What does not exist is the information of the type described in the paragraphs above.

So this issue proposes changing the transaction trace structure to provide more information about the order in which things were scheduled/notified as well as the scheduling/notifying/dispatching relationship between actions. Since the `global_sequence` information would still exist, clients could transform the returned trace structure back into the original structure, thus giving users a choice of which representation they wish to see.

Under this new system the transaction trace of the example above would instead be:

```
(alice::act1 -> alice) [1]
   ||
   ||===> (alice::act1 -> bob) [2]
   ||        ||
   ||        ||===> (bob::foo -> bob) [10]
   ||        ||
   ||        ||===> (alice::act1 -> david) [4]
   ||
   ||===> (alice::act2 -> alice) [5]
   ||        ||
   ||        ||===> (alice::act2 -> david) [6]
   ||        ||
   ||        ||===> (alice::act2 -> erin) [7]
   ||        ||
   ||        ||===> (alice::act4 -> alice) [8]
   ||
   ||===> (alice::act3 -> alice) [9]
   ||
   ||===> (alice::act1 -> charlie) [3]
             ||
             ||===> (charlie::bar -> charlie) [11]
```

Furthermore, the scenario of encountering a failure during execution of action 5 described earlier would now result in a transaction trace of the following:

```
(alice::act1 -> alice) [1]
   ||
   ||===> (alice::act1 -> bob) [2]
   ||        ||
   ||        ||===> (bob::foo -> bob) [10] (action trace would have an empty receipt)
   ||        ||
   ||        ||===> (alice::act1 -> david) [4]
   ||
   ||===> (alice::act2 -> alice) [5] (action trace would have an empty receipt but would include error encountered)
   ||        ||
   ||        ||===> (alice::act2 -> david) [6]  (action trace would have an empty receipt and empty parent_seq_num)
   ||
   ||===> (alice::act3 -> alice) [9]  (action trace would have an empty receipt)
   ||
   ||===> (alice::act1 -> charlie) [3]
             ||
             ||===> (charlie::bar -> charlie) [11]  (action trace would have an empty receipt)
```

More information is provided in the above structure which could help contract developers debugging their contracts. Notice that the actions that did not have a chance to execute fully (or at all) would have an empty receipt. Child traces of an action trace with an empty receipt would not necessarily be complete, but would be as accurate as possible with the information known by the time of the failure.

The `base_action_trace` structure would be modified as follows:
   * Change the type of the `receipt` field from `action_receipt` to `optional<action_receipt>`.
   * Add a new field `parent_seq_num` of type `optional<uint64_t>` which be set to the `global_sequence` number of the parent action (the action that sent this inline action or called `require_recipient` causing this action to be dispatched). This field would be empty for the original top-level actions of the transaction.

## Implementation

The goal is to only change the transaction trace structure as described above, and not to actually change the way actions get executed. Even if there is a more sensible way to do action dispatching, the order cannot be changed without breaking existing consensus rules (and this issue is not about a consensus protocol upgrade). Thus, the core code of the dispatch logic of actions, in `apply_context::exec`, will remain the same.

However, the general pattern of allocating new space in the `inline_traces` with an `emplace_back` at the time right before executing will be replaced by instead doing that trace allocation (as well as a copy of the action data to be eventually dispatched) during the time of scheduling or notification.

The `_notified` field of `apply_context` should be modified to be a `vector<pair<account_name, size_t>>`. The `_inline_actions` and `_cfa_inline_actions` fields of `apply_context` should be modified to be a `vector<size_t>`. A new field `_trace` of type `action_trace*` should be added to `apply_context`. On execution of `apply_context::exec( action_trace& trace )` this field will be set to `&trace`.

In `apply_context::require_recipient`, instead of pushing back `recipient` into `_notified`, the code should add `action_trace` initialized with the current `act` to the end of `trace->inline_traces` and then do  `_notified.emplace_back( recipient, trace->inline_traces.size() - 1 )`.

In `apply_context::execute_inline` and `apply_context::execute_context_free_inline`, instead of doing an `emplace_back( move(a) )` on the `_inline_action` or `_cfa_inline_actions` field, respectively, the code should add `action_trace` initialized with the moved `a` to the end of `trace->inline_traces` and then call the method `emplace_back( trace->inline_traces.size() - 1 )` on the `_inline_action` or `_cfa_inline_actions` field, respectively.

When the new `action_trace` is initialized as described above, the fields `act`, `context_free`, `trx_id`, `block_num`, `block_time`, `producer_block_id` can and should all be properly initialized (and would never have to modified again). The remaining fields should be initialized with their default constructed values and, with the exception of `parent_seq_num`, may only be changed during/after execution of the action in `apply_context::exec_one`. The `parent_seq_num` field of an action will be modified during the `apply_context::exec_one` execution for the parent action by iterating through all of the action traces in `trace.inline_traces` after `r.global_sequence` is determined.

In `apply_context::exec`, wherever `trace.inline_traces.emplace_back()` and `trace.inline_traces.back()` exist, they will have to be replaced with a lookup in the `trace.inline_traces` vector with the appropriate index retrieved either from one of the three vectors: `_retrieved`, `_cfa_inline_actions`, or `_inline_actions`. Furthermore, when looping through the actions in `_cfa_inline_actions` or `_inline_actions`, the `inline_action` variable can instead be replaced by the `act` field of the `action_trace` found via the index.

arhag:
## Change Description

Resolves #6897.

Modifies transaction trace structure:

   * Adds new optional field `account_ram_delta` to `transaction_trace` which, if present, tracks the delta in RAM usage for the payer for the deferred transaction. (Note that the RAM initially billed for a contract-generated deferred transaction is not included as part of this field.)
   * Removes `inline_traces` from `action_trace` and switches to flat structure for all action traces.
   * Adds the following fields to `action_trace`:
      + `receiver` which is the receiver of the action (identical to the one inside the `receipt`, if present);
      + `action_ordinal` an unsigned integer used to identify the action trace within the transaction;
      + `creator_action_ordinal` is the ordinal of the action trace that is the parent of the current action trace within the creation tree representation of the transaction trace;
      + `parent_action_ordinal` is the ordinal of the action trace that is the parent of the current action trace within the execution tree representation of the transaction trace.

On error, the transaction trace will now included the action traces for any actions that were scheduled but not yet successfully executed by the point of the error. Any action that has not successfully executed will have an empty `receipt`.

The `history_plugin`, `mongo_db_plugin`, and `state_history_plugin` have been modified to work with the `transaction_trace` structure.

The `push_transaction` and `push_transactions` RPCs of the chain API have been modified to convert the new transaction trace structure back into a format similar to the old structure prior to returning the converted result back to the client. The converted structure recovers the old "execution tree" representation (using the `inline_traces` field which is brought back) that existing clients may be expecting. However, the new fields in both `transaction_trace` and `action_trace` are still included in this converted structure.

A new RPC `send_transaction` has been added to the chain API which does the same thing as `push_transaction` except it does not do any conversion back to the old structure. It returns the `transaction_trace` as it is with the new flat structure.

New test cases have been added to `api_tests` to verify the structure of the `transaction_trace` after executing variations of the example scenario discussed in #6897.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

Add `send_transaction` RPC to chain API.

## Documentation Additions
- [x] Documentation Additions

Documentation will need to be updated to include the new `send_transaction` RPC as part of the chain API. `send_transaction` is preferred over `push_transaction` since it avoids doing the additional conversion work that `push_transaction` now does to support backwards compatibility.

The structure of `transaction_trace` has been modified which has an impact to any plugins that may be consuming the `applied_transaction` signal.


libraries/chain/transaction_context.cpp
```diff
@@ -566,14 +572,63 @@ namespace bacc = boost::accumulators;
...
-      acontext.receiver     = receiver;
+   action_trace& transaction_context::get_action_trace( uint32_t action_ordinal ) {
+      EOS_ASSERT( 0 < action_ordinal && action_ordinal <= trace->action_traces.size() ,
+                  transaction_exception, "invalid action_ordinal" );
+      return trace->action_traces[action_ordinal-1];
+   }
+
+   const action_trace& transaction_context::get_action_trace( uint32_t action_ordinal )const {
+      EOS_ASSERT( 0 < action_ordinal && action_ordinal <= trace->action_traces.size() ,
+                  transaction_exception, "invalid action_ordinal" );
```
heifner: Add what the `action_ordinal` value was and `action_traces.size()`

PR #7047, opened 2019-04-02
Remove unwinding frames from WAVM code generation

spoonincode:
## Change Description

WAVM, via LLVM, generates unwinding frames for each wasm that is instantiated. This allows exceptions to unwind through the JITed code. Unfortunately when hundreds or thousands of these unwinding frames are registered exception unwinding becomes exceptionally slow.

This change removes generation of the unwinding frames. Because exceptions can no longer unwind through WAVM’s compiled code, all exceptions must be caught before unwinding reaches the compiled code. This means none of the intrinsics — either the ones we register or the ones wavm registers itself — can allow an exception to escape. Instead, the exception_ptr is stored and we longjmp through the compiled code.

This can provide substantial performance improvement when there are thousands of wasm instances active at once.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/wasm-jit/Source/Runtime/WAVMIntrinsics.cpp
+++ b/libraries/wasm-jit/Source/Runtime/WAVMIntrinsics.cpp
@@ -9,9 +9,14 @@
 namespace Runtime
 {
 	static void causeIntrensicException(Exception::Cause cause) {
-		Platform::immediately_exit(std::make_exception_ptr(Exception{cause, std::vector<std::string>()}));
+		try {
+		   Platform::immediately_exit(std::make_exception_ptr(Exception{cause, std::vector<std::string>()}));
+		}
+		catch (...) {
+			Platform::immediately_exit(std::current_exception());
+		}
 		__builtin_unreachable();
-   }
+	}
 
 	template<typename Float>
 	Float quietNaN(Float value)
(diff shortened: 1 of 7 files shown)


libraries/wasm-jit/Source/Runtime/WAVMIntrinsics.cpp
```diff
@@ -152,18 +157,19 @@ namespace Runtime
 			actualSignature ? asString(actualSignature).c_str() : "nullptr",
 			ipDescription.c_str()
 			);
-		causeException(elementValue == nullptr ? Exception::Cause::undefinedTableElement : Exception::Cause::indirectCallSignatureMismatch);
+		causeIntrensicException(elementValue == nullptr ? Exception::Cause::undefinedTableElement : Exception::Cause::indirectCallSignatureMismatch);
 	}
```
heifner: Don't we need to wrap all these methods try-catch to catch `bad_alloc` or other exceptions that might leak through?

spoonincode: I didn't see any thing in there that could throw, did I miss something maybe? I would tend to agree that it's somewhat of a land mine to leave them naked in case something changes in the future. However, at this point, it's not like we would ever be able to blindly sync with upstream anyways.

heifner: The `std::string` could throw `bad_alloc`.

spoonincode: Okay, while we're probably going down the toilet anyways if that ever happens, I did try to protect the couple cases where that could occur

PR #7054, opened 2019-04-03
Allow cleos to query an API node behind Cloudflare

elmato:
## Change Description

Having _Content-Length_ header as mandatory for a valid http response prevents **cleos** from querying an API node which is behind Cloudflare.

Cloudflare treats **/v1/chain** endpoint as a dynamic resource so it will compress the content and
remove the _Content-Length_ header from the nodeos response.

https://support.cloudflare.com/hc/en-us/articles/200168386-Why-is-my-dynamic-content-being-sent-with-chunked-encoding-

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -90,20 +90,19 @@ namespace eosio { namespace client { namespace http {
             response_content_length = std::stoi(match[1]);
       }
 
-      std::stringstream re;
-      if( response_content_length >= 0 ) {
-         // Write whatever content we already have to output.
+      // Attempt to read the response body using the length indicated by the
+      // Content-length header. If the header was not present just read all available bytes.
+      if( response_content_length != -1 ) {
          response_content_length -= response.size();
-         if (response.size() > 0)
-            re << &response;
-
-         boost::asio::read(socket, response, boost::asio::transfer_exactly(response_content_length));
+         if( response_content_length > 0 )
+            boost::asio::read(socket, response, boost::asio::transfer_exactly(response_content_length));
       } else {
          boost::system::error_code ec;
          boost::asio::read(socket, response, boost::asio::transfer_all(), ec);
          EOS_ASSERT(!ec || ec == boost::asio::ssl::error::stream_truncated, http_exception, "Unable to read http response: ${err}", ("err",ec.message()));
       }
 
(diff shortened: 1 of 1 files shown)


spoonincode:
Do you have some example endpoints? Preferably some with https

elmato:
Sure,

```bash
cleos -u https://api-paper.eosfinex.com/ get code eosfinex
```
vs.
```bash
curl -X POST -d '{"account_name":"eosfinex"}' \
     -H 'Content-Type: application/json' \
     https://api-paper.eosfinex.com/v1/chain/get_raw_code_and_abi
```

programs/cleos/httpc.cpp
```diff
@@ -89,17 +89,22 @@ namespace eosio { namespace client { namespace http {
...
       }
-      EOS_ASSERT(response_content_length >= 0, invalid_http_response, "Invalid content-length response");
 
       std::stringstream re;
-      // Write whatever content we already have to output.
-      response_content_length -= response.size();
-      if (response.size() > 0)
-         re << &response;
+      if( response_content_length >= 0 ) {
+         // Write whatever content we already have to output.
```
spoonincode: Why is this part not required for the non-response_content_length branch?

elmato: The logic used when the _Content-Length_ header was found, was to read the exact amount of bytes indicated in it or to fail (boost::asio::transfer_exactly), so the code calculated the remaining amount of bytes to read based on what was already read.

On the other branch, since the _Content-Length_ header isn't present and we will read all the response until the end there is no need to do that.

elmato: We can put the partial output write outside the condition and always do a two step process to write the output

```c++
auto response_size = response.size();
if (response_size > 0)
  re << &response;

if( response_content_length >= 0 ) {
   response_content_length -= response_size;
 ....
} else { 
 ...
}

re << &response;
return re.str();

```

spoonincode: From what I can tell, the
```
         if (response.size() > 0)
            re << &response;
```
can just simply be removed. Any left over content in `response` will be there when we `re << &response;` in either case

spoonincode:
@elmato can you rebase and reopen a PR against develop? didn't catch that before moving forward

PR #7074, opened 2019-04-05
Boot script: add rex support in boot stage.

evsward:
## Change Description
Add rex support in boot stage.  Convenient for quickly building rex-based chain environment tests locally.

### DIFF
--- a/tutorials/bios-boot-tutorial/bios-boot-tutorial.py
+++ b/tutorials/bios-boot-tutorial/bios-boot-tutorial.py
@@ -26,6 +26,7 @@ systemAccounts = [
     'eosio.stake',
     'eosio.token',
     'eosio.vpay',
+    'eosio.rex'
 ]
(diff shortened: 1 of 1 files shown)


heifner:
Thanks @evsward. We are looking to remove contract specify capabilities from the `eos` repo. As such we are not looking to have the rex contract in eos repo (it lives in the eosio.contracts repo). I know we are not there yet. So in the meantime I think we would entertain taking the addition to the script of the eosio.rex account, but not the loading of the rex contract.

evsward:
Rex system depends on the latest eosio.contracts repo. So we need build the new version of eosio.system contract at first.  And then reboot the eos chain via this script.  I've tested it successfully obtain a chain environment with REX.

heifner:
`eosio.rex` account creation has been added to `release/1.8.x` branch.

PR #7095, opened 2019-04-09
Build script: support new options

mitza-oci:
## Change Description
Enhanced the build script with the ability to specify a prefix directory (instead of assuming HOME) and a pre-built Boost.
Cleaned up the usage message. 
Moved commands that make changes on the filesystem later (after preliminary checks are done).
Removed unsupported -a "Avoid Compiling"

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/eosio_build.sh
+++ b/scripts/eosio_build.sh
@@ -74,7 +74,7 @@ EOT
 }
 
 if [ $# -ne 0 ]; then
-   while getopts ":cdo:s:ahy" opt; do
+   while getopts ":cdo:s:p:b:hy" opt; do
       case "${opt}" in
          o )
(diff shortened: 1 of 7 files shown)


spoonincode:
Most of these changes look good but I'm concerned that the ability to set a custom prefix will mess up `eosio_install.sh`

mitza-oci:
> tried this on a centos7 image with
> `scripts/eosio_build.sh -p /eosroot`
> but failed with
> `/home/spoon/eos/scripts/../scripts/eosio_build_centos.sh: line 247: /home/spoon/bin/cmake: No such file or directory`

Did it print `Installing CMAKE...`?  Should it be using the distro's CMake?

PR #7107, opened 2019-04-10
c++17  and compiler pinning

larryk85:
## Change Description
Update libraries and CMake to build with C++ 17 instead of C++ 14. And pin all noninteractive builds (buildkite builds) to use clang 8.
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


scripts/eosio_build.sh
```diff
@@ -279,11 +290,10 @@ printf "## ENABLE_COVERAGE_TESTING=%s\\n" "${ENABLE_COVERAGE_TESTING}"
 mkdir -p $BUILD_DIR
 cd $BUILD_DIR
 
-$CMAKE -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_CXX_COMPILER="${CXX_COMPILER}" \
-   -DCMAKE_C_COMPILER="${C_COMPILER}" -DCORE_SYMBOL_NAME="${CORE_SYMBOL_NAME}" \
-   -DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=true \
-   -DENABLE_COVERAGE_TESTING="${ENABLE_COVERAGE_TESTING}" -DBUILD_DOXYGEN="${DOXYGEN}" \
+$CMAKE -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_CXX_COMPILER="${CXX}" -DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=false \
```
spoonincode: As far as I can tell this won't actually use llvm's libc++, for example:
```
opt/clang8/bin/clang++ -std=c++17 string_view.cpp -o string_view -v
#include <...> search starts here:
 /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8
 /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/x86_64-linux-gnu/c++/8
 /usr/lib/gcc/x86_64-linux-gnu/8/../../../../include/c++/8/backward
 /usr/local/include
 /home/ubuntu/opt/clang8/lib/clang/8.0.1/include
 /usr/include/x86_64-linux-gnu
 /usr/include
```
so we only get the system lib's `string_view`. Also the linkage is to the system's libc++, i.e.
```
$ ldd string_view
        linux-vdso.so.1 =>  (0x00007fff2a3e5000)
        libstdc++.so.6 => /usr/lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007fd0ec3e1000)
```

If we're going to pin the compiler, don't we want the stdc++ lib to be part of that pin?

CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ include( InstallDirectoryPermissions )
 include( MASSigning )
 
 set( BLOCKCHAIN_NAME "EOSIO" )
-set( CMAKE_CXX_STANDARD 14 )
+set( CMAKE_CXX_STANDARD 17 )
```
spoonincode: Also need to change the compiler check that's a few lines down from here

scripts/eosio_build_amazon.sh
```diff
@@ -248,6 +248,39 @@ if [ $? -ne 0 ]; then exit -1; fi
 cd ..
 printf "\\n"
 
+if [ $BUILD_CLANG8 ]; then
```
spoonincode: I believe this check is true no matter `BUILD_CLANG8` contains true or false

scripts/eosio_build.sh
```diff
@@ -179,6 +181,31 @@ if [ $STALE_SUBMODS -gt 0 ]; then
    exit 1
 fi
 
+BUILD_CLANG8=false
+if [ $NONINTERACTIVE -eq 0 ]; then
+   printf "#include <iostream>\nint main(){ std::cout << \"Hello, World!\" << std::endl; }" &> $TEMP_DIR/test.cpp
+   `c++ -c -std=c++17 $TEMP_DIR/test.cpp -o $TEMP_DIR/test.o &> /dev/null`
+   if [ $? -ne 0 ]; then
+      `CXX -c -std=c++17 $TEMP_DIR/test.cpp -o $TEMP_DIR/test.o &> /dev/null`
```
spoonincode: was this supposed to be `$CXX` instead of `CXX`?

scripts/eosio_build.sh
```diff
@@ -179,6 +181,31 @@ if [ $STALE_SUBMODS -gt 0 ]; then
    exit 1
 fi
 
+BUILD_CLANG8=false
+if [ $NONINTERACTIVE -eq 0 ]; then
+   printf "#include <iostream>\nint main(){ std::cout << \"Hello, World!\" << std::endl; }" &> $TEMP_DIR/test.cpp
```
spoonincode: This check won't work as is; even the gcc 5.5.0 on ubuntu 16.06 passes validation here (and then the build fails when the cmake 6.0 or greater check fails).

scripts/eosio_build.sh
```diff
@@ -71,6 +71,8 @@ export BOOST_LINK_LOCATION=${OPT_LOCATION}/boost
 export LLVM_VERSION=release_40
 export LLVM_ROOT=${OPT_LOCATION}/llvm
 export LLVM_DIR=${LLVM_ROOT}/lib/cmake/llvm
+export CLANG8_ROOT=${OPT_LOCATION}/clang8
+export PINNED_COMPILER_VERSION=release_80
```
spoonincode: I'm pretty sure this isn't a pin to a released version. The current compiler version I have after running the script is
`clang version 8.0.1 (https://git.llvm.org/git/clang.git a03da8be08a208122e292016cb6cea1f30229677)`

scripts/eosio_build_ubuntu.sh
```diff
@@ -163,7 +162,7 @@ if [ "${BOOSTVERSION}" != "${BOOST_VERSION_MAJOR}0${BOOST_VERSION_MINOR}0${BOOST
 	&& tar -xjf boost_$BOOST_VERSION.tar.bz2 \
 	&& cd $BOOST_ROOT \
 	&& ./bootstrap.sh --prefix=$BOOST_ROOT \
-	&& ./b2 -q -j"${JOBS}" install \
+	&& ./b2 -q -j2 install \
```
spoonincode: Lost the `-j"${JOBS}"` here

scripts/eosio_build_amazon.sh
```diff
@@ -12,7 +12,7 @@ if [[ "${OS_NAME}" == "Amazon Linux AMI" ]]; then # Amazonlinux1
 	DEP_ARRAY=( 
 		sudo procps util-linux which gcc72 gcc72-c++ autoconf automake libtool make doxygen graphviz \
 		bzip2 bzip2-devel openssl-devel gmp gmp-devel libstdc++72 python27 python27-devel python36 python36-devel \
-		libedit-devel ncurses-devel swig wget file libcurl-devel libusb1-devel
+		libedit-devel ncurses-devel swig curl -LO file libcurl-devel libusb1-devel
```
spoonincode: breaks amz1 but may not matter as it will be removed soon

scripts/eosio_build_ubuntu.sh
larryk85: Oof.

scripts/eosio_build_amazon.sh
larryk85: Search and replace badness, will fix.

spoonincode:
I'm fairly surprised and concerned with the complexity this grew to vs the original change. Of most technical concern is needing to build seemingly random versions of gmp, mpc, gcc & gnu's stdc++ library (why was a two year old version of 7.1.0 picked vs something new? AFAIK gcc7's c++ lib isn't even 100% c++17 compliant because it still places `filesystem` in `experimental::`. Also, the ucontext change needed was fixed already in gcc8). Because we have to maintain this toolchain for long in to the future, more complexity seems to imply more fragility.

Also IMO linking statically with gnu's stdc++ via LLVM needs to involve legal approval. The runtime license exception isn't entirely clear to me if it would apply to this usage.

I went to investigate in more detail why the original approach -- just using the c++lib provided by llvm -- would not work. I didn't run in to any remarkable issues, but to be clear I only tried on Ubuntu 18.04, with LLVM 8.0 binaries downloaded from llvm.org, on develop (no c++17), and no mongo.

I needed a config for boost building:
```
$ cat ~/boost-config.jam 
using clang : 8.0 : /home/ubuntu/toolchain/bin/clang++ : <cxxflags>"-std=c++17 -nostdinc++ -I/home/ubuntu/toolchain/include/c++/v1" ;
```
and a cmake toolchain file:
```
$ cat toolchain.cmake 
set(CMAKE_C_COMPILER /home/ubuntu/toolchain/bin/clang)
set(CMAKE_CXX_COMPILER /home/ubuntu/toolchain/bin/clang++)
set(CMAKE_CXX_STANDARD_INCLUDE_DIRECTORIES /home/ubuntu/toolchain/include/c++/v1)
set(CMAKE_CXX_FLAGS_INIT "-nostdinc++")
set(CMAKE_EXE_LINKER_FLAGS_INIT "-nostdlib++")
set(CMAKE_SHARED_LINKER_FLAGS_INIT "-nostdlib++")
set(CMAKE_MODULE_LINKER_FLAGS_INIT "-nostdlib++")
link_libraries("/home/ubuntu/toolchain/lib/libc++.a" "/home/ubuntu/toolchain/lib/libc++abi.a")
```

Then I built & installed boost
```
$ ./b2 --toolset=clang-8.0  --user-config=$HOME/boost-config.jam --prefix=$HOME/boost link=static threading=multi install --with-iostreams --with-date_time --with-filesystem --with-system --with-program_options --with-chrono --with-test
```
Then I built & installed LLVM4
```
$ cmake -DCMAKE_INSTALL_PREFIX=$HOME/llvmbuild -DLLVM_TARGETS_TO_BUILD=host -DLLVM_BUILD_TOOLS=false -DLLVM_ENABLE_RTTI=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_TOOLCHAIN_FILE=$HOME/toolchain.cmake ..
$ make install
```
Then I built eosio
```
$ cmake -DCMAKE_TOOLCHAIN_FILE=$HOME/toolchain.cmake -DCMAKE_BUILD_TYPE=Debug -DBOOST_ROOT=$HOME/boost -DCMAKE_PREFIX_PATH=$HOME/llvmbuild ..
$ make
```
I did need a little modification to eosio CMakeLists.txts, but these were buggy things anyways
```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 17c3df724..01d876662 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,9 +113,7 @@ FIND_PACKAGE(Boost 1.67 REQUIRED COMPONENTS
     iostreams)
 
 # Some new stdlibc++s will #error on <experimental/string_view>; a problem for boost pre-1.69
-if( APPLE AND UNIX )
    add_definitions(-DBOOST_ASIO_DISABLE_STD_EXPERIMENTAL_STRING_VIEW)
-endif()
 
 set(THREADS_PREFER_PTHREAD_FLAG 1)
 find_package(Threads)
@@ -159,11 +157,11 @@ else( WIN32 ) # Apple AND Linux
     if( APPLE )
         # Apple Specific Options Here
         message( STATUS "Configuring EOSIO on OS X" )
-        set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -Wall -Wno-deprecated-declarations" )
+        set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wno-deprecated-declarations" )
     else( APPLE )
         # Linux Specific Options Here
         message( STATUS "Configuring EOSIO on Linux" )
-        set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -Wall" )
+        set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall" )
         if ( FULL_STATIC_BUILD )
           set( CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -static-libstdc++ -static-libgcc")
         endif ( FULL_STATIC_BUILD )
Entering 'libraries/appbase'
diff --git a/CMakeLists.txt b/CMakeLists.txt
index eee8d3c..f2be62e 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,11 +23,11 @@ set( Boost_USE_STATIC_LIBS ON CACHE STRING "ON or OFF" )
 if( APPLE )
   # Apple Specific Options Here
   message( STATUS "Configuring AppBase on OS X" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=c++14 -stdlib=libc++ -Wall -Wno-conversion -Wno-deprecated-declarations" )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -stdlib=libc++ -Wall -Wno-conversion -Wno-deprecated-declarations" )
 else( APPLE )
   # Linux Specific Options Here
   message( STATUS "Configuring AppBase on Linux" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=c++14 -Wall" )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14 -Wall" )
   set( rt_library rt )
   set( pthread_library pthread)
   if ( FULL_STATIC_BUILD )
Entering 'libraries/chainbase'
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 23f6547..83ffd72 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,11 +28,11 @@ SET(PLATFORM_LIBRARIES)
 if( APPLE )
   # Apple Specific Options Here
   message( STATUS "Configuring ChainBase on OS X" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=c++11 -stdlib=libc++ -Wall -Wno-conversion" )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -stdlib=libc++ -Wall -Wno-conversion" )
 else( APPLE )
   # Linux Specific Options Here
   message( STATUS "Configuring ChainBase on Linux" )
-  set( CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -std=c++11 -Wall" )
+  set( CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall" )
   set( rt_library rt )
   set( pthread_library pthread)
   if ( FULL_STATIC_BUILD )
```
It looks like everything works (parallelizable tests pass); and the executables are linked as expected:
```
$ ldd programs/nodeos/nodeos 
	linux-vdso.so.1 (0x00007fffea1b5000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f0327807000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f03275e8000)
	libtinfo.so.5 => /lib/x86_64-linux-gnu/libtinfo.so.5 (0x00007f03273be000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f03271a1000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f0326e03000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f0326bfb000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f03269e3000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f03265f2000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f0327a0b000)
```
That said, there is a little wonkyness in how the toolchain.cmake worked. For one, `-nostdinc++` gets used even when linking causing an unused option warning. And two, the two static libraries are included twice on the linker command line. I suspect doing `link_libraries()` in the toolchain file is probably not a good idea; there might be a better way of accomplishing both these issues I didn't research much more. (also considering our expertise in customizing llvm -- EOSIO.cdt -- it may make more sense to just patch something in the clang driver to handle these options automatically)

I think minimally we need to have a fully compliant c++17 library (which means gcc8) and get legal approval. Even then I'm not personally comfortable approving the change with the added complexity vs just using LLVM's libc++, but I won't be against it.

spoonincode:
ah I see though that on macos we would just using the c++ library included with xcode. Maybe it's not a big problem to lack full c++17 support on our linux builds then, as xcode 10.2 also is not fully c++17 (for one we'd never be able to use std::filesystem, dunno what other surprises may lurk)

Still think it's important to get the licensing question resolved though

larryk85:
> ah I see though that on macos we would just using the c++ library included with xcode. Maybe it's not a big problem to lack full c++17 support on our linux builds then, as xcode 10.2 also is not fully c++17 (for one we'd never be able to use std::filesystem, dunno what other surprises may lurk)
> 
> Still think it's important to get the licensing question resolved though

So a few points to be made.  I just now try your approach and I'm not sure what the difference is in terms of the base setup, but I am still getting undefined symbols with libcxx.  I'm sure there is a way around what issues I am seeing since you got it to work on your machine, but I don't see the need unless legal says it's a no go on the static linkage of libstdc++.  The reason I say that is that the versions of gcc/gmp/mpfr/mpc are not random and were not picked out of a bag.  These are the exact versions that are from the build log for ubuntu 18.04s package of gcc7, which is what you get when you apt-get install build-essentials.  I picked this because we know what the performance of this is already, our build scripts have been inadvertently building with this configuration for a while, but with clang 4 and not 8.  So any performance impact should be an increase in performance and not a possible decrease with going with a version that has not been tested.  I would be interested in a matrix of gcc7, gcc8, libcxx to see what impacts they have, but I don't have the bandwidth and we don't have the time to do so.  I feel more comfortable with sticking with a configuration that we know works well and we can change that later if the need is high enough.

scripts/eosio_build_amazon.sh
```diff
@@ -248,6 +248,105 @@ if [ $? -ne 0 ]; then exit -1; fi
...
+      && cd libcxx && git checkout $PINNED_COMPILER_LIBCXX_COMMIT && cd ../ \
+      && git clone --depth 1 --single-branch --branch $PINNED_COMPILER_BRANCH https://git.llvm.org/git/libcxxabi.git \
+      && cd libcxxabi && git checkout $PINNED_COMPILER_LIBCXXABI_COMMIT && cd ../ \
+      && git clone --depth 1 --single-branch --branch $PINNED_COMPILER_BRANCH https://git.llvm.org/git/libunwind.git \
+      && cd libunwind && git checkout $PINNED_COMPILER_LIBUNWIND_COMMIT && cd ../ \
+      && git clone --depth 1 --single-branch --branch $PINNED_COMPILER_BRANCH https://git.llvm.org/git/compiler-rt.git \
+      && cd compiler-rt && git checkout $PINNED_COMPILER_COMPILER_RT_COMMIT && cd ../ \
+      && cd ${TMP_LOCATION}/clang8 \
+      && mkdir build && cd build \
+      && $CMAKE -G "Unix Makefiles" -DCMAKE_INSTALL_PREFIX="${CLANG8_ROOT}" -DLLVM_BUILD_EXTERNAL_COMPILER_RT=ON -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_ENABLE_LIBCXX=ON -DLLVM_ENABLE_RTTI=ON -DLLVM_INCLUDE_DOCS=OFF -DLLVM_OPTIMIZED_TABLEGEN=ON -DLLVM_TARGETS_TO_BUILD=all -DCMAKE_BUILD_TYPE=Release .. \
```
spoonincode: > -DLLVM_TARGETS_TO_BUILD=all

Is it worth making this `-DLLVM_TARGETS_TO_BUILD=host` to speed up compile for users or not?

spoonincode:
When I run --version on the binaries I got from LLVM I see
`clang version 8.0.0 (tags/RELEASE_800/final)`
But for the version we compile
`clang version 8.0.1 (https://git.llvm.org/git/clang.git a03da8be08a208122e292016cb6cea1f30229677) (https://git.llvm.org/git/llvm.git 18e41dc964f916504ec90dba523826ac74d235c4)`

Why are we not using the released 8.0.0?

CMakeLists.txt
```diff
@@ -121,6 +121,15 @@ set(THREADS_PREFER_PTHREAD_FLAG 1)
 find_package(Threads)
 link_libraries(Threads::Threads)
 
+if(${EOSIO_PIN_COMPILER})
+   if(NOT APPLE AND UNIX)
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -nostdinc++ -nostdlib++ -I${LIBSTDCPP_DIR}/include/c++/7.1.0")
+      set(CMAKE_CXX_STANDARD_LIBRARIES "${LIBSTDCPP_DIR}/lib64/libstdc++.a")
+   else()
+      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++")
```
spoonincode: This is actually a noop, because CMAKE_CXX_FLAGS gets stomped on below.

spoonincode:
The current configuration on macos seems particularly dangerous: LLVM's libc++ headers are used for compilation but then the executable is linked with the system provided libc++. There is no guarantee these "match up".

larryk85:
> When I run --version on the binaries I got from LLVM I see
> `clang version 8.0.0 (tags/RELEASE_800/final)`
> But for the version we compile
> `clang version 8.0.1 (https://git.llvm.org/git/clang.git a03da8be08a208122e292016cb6cea1f30229677) (https://git.llvm.org/git/llvm.git 18e41dc964f916504ec90dba523826ac74d235c4)`
> 
> Why are we not using the released 8.0.0?

Because I am pulling this from github and it doesn't have any tags.  I assume you are getting yours from the ftp? We can do that, but I assumed that it would be nice to use the git version so that we could apply any patches that need to be applied and not wait on a "release" to be made by LLVM.

spoonincode:
I didn't look at the log to see what (if any) measurable difference there was between 8.0.0 and "8.0.1 -- but not really 8.0.1".  I'm not sure how clean LLVM keeps their trees such that would we be getting something intermediary.

larryk85:
> I didn't look at the log to see what (if any) measurable difference there was between 8.0.0 and "8.0.1 -- but not really 8.0.1". I'm not sure how clean LLVM keeps their trees such that would we be getting something intermediary.

There is no difference other than PRs that bump version or add meta data to files that have no impact.

CMakeLists.txt
```diff
@@ -113,11 +113,13 @@ FIND_PACKAGE(Boost 1.67 REQUIRED COMPONENTS
     iostreams)
 
 # Some new stdlibc++s will #error on <experimental/string_view>; a problem for boost pre-1.69
+add_definitions(-DBOOST_ASIO_DISABLE_STD_EXPERIMENTAL_STRING_VIEW)
 if( APPLE AND UNIX )
    add_definitions(-DBOOST_ASIO_DISABLE_STD_EXPERIMENTAL_STRING_VIEW)
 endif()
```
spoonincode: these 3 lines can go now

scripts/eosio_build.sh
```diff
@@ -184,6 +215,64 @@ if [ $STALE_SUBMODS -gt 0 ]; then
...
+      fi
+   fi
+else
+   ### gcc version 7
+   if [[ `c++ -dumpversion | cut -d '.' -f 1` -lt 7 ]]; then
+      NO_CPP17=true
+   fi
+fi
+
+if $PIN_COMPILER; then
```
spoonincode: always true

scripts/eosio_build.sh
```diff
@@ -291,15 +364,25 @@ printf "======================= Starting EOSIO Build =======================\\n"
...
 cd $BUILD_DIR
 
-$CMAKE -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_CXX_COMPILER="${CXX_COMPILER}" \
-   -DCMAKE_C_COMPILER="${C_COMPILER}" -DCORE_SYMBOL_NAME="${CORE_SYMBOL_NAME}" \
-   -DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=true \
-   -DENABLE_COVERAGE_TESTING="${ENABLE_COVERAGE_TESTING}" -DBUILD_DOXYGEN="${DOXYGEN}" \
-   -DCMAKE_PREFIX_PATH=$PREFIX \
-   -DCMAKE_INSTALL_PREFIX=$OPT_LOCATION/eosio $LOCAL_CMAKE_FLAGS "${REPO_ROOT}"
+
+if $PIN_COMPILER; then
```
spoonincode: always true

scripts/eosio_build.sh
```diff
@@ -291,15 +364,25 @@ printf "======================= Starting EOSIO Build =======================\\n"
...
+   $CMAKE -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_TOOLCHAIN_FILE=$BUILD_DIR/pinned_toolchain.cmake \
+      -DCORE_SYMBOL_NAME="${CORE_SYMBOL_NAME}" \
+      -DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=false \
```
spoonincode: doesn't seem to respect the build script's command line option for enabling mongo

scripts/eosio_build_ubuntu.sh
```diff
@@ -155,102 +162,198 @@ if [ $? -ne 0 ]; then exit -1; fi
...
+   else
+      printf " - Clang 8 found @ ${CLANG8_ROOT}.\\n"
+   fi
+   if [ $? -ne 0 ]; then exit -1; fi
 
+   printf "\\n"
 
-printf "\\n"
+   printf "Checking LLVM 4 installation...\\n"
+   if [ ! -d $OPT_LOCATION/llvm4 ] || [ $FORCE_BUILD ]; then
```
spoonincode: don't we only want to build this when not pinning compiler? otherwise it won't work right -- because it's using the toolchain file

spoonincode: n/m I see it's good

larryk85: Huh? We only want to build when pinning because we are using a toolchain file.  We are pinning to libcxx for all of pinned builds.

scripts/eosio_build_amazon.sh
```diff
@@ -140,113 +142,201 @@ if [ $? -ne 0 ]; then exit -1; fi
 
 printf "\\n"
 
+if $PIN_COMPILER; then
```
spoonincode: always true

scripts/eosio_build_centos.sh
```diff
@@ -192,112 +194,200 @@ printf "\\n"
...
-	&& tar -xzf mongo-cxx-driver-r${MONGO_CXX_DRIVER_VERSION}.tar.gz \
-	&& cd mongo-cxx-driver-r$MONGO_CXX_DRIVER_VERSION/build \
-	&& $CMAKE -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$PREFIX .. \
-	&& make -j"${JOBS}" VERBOSE=1 \
-	&& make install \
-	&& cd ../.. \
-	&& rm -f mongo-cxx-driver-r$MONGO_CXX_DRIVER_VERSION.tar.gz \
-	|| exit 1
-	printf " - MongoDB C++ driver successfully installed @ ${MONGO_CXX_DRIVER_ROOT}.\\n"
+if $PIN_COMPILER; then
```
spoonincode: always true

scripts/eosio_build_darwin.sh
```diff
@@ -262,6 +263,48 @@ fi
 cd ..
 printf "\\n"
 
+if $PIN_COMPILER; then
```
spoonincode: always true

scripts/eosio_build_ubuntu.sh
```diff
@@ -155,102 +162,198 @@ if [ $? -ne 0 ]; then exit -1; fi
 
 printf "\\n"
 
+if $PIN_COMPILER; then
```
spoonincode: always true

scripts/eosio_build.sh
```diff
@@ -184,6 +215,64 @@ if [ $STALE_SUBMODS -gt 0 ]; then
...
+      NO_CPP17=true
+   fi
+fi
+
+if [ $PIN_COMPILER ]; then
+   BUILD_CLANG8=true
+   CPP_COMP=${OPT_LOCATION}/clang8/bin/clang++
+   CC_COMP=${OPT_LOCATION}/clang8/bin/clang
+   PIN_COMPILER_CMAKE="-DEOSIO_PIN_COMPILER=1 -DLIBSTDCPP_DIR=${OPT_LOCATION}/gcc -DLIBCPP_DIR=${OPT_LOCATION}/clang8/"
+elif $NO_CPP17; then
```
spoonincode: always true

scripts/eosio_build.sh
```diff
@@ -291,15 +364,25 @@ printf "======================= Starting EOSIO Build =======================\\n"
...
+   $CMAKE -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_TOOLCHAIN_FILE=$BUILD_DIR/pinned_toolchain.cmake \
+      -DCORE_SYMBOL_NAME="${CORE_SYMBOL_NAME}" \
+      -DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=$ENABLE_MONGO \
+      -DENABLE_COVERAGE_TESTING="${ENABLE_COVERAGE_TESTING}" -DBUILD_DOXYGEN="${DOXYGEN}" \
+      -DCMAKE_PREFIX_PATH=$PREFIX -DCMAKE_PREFIX_PATH=$OPT_LOCATION/llvm4\
+      -DCMAKE_INSTALL_PREFIX=$OPT_LOCATION/eosio $LOCAL_CMAKE_FLAGS -DEOSIO_PIN_COMPILER=1 "${REPO_ROOT}"
+else
+   $CMAKE -DCMAKE_BUILD_TYPE="${CMAKE_BUILD_TYPE}" -DCMAKE_CXX_COMPILER="${CXX}" \
+      -DCMAKE_C_COMPILER="${CC}" -DCORE_SYMBOL_NAME="${CORE_SYMBOL_NAME}" \
+      -DOPENSSL_ROOT_DIR="${OPENSSL_ROOT_DIR}" -DBUILD_MONGO_DB_PLUGIN=$ENABLE_MONGO \
```
spoonincode: I can't find where `ENABLE_MONGO` is ever defined

scripts/eosio_build_centos.sh
```diff
@@ -192,112 +194,200 @@ printf "\\n"
...
+   fi
+   if [ $? -ne 0 ]; then exit -1; fi
+
+   printf "\\n"
+
+   printf "Checking LLVM 4 installation...\\n"
+   if [ ! -d $OPT_LOCATION/llvm4 ] || [ $FORCE_BUILD ]; then
+      printf "Installing LLVM 4...\\n"
+      cd $SRC_LOCATION \
+      && git clone https://github.com/llvm-mirror/llvm --single-branch --branch $LLVM_VERSION && cd llvm \
```
spoonincode: `FORCE_BUILD` is documented as "Force rebuilding of dependencies" but there are many places throughout the scripts where that doesn't actually work. This is one case, where when the clone is run a second time git errors out with `fatal: destination path 'llvm' already exists and is not an empty directory.` There are also places where tar just untars over an existing directory. It seems like for `FORCE_BUILD` to do what it says on the tin all these places needed to be further cleaned up to remove the source dir before doing the clone/untar/whatevs

spoonincode:
I am unable to determine why building zlib is required. I made a branch where I removed that:
https://github.com/EOSIO/eos/compare/feature/cpp17_compiler_pinning...cpp17_remove_zlib

and the builds seemed to be fine:
https://buildkite.com/EOSIO/eosio/builds/11071

(mac build failed for something unrelated; I didn't change mac builds)

larryk85:
> I am unable to determine why building zlib is required. I made a branch where I removed that:
> [feature/cpp17_compiler_pinning...cpp17_remove_zlib](https://github.com/EOSIO/eos/compare/feature/cpp17_compiler_pinning...cpp17_remove_zlib)
> 
> and the builds seemed to be fine:
> https://buildkite.com/EOSIO/eosio/builds/11071
> 
> (mac build failed for something unrelated; I didn't change mac builds)

That is fine, we can change it after we get this merged.  I really don't want to keep tinkering with this right now.  It builds fine currently, but it might be a little bit aggressive in terms of dependencies.

larryk85:
> Looks like on macos we don't build our own llvm4 when pinned? Seems potentially dangerous going forward but probably not an issue for now

Shouldn't be dangerous, as the mac build by default is libcxx and we are statically linking libcxx so even though we don't build llvm4 I don't see it as an issue.  Secondly, the plan is that llvm4 is a temporary dependency at this point.  Either we replace WAVM or port it up to use llvm 8 which we are already building.

PR #7108, opened 2019-04-10
Add optional error_code to transaction and action traces

### LINKED ISSUE #6898: Add optional error code to action trace
Adding onto the major changes to transaction and action traces described in #6897, this will be a good opportunity (since transaction trace changes force replay for history_plugin users and may also have an impact on the definitions of the still alpha state history plugin) to add another value piece of information to the action trace: the error code.

The `eosio_assert_code` intrinsic allows contracts to signal the error with a simple `uint64_t` code rather than a full string. Error message maps in the ABI could be used to associate this code with a string to be included in the error message that is thrown (or included in action trace through maybe null `except` field). However, regardless of whether such a mapping exists, if an error code is thrown, the error code value should be included in the action trace.

So at a minimum a new field `error_code` of type `optional<uint64_t>` should be added to both `base_action_trace` and `transaction_trace`, and they should be set to the appropriate value if an action fails due to a call to `eosio_assert_code`. It would be good to also make the modifications to nodeos so that the exception thrown (and stored in the `except` field) includes the string associated with the error code in the contract's ABI (if present).

arhag:
## Change Description

Resolves #6898.

Rename `parent_action_ordinal` to `closest_unnotified_ancestor_action_ordinal` and remove it from the traces stored by the state history plugin.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

The new `parent_action_ordinal` field from PR #7044 has been renamed to `closest_unnotified_ancestor_action_ordinal`.

New optional field `error_code` has been added to `action_trace` and `transaction_trace`. This field should always be empty if the `except` field in the same object is empty. However, it is possible for `except` to be non-empty but `error_code` to be empty (this will be the case for any errors that were not due to the `eosio_assert_code` intrinsic). If `error_code` is non-empty, it means the error was due to an assertion failure on a call to `eosio_assert_code`, and the value will be the error code passed into the `eosio_assert_code` intrinsic.


plugins/state_history_plugin/state_history_plugin_abi.cpp
```diff
@@ -95,15 +95,15 @@ extern const char* const state_history_plugin_abi = R"({
             "name": "action_trace_v0", "fields": [
                 { "name": "action_ordinal", "type": "varuint32" },
                 { "name": "creator_action_ordinal", "type": "varuint32" },
-                { "name": "parent_action_ordinal", "type": "varuint32" },
                 { "name": "receipt", "type": "action_receipt?" },
```
heifner: What about `closest_unnotified_ancestor_action_ordinal` ?

libraries/chain/controller.cpp
```diff
@@ -2983,6 +2997,30 @@ bool controller::all_subjective_mitigations_disabled()const {
...
+
+   try {
+      error_code = std::strtoull( msg.c_str() + pos, nullptr, 10 );
+   } catch( ... ) {
+      return {};
+   }
+
+   return error_code;
+}
+
```
heifner: Instead of this, can we either manually derive `eosio_assert_code_exception` from `action_validate_exception` or create a new macro that adds an `error_code` attribute to the exception so we can just grab it directly instead of this string parsing.

plugins/state_history_plugin/state_history_plugin_abi.cpp
arhag: We decided not to include it in state_history_plugin since we don't want to encourage clients to deal with the ambiguous execution tree.

PR #7111, opened 2019-04-11
Implement GET_SENDER protocol feature

### LINKED ISSUE #7028: Consensus protocol upgrade to get the sending account of an inline action
## Background

A contract may wish to simply notify another contract of some event (but not via `require_recipient`). Inline actions are a good way of doing this. However, if the sender does not wish to allow the receiver to bill the sender for any RAM, it must keep the `authorization` vector of the inline action empty. But this creates a spoofing problem. Another contract (or just a signed transaction) can send that action while pretending to be from some other contract.

If a contract could use an intrinsic to get the account that an inline action was sent from, it could know with certainty whether the message was authentic without relying on the authorization system that opens up the issue of RAM billing for the sender. 

## Consensus protocol upgrade feature

This protocol feature (codename: `GET_SENDER`) allows contracts to determine which account is the sender of an inline action. 

### New intrinsic `get_sender`

A new intrinsic `account_name get_sender()` should be added to the `system_api`. The C intrinsic for the WebAssembly side would have a signature such as `uint64_t get_sender()` where the return type is the `uint64_t` raw representation of the account name of the sender of the inline action. If the current action is not an inline action (for example the top-level actions of a transaction), then `get_sender` will return 0.

taokayan:
## Change Description

Resolves #7028.

This PR adds support for the `GET_SENDER` protocol feature which allows contracts to determine which account is the sender of an inline action.

This PR introduces a new unit test `protocol_feature_tests/get_sender_test` to test the behavior of this new protocol feature. The `get_sender_test` test contract was added to meet the needs of the new unit test.

## Consensus Changes
- [x] Consensus Changes

Makes the `GET_SENDER` protocol feature available. See above description and #7028 for details.

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tbfleming:
After this change, there will be 2 event systems:
* `require_recipient`
* inline action, with receiver verifying the sender using `get_sender`

This brings up a question: how should they interact?

* Imagine an inline action with no args that's a simple ack.
* Intended behavior: if contract A sends an ack to contract B, contract B can verify authenticity by only looking at `get_sender`
* Setup:
   * A sends an ack to contract C
   * C uses `require_recipient` to send it to B
   * B uses `get_sender`
* Options:
   * `get_sender` returns A: B incorrectly assumes the ack was for it
   * `get_sender` returns 0: B doesn't fall for the attack
   * `get_sender` returns C: B doesn't fall for the attack

arhag:
> * Setup:
>   
>   * A sends an ack to contract C
>   * C uses `require_recipient` to send it to B
>   * B uses `get_sender`
> * Options:
>   
>   * `get_sender` returns A: B incorrectly assumes the ack was for it
>   * `get_sender` returns 0: B doesn't fall for the attack
>   * `get_sender` returns C: B doesn't fall for the attack

This PR goes with the third option where it would return C in the situation described above.

PR #7118, opened 2019-04-12
Make executable names configurable

conr2d:
## Change Description
`CMakeLists.txt` has options for changing the name of executables like nodeos, keosd or cleos, but there are hard-coded strings in their source code. This PR makes executable names configurable by setting path to changed name or applying new name in help text.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/eosio-launcher/main.cpp
+++ b/programs/eosio-launcher/main.cpp
@@ -1558,7 +1558,7 @@ launcher_def::launch (eosd_def &instance, string &gts) {
         install_path = specific_nodeos_installation_paths[node_num] + "/";
      }
   }
-  string eosdcmd = install_path + "programs/nodeos/" + string(node_executable_name);
+  string eosdcmd = install_path + "programs/nodeos/" + string(node_executable_name) + " ";
   if (skip_transaction_signatures) {
     eosdcmd += "--skip-transaction-signatures ";
   }
--- a/programs/eosio-launcher/main.cpp
+++ b/programs/eosio-launcher/main.cpp
@@ -48,6 +48,7 @@ using bpo::options_description;
 using bpo::variables_map;
 using public_key_type = fc::crypto::public_key;
 using private_key_type = fc::crypto::private_key;
+using namespace eosio::launcher::config;
 
 const string block_dir = "blocks";
 const string shared_mem_dir = "state";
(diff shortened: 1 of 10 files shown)


programs/keosd/main.cpp
```diff
@@ -48,7 +48,7 @@ int main(int argc, char** argv)
       if(!app().initialize<wallet_plugin, wallet_api_plugin, http_plugin>(argc, argv))
          return -1;
       auto& http = app().get_plugin<http_plugin>();
-      http.add_handler("/v1/keosd/stop", [](string, string, url_response_callback cb) { cb(200, fc::variant(fc::variant_object())); std::raise(SIGTERM); } );
+      http.add_handler("/v1/" + keosd::config::key_store_executable_name + "/stop", [](string, string, url_response_callback cb) { cb(200, fc::variant(fc::variant_object())); std::raise(SIGTERM); } );
```
spoonincode: need to change the corresponding line in httpc.hpp too

programs/eosio-launcher/main.cpp
```diff
@@ -1557,7 +1558,7 @@ launcher_def::launch (eosd_def &instance, string &gts) {
         install_path = specific_nodeos_installation_paths[node_num] + "/";
      }
   }
-  string eosdcmd = install_path + "programs/nodeos/nodeos ";
+  string eosdcmd = install_path + "programs/nodeos/" + string(node_executable_name);
```
spoonincode: you lost a trailing space here, and it might be why tests are failing

conr2d: Thank you, I applied it.

programs/eosio-launcher/main.cpp
```diff
@@ -490,18 +491,18 @@ launcher_def::set_options (bpo::options_description &cfg) {
...
+    ("skip-signature", bpo::bool_switch(&skip_transaction_signatures)->default_value(false), (string(node_executable_name) + " does not require transaction signatures.").c_str())
+    (node_executable_name, bpo::value<string>(&eosd_extra_args), ("forward " + string(node_executable_name) + " command line argument(s) to each instance of " + string(node_executable_name) + ", enclose arg(s) in quotes").c_str())
+    ("specific-num", bpo::value<vector<uint>>()->composing(), ("forward " + string(node_executable_name) + " command line argument(s) (using \"--specific-" + string(node_executable_name) + "\" flag) to this specific instance of " + string(node_executable_name) + ". This parameter can be entered multiple times and requires a paired \"--specific-" + string(node_executable_name) +"\" flag each time it is used").c_str())
+    (("specific-" + string(node_executable_name)).c_str(), bpo::value<vector<string>>()->composing(), ("forward " + string(node_executable_name) + " command line argument(s) to its paired specific instance of " + string(node_executable_name) + "(using \"--specific-num\"), enclose arg(s) in quotes").c_str())
```
spoonincode: unfortunately changing these config names will mean that the python test scripts will not operate when node_executable_name is not 'nodeos'

maybe these two config items should be renamed to something else that doesn't include "nodeos"? (to avoid having to preprocess all the various python scripts with whatever is configured)

conr2d: I found the scripts you mentioned, and I guess there can be another cases which contain hard-coded these options. I don't like this way much, but what about adding new option like `-node-exec`, but also leaving `-nodeos` for backward compatibility?

spoonincode: come to think of it, those scripts won't work before your change anyways because of hardcoded "nodeos", "keosd", etc in the scripts too. Would be nice to take care of that in the future, but I think PR is ready as-is

PR #7117, opened 2019-04-12
store all wasm code in a separate index

spoonincode:
Store all WASM code in a separate index outside of the account index. This allows for deduplication of identical code in memory and will also allow greater accuracy in pruning the WASM instantiation cache because reference counts are updated via the DB

Leaving this as draft at the moment because:
* state_history needs attention
* Unsure what to do about vm_type/vm_version

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #7134, opened 2019-04-16
store all wasm code in a separate index

spoonincode:
## Change Description

Store all WASM code in a separate index outside of the account index. This allows for deduplication of identical code in memory and will also allow greater accuracy in pruning the WASM instantiation cache because reference counts are updated via the DB

Replaces #7117

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/eosio_contract.cpp
+++ b/libraries/chain/eosio_contract.cpp
@@ -12,6 +12,7 @@
 #include <eosio/chain/exceptions.hpp>
 
 #include <eosio/chain/account_object.hpp>
+#include <eosio/chain/code_object.hpp>
 #include <eosio/chain/permission_object.hpp>
 #include <eosio/chain/permission_link_object.hpp>
 #include <eosio/chain/global_property_object.hpp>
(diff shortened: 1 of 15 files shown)


libraries/chain/eosio_contract.cpp
```diff
@@ -167,19 +167,15 @@ void apply_eosio_setcode(apply_context& context) {
...
          db.modify(*new_code_entry, [&](code_object& o) {
-            code_id = o.id;
             ++o.code_ref_count;
          });
       } else {
          db.create<code_object>([&](code_object& o) {
-            code_id = o.id;
             o.code_hash = code_hash;
             o.code.assign(act.code.data(), code_size);
             o.code_ref_count = 1;
```
arhag: `o.first_block_used` in line below should be set to `context.control.head_block_num() + 1`.

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_serialization.hpp
```diff
@@ -117,22 +117,9 @@ datastream<ST>& operator<<(datastream<ST>& ds, const history_serial_wrapper<eosi
...
-         auto  it   = undo.removed_values.find(obj.obj.code_id);
-         EOS_ASSERT(it != undo.removed_values.end(), eosio::chain::plugin_exception,
-                    "can not find code_object");
-         code_obj = &it->second;
-      }
-      fc::raw::pack(ds, as_type<uint8_t>(code_obj->vm_type));
-      fc::raw::pack(ds, as_type<uint8_t>(code_obj->vm_version));
-      fc::raw::pack(ds, as_type<eosio::chain::digest_type>(code_obj->code_hash));
-   }
+   fc::raw::pack(ds, as_type<uint8_t>(obj.obj.vm_type));
```
arhag: I think this should still use `code_id?` as part of the `account_metadata_v0` object. So if `obj.obj.code_hash` is empty, only pack the 0 byte. Otherwise pack the 1 byte and then these three fields.

plugins/state_history_plugin/state_history_plugin_abi.cpp
```diff
@@ -206,7 +199,9 @@ extern const char* const state_history_plugin_abi = R"({
                 { "type": "name", "name": "name" },
                 { "type": "bool", "name": "privileged" },
                 { "type": "time_point", "name": "last_code_update" },
-                { "type": "code_id?", "name": "code" }
+                { "type": "uint8", "name": "vm_type" },
```
arhag: Revert back to `code_id?`.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -64,30 +64,36 @@ namespace eosio { namespace chain {
...
-                                                                                    transaction_context& trx_context )
+      const std::unique_ptr<wasm_instantiated_module_interface>& get_instantiated_module( const digest_type& code_hash, const uint8_t& vm_type,
+                                                                                 const uint8_t& vm_version, transaction_context& trx_context )
       {
          wasm_cache_index::iterator it = wasm_instantiation_cache.find(
-                                             boost::make_tuple(code.code_hash, code.vm_type, code.vm_version) );
+                                             boost::make_tuple(code_hash, vm_type, vm_version) );
+         const code_object* codeobject = nullptr;
          if(it == wasm_instantiation_cache.end()) {
+            codeobject = db.find<code_object,by_code_hash>(boost::make_tuple(code_hash, vm_type, vm_version));
```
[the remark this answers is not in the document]
spoonincode: I want to maintain a find here so I don't have to potentially do the look up twice. I will put an EOS_ASSERT in to cover the nullptr pathological case

libraries/chain/eosio_contract.cpp
```diff
@@ -135,36 +136,62 @@ void apply_eosio_setcode(apply_context& context) {
...
+
+   int64_t old_size  = 0;
    int64_t new_size  = code_size * config::setcode_ram_bytes_multiplier;
 
-   EOS_ASSERT( account.code_version != code_id, set_exact_code, "contract is already running this version of code" );
+   if( existing_code ) {
+      const code_object& old_code_entry = db.get<code_object, by_code_hash>(boost::make_tuple(account.code_hash, account.vm_type, account.vm_version));
+      EOS_ASSERT( old_code_entry.code_hash != code_hash, set_exact_code,
+                  "contract is already running this version of code" );
+      int64_t old_size  = (int64_t)old_code_entry.code.size() * config::setcode_ram_bytes_multiplier;
```
arhag: The spurious `int64_t` causes consensus failure. Resolved in #7156.

PR #7113, opened 2019-04-11
Update ship's file format

tbfleming:
## Change Description

Replace ship's header to address these concerns in #5970:
* Switch to serialization instead of memcpy
* Rework header versioning
* Drop block_num since it's included in block_id
* Check header version in recover_blocks

These changes, plus changes in other PRs, make ship files incompatible with previous nodeos versions.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #6843, opened 2019-02-27
Add automated test (python) for irreversible mode

andriantolie:
## Change Description

Add automated test (python) for the irreversible mode (#6704) which encompasses the following 8 test cases:

- 1st test case: Replay in irreversible mode with reversible blks
    - Expectation: Node replays and launches successfully with head == libBeforeSwitchMode == lib and forkDbHead == headBeforeSwitchMode == forkDbHeadBeforeSwitchMode
    - Current Bug: duplicate blk added error

- 2nd test case: Replay in the irreversible mode without reversible blks
    - Expectation: Node replays and launches successfully with head == libBeforeSwitchMode == lib and forkDbHead == headBeforeSwitchMode == forkDbHeadBeforeSwitchMode
    - Current Bug: lib != libBeforeSwitchMode

- 3rd test case: Switch mode speculative -> irreversible without replay
    - Expectation: Node switches mode successfully with head == libBeforeSwitchMode == lib and forkDbHead == headBeforeSwitchMode == forkDbHeadBeforeSwitchMode
    - Current Bug: head != lib

- 4th test case: Switch mode irreversible -> speculative without replay
    - Expectation: Node switches mode successfully with head == forkDbHeadBeforeSwitchMode == forkDbHead and lib == headBeforeSwitchMode == libBeforeSwitchMode
    - Current Bug: head != forkDbHead and head != forkDbHeadBeforeSwitchMode and lib != libBeforeSwitchMode

- 5th test case: Switch mode speculative -> irreversible without replay and connected to producing node
    - Expectation: Node switches mode successfully and the head and lib should be advancing after some blocks produced with head == libBeforeSwitchMode == lib and forkDbHead == headBeforeSwitchMode == forkDbHeadBeforeSwitchMode
    - Current Bug: Fail to switch to irreversible mode, blk_validate_exception next blk in the future will be thrown

- 6th test case: Switch mode irreversible -> speculative without replay and connected to producing node
    - Expectation: Node switches mode successfully and the head and lib should be advancing after some blocks produced with head == forkDbHeadBeforeSwitchMode == forkDbHead and lib == headBeforeSwitchMode == libBeforeSwitchMode
    - Current Bug: Node switches mode successfully, however, it fails to establish connection with the producing node

- 7th test case: Replay in the irreversible mode with reversible blks while production is enabled
[...]

Issue #6736, opened 2019-02-13
Expected tests for protocol upgrade functionality


arhag:
## Plan for developing the protocol upgrade functionality

Several changes to the core code base are needed to facilitate protocol upgrade features.

First, it is necessary to build the foundations for protocol feature upgrades before a single (pre-activated) protocol feature of interest can actually be implemented. 

The first part of building the foundations has already been completed. PR #6588 and PR #6624 which complete the refactors of block header state and fork database in preparation for the changes needed for protocol features have already been merged into a separate branch `protocol-feature-foundations`. This branch (which is forked off of a recent develop branch) will host the changes for the protocol feature foundations. In addition to changes from PR #6588 and PR #6624 (and possibly some code changes that may be required to satisfy the requirements on irreversible mode documented in #6704), the main portion of the protocol feature foundations are the changes tracked by issues #6429, #6431, and #6437.

After the protocol feature foundations are completed, the `protocol-feature-foundations` branch will be forked off to the `forced-replay` branch which will be kept in sync with `develop`. The idea will be for the individual protocol features to merged into this branch (and any other PRs assuming replay from genesis is required) as they are developed rather than into `develop` directly. The protocol feature foundations include large changes that will force a replay from genesis. We do not want to merge these changes into `develop`, which would commit those changes to the next minor release of eosio release candidate, until we are confident that we will be able to get the desired set of initial protocol features in together with the protocol feature foundations for an RC release. 

In particular, it is important to ensure that the protocol feature tracked in issue #6103 is at least included in this initial set prior to a release, since that is the protocol feature that imposes the requirement of forced replay from genesis. In general, it would be ideal to include the protocol features intended for the initial set of protocol features that require a change to ChainBase data structures. The protocol features that require a change to the ChainBase data structures (and thus also a change to the deterministic snapshot specification) are labeled with `SNAPSHOT` on GitHub; not all protocol feature issues labeled `SNAPSHOTS` on the eos GitHub will be included in the initial set of protocol features to be released.

See the section below called "Candidates for initial protocol features" for the candidate set of protocol upgrade features to be considered for the initial release and the tests (at a high-level) expected for each.

## Testing the protocol feature foundations

The protocol feature foundations will require testing. 

The protocol upgrade tracked by #6431 (pre-activation intrinsic) itself acts as a good test of #6429 and especially #6437. 

A sufficient automated test of #6437 (and a significant part of #6429) would be to:

[...]

PR #7147, opened 2019-04-17
Block header state, fork database, chainbase, and transaction trace refactors; protocol feature foundations; initial set of new protocol features

arhag:
## Change Description

This PR brings in several changes from the forced-replay branch. They are broken down into the following categories:

   * Refactors of existing components
   * Protocol feature foundations
   * Initial set of new protocol features

All of these changes take advantage of the fact that replay from genesis is forced when upgrading from v1.7.x and earlier to v1.8.x. For example some of the refactors change the fork database structure and chainbase database format. And many changes are made to the indices storing consensus state information in the chainbase state database, which also means changes to what is stored and expected in the snapshot (chain snapshot version has been bumped to version 2) as well as in the state history files and table delta structures sent via its websocket API (no version bumps needed for the state history plugin since it was an alpha product that was expected to have breaking changes).

### Refactors of existing components

This PR brings in several changes from prior PRs merged into the `forced-replay` branch (or the `protocol-feature-foundations` branch prior to that) that refactored important components nodeos in preparation for the further changes that the protocol feature foundations bring.

The code related to `block_header_state` and its transitions have been significantly refactored. See PR #6588 for the details of the bulk of the refactoring. However, related PRs #6856 and #6864 fixed a bug resulting from that refactor and added a unit test related to that fixed bug, respectively.

The fork database was refactored in PR #6624. That PR also made significant changes to enable irreversible mode in nodeos. There were some additional bugs remaining with the newly activated irreversible mode that were fixed in PR #6842. Additionally a new Python long-running test was added in PR #6843 to test the additional functions of irreversible mode that unit tests were unable to test.

Also, this PR brings the chainbase submodule up to the latest master which incorporates three PRs merged in the chainbase repository: EOSIO/chainbase#41, EOSIO/chainbase#42, EOSIO/chainbase#43. Of particular importance, is EOSIO/chainbase#42 which uses a new portable header at the start of the shared_memory.bin files used for the chainbase databases to record the environment record with which the database was created with (as well as the dirty flag). This enables a safe and reliable way to check if the current environment matches the environment record stored in the file before opening up the database.

Furthermore, PR #7134 separates out the index to store the actual WebAssembly code from the indices storing account information. It also moves some account-related fields from the `account_object` to what is now known as the `account_metadata_object` (formerly the `account_sequence_object`).  PR #7156 fixes a bug introduced in PR #7134.

Finally, PR #7044 restructures transaction traces to use a flatten structure for all the action traces it contains, but with extra metadata information to allow reconstruction of the legacy execution tree as well as the recommended creation tree that provides more information and resolves certain ambiguities that existed with the legacy tree. The `push_transaction` and `push_transactions` chain API calls reconstruct the legacy tree before returning the trace back to the caller in an effort to be backwards compatible with existing clients. However, this comes at a cost of additional work that must be done by the server. Updated clients can use the new `send_transaction` RPC in chain API to get back the new flat transaction trace and avoid the additional computations (and thus slow downs) that `push_transaction` must do. In additional, PR #7108 adds a new `error_code` field to the transaction traces.

### Protocol features

The foundations for protocol features were implemented in PR #6831. See the PR for details of the changes as well as the some documentation on the new API endpoints that it introduces. That PR also introduces the `PREACTIVATE_FEATURE` protocol feature which is a critical part of the protocol feature foundations. PR #7167 adds on additional changes to better handle error cases when handling the JSON files specifying the protocol features.

The release that introduces the protocol feature foundations also includes an initial set of protocol features (see #6736 for details). This PR includes implementations for the following set of protocol features that are intended to be included in the initial release:

1. `ONLY_LINK_TO_EXISTING_PERMISSION` (implemented in PR #6831 which is the same PR that implements the protocol feature foundations)
2. `REPLACE_DEFERRED` (#6997, #7185)
3. `NO_DUPLICATE_DEFERRED_ID` (#7072, #7173)
4. `FIX_LINKAUTH_RESTRICTION` (#7025)
5. `DISALLOW_EMPTY_PRODUCER_SCHEDULE` (#7026)
6. `RESTRICT_ACTION_TO_SELF` (#7088)
7. `ONLY_BILL_FIRST_AUTHORIZER` (#7089)
8. `FORWARD_SETCODE` (#7109)
9. `GET_SENDER` (#7111)
10. `RAM_RESTRICTIONS` (#7131)

### Other changes

PR #7123 updates the wabt submodule to include changes that may possible improve performance.

PR #7113 makes changes to the structure of files used by the state history plugin. PR #7139 makes adjustments to the state history plugin to avoid outputting table deltas when a modification to the table row does not change any of the fields exposed to state history consumers. PR #7150 augments the `transaction_trace` structure stored and returned by the state history plugin to include additional data from the original transaction which, in the case of non-delayed input transactions, enables consumers to reconstruct the `signed_transaction`. PR #7163 adds a limited form of the `protocol_state_object` table deltas (only including the `activated_protocol_features` field) to the state history plugin.

PR #7143 updates the bios contract to one supporting the new `activate` action and updates the tests to work properly with the new bios contract as well as the restrictions of the `RESTRICT_ACTION_TO_SELF` protocol feature (specifically it updates the bootstrapping procedure of the Python tests to use `eosio` rather than `eosio.token` as the token issuer).

PR #7172 updates the behavior of the `create_snapshot` RPC in the `producer_plugin` to immediately return the final snapshot path when the node is in the irreversible read mode. 

PR #7179 evicts the cached instances of WebAssembly code when the last block in which it existed becomes irreversible.

## Consensus Changes
- [x] Consensus Changes

See #6831, #6997, #7072, #7025, #7026, #7088, #7089, #7109, #7111, and #7131.

## API Changes
- [x] API Changes

See #6831, #7044, #7113, #7139, #7150, #7163, #7172, and #7185.

## Documentation Additions
- [x] Documentation Additions

Documentation changes need to describe the new irreversible mode (see #6624 and #6842 and #6843 for related PRs).

Also see #6831, #7088, #7044, #7108, #7113, #7139, #7150, and #7163.


libraries/chain/apply_context.cpp
```diff
@@ -29,68 +30,98 @@ static inline void print_debug(account_name receiver, const action_trace& ar) {
...
 
-         if( a.code.size() > 0
-             && !(act.account == config::system_account_name && act.name == N( setcode ) &&
-                  receiver == config::system_account_name) ) {
+         if( ( receiver_account->code_hash != digest_type() ) &&
+               ( control.is_builtin_activated( builtin_protocol_feature_t::forward_setcode )
+                  || !( act->account == config::system_account_name
+                        && act->name == N( setcode )
+                        && receiver == config::system_account_name )
+               )
```
heifner: Seems like the right-hand side of `||` would be faster to check. Can we check it first.

heifner: Precedence order is `&&` then `||` but I think another set of parens would make it clearer.

arhag: Resolved by https://github.com/EOSIO/eos/pull/7147/commits/a0f40f517b150f9a3162068aa032c3a95f721332.

libraries/chain/apply_context.cpp
```diff
@@ -363,39 +464,53 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
+         gtx.delay_until = gtx.published + delay;
+         gtx.expiration  = gtx.delay_until + fc::seconds(control.get_global_properties().configuration.deferred_trx_expiration_window);
+
+         trx_size = gtx.set( trx );
+      } );
    }
 
-   EOS_ASSERT( control.is_ram_billing_in_notify_allowed() || (receiver == act.account) || (receiver == payer) || privileged,
-               subjective_block_production_exception, "Cannot charge RAM to other accounts during notify." );
+   EOS_ASSERT( control.is_builtin_activated( builtin_protocol_feature_t::ram_restrictions )
```
heifner: Since `control.is_builtin_activated( builtin_protocol_feature_t::ram_restrictions )` is always called, it can be called above and stored in a local variable to avoid calling it twice in this method.

arhag: Resolved by https://github.com/EOSIO/eos/pull/7147/commits/10b8ae10d0c6f832f5e2d787f0f91295a1d2c2ae

PR #7165, opened 2019-04-19
add mageia linux to build scripts

joequant:
## Change Description

This adds the build script for mageia linux.  I am a maintainer for mageia linux and would like to 
add eos to the distribution.
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/eosio_build.sh
+++ b/scripts/eosio_build.sh
@@ -229,6 +229,12 @@ if [ "$ARCH" == "Linux" ]; then
          CXX_COMPILER=g++
          C_COMPILER=gcc
       ;;
+      "Mageia")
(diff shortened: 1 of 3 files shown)


spoonincode:
hey, thanks for your PR. Unfortunately, we've decided that we only want the build scripts to run on distributions that we support. Supported distributions get continual attention from us such as every PR running on CI across all of them.

If a distribution has a c++17 compiler, boost 1.67+, LLVM 4.0, and some other very basic development packages, then it's typically possible to build eosio without the build script. I realize the LLVM 4.0 requirement tends to be the tricky one to come by and hopefully that requirement won't be as rigid in the future.

PR #7171, opened 2019-04-19
reduction of wavm memory usage

spoonincode:
## Change Description
wavm tends to cause significant memory usage even when fixes like #6983 are in place. This PR includes two changes that reduce the memory usage of WAVM by a large margin. First: avoid generation of debug information; this was actually done at a per-wasm-opcode resolution so it was very memory intensive and also tended to wedge things in to a global cache that would never be freed. Second: free the WAVM Module after we no longer need it once a ModuleInstance is created from it

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/webassembly/wavm.cpp
+++ b/libraries/chain/webassembly/wavm.cpp
@@ -31,7 +31,7 @@ class wavm_instantiated_module : public wasm_instantiated_module_interface {
          // that didn't declare "memory", getDefaultMemory() won't see it. It would also be possible
          // to say something like if(module->memories.size()) here I believe
          if(getDefaultMemory(_instance))
-            _initial_memory_config = module->memories.defs[0].type;
+            _initial_memory_config = module->memories.defs.at(0).type;
       }
 
       void apply(apply_context& context) override {
--- a/libraries/chain/webassembly/wavm.cpp
+++ b/libraries/chain/webassembly/wavm.cpp
@@ -25,9 +25,14 @@ class wavm_instantiated_module : public wasm_instantiated_module_interface {
    public:
       wavm_instantiated_module(ModuleInstance* instance, std::unique_ptr<Module> module, std::vector<uint8_t> initial_mem) :
          _initial_memory(initial_mem),
-         _instance(instance),
-         _module(std::move(module))
-      {}
+         _instance(instance)
+      {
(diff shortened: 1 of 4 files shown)


libraries/chain/webassembly/wavm.cpp
```diff
@@ -25,9 +25,14 @@ class wavm_instantiated_module : public wasm_instantiated_module_interface {
...
-         _instance(instance),
-         _module(std::move(module))
-      {}
+         _instance(instance)
+      {
+         //The memory instance is reused across all wavm_instantiated_modules, but for wasm instances
+         // that didn't declare "memory", getDefaultMemory() won't see it. It would also be possible
+         // to say something like if(module->memories.size()) here I believe
+         if(getDefaultMemory(_instance))
+            _initial_memory_config = module->memories.defs[0].type;
```
heifner: This would be safer to use `.defs.at(0)`

libraries/wasm-jit/Source/Runtime/LLVMJIT.cpp
```diff
@@ -514,16 +497,12 @@ namespace LLVMJIT
...
-					llvm::DILineInfoTable lineInfoTable = dwarfContext->getLineInfoForAddressRange(loadedAddress,symbolSizePair.second);
-					std::map<U32,U32> offsetToOpIndexMap;
-					for(auto lineInfo : lineInfoTable) { offsetToOpIndexMap.emplace(U32(lineInfo.first - loadedAddress),lineInfo.second.Line); }
-					
 					#if PRINT_DISASSEMBLY
 					Log::printf(Log::Category::error,"Disassembly for function %s\n",name.get().data());
 					disassembleFunction(reinterpret_cast<U8*>(loadedAddress),Uptr(symbolSizePair.second));
 					#endif
 
+					std::map<U32,U32> offsetToOpIndexMap;
```
heifner: This does not appear to be used.

spoonincode: you're right that it's ultimately not used once it's moved in to `notifySymbolLoaded`. I'd like a pass on this one: there is even more unused dead code in wavm from this change and other changes too. I'm too nervous about tearing out more code at 11th hour

PR #7175, opened 2019-04-20
Mitigate coupling among plugins

conr2d:
## Change Description

This PR is written for the purpose of experiment. Feel free to reject, if this conflicts with original design.

Remove strong dependencies on chain_plugin from plugins:
- history_plugin
- mongo_db_plugin
- state_history_plugin
- net_plugin
- producer_plugin

### Rationale

I'm writing a program which allows contract developers to write unit tests easily with JavaScript. Its executable works like nodeos, but it uses internally `eosio::testing::tester` and exposes its methods to RPC APIs. `test_rpc_plugin` substitutes for chain_plugin and producer_plugin in my program, but other plugins cannot be used, because they depend on chain_plugin directly. (usually containing a pointer to chain_plugin and call `chain_plug->chain()` to retrieve existing `chain::controller`)

### Need advise

1. Is it all right to use `eosio::testing::tester` in my program?

1. If chain_plugin is removed from `APPBASE_PLUGIN_REQUIRES` macro, an instantiation of `chain::controller` before other plugins initialization is not guaranteed. I think there can be a wrapper named like `chain_provider_plugin` which initializes `chain_plugin` by default, but when `--chain-provider` options is set explicitly (in my case, `--chain-provider=eosio::test_rpc_plugin`), initializes specified one. Other plugins will depend on `chain_provider_plugin` instead of `chain_plugin`. 

1. `appbase::application::get_method` returns default method when no implementation is provided, but what about adding a method like `require_method` or boolean parameter `get_method(bool required = false)` can throw an exception when no explicit implementation is provided?

1. I think it doesn't need to remove dependencies from producer_plugin and net_plugin, because their functionalities strongly bound to chain_plugin. Especially, `net_plugin` calls `chain_plugin::accept_transaction` directly. If this PR can be merged, is it better keeping dependencies in the case of producer_plugin and net_plugin?

### Remaining works
- [ ] Remove chain_plugin dependencies completely
- [ ] Implement check whether controller is provided correctly

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


conr2d:
If `chain_provider_plugin` that I mentioned above is added, it would be possible that other plugins have a pointer to `chain_provider_plugin` in place of `chain_plugin`. If the target plugin (chain_plugin or similar ones) initialized by chain_provider_plugin overrides pure virtual methods, `chain()` and `get_abi_serializer_max_time()` (by inheriting from abstract class like chain_provider_interface), then indirect call by `app().get_method<methods::get_controller>()()` might not be necessary.

b1bart:
@conr2d Thanks for tackling this.  I have some high level feedback that I wanted your input on:

1. YES 

2. Instead of a "provider" plugin that resolves the dependency problem, what are your opinions on moving to less a less direct form of dependency resolution in `appbase`?  For instance, each plugin could `require` and `provide` a set of methods and channels (like `methods::get_controller`) in addition to the explicit plugin dependencies.   Then at start-up we can use this information to resolve the initialization order of the plugins and detect any errors (like cycles).  

3. I think this should be covered by (2) but if it isn't, I agree some sort of stricter form of `get_method` would be useful

4. For now `producer_plugin` and `chain_plugin` are tightly coupled.  We would like in the future to de-couple `net_plugin` however the amount of plumbing required to do so now (as opposed to through channels/methods) is a fair bit.  If you want to leave these coupled for now, that is ok.

One last note, our internal etiquette for this repo ( documented nowhere 😇 ) is to **never force push a branch that is in PR** .  It makes it very difficult to track where the review comments applied etc.  Even if it leaves a bad log, we prefer to have the intact git history.  There is no penalty or enforcement, it's just a friendly request.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -184,6 +184,8 @@ class chain_plugin_impl {
    methods::get_block_by_id::method_type::handle                     get_block_by_id_provider;
    methods::get_head_block_id::method_type::handle                   get_head_block_id_provider;
    methods::get_last_irreversible_block_number::method_type::handle  get_last_irreversible_block_number_provider;
+   methods::get_controller::method_type::handle                      get_controller_provider;
+   methods::get_abi_serializer_max_time::method_type::handle         get_abi_serializer_max_time_provider;
```
heifner: `abi_serializer_max_time` I think can be grabbed from command line for plugings. See how mongo_db_plugin accesses it. I don't think we need a `methods` for it.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -205,7 +205,7 @@ namespace eosio {
          }
 
          void record_account_action( account_name n, const action_trace& act ) {
-            auto& chain = chain_plug->chain();
+            auto& chain = app().get_method<methods::get_controller>()();
```
heifner: I would rather see this cached on the plugin instead of looking up the method each time as these are called very offen.

conr2d:
Thank you for review. I'll rewrite PR soon.

conr2d:
I'll open this later when writing new patch. :)

PR #7176, opened 2019-04-22
Destroy wavm module instances when deleting the wavm_instantiated_module

spoonincode:
## Change Description
When we instantiate a wasm with wavm we get back a ModuleInstance. But deleting this object is a noop because wavm has its own garpage collection thing. Implement the code to interface with the garbage collection so that deleting a ModuleInstance can really free up the objects that are associated with it. Otherwise, evictions from the wasm_cache don't actually free up wavm (or system) resources

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/webassembly/wavm.cpp
```diff
@@ -21,14 +22,47 @@ namespace eosio { namespace chain { namespace webassembly { namespace wavm {
...
+using live_module_ref = std::list<ObjectInstance*>::iterator;
+
+struct wavm_live_modules {
+   live_module_ref add_live_module(ModuleInstance* module_instance) {
+      return live_modules.insert(live_modules.begin(), asObject(module_instance));
+   }
+
+   void remove_live_module(live_module_ref it) {
+      live_modules.erase(it);
+      std::vector<ObjectInstance*> root;
```
heifner: This looks like you are doing garbage collection here because this is a convenient place to do it. Maybe pull into its own method so it is clearer that is what is going on.

spoonincode: Not following what you're suggesting here. It's only 4 line method there is not much to pull out.

The situation is that, internally, WAVM is maintaining a list of every object that is created. By calling WAVM's `Runtime::freeUnreferencedObjects()` I'm giving it a list of objects to retain and then it does GC removing what is not in the intersection of `root` and its internal list. (it's more complicated than this internally to wavm because there are child objects and stuff that are retained if the parent is retained)

Unfortunately `Runtime::freeUnreferencedObjects()` expects to mutate the list that I hand it which is why I have to create a copy from a std::list here.

heifner: My point is that the garbage collection is tangential to the remove, so it makes more sense to have it in a diff method.

```
void remove_live_module(live_module_ref it) {
  live_modules.erase(it);
  garbage_collect();
}

void garbage_collect() {
   std::vector<ObjectInstance*> root;
   std::copy(live_modules.begin(), live_modules.end(), std::back_inserter(root));
   Runtime::freeUnreferencedObjects(std::move(root));
}
```

PR #7177, opened 2019-04-22
Avoid possibly lengthy destruction of wasm instances on shutdown

spoonincode:
## Change Description

When the wasm_interface is destroyed, all unique_ptrs to the current instantiated modules are also destroyed. Unfortunately for some wasm runtimes -- wavm with #7176 -- this incurs a lengthy garbage collection process for each module that is deleted. This can cause destruction of wasm_interface to take on the order of multiple minutes when there are thousands of instances! We need to violate the abstraction layer here and drop these instances on the floor during shutdown so the user isn't waiting minutes

I am completely open to other approaches here

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -32,6 +32,11 @@ namespace eosio { namespace chain {
             EOS_THROW(wasm_exception, "wasm_interface_impl fall through");
       }
 
+      ~wasm_interface_impl() {
+         if(is_shutting_down)
+            std::for_each(instantiation_cache.begin(), instantiation_cache.end(), [](auto& i) {i.second.release();});
```
heifner: Add an `instantiation_cache.clear()`

spoonincode: why is that needed? `instantiation_cache` will be destroyed once dtor exits

heifner: Ah, yes, of course.

PR #7187, opened 2019-04-23
app shutdown when corrupt blocks during reconstruction

dorian-yeager:
## Change Description
Prevents a crash when index reconstruction fails on corrupt blocks.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
Thanks for this. However, we addressed at least one case where this is a problem here: https://github.com/EOSIO/eos/pull/7188
In other cases, we want to do a forced replay.

PR #7212, opened 2019-04-25
commit db during replays when replay optimizations are disabled (1.7 backport)

wanderingbort:
## Change Description

porting fix made to develop for committing database on replay with disable-replay-opts

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


wanderingbort:
This is actually not needed as the call to `fork_db.add` will eventually call the `db.commit`

PR #7273, opened 2019-05-06
Added functionality to skip rows when calling get table

jasonbert:
## Change Description
The change introduces the ability to skip table rows, thus enabling the ability to paginate results. The skip functionality is independent of the lower & upper bounds. A primary example is iterating over all rows in a table, where there could be thousands, in a controlled manner by utilising skip & limit parameters.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

A skip parameter has been added, as this is an addition there should be no breaking changes.

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -286,6 +286,7 @@ public:
       string      lower_bound;
       string      upper_bound;
       uint32_t    limit = 10;
+      uint64_t    skip = 0;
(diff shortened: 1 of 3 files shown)


tbfleming:
Since the indexes are implemented by trees, `std::next` is a linear operation instead of a constant-time operation. This could lead to a DOS attack.

tbfleming:
`std::next` can also attempt to go past the end, causing undefined behavior.

jasonbert:
Thanks for the review @tbfleming , I'll re-work the PR.

b1bart:

PR #7285, opened 2019-05-06
Use of Mac Anka Fleet instead of iMac fleet

NorseGaud:
Problem: Current Parallels/iMac fleet requires a human to create new images anytime dependencies change in the build scripts. Not only that, but we cannot free up a buildkite agent to pick up a different kind of job if it's just sitting around not doing anything and it's setup for a specific purpose (mac osx with xcode 10.2 OR nightly "clean" vms).

Solution: The Mac Anka fleet provides the flexibility of on-demand buildkite runs across mac minis with 12 cores and 32G of memory. Half of the fleet is limited to a single buildkite-agent, using the entire 12 cores; The other half runs two buildkite-agents with 6 cores each. This should greatly increase the performance and speed of running builds (which will use 12 cores) and tests (which only need 6). VM image layers are created just like docker and then pulled onto the host when we need them (or it uses the local copy if it's already been downloaded before).


NorseGaud:
Building new images for linux...

scripts/serial-test.sh
```diff
@@ -1,36 +1,21 @@
...
 ctest -L nonparallelizable_tests --output-on-failure -T Test
 EXIT_STATUS=$?
-[[ "$EXIT_STATUS" == 0 ]] && set -e
-echo "Done running non-parallelizable tests."
-# upload artifacts
-echo "Uploading artifacts..."
-XML_FILENAME="test-results.xml"
-mv $(pwd)/Testing/$(ls $(pwd)/Testing/ | grep '20' | tail -n 1)/Test.xml $XML_FILENAME
-buildkite-agent artifact upload config.ini
-buildkite-agent artifact upload genesis.json
```
zreyn: As far as I know we don't use these artifacts, but someone put them there for a reason.  Are you sure they're not used?

.buildkite/pipeline.yml
```diff
@@ -32,20 +59,18 @@ steps:
         account_ids: "436617320021"
         no-include-email: true
         region: "us-west-2"
-      docker#v2.1.0:
+      docker#v3.1.0:
```
zreyn: I'm assuming you've tested this change and it works?  Sometimes upgrading versions will break existing code.

.buildkite/pipeline.yml
```diff
@@ -401,32 +435,36 @@ steps:
...
-        buildkite-agent artifact download "build.tar.gz" . --step ":darwin: Mojave Build"
-        tar -zxf build.tar.gz
-        echo "+++ :microscope: Starting package build"
-        ln -s "$(pwd)" /data/job
-        cd /data/job/build/packages
-        bash generate_package.sh brew
-    label: ":darwin: Mojave Package Builder"
+  - label: ":darwin: Mojave Package Builder"
+    command:
+      - "tar -xzf /Network/NAS/MAC_FLEET/BUILDKITE/artifacts/${ANKA_MOJAVE_TEMPLATE}-${BUILDKITE_PIPELINE_SLUG}-${BUILDKITE_BUILD_ID}.tar.gz"
```
zreyn: Why use the NAS vs buildkite artifact?

scripts/eosio_build_ubuntu.sh
```diff
@@ -219,7 +219,7 @@ if [ "$BUILD_CLANG8" = "true" ]; then
 
    cd $SRC_LOCATION
    printf "Checking zlib library installation...\\n"
-   if [ ! -d $OPT_LOCATION/zlib || $FORCE_BUILD ]; then
+   if [ ! -d $OPT_LOCATION/zlib ] || $FORCE_BUILD; then
```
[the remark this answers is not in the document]
NorseGaud: [ ] is redundant there

scripts/serial-test.sh
```diff
@@ -1,36 +1,21 @@
...
-echo "Extracting build directory..."
-[[ -z "${1}" ]] && tar -zxf build.tar.gz || tar -xzf $1
-echo "Starting MongoDB..."
-mongod --fork --dbpath ~/data/mongodb -f ~/etc/mongod.conf --logpath "$(pwd)"/mongod.log
-cd /data/job/build
-# run tests
-echo "Running tests..."
-TEST_COUNT=$(ctest -N -L nonparallelizable_tests | grep -i 'Total Tests: ' | cut -d ':' -f 2 | awk '{print $1}')
-[[ $TEST_COUNT > 0 ]] && echo "$TEST_COUNT tests found." || (echo "ERROR: No tests registered with ctest! Exiting..." && exit 1)
+mongod --fork --dbpath ~/data/mongodb -f ~/etc/mongod.conf --logpath ~/var/log/mongodb/mongod.log
```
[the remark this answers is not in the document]
NorseGaud: If you don't move it before uploading the artifact ends up being the absolute path to the file /Users/blah/var/log/mongodb/mongod.log in buildkite. If we try and download it, we have to specify that full path/name for the artifact. 

The build directory is always writable, though I've added -f for some mv assurance that it will overwrite. We added an ENV that forces the "working directory" for buildkite to clean itself before doing the git clone/checkout, so that should definitely prevent any sort of collisions.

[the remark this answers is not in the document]
NorseGaud: You're right. I wanted to keep things consistent with the build instructions, but I realize now that's futile.

.buildkite/pipeline.yml
NorseGaud: Yep, the pipeline step is passing with the new docker plugin version so it's doing what it should.

.buildkite/pipeline.yml
NorseGaud: Much faster and eliminates data usage from the local network out (Nick V requested this)

scripts/parallel-test.sh
```diff
@@ -1,35 +1,20 @@
 #!/bin/bash
-set -e # exit on failure of any "simple" command (excludes &&, ||, or | chains)
-# prepare environment
-PATH=$PATH:~/opt/mongodb/bin
-echo "Extracting build directory..."
```
kj4ezj: These types of print statements are used for debugging. It is okay if you want to reformat them....    
`echo "[Extracting build directory]"`     
-- or --    
`echo "+++ :compression: Extracting build directory..."`    
...but please do not remove them entirely because when something fails I want people to come up to me saying "My build failed on 'extracting build directory'," rather than "my build crashed before it started". The more information we give engineers, the more they can diagnose themselves.

scripts/parallel-test.sh
```diff
@@ -1,35 +1,20 @@
...
-[[ -z "${1}" ]] && tar -zxf build.tar.gz || tar -xzf $1
-cd /data/job/build
-# run tests
-echo "Running tests..."
-CPU_CORES=$(getconf _NPROCESSORS_ONLN)
-echo "$CPU_CORES cpu cores detected."
-TEST_COUNT=$(ctest -N -LE _tests | grep -i 'Total Tests: ' | cut -d ':' -f 2 | awk '{print $1}')
```
kj4ezj: This check here...
```
TEST_COUNT=$(ctest -N -LE _tests | grep -i 'Total Tests: ' | cut -d ':' -f 2 | awk '{print $1}')
[[ $TEST_COUNT > 0 ]] && echo "$TEST_COUNT tests found." || (echo "ERROR: No tests registered with ctest! Exiting..." && exit 1)
```
...needs to stay. The purpose of this check is to verify that a positive number of tests are registered with ctest. The reason for doing this is that if someone messes with the CMake files and accidentally un-registers all tests, ctest will execute zero tests and, with zero failures, the build will pass. This has happened before and this check prevents that.

scripts/parallel-test.sh
```diff
@@ -1,35 +1,20 @@
...
-[[ $TEST_COUNT > 0 ]] && echo "$TEST_COUNT tests found." || (echo "ERROR: No tests registered with ctest! Exiting..." && exit 1)
+set -eo pipefail
+[[ -f build.tar.gz ]] && tar -xzf build.tar.gz
+cd ./build
```
kj4ezj: You've done a great job of minimizing this code!   
In that spirit, `cd build` is equivalent to `cd ./build`.

scripts/parallel-test.sh
```diff
@@ -1,35 +1,20 @@
...
+echo "[Running tests]"
 set +e # defer ctest error handling to end
-echo "$ ctest -j $CPU_CORES -LE _tests --output-on-failure -T Test"
```
kj4ezj: I am not opposed to removing this echo statement for the reasons Matt brought up in the past, but the purpose of it is so the engineers can see (in the logs) the exact command Buildkite is running to replicate it on their own machines.

scripts/parallel-test.sh
```diff
@@ -1,35 +1,20 @@
...
+echo "[Uploading artifacts]"
+mv -f ./Testing/$(ls ./Testing/ | grep '20' | tail -n 1)/Test.xml test-results.xml
+buildkite-agent artifact upload test-results.xml
 # ctest error handling
-if [[ "$EXIT_STATUS" != 0 ]]; then
+if [[ $EXIT_STATUS != 0 ]]; then
     echo "Failing due to non-zero exit status from ctest: $EXIT_STATUS"
-    echo '    ^^^ scroll up for more information ^^^'
+    buildkite-agent artifact upload config.ini
+    buildkite-agent artifact upload genesis.json
```
kj4ezj: Nice work, making this conditional on failure. I like it.

scripts/serial-test.sh
```diff
@@ -1,36 +1,25 @@
 #!/bin/bash
-set -e # exit on failure of any "simple" command (excludes &&, ||, or | chains)
-# prepare environment
+set -eo pipefail
+[[ -f build.tar.gz ]] && tar -xzf build.tar.gz
+cd ./build
+echo "[Killing old MongoDB]"
+$(pgrep mongod | xargs kill -9) || true
```
kj4ezj: `$(pgrep mongod | xargs kill -9) || true`    
is equivalent to  
`$(pgrep mongod | xargs kill -9) || :`    
The `:` character is no-op in BASH.

scripts/serial-test.sh
```diff
@@ -1,36 +1,25 @@
 #!/bin/bash
-set -e # exit on failure of any "simple" command (excludes &&, ||, or | chains)
-# prepare environment
+set -eo pipefail
+[[ -f build.tar.gz ]] && tar -xzf build.tar.gz
```
kj4ezj: Add `echo "[Extracting build directory]"` for reasons stated above.

scripts/serial-test.sh
```diff
@@ -1,36 +1,25 @@
...
+echo "[Starting MongoDB]"
 PATH=$PATH:~/opt/mongodb/bin
-echo "Extracting build directory..."
-[[ -z "${1}" ]] && tar -zxf build.tar.gz || tar -xzf $1
-echo "Starting MongoDB..."
-mongod --fork --dbpath ~/data/mongodb -f ~/etc/mongod.conf --logpath "$(pwd)"/mongod.log
-cd /data/job/build
-# run tests
-echo "Running tests..."
-TEST_COUNT=$(ctest -N -L nonparallelizable_tests | grep -i 'Total Tests: ' | cut -d ':' -f 2 | awk '{print $1}')
```
kj4ezj: Restore the empty test set check, for reasons stated above:    
```
TEST_COUNT=$(ctest -N -L nonparallelizable_tests | grep -i 'Total Tests: ' | cut -d ':' -f 2 | awk '{print $1}')
[[ $TEST_COUNT > 0 ]] && echo "$TEST_COUNT tests found." || (echo "ERROR: No tests registered with ctest! Exiting..." && exit 1)
```

.buildkite/pipeline.yml
```diff
@@ -1,28 +1,55 @@
...
-      - "role=builder-v2-2"
-      - "os=mojave"
-    artifact_paths: "build.tar.gz"
+      - "queue=mac-anka-large-node-fleet"
     timeout: 120
 
-  - command: | # Amazon Linux 2 Build
-        echo "+++ :hammer: Building"
-        ./scripts/eosio_build.sh -y -P -m
-        echo "--- :compression: Compressing build directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -32,20 +59,18 @@ steps:
         account_ids: "436617320021"
         no-include-email: true
         region: "us-west-2"
-      docker#v2.1.0:
+      docker#v3.1.0:
+        shell: ["/bin/bash", "-i", "-e", "-c"]
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2-4"
```
kj4ezj: We should be using labels on docker images rather than changing the image names.    
Ex:    
`image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2:latest"`

.buildkite/pipeline.yml
```diff
@@ -32,20 +59,18 @@ steps:
...
         propagate-environment: true
         workdir: /data/job
     timeout: 120
 
-  - command: | # CentOS 7 Build
-        echo "+++ :hammer: Building"
-        ./scripts/eosio_build.sh -y -P -m
-        echo "--- :compression: Compressing build directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -55,19 +80,18 @@ steps:
...
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:centos7_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:centos7_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 120
 
-  - command: | # Ubuntu 16.04 Build
-        echo "+++ :hammer: Building"
-        ./scripts/eosio_build.sh -y -P -m
-        echo "--- :compression: Compressing build directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -78,20 +102,18 @@ steps:
...
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu16_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu16_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 120
 
-  - command: | # Ubuntu 18.04 Build
-        echo "+++ :hammer: Building"
-        ./scripts/eosio_build.sh -y -P -m
-        echo "--- :compression: Compressing build directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -125,18 +187,18 @@ steps:
...
+        shell: ["/bin/bash", "-i", "-e", "-c"]
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -146,20 +208,19 @@ steps:
...
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2-2"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:amazonlinux2_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  # centOS 7 Tests
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -168,19 +229,18 @@ steps:
...
+        shell: ["/bin/bash", "-i", "-e", "-c"]
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:centos7_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:centos7_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -189,20 +249,19 @@ steps:
...
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:centos7_2-2"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:centos7_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  # Ubuntu 16.04 Tests
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -211,19 +270,18 @@ steps:
...
+        shell: ["/bin/bash", "-i", "-e", "-c"]
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu16_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu16_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -232,20 +290,18 @@ steps:
...
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu16_2-2"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu16_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  # Ubuntu 18.04 Tests
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
```diff
@@ -254,19 +310,18 @@ steps:
...
+        shell: ["/bin/bash", "-i", "-e", "-c"]
         debug: true
-        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu18_2-3"
+        image: "436617320021.dkr.ecr.us-west-2.amazonaws.com/ci:ubuntu18_2-4"
         propagate-environment: true
         workdir: /data/job
     timeout: 60
 
-  - command: |
-        echo "--- :arrow_down: Downloading Build Directory"
```
kj4ezj: Please don't delete these types of print statements.
Without this print statement, it will look like the agent has hanged.

.buildkite/pipeline.yml
NorseGaud: The commands executed after the echo have output.

PR #7288, opened 2019-05-06
Changes for Boost 1_70_0

### LINKED ISSUE #7098: boost 1.70 build problems
boost 1.70 removes some methods that were depreciated in 1.67. At least http_client and bnet_plugin are affected.

heifner:
## Change Description

- Changes so compile with boost 1_70_0
- Resolves #7098 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/local_endpoint.hpp
@@ -82,8 +82,7 @@ protected:
             return socket::make_error_code(socket::error::invalid_state);
         }
 
-        m_socket = lib::make_shared<lib::asio::local::stream_protocol::socket>(
-            lib::ref(*service));
+        m_socket = lib::make_shared<lib::asio::local::stream_protocol::socket>(*service);
 
         m_state = READY;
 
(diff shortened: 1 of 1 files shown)


spoonincode:
This isn't enough for me when I tried it on macos, I still needed to pass `-DBoost_NO_BOOST_CMAKE=ON`. It's not yet clear to me why this is needed & what changed.

heifner:
I did not see the cmake issue on Ubuntu.

As to what changed, the deprecated `get_io_service` of socket was removed and they modified the constructor to take a templated type instead of an io_context.

https://www.boost.org/doc/libs/1_67_0/doc/html/boost_asio/reference/basic_stream_socket/basic_stream_socket.html
vs
https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/basic_stream_socket/basic_stream_socket.html

spoonincode:
boost 1.70 appears to be the first version that includes its own cmake Config files. The FindBoost function in cmake (either the ancient one we still have in fc or even the newest in cmake 3.14+) will completely delegate boost configuration to the Config files if it finds them. The Config files aren't entirely compatible with the older FindBoost usage -- they only set up imported targets; something we don't use -- so our current cmake usage fails.

However, it seems like it only searches for the Config files in the default search paths, not in BOOST_ROOT. So I think you'd have to install 1.70 such that you don't need to specify BOOST_ROOT to suffer from the problem.

These changes look good but I don't think we can clear #7098 until we fix the cmake issues. There are a number of approaches we could take.

PR #7294, opened 2019-05-07
quick fix for zlib

NorseGaud:
zlib test fails

### DIFF
--- a/scripts/eosio_build_amazon.sh
+++ b/scripts/eosio_build_amazon.sh
@@ -46,7 +46,7 @@ printf "Yum installation found at ${YUM}.\\n"
 if [ $ANSWER != 1 ]; then read -p "Do you wish to update YUM repositories? (y/n) " ANSWER; fi
 case $ANSWER in
 	1 | [Yy]* )
-		if ! sudo $YUM -y update; then
+		if ! sudo -E $YUM -y update; then
 			printf " - YUM update failed.\\n"
 			exit 1;
 		else
@@ -76,7 +76,7 @@ if [ "${COUNT}" -gt 1 ]; then
 	if [ $ANSWER != 1 ]; then read -p "Do you wish to install these dependencies? (y/n) " ANSWER; fi
 	case $ANSWER in
 		1 | [Yy]* )
-			if ! sudo $YUM -y install ${DEP}; then
+			if ! sudo -E $YUM -y install ${DEP}; then
 				printf " - YUM dependency installation failed!\\n"
 				exit 1;
 			else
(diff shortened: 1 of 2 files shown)


scripts/eosio_build_ubuntu.sh
```diff
@@ -219,7 +219,7 @@ if [ "$BUILD_CLANG8" = "true" ]; then
 
    cd $SRC_LOCATION
    printf "Checking zlib library installation...\\n"
-   if [ ! -d $OPT_LOCATION/zlib || $FORCE_BUILD ]; then
+   if [ ! -d $OPT_LOCATION/zlib ] || $FORCE_BUILD; then
       printf "Installing zlib...\\n"
```
heifner: Above it is `[ $FORCE_BUILD ];` I think, make it the same for consistency.

heifner: Well that took an unexpected turn ;)

scripts/eosio_build_amazon.sh
```diff
@@ -115,7 +115,7 @@ fi
 printf "\\n"
 
 ### clean up force build before starting
-if [ $FORCE_BUILD ];then
+if $FORCE_BUILD;then
```
heifner: I thought that only worked if `$FORCE_BUILD` was `true` or `false` not 1 or 0 which `eosio_build.sh` uses. Should we not use `[]` for most portable use?

NorseGaud: I got confused since I moved all of the new scripts I've been writing to use booleans. This actually exposes the reason I dislike using integers: [ $FORCE_BUILD ] if FORCE_BUILD=0 or 1, or even true or false, will ALWAYS return true:
```
 nathan.pierce@npierce-mbp  ~/BLOCKONE/eos   minor-fixes  bash -c "TEST=0; if [ \$TEST ]; then echo 'GOT IT'; fi"
GOT IT
 nathan.pierce@npierce-mbp  ~/BLOCKONE/eos   minor-fixes  bash -c "TEST=1; if [ \$TEST ]; then echo 'GOT IT'; fi"
GOT IT
 nathan.pierce@npierce-mbp  ~/BLOCKONE/eos   minor-fixes  bash -c "TEST=true; if [ \$TEST ]; then echo 'GOT IT'; fi"
GOT IT
 nathan.pierce@npierce-mbp  ~/BLOCKONE/eos   minor-fixes  bash -c "TEST=false; if [ \$TEST ]; then echo 'GOT IT'; fi"
GOT IT
 nathan.pierce@npierce-mbp  ~/BLOCKONE/eos   minor-fixes  bash -c "TEST=; if [ \$TEST ]; then echo 'GOT IT'; fi"
 nathan.pierce@npierce-mbp  ~/BLOCKONE/eos   minor-fixes                                                  
```
Fortunately, if it's not set (which it doesn't get set by default right now), we'll be ok.

NorseGaud: Sorry about that. Reverting now.

NorseGaud:
I'm building new images for this since the existing ubuntu one doesn't have zlib properly installed.

spoonincode:
Should look in to removing zlib from the scripts. AFAIK there is no reason for it

PR #7297, opened 2019-05-07
added sudo -E to installs for Ben on automation + zlib if fix

NorseGaud:
-  Fixed if for zlib
- Added sudo -E for proxies to work (Ben on Automation needs this)


NorseGaud:
@heifner Sorry for the second PR for the same thing. I royally screwed up the previous branch and figured I'd just start off fresh to be safe.

spoonincode:
Is setting the proxy in yum's config file not workable?

NorseGaud:
@benbenben1010

benbenben1010:
The -E is nice because it allows us to use the same Dockerfile in the sandbox (which requires the proxy) and outside the sandbox (which doesn't) by conditionally passing in environment variables to the docker build command.

That being said, building in the sandbox has been de-prioritized, so for now, I don't have a strong opinion either way.

Issue #7232, opened 2019-04-27
linux build scripts "which/c++: command not found"


NorseGaud:
./bootstrap.sh: line 27: which: command not found
./scripts/eosio_build.sh: line 231: which: command not found
./scripts/eosio_build.sh: line 247: c++: command not found

NorseGaud:
I will address in new bash scripts for BATS tests

PR #7317, opened 2019-05-09
Various improvements for 1.8.x

NorseGaud:
- https://github.com/EOSIO/eos/issues/7232: New compiler check code
- Using Anka
- New ARCH and OS NAME setting


NorseGaud:
Because I based it on develop at first (by accident), the PULL_REQUEST_BRANCH env is set to develop and won't change. I'll close this and re-open

Issue #7260, opened 2019-05-03
build script on clean install has compiler detection error


spoonincode:
When running the build script with no options on a fresh Ubuntu 18.04 install it builds the clang8 compiler. That is not expected without the -P option. The reason for this is due to a couple errors in the build script and ordering of detection.

The script has a couple logic defects that should probably be fixed like so:
```diff
--- a/scripts/eosio_build.sh
+++ b/scripts/eosio_build.sh
@@ -253,8 +253,8 @@ if [ $PIN_COMPILER ]; then
    BUILD_CLANG8=true
    CPP_COMP=${OPT_LOCATION}/clang8/bin/clang++
    CC_COMP=${OPT_LOCATION}/clang8/bin/clang
-elif $NO_CPP17; then
-   if [ $NONINTERACTIVE -eq 0 ]; then
+elif [ $NO_CPP17 ]; then
+   if [ $NONINTERACTIVE -eq 1 ]; then
       BUILD_CLANG8=true
       CPP_COMP=${OPT_LOCATION}/clang8/bin/clang++
       CC_COMP=${OPT_LOCATION}/clang8/bin/clang
```

But this doesn't fix the underlying problem: this compiler check happens before a compiler is possibly installed by the eosio_build_ubuntu.sh (or other platform) script. So by making this change above it will prompt users to install pinned compiler when a platform compiler would pass muster any other time.

But there's more: the eosio_build_ubuntu.sh script will end up installing gcc still. This means the next time the user runs the script with no options (as they probably would, since that's what they did the first time), the script will try doing a non-pinned build. And of course that won't work -- giving bizarre errors -- because you'd need to give a -f to switch between pinned vs nonpinned.

spoonincode:
This error also seems to prevent eosio_build.sh from working on 16.04 when no arguments are passed

NorseGaud:
I noticed the same with my tests in #7263... Strange... I'm collecting failure logs for each distro so we can swarm on what's causing it.

spoonincode:
#7263 still has the faulty if statement I documented in this defect, that's why it too has problems

NorseGaud:
There is a bit of a chicken and egg problem here. You can't compile cmake without clang/gcc-c++ installed, so you can't compile clang8. That's why the user is prompted that their version will not work and -y will answer yes to installing our clang8.

[...]

PR #7318, opened 2019-05-09
Various improvements

NorseGaud:
- https://github.com/EOSIO/eos/issues/7232: New compiler check code
- Using Anka
- New ARCH and OS NAME setting
- https://github.com/EOSIO/eos/issues/7260: Code to handle warning users about previous install
- export added so we can simplify NONINTERACTIVE and a few other vars


scripts/eosio_build_amazon.sh
```diff
@@ -183,7 +152,8 @@ if [ "$BUILD_CLANG8" = "true" ]; then
       && rm -rf ${TMP_LOCATION}/clang8 \
       && cd ../.. \
       || exit 1
-
+      export CXX=$CPP_COMP
```
arhag: I don't think these two lines shouldn't be here. (Same for the other OSes.)

scripts/eosio_build.sh
```diff
@@ -217,63 +227,80 @@ if [ $STALE_SUBMODS -gt 0 ]; then
...
-else
-   ### gcc version 7
-   if [[ `c++ -dumpversion | cut -d '.' -f 1` -lt 7 ]]; then
-      NO_CPP17=true
+      ## Check for c++ version 7 or higher
+      [[ $( $(which $CXX) -dumpversion | cut -d '.' -f 1 ) -lt 7 ]] && export NO_CPP17=true
    fi
+elif $PIN_COMPILER; then
+   export BUILD_CLANG8=true
+   export CPP_COMP=$CLANG8_ROOT/bin/clang++
```
arhag: I think this can be:
```
CXX=$CLANG8_ROOT/bin/clang++
```

scripts/eosio_build.sh
```diff
@@ -217,63 +227,80 @@ if [ $STALE_SUBMODS -gt 0 ]; then
...
+   export CC_COMP=$CLANG8_ROOT/bin/clang
```
arhag: I think this can be:
```
CC=$CLANG8_ROOT/bin/clang
```

scripts/eosio_build.sh
```diff
@@ -217,63 +227,80 @@ if [ $STALE_SUBMODS -gt 0 ]; then
...
+if $NO_CPP17; then
+   while true; do
+      echo "Unable to find C++17 support!"
+      echo "If you already have a C++17 compiler installed or would like to install your own, export CXX to point to the compiler of your choosing."
+      [[ $NONINTERACTIVE == false ]] && read -p "Do you wish to download and build C++17? (y/n)? " PROCEED
+      case $PROCEED in
+            "" ) echo "What would you like to do?";;
+            0 | true | [Yy]* )
+               export BUILD_CLANG8=true
+               export CPP_COMP=$CLANG8_ROOT/bin/clang++
```
arhag: Same for these three lines as the comments above.

scripts/eosio_build.sh
```diff
@@ -217,63 +227,80 @@ if [ $STALE_SUBMODS -gt 0 ]; then
...
-   if [[ `c++ -dumpversion | cut -d '.' -f 1` -lt 7 ]]; then
-      NO_CPP17=true
+      ## Check for c++ version 7 or higher
+      [[ $( $(which $CXX) -dumpversion | cut -d '.' -f 1 ) -lt 7 ]] && export NO_CPP17=true
    fi
+elif $PIN_COMPILER; then
+   export BUILD_CLANG8=true
+   export CPP_COMP=$CLANG8_ROOT/bin/clang++
+   export CC_COMP=$CLANG8_ROOT/bin/clang
+   export PATH=$CLANG8_ROOT/bin:$PATH
```
arhag: I don't think this is necessary.

scripts/eosio_build_amazon.sh
NorseGaud: They set CXX to use the new clang: LLVM, boost, etc, look for CXX and without it the cpp/c++ version on the machine is used instead of the new clang8.

scripts/eosio_build.sh
NorseGaud: That will cause cmake to not install (https://github.com/EOSIO/eos/issues/7260). That's why I set CXX later, after CLANG8 is built.

scripts/eosio_build.sh
NorseGaud: Sorry --Do you mean $PATH?

PR #7343, opened 2019-05-13
correct signed mismatch warning in chain controller - 1.8

spoonincode:
## Change Description

cleans up a handful of gcc7 warnings coming from this function. I elected to stick with auto

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -515,7 +515,7 @@ struct controller_impl {
 
    void init(std::function<bool()> shutdown, const snapshot_reader_ptr& snapshot) {
       // Setup state if necessary (or in the default case stay with already loaded state):
-      auto lib_num = 1;
+      auto lib_num = UINT32_C(1);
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -515,7 +515,7 @@ struct controller_impl {
 
    void init(std::function<bool()> shutdown, const snapshot_reader_ptr& snapshot) {
       // Setup state if necessary (or in the default case stay with already loaded state):
-      auto lib_num = 1;
+      auto lib_num = UINT32_C(1);
```
arhag: Since everywhere `lib_num` is being used expects it to be `uint32_t` anyway, I suggest just defining this variable as `uint32_t` explicitly rather than preserving the existing (unintentional) behavior of automatically deducing the `uint_least32_t` type.

PR #7377, opened 2019-05-17
Allow aliases of variants in ABI

### LINKED ISSUE #7254: ABI Serializer throws on type alias to variant
Using a type alias for a variant eg:

```json
{
    "version": "eosio::abi/1.1",
    "types": [
        {
            "new_type_name": "my_things",
            "type": "variant_many_things"
        }
    ],
    "structs": [
        {
            "name": "foo",
            "fields": [
                {
                    "name": "things",
                    "type": "my_things"
                }
            ]
        }
    ],
    "variants": [
        {
            "name": "variant_many_things",
            "types": [
                "string",
                "uint8",
                "uint16",
                "uint32",
                "uint64"
            ]
        }
    ]
}
```
causes: 

```Exception Details: 3015004 invalid_type_inside_abi: The type defined in the ABI is invalid
invalid type variant_block_nodes
    {"type":"variant_block_nodes"}
    thread-0  abi_serializer.cpp:122 set_abi
```

The error thrown is caused by `_is_type` not finding the variant because they are populated after type aliases are resolved.

https://github.com/EOSIO/eos/blob/3fddb727b8f3615917707281dfd3dd3cc5d3d66d/libraries/chain/abi_serializer.cpp#L118-L137

https://github.com/EOSIO/eos/blob/3fddb727b8f3615917707281dfd3dd3cc5d3d66d/libraries/chain/abi_serializer.cpp#L206-L214

wanderingbort:
## Change Description
This resolves #7254

We had a redundant validation that attempted to prevent aliasing to an unknown type however, this was done before variants were loaded so, it would reject an alias of a variant.  `::validate` will perform this validation after all types are known so, this assert was not needed

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -119,7 +119,6 @@ namespace eosio { namespace chain {
          structs[st.name] = st;
 
       for( const auto& td : abi.types ) {
-         EOS_ASSERT(_is_type(td.type, ctx), invalid_type_inside_abi, "invalid type ${type}", ("type",td.type));
          EOS_ASSERT(!_is_type(td.new_type_name, ctx), duplicate_abi_type_def_exception, "type already exists", ("new_type_name",td.new_type_name));
          typedefs[td.new_type_name] = td.type;
       }
--- a/unittests/abi_tests.cpp
+++ b/unittests/abi_tests.cpp
@@ -1955,8 +1955,8 @@ BOOST_AUTO_TEST_CASE(abi_type_redefine)
    }
    )=====";
 
-   auto is_type_exception = [](fc::exception const & e) -> bool { return e.to_detail_string().find("invalid type") != std::string::npos; };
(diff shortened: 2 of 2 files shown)


wanderingbort:
We may want to backport this to supported versions as it is preventing tooling based on those versions from accepting well-formed ABI's from CDT

libraries/chain/abi_serializer.cpp
```diff
@@ -119,7 +119,6 @@ namespace eosio { namespace chain {
          structs[st.name] = st;
 
       for( const auto& td : abi.types ) {
-         EOS_ASSERT(_is_type(td.type, ctx), invalid_type_inside_abi, "invalid type ${type}", ("type",td.type));
          EOS_ASSERT(!_is_type(td.new_type_name, ctx), duplicate_abi_type_def_exception, "type already exists", ("new_type_name",td.new_type_name));
```
heifner: Can you add a unittest please.

wanderingbort: Yep, and fix the one I broke....

PR #7379, opened 2019-05-17
transaction deadline cleanup

heifner:
## Change Description

- Fixed issue in `transaction_metadata` `start_key_recovery` where exceptions when `next()` provided didn't call `next()` causing silent dropping of exceptions. This produced a confusing unable to connect error to `cleos` instead of a deadline exception error. 
- Added elapsed time to transaction deadline exception strings so users can see how long the transaction executed. 
-  Moved `CATCH_AND_CALL` macro to `exceptions.hpp` and handle `bad_alloc` in `EOS_RETHROW_EXCEPTIONS` and `EOS_CAPTURE_AND_RETHROW`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/transaction_metadata.cpp
+++ b/libraries/chain/transaction_metadata.cpp
@@ -16,6 +16,7 @@ recovery_keys_type transaction_metadata::recover_keys( const chain_id_type& chai
 
    g.unlock();
    // shared_keys_future not created or different chain_id
+   wlog( "recovering keys in recover_keys" );
    auto recovered_pub_keys = std::make_shared<flat_set<public_key_type>>();
    const signed_transaction& trn = _packed_trx->get_signed_transaction();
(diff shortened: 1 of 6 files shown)


libraries/chain/transaction_metadata.cpp
```diff
@@ -16,6 +16,7 @@ recovery_keys_type transaction_metadata::recover_keys( const chain_id_type& chai
 
    g.unlock();
    // shared_keys_future not created or different chain_id
+   wlog( "recovering keys in recover_keys" );
```
b1bart: This seems like a `debug` log at best.

heifner: Could change to debug. If developer called `start_key_recovery` as they should then you will never hit this line.

heifner: Changed to `debug`.

PR #7380, opened 2019-05-17
Producer incoming-transaction-queue-size-mb

heifner:
## Change Description

- Add `producer_plugin` `incoming-transaction-queue-size-mb` so that incoming transactions can be rejected with `tx_resource_exhaustion exception` when queue is full.
- This is useful for a new transaction only connection option that is in work for the net_plugin.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -604,6 +604,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       void schedule_delayed_production_loop(const std::weak_ptr<producer_plugin_impl>& weak_this, const block_timestamp_type& current_block_time);
 };
 
+uint64_t producer_plugin_impl::_max_incoming_transaction_queue_size = 0;
+
 void new_chain_banner(const eosio::chain::controller& db)
 {
    std::cerr << "\n"
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -429,7 +429,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          }
       }
 
-      class incoming_transactions {
+      class incoming_transaction_queue {
          uint64_t size_in_bytes = 0;
          std::deque<std::tuple<transaction_metadata_ptr, bool, next_function<transaction_trace_ptr>>> _incoming_transactions;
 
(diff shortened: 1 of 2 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -428,7 +429,38 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          }
       }
 
-      std::deque<std::tuple<transaction_metadata_ptr, bool, next_function<transaction_trace_ptr>>> _pending_incoming_transactions;
+      class incoming_transaction_queue {
+         uint64_t size_in_bytes = 0;
+         std::deque<std::tuple<transaction_metadata_ptr, bool, next_function<transaction_trace_ptr>>> _incoming_transactions;
+
+      private:
+         uint64_t calc_size( const transaction_metadata_ptr& trx ) {
```
b1bart: this could be `const` or `static`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -201,6 +201,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
       std::atomic<int32_t>                                      _max_transaction_time_ms; // modified by app thread, read by net_plugin thread pool
       fc::microseconds                                          _max_irreversible_block_age_us;
+      static uint64_t                                           _max_incoming_transaction_queue_size;
```
b1bart: it seems odd that this is `static`.  it is only used (other than configuration) inside the non-static member of type `incoming_transaction_queue`.  Can it just be stored there?  meaning configuration pushes it into that class.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -484,53 +510,62 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
 
-         auto deadline = fc::time_point::now() + fc::milliseconds(_max_transaction_time_ms);
-         bool deadline_is_subjective = false;
-         const auto block_deadline = calculate_block_deadline(block_time);
-         if (_max_transaction_time_ms < 0 || (_pending_block_mode == pending_block_mode::producing && block_deadline < deadline) ) {
-            deadline_is_subjective = true;
-            deadline = block_deadline;
-         }
+            const auto& id = trx->id();
+            if( fc::time_point( trx->packed_trx()->expiration()) < block_time ) {
```
b1bart: we used to do this check first, is there concern that `block_time` may move backwards? otherwise this seems like it may store an expired transaction in the `_pending_incoming_transaction` queue

b1bart: oh, I'm incorrect it was always in this order.  But perhaps we should swap these now such that we check expiration before storing on the pending list.

Same with other early outs like `is_known_unexpired_transaction` .  I am relatively sure that even when faced with forking conditions any transaction that returns `true` will end up on the unapplied transactions list and therefore not need to be duplicated in our pending queue.

plugins/producer_plugin/producer_plugin.cpp
heifner: I made it static so the embedded class has direct access to it. I'll plumb it in.

plugins/producer_plugin/producer_plugin.cpp
heifner: Made it static.

plugins/producer_plugin/producer_plugin.cpp
heifner: I left it that way because `pending_block_time` is not available if `!chain.is_building_block()`. Changed it to use either pending block time or head block time.

Issue #7403, opened 2019-05-24
Weighted-Threshold-Multi-Signature (WTMsig) Block Production


b1bart:
## Background

Block producers should provide high availability for their core service of extending the blockchain (aka producing blocks). In order to achieve this they need support for redundant infrastructure in a way that safely maintains the other invariants that are expected of them. 
 For instance, not double signing blocks.  This is the first of many features that attempt to enable a safer paradigm for highly available block production.

### Goals 

1. Allow BPs to have redundant block production infrastructure that does not share private key material
2. Allow Block Production to be resilient to the compromise of some portion of this infrastructure by a malicious actor

## Consensus upgrade feature

In service of the above goals, the `WTMSIG_BLOCK_SIGNATURES` feature will transition the `block_header` and `signed_block` such that they allow for the declaration and use of a full **key-only** weighted-threshold multi-signature authority that is similar to the one used in account permissions.  

We have discussed alternatives, such as 1-of-N multisig or M-of-N multisig. We felt that 1-of-N did not actually achieve goals (1) and (2) as each of the keys represents a valuable target for forging blocks ( violating (2) ) and any system that would attempt to mitigate this would effectively require sharing private key data ( violating (1) ). M-of-N was sufficient, but as our authority structure for validating transaction signatures is based on WTMSig and WTMSig is a super set of the functionality of M-of-N, we felt that the incremental reduction in technical complexity did not outweigh the additional cognitive complexity for users who must now understand multiple schemes for multisig.

In this scheme, a Block Producer can define a threshold and a set of weighted public keys, signing a block with as many signatures as necessary to satisfy the threshold with the sum of associated weights.

#### Upon activation:

* validation logic will enforce that the existing `new_producers` field in `block_header` MUST be empty for all new blocks regardless of source.
* A new `block_header_extension` will be introduced that is used to announce new producer * schedules based on the new `producer_authority` type.
* a new intrinsic `set_proposed_producers_ex` will be valid and linkable
   * this intrinsic allows the legacy format to be used as well as a new format based on `producer_authority`
   * **note** the existing `set_proposed_producers` intrinsic will remain valid and only support the legacy format (which will be upgraded to a full `producer_authority` schedule automatically)

arhag:
Resolved by #7404.

Issue #7012, opened 2019-03-28
webauthn key & signature type


spoonincode:
Finish adding webauthn key & signatures to eosio. There are still some unresolved questions (now solved; answers in bold) that came out of the prototype:

* where/how/if credential ID is stored - **will not be stored as part of consensus; public storage discouraged regardless**
* JSON parsing vs not parsing - **We need to parse JSON, the "tricks" we discussed internally will not work due to item 4 on this list**
* Compression of JSON or not - **No**
* Validation of HTTPS origin in signature or not (a public key would have an origin it is scoped to) - **Yes we need to validate this**
* Validation of user presence and/or user verification in signature (a public key would state some minimum level of verification requirement) - **Doing this and maintaining the "symmetry" of recovered keys to the public keys in the authority is unwieldy, but it's probably useful to have this ability so will do it**

Ideally all of these are answered before commencing work on the production implementation.

jafri:
If credential ID is not stored as part of consensus or public storage, how will credentials be synced between nodes?

spoonincode:
>If credential ID is not stored as part of consensus or public storage, how will credentials be synced between nodes?

The public key that a credential ID references is part of consensus. It will be a `PUB_WA_...` thing. But the storage of the credential ID that is required to sign via the private key associated with that public key is left outside the scope of consensus.

nksanthosh:
EOSJS related changes will be tracked in https://github.com/EOSIO/eosjs/issues/537

spoonincode:
#7421

PR #7404, opened 2019-05-24
wtmsig block production

b1bart:
## Change Description

This change implements #7403 

This now encompasses (and fixes conflicts from) #7421 
 
Remaining work: 
- [x] unit tests for the protocol feature activation
- [ ] producer plugin work to allow BPs to use the new feature 

## Consensus Changes
- [x] Consensus Changes

see #7403

## API Changes
- [x] API Changes

new deployments of the API will emit  `producer_authority` 's for all block signing data regardless of feature activation (it will not backport to the legacy structure)

specifically the `get_producers` API will no longer guarantee the result has a field `producer_key`.  Instead it will return `producer_authority`.  In instances where the `producer_authority` appears to be a canonical "upsert" of the legacy single key, this API will still emit the `producer_key` field as expected.  In those instances, the `producer_authority` field will contain a 1-of-1 wtmsig where both the threshold and weight are exactly `1`

## Documentation Additions
- [x] Documentation Additions

need to document `set_proposed_producers_ex` , see TBD PR in eosio.cdt

## Other notes

~this PR is likely to conflict with #7012 as that will also bump the snapshot version.  This PR **should** give way to a PR regarding that feature~ Resolved as #7012 was merged to `develop` and this PR is caught up.

this PR includes #7401 as it introduces an otherwise breaking change to the in-memory format


b1bart:
This is now up for review, I've decided not to include `producer plugin work to allow BPs to use the new feature` in this PR as it can safely be added in a future PR.

libraries/chain/wasm_interface.cpp
```diff
@@ -175,29 +175,79 @@ class privileged_api : public context_aware_api {
...
-         return context.control.set_proposed_producers( std::move(producers) );
+
+         return context.control.set_proposed_producers( std::forward<vector<producer_authority>>(producers) );
+      }
+
+      int64_t set_proposed_producers( array_ptr<char> packed_producer_schedule, size_t datalen) {
+         datastream<const char*> ds( packed_producer_schedule, datalen );
+         vector<producer_authority> producers;
+
+//         if ( context.control.is_builtin_activated( builtin_protocol_feature_t::wtmsig_block_signatures )) {
```
heifner: Remove commented out code.

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -2,57 +2,334 @@
...
-      public_key_type   block_signing_key;
+   struct block_signing_authority_v0 {
+      uint32_t                    threshold;
+      vector<key_weight>          keys;
+
+      bool key_is_relevant( const public_key_type& key ) const {
+         return std::find_if(keys.begin(), keys.end(), [&key](const auto& kw){
+            return kw.key == key;
+         }) != keys.end();
+      }
```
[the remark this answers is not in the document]
b1bart: the type of `keys` is a `key_weight` so, this is needed to inspect the contained key

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -2,57 +2,334 @@
...
+      auto shared_convert( const T& src ) {
+         return shared_converter<std::decay_t<decltype(src)>>::convert(src);
+      }
+
+      template<>
+      struct shared_converter<block_signing_authority_v0> {
+         static shared_block_signing_authority_v0 convert (chainbase::allocator<char> alloc, const block_signing_authority_v0& src) {
+            shared_block_signing_authority_v0 result(alloc);
+            result.threshold = src.threshold;
+            result.keys.clear();
```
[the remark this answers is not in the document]
b1bart: this is potentially still needed to address the odd edge case with vectors and the interprocess allocator.   We should investigate in a separate task whether this bug is still present in boost 1.70+

libraries/chain/controller.cpp
```diff
@@ -2812,17 +2836,24 @@ const producer_schedule_type&    controller::active_producers()const {
...
       return my->pending->_block_stage.get<completed_block>()._block_state->pending_schedule.schedule;
 
    if( my->pending->_block_stage.contains<assembled_block>() ) {
-      const auto& np = my->pending->_block_stage.get<assembled_block>()._unsigned_block->new_producers;
-      if( np )
-         return *np;
+      if (is_builtin_activated(builtin_protocol_feature_t::wtmsig_block_signatures)) {
+         auto exts = my->pending->_block_stage.get<assembled_block>()._unsigned_block->validate_and_extract_header_extensions();
+         if (exts.count(producer_schedule_change_extension::extension_id()))
+            return exts.lower_bound(producer_schedule_change_extension::extension_id())->second.get<producer_schedule_change_extension>();
```
arhag: Dangling reference.

arhag: Can't you return `*my->pending->_block_stage.get<assembled_block>()._new_producer_authority_cache` (assuming `has_value`) like you do below whether `WTMSIG_BLOCK_SIGNATURES` has activated or not? That would resolve the dangling reference issue and be more efficient.

libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -559,6 +562,10 @@ namespace eosio { namespace chain {
                                     3170009, "Snapshot Finalization Exception" )
       FC_DECLARE_DERIVED_EXCEPTION( invalid_protocol_features_to_activate,  producer_exception,
                                     3170010, "The protocol features to be activated were not valid" )
+      FC_DECLARE_DERIVED_EXCEPTION( no_block_signatures,  producer_exception,
+                                    3170011, "The signer returned no valid block signatures" )
+      FC_DECLARE_DERIVED_EXCEPTION( unsupported_multiple_block_signatures,  producer_exception,
+                                    3170011, "The signer returned multiple signatures but that is not supported" )
```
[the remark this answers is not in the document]
arhag: This was already resolved by commit 510118a.

libraries/chain/block_state.cpp
```diff
@@ -3,26 +3,131 @@
...
+         if (!result.additional_signatures.empty()) {
+            EOS_ASSERT(wtmsig_enabled, block_validate_exception,
+                       "Block has multiple signatures before activation of WTMsig Block Signatures");
+
+            // as an optimization we don't copy this out into the legitimate extension structure as it serializes
+            // the same way as the vector of signatures
+            static_assert(fc::reflector<additional_block_signatures_extension>::total_member_count == 1);
+            static_assert(std::is_same_v<decltype(additional_block_signatures_extension::signatures), std::vector<signature_type>>);
+
+            b.block_extensions.emplace_back(
```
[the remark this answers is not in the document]
arhag: Resolved by an even better approach of using the new `emplace_extension` helper that ensures the extensions are in a valid order regardless of order of `emplace_extension` calls.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -243,28 +243,6 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             active_producers.insert(p.producer_name);
          }
```
arhag: Also, remove everything else above this line within this function. And remove the member variables `_last_signed_block_time`, `_start_time`, and `_last_signed_block_num` from `producer_plugin_impl`.

Although maybe we still need to keep `_start_time` after correcting the watermark logic, since that effectively acts as the time watermark for all locally configured producers even if an entry doesn't yet exist for them in `_producer_watermarks`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -243,28 +243,6 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-                     _last_signed_block_num  = bsp->block_num;
-
-   //                  ilog( "${n} confirmed", ("n",name(producer)) );
-                     _self->confirmed_block( { bsp->id, d, producer, sig } );
-                  }
-               }
-            }
-         } ) );
-
          // simplify handling of watermark in on_block
```
arhag: It seems like we need `_producer_watermarks` to map to both the watermark block number and the watermark block time. We would update the watermark time both here and in `produce_block` (unless it would cause it to decrease).

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -98,27 +98,33 @@ namespace eosio { namespace chain {
...
+
       uint32_t                    threshold = 0;
       vector<key_weight>          keys;
 
-      bool key_is_relevant( const public_key_type& key ) const {
-         return std::find_if(keys.begin(), keys.end(), [&key](const auto& kw){
-            return kw.key == key;
-         }) != keys.end();
+      template<typename Op>
+      void for_each_key( Op op ) const {
```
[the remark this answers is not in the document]
b1bart: this is modeled after `std::for_each` which takes the `UnaryFunction` by value.  [reference](https://en.cppreference.com/w/cpp/algorithm/for_each)

I'm willing to change this to a universal reference and do perfect forwarding however, it seems that if you are always using a lambda that compilers do the same thing.  If our Op was a standard function then we would create a modest amount of overhead passing by value: https://godbolt.org/z/Xa6bJh

plugins/producer_plugin/producer_plugin.cpp
b1bart: For the context of this PR I'm going to ignore this feedback as it should be placed into an issue and resolved in the context of working on watermark safety.

libraries/chain/apply_context.cpp
```diff
@@ -372,8 +372,8 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
                      ("expected", trx_context.id)("actual", context.sender_trx_id)
          );
       } else {
```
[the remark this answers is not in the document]
b1bart: Nope, but I'd be willing to turn this feedback into an issue for another PR :)

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1691,7 +1691,7 @@ read_only::get_producers_result read_only::get_producers( const read_only::get_p
    for (auto p : db.active_producers().producers) {
       fc::variant row = fc::mutable_variant_object()
          ("owner", p.producer_name)
-         ("producer_key", p.block_signing_key)
```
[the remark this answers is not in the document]
b1bart: opted to maintain legacy results when I can by detecting a canonical 1-of-1 wtmsig and injecting the old `producer_key` field.  Noted in the comments as well so that it can be labeled as "deprecated" or something similar in the API docs

plugins/producer_plugin/producer_plugin.cpp
b1bart: kept the `_start_time` in pending a deeper review.

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -2,57 +2,334 @@
...
+      };
+
+      template<>
+      struct shared_converter<shared_block_signing_authority_v0> {
+         static block_signing_authority_v0 convert (const shared_block_signing_authority_v0& src) {
+            block_signing_authority_v0 result;
+            result.threshold = src.threshold;
+            result.keys.reserve(src.keys.size());
+            for (const auto& k: src.keys) {
+               result.keys.push_back(k);
```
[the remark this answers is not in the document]
b1bart: As this was the pattern established in #7421 I'm going to promote this to an issue to handle outside of this PR

libraries/chain/block.cpp
```diff
@@ -0,0 +1,64 @@
+#include <eosio/chain/block.hpp>
```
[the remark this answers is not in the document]
b1bart: I'm going to promote this to a global issue

plugins/state_history_plugin/state_history_plugin_abi.cpp
```diff
@@ -311,6 +311,24 @@ extern const char* const state_history_plugin_abi = R"({
                 { "type": "producer_key[]", "name": "producers" }
             ]
         },
+        {
+            "name": "block_signing_authority_v0", "fields": [
+                { "type": "uint32", "name": "weight" },
```
tbfleming: `threshold`?

unittests/producer_schedule_tests.cpp
```diff
@@ -503,4 +503,224 @@ BOOST_AUTO_TEST_CASE( producer_watermark_test ) try {
...
+   BOOST_CHECK_EQUAL( control->pending_producers().version, 0u );
+   produce_block(); // Starts new block which promotes the proposed schedule to pending
+   BOOST_CHECK_EQUAL( control->pending_producers().version, 1u );
+   BOOST_CHECK_EQUAL( true, compare_schedules( sch1, control->pending_producers() ) );
+   BOOST_CHECK_EQUAL( control->active_producers().version, 0u );
+   produce_block();
+   produce_block(); // Starts new block which promotes the pending schedule to active
+   BOOST_CHECK_EQUAL( control->active_producers().version, 1u );
+   BOOST_CHECK_EQUAL( true, compare_schedules( sch1, control->active_producers() ) );
+   produce_blocks(18);
```
arhag: For example this could be replaced by `produce_empty_blocks_until( *this, N(bob), N(alice) );`.

You could even put a limit on the number of blocks to produce to avoid getting into an infinite loop if something went wrong earlier with setting up the producer schedule: `produce_empty_blocks_until( *this, N(bob), N(alice), 36 );`. In that case, one of the `BOOST_REQUIRE_EQUAL` assertions would fail and the test would terminate.

Or even better, since the `produce_empty_blocks_until` returns true iff the desired condition is met, you can avoid the next two `BOOST_REQUIRE_EQUAL` lines as well and just do the following:
```cpp
BOOST_REQUIRE( produce_empty_blocks_until( *this, N(bob), N(alice), 36 ) );
```

unittests/producer_schedule_tests.cpp
```diff
@@ -503,4 +503,224 @@ BOOST_AUTO_TEST_CASE( producer_watermark_test ) try {
...
```
arhag: See earlier comment about `BOOST_REQUIRE( produce_empty_blocks_until( *this, N(bob), N(alice), 36 ) );`.

unittests/producer_schedule_tests.cpp
b1bart: I ended up deciding that I did not need to re-test the schedule promotion logic in all these cases.  What those tests should show is that 1-of-n and m-of-n wtmsig block signing authorities are valid.   Most of the logic was replace by a new version of `produce_empty_blocks_until` that will produce blocks until all producers in a `std::set` have had a turn (or until a threshold of blocks is reached).  This makes those tests a lot shorter.  It's just setting a schedule and guaranteeing that in the first N blocks those producers can produce valid blocks.

piqjen-1xofFi-sikxyh (commented):
@b1bart

PR #7413, opened 2019-05-24
--replace-producer-keys

tbfleming:
## Change Description

`--replace-producer-keys {pubkey}` replaces the block signing key for all active produces with pubkey. It also resets their active and owner permissions to that key.

To create a snapshot forked off an existing chain:
* Start up nodeos disconnected from the p2p net. Use these options:
  * `-e`
  * `-p {producer}` for each active producer
  * `--replace-producer-keys {pubkey}`
* Quickly use `cleos system regproducer` on all the active producers
* Stop nodeos
* Restart nodeos with same options as above
* Create snapshot

The newly created snapshot is restorable, but not replayable. For best results, do not use a block log when restoring the snapshot.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

See above

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2211,4 +2211,31 @@ const flat_set<account_name> &controller::get_resource_greylist() const {
    return  my->conf.resource_greylist;
 }
 
+void controller::replace_producer_keys( const public_key_type& key ) {
+   ilog("Replace producer keys with ${k}", ("k", key));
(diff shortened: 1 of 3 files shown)


jeffreyssmith2nd:
The chainlib part looks to have been handled in #8929, do we still need the chain_plugin part of this PR?

brianjohnson5972:
@tbfleming Can you respond to @jeffreyssmith2nd 's last question?  Just wanting to close this and get it out of the way if we can.

Issue #7396, opened 2019-05-24
cleos needs —version 


tbfleming:

ekkis:
yes, and can we please output something rational like 1.7.3 instead of Build version: 448287d5?

johndebord:
@tbfleming @ekkis Fixed with [PR #7414](https://github.com/EOSIO/eos/pull/7414).

PR #7414, opened 2019-05-25
Add `cleos` version format changes

johndebord:
## Change Description

In reference to [Issue #7396](https://github.com/EOSIO/eos/issues/7396), this PR alters the format of printing the `cleos` version in a more intuitive way; in addition, a new `cleos version` sub-command has been added as well.

Examples:
```
~/eos/build/bin $ ./cleos version client
1.9.0-develop
```

```
~/eos/build/bin $ ./cleos version full
1.9.0-develop-7de45825-dirty
```

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

New `cleos version` sub-command: `cleos version full`.

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/programs/cleos/CMakeLists.txt
+++ b/programs/cleos/CMakeLists.txt
@@ -16,9 +16,16 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../../.git)
     execute_process(
       COMMAND ${GIT_EXECUTABLE} rev-parse --short=8 HEAD
       WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/../.."
-      OUTPUT_VARIABLE "cleos_BUILD_VERSION"
+      OUTPUT_VARIABLE "cleos_BUILD_HASH"
       ERROR_QUIET
       OUTPUT_STRIP_TRAILING_WHITESPACE)
+    execute_process(
+      COMMAND ${GIT_EXECUTABLE} describe --tags --dirty
+      WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/../.."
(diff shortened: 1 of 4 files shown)


programs/cleos/CMakeLists.txt
```diff
@@ -16,9 +16,16 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/../../.git)
     execute_process(
       COMMAND ${GIT_EXECUTABLE} rev-parse --short=8 HEAD
       WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/../.."
-      OUTPUT_VARIABLE "cleos_BUILD_VERSION"
+      OUTPUT_VARIABLE "cleos_BUILD_HASH"
       ERROR_QUIET
       OUTPUT_STRIP_TRAILING_WHITESPACE)
+    execute_process(
```
spoonincode: this approach is only evaluated during cmake generation so it's not as sensitive as it should be. Compare to how it was implemented for nodeos:
```
$ ./programs/nodeos/nodeos --version
v1.8.0-rc1-501-gafa686d8c
$ ./programs/cleos/cleos version full
1.9.0-develop-afa686d8
$ echo "// spoon was here" >> ../src/programs/cleos/main.cpp 
$ ninja
...
$ ./programs/nodeos/nodeos --version
v1.8.0-rc1-501-gafa686d8c-dirty
$ ./programs/cleos/cleos version full
1.9.0-develop-afa686d8
```

I think we should consider refactoring out the more advanced appbase/nodeos code in a common manner so that it can be used by both nodeos as well as non-appbase based targets like cleos. Probably never belonged in appbase in the first place. Plus that will reduce duplication. And if we're lucky it'll resolve that annoying double-tap we get from cmake.

PR #7419, opened 2019-05-27
fc varint unpacking out of bound check

### LINKED ISSUE #7378: unpacking variable-length integer should throw if number is out of bounds
When unpacking `fc::unsigned_int` and `fc::signed_int`, the implementation should verify that the serialized variable-length integer is within the appropriate bounds (within bounds of `uint32_t` for `fc::unsigned_int` and `int32_t` for `fc::signed_int`). If not, it should throw an exception.

taokayan:
fix #7378 

## Change Description
out of bound check for unpacking fc::signed_int and fc::unsigned_int. add test cases

## Consensus Changes
N/A

## API Changes
N/A

## Documentation Additions
N/A


unittests/misc_tests.cpp
```diff
@@ -1128,6 +1128,91 @@ BOOST_AUTO_TEST_CASE(stable_priority_queue_test) {
   } FC_LOG_AND_RETHROW()
 }
 
+BOOST_AUTO_TEST_CASE(varint_unpack) {
```
b1bart: `libraries/fc` has its own unit tests and this should probably be there with other basic unit tests for varint encoding.

taokayan: sure will fix it.

NorseGaud:
You'll need to merge in changes from develop too. Looks like you're using older build scripts that don't install c++ properly and allow it to build on our new base images.

taokayan:
> Please submit the changes to `fc` in a PR to that repo so that it can be integrated into the master branch there and then pulled into this submodule.
> 
> In addition, please place the new test cases into their own test module, perhaps `test_io_raw.cpp` and follow the pattern(s) established in the other test modules so that it is logically easier to find test cases related to certain systems.

Sure. Would you check https://github.com/EOSIO/fc/pull/99?

brianjohnson5972:
@taokayan @b1bart This is so old, I'm presuming that we should start again, rather than keeping this hanging around.

Issue #7407, opened 2019-05-24
cleos sign should allow signing with keosd


arhag:
The `cleos sign` command currently provides two ways to sign a transaction:
1. Pass in the private key directly as a command-line argument with the `-k` option.
2. Avoid the `-k` option in which case cleos will prompt the user the enter in the private key (it won't be visible in the terminal).

It should be possible to pass in the public key (corresponding to the private key to sign with) as a command line option which would trigger cleos to request the corresponding private key from keosd.

andriantolie:
PR #7430 is created for this

heifner:
Resolved by #7430

PR #7430, opened 2019-05-29
Add option for cleos sign subcommand to ask keosd for signing

andriantolie:
## Change Description

Related to https://github.com/EOSIO/eos/issues/7407
Currently "cleos sign" subcommand doesn't allow the user to sign using the private key which is stored in keosd. This PR will enable this capability for cleos.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -3300,20 +3300,21 @@ int main( int argc, char** argv ) {
    string str_private_key;
    string str_chain_id;
    string str_private_key_file;
-   string str_keosd_pub_key;
+   string str_public_key;
    bool push_trx = false;
 
    auto sign = app.add_subcommand("sign", localized("Sign a transaction"), false);
    sign->add_option("transaction", trx_json_to_sign,
                                  localized("The JSON string or filename defining the transaction to sign"), true)->required();
    sign->add_option("-k,--private-key", str_private_key, localized("The private key that will be used to sign the transaction"));
-   sign->add_option("-w,--wallet", str_keosd_pub_key, localized("Ask keosd to sign with the given public key"));
+   sign->add_option("--public-key", str_public_key, localized(string("Ask ") + string(key_store_executable_name +
+                                                                 "to sign with the corresponding private key of the given public key"));
    sign->add_option("-c,--chain-id", str_chain_id, localized("The chain id that will be used to sign the transaction"));
-   sign->add_flag( "-p,--push-transaction", push_trx, localized("Push transaction after signing"));
+   sign->add_flag("-p,--push-transaction", push_trx, localized("Push transaction after signing"));
 
    sign->set_callback([&] {
 
-      EOSC_ASSERT( str_private_key.empty() || str_keosd_pub_key.empty(), "ERROR: Either -k/--private-key or -w/--wallet or none of them can be set" );
+      EOSC_ASSERT( str_private_key.empty() || str_public_key.empty(), "ERROR: Either -k/--private-key or --public-key or none of them can be set" );
 
       signed_transaction trx = json_from_file_or_string(trx_json_to_sign).as<signed_transaction>();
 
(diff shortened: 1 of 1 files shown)


programs/cleos/main.cpp
```diff
@@ -3299,16 +3299,22 @@ int main( int argc, char** argv ) {
...
    string str_chain_id;
+   string str_private_key_file;
+   string str_keosd_pub_key;
    bool push_trx = false;
 
    auto sign = app.add_subcommand("sign", localized("Sign a transaction"), false);
    sign->add_option("transaction", trx_json_to_sign,
                                  localized("The JSON string or filename defining the transaction to sign"), true)->required();
    sign->add_option("-k,--private-key", str_private_key, localized("The private key that will be used to sign the transaction"));
+   sign->add_option("-w,--wallet", str_keosd_pub_key, localized("Ask keosd to sign with the given public key"));
```
heifner: I think `--public-key` would be a better option name.
Use `key_store_executable_name` instead of `keosd`
The help text should say, ".. sign with the corresponding private key of the given public key" or something along those lines. It doesn't sign with the public key.

programs/cleos/main.cpp
```diff
@@ -3299,16 +3299,22 @@ int main( int argc, char** argv ) {
...
    sign->add_option("-c,--chain-id", str_chain_id, localized("The chain id that will be used to sign the transaction"));
    sign->add_flag( "-p,--push-transaction", push_trx, localized("Push transaction after signing"));
 
    sign->set_callback([&] {
+
+      EOSC_ASSERT( str_private_key.empty() || str_keosd_pub_key.empty(), "ERROR: Either -k/--private-key or -w/--wallet or none of them can be set" );
```
heifner: If option name changed above, make sure to change here also.

programs/cleos/main.cpp
```diff
@@ -3321,16 +3327,26 @@ int main( int argc, char** argv ) {
          chain_id = chain_id_type(str_chain_id);
       }
 
-      if( str_private_key.size() == 0 ) {
+      if( str_private_key.size() > 0 ) {
+         private_key_type priv_key;
+         try {
+            priv_key = private_key_type(str_private_key);
+         } EOS_RETHROW_EXCEPTIONS(private_key_type_exception, "Invalid private key: ${private_key}", ("private_key", str_private_key))
```
heifner: Don't echo the private key.

programs/cleos/main.cpp
```diff
@@ -3321,16 +3327,26 @@ int main( int argc, char** argv ) {
...
+            pub_key = public_key_type(str_keosd_pub_key);
+         } EOS_RETHROW_EXCEPTIONS(public_key_type_exception, "Invalid public key: ${public_key}", ("public_key", str_keosd_pub_key))
+         fc::variant keys_var(flat_set<public_key_type>{ pub_key });
+         sign_transaction(trx, keys_var, *chain_id);
+      } else {
          std::cerr << localized("private key: ");
          fc::set_console_echo(false);
          std::getline( std::cin, str_private_key, '\n' );
          fc::set_console_echo(true);
       }
```
heifner: Looks like you can no longer sign with a private key provided to `getline`

programs/cleos/main.cpp
heifner: Do we need option to sign via permission(s)? account@permission ?

programs/cleos/main.cpp
andriantolie: Right, my bad. I forgot to use the str_private_key after refactoring the code. Thanks a lot for catching this!

programs/cleos/main.cpp
andriantolie: PR updated. No longer echoing invalid private key

programs/cleos/main.cpp
andriantolie: Make sense. PR updated with a better option name and description.

andriantolie: I think signing via permission is more like a convenient option to add. But if we add it, I guess we will need to bring the auth_checker logic to cleos? Say, for example, alice@active with threshold 2 is satisfied by a key with weight 1 and bob@active with weight 1, and bob@weight is satisfied with a key. In this situation, if we decide to use `cleos sign` with `alice@active`, cleos will need to recursively decode the permission to end up with alice's key and bob's key and then sign with these two keys. These can already be achieved by passing public key of alice and bob to `cleos sign`'s `--public-key` option. Personally, I prefer to keep the logic in cleos simple and not bloated the software, so I'm not sure whether it's worth it or not to add this option.

PR #7441, opened 2019-05-30
Net plugin block id notification

heifner:
## Change Description

- Send `notice_message` with block id as soon as block is received. This allows the receiver to avoid sending block to a node that has already received it.
- This should reduce block network traffic.
- Change requires no protocol change as `notice_message` already allowed for block id notify.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
Replaced by #7598

Issue #7443, opened 2019-05-31
Parallel long_running test cases


taokayan:
make the current test framework parallelizable for long running test cases:
```
    Start 63: nodeos_sanity_lr_test
1/9 Test #63: nodeos_sanity_lr_test .................   Passed    5.69 sec
    Start 64: nodeos_run_check_lr_test
2/9 Test #64: nodeos_run_check_lr_test ..............   Passed   79.59 sec
    Start 65: nodeos_remote_lr_test
3/9 Test #65: nodeos_remote_lr_test .................   Passed   76.96 sec
    Start 66: nodeos_forked_chain_lr_test
4/9 Test #66: nodeos_forked_chain_lr_test ...........   Passed  717.55 sec
    Start 67: nodeos_voting_lr_test
5/9 Test #67: nodeos_voting_lr_test .................   Passed  1148.16 sec
    Start 68: nodeos_under_min_avail_ram_lr_test
6/9 Test #68: nodeos_under_min_avail_ram_lr_test ....   Passed  316.36 sec
    Start 69: nodeos_irreversible_mode_lr_test
7/9 Test #69: nodeos_irreversible_mode_lr_test ......   Passed  575.07 sec
    Start 70: nodeos_chainbase_allocation_lr_test
8/9 Test #70: nodeos_chainbase_allocation_lr_test ...   Passed   44.73 sec
    Start 71: nodeos_startup_catchup_lr_test
9/9 Test #71: nodeos_startup_catchup_lr_test ........   Passed  366.80 sec
```

Issue #78, opened 2017-07-13
Create special account names (nobody, anybody, producers)


elmato:

elmato:
Update "producers" account active authority at the end of each round

Issue #76, opened 2017-07-12
char_to_symbol inconsistency


jcalfee:
Is it 26?
```c++
return (c - '1') + 26;
```
https://github.com/EOSIO/eos/blob/c5dba14c09bb1b039165ded55e1481b1a9c8e0dc/contracts/eoslib/eos.hpp#L168

or 27?
```c++
return (c - '1') + 27;
```
https://github.com/EOSIO/eos/blob/c5dba14c09bb1b039165ded55e1481b1a9c8e0dc/libraries/types/include/eos/types/native.hpp#L88

bytemaster:
It should be 27.

bytemaster:
Fixed

Issue #74, opened 2017-07-11
Cannot use from US?


gfxguru:
I'm quite interested in the project,code and wondering if I can even use if I'm located in the US? I'm unable to participate from eos site. Is this true that I can't connect or utilize if in the US?... or does this even apply?

I was even interested in investing but not if we're unable to use.

bytemaster:
The EOS.IO software is available to everyone in all jurisdictions.

PR #7444, opened 2019-05-31
test parallel long_running tests

taokayan:
## Change Description
implement #7443. allow some of the long running tests to be executed in parallel.

To configure it, each python test need to be specified with a different "--cluster-id", different "--wallet_port", and "--clean-run" must NOT be specified. Test cases involving mongodb can not be parallelizable. 

users can run those parallelizable long running test cases by the following command:
```ctest -L parallel_tests_group_1 --parallel 30```

This is a sample output. Around 2200 seconds are saved thanks to parallelism:
```
Test project /Users/jiaendu/workspaces/eos
    Start 78: nodeos_under_min_avail_ram_lr_test_parallel
    Start 77: nodeos_voting_lr_test_parallel
    Start 75: nodeos_irreversible_mode_lr_test_parallel
    Start 74: nodeos_forked_chain_lr_test_parallel
    Start 76: nodeos_startup_catchup_lr_test_parallel
    Start 73: nodeos_run_check_lr_test_parallel
    Start 72: nodeos_sanity_lr_test_parallel
1/7 Test #72: nodeos_sanity_lr_test_parallel ................   Passed    5.76 sec
2/7 Test #73: nodeos_run_check_lr_test_parallel .............   Passed  115.45 sec
3/7 Test #76: nodeos_startup_catchup_lr_test_parallel .......   Passed  319.92 sec
4/7 Test #78: nodeos_under_min_avail_ram_lr_test_parallel ...   Passed  375.10 sec
5/7 Test #75: nodeos_irreversible_mode_lr_test_parallel .....   Passed  596.41 sec
6/7 Test #74: nodeos_forked_chain_lr_test_parallel ..........   Passed  788.64 sec
7/7 Test #77: nodeos_voting_lr_test_parallel ................   Passed  1208.61 sec

100% tests passed, 0 tests failed out of 7

Label Time Summary:
parallel_tests_group_1    = 3409.89 sec*proc (7 tests)

Total Test time (real) = 1208.67 sec
```

to run it with verbose log:
```ctest -L parallel_tests_group_1 --parallel 30 -V```

## Consensus Changes
N/A

## API Changes
N/A

## Documentation Additions
N/A

### CODE REVIEW

**Comment by NorseGaud** [comment]
Is there something special we need to change in the ctest commands for the cicd NP tests?

Also, do you have read access to buildkite to view the failure reasons/logs?

**Comment by NorseGaud** [comment]
Can you add the following to your pipeline.jsonc file right at the top?
```
    "eosio": {
        "pipeline-branch": "kayan_parallel_tests",
        "environment": {
            "DEBUG": true
        }
    },
```

Once you do, I will setup the test scripts to run with the new ctest command to get them passing.

File: programs/eosio-launcher/main.cpp [hunk 1]
```diff
@@ -150,19 +153,19 @@ class host_def {
   vector<eosd_def> instances;
 
   uint16_t p2p_port() {
-    return base_p2p_port + p2p_count++;
+    return cluster_id * cluster_port_offset + base_p2p_port + p2p_count++;
```
arhag: I think the way ports are calculated should be updated to reflect the limits imposed by `cluster_port_offset` and the fact that the port number must be a 16-bit unsigned integer.

I think ports could be calculated with the help of the following code:

```cpp
enum class port_type : uint8_t {
   p2p,
   http,
   NUM_PORT_TYPES
};

namespace port_constants {
   static constexpr uint16_t base = 1536;
   static constexpr uint16_t cluster_stride = 2000;
   static constexpr uint8_t  node_stride = 5;
   // maximum number of different types of ports = node_stride

   static_assert( cluster_stride > 0 );
   static_assert( base + cluster_stride <= std::numeric_limits<uint16_t>::max() );

   static constexpr uint16_t max_num_clusters = (std::numeric_limits<uint16_t>::max() - base) / cluster_stride;
   // maximum number of clusters = max_num_clusters

   static_assert( node_stride > 0 );
   static_assert( node_stride <= cluster_stride );
   static_assert( node_stride >= static_cast<uint8_t>(port_type::NUM_PORT_TYPES) );

   static constexpr uint16_t max_num_nodes_per_cluster = cluster_stride / node_stride;
   // maximum number of nodes (including bios node) per cluster = max_num_nodes_per_cluster
}

constexpr uint16_t calculate_port( uint16_t cluster_index, uint16_t node_index, port_type t ) {
   if( cluster_index >= port_constants::max_num_clusters )
     throw std::out_of_range( "invalid cluster index" );
   if( node_index >= port_constants::max_num_nodes_per_cluster )
     throw std::out_of_range( "invalid node index" );
    
   return port_constants::base 
            + (port_constants::cluster_stride * cluster_index)
            + (port_constants::node_stride * node_index)
            + static_cast<uint16_t>(t);
}
```

So for example the `http_bios_port()` would return `calculate_port( cluster_id, 0, port_type::http )` which given the example port constants defined above would be equal to 1537 if `cluster_id == 0` and 3537 if `cluster_id == 1`.

The numbers defined above for `port_constants::base`, `port_constants::cluster_stride`, and `port_constants::node_stride` are just examples. But with those example values, this system could support a maximum of 32 (`max_num_clusters`) simultaneous clusters on a single machine and a maximum of 400 (`max_num_nodes_per_cluster`) nodes per each cluster (or a maximum of 399 nodes if you don't count the bios node). And it could do this while also supporting an expansion of the number of port types from the current two (`p2p` and `http`) to 5 to accommodate other ports we may want in future tests, e.g. the state history plugin. (Also what about the keosd port? Should a separate keosd run for each cluster, or should multiple clusters running on a single machine share a single keosd instance?)

programs/eosio-launcher/main.cpp
```diff
@@ -150,19 +153,19 @@ class host_def {
...
   }
 
   uint16_t http_port() {
-    return base_http_port + http_count++;
+    return cluster_id * cluster_port_offset + base_http_port + http_count++;
```
arhag: So for example, using the above mentioned scheme, this implementation could be:
```cpp
   return calculate_port( cluster_id, ++http_count, port_type::http );
```

programs/eosio-launcher/main.cpp
```diff
@@ -555,6 +560,10 @@ launcher_def::initialize (const variables_map &vmap) {
     }
   }
 
+  if (vmap.count("cluster-id")) {
+    cluster_id = vmap["cluster-id"].as<uint32_t>();
```
arhag: `cluster_id` could be a `uint16_t` and you could assert that it is less than `max_num_clusters` here.

programs/eosio-launcher/main.cpp
```diff
@@ -605,8 +615,17 @@ launcher_def::initialize (const variables_map &vmap) {
...
   }
 
-  config_dir_base = "etc/eosio";
-  data_dir_base = "var/lib";
+  config_dir_base_cluster0 = "etc/eosio";
+  if (cluster_id == 0) {
+    config_dir_base = config_dir_base_cluster0;
+    data_dir_base = "var/lib";
+  } else {
+    std::stringstream config_s, data_s;
```
arhag: It is faster to just convert the number to a string and then append it to the existing string than use `std::stringstream` here.

Also when converting the number to a string it would be good to always have 5 digits with padded zeros as necessary. That way lexicographical sorting of the directories would always work correctly.

programs/eosio-launcher/main.cpp
```diff
@@ -605,8 +615,17 @@ launcher_def::initialize (const variables_map &vmap) {
     }
   }
 
-  config_dir_base = "etc/eosio";
-  data_dir_base = "var/lib";
+  config_dir_base_cluster0 = "etc/eosio";
+  if (cluster_id == 0) {
+    config_dir_base = config_dir_base_cluster0;
```
arhag: Why not be consistent and do the same thing as when `cluster_id != 0`? 

So in this case `config_dir_base` could be the path "etc/eosio/cluster00000" and `data_dir_base` could be the path "var/lib/cluster00000".

programs/eosio-launcher/main.cpp
```diff
@@ -1635,6 +1654,11 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
   case LM_LOCAL:
   case LM_REMOTE : {
     bfs::path source = "last_run.json";
+    if (cluster_id) {
+      std::stringstream s;
```
arhag: Similar change as before: use string concatenation instead and enforce 5 digits, e.g. `last_run_cluster00001.json`.

programs/eosio-launcher/main.cpp
```diff
@@ -1866,6 +1890,11 @@ launcher_def::start_all (string &gts, launch_modes mode) {
   }
   }
   bfs::path savefile = "last_run.json";
+  if (cluster_id) {
+    std::stringstream s;
```
arhag: Similarly here.

tests/Cluster.py
```diff
@@ -65,6 +66,12 @@ def __init__(self, walletd=False, localCluster=True, host="localhost", port=8888
         defproducerbPrvtKey: Defproducerb account private key
         """
         self.accounts={}
+        self.clusterID=clusterID
+        if clusterID != 0:
```
arhag: This and the related lines below would have to also be changed to reflect the new port assignment scheme I proposed above, as well as to not treat a `clusterID` of 0 as a special case and also to ensure 5 digits are always included in the name of the bios boot shell script.

tests/WalletMgr.py
```diff
@@ -19,14 +19,22 @@ class WalletMgr(object):
...
         self.port=port
         self.host=host
         self.wallets={}
         self.__walletPid=None
+        if clusterID != 0:
+            WalletMgr.__walletLogOutFile = "test_keosd_cluster" + str(clusterID) + "_out.log"
+            WalletMgr.__walletLogErrFile="test_keosd_cluster" + str(clusterID) + "_err.log"
+            WalletMgr.__walletDataDir="test_wallet_" + str(clusterID) 
+            if self.port == 9899:
+                self.port += clusterID
```
arhag: If separate keosd instances are going to run per cluster, then it would need to be allocated in the `port_type` `enum class` described above to use that port assignment scheme. Also, keosd could also use Unix sockets instead which would save a port type.

taokayan:
> Can you add the following to your pipeline.jsonc file right at the top?
> 
> ```
>     "eosio": {
>         "pipeline-branch": "kayan_parallel_tests",
>         "environment": {
>             "DEBUG": true
>         }
>     },
> ```
> 
> Once you do, I will setup the test scripts to run with the new ctest command to get them passing.

hi, I guess I need to add ```--output-on-failure -T Test``` to the ctest command for parallel test cases.

NorseGaud:
Do we need to run the previous `ctest -L nonparallelizable_tests --output-on-failure -T Test` and also the new `ctest -L parallel_tests_group_1 --parallel 4 --output-on-failure -T Test`?

Or, can you add the new tests to the ones that run when we execute `ctest -LE _tests --output-on-failure -T Test`?

NorseGaud:
Sorry about that -- I've fixed the pipeline problem.

NorseGaud:
Once last fix to make. Need you to change your piperline.jsonc to have the following (replacing the current eosio block):

```
    "eosio": {
        "environment": {
            "HELPER_ENABLED": true,
            "CONTAINER_LEVEL_PARALLELIZATION": true,
            "IMAGE_TAG": "1.9.0-develop"
        }
    }
```

taokayan:
> Once last fix to make. Need you to change your piperline.jsonc to have the following (replacing the current eosio block):
> 
> ```
>     "eosio": {
>         "environment": {
>             "HELPER_ENABLED": true,
>             "CONTAINER_LEVEL_PARALLELIZATION": false,
>             "IMAGE_TAG": "1.9.0-develop"
>         }
>     }
> ```

sure. I'll fix it and try.

NorseGaud:
Hey Kayan, things are stable again. However, I still need my question answered about where these tests are going to run?

Can we add these to the parallel tests that currently exist and execute when we run `ctest -LE _tests --output-on-failure -T Test`? You can change that in the cmake stuff I believe and then we can avoid any pipeline changes.

kj4ezj:
> Hey Kayan, things are stable again. However, I still need my question answered about where these tests are going to run.
> 
> Can we add these to the parallel tests that currently exist and execute when we run `ctest -LE _tests --output-on-failure -T Test`? You can change that in the cmake stuff I believe and then we can avoid any pipeline changes.

I agree that adding them to the unit test group is the best approach because the unit test group is already run in parallel. It sucks that the tests will take almost 20 minutes again, but we want to add the long-running tests back into the main pipeline anyways now that we have container-level parallelization so that will eventually happen either way.

taokayan:
> Hey Kayan, things are stable again. However, I still need my question answered about where these tests are going to run.
> 
> Can we add these to the parallel tests that currently exist and execute when we run `ctest -LE _tests --output-on-failure -T Test`? You can change that in the cmake stuff I believe and then we can avoid any pipeline changes.

I think at this moment we don't support running nonparallelizable tests along with parallelizable tests, although sometime we can do it successfully. This is because most of the nonparallelizable tests implicitly kill all "nodeos" processes or "keosd" processes regardless of cluster id before they exit. However, in the long term we can migrate each testing case from nonparallelizable to paralleblizable.

kj4ezj:
> Can I know if `ctest -LE _tests --output-on-failure -T Test` only executes unit_tests but not those tests from scripts?

Currently it runs unit tests, and a couple of other tests which are in the integration test folder but which are parallelizable, including `plugin_test`, `version-label-test`, `test_cypher_suites`, etc. You can see which tests that command will run without actually running them by using the `-N` flag:  
`$ ctest -N -LE _tests --output-on-failure -T Test`

> If yes then we can run it along with parallel tests, although one of the test in parallel test group can take as much as 1208.61 sec which might not be a good idea to be part of the building process.

Yeah, it is unfortunate that the `nodeos_voting_lr_test` takes ~20 minutes, but I see that as a separate issue. I actually [started fixing it](https://github.com/EOSIO/eos/pull/6987), but that has gotten deprioritized in favor of other work for now. Ultimately our medium-term goal is to run all tests against every pull request. If this test increases our build time by 20 minutes, then fixing that test will become higher priority again.

programs/eosio-launcher/main.cpp
```diff
@@ -1271,6 +1304,12 @@ launcher_def::write_bios_boot () {
          string key = line.substr(len);
          if (key == "envars") {
             brb << "bioshost=" << bhost << "\nbiosport=" << biosport << "\n";
+            char cluster_str[20];
+            sprintf(cluster_str, "cluster%05d", cluster_id);
+            brb << "wddir=$wddir/" << (const char *)cluster_str << "\n";
+            brb << "wdaddr=localhost:" << calculate_port(cluster_id, 0, port_type::keosd_boot) << "\n";
+            brb << "wdurl=http://$wdaddr\n";
+            brb << "wpidfile=ignition_wallet" << cluster_str << ".pid\n";
```
arhag: This could say, for example, "wpidfile=ignition_walletcluster00000". Did you mean to add a _ to separate them?

Better yet I think this line should be:
```cpp
brb << "wpidfile=$wddir/ignition_wallet.pid\n";
```

That way all these files are kept together in the appropriate cluster subdirectory.

In fact, at that point, both this line and the one above it can just be kept in the testnet.template file after the `###INSERT envars` line.

programs/eosio-launcher/main.cpp
```diff
@@ -1271,6 +1304,12 @@ launcher_def::write_bios_boot () {
          string key = line.substr(len);
          if (key == "envars") {
             brb << "bioshost=" << bhost << "\nbiosport=" << biosport << "\n";
+            char cluster_str[20];
+            sprintf(cluster_str, "cluster%05d", cluster_id);
+            brb << "wddir=$wddir/" << (const char *)cluster_str << "\n";
+            brb << "wdaddr=localhost:" << calculate_port(cluster_id, 0, port_type::keosd_boot) << "\n";
```
arhag: Why is a `keosd_boot` port type necessary that is separate from the other `keosd` port type?

testnet.template
```diff
@@ -44,8 +42,12 @@ echo "FEATURE_DIGESTS: $FEATURE_DIGESTS" >> $logfile
 
 echo "http-server-address = $wdaddr" > $wddir/config.ini
 
+echo "starting keosd" >> $logfile
+
 programs/keosd/keosd --config-dir $wddir --data-dir $wddir 2> $wddir/wdlog.txt &
-echo $$ > ignition_wallet.pid
+wpid=$!
+echo "pid is $wpid, saving to $pwidfile" >> $logfile
```
arhag: Should be "pid is $wpid, saving to $wpidfile".

testnet.template
```diff
@@ -133,4 +142,9 @@ programs/cleos/cleos --wallet-url $wdurl --url http://$bioshost:$biosport push a
 # ------ DO NOT ALTER THE NEXT LINE -------
 ###INSERT cacmd
 
-pkill -15 keosd
+killcmd="kill -15 \`cat $wpidfile\`"
+echo "ready to kill keosd with cmd:$killcmd" >> $logfile
```
arhag: Add space after the colon.

tests/Cluster.py
```diff
@@ -46,11 +46,18 @@ class Cluster(object):
     __BiosPort=8788
     __LauncherCmdArr=[]
     __bootlog="eosio-ignition-wd/bootlog.txt"
```
arhag: Wouldn't this now be "eosio-ignition-wd/cluster00000/bootlog.txt" for cluster 0? Also is this even needed? Shouldn't it be replaced during initialization?

tests/Cluster.py
```diff
@@ -46,11 +46,18 @@ class Cluster(object):
...
-    __BootScript="bios_boot.sh"
+    __BootScript="bios_boot_cluster0000.sh"
```
arhag: Missing a 0. Also is this even necessary anymore?

tests/Cluster.py
```diff
@@ -46,11 +46,18 @@ class Cluster(object):
     __BiosPort=8788
```
arhag: Old port number still here.

tests/Cluster.py
```diff
@@ -46,11 +46,18 @@ class Cluster(object):
...
+    __portBase = Utils.portBase
+    __cluster_stride = Utils.cluster_stride
+    __node_stride = Utils.node_stride
+    __port_type_p2p = Utils.port_type_p2p
+    __port_type_http = Utils.port_type_http
+    __port_type_keosd = Utils.port_type_keosd
 
     # pylint: disable=too-many-arguments
     # walletd [True|False] Is keosd running. If not load the wallet plugin
-    def __init__(self, walletd=False, localCluster=True, host="localhost", port=8888, walletHost="localhost", walletPort=9899, enableMongo=False
```
arhag: Why is there still a default `walletPort` using the old numbering scheme?

tests/Cluster.py
```diff
@@ -67,10 +74,13 @@ def __init__(self, walletd=False, localCluster=True, host="localhost", port=8888
         """
         self.accounts={}
         self.clusterID=clusterID
-        if clusterID != 0:
-            port += clusterID * 2000
-            Cluster.__BiosPort = 8788 + clusterID * 2000
-            Cluster.__BootScript = "bios_boot_cluster" + str(clusterID) + ".sh"
+        Utils.clusterID=clusterID
```
arhag: Why is this stopping over `Utils.clusterID`?

If `Utils.getNodeDataDir` and `Utils.getNodeConfigDir` need the cluster ID, then it should be passed in as an argument to the function.

tests/Cluster.py
```diff
@@ -67,10 +74,13 @@ def __init__(self, walletd=False, localCluster=True, host="localhost", port=8888
...
+
+        Cluster.__BiosPort = Cluster.__portBase + clusterID * Cluster.__cluster_stride + Cluster.__port_type_http
+        Cluster.__BootScript = "bios_boot_cluster" + ("%05d" % clusterID) + ".sh"
```
arhag: I would prefer to see the generated boot script go in the same cluster sub-directory as all the other cluster specific files.

tests/WalletMgr.py
```diff
@@ -12,29 +12,29 @@
...
-        if clusterID != 0:
-            WalletMgr.__walletLogOutFile = "test_keosd_cluster" + str(clusterID) + "_out.log"
-            WalletMgr.__walletLogErrFile="test_keosd_cluster" + str(clusterID) + "_err.log"
-            WalletMgr.__walletDataDir="test_wallet_" + str(clusterID) 
-            if self.port == 9899:
-                self.port += clusterID     
-            if self.nodeosPort == 8888:
-                self.nodeosPort += clusterID * 2000
+
+        WalletMgr.__walletLogOutFile = "test_keosd_cluster" + ("%05d" % clusterID) + "_out.log"
```
arhag: I would again prefer to see all cluster specific files to be contained in the same directory rather than augmenting the files names with the cluster IDs.

programs/eosio-launcher/main.cpp
arhag: Latest changes switch to using `uint16_t`. But I would still like to see an assertion that `cluster_id <= port_constants::max_num_clusters` to provide a helpful error message early on if the user provides a cluster ID that is too large.

programs/eosio-launcher/main.cpp
```diff
@@ -798,7 +820,7 @@ launcher_def::generate () {
       sf.close();
     }
     if (host_map_file.empty()) {
-      savefile = bfs::path (output.stem().string() + "_hosts.json");
+      savefile = bfs::path (bfs::path(cluster_id_str()) / (output.stem().string() + "_hosts.json"));
     }
     else {
       savefile = bfs::path (host_map_file);
```
arhag: Here you could bring in `cluster_id_str()` since `host_map_file` wouldn't be modified earlier. But more importantly, why was this being done in the original code in the first place. It seems like a bad idea for the input to also maybe be an output file.

tests/WalletMgr.py
```diff
@@ -12,9 +12,9 @@
 Wallet=namedtuple("Wallet", "name password host port")
 # pylint: disable=too-many-instance-attributes
 class WalletMgr(object):
```
arhag: We shouldn't be using `findAvailablePort` since the port numbers that keosd is allowed to use are strongly constrained now. We should be using a range of ports that are hopefully not taken on the machines these tests run on. If so, as long as two tests don't run with the same cluster ID, the hope is that the assigned port number could be used without conflict. If there is a conflict, then the test must shutdown. So this will require changes to the `launch` function to not use `findAvailablePort` and furthermore, the port should always be specified to the `WalletMgr`.

programs/eosio-launcher/main.cpp
taokayan: There's are 2 keosd processes for a cluster: the ignition wallet which is inside the bios_boot.sh, and the normal wallet which is managed in Cluster.py. I am thinking to make the ignition wallet to use port of ```calculate_port(cluster_id, 0, port_type::keosd)```, and the normal wallet use ```calculate_port(cluster_id, 1, port_type::keosd)```

tests/Cluster.py
taokayan: I'll only put clusterID into Utils only, which makes thing much easier and can be used in Cluster, WalletMgr, and Node.

programs/eosio-launcher/main.cpp
```diff
@@ -466,6 +466,27 @@ struct launcher_def {
    fc::optional<uint32_t> max_transaction_cpu_usage;
    eosio::chain::genesis_state genesis_from_file;
 
+   std::string cluster_id_str() const {
+      char cluster_id_str[20];
+	   sprintf(cluster_id_str, "cluster%05u", cluster_id);
+      std::string str = cluster_id_str;
+      if (!bfs::exists(str)) {
```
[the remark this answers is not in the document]
taokayan: my idea is to create ```bfs::current_path() / cluster00000``` and the following files will be there:
```
shape_bridge_hosts.json
shape_bridge.json
setprods.json
bios_boot.sh
testnet.dot
last_run.json
test_keosd_out.log
test_keosd_err.log
```
There files are originally in bfs::current_path(), but not etc/eosio/. I think it's better not to put them into ```etc/eosio/cluster00000``` because the contexts of ```etc/eosio/cluster00000``` will be cleaned up after the tests and it makes user difficult to check those files.

programs/eosio-launcher/main.cpp
```diff
@@ -1692,9 +1712,7 @@ launcher_def::kill (launch_modes mode, string sig_opt) {
   case LM_ALL:
   case LM_LOCAL:
   case LM_REMOTE : {
-    char last_run_str[100];
-	  sprintf(last_run_str, "last_run_cluster%05u.json", cluster_id);
-    std::string source = (const char *)last_run_str;
+    std::string source = last_run_json_str();
```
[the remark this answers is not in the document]
taokayan: will fix it

programs/eosio-launcher/main.cpp
```diff
@@ -650,14 +671,15 @@ launcher_def::initialize (const variables_map &vmap) {
     }
   }
 
-	char cluster_id_str[10];
-	sprintf(cluster_id_str, "%05u", cluster_id);
+  if (!host_map_file.empty() && host_map_file.string().find(cluster_id_str()) == std::string::npos) {
```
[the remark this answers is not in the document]
taokayan: will fix it

programs/eosio-launcher/main.cpp
taokayan: changed it to ```savefile = cluster_dir_path() / (output.stem().string() + "_hosts.json"));```

tests/WalletMgr.py
taokayan: will remove findAvailablePort

taokayan: it can be specified, or by default the port is ```calculate_port(cluster_id, 1, port_type::keosd)```

brianjohnson5972:
This effort appears to have been an earlier attempt than the launcher-service to have parallel integration tests.  It is so old, I'm presuming it isn't worth trying to bring up to date. @taokayan

taokayan:
Yeah I agree it's not worth to bring it up.

PR #7458, opened 2019-06-03
[1.8.x] BATS bash tests for build scripts + various other improvements and fixes

NorseGaud:
## Change Description
- BATS-CORE is being used for bash unit testing: This will allow us to ensure bash scripts function as they should for users. It's perfect for regression testing and consistency. (see tests/bash-bats/README.md)
- CICD Pipeline/steps run as the user cicd (instead of root)
  - Improvements to log output in buildkite showing a collapsed tab for each and every command the scripts run.
  - Auto retry failed step on -1 (Agent Lost)
- Users are prompted if 'which' is missing (we will install it for them if they want): https://github.com/EOSIO/eos/issues/7232
- All prompts are now properly wrapped to be user friendly: https://github.com/EOSIO/eos/issues/6931
- New eosio path: `~/eosio/$VersionFromCmakeFile/` to store dependencies and eosio: https://github.com/EOSIO/eos/issues/6931
- Script filename extensions are now .bash, indicating that bash is required to properly execute.
- Scripts now use #!/usr/bin/env to pull bash if it's installed into nonstandard /bin/ location
- Better uninstaller script, removing old "full_uninstaller" (readme updated)
- Build scripts support DRYRUN=true (doesn't execute anything, but shows you what it would do) and VERBOSE=true (shows the exact commands executed throughout script) for better troubleshooting.
- Scripts use `set -ieo pipefail` so they fail properly when a command fails and also show colors for users executing the script (colors not supported on mac yet).
- Using *_deps file for each distro
  - Package dependency files include flexibility by allowing us to specify what command is used to check if it exists or not (dpkg -s/rpm -qa)
  - no more extra space at end of deps file! (yay)
- Using helpers/functions for modularity, avoiding having to repeat ourselves in each eosio_build_* script
- Environment variables have been moved into a separate file for better flexibility
  - Test scripts can pull in the same variables
  - comments added for better comprehension 

### Compiler support
_You can see almost all of these tests passing in https://buildkite.com/EOSIO/eosio-base-images/builds/170

| OS | -P | existing compiler | existing c++/clang < required version | no existing compiler |
| --- | --- | --- | --- | --- |
| Darwin | :white_check_mark: | :white_check_mark: | :white_check_mark: Will prompt for pin/install clang8 | :heavy_check_mark: (Hasn't been tested) |
[...]

PR #7479, opened 2019-06-10
Quick fix for 1.8.x IMAGE_TAG

NorseGaud:


NorseGaud:
> It needs to be `IMAGE_TAG='_2-5'` as the underscore is no longer included in the hard-coded part of the image tag for backwards-compatibility with `release/1.8.x`.

I moved everything >= 1.8 to using IMAGE_TAG without _ in it. That's part of the .yml files now.

NorseGaud:
merged in https://github.com/EOSIO/eos/pull/7458

Issue #7402, opened 2019-05-24
1.8+ performance regression on 18.04 when using build script's default options


spoonincode:
Pre-1.8 the build script on 18.04 used clang 4.0 as the compiler. 1.8+ uses gcc 7. gcc seems to really struggle with wabt and this causes a serious performance regression for users who use the default build script options.

I measured a popular benchmarking contract on a 9600k with a core forced to C0 (4.6Ghz). The execution time was:

compiler | execution time
------------ | -------------
clang 8 | 1.2ms
gcc 8 | 1.7ms
gcc 7 | 1.8ms
clang 6 | 1.2ms

We need to seriously consider using the system's clang compiler instead of gcc. For ubuntu 18.04 this means clang6, for amz2 this means clang7, and for rhel7 clones this may possibly mean clang5 (there are new versions of clang available from RH but the centos software collections have not surpassed 5.0).

Obviously users who use the pinned build script option are not affected since they get clang8.

NorseGaud:
I've added a nice little warning/prompt for this in bats changes.

PR #7480, opened 2019-06-10
on supported platforms build with system clang by default

spoonincode:
## Change Description

compared to clang, building with gcc has shown significant performance degradation particularly for wabt execution. See #7402. Change the build scripts to use a system provided clang for platforms that provide it. For supported platforms we will now end up building with:

platform | compiler
------------ | -------------
Ubuntu 16.04 | clang 8 (via forced -P option)
Ubuntu 18.04 | clang 6
Centos 7 | clang 5
Amazon Linux 2 | clang 7
macOS | clang provided by Xcode

🚨🚨🚨I don't believe the clang5 in centos7 is completely c++17 compliant. It didn't seem able to use the c++17 variant properly. So maybe this isn't a good idea?

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
replaced with #7516

PR #7485, opened 2019-06-11
[1.8.x] Ensure we're in repo root

NorseGaud:
Need a way to ensure we're in the repo root. This allows people to ./eos/scripts/eosio_build.sh or even run from the scripts directory.

### DIFF
--- a/scripts/eosio_build.sh
+++ b/scripts/eosio_build.sh
@@ -111,7 +111,7 @@ if [ $# -ne 0 ]; then
    done
 fi
 
-[[ ! -d scripts ]] && echo "You must run this script from the eos root!" && exit 1
+[[ ! -d scripts ]] && echo "You must run this script from the eos root." && exit 1
 
 # Load eosio specific helper functions
 . ./scripts/helpers/eosio.sh
--- a/scripts/eosio_build.sh
+++ b/scripts/eosio_build.sh
@@ -111,6 +111,8 @@ if [ $# -ne 0 ]; then
    done
 fi
 
+[[ ! -d scripts ]] && echo "You must run this script from the eos root!" && exit 1
+
 # Load eosio specific helper functions
 . ./scripts/helpers/eosio.sh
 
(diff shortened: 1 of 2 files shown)


scripts/eosio_build.sh
```diff
@@ -111,6 +111,8 @@ if [ $# -ne 0 ]; then
    done
 fi
 
+[[ ! -d scripts ]] && echo "You must run this script from the eos root!" && exit 1
```
kj4ezj: Failing for arbitrary reasons is frustrating to users.
Instead of failing, we can automatically switch to the root of the repository for them as follows:
`cd $(git rev-parse --show-toplevel)`

NorseGaud: There may be unintended/unknown consequences to doing this. Let's wait for someone from blockchain to review.

kj4ezj: I am on blockchain.   
I can do you one better:   
```
[[ ! -d scripts && "$(git rev-parse --is-inside-work-tree)" == 'true' ]] && cd "$(git rev-parse --show-toplevel)"
[[ ! -d scripts ]] && (echo 'ERROR: Cannot find the scripts directory!' && echo 'This script must be run from within the eos repository!' && exit 1)
```
This checks if we are inside a git repository and attempts to move to the repo root, then checks for the scripts directory. The error message is also more clear and more specific.   
   
By the way, the command `echo "You must run this script from the eos root!"` returns an error on my terminal because of BASH substitution of the exclamation point. Use single-quotes to avoid the substitution. Single quotes also almost certainly print faster because BASH doesn't try to perform substitution before printing.

arhag: Why not just replace the line with the following?

```sh
cd $( dirname "${BASH_SOURCE[0]}" )/..
```

kj4ezj: arhag's solution is better than mine because it works even if eos is not a git repo.

PR #7486, opened 2019-06-11
[develop] Ensure we're in repo root

NorseGaud:
Need a way to ensure we're in the repo root. This allows people to ./eos/scripts/eosio_build.sh or even run from the scripts directory.

### DIFF
--- a/scripts/eosio_build.sh
+++ b/scripts/eosio_build.sh
@@ -108,6 +108,8 @@ if [ $# -ne 0 ]; then
    done
 fi
 
+[[ ! -d scripts ]] && echo "You must run this script from the eos root!" && exit 1
+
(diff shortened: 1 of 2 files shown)


scripts/eosio_build.sh
```diff
@@ -108,6 +108,8 @@ if [ $# -ne 0 ]; then
    done
 fi
 
+[[ ! -d scripts ]] && echo "You must run this script from the eos root!" && exit 1
```
kj4ezj: Failing for arbitrary reasons is frustrating to users.
Instead of failing, we can automatically switch to the root of the repository for them as follows:
`[[ ! -d scripts ]] && cd $(git rev-parse --show-toplevel)`

PR #7500, opened 2019-06-14
[master] Anka support

NorseGaud:
Various changes needed for supporting the new pipeline-upload and Anka.

### DIFF
--- a/pipeline.jsonc
+++ b/pipeline.jsonc
@@ -1,4 +1,12 @@
 {
+    "eosio": {
+        "environment": {
+            "ENABLE_HIGH_SIERRA": true,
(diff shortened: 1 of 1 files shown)


kj4ezj:
Blockchain has asked us to use the `pipeline.jsonc` file only when we need to override default pipeline behavior, but these variables are intended to create default pipeline behavior and should be implemented in one of the following places:
- Pipeline upload script
- Pipeline YAML
- Pipeline terraform

They also do not approve pull requests to master unless it is really, really necessary. I have opened three pull requests to master and they have all been closed as another workaround was possible. In this case, another workaround is possible so, respectfully, I am closing this pull request.

PR #7501, opened 2019-06-14
[1.8.x] Versioned images (prep for hashing)

NorseGaud:
Adding version to end of image names to prep for hashing and support flexible development/testing.

### DIFF
--- a/pipeline.jsonc
+++ b/pipeline.jsonc
@@ -1,17 +1,4 @@
 {
-    "eosio": {
-        "environment": {
-            "HELPER_ENABLED": true,
-            "IMAGE_TAG_VERSION": "v1"
-        }
-    },
-    "eosio-build-unpinned":
-    {
-        "environment":
-        {
-            "HELPER_ENABLED": true
-        }
-    },
     "eosio-lrt":
     {
         "environment":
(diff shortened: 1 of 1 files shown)


pipeline.jsonc
```diff
@@ -1,8 +1,8 @@
 {
     "eosio": {
         "environment": {
-            "IMAGE_TAG": "1.8.0-rc2",
-            "HELPER_ENABLED": true
+            "HELPER_ENABLED": true,
```
kj4ezj: Blockchain has asked us to use the `pipeline.jsonc` file only when we need to override default pipeline behavior, but `HELPER_ENABLED='true'` is required for the pipeline to function and so should be implemented in one of the following places:
- Pipeline upload script
- Pipeline YAML
- Pipeline terraform

As such, it should be deleted here.

pipeline.jsonc
```diff
@@ -1,8 +1,8 @@
...
+            "IMAGE_TAG_VERSION": "v1"
```
kj4ezj: Same as above, this should be deleted here and implemented in `auto-buildkite-pipelines` unless `IMAGE_TAG_VERSION` is overriding some default behavior.

NorseGaud: It is...

PR #7502, opened 2019-06-14
[develop] Versioned images (prep for hashing)

NorseGaud:
Adding version to end of image names to prep for hashing and support flexible development/testing.

### DIFF
--- a/pipeline.jsonc
+++ b/pipeline.jsonc
@@ -4,7 +4,7 @@
         "environment":
         {
             "HELPER_ENABLED": true,
-            "IMAGE_TAG_VERSION": "v2"
+            "IMAGE_TAG_VERSION": "v1"
         }
     },
     "eosio-build-unpinned":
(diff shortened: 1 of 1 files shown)


pipeline.jsonc
```diff
@@ -4,7 +4,7 @@
         "environment":
         {
             "HELPER_ENABLED": true,
-            "IMAGE_TAG": "1.9.0-develop"
+            "IMAGE_TAG_VERSION": "v1"
```
kj4ezj: Blockchain has asked us to use the `pipeline.jsonc` file only when we need to override default pipeline behavior, but these variables are intended to create default pipeline behavior and should be implemented in one of the following places:
- Pipeline upload script
- Pipeline YAML
- Pipeline terraform

`IMAGE_TAG_VERSION` should be deleted here and implemented in `auto-buildkite-pipelines` unless it is overriding some default behavior.

Issue #7400, opened 2019-05-24
1.8 build scripts don't handle system clang via alternatives on 18.04


spoonincode:
if I set my 18.04 ubuntu to use clang++ for c++ via update-alternatives it isn't able to detect clang. This is at least due to an error in the readlink part of the script but possibly also the hard check against `clang++`. It appears that #7366 may fix this just by reading the changes there

The 1.8 script does
```
$ readlink `which c++`
/etc/alternatives/c++
```
and expects that to be `clang++` if clang is in use. Even adding a `-f` won't fix the logic in the scripts
```
$ readlink -f `which c++`
/usr/lib/llvm-6.0/bin/clang
```

spoonincode:
I'm going to close this one because the way the 1.8/develop scripts work now is via explicitly asking for gcc or clang -- not trying to use whatever the system "alternative" is via cc

PR #7525, opened 2019-06-21
[develop] Readlink quick fix and BATS test fixes

NorseGaud:
The readlink tool throws a return code of 1 if the path is not a symlink. It used to be a symlink /usr/bin/c++ -> /usr/bin/clang++ but since https://github.com/EOSIO/eos/pull/7516, we need to return true regardless of readlink failing and check if CXX =~ clang already.


scottarnette:
In addition, this also includes fixes and adjustments for the BATS tests to support both the new installation path prompt and changes to the ```ensure-compiler``` helper function. I am continuing the refine the BATS tests while we investigate issues with builds on clean containers.

NorseGaud:
Related https://github.com/EOSIO/eos/issues/7400

PR #7528, opened 2019-06-24
Readlink bats fix r1.8

scottarnette:
## Change Description
The readlink tool throws a return code of 1 if the path is not a symlink. It used to be a symlink /usr/bin/c++ -> /usr/bin/clang++ but since #7516, we need to return true regardless of readlink failing and check if CXX =~ clang already.

In addition, this also includes fixes and adjustments for the BATS tests to support both the new installation path prompt and changes to the ensure-compiler helper function. I am continuing to refine the BATS tests while we investigate issues with builds on clean containers to ensure we check for the correct dependencies.

**NOTE:** This is a backport of #7525 for release 1.8.x.


scottarnette:
See BATS tests passing here: https://buildkite.com/EOSIO/eosio-base-images/builds/285

Issue #6988, opened 2019-03-22
Consensus protocol upgrade to forward setcode actions to the contract on the eosio account


arhag:
## Background

Normal execution of actions involve first calling any native handler that may be registered for the action and then calling the WebAssembly code if a contract is in place on the receiver account (see `apply_context::exec_one`). The condition `a.code.size() > 0` (where `a` is the `account_object` of the `receiver`) is used in that code to check that a contract exists on the receiver and if not it will not attempt to execute WebAssembly code.

However, the existing code has additional restrictions that prevents the execution of WebAssembly code for a `eosio::setcode` action even when a contract exists. The full `if` statement to enter the part of the code that executes the contract is:
```cpp
if( a.code.size() > 0 && !(act.account == config::system_account_name && act.name == N( setcode ) && receiver == config::system_account_name) ) {
```
Note that `act` is the currently executing action and `config::system_account_name == N(eosio)`.

The original intention behind this condition, which is one we no longer find necessary and may even find undesirable to maintain, was to prevent the contract from executing if the `eosio::setcode` was setting a new contract on the `eosio` account. But due to a bug in the condition, it ended up preventing WebAssembly code from executing on any `eosio::setcode` actions, regardless of which account the `eosio::setcode` action was for.

Use cases are enabled by fixing the bug. For example, it enables operators of the network to deploy a system contract that is selective about the contracts it allows other accounts to deploy. 

There is not as strong of a use case identified for allowing an `eosio::setcode` action that tries to set a contract on the `eosio` account to also be forwarded to execute on the contract deployed on `eosio`; but we haven't identified any harm in allowing that to occur as well. If it was to be allowed, there is a choice available between two ways of doing it. One option is for the old WebAssembly code  to run in the context of the `eosio::setcode` action, essentially acting as a pre-check to the code that is to replace it. The other option is for the new WebAssembly code to run in the context of the `eosio::setcode` action. We have identified three arguments in favor of taking the latter approach:

1. The code is currently structured to support easier implementation of the latter approach. 
2. Any subsequent actions that run because of the execution of the WebAssembly code running on `eosio` (whether before or after contract replacement) via the `require_recipient` or `send_inline` intrinsics would necessarily run in an environment in which the new contract was already deployed. So going with the latter approach provides more consistency to contract developers.
[...]

Issue #6168, opened 2018-10-26
Consensus upgrade to allow contracts to read code version of any account


arhag:
## Background

A smart contract deployed on an EOSIO blockchain may depend on another contract deployed to another account. The dependency can be in the form of read-only access to the tables of the other contract directly or in the form of calling certain actions of the other contract and expecting them to behave in a certain way. However, smart contracts will typically be upgraded over time to fix bugs and add features. 

While it is possible for the account that the contract is deployed on to nullify its permissions in a way that makes the contract immutable (short of privileged actions taken by a super-majority of BPs), this is not recommended practice since there will always be bugs discovered or reasons to need to update the contract code in order to just maintain the same behavior that users of the contract _expect_ (note that giving up one extreme of immutability does not require going to the other extreme of a single developer unilaterally and instantly being able to update the contract; EOSIO enables the possibility of middle-ground solutions, but that is beyond the scope of this document). Even for smart contracts that are not meant to change, the specific behavior of the code may need to be tweaked either because: a bug was discovered and the code needs to be modified to match the intent of the code that both the developers and users of the contract agreed upon (as described by the Ricardian contract); or a dependency of the contract (or the platform itself) changes behavior after a sufficient deprecation period in a way that requires the contract to be updated to use new patterns/intrinsics/actions to maintain its same behavior going forward.

A smart contract that depends on another smart contract may wish to play it safe and automatically shut down its functionality (in an effort to protect its users) if a smart contract that it depends on but does not fully control has changed compared to the pinned version that it knows was safe (perhaps because auditors it trusts have fed in a specific version of the contract dependency as safe to use). 

For example, it may be desirable for a contract deployed on account `contracta` to only send an inline action `contractb::foo` authorized with the `contracta@active` permission to the contract deployed on account `contractb` if it is confident that the behavior of `contractb::bar` will be identical to that which was concluded by the contract developer at the time of writing the contract. Perhaps they are concerned about `contractb` changing its behavior during the `contractb::bar` action to bill `contracta` more RAM than expected. So the contract developer may want to include the hash of the WASM code that is deployed on `contractb` in a table accessible to `contracta` and write code in the contract deployed on `contracta` to actually assert that the current code hash of `contractb` matches the store hash in the table prior to sending the inline action.

[...]

PR #7529, opened 2019-06-25
add api to get last code update time of account

xinyijun:
## Change Description
Sometimes we need to detect whether a specific contract code has changed since last time from within contract execution.

## API Changes
Add a API to get last update time of contract.

### DIFF
--- a/libraries/chain/wasm_interface.cpp
+++ b/libraries/chain/wasm_interface.cpp
@@ -877,6 +877,13 @@ class permission_api : public context_aware_api {
          return time_point(acct->creation_date).time_since_epoch().count();
       }
 
+      int64_t get_account_last_code_update_time( account_name account ) {
+         auto* acct = context.db.find<account_object, by_name>(account);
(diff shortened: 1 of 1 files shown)


jafri:
Just usee get_account, it has last code update

jafri:
Modify the PR to use the protocol feature mechanism, similar to: https://github.com/EOSIO/eos/issues/6988

arhag:
This PR introduces a consensus change without the appropriate guards needed for consensus protocol feature. Please see how other protocol features introduced in v1.8.0 were implemented.

Also, we have an issue https://github.com/EOSIO/eos/issues/6168 for the design of a new intrinsic that would allow getting the last code update time as well as the code hash and code sequence number. I would expect a protocol feature introducing an intrinsic that gets metadata about the contract deployed on an account to provide all that info as discussed in the design of the `CODE_VERSION` protocol feature.

b1bart:

Issue #7313, opened 2019-05-08
add more2 field to results of get_table_rows to better support pagination


arhag:
Other RPC calls in chain API that allow pagination have  a `more` field that can support the same type as the `lower_bound` and `upper_bound` fields of the request. This enables nodeos to include the key that would need to be passed into the `lower_bound` on a subsequent call to continue retrieving the remaining results from the initial request (ignoring any state changes since that time).

However, the `get_table_rows` RPC was implemented using the old pattern of providing a boolean `more` which does not convey enough information to allow a client to generally handle pagination. 

So a new `string` field `more2` should be added to `get_table_rows_result` which is kept as the empty string when `more == false`, otherwise the string will be set to the key of the next row that wasn't returned in the response in the same format that was expected for the `lower_bound` and `upper_bound` fields of the request.

heipacker:
sounds good

andresberrios:
Why would you call it `more2` instead of something meaningful like `next_key` or something?

arhag:
> Why would you call it `more2` instead of something meaningful like `next_key` or something?

Sure. That sounds good.

mharrisb1:
@lparisc tagging you and labeling documentation because work is done for this but it is not documented. Please add new `next_key` field to `get_table_rows` documentation https://developers.eos.io/manuals/eos/latest/nodeos/plugins/chain_api_plugin/api-reference/index#operation/get_table_rows

PR #7530, opened 2019-06-25
Add more2 to get_table_rows API

andriantolie:
## Change Description

Related to https://github.com/EOSIO/eos/issues/7313. This PR adds a new field called `more2` to the result of get_table_rows chain API. The `more2` field contains the key of the next row that wasn't returned in the response in the same format that was expected for the `lower_bound` and `upper_bound` fields of the request, i.e.
- without --key-type -> the more2 field will contain uint64_t of the next row primary key; 
- --key-type sha256 ->  the more2 field will contain sha256 of the next row secondary key
- --key-type float128, the more2 field will contain floating number represented as uint128 of the next row secondary key
- etc...

This PR removes fixed_key.hpp as it's unused anymore and bring in fixed_bytes.hpp from eosio.cdt.

This PR also introduces breaking API changes for the input of get_table_rows for key-type sha256, ripemd160, i256. It also introduces breaking changes in the representation of uint128 in the JSON response (details below). 

A new PR is also added in the FC repo https://github.com/EOSIO/fc/pull/111

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

This PR changes the input of get_table_rows for the following key-type:
- sha256 -> convert the input to an array of two uint128_t in little endian (which is the same as how fixed_bytes represents the data)
- i256 -> convert the input to an array of two uint128_t in little endian (which is the same as how fixed_bytes represents the data)
- ripemd160  -> convert the input to an array of two uint128_t in little endian (which is the same as how fixed_bytes represents the data)
And this PR changes the way uint128 being represented in the JSON response, from previously being represented as binary, it is now being represented as a decimal which is easier to be understood by the user.

## Documentation Additions
- [ ] Documentation Additions


jafri:
Any consideration on giving it a better name like `next_key` instead of `more2`?

andriantolie:
> Any consideration on giving it a better name like `next_key` instead of `more2`?
That sounds good, thanks for the input. I've made the change in this PR.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -625,13 +642,11 @@ class read_write {
...
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wstrict-aliasing"
-            k[0] = ((uint128_t *)&v._hash)[0]; //0-127
-            k[1] = ((uint128_t *)&v._hash)[1]; //127-256
-#pragma GCC diagnostic pop
-            return k;
+            // The input is in big endian, i.e. f58262c8005bb64b8f99ec6083faf050c502d099d9929ae37ffed2fe1bb954fb
+            // fixed_bytes will convert the input to array of 2 uint128_t in little endian, i.e. 50f0fa8360ec998f4bb65b00c86282f5 fb54b91bfed2fe7fe39a92d999d002c5
+            // which is the format used by secondary index
+            fixed_bytes<32> fb(*reinterpret_cast<const std::array<const uint8_t, 32>*>(v.data()));
```
tbfleming: Since `std::array` is a struct, this is a strict aliasing violation. It's probably safest to memcpy from `v.data()` into a local array, then pass that array to the `fixed_bytes` constructor.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -643,10 +658,11 @@ class read_write {
...
      static auto function() {
         return [](const input_type& v) {
-            chain::key256_t k;
-            memset(k.data(), 0, sizeof(k));
-            memcpy(k.data(), v._hash, sizeof(v._hash));
-            return k;
+            // The input is in big endian, i.e. 83a83a3876c64c33f66f33c54f1869edef5b5d4a000000000000000000000000
+            // fixed_bytes will convert the input to array of 2 uint128_t in little endian, i.e. ed69184fc5336ff6334cc676383aa883 0000000000000000000000004a5d5bef
+            // which is the format used by secondary index
+            fixed_bytes<20> fb(*reinterpret_cast<const std::array<const uint8_t, 20>*>(v.data()));
```
tbfleming: also here

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -657,9 +673,12 @@ class read_write {
      using index_type = chain::index256_index;
      static auto function() {
         return [](const input_type v) {
+            // The input is in little endian of uint256_t, i.e. fb54b91bfed2fe7fe39a92d999d002c550f0fa8360ec998f4bb65b00c86282f5
+            // fixed_bytes will convert the input to array of 2 uint128_t in little endian, i.e. 50f0fa8360ec998f4bb65b00c86282f5 fb54b91bfed2fe7fe39a92d999d002c5
+            // which is the format used by secondary index
             chain::key256_t k;
-            k[0] = ((uint128_t *)&v)[0]; //0-127
-            k[1] = ((uint128_t *)&v)[1]; //127-256
+            k[0] = ((uint128_t *)&v)[1]; //128-255
```
tbfleming: `boost::multiprecision::uint256_t` is a class type, so this is a strict-aliasing violation.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1476,6 +1476,16 @@ uint64_t convert_to_type(const string& str, const string& desc) {
    }
 }
 
+template<>
+uint128_t convert_to_type(const string& str, const string& desc) {
+   uint128_t val;
+   try {
+      auto bmp = boost::multiprecision::uint128_t(str);
+      val = *reinterpret_cast<uint128_t*>(&bmp);
```
tbfleming: Strict aliasing violation; `boost::multiprecision::uint128_t` is a class type

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1548,6 +1595,11 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
          });
       }
       else if (p.key_type == chain_apis::float128) {
+         if ( p.encode_type == chain_apis::hex) {
+            return get_table_rows_by_seckey<index_long_double_index, uint128_t>(p, abi, [](uint128_t v)->float128_t{
+               return *reinterpret_cast<float128_t *>(&v);
```
tbfleming: Strict aliasing. `float128_t` is a struct.

tbfleming:
The `float128_t` aliasing problem also occurs in existing consensus-handling code. For now, keep the `float128_t` handling as is, but fix the other aliasing problems. `float128_t` fixes can come later: #7960

andriantolie:
@tbfleming Thanks for the feedback. I've updated my PR to fix the strict alias violation.

aclark-b1:
Docs team - I don't believe this change ever made it into our documentation.  Can you please add it?

PR #7540, opened 2019-06-27
bios-boot-tutorial.py: bugfix, SYS hardcoded instead of using command…

cc32d9:
## Change Description

Minor bugfix in testnet spin-up script. In one place, it uses hardcoded SYS instead of symbol used in --symbol command-line argument

## Consensus Changes

none

## API Changes

none

## Documentation Additions

none


heifner:
Note this needs to be targeted to `develop` and `release/1.8.x`

PR #7553, opened 2019-06-28
[develop] CMAKE version check before dependencies are installed

NorseGaud:
## Change Description
Problem: People with versions less than the required don't see the requirement failure until too late (after dependencies are installed). 

Solution: Add a check for the required version of CMAKE (using CMakeLists.txt) and compare it to the currently installed version.

Base Images pipeline tests & builds : https://buildkite.com/EOSIO/eosio-base-images/builds/339

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/.build_vars
+++ b/scripts/.build_vars
@@ -7,7 +7,6 @@ export LIB_DIR=${EOSIO_INSTALL_DIR}/lib
 export DATA_DIR=${EOSIO_INSTALL_DIR}/data
 
 # CMAKE
-export CMAKE_REQUIRED_VERSION=$(cat $REPO_ROOT/CMakeLists.txt | grep -E "^[[:blank:]]*cmake_minimum_required[[:blank:]]*\([[:blank:]]*VERSION" | tail -1 | sed 's/.*VERSION //g' | sed 's/ //g' | sed 's/"//g' | cut -d\) -f1)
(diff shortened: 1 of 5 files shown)


scripts/.build_vars
```diff
@@ -7,6 +7,7 @@ export LIB_DIR=${EOSIO_INSTALL_DIR}/lib
 export DATA_DIR=${EOSIO_INSTALL_DIR}/data
 
 # CMAKE
+export CMAKE_REQUIRED_VERSION=$(cat $REPO_ROOT/CMakeLists.txt | grep -E "cmake_minimum_required\([[:blank:]]*VERSION" | tail -1 | sed 's/.*VERSION //g' | cut -d\) -f1 | sed 's/ //g')
```
arhag: This could safely go into `.environment` instead so it is only called once in the script.

Also for further resilience it could be:
```
export CMAKE_REQUIRED_VERSION=$(cat $REPO_ROOT/CMakeLists.txt | grep -E "^[[:blank:]]*cmake_minimum_required[[:blank:]]*\([[:blank:]]*VERSION" | tail -1 | sed 's/.*VERSION //g' | sed 's/ //g' | sed 's/"//g' | cut -d\) -f1)
```

Issue #7560, opened 2019-06-29
build script INSTALL_LOCATION should support relative paths


arhag:
Specifying a relative path with the `-i` option of the build script doesn't work correctly. It should set the `INSTALL_LOCATION` variable to an absolute path computed from the provided relative path taken relative to the current working directory.

dawnFromDark:
got same error about this

PR #7578, opened 2019-07-01
[develop] -i support for relative paths

NorseGaud:
## Change Description

https://github.com/EOSIO/eos/issues/7560

Base Images (with BATS)
https://buildkite.com/EOSIO/eosio-base-images/builds/368

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,9 +68,10 @@ $ sudo yum remove eosio
 
 #### Build Script Uninstall
 
-If you have previously installed EOSIO using build scripts, you may execute `./scripts/eosio_uninstall.sh` to uninstall.
-- Passing `--force` will answer yes to all prompts
-- Passing `--full` will remove data directories (be very careful with this)
+If you have previously installed EOSIO using build scripts, you can execute `eosio_uninstall.sh` to uninstall.
+- Passing `-y` will answer yes to all prompts
+- Passing `-f` will remove data directories (be very careful with this)
+- Passing in `-i` allows you to specify where your eosio installation is located
 
 ## Supported Operating Systems
 EOSIO currently supports the following operating systems:  
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ $ sudo yum remove eosio
 #### Build Script Uninstall
 
 If you have previously installed EOSIO using build scripts, you can execute `eosio_uninstall.sh` to uninstall.
-- Passing `-y` will answer yes to all prompts (forces `-f`)
+- Passing `-y` will answer yes to all prompts (does not remove data directories)
 - Passing `-f` will remove data directories (be very careful with this)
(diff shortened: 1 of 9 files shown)


scripts/eosio_uninstall.sh
```diff
@@ -1,89 +1,125 @@
...
+VERSION=2.1
+# User input handling
+DEP_PROCEED=false
+FULL=false
+
+function usage() {
+   printf "Usage: $0 OPTION...
+  -i DIR      Directory where eosio is installed)
+  -y          Noninteractive mode (answers yes to every prompt)
+  -f          Removal of data directory (be absolutely sure you want to delete it before using this!)
```
arhag: Please update the README as well to reflect these changes to the uninstall script.

scripts/eosio_uninstall.sh
```diff
@@ -1,89 +1,125 @@
...
+      read -p "By specifying -f, removal of the eosio data directory will require a resync of data which can take days. Do you wish to proceed? (y/n) " PROCEED
+      case $PROCEED in
+         "" ) echo "What would you like to do?";;
+         0 | true | [Yy]* ) break;;
+         1 | false | [Nn]* ) exit 0;;
+         * ) echo "Please type 'y' for yes or 'n' for no.";;
+      esac
+   done
+elif $NONINTERACTIVE && $FULL; then
+   echo "${COLOR_RED}By specifying -f, removal of the eosio data directory will require a resync of data which can take days. The script is sleeping for 15 seconds to give you a chance to cancel if this is not what you want.${COLOR_NC}"
```
arhag: If the user has specified `-y` for non-interactive (which is not the default), I don't think we should be waiting in the script for a user action that is likely not going to come.

README.md
```diff
@@ -68,9 +68,10 @@ $ sudo yum remove eosio
 
 #### Build Script Uninstall
 
-If you have previously installed EOSIO using build scripts, you may execute `./scripts/eosio_uninstall.sh` to uninstall.
-- Passing `--force` will answer yes to all prompts
-- Passing `--full` will remove data directories (be very careful with this)
+If you have previously installed EOSIO using build scripts, you can execute `eosio_uninstall.sh` to uninstall.
+- Passing `-y` will answer yes to all prompts (forces `-f`)
```
arhag: I don't think `-y` should force `-f`. They should be independent.

`-y` alone should not remove the data directories and not prompt the user for anything. `-f` alone should remove the data directories after prompting the user if they are sure. `-f -y` should proceed forward with removing the data directories without prompting the user.

NorseGaud: Ah, I misunderstood. I'll get that fixed

Issue #7536, opened 2019-06-26
cleos send transaction command


dixia:
it looks like the new RPC API send_tranasction has not been added to cleos.

this is send_tranasction function in cleos main.cpp but it is not added to the command list yet or maybe it is not meant to implement this RPC API.

https://github.com/EOSIO/eos/blob/5082391c60b0fa5e68157c385cd402bf25aea934/programs/cleos/main.cpp#L513

Could the new send transaction API be added to cleos command list and pointed to the correct endpoint?

arhag:
The function you linked to is an implementation-specific function for cleos that has been there for a while. However, coincidentally, EOSIO v1.8.x does introduce a new RPC called `send_transaction` to the chain API.

You are correct that cleos does not currently support using the `send_transaction` RPC. One option is to add an option to the `push action` and `push transaction` cleos commands to force the use of either the `send_transaction` RPC or the `push_transaction` RPC (they basically do the same thing, except `push_transaction` rearranges the returned trace to avoid breaking compatibility with existing clients that don't support the new transaction trace structure). 

I would prefer if the default became using `send_transaction` because that is the more efficient RPC of the two (it doesn't have to go through the additional rearrangement step before returning the response). However, that won't work if cleos is talking to an older version of nodeos (before v1.8.x) that doesn't support the `send_transaction` RPC. Cleos could use the `/v1/node/get_supported_apis` endpoint to discover whether the node it is talking to supports `send_transaction`, and if it doesn't support it, it would fall back to the `push_transaction`.

aaron-fu-block-one:
I would like to take this issue, and enhance `cleos push transaction` as proposed by @arhag:

- When `cleos push transaction` is called, use `get_supported_apis` to detect whether the node supports the newly introduced `send_transaction` RPC or not:
    - If the node supports, the more efficient `send_transaction` will be called
    - If it does not, the traditional `push_transaction` will be called
- If possible, will add an option to `cleos push transaction`, which will force using the old RPC `push_transaction` across all the cases.

matthewdarwin:
FWIW, as a block producer, I don't expose the `get_supported_apis` endpoint as there is no way to control what APIs are listed.  I have more APIs available in nodeos that I don't make publicly available.  So using this logic while I would support send_transaction, you wouldn't know. (eg I might enable the 'net' plugin, so I can use the functions, but not make that available publicly).

I also find it interesting that `get_supported_apis` is not listed in the available api list it returns.

[...]

PR #7585, opened 2019-07-02
Enhance cleos to enable new RPC send_transaction

aaron-fu-block-one:
## Change Description

With regard to issue #7536, enhance `cleos` so as to enable `send_transaction`, an RPC newly introduced in  EOSIO `v1.8`.

Changes are made to `programs/cleos/main.cpp` and `programs/cleos/httpc.hpp`.

**Background**

- For versions `v1.8` or after, both RPC `send_transaction` and `push_transaction` are supported.
- For versions before `v1.8`, only RPC `push_transaction` is supported.
- RPC `send_transaction` and `push_transaction` are almost the same, except the traces they return, which can be viewed in JSON format by adding a `-j` option in the command calling them.

**Flow Control**

- Users can enforce always using old RPC `push_transaction` by adding a flag `--use-old-rpc`.
- Otherwise, which RPC will be called depends on the current `nodeos` version:
    - If the `nodeos` version is between `v1.0` and `v1.7`, call old RPC `push_transaction`;
    - Otherwise, call new RPC `send_transaction`.

**Test**

Have tested behavior such as `create account`, `issue` and `transfer` for `nodeos` version `v1.8.0` and version `v1.7.4`.

**Sidenote**

In `programs/cleos/main.cpp` there is a `void send_transaction(...)` function, which, according to @arhag in issue #7536, is simply ***unrelated*** to this topic. Just ignore it.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -343,14 +343,17 @@ fc::variant push_transaction( signed_transaction& trx, packed_transaction::compr
    }
 
    if (!tx_dont_broadcast) {
-      bool send_txn_func_supported = true;
-      if (info.server_version_string) {
-        int major, minor;
-        if (sscanf(info.server_version_string->c_str(), "v%d.%d", &major, &minor) == 2) {
-          send_txn_func_supported = !(major == 1 && minor < 8);
-        }
+      if (tx_use_old_rpc) {
+         return call(push_txn_func, packed_transaction(trx, compression));
+      } else {
+         try {
+            return call(send_txn_func, packed_transaction(trx, compression));
+         }
+         catch (chain::missing_chain_api_plugin_exception &) {
+            std::cerr << "New RPC send_transaction may not be supported. Add flag --use-old-rpc to use old RPC push_transaction instead." << std::endl;
+            throw;
+         }
       }
-      return call(!tx_use_old_rpc && send_txn_func_supported ? send_txn_func : push_txn_func, packed_transaction(trx, compression));
    } else {
       if (!tx_return_packed) {
         return fc::variant(trx);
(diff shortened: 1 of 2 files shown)


programs/cleos/main.cpp
```diff
@@ -341,7 +343,14 @@ fc::variant push_transaction( signed_transaction& trx, packed_transaction::compr
    }
 
    if (!tx_dont_broadcast) {
-      return call(push_txn_func, packed_transaction(trx, compression));
+      bool send_txn_func_supported = true;
+      if (info.server_version_string) {
+        int major, minor;
+        if (sscanf(info.server_version_string->c_str(), "v%d.%d", &major, &minor) == 2) {
```
spoonincode: I'm not sure this is a good idea; not everyone who forks eosio (telos, mainnet, graymass) maintains this version string format.

arhag: I agree with @spoonincode. Furthermore, there is a good reason not to change which RPC is called due to reasons outside of the users control (the reported version of the node cleos talks to is outside the user's control), because the different RPCs return slightly different traces.

The code should simply use `push_txn_func` when `tx_use_old_rpc` is true; otherwise it should use ``send_txn_func`.

However, it would be good to go further and recognize if the call to `send_txn_func` failed because the endpoint wasn't recognized, and instead provide the user with a more meaningful error message that recommends they use the `--use-old-rpc` option.

aaron-fu-block-one:
I have modified the code adopting @arhag 's suggestion. Now the new flow is

**Flow**

- Users can enforce using old RPC `push_transaction` by adding a flag `--use-old-rpc`.
- Otherwise, without such a flag, `cleos` will always call the new RPC `send_transaction`.
    - If it happens that the call to `send_transaction` fails, an error message will be printed out, saying that the new RPC may not be supported, and suggesting the user add the flag `--use-old-rpc` to call `push_transaction` instead.

PR #7601, opened 2019-07-05
fix 7600 double confirm after changing sign key

### LINKED ISSUE #7600: Producer double confirm bug after changing sign key that results in producing blocks by different nodeos
similar to https://github.com/EOSIO/eos/issues/3835,  Under following scenario there would be double confirmation bug:

- node x is producing under account a1 using signing key k1
- a client change a1's singing key from k1 to k2
- new schedule takes place after a while
- node y receive block that produced by a1 and signed by k1
- node y start to produce under account a1 using signing key k2, but in node y's point of view, the watermark of a1 is the first time a1 was selected, resulting in double confirm bug.

proposed fix:
- reset water mark for account x if we don't have the old sign key of x (which means we were not producing for account x) upon new schedule
- also check with last produced block number in head_block_state, as we may receive a block with same producer but old signing key.

taokayan:
## Change Description
This fix #7600  double confirm bug after changing signing key which results in producing by different nodeos

## Consensus Changes
No. Changes are subjective.

## API Changes
No
## Documentation Additions
No

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -280,31 +280,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             }
          } ) );
 
-         // since the watermark has to be set before a block is created, we are looking into the future to
-         // determine the new schedule to identify producers that have become active
-         chain::controller& chain = chain_plug->chain();
-         const auto hbn = bsp->block_num;
-         auto new_pbhs = bsp->next(bsp->header.timestamp.next(), 0);
-
-         // for newly installed producers we can set their watermarks to the block they became active
-         if( bsp->active_schedule.version != new_pbhs.active_schedule.version ) {
-            flat_set<account_name> new_producers;
-            new_producers.reserve(new_pbhs.active_schedule.producers.size());
-            for( const auto& p: new_pbhs.active_schedule.producers) {
-               if (_producers.count(p.producer_name) > 0)
-                  new_producers.insert(p.producer_name);
-            }
-
-            for( const auto& p: bsp->active_schedule.producers) {
-               // clear watermark unless we have the old key (we were producing before)
-               if (_signature_providers.count(p.block_signing_key) > 0) {
-                  new_producers.erase(p.producer_name);
-               }
-            }
-
-            for (const auto& new_producer: new_producers) {
-               _producer_watermarks[new_producer] = hbn;
-            }
+         // simplify handling of watermark in on_block
+         if (bsp->block) {
+            _producer_watermarks[bsp->block->producer] = bsp->block_num;
          }
       }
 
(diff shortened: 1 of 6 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -280,28 +280,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-            for( const auto& p: bsp->active_schedule.producers) {
-               new_producers.erase(p.producer_name);
-            }
-
-            for (const auto& new_producer: new_producers) {
-               _producer_watermarks[new_producer] = hbn;
-            }
+         // simplify handling of watermark in on_block
+         if (bsp->block) {
+            _producer_watermarks[bsp->block->producer] = bsp->block_num;
```
arhag: You don't need the extra indirection to the through the `block` shared pointer to get to the producer. You can access it via `bsp->header.producer`. This would also work if `bsp->block` isn't present, although I suppose that should never be the case in this context.

This only needs to be done for producers that are configured on the local node. So only if `bsp->header.producer` can be found inside `_producers`.

tests/nodeos_producer_watermark_test.py
```diff
@@ -0,0 +1,271 @@
...
+ClientName="cleos"
+
+try:
+    TestHelper.printSystemInfo("BEGIN")
+    cluster.setWalletMgr(walletMgr)
+
+    cluster.killall(allInstances=killAll)
+    cluster.cleanup()
+    Print("Stand up cluster")
+    if cluster.launch(prodCount=prodCount, onlyBios=False, pnodes=totalNodes, totalNodes=totalNodes, totalProducers=totalNodes, useBiosBootFile=False, sharedProducers=1) is False:
```
arhag: I don't want this test to do all the unnecessary work of bootstrapping token and system contracts. The only contract that needs to be deployed for this test is the `eosio.bios` contract so that the `setprods` action can be used to change the producer schedule.

tests/nodeos_producer_watermark_test.py
```diff
@@ -0,0 +1,271 @@
...
+###############################################################
+# nodeos_voting_test
+# --dump-error-details <Upon error print etc/eosio/node_*/config.ini and var/lib/node_*/stderr.log to stdout>
+# --keep-logs <Don't delete var/lib/node_* folders upon test completion>
+###############################################################
+class ProducerToNode:
+    map={}
+
+    @staticmethod
+    def populate(node, num):
```
arhag: Why is map a static variable modified by a static function? Avoid globals. Make it an instance variable if necessary, or just define the function to mutate the map that is passed into the function as an argument.

arhag: In fact, is this even necessary? It seems it is used only for extra information given in error messages and I'm not sure how much value it actually adds. 

That data is also not exactly accurate for shared producers because they belong to more than one node. 

I think it is totally reasonable to just print out the producers assigned to each node, then if someone is investigating the logs to debug an error, they can easily figure out which nodes they belonged to by looking it up in the assignments printed in the logs.

tests/nodeos_producer_watermark_test.py
```diff
@@ -0,0 +1,271 @@
...
+    if blockProducer not in prodsActive:
+        Utils.cmdError("unexpected block producer %s at blockNum=%s" % (blockProducer,blockNum))
+        Utils.errorExit("Failed because of invalid block producer")
+    if not prodsActive[blockProducer]:
+        Utils.cmdError("block producer %s for blockNum=%s not elected, belongs to node %s" % (blockProducer, blockNum, ProducerToNode.map[blockProducer]))
+        Utils.errorExit("Failed because of incorrect block producer")
+    prodsSeen[blockProducer]=True
+    return blockProducer
+
+def setActiveProducers(prodsActive, activeProducers):
```
arhag: This function is not used.

PR #7602, opened 2019-07-05
Reuse producer threads

baegjae:
## Change Description

**Description**
- Producer threads perform core operations in producer_plugin, such as `push_transaction`, `push_block`, and `produce_block`. 
- The purpose of this PR is to observe the impact of separating these operations from the application thread. 
- Note that this code works well, but not concrete.

**Motivation and Work Notes**
- Application thread is still overhead despite the help of chain-threads, net-threads, and http-threads.
- Our first intuition was to separate `push_transaction` from the application thread.
- However, `push_transaction` is tightly coupled with `push_block` and `produce_block`, therefore we needed to separate them all.
- The producer threads were idle since #6845 (https://github.com/EOSIO/eos/pull/6845/commits/eceaa9bb5cd551e3d38b15ddeeb5404421fdc786) and were suitable for performing separated operations.
- We limit the number of producer threads to 1, because currently sequential processing must be guaranteed.
- We carefully changed the processing of `push_transaction`, `push_block`, `produce_block` from application thread to producer threads.

**Experiment Environment**
- Nodes: 8 nodes
- HW: Each Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz (44 cores), 128GB
- Options: 2 chain-threads, 2 net-threads, 2 http-threads, 2 txn_test_gen-threads, 1 producer-thread, wavm
- Benchmark: txn_test_gen - generate 200 transactions every 170 ms on each 8 node (intended 9,400 TPS) for 30 seconds

**Results**
- Before: 7,832 TPS, 99% CPU usage (application thread)
- After: 8,512 TPS, 7.1% CPU usage (application thread), 99% CPU usage (producer thread)

**Discussion**
- Producer threads increases TPS by 9% and reduces application thread CPU usage by 99% to 7%.
- We can observe that the core operations in producer_plugin accounted for a large portion of the application thread previously.
- If we evenly distribute the core operations in producer_plugin to producer threads and application thread, we can achieve higher throughput.
- If we can increase the number of producer threads (e.g., the parallel processing of push transaction is supported), we can easily achieve much higher throughput. Because the CPU usage of the application thread is low currently.
- We hope that this PR will help design multithreading techniques.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -654,7 +654,7 @@ void producer_plugin::set_program_options(
           "ratio between incoming transations and deferred transactions when both are exhausted")
          ("incoming-transaction-queue-size-mb", bpo::value<uint16_t>()->default_value( 1024 ),
           "Maximum size (in MiB) of the incoming transaction queue. Exceeding this value will subjectively drop transaction with resource exhaustion.")
-         ("producer-threads", bpo::value<uint16_t>()->default_value(config::default_controller_thread_pool_size),
+         ("producer-threads", bpo::value<uint16_t>()->default_value( 1 ),
           "Number of worker threads in producer thread pool")
          ("snapshots-dir", bpo::value<bfs::path>()->default_value("snapshots"),
           "the location of the snapshots directory (absolute path or relative to application data dir)")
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -449,7 +452,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
          auto after_sig_recovery =
                [self = this, trx, persist_until_expired, next{std::move(next)}]() mutable {
-                  app().post(priority::low, [self, trx{std::move(trx)}, persist_until_expired, next{std::move(next)}]() {
+                  self->_self->post(priority::low, [self, trx{std::move(trx)}, persist_until_expired, next{std::move(next)}]() {
                      self->process_incoming_transaction_async( trx, persist_until_expired, next );
```
heifner: `chain::controller` is only safe to be accessed from the application thread. All calls into `controller` must be on the application thread.

`process_incoming_transaction_async` calls `chain.push_transaction` as well as other controller methods.

baegjae:
> Unfortunately these changes are not thread-safe. The controller must only be accessed by the main application thread.

Thank you for your comments and interest. The purpose of this PR was to experimentally observe the behavior when processing `push_transaction` in producer threads. More things need to be considered in order to actually merge. I will set up this practical implementation as a future plan, and I will close this PR.

PR #7625, opened 2019-07-09
Fix flaky tests - mainly net_plugin fixes

heifner:
## Change Description

This PR restores stability to `develop` tests including the Long Running Tests (LRT). It includes a number of changes, mostly fixes to `net_plugin` which have been lingering for a number of releases as well as new `develop` only issues introduced in the multi-threaded `net_plugin` work.

## `net_plugin` fixes
- Access boost asio acceptor only from connection strand
- Use non-deprecated boost `async_resolve` method
- Use non-deprecated boost `async_connect` method
- Dispatch blocks via a `dispatch_manager` strand to avoid blocks being sent out of order
- Recreate socket object on reuse
- Use shared_ptr to socket to make sure it lives for lifetime of use
- Avoid doing extra non-needed work on invalid peer_requested blocks
- Fix issue with `head_catch`  not sending a `request_message`. This is similar to  #7660 only in another location where `verify_catchup` is used.
- Fix issue with block already received optimization that was incorrectly recording rejected blocks. This included rejected blocks that were unlinkable because they arrived too early. This prevented the blocks from being processed when they came again in the correct order.
- Fixed issue with non-accepted blocks via `fc::exception` being propagated and recorded as valid.

## test fixes

- Fixed error in python `tests/Node.py` to correctly log block NOT found when not found
- Added 1 second delay to mongo getHeadBlockNum since mongo is very slow on mac test machines
- Add additional time to `db_modes_test` since current delay doesn't appear to be long enough to release resources
- Note not all the above additional time may not be required as they were added early in the investigation of this PR

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/db_modes_test.py
```diff
@@ -42,10 +42,11 @@ def run_and_kill(extraCmd="", killsig=signal.SIGTERM, preCmd=""):
             tries = tries - 1
             port = port + 1
     # now start nodeos
+    time.sleep(5)
```
spoonincode: Is this intended to be part of the PR?

tests/nodeos_startup_catchup.py
```diff
@@ -142,6 +142,7 @@ def waitForNodeStarted(node):
     Print("Cycle through catchup scenarios")
     twoRounds=21*2*12
     for catchup_num in range(0, catchupCount):
+        time.sleep(5)
```
spoonincode: likewise

tests/db_modes_test.py
heifner: Yes. This PR morphed into get the tests passing PR.

tests/nodeos_startup_catchup.py
heifner: Yes. This PR morphed into get the tests passing PR.

PR #7654, opened 2019-07-16
exclusively use timer for checktime

spoonincode:
## Change Description
pre-1.4 nodeos would frequently poll the system time during contract execution (to ensure a transaction wasn’t executing beyond its allowed time). For most platforms this isn’t a syscall (see vDSO) but it still was a significant overhead.

1.4 (via #5799) introduced a checktime timer where a kernel timer — via BSD interval timer interface — would be scheduled instead. This allows nodeos to skip polling of system time until the timer expires. #7449 improves this code further by migrating to the POSIX timer API on Linux and using the kevent timer on macOS.

The way this timer mechanism was structured was to expire the timer early and then fall back to polling time when the elapsed time for a transaction approached its deadline. The reason behind this was to ensure the utmost accuracy is maintained for checktime.

However, in practice, the checktime timer is very accurate (very low latency) across platforms we support. Some average timer expiry latency on various platforms running nodeos that I measured:

Platform | Timer Latency
--- | ---
2700X & 9600k Linux | `min:1us max:6us mean:2us stddev:1us`
AWS C5 instance (Nitro based) | `min:3us max:16us mean:6us stddev:1us`
AWS C4 instance (xen based) | `min:18us max:171us mean:77us stddev:32us`
AWS A1 instance (ARM8) | `min:6us max:19us mean:10us stddev:2us`
Raspberry Pi 3 B+ Linux | `min:15us max:64us mean:22us stddev:15us`
2018 MBP, macOS 10.14 | `min:3us max:102us mean:27us stddev:22us`

Across the board the average latency for the timer is under 100us; and if we only look at non-xen Linux platforms the average is under 25us inclusive of a Raspberry Pi 3. This is plenty accurate enough alone.

This PR removes the "gear down" from timer to poll and instead strictly uses a timer. It no longer even tries to adjust the timer expiration to accommodate latency. Though a print at the beginning still gives the latency statistics that were measured.

Motivation for this change is that significant optimizations to wasm performance can be achieved when strictly using a timer. (this PR contains no such performance improvement; it simply paves the way for it)

Some platforms don't have an accurate timer. For example, Win32 and FreeBSD. These platforms will simply not be ideal to run a nodeos on when accurate deadline enforcement is required. It is worth noting that eosio does not currently run on nor support those two example platforms. Nevertheless, a boost::asio based checktime timer is included for compilation when a platform does not have POSIX timers (such as win32). It was validated as passing unit tests on Linux.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/transaction_context.cpp
```diff
@@ -156,13 +156,13 @@ namespace eosio { namespace chain {
...
 
-      checktime(); // Fail early if deadline has already been exceeded
-
       if(control.skip_trx_checks())
          deadline_timer.start(fc::time_point::maximum());
       else
          deadline_timer.start(_deadline);
 
+      checktime(); // Fail early if deadline has already been exceeded
+
```
spoonincode: There is somewhat of a landmine above. The checktime timer needs to start before `checktime()` is ever called, otherwise the timer will still be still be in an expired state and cause the `checktime()` to fail. This wasn't a problem previously because an expired checktime timer simply caused `checktime()` to poll the time.

Not sure the best way of trying to guard against someone falling in to that trap in the future.

b1bart: Can the `deadline_timer` be something that the calling code constructs/initializes and moves into a `transaction_context` as part of construction? that would make it impossible to write a call to `transaction_context::checktime` without having a valid and initialized `deadline_timer`.  That would move the `deadline_timer.start` call quite a bit and I'm not sure what consequences that would have

spoonincode: >Can the deadline_timer be something that the calling code constructs/initializes and moves into a transaction_context as part of construction?

Unless I'm misunderstanding, this won't work very well post #7449 because now checktime_timer on construction acquires a system resource i.e. it does some syscalls and thus has overhead I'd really like to avoid performing for each transaction. That is why the timer was moved to live in controller, so that it is only created once.

I could add something like a `checktime::reset()` that would set `expired = 0`; the transaction_context could call this in its ctor. This would ensure any `checktime()` calls before transaction_context got around to starting the timer would succeed. Still a bit hokey I know. Any other ideas?

b1bart: Right sorry, didn't remember its a passed in reference.  I think its the responsibility of the calling code (which is providing that reference) to make sure it provides a "ready to use" `checktime_timer`, I agree it's pretty fragile to implicitly assume someone has handled this for you so... my hair-brained idea is to create an abstraction around it. 

 Something like a `checktime_timer_pool` which creates the system resource(s) and "loans" them out to `checktime_timer` objects (which come into life properly reset).  Those objects would return the system resource to the pool when they are destroyed (if nothing stole its resource with `std::move`).  

`transaction_context` would then expect a `checktime_timer&&` in its constructor and move it to a member.  The calling frame would never get that object back so, it cannot pass it to another `transaction_context`.  It would always be calling `checktime_timer_pool::take_one()` or something.   This makes the lifetime of the `checktime_timer` "safe" from the point of view of the `transaction_context` and provides an RAII style lifecycle management of the expensive system resource and the expected `reset`

spoonincode: I don't think we should get too carried away just yet with an actual pool. We all know multiple timers are going to be needed when multithreading comes but it's not clear at all how that truly comes together just yet. Don't want to try and predict the future too much here.

I tried to clean up the lack of RAII-ness by making controller move a transaction_checktime_timer in to the transaction_context upon its construction. This still isn't very pool-like since transaction_checktime_timer doesn't phone anything up to say it has been destroyed. But, at least it hides the initialization and cleanup from transaction_context unlike before where transaction_context had to manually create another scoped exit of sorts.

PR #7657, opened 2019-07-16
(chainbase sync) print name of DB causing failure condition & win32 fixes

spoonincode:
## Change Description

Gets EOSIO/chainbase#48 & EOSIO/chainbase#49

EOSIO/chainbase#49 (print name of DB causing failure condition) may be a 1.8 candidate

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
Though what's interesting is that for the nodeos case those can never be hit because the exception is swallowed up in appbase's init

Issue #3516, opened 2018-05-28
on system/compiler/boost change detection, (hard) replay takes 3 tries


spoonincode:
When the system/compiler/boost change detection is activated it takes 3 tries to recover.

```
$ programs/nodeos/nodeos -d ~/eos.data/f --config-dir ~/eos.data/f
3470986ms thread-0   chain_plugin.cpp:151          plugin_initialize    ] initializing chain plugin
3470987ms thread-0   main.cpp:119                  main                 ] /Users/spoon/eos-noon-net/libraries/chainbase/src/chainbase.cpp(70): Throw in function chainbase::database::database(const bfs::path &, chainbase::database::open_flags, uint64_t, bool)
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
std::exception::what: database created by a different compiler, build, boost version, or operating system
```
try 1:
```
$ programs/nodeos/nodeos -d ~/eos.data/f --config-dir ~/eos.data/f --hard-replay-blockchain
3512872ms thread-0   chain_plugin.cpp:151          plugin_initialize    ] initializing chain plugin
3512872ms thread-0   chain_plugin.cpp:233          plugin_initialize    ] Hard replay requested: deleting state database
3512873ms thread-0   block_log.cpp:324             repair_log           ] Recovering Block Log...
3512874ms thread-0   block_log.cpp:344             repair_log           ] Moved existing blocks directory to backup location: '/Users/spoon/eos.data/f/blocks-2018-05-28T21:58:32.873'
3512874ms thread-0   block_log.cpp:349             repair_log           ] Reconstructing '/Users/spoon/eos.data/f/blocks/blocks.log' from backed up block log
3512874ms thread-0   block_log.cpp:455             repair_log           ] Existing block log was undamaged. Recovered all irreversible blocks up to block number 1.
3512875ms thread-0   chain_plugin.cpp:431          recover_reversible_b ] Reconstructing '/Users/spoon/eos.data/f/blocks/reversible' from backed up reversible directory
3512875ms thread-0   main.cpp:119                  main                 ] /Users/spoon/eos-noon-net/libraries/chainbase/src/chainbase.cpp(70): Throw in function chainbase::database::database(const bfs::path &, chainbase::database::open_flags, uint64_t, bool)
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
[...]

PR #7678, opened 2019-07-24
Issue 3516 fix

ndcgundlach:
## Change Description

Extended try-catch to handle std::runtime_error when opening the old reversible database when doing a hard replay.  The database is opened permissively (allowing dirty) - this is no change, but the code now catches the exception when the database format is incompatible.  In this case, the old reversible database is not used.  It is backed up however. #3516 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -536,6 +536,27 @@ protocol_feature_set initialize_protocol_features( const fc::path& p, bool popul
    return pfs;
 }
 
+void
+chain_plugin::do_hard_replay(const variables_map& options) {
+         ilog( "Hard replay requested: deleting state database" );
+         clear_directory_contents( my->chain_config->state_dir );
+         auto backup_dir = block_log::repair_log( my->blocks_dir, options.at( "truncate-at-block" ).as<uint32_t>());
+         if( fc::exists( backup_dir / config::reversible_blocks_dir_name ) ||
+             options.at( "fix-reversible-blocks" ).as<bool>()) {
+            // Do not try to recover reversible blocks if the directory does not exist, unless the option was explicitly provided.
+            if( !recover_reversible_blocks( backup_dir / config::reversible_blocks_dir_name,
+                                            my->chain_config->reversible_cache_size,
+                                            my->chain_config->blocks_dir / config::reversible_blocks_dir_name,
+                                            options.at( "truncate-at-block" ).as<uint32_t>())) {
+               ilog( "Reversible blocks database was not corrupted. Copying from backup to blocks directory." );
+               fc::copy( backup_dir / config::reversible_blocks_dir_name,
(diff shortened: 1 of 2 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1062,57 +1067,60 @@ bool chain_plugin::recover_reversible_blocks( const fc::path& db_dir, uint32_t c
 
    ilog( "Reconstructing '${reversible_dir}' from backed up reversible directory", ("reversible_dir", reversible_dir) );
 
-   chainbase::database  old_reversible( backup_dir, database::read_only, 0, true );
-   chainbase::database  new_reversible( reversible_dir, database::read_write, cache_size );
-   std::fstream         reversible_blocks;
-   reversible_blocks.open( (reversible_dir.parent_path() / std::string("portable-reversible-blocks-").append( now ) ).generic_string().c_str(),
+   try {
```
brianjohnson5972: I think we should change this to:

fc::optional<chainbase::databes> old_reversible;
try {
   old_reversible = chainbase::database( backup_dir, database::read_only, 0, true );
} catch(const std::runtime_error&) {
   ilog( "Did not recover any reversible blocks since reversible database incompatible");
   // since we are allowing for dirty, it must be incompatible
   return true;
}

And keep the rest as it was, so that we do not scoop up any other exceptions that we have not analyzed.

ndcgundlach: I like that.  C++17 has an optional, can we use C++17 features?

heifner: We can use C++17 features. However, fc has an optional. Please use that until we move from fc option to C++17 optional.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1062,37 +1067,45 @@ bool chain_plugin::recover_reversible_blocks( const fc::path& db_dir, uint32_t c
...
-      ilog( "Did not recover any reversible blocks since the specified block number to stop at (${stop}) is less than first block in the reversible database (${start}).", ("stop", truncate_at_block)("start", start) );
-      return true;
-   }
-   try {
-      for( ; itr != ubi.end(); ++itr ) {
-         EOS_ASSERT( itr->blocknum == end + 1, gap_in_reversible_blocks_db,
+   	uint32_t num = 0;
+   	uint32_t start = 0;
+   	uint32_t end = 0;
+   	old_reversible->add_index<reversible_block_index>();
```
brianjohnson5972: Please remove all tabs from the code.  The only diffs that should show up between 1078 and 1118 should be old_reversible using "->" instead of "."

heifner: Evil tabs, nice catch @brianjohnson5972

PR #7686, opened 2019-07-25
Unapplied transaction queue performance

heifner:
## Change Description

- Applied transactions used to be removed from the unapplied transaction map when they were applied in `controller::push_transaction`. The removal of these applied transactions was moved into the `producer_plugin` during the processing of unapplied transactions. See #7453.  However, `push_block` transactions from other producers were not removed causing the `unapplied_transaction_queue` to grow with transactions that were already applied. This slowed down block production as all these transactions had to be processed to determine that they were duplicates (already applied).

Modifications:
- Remove from the `unapplied_transaction_queue` all applied transactions when signaled via `accepted_block`. 
- Add aborted transactions into the `unapplied_transaction_queue` before processing the next `push_block` so that applied transactions can be removed via `accepted_block`. Otherwise, If aborted transactions are added into the `unapplied_transaction_queue` after `push_block` many/most of these transactions are actually already applied as they came in via the `push_block`.
- Add forked transactions into the `unapplied_transaction_queue` before applying new branch so that applied transactions can be removed as applying new branch.
- Changed to only store forked out transactions when in speculative mode.
- Modified `unapplied_transaction_queue` to not store any transactions when not in speculative mode.
- Added some tests that verify duplicates are not allowed in the `unapplied_transaction_queue`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/controller.cpp
```diff
@@ -1962,7 +1962,9 @@ struct controller_impl {
             } // end if exception
          } /// end for each block in branch
 
-         unapplied_branch = std::move( branches.second );
+         if ( read_mode == db_read_mode::SPECULATIVE ) {
+            unapplied_branch = std::move( branches.second );
+         }
```
heifner: This is not ideal in that we add forked branches after applying the new branch which might have many of the same transactions.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
```diff
@@ -65,17 +66,25 @@ class unapplied_transaction_queue {
    > unapplied_trx_queue_type;
 
    unapplied_trx_queue_type queue;
+   db_read_mode read_mode = db_read_mode::SPECULATIVE;
```
arhag: Avoid bringing in controller data structures (like `db_read_mode`) into `unapplied_transaction_queue`. Instead replace both `read_mode` and `only_track_persisted` with something that tracks the add mode for `unapplied_transaction_queue`. Then use a member function to set the add mode at the beginning based on information like the controller read mode and whether any producers are configured on the node.

Here is one option:

Redefine the `trx_enum_type` enum class as:
```cpp
enum class trx_enum_type : uint64_t {
   unknown   = 1,
   persisted = 2,
   forked    = 4,
   aborted   = 8 
};
```
Then add a field into `unapplied_transaction_queue`:
```cpp
    std::underlying_type_t<trx_enum_type> add_mode{};
```

Replace `set_only_track_persisted` and `set_chain_read_mode` with:
```cpp
template<typename... Es>
auto set_add_mode( Es... allowed_types ) -> std::underlying_type_t<trx_enum_type> {
    static_assert( (std::is_same_v<Es, trx_enum_type> && ...) );
    FC_ASSERT( empty(), "set_add_mode, queue required to be empty" );
    return create_mask( allowed_types... );
}

auto set_add_mode() {
    FC_ASSERT( empty(), "set_add_mode, queue required to be empty" );
    return create_mask<trx_enum_type>();
}
```

where the `create_mask` is a utility function added to types.hpp:
```cpp
template<typename E, typename... Es>
inline constexpr auto create_mask( E first, Es... rest )
-> std::enable_if_t<
      std::is_enum_v<E>
      && std::is_integral_v<std::underlying_type_t<E>> 
      && std::is_unsigned_v<std::underlying_type_t<E>>
      && (std::is_same_v<Es, E> && ...),
      std::underlying_type_t<E> 
   >
{
    using F = std::underlying_type_t<E>;
    return (static_cast<F>(first) | ... | static_cast<F>(rest));
}

template<typename E>
inline constexpr auto create_mask() 
-> std::enable_if_t<
      std::is_enum_v<E>
      && std::is_integral_v<std::underlying_type_t<E>> 
      && std::is_unsigned_v<std::underlying_type_t<E>>,
      std::underlying_type_t<E> 
   >
{
    return std::underlying_type_t<E>{};
}
```

Then you check for the appropriate flag set in the `add_mode` member variable in each of the add member functions. For example:
```cpp
void add_aborted( std::vector<transaction_metadata_ptr> aborted_trxs ) {
   if( aborted_trxs.empty() || !has_field( add_mode, trx_enum_type::aborted ) ) return;
   ...
}
```

libraries/chain/controller.cpp
```diff
@@ -1920,8 +1918,11 @@ struct controller_impl {
          auto branches = fork_db.fetch_branch_from( new_head->id, head->id );
 
          if( branches.second.size() > 0 ) {
-            for( auto itr = branches.second.begin(); itr != branches.second.end(); ++itr ) {
+            // forked branches are in reverse order of how they were applied, signal callback in the order applied
+            for( auto ritr = branches.second.rbegin(), rend = branches.second.rend(); ritr != rend; ++ritr ) {
+               const block_state_ptr& bsptr = *ritr;
                pop_block();
+               forked_trxs_cb( bsptr->trxs );
```
arhag: Call `forked_trxs_cb` on `branches.second` once, rather than calling the callback for each block.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
```diff
@@ -118,28 +127,24 @@ class unapplied_transaction_queue {
...
-      if( only_track_persisted ) return;
-      // forked_branch is in reverse order
-      for( auto ritr = forked_branch.rbegin(), rend = forked_branch.rend(); ritr != rend; ++ritr ) {
-         const block_state_ptr& bsptr = *ritr;
-         for( auto itr = bsptr->trxs.begin(), end = bsptr->trxs.end(); itr != end; ++itr ) {
-            const auto& trx = *itr;
-            fc::time_point expiry = trx->packed_trx()->expiration();
-            queue.insert( { trx, expiry, trx_enum_type::forked } );
-         }
+   void add_forked( const std::vector<transaction_metadata_ptr>& forked_trxs ) {
```
arhag: Change this back to taking the `const branch_type&`.

libraries/chain/controller.cpp
arhag: Also check to see if `forked_trxs_cb` converts to true (i.e contains valid target) before calling.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
arhag: After talking with @heifner: 

The intention is for `unapplied_transaction_queue` to be aware of the logic of when/how transactions get abandoned due to forks, aborts, etc., and to impose the policy for ordering these transactions. So the decision of which add functions do nothing under certain environments is to be made by `unapplied_transaction_queue`. It just needs to know what the relevant environment is to make its decisions.

I still think it isn't good to bring in ` db_read_mode` into this class however. So there can be a new `enum class` that describes the relevant environment and is fed into a single function that sets the mode (i.e. sets the value of member variable appropriately so that future add calls work as intended). As of now, we only need three cases to distinguish for the environment: `speculative_producer`, `speculative_non_producer`, `non_speculative`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -849,13 +846,17 @@ void producer_plugin::plugin_startup()
...
               "node cannot have any producer-name configured because block production is impossible when read_mode is not \"speculative\"" );
 
    EOS_ASSERT( my->_producers.empty() || chain.get_validation_mode() == chain::validation_mode::FULL, plugin_config_exception,
               "node cannot have any producer-name configured because block production is not safe when validation_mode is not \"full\"" );
 
+   unapplied_transaction_queue::process_mode unapplied_mode =
+         (chain.get_read_mode() != chain::db_read_mode::SPECULATIVE) ? unapplied_transaction_queue::process_mode::non_speculative :
+            my->_producers.empty() ? unapplied_transaction_queue::process_mode::speculative_non_producer :
+               unapplied_transaction_queue::process_mode::speculative_producer;
+   my->_unapplied_transactions.set_mode( unapplied_mode );
```
arhag: Do this in `producer_plugin::plugin_initiailize`. Since producer_plugin depends on chain_plugin, it is safe to get the read_mode from the controller at that point.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -406,13 +406,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-            auto size = calc_size( trx );
-            EOS_ASSERT( size_in_bytes + size < max_incoming_transaction_queue_size, tx_resource_exhaustion, "Transaction exceeded producer resource limit" );
-            size_in_bytes += size;
+            add_size( trx );
+            _incoming_transactions.emplace_back( trx, persist_until_expired, std::move( next ) );
+         }
+
+         void add_front( const transaction_metadata_ptr& trx, bool persist_until_expired, next_function<transaction_trace_ptr> next ) {
+            add_size( trx );
             _incoming_transactions.emplace_back( trx, persist_until_expired, std::move( next ) );
```
arhag: `emplace_front`.

heifner: Duh. That was the whole point.

PR #7689, opened 2019-07-26
WABT Lightweight callback

mitza-oci:
Make use of https://github.com/EOSIO/wabt/pull/8 (will need to update submodule here as well)


heifner:
Overcome by events.

PR #7690, opened 2019-07-27
Make mongoDB store genesis block

### LINKED ISSUE #7687: mongo_db_plugin doesn't store genesis block
mongo_db_plugin handles `accepted_block` and `applied_irreversible_block` events, and for genesis block, both events aren't emitted, so mongoDB doesn't store genesis block.

When MongoActionReader in demux-eos tries reading blocks from the beginning, because of missing block 1, its process is aborted. To keep consistency, (NodeosActionReader can read block 1 from history_api_plugin), mongoDB needs to store genesis block.

I found that block_state_ptr for genesis block isn't stored in fork_database, so `controller::fetch_block_state_by_[id/number]` don't work in this case. (signed_block_ptr for block 1 is stored in block_log, so it can be retrieved by `controller::fetch_block_by_number`)

I think that mongo_db_plugin needs to store genesis block in its `plugin_startup` step, and there are some possibilities to get genesis block_state_ptr.

1. Fresh started blockchain case, `controller::head_block_state()` will return genesis block_state_ptr.
2. Not fresh blockchain case, I could't find where to get genesis block_state_ptr, so extract build genesis block_state_ptr process from `controller_impl::initialize_blockchain_state()` and name it `controller::genesis_block_state()`. mongo_db_plugin calls it to retrieve genesis block_state_ptr. The one thing I concerned that block_state_ptr, returned by this method, isn't marked validated, because `fork_database::reset()` marks validated as true during `controller_impl::init()`.

Or, please give me any suggestion to solve this issue. Thank you.

conr2d:
## Change Description
Fixes #7687 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/mongo_db_plugin/mongo_db_plugin.cpp
```diff
@@ -1662,6 +1671,13 @@ void mongo_db_plugin::plugin_initialize(const variables_map& options)
 
 void mongo_db_plugin::plugin_startup()
 {
+   chain_plugin* chain_plug = app().find_plugin<chain_plugin>();
+   EOS_ASSERT( chain_plug, chain::missing_chain_plugin_exception, ""  );
+   auto& chain = chain_plug->chain();
+
+   auto bs = chain.genesis_block_state();
+   bs->validated = true; // consider genesis block validated
```
heifner: This should be moved to `genesis_block_state()` method.

heifner:
Thanks for this contribution. However, after internal discussion we have decided to not take this PR. For eosio based chains, block 1 is not really considered a valid block. This hack would provide a work-around for a bug in `demux-eos`. Instead please file a bug with `demux-eos`. Note block 1 does not contain any actions.

The history plugin is deprecated and the mongo plugin is also deprecated. The new state_history plugin does not provide block 1 and we do not intend to add the capability.

Issue #7575, opened 2019-07-01
nodeos 1.7.4 will not connect to nodeos 1.8.0 with allowed-connection = specified


matthewdarwin:
If you have a nodeos 1.7.4 server and try to connect to a 1.8.0 server with `allowed-connection = specified`, it will fail.  (1.8.0 to 1.7.4 also fails)

This is seen in the log:

```
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: info  2019-07-01T13:03:24.246 thread-0  net_plugin.cpp:1827           connect              ] host: 192.168.122.37 port: 9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: info  2019-07-01T13:03:24.246 thread-0  net_plugin.cpp:1827           connect              ] host: 192.168.122.131 port: 9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: info  2019-07-01T13:03:24.246 thread-0  net_plugin.cpp:1827           connect              ] host: 192.168.122.157 port: 9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: info  2019-07-01T13:03:24.247 thread-0  net_plugin.cpp:1827           connect              ] host: 192.168.122.175 port: 9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: error 2019-07-01T13:03:24.249 thread-0  net_plugin.cpp:1108           operator()           ] sent a go away message: authentication failure, closing connection to 192.168.122.175:9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: error 2019-07-01T13:03:24.249 thread-0  net_plugin.cpp:1108           operator()           ] sent a go away message: authentication failure, closing connection to 192.168.122.131:9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: error 2019-07-01T13:03:24.249 thread-0  net_plugin.cpp:1108           operator()           ] sent a go away message: authentication failure, closing connection to 192.168.122.37:9876
Jul  1 13:03:24 mainnet-peer8 nodeos[42247]: error 2019-07-01T13:03:24.250 thread-0  net_plugin.cpp:1108           operator()           ] sent a go away message: authentication failure, closing connection to 192.168.122.157:9876
```

The servers will connect to each other fine if `allowed-connections = any`
They also connect fine if they are all on 1.8 or all on 1.7.

taokayan:
[...]

PR #7691, opened 2019-07-27
Fix nodeos 1.8.x to > 1.7.x peering issue (allowed-connection not equal to "any")

rdewilder:
Fixes https://github.com/EOSIO/eos/issues/7575

## Change Description

Issue: #7575
The default std::chrono::duration resolution changed from nanoseconds in 1.7.x and below to microseconds in 1.8.x.  When allowed-connection not equal to "any" the time skew between peers needs to be less than 1 sec.  The time skew check will always fail due to the resolution mismatch.

Modifications:

added explicit cast of the current time to nanoseconds during the peering handshake intialization.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3160,11 +3160,12 @@ namespace eosio {
 
    // call from connection strand
    void connection::populate_handshake( handshake_message& hello ) {
+      namespace sc = std::chrono;
       hello.network_version = net_version_base + net_version;
       hello.chain_id = my_impl->chain_id;
       hello.node_id = my_impl->node_id;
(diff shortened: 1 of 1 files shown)


spoonincode:
> When allowed-connection = "any" the time skew between peers needs to be less than 1 sec

I think you mean when it's _not_ set to "any"? When set to any it seems to pass before getting to that check.
https://github.com/EOSIO/eos/blob/1418543149b7caf8fc69a23621e3db7f3c6d18ad/plugins/net_plugin/net_plugin.cpp#L2754-L2776

rdewilder:
My mistake.  I have updated the comment and description.

PR #7702, opened 2019-07-29
fix producer_plugin watermark tracking - develop

arhag:
## Change Description

The watermarks will not be allowed to decrease.

Update watermark marks as soon as block is accepted into fork database (meaning it passed header validation which means the block was correctly signed by the producer) rather than waiting until the block is applied.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1865,8 +1865,6 @@ void producer_plugin_impl::produce_block() {
 
    block_state_ptr new_bs = chain.head_block_state();
 
-   consider_new_watermark( new_bs->header.producer, new_bs->block_num );
-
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1856,7 +1864,8 @@ void producer_plugin_impl::produce_block() {
    chain.commit_block();
 
    block_state_ptr new_bs = chain.head_block_state();
-   _producer_watermarks[new_bs->header.producer] = chain.head_block_num();
+
+   consider_new_watermark( new_bs->header.producer, new_bs->block_num );
```
b1bart: it seems like `controller::commit_block()` will emit an `accepted_block_header` signal making this redundant

PR #7703, opened 2019-07-30
return flat_multimap from transaction::validate_and_extract_extensions

### LINKED ISSUE #7681: update `transaction::validate_and_extract_extensions` to return a `flat_multimap` to be consistent with the other type extensions
_Originally posted by @arhag in https://github.com/EOSIO/eos/pull/7404_

arhag:
## Change Description

Resolves #7681.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/transaction.cpp
+++ b/libraries/chain/transaction.cpp
@@ -144,14 +144,10 @@ fc::microseconds transaction::get_signature_keys( const vector<signature_type>& 
    return sig_cpu_usage;
 } FC_CAPTURE_AND_RETHROW() }
 
-vector<transaction_extensions> transaction::validate_and_extract_extensions()const {
-   using transaction_extensions_t = transaction_extension_types::transaction_extensions_t;
+flat_multimap<uint16_t, transaction_extensions> transaction::validate_and_extract_extensions()const {
    using decompose_t = transaction_extension_types::decompose_t;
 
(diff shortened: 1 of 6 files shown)


libraries/chain/transaction.cpp
```diff
@@ -163,9 +159,12 @@ vector<transaction_extensions> transaction::validate_and_extract_extensions()con
...
       );
 
-      results.emplace_back();
+      auto iter = results.emplace(std::piecewise_construct,
+         std::forward_as_tuple(id),
+         std::forward_as_tuple()
+      );
 
-      auto match = decompose_t::extract<transaction_extensions_t>( id, e.second, results.back() );
+      auto match = decompose_t::extract<eosio::chain::transaction_extensions>( id, e.second, iter->second );
```
b1bart: I am not a fan of the 3x different "standards" for this typename:

`eosio::chain::transaction_extensions` , presumably fully-qualified because it collides with the member `transaction::transaction_extensions`

`block_header_extensions`, which isn't fully qualified because its associated member is `block_header::header_extensions`

`block_extension` (not plural), which isn't fully qualified because its associated member is `signed_block::block_extensions` (plural)

Is there a good way to create unity across these?  I suppose as the member names are API-relevant we'd be talking about unifying the typenames

PR #7706, opened 2019-07-30
[Travis POC] EOSIO pipeline parity

NorseGaud:
- Created all platform build and test steps with buildkite-agent upload
- Created packaging and submodule step with buildkite-agent download
- Reorganized folder structure to cut down on complexity (we can further do this by creating the buildkiite directory, but that requires a pipeline change)
- Moved Travis trigger to after base images
- Support for folding and printing each command in BOTH Travis and Buildkite added
- Support for disabling/enabling various tests:
```
${ENABLE_PARALLEL_TESTS:-true}
${ENABLE_SERIAL_TESTS:-true}
${ENABLE_LR_TESTS:-false}
```
- Test Metrics ( https://github.com/EOSIO/auto-gcp-buildfleet/pull/137 ) passing from https://buildkite.com/EOSIO/eosio-beta/builds/431

https://buildkite.com/EOSIO/eosio-beta/builds/528
https://travis-ci.org/EOSIO/eos/builds/565970798


.cicd/.helpers
```diff
@@ -42,8 +39,9 @@ function docker_tag_exists() {
     ( [[ $EXISTS =~ '404 page not found' ]] || [[ $EXISTS =~ 'manifest unknown' ]] ) && return 1 || return 0
 }
 
-if [[ $(uname) != 'Darwin' ]]; then # Darwin doesn't need hash (yet)
-    determine-hash ".cicd/${IMAGE_TAG}.dockerfile"
+# If you're on mac and you want to generate the image, use "force" as $2
```
kj4ezj: What does this mean?   
Running the following code....
```

File: .cicd/.helpers [hunk 2]
```diff
@@ -3,12 +3,11 @@ VERBOSE=${VERBOSE:-true}
 export PROJECT_NAME="eosio"
 
 # Set IMAGE_TAG using arg (generate-base-images.sh) or env (travis.yml)
-( [[ $(uname) != 'Darwin' ]] && [[ -z $IMAGE_TAG ]] && [[ -z $1 ]] ) && echo "You must provide the distro IMAGE_TAG name (example: ubuntu-18.04) as argument \$1 or set it within your ENV" && exit 1
 export IMAGE_TAG=${IMAGE_TAG:-$1}
 
 function execute() {
```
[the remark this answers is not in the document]
NorseGaud: Half way agree. `--- Executing:` is needed for buildkite folding in logs.

.cicd/.helpers
kj4ezj: What does this mean?
Running the following code on your branch...
```
.cicd$ grep -r '$2' .
Binary file ./metrics/test-metrics.tar.gz matches
./.helpers:# If you're on mac and you want to generate the image, use "force" as $2
```
...shows that $2 is not consumed anywhere. If it is, we should automate this using:
```
[[ $(uname) != 'Darwin' ]]
```
But we are not building base images on mac, so I don't think we need any of this.

.cicd/docker/.helpers-v33
```diff
@@ -0,0 +1,8 @@
+function fold-execute() {
```
kj4ezj: We don't need the code folding. This is unnecessary complexity. This can be the same as above, and should not be duplicated here IMO:
```
function execute()
{
    echo "\$ $@"
    "$@"
}
```

.cicd/docker/amazonlinux-2.dockerfile
```diff
@@ -72,11 +72,23 @@ ENV PATH=${PATH}:/mongodb-linux-x86_64-amazon-3.6.3/bin
...
-ENV PRE_COMMANDS="export PATH=/usr/lib64/ccache:$PATH &&"
+ENV PRE_COMMANDS="export PATH=/usr/lib64/ccache:\$PATH"
 ENV CMAKE_EXTRAS="-DCMAKE_CXX_COMPILER='clang++' -DCMAKE_C_COMPILER='clang'"
 
-CMD bash -c "$PRE_COMMANDS ccache -s && \
-    mkdir /workdir/build && cd /workdir/build && cmake -DCMAKE_BUILD_TYPE='Release' -DCORE_SYMBOL_NAME='SYS' -DOPENSSL_ROOT_DIR='/usr/include/openssl' -DBUILD_MONGO_DB_PLUGIN=true $CMAKE_EXTRAS /workdir && make -j $(getconf _NPROCESSORS_ONLN) && \
-    ctest -j$(getconf _NPROCESSORS_ONLN) -LE _tests --output-on-failure -T Test"
\ No newline at end of file
+# Bring in helpers that provides execute function so we can get better logging in BK and TRAV
+COPY ./docker/.helpers-v33 /tmp/.helpers
```
kj4ezj: Do we need multiple helper scripts?
Can they be named more clearly?

.cicd/docker/ubuntu-16.04.dockerfile
```diff
@@ -5,7 +5,7 @@ RUN apt-get update && apt-get upgrade -y \
   && DEBIAN_FRONTEND=noninteractive apt-get install -y build-essential git automake \
   libbz2-dev libssl-dev doxygen graphviz libgmp3-dev autotools-dev libicu-dev \
   python2.7 python2.7-dev python3 python3-dev autoconf libtool curl zlib1g-dev \
-  sudo ruby libusb-1.0-0-dev libcurl4-gnutls-dev pkg-config
+  sudo ruby libusb-1.0-0-dev libcurl4-gnutls-dev pkg-config apt-transport-https
```
kj4ezj: Is the new dependency for `buildkite-agent`?

.cicd/package-builder.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+set -eo pipefail
+cd $( dirname "${BASH_SOURCE[0]}" ) # Ensure we're in the .cicd dir
+. ./.helpers
```
kj4ezj: Where is this library used in this script? Seems to me like it doesn't need to be included here.

.cicd/pipeline.yml
```diff
@@ -3,78 +3,100 @@ env:
   TEST_TIMEOUT: 60
   TIMEOUT: 120
   VERBOSE: true
+  DEBUG: true
```
kj4ezj: This whole block of variables is unnecessary complexity.

.cicd/pipeline.yml
```diff
@@ -3,78 +3,100 @@ env:
...
 
 steps:
   - wait
 
   - label: ":webhook: Trigger Travis CI Build"
-    command: |
-      . ./.cicd/trigger-travis.sh
+    command:
+      - "./.cicd/travis/travis-trigger.sh"
```
kj4ezj: Why was the format of this command statement changed?   
Why was the script renamed?

.cicd/pipeline.yml
```diff
@@ -3,78 +3,100 @@ env:
...
 
-  - label: ":aws: Amazon_Linux 2 - Build"
+  - label: ":aws: Amazon_Linux 2 - Build & Test"
     agents:
       queue: "automation-eos-builder-fleet"
     plugins:
       - docker#v3.2.0:
           debug: $DEBUG
-          image: "eosio/producer:eosio-amazonlinux-2-1c266e25276ad1f1147a4df5c12921968079c49b"
+          image: "eosio/producer:eosio-amazonlinux-2-cc8d19ea479663d98189ce2473aacc2a7567a8be"
```
kj4ezj: Can this hash be automated?

.cicd/pipeline.yml
```diff
@@ -3,78 +3,100 @@ env:
...
+          environment:
+            - "BUILDKITE_AGENT_ACCESS_TOKEN" # Needed for buildkite-agent upload/download
+          mount-buildkite-agent: false # Mounting bk-agent doesn't work, so disable it
+          propagate-environment: true # Need for buildkite-agent upload/download (JOB_ID, etc)
     timeout: $BUILD_TIMEOUT
     skip: $SKIP_AMAZON_LINUX_2
 
-  - label: ":centos: CentOS 7.6 - Build"
+  - label: ":centos: CentOS 7 - Build & Test"
```
kj4ezj: This does not follow the label guidelines established [here](https://github.com/EOSIO/eos-buildkite-pipelines/pull/27).

.cicd/pipeline.yml
```diff
@@ -3,78 +3,100 @@ env:
...
     agents:
       queue: "automation-eos-builder-fleet"
     plugins:
       - docker#v3.2.0:
-          debug: $DEBUG
-          image: "eosio/producer:eosio-centos-7-d45789dfb1a5b830a62381afd205b34d3d7c2d63"
+          debug: true
+          image: "eosio/producer:eosio-centos-7-61150e3338126b91bd9384821be93de5e9f8b3fe"
+          environment:
+            # - "ENABLE_LR_TESTS=true"
```
kj4ezj: Remove commented-out code.

.cicd/pipeline.yml
```diff
@@ -3,78 +3,100 @@ env:
...
     command:
-      - "brew install git graphviz libtool gmp llvm@4 pkgconfig python python@2 doxygen libusb openssl boost@1.70 cmake"
+      - "brew install git graphviz libtool gmp llvm@4 pkgconfig python python@2 doxygen libusb openssl boost@1.70 cmake mongodb"
       - "git clone $BUILDKITE_REPO eos && cd eos && git checkout $BUILDKITE_COMMIT && git submodule update --init --recursive"
-      - "cd eos && mkdir build && cd build && cmake .. && make -j$(getconf _NPROCESSORS_ONLN)"
-      - "cd eos/build && ctest -j$(getconf _NPROCESSORS_ONLN) -LE _tests --output-on-failure -T Test"
+      - "cd eos && mkdir -p build && cd build && cmake .."
+      - "cd eos/build && make -j$(getconf _NPROCESSORS_ONLN)"
+      - "if ${ENABLE_PARALLEL_TESTS:-true}; then cd eos/build && ctest -j$(getconf _NPROCESSORS_ONLN) -LE _tests --output-on-failure -T Test; fi"
+      - "if ${ENABLE_SERIAL_TESTS:-true}; then cd eos/build && mkdir -p ./mongodb && mongod --dbpath ./mongodb --fork --logpath mongod.log && ctest -L nonparallelizable_tests --output-on-failure -T Test; fi"
```
kj4ezj: We should discuss serial tests as a group.

.cicd/pipeline.yml
```diff
@@ -84,4 +106,93 @@ steps:
       - "queue=mac-anka-large-node-fleet"
     timeout: $BUILD_TIMEOUT
     skip: $SKIP_MOJAVE
-    
\ No newline at end of file
+
+  - wait:
+    continue_on_failure: true
+
+  # - command: |
```
kj4ezj: Test metrics should be enabled.
Otherwise, delete commented-out code.

.cicd/pipeline.yml
```diff
@@ -84,4 +106,93 @@ steps:
...
+  #     node --max-old-space-size=32768 test-metrics.js
+  #   label: ":bar_chart: Test Metrics"
+  #   agents:
+  #     queue: "automation-eos-builder-fleet"
+  #   timeout: 10
+  #   soft_fail: true
+
+  # - wait
+
+  - label: ":centos: Centos 7 - PB"
```
kj4ezj: This does not follow the label guidelines established [here](https://github.com/EOSIO/eos-buildkite-pipelines/pull/27).

.cicd/docker/.helpers-v33
NorseGaud: I do; It makes troubleshooting failures way easier.

.cicd/docker/amazonlinux-2.dockerfile
NorseGaud: Agreed, changing the name (I forgot to do this before the PR)

.cicd/docker/ubuntu-16.04.dockerfile
NorseGaud: Yep. But it won't install when it's in the buildkite install section (I'm not sure why; it gets completely ignored even with a pruned environment/no cache)

.cicd/package-builder.sh
NorseGaud: Good catch, you're right

.cicd/pipeline.yml
NorseGaud: Should we fix the dockerfile to use FROM: centos:7.6.1810?

.cicd/pipeline.yml
NorseGaud: Agreed, removing

.cicd/pipeline.yml
NorseGaud: We've had this discussion before I believe; `command: |` is only for supporting porting over from travis and not needed for a single command.

.cicd/pipeline.yml
NorseGaud: That's the goal

.cicd/pipeline.yml
NorseGaud: It's there to show how it works/demo. I can remove it later

.cicd/pipeline.yml
NorseGaud: Also, why isn't this a problem for amazonlinux?

NorseGaud:
> Is `./.cicd/metrics/test-metrics.tar.gz` required in the repository now?

Nope, removing

.cicd/pipeline.yml
kj4ezj: The script name `travis-trigger.sh` is inconsistent with the rest of the repos, which are `trigger-travis.sh`. I think we should maintain consistency in all repos.

.cicd/pipeline.yml
kj4ezj: This was not addressed.

.cicd/pipeline.yml
kj4ezj: Remove commented-out code before merge.
This is the point of version control, [that code](https://github.com/EOSIO/eos/blob/3b21fdd536aa9057f98d4bfaac594dfa079735fd/.cicd/pipeline.yml#L42) is never gone.

kj4ezj:
> > Is `./.cicd/metrics/test-metrics.tar.gz` required in the repository now?
> 
> Nope, removing

We will need the test metrics code if we intend to run test metrics. We shouldn't commit >50 node files to the repo, so we'll need the `test-metrics.tar.gz`.

NorseGaud:
> > > Is `./.cicd/metrics/test-metrics.tar.gz` required in the repository now?
> > 
> > 
> > Nope, removing
> 
> We will need the test metrics code if we intend to run test metrics. We shouldn't commit >50 node files to the repo, so we'll need the `test-metrics.tar.gz`.

I added it back after Scott and I's comments 👍

.cicd/pipeline.yml
NorseGaud: What about travis-build.sh? Is that not a naming convention? Why would we not change the other repos instead?

kj4ezj: > Is that not a naming convention?
  
Not as far as I am aware.
   
EOSIO, contracts, and eos-vm all have `trigger-travis.sh`. I don't see why we should deviate. The `.cicd` folder shouldn't be complicated.

NorseGaud: I changed it back to get this unblocked.

PR #7707, opened 2019-07-30
add softfloat only injection mode

spoonincode:
## Change Description
refactor the injection code to offer a softfloat only injection mode along with its classical "full" injection that includes checktime & depth checking. This new softfloat only injection is useful for potential new wasm runtimes that don't need checktime/depth validation

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
@@ -677,9 +677,6 @@ namespace eosio { namespace chain { namespace wasm_injections {
    };
 
    struct pre_op_injectors : wasm_ops::op_types<pass_injector> {
-      using call_t            = wasm_ops::call                    <call_depth_check_and_insert_checktime>;
-      using call_indirect_t   = wasm_ops::call_indirect           <call_depth_check_and_insert_checktime>;
-      
       // float binops 
       using f32_add_t         = wasm_ops::f32_add                 <f32_binop_injector<wasm_ops::f32_add_code>>;
       using f32_sub_t         = wasm_ops::f32_sub                 <f32_binop_injector<wasm_ops::f32_sub_code>>;
@@ -752,10 +749,17 @@ namespace eosio { namespace chain { namespace wasm_injections {
       using f64_convert_u_i64 = wasm_ops::f64_convert_u_i64       <i64_convert_f64_injector<wasm_ops::f64_convert_u_i64_code>>;
    }; // pre_op_injectors
 
+   struct pre_op_full_injectors : pre_op_injectors {
+      using call_t            = wasm_ops::call                    <call_depth_check_and_insert_checktime>;
+      using call_indirect_t   = wasm_ops::call_indirect           <call_depth_check_and_insert_checktime>;
+   };
 
    struct post_op_injectors : wasm_ops::op_types<pass_injector> {
-      using loop_t   = wasm_ops::loop        <checktime_injection>;
       using call_t   = wasm_ops::call        <fix_call_index>;
+   };
+
+   struct post_op_full_injectors : post_op_injectors {
+      using loop_t   = wasm_ops::loop        <checktime_injection>;
    };
 
    template <typename ... Visitors>
@@ -773,7 +777,9 @@ namespace eosio { namespace chain { namespace wasm_injections {
       }
    };
  
-   // inherit from this class and define your own injectors 
+   // "full injection" gives checktime & depth counting along with softfloat injection.
+   // Otherwise you'll just get softfloat injection
+   template<bool full_injection>
    class wasm_binary_injection {
       using standard_module_injectors = module_injectors< max_memory_injection_visitor >;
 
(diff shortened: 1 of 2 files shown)


libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
```diff
@@ -752,12 +749,19 @@ namespace eosio { namespace chain { namespace wasm_injections {
...
 
    struct post_op_injectors : wasm_ops::op_types<pass_injector> {
-      using loop_t   = wasm_ops::loop        <checktime_injection>;
       using call_t   = wasm_ops::call        <fix_call_index>;
    };
 
+   struct post_op_full_injectors : post_op_injectors {
+      using loop_t   = wasm_ops::loop        <checktime_injection>;
+   };
+
```
swatanabe-b1: Is there a reason why loop checktime injection is in the post_op_injectors instead of the pre-op injectors?  It would be simpler if you only needed to switch one set of injectors.

libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
```diff
@@ -789,10 +795,11 @@ namespace eosio { namespace chain { namespace wasm_injections {
          void inject() {
             _module_injectors.inject( *_module );
             // inject checktime first
-            injector_utils::add_import<ResultType::none>( *_module, u8"checktime", checktime_injection::chktm_idx );
+            if constexpr (full_injection)
+               injector_utils::add_import<ResultType::none>( *_module, u8"checktime", checktime_injection::chktm_idx );
```
swatanabe-b1: Why does checktime need to be the first import?  It seems like a pretty arbitrary requirement imposed by checktime_injection, which is inconsistent with the other injected intrinsics.

libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
```diff
@@ -773,7 +777,9 @@ namespace eosio { namespace chain { namespace wasm_injections {
       }
    };
  
-   // inherit from this class and define your own injectors 
+   // "full injection" gives checktime & depth counting along with softfloat injection.
+   // Otherwise you'll just get softfloat injection
+   template<bool full_injection>
```
swatanabe-b1: It would be cleaner and more general to pass the injector(s) as template parameters.

libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
spoonincode: @larryk85

libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
spoonincode: I'm struggling to remember why; maybe @larryk85 knows.

spoonincode:
> "potential new wasm runtimes" is quite vague.

yep for sure. The idea here is to carve out pieces from the `rodeos_wip` branch that can be PRed & reviewed standalone so that there isn't a single massive code drop of months of work. Since we're averaging weeks to complete reviews on PRs it's worthwhile to get these up early. But of course it's still not clear if/when `rodeos_wip` will be merged.

libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
spoonincode: I thought about that but I still need to do something like
```
               if constexpr (full_injection) {
	                  wasm_ops::op_types<>::call_t chktm;
                          chktm.field = injector_utils::injected_index_mapping.find(checktime_injection::chktm_idx)->second;
                          chktm.pack(&post_code);
               }
```

swatanabe-b1: This could be moved into pre/post_injector as a function prefix (which defaults to a no-op).  That would also allow the two loops in inject to be unduplicated.  Also, while we're on this topic, why is it necessary to insert checktime both at the beginning of every function and before each function call.

spoonincode: If the concern is that a boolean is not clear enough here, would an enum be better?

> Also, while we're on this topic, why is it necessary to insert checktime both at the beginning of every function and before each function call.

I can elaborate on this in private.

swatanabe-b1: It's not just the boolean.  It's more that there isn't a clean separation of responsibilities and that it's almost generic, but not quite.

libraries/chain/include/eosio/chain/wasm_eosio_injection.hpp
larryk85: It was made that way so that we could ensure that we knew exactly which function was checktime without having to parse much by external tools, i.e. batch dump wasms and see how many checktimes where called (it is always 0).  This was also done to allow for checktimes to be added around other injected functions without having to keep any real state of the checktime function.  None the less, it doesn't have to do it that way, it was just a bit of a micro optimization at the time.

PR #7717, opened 2019-08-01
Support for v2 snapshots with pending producer schedules

### LINKED ISSUE #7715: Support importing v2 snapshot even when global_property_object::proposed_schedule is not empty
PR #7404 introduced the `WTMSIG_BLOCK_SIGNATURES` protocol feature and in doing so required changes to the `global_property_object`. The intention of the changes in that PR were to still successfully import old serialized `global_property_object`s in v2 snapshots and upconvert as necessary to store as the new `global_property_object` in the state database. However, the code path to handle the legacy version in v2 snapshots does not actually exist. The reason it is still possible to import v2 most of the time (and indeed possible in the tests that explicitly test this) is due to the fact that the serialization of `global_property_object` in the new and old versions are the same if `proposed_schedule.producers` is the empty vector.

Code to explicitly add support for importing in the old `global_property_object` from v2 snapshots needs to be added to `controller_impl::read_from_snapshot` which will upconvert `legacy:: producer_schedule_type` to `producer_authority_schedule` (similar to how it is done already for the `block_state`). Furthermore, a test should be added (or existing one modified) to verify that import of a v2 snapshot is successful if the `proposed_schedule.producers` is not empty.

b1bart:
## Change Description

this resolves #7715 

as part of #7404 , the contents of `global_property_object` was changed such that it stored a pending producer schedule composed of `producer_authority` structures instead of `legacy::producer_key` structures.  The intent of this PR was to be backwards compatible with snapshots at version `2` and efforts were made in other structures to support that goal however, `global_property_object` was accidentally omitted from special handling.

As a result, snapshots of version `2` taken while there was a pending producer schedule change were **not** forwards compatible.

This PR addresses that omission.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


unittests/snapshot_tests.cpp
```diff
@@ -384,4 +385,60 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_compatible_versions, SNAPSHOT_SUITE, snapshot
...
+
+   BOOST_REQUIRE_EQUAL(gpo.proposed_schedule.version, 1);
+   BOOST_REQUIRE_EQUAL(gpo.proposed_schedule.producers.size(), 1);
+   BOOST_REQUIRE_EQUAL(gpo.proposed_schedule.producers[0].producer_name.to_string(), "snapshot");
+
+   static_assert(chain_snapshot_header::minimum_compatible_version <= 2, "version 2 unit test is no longer needed.  Please clean up data files");
+   auto v2 = SNAPSHOT_SUITE::template load_from_file<snapshots::snap_v2_prod_sched>();
+   snapshotted_tester v2_tester(chain.get_config(), SNAPSHOT_SUITE::get_reader(v2), 0);
+   auto v2_integrity_value = v2_tester.control->calculate_integrity_hash();
+   if (std::is_same_v<SNAPSHOT_SUITE, variant_snapshot_suite> && v2_integrity_value.str() != base_integrity_value.str()) {
```
b1bart: This conditional means that we are not calling `BOOST_REQUIRE_EQUAL(v2_integrity_value.str(), base_integrity_value.str());` on the binary form of the snapshots below and therefore not actually validating that the binary snapshot is loadable with the same integrity value.

unittests/snapshot_tests.cpp
```diff
@@ -384,4 +385,60 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_compatible_versions, SNAPSHOT_SUITE, snapshot
...
+      // create a latest snapshot
+      auto latest_writer = SNAPSHOT_SUITE::get_writer();
+      chain.control->write_snapshot(latest_writer);
+      auto chain_latest = SNAPSHOT_SUITE::finalize(latest_writer);
+      auto latest_writer2 = SNAPSHOT_SUITE::get_writer();
+      v2_tester.control->write_snapshot(latest_writer2);
+      auto v2_tester_latest = SNAPSHOT_SUITE::finalize(latest_writer2);
+      print_variant_diff(chain_latest, v2_tester_latest);
+      BOOST_REQUIRE_EQUAL(v2_integrity_value.str(), base_integrity_value.str());
```
b1bart: in the similar test `test_compatible_versions` we also validate a round-trip for the "v2" snapshot in that it can be loaded into controller, form it a new latest-version snapshot can be created and when that snapshot is loaded it still produces the same integrity value.   I'd like to see that round-trip here.

arhag: I'll echo what @b1bart said here. I would also like to see the `v2_tester_latest` snapshot loaded and then compare the integrity hash of that to `v2_integrity_value`.

b1bart: @arhag I pushed up the change to make that happen.  Do you mind reviewing the whole of the PR and rendering a call?

PR #7721, opened 2019-08-02
Improve signature recovery

heifner:
## Change Description

- Improve the performance of signature recovery by reverting back to 1.8.x method of thread_pool usage. The `develop` branch approach of passing a `next` function does not perform as well as just calling `wait()` on the future.
- Transaction id is now cached on `packed_transaction` instead of `transaction_metadata`.
- Changed the interface for `accept_transaction` to take a `packed_transaction` instead of `transaction_metadata`.
- `transaction_metadata` is now constructed with the recovered public keys and signature cpu usage.
- `transaction_metadata` is now constructed with `implicit` and `scheduled` const values set.
- `transaction_metadata` is now immutable except for the `accepted` flag.
- `transaction_metadata` construction is now restricted to static factory methods `start_recover_keys` and `create_no_recover_keys`.
- Use cached `transaction_metadata` in `apply_block` if a previously applied block with cached `transaction_metadata`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ x ] API Changes (plugin interface)
- `eosio::chain::plugin_interface::incoming::channels::transaction` changed from `transaction_metadata_ptr` to `packed_transaction_ptr`
- `eosio::chain::plugin_interface::incoming::methods::transaction_async` changed from `void(const transaction_metadata_ptr&, bool, next_function<transaction_trace_ptr>), first_provider_policy>` to `void(const packed_transaction_ptr&, bool, next_function<transaction_trace_ptr>), first_provider_policy>`

## Documentation Additions
- [ ] Documentation Additions


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -406,20 +406,23 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+            if( future.valid() ) {
+               future.wait();
+               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}]() mutable {
+                  try {
+                     self->process_incoming_transaction_async( future.get(), persist_until_expired, std::move( next ) );
+                  } CATCH_AND_CALL(next);
+               } );
+            }
+         });
       }
```
heifner: This is the key performance change. As odd as this looks, it is demonstratively better than before. And it is not the switch back to producer thread pool. Either thread pool (chain or producer) showed same results. The difference is in passing in a next `std::function` over calling `future.wait()`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -406,20 +406,23 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-      void on_incoming_transaction_async(const transaction_metadata_ptr& trx, bool persist_until_expired, next_function<transaction_trace_ptr> next) {
+      void on_incoming_transaction_async(const packed_transaction_ptr& trx, bool persist_until_expired, next_function<transaction_trace_ptr> next) {
          chain::controller& chain = chain_plug->chain();
          const auto max_trx_time_ms = _max_transaction_time_ms.load();
          fc::microseconds max_trx_cpu_usage = max_trx_time_ms < 0 ? fc::microseconds::maximum() : fc::milliseconds( max_trx_time_ms );
 
-         auto after_sig_recovery =
-               [self = this, trx, persist_until_expired, next{std::move(next)}]() mutable {
-                  app().post(priority::low, [self, trx{std::move(trx)}, persist_until_expired, next{std::move(next)}]() {
-                     self->process_incoming_transaction_async( trx, persist_until_expired, next );
```
swatanabe-b1: I don't know whether this matters, but next is being copied here.

heifner: Yes it was. I tested a version where the lambda was mutable and it `std::move` but didn't make any difference.

https://github.com/EOSIO/eos/pull/7721/commits/b734e56e41b2875eb48e491feef6af3a6e8ea9e5

libraries/chain/fork_database.cpp
```diff
@@ -128,7 +128,7 @@ namespace eosio { namespace chain {
                for( const auto& receipt : s.block->transactions ) {
```
arhag: We should probably skip this whole loop and temporarily remove the `trxs` field because we don't actually use it anyway. We really need a separate PR to more sensibly handle caching transaction metadatas that are created as part of applying blocks. Currently the fork database will only cache `transaction_metadata`s for blocks produced by a local producer.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1900,13 +1896,12 @@ void read_write::push_transaction(const read_write::push_transaction_params& par
    try {
       auto pretty_input = std::make_shared<packed_transaction>();
       auto resolver = make_resolver(this, abi_serializer_max_time);
-      transaction_metadata_ptr ptrx;
       try {
          abi_serializer::from_variant(params, *pretty_input, resolver, abi_serializer_max_time);
-         ptrx = std::make_shared<transaction_metadata>( pretty_input, db.configured_subjective_signature_length_limit() );
```
arhag: Before the `check_variable_sig_size` check was done here and it would throw the `packed_transaction_type_exception` exception. Now that check is delayed until key recovery and I'm not sure how the exception will be handled.

Same goes for `send_transaction`.

It isn't clear whether this is a problem. It depends on what the point of that validation was and when it is expected to be done.

@spoonincode Can I get your input on this?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2801,20 +2800,20 @@ namespace eosio {
          return;
       }
 
-      auto ptrx = std::make_shared<transaction_metadata>( trx, my_impl->chain_plug->chain().configured_subjective_signature_length_limit() );
```
arhag: Before the `check_variable_sig_size` check was done here and it would throw an exception out of this function. Now that check is delayed until key recovery and I'm not sure how the exception will be handled.

It isn't clear whether this is actually a problem. It depends on what the point of that validation was and when it is expected to be done.

unittests/misc_tests.cpp
```diff
@@ -832,47 +832,37 @@ BOOST_AUTO_TEST_CASE(transaction_metadata_test) { try {
...
+      BOOST_CHECK_EQUAL(trx.id(), ptrx2->id());
 
       named_thread_pool thread_pool( "misc", 5 );
 
-      transaction_metadata::start_recover_keys( mtrx, thread_pool.get_executor(), test.control->get_chain_id(), fc::microseconds::maximum() );
-      transaction_metadata::start_recover_keys( mtrx2, thread_pool.get_executor(), test.control->get_chain_id(), fc::microseconds::maximum() );
+      auto fut = transaction_metadata::start_recover_keys( ptrx, thread_pool.get_executor(), test.control->get_chain_id(), fc::microseconds::maximum() );
+      auto fut2 = transaction_metadata::start_recover_keys( ptrx2, thread_pool.get_executor(), test.control->get_chain_id(), fc::microseconds::maximum() );
 
       // no-op
```
arhag: These wouldn't be no-ops anymore right? It would actually be starting another key recovery process now.

plugins/chain_plugin/chain_plugin.cpp
spoonincode: Not `packed_transaction_type_exception`, but rather `sig_variable_size_limit_exception`. It's a subjective check, and a subjective exception. I haven't studied this change enough to know the ramifications for moving this (does it keep the subjectivity?)

unittests/misc_tests.cpp
heifner: Correct, so no-op in terms of nothing bad happens. I'll change it to:
`// start another key recovery on same transaction_metadata`

plugins/chain_plugin/chain_plugin.cpp
heifner: It is still subjective, but it is now done in `producer_plugin::on_incoming_transaction_async` where the transaction_metadata is created via a future.

plugins/net_plugin/net_plugin.cpp
heifner: The check is now in `producer_plugin::on_incoming_transaction_async` where the transaction_metadata is created via a future. The `future.get()` here https://github.com/EOSIO/eos/pull/7721/files#diff-a2a21b62383e9cc5c5349676661db760R422 will pass the exception on to `next()` which will log it. Before the exception would have been handled in `net_plugin::process_next_message` which would have `close()` the connection which was probably not what was intended.

plugins/chain_plugin/chain_plugin.cpp
spoonincode: Yeah I think it's okay -- the check still occurs before the actual recovery starts, so the purpose of the check is still clutched in

libraries/chain/fork_database.cpp
heifner: I'm going to go ahead and clean this up on this PR.

heifner: Reused cached `transaction_metadata` if available and created `completed_block` with existing `block_state_ptr` in `apply_block`. This also has the benefit of not having to create a new `block_state`.

heifner:
@arhag ready for review.

libraries/chain/controller.cpp
```diff
@@ -1799,18 +1802,9 @@ struct controller_impl {
...
-                        []( block_timestamp_type timestamp,
-                            const flat_set<digest_type>& cur_features,
-                            const vector<digest_type>& new_features )
-                        {}, // validation of any new protocol features should have already occurred prior to apply_block
-                        true // signature should have already been verified (assuming untrusted) prior to apply_block
-                    );
-
+         // create completed block with existing block_state along with newly created transaction_metadata as
+         // the newly created may contain recovered keys
+         bsp->trxs = std::move( ab._trx_metas );
```
arhag:
```cpp
if( !existing_trxs_metas && !skip_auth_checks ) {
   bsp->trxs = std::move( ab._trx_metas );
}
```

Only modify the `trxs` field if the `transaction_metadata`s have the keys recovered from the signatures.

Unfortunately, this would create the same potential thread safety issues as mutating `bsp->trxs` did above. Currently, no plugins actually use the `trxs` field. Maybe we should document that it should not be used by plugins since it is not thread safe? Or come up with a better solution. Although I really prefer not having to construct a new `block_state` here as was done before because removing it really cleans up the code nicely.

libraries/chain/controller.cpp
```diff
@@ -1730,10 +1730,13 @@ struct controller_impl {
          auto producer_block_id = b->id();
          start_block( b->timestamp, b->confirmed, new_protocol_feature_activations, s, producer_block_id);
 
+         const bool existing_trxs_metas = !bsp->trxs.empty();
          std::vector<recover_keys_future> trx_futures;
          std::vector<transaction_metadata_ptr> trx_metas;
          const bool skip_auth_checks = self.skip_auth_check();
-         if( skip_auth_checks ) {
+         if( existing_trxs_metas ) {
+            trx_metas = std::move( bsp->trxs );
```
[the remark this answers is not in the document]
heifner: Made the mutable non-thread-safe attributes private and added a comment indicating they are for internal use.

libraries/chain/controller.cpp
```diff
@@ -1799,18 +1802,9 @@ struct controller_impl {
          EOS_ASSERT( producer_block_id == ab._id, block_validate_exception, "Block ID does not match",
                      ("producer_block_id",producer_block_id)("validator_block_id",ab._id) );
 
-         auto bsp = std::make_shared<block_state>(
```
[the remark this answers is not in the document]
heifner: Removed.

libraries/chain/controller.cpp
```diff
@@ -1759,7 +1762,7 @@ struct controller_impl {
             const auto& trx_receipts = pending->_block_stage.get<building_block>()._pending_trx_receipts;
             auto num_pending_receipts = trx_receipts.size();
             if( receipt.trx.contains<packed_transaction>() ) {
-               if( skip_auth_checks ) {
+               if( skip_auth_checks || existing_trxs_metas ) {
```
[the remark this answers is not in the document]
heifner: Not really convinced that is clearer, but put it in.

libraries/chain/controller.cpp
heifner: I first added an indication on `block_state` of the state of the `cached_trxs` but then realized it was really only used in `apply_block` and I think what I pushed is safer as the indicator in `block_state` can't be out of sync with `cached_trxs` now.

heifner: Added tracking of public keys were recovered to `block_state` so we do not rely on every `transaction` having a signature.

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -46,13 +46,24 @@ namespace eosio { namespace chain {
       friend struct pending_state;
 
       bool is_valid()const { return validated; }
+      bool is_pub_keys_recovered()const { return _pub_keys_recovered; }
 
+      vector<transaction_metadata_ptr> extract_trxs_metas() {
+         _pub_keys_recovered = false;
+         return std::move( _cached_trxs );
```
arhag:
```cpp
_pub_keys_recovered = false;
vector<transaction_metadata_ptr> result( std::move( _cached_trxs ) );
_cached_trxs.clear();
return result;
```

Issue #6979, opened 2019-03-21
Several locations swallow std::bad_alloc


tbfleming:
If a plugin throws `std::bad_alloc` from a handler, `controller_impl::emit` will report `signal handler threw exception` and keep running. Many macros throughout the codebase, including fc, also swallow this exception.

Possibly related to #6377.

aclark-b1:
In order to focus our efforts on issues that are currently creating difficulty for the community we are closing tickets that were created prior to the EOSIO 2.0 release.  If you believe this issue is still relevant please feel free to reopen it or create a new one.  Thank you for your continued support of EOSIO!

PR #7734, opened 2019-08-07
Fix db exhaustion

heifner:
## Change Description

- `nodeos_under_min_avail_ram_lr_test` periodically fails because we are not properly handling `database_guard_exception` in all places.
- Add additional catch clauses for `database_guard_exception` via new catch macro in producer_plugin.
- Also address `std::bad_alloc` in controller emit as mentioned in #6979 
- fc exceptions still do not correctly handle `std::bad_alloc`. A different PR to `fc` repo needs to be created to address that.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -44,6 +44,28 @@ using std::vector;
 using std::deque;
 using boost::signals2::scoped_connection;
 
+#undef FC_LOG_AND_DROP
+#define LOG_AND_DROP()  \
+   catch ( const guard_exception& e ) { \
+      chain_plugin::handle_guard_exception(e); \
+   } catch ( boost::interprocess::bad_alloc& ) { \
+      chain_plugin::handle_db_exhaustion(); \
+   } catch( fc::exception& er ) { \
+      wlog( "${details}", ("details",er.to_detail_string()) ); \
+   } catch( const std::exception& e ) {  \
+      fc::exception fce( \
+                FC_LOG_MESSAGE( warn, "std::exception: ${what}: ",("what",e.what()) ), \
+                fc::std_exception_code,\
+                BOOST_CORE_TYPEID(e).name(), \
+                e.what() ) ; \
+      wlog( "${details}", ("details",fce.to_detail_string()) ); \
+   } catch( ... ) {  \
+      fc::unhandled_exception e( \
+                FC_LOG_MESSAGE( warn, "unknown: ",  ), \
+                std::current_exception() ); \
+      wlog( "${details}", ("details",e.to_detail_string()) ); \
+   }
+
 const fc::string logger_name("producer_plugin");
 fc::logger _log;
 
@@ -342,7 +364,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                _unapplied_transactions.add_forked( forked_branch );
             } );
          } catch ( const guard_exception& e ) {
-            chain_plug->handle_guard_exception(e);
+            chain_plugin::handle_guard_exception(e);
             return;
          } catch( const fc::exception& e ) {
             elog((e.to_detail_string()));
@@ -521,7 +543,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             }
 
          } catch ( const guard_exception& e ) {
-            chain_plug->handle_guard_exception(e);
+            chain_plugin::handle_guard_exception(e);
          } catch ( boost::interprocess::bad_alloc& ) {
             chain_plugin::handle_db_exhaustion();
          } CATCH_AND_CALL(send_response);
@@ -703,6 +725,7 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
 { try {
    my->chain_plug = app().find_plugin<chain_plugin>();
    EOS_ASSERT( my->chain_plug, plugin_config_exception, "chain_plugin not found" );
+   chain_plugin* chain_plug = my->chain_plug;
    my->_options = &options;
    LOAD_VALUE_SET(options, "producer-name", my->_producers)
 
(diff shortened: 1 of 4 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -801,13 +827,13 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
    my->_incoming_block_subscription = app().get_channel<incoming::channels::block>().subscribe([this](const signed_block_ptr& block){
       try {
          my->on_incoming_block(block);
-      } FC_LOG_AND_DROP();
+      } LOG_AND_DROP();
```
b1bart: This is still swallowing `std::bad_alloc` by re-throwing it up to :

https://github.com/EOSIO/appbase/blob/4b95cd33659e18bd4f6ea2327202cfa20b09c658/include/appbase/channel.hpp#L28-L30

What is the expected resolution when a `std::bad_alloc` is thrown?  should we call a method like `chain_plugin::handle_db_exhaustion()` specific to heap exhaustion?  is this only thrown from heap exhaustion?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -801,13 +827,13 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
...
    });
 
    my->_incoming_transaction_subscription = app().get_channel<incoming::channels::transaction>().subscribe([this](const transaction_metadata_ptr& trx){
       try {
          my->on_incoming_transaction_async(trx, false, [](const auto&){});
-      } FC_LOG_AND_DROP();
+      } LOG_AND_DROP();
```
b1bart: See other comment about this throwing `std::bad_alloc` into something that will swallow it.

plugins/producer_plugin/producer_plugin.cpp
b1bart: more detail, appbase's channels are not meant to handle exceptions thrown out of subscriptions by default.   We can install handlers on these channels that do something specific with exceptions (other than swallowing them) but it is likely better practice to sanitize the handlers such that they don't throw.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1420,7 +1446,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       }
 
       chain.start_block( block_time, blocks_to_confirm, features_to_activate );
-   } FC_LOG_AND_DROP();
+   } LOG_AND_DROP();
```
b1bart: This will lead to `producer_plugin_impl::start_block()` rethrowing on a `std::bad_alloc` which in turn leads to `producer_plugin_impl::schedule_production_loop()` throwing and I'm pretty sure thats unexpected behavior in the rest of the code.  

We should probably figure out what to do with `std::bad_alloc` and change `LOG_AND_DROP` to properly do that in all of these cases.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1516,10 +1542,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
                      itr = _unapplied_transactions.erase( itr );
                      continue;
                   }
-               } catch( const guard_exception& e ) {
-                  chain_plug->handle_guard_exception( e );
-                  return start_block_result::failed;
-               } FC_LOG_AND_DROP();
+               } LOG_AND_DROP();
```
b1bart: This change ultimately leads to the final result of `producer_plugin_impl::start_block()` being `start_block_result::succeeded` as it removes the early return.  It's possible that other calls into the controller will probably also throw and be `LOG_AND_DROP` 'ed.  That may also lead to `chain_plugin::handle_guard_exception(e);` being called multiple times before the io_service can address the graceful shutdown.  

While that is messy, I'm pretty sure we were already vulnerable to this if we threw guards while processing multiple `_pending_incoming_transactions` at the end of this method.  (each has its own internal handling/dropping of the guard exception).  

I think we need to make sure that once a guard exception is thrown we are on a path out of this method that is guaranteed not to throw/log another guard exception.  It may be more difficult to guarantee that system wide so at maybe protecting the call to `App().quit()` in `chain_plugin` with a check against `App().is_quiting()` can help prevent unexpected situations?

heifner: It is safe to call `app.quit()` multiple times.

plugins/producer_plugin/producer_plugin.cpp
heifner: Why is this `LOG_AND_DROP` here? Seems like any exception from `start_block` should not be ignored.

plugins/producer_plugin/producer_plugin.cpp
heifner: Added a check for `app().is_quiting()` and return failed in that case.

plugins/producer_plugin/producer_plugin.cpp
b1bart: outside of the "special cases" that result in app-terminating behavior like bad_allocs the other exceptions would result in `chain.is_building_block()` on the next line returning `false` ultimately leading to a return of `start_block_result::failed` 

I agree this could be handled better but the callers are not expecting exceptions thrown out of this (we may even want to throw on `noexcept` to make that obvious) so, they all need to be "handled" in some way and reflected in the returned value. 

This is important for any "recoverable" error as this is usually called inside some async task handler, as part of a destructor for an `fc::scoped_exit` that calls `schedule_production_loop()` in its destructor.  Even if we handled it better, throwing exceptions into the `io_service` is not going to be the right move so, we'd be swallowing it somewhere.

heifner: `scoped_exit()` does make it "interesting". Ok, I added he `is_quiting()` check which will return `failed` now.

PR #7736, opened 2019-08-08
unit test ensuring that OOB table init allowed on set code; fails on action

spoonincode:
## Change Description

Out of bound table elements are accepted during setcode but then fail later when the wasm module is actually instantiated when an action is executed. This is arguably the correct behavior when viewed from the standpoint of the wasm spec. However, it's probably not what we intended -- OOB memory init _is_ caught during setcode.

Regardless, this goof isn't really a problem. But, this goof is now consensus. So add a unit test for this goofy behavior.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/wasm_tests.cpp
+++ b/unittests/wasm_tests.cpp
@@ -702,9 +702,7 @@ BOOST_FIXTURE_TEST_CASE( table_init_oob, TESTER ) try {
       
       //the unspecified_exception_code comes from WAVM, which manages to throw a WAVM specific exception
       // up to where exec_one captures it and doesn't understand it
-      BOOST_CHECK_EXCEPTION(push_transaction(trx), fc::exception, [](auto& e) {
-         return e.code() == unspecified_exception_code || e.code() == wasm_execution_error::code_value;
-      });
+      BOOST_CHECK_THROW(push_transaction(trx), eosio::chain::wasm_execution_error);
    };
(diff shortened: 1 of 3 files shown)


unittests/wasm_tests.cpp
```diff
@@ -686,6 +686,44 @@ BOOST_FIXTURE_TEST_CASE( table_init_tests, TESTER ) try {
...
+   trx.actions[0].authorization = vector<permission_level>{{N(tableinitoob),config::active_name}};
+
+    auto pushit_and_expect_fail = [&]() {
+      produce_block();
+      trx.signatures.clear();
+      set_transaction_headers(trx);
+      trx.sign(get_private_key(N(tableinitoob), "active"), control->get_chain_id());
+      
+      //the unspecified_exception_code comes from WAVM, which manages to throw a WAVM specific exception
+      // up to where exec_one captures it and doesn't understand it
```
b1bart: can we catch it somewhere in the wavm plumbing and re-code it to a common `wasm_execution_error` when we still have a chance at understanding it? 

Allowing anything that `FC_RETHROW_EXCEPTIONS` doesn't understand seems very broad and I'd expect us to want to know if this moves on us.

Issue #7658, opened 2019-07-16
nodeos not returning code of '2' when database is dirty


spoonincode:
There is code in nodeos that intends to have nodeos exit with a code of 2 when the database is dirty. It does this by looking at the text in the exception that would typically be fired from chainbase, specifically pinnable_mapped_file

However, appbase swallows exceptions from plugin init and simply returns false from `initialize<>()` causing nodeos to exit with a -1 instead.

This is almost certainly a regression from some point in time; though it's not immediately clear to me if it's worth fixing this behavior or not?

ndcgundlach:
Fixed and merged into develop.

PR #7747, opened 2019-08-10
two useful chainbase backports for 1.8

spoonincode:
## Change Description
This backports EOSIO/chainbase#50 (the fix for boost 1.71) and EOSIO/chainbase#49 (printing which DB causes typical startup failures). These are small and useful for 1.8 users.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/eosio-blocklog/main.cpp
+++ b/programs/eosio-blocklog/main.cpp
@@ -60,7 +60,7 @@ void blocklog::read_log() {
          reversible_blocks.reset();
       }
    } catch( const std::runtime_error& e ) {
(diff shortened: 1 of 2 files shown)


heifner:
@spoonincode This needs to fix applied in https://github.com/EOSIO/eos/pull/7657

spoonincode:
Those don't actually work, AFAIK, see #7658. But fair enough I'll cherry-pick that change in to here to keep consistency.

PR #7751, opened 2019-08-12
[TRAVIS POC] Standardization with eos-vm + various other improvements

NorseGaud:
Modifying various things:

1. Added SKIP_* to each step, defaulting LRTs off
2. We're really close to hitting the max cache size for ccache (500MB). I've added ccache --max-size=1G to get ahead of this.
3. Added generate-serial-tests.sh to automatically generate serial tests for each platform. This required a change from amazonlinux-2.dockerfile to amazon_linux-2.dockerfile since it's now the source of truth (see code)
4. Upgrade to 1.8.1 to ensure that we're running tests against the newest 1.8.x code

SKIP_LONG_RUNNING_TESTS=false : https://buildkite.com/EOSIO/eosio-beta/builds/1079
SKIP_LONG_RUNNING_TESTS default (true) : https://buildkite.com/EOSIO/eosio-beta/builds/1237
https://travis-ci.com/EOSIO/eos/builds/123750028


CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ set( CXX_STANDARD_REQUIRED ON)
 
 set(VERSION_MAJOR 1)
 set(VERSION_MINOR 8)
-set(VERSION_PATCH 0)
+set(VERSION_PATCH 1)
```
kj4ezj: This is not relevant to our pipeline work, so let's make this a separate PR to `release/1.8.x`.

kj4ezj: This is not relevant to our pipeline work, so let's make this a separate PR to `release/1.8.x`.

README.md
```diff
@@ -44,22 +44,22 @@ $ brew remove eosio
 
 #### Ubuntu 18.04 Package Install
 ```sh
-$ wget https://github.com/eosio/eos/releases/download/v1.8.0/eosio_1.8.0-1-ubuntu-18.04_amd64.deb
-$ sudo apt install ./eosio_1.8.0-1-ubuntu-18.04_amd64.deb
+$ wget https://github.com/eosio/eos/releases/download/v1.8.1/eosio_1.8.1-1-ubuntu-18.04_amd64.deb
+$ sudo apt install ./eosio_1.8.1-1-ubuntu-18.04_amd64.deb
```
kj4ezj: This is not relevant to our pipeline work, so let's make this a separate PR to `release/1.8.x`.

.cicd/build.sh
```diff
@@ -0,0 +1,64 @@
...
+        CMAKE_EXTRAS="$CMAKE_EXTRAS -DCMAKE_CXX_COMPILER='clang++' -DCMAKE_C_COMPILER='clang'"
+    fi
+
+    BUILD_COMMANDS="cmake $CMAKE_EXTRAS -DBUILD_MONGO_DB_PLUGIN=true .. && make -j$JOBS"
+
+    # Docker Commands
+    if [[ $BUILDKITE == true ]]; then
+        # Generate Base Images
+        $CICD_DIR/generate-base-images.sh
+        [[ $ENABLE_INSTALL == true ]] && COMMANDS="cp -r $MOUNTED_DIR ~/eosio && cd ~/eosio/build &&"
```
zreyn: Is there a way we can specify a directory, rather than relying on HOME?  I would rather fail in the pipelines if the directory doesn't exist than push images that are inconsistent with others.

.cicd/build.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env bash
+set -eo pipefail
+. ./.cicd/helpers/general.sh
```
zreyn: It looked like in eos-vm, the plan was to move away from helpers. No?  This is not a requirement to change, but would consider flattening the .cicd directory a little to encourage keeping it simple going forward.

* The installation-build.sh could be put into the eosio-installation-images.yml directly
* The docker-hash.sh, general.sh, and clang.make seems to make sense because they reduce code duplication, but they don't really need to be in a subdirectory.

.cicd/build.sh
NorseGaud: Great idea! Adding

.cicd/build.sh
NorseGaud: We're actually able to delete eosio-installation-images.yml and base-images.yml completely which helps. I moved installation-build.sh out into the root since it's not really a helper.

I believe the pattern we've been using for all repos is to keep helpers, dockerfiles, and anything else under directories. Here are some of the reasons I think we should stick to folders:

1. Let's say we delete test-metrics. If it's all contained to a folder, we just delete the folder. This prevents us from missing a file and leaving stuff laying around which isn't used anymore.
2. Visually, it's better looking and easier to find things. If I need to find and replace stuff in dockerfiles,  they're all in docker/, versus spread out in the main directory.

PR #7756, opened 2019-08-13
block_log performance improvement, and misc.

heifner:
## Change Description

- Main performance enhancement in this PR is the switch to using the new `fc::cfile` https://github.com/EOSIO/fc/pull/107 in `block_log`
- Minor performance improvement in `net_plugin` of creating `packed_transaction` and `signed_block` directly from `net_message` without a `std::move`
- Improve logging in `producer_plugin`

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/net_plugin/net_plugin.cpp
```diff
@@ -772,16 +769,11 @@ namespace eosio {
       }
 
       void operator()( signed_block&& msg ) const {
-         // continue call to handle_message on connection strand
-         shared_ptr<signed_block> ptr = std::make_shared<signed_block>( std::move( msg ) );
-         c->handle_message( ptr );
+         EOS_ASSERT( false, plugin_config_exception, "operator()(signed_block&&) should call handle_message" );
```
jgiszczak: I don't understand the phrasing of this error message.  Or the one on the new line 776.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2409,6 +2394,7 @@ namespace eosio {
    bool connection::process_next_message( uint32_t message_length ) {
       try {
          // if next message is a block we already have, exit early
+         auto ds = pending_message_buffer.create_datastream();
```
jgiszczak: What was the reason to relocate the construction of this datastream here?  The immediately preceding comment plus the next 15 lines seem to exist for the express purpose of bypassing the (presumably expensive) construction of the datastream.  If creating the datastream is so cheap, I question the presence of the peek_datastream and its usage here.  It seems like extra work creating two objects, then rereading the same data twice (the block number), only to throw it away the second time.

plugins/net_plugin/net_plugin.cpp
heifner: Updated these all to say: `"Not implemented, call handle_message directly instead"`

plugins/net_plugin/net_plugin.cpp
heifner: Creating the datastream is cheap. Using the datastream is not. But you are correct that this could be better although I would assume the optimizer would do something similar. I've moved the datastream construction closer to use.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -759,29 +756,22 @@ namespace eosio {
       explicit msg_handler( const connection_ptr& conn) : c(conn) {}
 
       void operator()( const signed_block& msg ) const {
-         EOS_ASSERT( false, plugin_config_exception, "operator()(signed_block&&) should be called" );
+         EOS_ASSERT( false, plugin_config_exception, "Not implemented, call handle_message directly instead" );
```
b1bart: can this not be a templated `operator()` that throws the assert and catches all overloads that are not explicitly specified?

heifner: Yes, it can now that no value category of type `packed_transaction[_ptr]` and `signed_block[_ptr]` are allowed.

Issue #7639, opened 2019-07-11
nodeos 1.7 and 1.8 get stuck at 100% CPU usage under heavy p2p load


n8d:
We have some nodes that typically have in the range of 200-400 incoming p2p connections open, and after upgrading to 1.7 we have found that they occasionally jump from about 20% CPU to 100% and stay that way until nodeos restart. If we don't restart they don't recover, and they eventually may have issues where they can't stay in sync with the head block. Typically it takes around 2-3 days after start to have the 100% CPU issue, and then anywhere from hours to days to have sync issues.

This does not happen to the same nodes (same load) when running on 1.6. It also doesn't happen to our other node types (API, BP) on 1.7; just the ones with heavy p2p. We haven't yet tested 1.8 on these nodes.

Here is what the CPU usage looks like when it happens:

![chart2-7](https://user-images.githubusercontent.com/787401/61066495-e7d52800-a3c2-11e9-96ce-382138de4a62.png)

And the peer count during the same timeframe:

![chart2-5](https://user-images.githubusercontent.com/787401/61066415-bc523d80-a3c2-11e9-9e8c-7d952c23e484.png)

Memory utilization (no change, and these are 64GB systems with plenty free):

![chart2-6](https://user-images.githubusercontent.com/787401/61066453-d1c76780-a3c2-11e9-9ff2-4b822aa2c177.png)

Here is a ~30 second perf report while CPU is pegged on the 1.7 node:

<img width="706" alt="image" src="https://user-images.githubusercontent.com/787401/61066635-266ae280-a3c3-11e9-8d24-3b35de2bb353.png">

For comparison, here is a ~30 second perf report on a healthy 1.6 node that is under roughly the same p2p load:

<img width="1006" alt="image" src="https://user-images.githubusercontent.com/787401/61066800-777ad680-a3c3-11e9-946f-7347fa9d68b4.png">

Please let me know what else we can do/collect to help troubleshoot this.

matthewdarwin:
Have you tried 1.8.1?  It seems better than 1.7 from my perspective.

n8d:
> Have you tried 1.8.1? It seems better than 1.7 from my perspective.

Not under this same load, but I will test that soon.

heifner:
Are you running `1.7.0`, if not which exact version are you seeing this on? Please update us when you try 1.8.1. Do you ever see a crash or just cpu spin?

n8d:
[...]

PR #7769, opened 2019-08-16
acceptor thread-safety

heifner:
## Change Description

- `boost::asio::acceptor` is not thread safe; protect use with a `mutex`

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -206,6 +206,7 @@ namespace eosio {
 
    class net_plugin_impl : public std::enable_shared_from_this<net_plugin_impl> {
    public:
+      std::mutex                       acceptor_mtx;
       unique_ptr<tcp::acceptor>        acceptor;
       std::atomic<uint32_t>            current_connection_id{0};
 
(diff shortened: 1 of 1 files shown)


spoonincode:
Where was the problem? Only at shutdown?

heifner:
@spoonincode This was an observation, not any known bug. Although #7639 might be this.

heifner:
Note since it is `async_accept` it is possible for multiple strands to attempt to use the same `acceptor` at the same time.

spoonincode:
>Note since it is async_accept it is possible for multiple strands to attempt to use the same acceptor at the same time.

I'm trying to figure out how that works. It seems like there can only be one `async_accept` outstanding because only after the `async_accept` completes is `start_listen_loop` called again (which does the next `async_accept`)

heifner:
I missed that `start_listen_loop` was inside the lambda. So yes, this is not needed.

spoonincode:
@heifner there might still technically be a problem at shutdown. Maybe the closing of the acceptor should be after the thread_pool is shut down.

heifner:
Agreed. I have another net_plugin PR I'll add that to.

Issue #7767, opened 2019-08-15
node never identifies that its peer is done syncing


brianjohnson5972:
-If node_00 has a lib of 100 and head of 200 and sends out a notify message and handshake message indicating the id of the head to node_01 which has caught up  to lib but not head

-then node_00 produces block num 201

-node_00 has determined that node_01 needs block num 101 (lib+1) to block num 201
since node_01 last saw that node_00's handshake message with block num 200 as head, 

-node_01 determines that it needs to send a handshake message and it has node_01's head which is 200

-after sending the handshake message, node_01 then processes block num 201 from node_00

-node_00 receives the handshake and sees that  node_01 is still behind (block num 200) so it sets the connection->syncing=true, issues a notify message with the current head and sends all the needed blocks up to block num 201

-node_01 receives the blocks and drops them, since they have all been applied

So at this point, node_00 will not send new blocks to node_01 and node_01 has no reason to request new blocks, since it doesn't know about the new blocks.

Solution: In sync_manager::sync_recv_notice, we should check to see if the msg indicates a block number that is equal to or lesser than our current block number, since that would indicate that our understanding of the peer's state is incorrect.  This will correct the sender's assessment that the peer connection is syncing.

heifner:
Fixed by #7770

PR #7770, opened 2019-08-16
Net plugin sync fix

heifner:
## Change Description

- Fix for #7767. We now send a handshake if peer reports a notify for a block we already have since that would indicate our views of the world do not match.
- Added log message for #7767 
- Safer shutdown of `acceptor`, close after thread pool is done.
- Clear `trx_in_progress_size` on close, otherwise connection can never work again until restart. #7639 
- Handle read delay timer error by closing connection to prevent infinite loop. #7639
- Note these are candidate changes for 1.8.x

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1667,7 +1667,13 @@ namespace eosio {
          if (msg.known_blocks.ids.size() == 0) {
             fc_elog( logger,"got a catch up with ids size = 0" );
          } else {
-            verify_catchup(c, msg.known_blocks.pending, msg.known_blocks.ids.back());
+            const block_id_type& id = msg.known_blocks.ids.back();
+            if( !my_impl->dispatcher->have_block( id ) ) {
+               verify_catchup( c, msg.known_blocks.pending, id );
+            } else {
+               // we already have the block, so update peer with our view of the world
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -957,7 +958,8 @@ namespace eosio {
       self->sent_handshake_count = 0;
       self->node_id = fc::sha256();
       my_impl->sync_master->sync_reset_lib_num( self->shared_from_this() );
-      fc_dlog( logger, "closed, canceling wait on ${p}", ("p", self->peer_name()) ); // peer_name(), do not hold conn_mtx
+      fc_ilog( logger, "closing '${a}', ${p}", ("a", self->peer_address())("p", self->peer_name()) );
+      fc_dlog( logger, "canceling wait on ${p}", ("p", self->peer_name()) ); // peer_name(), do not hold conn_mtx
```
brianjohnson5972: Do we want the extra log statement just to make it clear, even though the same info is in the ilog statement?

heifner: It has the additional info of `peer_address()` as requested by #7767

PR #7774, opened 2019-08-19
update WAVM to be compatible with LLVM 7 through 9

spoonincode:
## Change Description
WAVM as used in EOSIO requires LLVM 4.0 -- nothing higher nor lower. This requirement is starting to become a little cumbersome: most distros don't provide an llvm4 package (Ubuntu's last was in 18.10) and building llvm4 from source with recent gcc versions requires patching llvm's source code. Plus, requiring LLVM 4 will conflict with potential new wasm runtimes.

This PR updates WAVM to use LLVM APIs compatible from version 7 to 9.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


CMakeModules/CMakeASM-LLVMWARInformation.cmake
```diff
@@ -0,0 +1,9 @@
+# These trio of files implement a workaround for LLVM bug 39427
+
+set(ASM_DIALECT "-LLVMWAR")
```
swatanabe-b1: What does WAR mean?

libraries/wasm-jit/Source/Runtime/LLVMJIT.cpp
```diff
@@ -21,9 +29,34 @@
...
+      const Uptr numInstructionBytes = LLVMDisasmInstruction(
+         disasmRef,
+         nextByte,
+         numBytesRemaining,
+         reinterpret_cast<Uptr>(nextByte),
+         instructionBuffer,
+         sizeof(instructionBuffer)
+         );
+      WAVM_ASSERT_THROW(numInstructionBytes > 0);
+      WAVM_ASSERT_THROW(numInstructionBytes <= numBytesRemaining);
```
swatanabe-b1: These will leak disasmRef if they trigger.

libraries/wasm-jit/Source/Runtime/LLVMJIT.cpp
```diff
@@ -215,69 +218,48 @@ namespace LLVMJIT
...
+                              auto symbol = symbolSizePair.first;
+                              auto name = symbol.getName();
+                              auto address = symbol.getAddress();
+                              if(symbol.getType() && symbol.getType().get() == llvm::object::SymbolRef::ST_Function && name && address) {
+                                 Uptr loadedAddress = Uptr(*address);
+                                 auto symbolSection = symbol.getSection();
+                                 if(symbolSection)
+                                    loadedAddress += (Uptr)o.getSectionLoadAddress(*symbolSection.get());
+                                 //printf(">>> %s 0x%lX\n", symbolSizePair.first.getName()->data(), loadedAddress);
+                                 std::map<U32,U32> offsetToOpIndexMap;
```
swatanabe-b1: This parameter looks dead to me.  I couldn't find any other place that calls notifySymbolLoaded, and nothing ever sets the map to non-empty.  Am I missing something?

CMakeLists.txt
```diff
@@ -11,7 +11,7 @@ endif()
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules")
 
 if (UNIX AND APPLE)
-   list(APPEND CMAKE_PREFIX_PATH "/usr/local/opt/llvm@4" "/usr/local/opt/gettext")
+   list(APPEND CMAKE_PREFIX_PATH "/usr/local/opt/llvm@7" "/usr/local/opt/gettext")
```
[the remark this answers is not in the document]
spoonincode: I think so, because as is it is not compatible with 10 🙄 At the moment the only "pinned" version of LLVM between 7-9 on homebrew is 7.

CMakeModules/CMakeASM-LLVMWARInformation.cmake
spoonincode: workaround

libraries/wasm-jit/Source/Runtime/LLVMJIT.cpp
spoonincode: You are correct. Originally, WAVM's `notifySymbolLoaded()` would consume this and create debug information mapping each wasm opcode to the x86 instruction(s) it generated. As you can imagine that created quite the memory bloat 😳 I removed that code from `notifySymbolLoaded()` but decided not to refactor away the function signature as I was trying to minimize the changes to WAVM. Since WAVM will be deprecated in 2.0 I would rather not continue hacking it up further here.

libraries/wasm-jit/Source/Runtime/LLVMJIT.cpp
```diff
@@ -215,69 +218,48 @@ namespace LLVMJIT
...
-			NotifyFinalizedFunctor(JITUnit* inJITUnit): jitUnit(inJITUnit) {}
-			void operator()(const llvm::orc::ObjectLinkingLayerBase::ObjSetHandleT& objectSetHandle);
-		};
-		typedef llvm::orc::ObjectLinkingLayer<NotifyLoadedFunctor> ObjectLayer;
-		typedef llvm::orc::IRCompileLayer<ObjectLayer> CompileLayer;
+		typedef llvm::orc::LegacyIRCompileLayer<llvm::orc::LegacyRTDyldObjectLinkingLayer, llvm::orc::SimpleCompiler> CompileLayer;
 
-		UnitMemoryManager memoryManager;
-		std::unique_ptr<ObjectLayer> objectLayer;
+      llvm::orc::ExecutionSession ES;
```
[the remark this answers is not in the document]
spoonincode: fair enough, I can change it back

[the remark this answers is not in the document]
swatanabe-b1: The whole file is a mess in terms of whitespace, but at least avoid making in worse.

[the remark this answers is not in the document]
spoonincode: it just has to last one more release so hold your 👃at most formatting issues

PR #7783, opened 2019-08-21
Add supporting code for WASM spec tests

jeffreyssmith2nd:
## Change Description

This PR adds the supporting code (CMake and main + a header file) to allow us to run the WASM spec tests as EOSIO unittests. #7784 contains the actual wasms and cpp files.

CMake and compilation take a significant amount of time with these additions, so they are guarded by the WASM_SPEC_TESTS CMake option.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
Indeed, compilation time is dreadful. Is there a reason each one is in its own .cpp file? Are they being auto-generated or something? My guess is if we consolidated them some the compile time would be more manageable and we could leave them always on.

jeffreyssmith2nd:
@spoonincode Yes, they're being auto generated using the code in this PR on eos-vm https://github.com/EOSIO/eos-vm/pull/79

jeffreyssmith2nd:
> @spoonincode Yes, they're being auto generated using the code in this PR on eos-vm [EOSIO/eos-vm#79](https://github.com/EOSIO/eos-vm/pull/79)

That being said, that doesn't necessarily mean they have to be broken out the way they currently are.

spoonincode:
Just combining them all was still over 4 minutes for me

swatanabe-b1:
It looks like the only thing that varies between the test cases is the test_index.  You can use BOOST_DATA_TEST_CASE and drastically reduce the amount of generated code.

jeffreyssmith2nd:
Good suggestion, that does get rid of a massive amount of code. I'm going to update the generator to use that as well as look at combining files to see if there's any significant compilation improvement.

jeffreyssmith2nd:
Here's a comparison of compiling without tests, with them as they currently are, and switching to BOOST_DATA_TEST_CASE and combining by name (all `address` in one file, all `call` in another, etc.)

| Type  | CMake | Build |
| ------------- | ------------- | ------------ |
| No Spec Tests | 21s | 174s |
| Spec Tests in different files  | 302s | 613s |
| Spec Tests in the same files | 48s | 235s |

jeffreyssmith2nd:
Closing in favor of #7955

PR #7785, opened 2019-08-22
Optimize block log usage

heifner:
## Change Description

- Emit accepted block signal before updating block log since accepted block does not depend on lib. This allows the `net_plugin` to start sending out the block sooner.
- Optimize getting `block_id` via `block_num` out of block log by reading only the `block_header` instead of the entire block.
- Provide access to `head_id` of block log instead of calculating it from the `head` block.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -377,7 +377,7 @@ struct controller_impl {
       auto root_id = fork_db.root()->id;
 
       if( log_head ) {
-         EOS_ASSERT( root_id == log_head->id(), fork_database_exception, "fork database root does not match block log head" );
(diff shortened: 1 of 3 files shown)


libraries/chain/controller.cpp
```diff
@@ -2697,12 +2697,12 @@ block_id_type controller::last_irreversible_block_id() const {
    if( block_header::num_from_id(tapos_block_summary.block_id) == lib_num )
       return tapos_block_summary.block_id;
 
-   auto signed_blk = my->blog.read_block_by_num( lib_num );
+   auto id = my->blog.read_block_id_by_num( lib_num );
```
brianjohnson5972: Could we change the contents of this method to just call get_block_id_for_num?

PR #7791, opened 2019-08-22
Producer plugin

heifner:
## Change Description

- Producer plugin was waking up a block early so that it would for sure not miss its production slot. However, this causes it to work twice as hard for the first block production. Under heavy load, setting exact wake up time works much better.
- When block is exhausted by processing `start_block` transactions (scheduled, pending, un-applied) do not set a timer, but immediately produce block.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1322,12 +1322,6 @@ fc::time_point producer_plugin_impl::calculate_block_deadline( const fc::time_po
    return block_time + fc::microseconds(last_block ? _last_block_time_offset_us : _produce_time_offset_us);
 }
 
-enum class tx_category {
-   PERSISTED,
-   UNEXPIRED_UNPERSISTED
-};
-
-
 producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    chain::controller& chain = chain_plug->chain();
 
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1750,9 +1746,13 @@ void producer_plugin_impl::schedule_production_loop() {
          auto expect_time = chain.pending_block_time() - fc::microseconds(config::block_interval_us);
          // ship this block off up to 1 block time earlier or immediately
          if (fc::time_point::now() >= expect_time) {
-            _timer.expires_from_now( boost::posix_time::microseconds( 0 ));
-            fc_dlog(_log, "Scheduling Block Production on Exhausted Block #${num} immediately",
+            // produce block immediately
```
b1bart: While I cannot find a direct reason this is an issue, there are a few implicit requirements that this may one day run-afoul of that I think we should attempt to make more explicit

1. `schedule_production_loop` is almost always called from a destructor of an `fc::scoped_exit` which means it really shouldn't throw (maybe one day we should consider requiring `noexcept` qualifiers on lambdas for `fc::scoped_exit).  Right now, `maybe_produce_block` appears to handle all of the exceptions and is clean but I'd be interested in how we make sure the implicit requirement that it not throw is made explicit. 

2. `maybe_produce_block` calls `schedule_production_loop` making this recursive.  I don't think it can get into an infinite recursion because this branch is taken based on time but, there are a lot of variables and if this stays direct, perhaps we should "schedule" the call to `schedule_production_loop` in `maybe_produce_block` so that it at least clears the stack ?

at the least we should probably make the log messages the same (other than the  notice about exhaustion) and turn the call and the `fc_dlog`calls into a local lambda so that the messages are kept consistent.

heifner: Reverted this change as I think it is clearer and more consistent the way it was.

PR #7801, opened 2019-08-26
Add keosd docs

dixia:
Add docs for keosd

## Change Description

Add the first batch of docs for keosd.  Some of the content is adopted from the communication with Bart and Matt.

As another PR(docs/starter) is not merged, I created this PR against `docs/starter` to avoid duplicated reviews. Given there are no `keosd` related docs in `eos` repo, it should be immaterial if this PR switch to compare with `develop` later.

if possible, it would be good to start to review the checked-in changes now

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


docs/keosd/basic-usage.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Stopping keosd
+
+The most effective way to stop `keosd` is to find the keosd process and send a SIGTERM signal to it.
+
+## Other all options
+
+For a list of all commands known to `keosd`, simply run it with no arguments:
+
+```bash
```
b1bart: Is this section auto generated from the actual output of the command?  it seems highly likely to drift as we add things to the options

docs/keosd/how-to-guides/how-to-attach-a-YubiHSN.md
```diff
@@ -0,0 +1,69 @@
+## Goal
+
+Attach a YubiHSM as a hard wallet
+
+## Before you begin
+
+* Install the currently supported version of kleos
```
b1bart: `kleos` is `cleos` and `keosd`?

b1bart: also elsewhere it has backticks which I think it should have here unless we want to refer to a product suite

docs/keosd/introduction.md
```diff
@@ -1,7 +1,5 @@
-## Introduction
+`keosd` is a key manager for store private keys and sign transactions created by `cleos`. It providdes a secure key storage medium so that when the private keys is stored in a wallet file , it will be encrypted at rest. Only when a wallet is unlocked with the wallet password, `cleos` can request `keosd` to sign a transaction with these private keys. `keosd` provides support for hardware-based wallets as well such as Secure Encalve and YubiHSM
 
-What is keosd. What is its goal and scope
+`keosd` is intended to be used by EOSIO developers and power users
```
b1bart: I don't think we want to frame `keosd` as a tool for power users, only developers really.  Basically, it has not had the scrutiny applied to it would need to hold keys for large amounts of real value.   It may be up to that task but we certainly do not want to position it as such.

docs/keosd/basic-usage.md
dixia: It is not auto-generated at the moment. I think it is a good idea to programmatically do this though

dskvr: @b1bart As discussed last January, a custom parser would have to be written and maintained for cleos/keosd cli's. This is absolutely the mid-longterm goal.

dixia:
> I did not bother to review anything outside of `docs/keosd` markdown files as there are a few hundred merged file changes in this PR. Specific comments are called out and my one general comment is that there are grammar and spelling mistakes in most of the paragraphs. I felt it was unproductive to call out by hand when automated tools exist to check and correct these issues.

Yep, unfortunately, there are many unrelated commits to make sure the pipeline could be kick start. Only the content under the `docs/keosd` need to be review. Typo should be fixed as well. Apparently, there was an issue with my VSCode.

dixia:
> Some small issues that need to be rectified before being published, will also be committing some index files for navigation consistency.

Yes, that was due to a bad conflict merge. It has been resolved now

PR #7803, opened 2019-08-26
callback support for checktime timer expiry

spoonincode:
## Change Description

When a transaction's checktimer timer expires, call a user specified function. This can allow potential future wasm runtimes to be notified of expiry in a very low overhead manner (they don't need to poll expiry).

This PR tries to KISS and just adds support for a single callback by storing two `void*`s in the platform_timer. Since the callback could be called within a signal handling context I didn't want to complicate manners with `std::function`s or anything that required resizing to support more than one.

Usage could be something like this:
```c++
super_fast_wasm_runtime::execute_action(apply_context& context, ...) {
   context.trx_context.transaction_timer.set_expiry_callback([](void* user) {
      //this code could be called from signal handling context!
      super_fast_wasm_runtime* self = (super_fast_wasm_runtime*)user;
      self->stop_wasm_execution();
   }, this);
   context.trx_context.checktime(); //catch any expiration that might have occurred before setting up callback

   //when wasm code is done, disable callback
   auto reset_expiry_cb = fc::make_scoped_exit([&tt=context.trx_context.transaction_timer](){tt.set_expiry_callback(nullptr, nullptr);});

   run_wasm();
}
```

You will notice what technically amounts to a very slight race condition here. If the timer expires at the very same time that the callback is attempting to be disabled (by calling `set_expiry_callback(nullptr, nullptr)`) it's conceivable that the callback will still be called even though it has been disabled at that point.

It's not clear what the best way is to solve that problem. The expiry code being run from a signal handling context ties my hands on many typical solutions. Any thoughts on this are welcome.

However, this issue isn't a problem for potential new wasm runtime given the current state of single threaded execution -- slight late execution of the callback even after disabling it doesn't cause a problem (as long it isn't so late that it is called after the next transaction's wasm starts executing, which would be absurd in practice). In fact, the only reason potential new wasm runtime needs to disable the callback at all is just to make platform_timer's assert() happy that no one is trying to pile up more than one callback.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/platform_timer.hpp
+++ b/libraries/chain/include/eosio/chain/platform_timer.hpp
@@ -4,6 +4,8 @@
 
 #include <eosio/chain/exceptions.hpp>
 
+#include <atomic>
+
 #include <signal.h>
 
 namespace eosio { namespace chain {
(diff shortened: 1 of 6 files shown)


spoonincode:
> simpler to reason about and likely safer

At this point it's probably easier & safer to just use the threaded approach. Since we could safely use a real lock it would allow solving all problems (even prevent late firing of the callback).

spoonincode:
Unfortunately using a dedicated thread for signal handling causes high variation in timer quality. Probably due to a dormant thread needing to be completely woke up and scheduled when the timer expires. No go here.

PR #7835, opened 2019-08-31
Add protocol feature CODE_VERSION

### LINKED ISSUE #6168: Consensus upgrade to allow contracts to read code version of any account

conr2d:
## Change Description

Fix #6168.

## Consensus Changes
- [x] Consensus Changes

Consensus upgrade.

## API Changes
- [x] API Changes

New intrinsic API `get_code_version` is introduced by this PR.

## Documentation Additions
- [x] Documentation Additions

The documentation of eosio.cdt needs to include `get_code_version`.


libraries/chain/wasm_interface.cpp
```diff
@@ -987,6 +987,13 @@ class authorization_api : public context_aware_api {
       return context.is_account( account );
    }
 
+   void get_code_version( const account_name& account, int64_t& last_code_update_time, fc::sha256& code_hash ) {
```
swatanabe-b1: intrinsics with multiple output parameters have very strange semantics when the pointers alias.  I would strongly recommend either using a packed struct or splitting this into two intrinsics.

unittests/test-contracts/code_version_test/code_version_test.cpp
```diff
@@ -0,0 +1,11 @@
+#include "code_version_test.hpp"
+
+using namespace eosio;
+
+void code_version_test::reqversion( name name, checksum256 version ) {
+   check( get_code_version(name) == version, "code version not match" );
```
swatanabe-b1: The error message is not grammatically correct.

unittests/test-contracts/code_version_test/code_version_test.cpp
```diff
@@ -0,0 +1,11 @@
...
+}
+
+void code_version_test::reqtime( name name, time_point last_code_update ) {
+   check( get_code_last_update_time(name) == last_code_update, "last update time not match" );
```
swatanabe-b1: Same as above.

unittests/test-contracts/code_version_test/code_version_test.hpp
```diff
@@ -0,0 +1,39 @@
+#pragma once
+
+#include <eosio/eosio.hpp>
+
+namespace eosio {
```
swatanabe-b1: It's better not to put this stuff in namespace eosio, as that causes a headache when cdt is eventually updated.

PR #7836, opened 2019-09-03
bandwidth - 1.8.x

### LINKED ISSUE #7829: nodeos 1.8.2 takes too much bandwith at times
Upgrade nodeos and all connected peers to 1.8.2 on EOS Mainnet.  Sometimes get very high network bandwidth usage.  Restarting nodeos clears the problem.

Example "iftop" output:

image

detailed logging starts at 14:27
[log.txt.gz](https://github.com/EOSIO/eos/files/3556095/log.txt.gz)

heifner:
## Change Description

- Resolves #7829 
- Only `request_message` for our head greater than `msg.head` when we suspect they are on a fork.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1441,6 +1441,8 @@ namespace eosio {
       //
       // 3  my head block num < peer head block num - update sync state and send a catchup request
       // 4  my head block num >= peer block num send a notice catchup if this is not the first generation
+      //    4.1 if peer appears to be on a different fork ( our_id_for( msg.head_num ) != msg.head_id )
+      //        then request peer's blocks
       //
       //-----------------------------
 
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1499,10 +1499,12 @@ namespace eosio {
             c->enqueue( note );
          }
          c->syncing = true;
-         request_message req;
-         req.req_blocks.mode = catch_up;
-         req.req_trx.mode = none;
-         c->enqueue( req );
+         if( cc.get_block_id_for_num( msg.head_num ) != msg.head_id ) {
```
b1bart: There is a comment block above where we should probably note what this branch "means"

Something like 
```
4.1 My head block number >= peer head block number but they appear to be on a different fork which may be better and should be requested
```

b1bart: At the very least we should note the difference in comments in `develop` I think these backports also should bear the comments though.

heifner: Good point. I'll update all of them.

PR #7845, opened 2019-09-03
Deprecate network version match - develop

heifner:
## Change Description

- `nodeos` option `network-version-match` when set to non-default `true`, needlessly restricted connections between two `nodeos`. The option requires both `nodeos` to have the exact same peer network version even if the versions are compatible. Since the next major release will bump the network version but still support and work seamlessly with the older network version, this option is being deprecated as enabling it provides no benefit.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions
- Note deprecation and planned removal of `network-version-match`.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3239,6 +3239,8 @@ namespace eosio {
          peer_log_format = options.at( "peer-log-format" ).as<string>();
 
          my->network_version_match = options.at( "network-version-match" ).as<bool>();
+         if( my->network_version_match )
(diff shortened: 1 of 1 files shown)


spoonincode:
Should we log a warning when someone uses the depreciated option? Like we do for private-key.

heifner:
I started to add that this morning, but didn't know if we wanted it or not.

spoonincode:
It would at least make it consistent with the other depreciated option. And it would probably hasten its deactivation by users, or at least hasten the visibility of the depreciation. Seems like a good thing to me.

PR #7847, opened 2019-09-05
Introduce get_transaction_merkle_proof RPC

conr2d:
## Change Description

Merkle proof is very basic feature used in inter-blockchain communication. It can be done in application level, but it would be helpful provided by RPC API.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

`/v1/chain/get_transaction_merkle_proof` is added.

## Documentation Additions
- [x] Documentation Additions

API needs to be documented.

### DIFF
--- a/libraries/chain/include/eosio/chain/merkle.hpp
+++ b/libraries/chain/include/eosio/chain/merkle.hpp
@@ -19,4 +19,14 @@ namespace eosio { namespace chain {
     */
    digest_type merkle( vector<digest_type> ids );
 
+   /**
+    * Generates a merkle proof, a set of digests from leaf to root
+    */
+   vector<digest_type> generate_merkle_proof( digest_type id, vector<digest_type> ids );
+
+   /**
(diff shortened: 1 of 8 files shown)


conr2d:
@swatanabe-b1 
Thank you for your review. Is there any conventional way to test RPC in this repository? I couldn't find any tests for other RPCs, so I'm not sure where to put my test in.

swatanabe-b1:
It should probably go in tests/.  You can add a getProof method in tests/Node.py.

swatanabe-b1:
This should really be handled in the SDKs instead as a new RPC endpoint.

conr2d:
@swatanabe-b1 This proof can be acquired from the result of `/v1/chain/get_block` with full list of transactions, but I thought that getting whole block costs too much for getting a proof of one transaction. Can you let me know what SDK you mentioned means? (eosjs or history-tools?)

PR #7864, opened 2019-09-06
Add output of build info on nodeos startup

heifner:
## Change Description

- Add output of build info of `nodeos` via `--print-build-info` & `--extract-build-info`
- Example:
```
10:28:58@heifnerk:~/ext/eosio/programs/nodeos$ ./nodeos --print-build-info
info  2019-09-09T15:29:17.655 nodeos    chain_plugin.cpp:568          plugin_initialize    ] initializing chain plugin
info  2019-09-09T15:29:17.655 nodeos    chain_plugin.cpp:582          plugin_initialize    ] Build environment JSON:
{
  "debug": false,
  "os": "OS_LINUX",
  "arch": "ARCH_X86_64",
  "boost_version": 107000,
  "compiler": "4.2.1 Compatible Clang 7.0.0 (tags/RELEASE_700/final)"
}
warn  2019-09-09T15:29:17.655 nodeos    chain_plugin.cpp:960          plugin_initialize    ] 3100009 node_management_success: Node management operation successfully executed
reported build environment information
    {}
    nodeos  chain_plugin.cpp:583 plugin_initialize
Failed to initialize
```

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

- Add documentation for new `--print-build-info` & `--extract-build-info`

```
  --print-build-info                    print build environment information to 
                                        console as JSON and exit
  --extract-build-info arg              extract build environment information 
                                        as JSON, write into specified file, and
                                        exit
```

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -274,6 +274,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "extract genesis_state from blocks.log as JSON, write into specified file, and exit")
          ("print-build-info", bpo::bool_switch()->default_value(false),
           "print build environment information to console as JSON and exit")
+         ("extract-build-info", bpo::value<bfs::path>(),
+          "extract build environment information as JSON, write into specified file, and exit")
          ("fix-reversible-blocks", bpo::bool_switch()->default_value(false),
           "recovers reversible block database if that database is in a bad state")
          ("force-all-checks", bpo::bool_switch()->default_value(false),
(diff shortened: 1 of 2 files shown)


heifner:
@kj4ezj @arhag thoughts on using this info for a test.

Options:
1. Create a unittest that has a list of files with this info for all platforms we support and pass if any are found that match. One issue with this approach is this will need to be disable for debug (easy enough with `#ifdef NDEBUG`, but don't know how to determine if unpinned build.

2. Create a python test that looks at this output. The python might be able to more easily look at our configuration and know correct values. The correct values can just be part of the python file.

kj4ezj:
Personally, I am a fan of using something like this for the integration test:
```
./nodeos --config-dir $(pwd)/config --data-dir $(pwd)/data & sleep 10; kill $!
[[ $(xxd -seek 9 -l 1 data/state/shared_memory.bin | awk '{print $2}') > 00 ]] && echo 'FAIL' || echo 'pass'
```
This code is already supported by `release/1.8.x`, `master`, and `develop`. It is also much easier to deterministically parse the `shared_memory.bin` file in hexadecimal than it is to parse log files. We've had a lot of trouble parsing log files for test results. I also think developers will be less likely to change the binary file header than to change what `nodeos` is outputting.   
   
That being said, while I don't think this print statement is so useful for the machines, I think it is incredibly useful for us humans. I'd like to commit it.

spoonincode:
This is a good idea but I wonder if it should be a command line option instead e.g. `nodeos --print-build-environment` and/or if the output should be in a parsable manner like maybe, yes I'm going to suggest it, JSON 🙊 

```
$ nodeos --print-build-environment
{"compiler":"Compiler: 4.2.1 Compatible Clang 7.0.0 (tags/RELEASE_700/final)", "debug":false}, etc
```

heifner:
> This is a good idea but I wonder if it should be a command line option instead e.g. nodeos --print-build-environment and/or if the output should be in a parsable manner like maybe, yes I'm going to suggest it, JSON 🙊

I think prettified JSON would work, so you get human and machine readable.

```
$ nodeos --print-build-environment
{
  "compiler":"Compiler: 4.2.1 Compatible Clang 7.0.0 (tags/RELEASE_700/final)", 
  "debug":false,
  "OS":"Linux",
  "Arch": "x86_64",
  "Boost":"1.70.0"
}
```

heifner:
```
10:28:58@heifnerk:~/ext/eosio/programs/nodeos$ ./nodeos --print-build-info
info  2019-09-09T15:29:17.655 nodeos    chain_plugin.cpp:568          plugin_initialize    ] initializing chain plugin
info  2019-09-09T15:29:17.655 nodeos    chain_plugin.cpp:582          plugin_initialize    ] Build environment JSON:
{
  "debug": false,
  "os": "OS_LINUX",
  "arch": "ARCH_X86_64",
  "boost_version": 107000,
  "compiler": "4.2.1 Compatible Clang 7.0.0 (tags/RELEASE_700/final)"
}
warn  2019-09-09T15:29:17.655 nodeos    chain_plugin.cpp:960          plugin_initialize    ] 3100009 node_management_success: Node management operation successfully executed
reported build environment information
    {}
    nodeos  chain_plugin.cpp:583 plugin_initialize
Failed to initialize
```

PR #7877, opened 2019-09-10
Create Release Build Test

kj4ezj:
## Change Description
One problem we identified on Friday is that we have no good way of knowing whether a binary was built with debugging symbols or not. Debugging symbols enable software engineers to inspect and control a running program with a debugging tool, but they significantly slow down
performance-critical applications like `nodeos`. I created an integration test which fails when EOSIO is built with debugging symbols.  
  
Though I can write the test to only fail in a CI environment like Buildkite, it currently fails for local debugging builds as well. I know a lot of members of our community build and install EOSIO for production systems from source, and I wanted this test to be able to catch build-type errors for those users despite any inconvenience to our own developers. This test behavior is open to persuasion from any stakeholders.  
  
The test provides useful error messages when test prerequisites are not met, and when the test fails.
   
## Consensus Changes
- [ ] Consensus Changes
None.

## API Changes
- [ ] API Changes
None.

## Documentation Additions
- [ ] Documentation Additions
None.

### DIFF
--- a/tests/release-build.sh
+++ b/tests/release-build.sh
@@ -34,6 +34,7 @@ if [[ ! -f data/state/shared_memory.bin ]]; then
     echo "$ ls -la \"$(pwd)/data/state\""
     ls -la "$(pwd)/data/state"
     echo 'Release build test not run.'
+    rm -rf config data
     exit 2
 fi
 # test state files for debug flag
@@ -48,4 +49,5 @@ echo "Debug Byte = 0x$DEBUG_BYTE"
 echo 'First kilobyte of shared_memory.bin:'
 echo '$ xxd -l 1024 shared_memory.bin'
 xxd -l 1024 data/state/shared_memory.bin
+rm -rf config data
 exit 3
--- a/tests/release-build.sh
+++ b/tests/release-build.sh
@@ -10,7 +10,7 @@ echo 'performance-critical applications like nodeos. Anyone intending to build a
 echo 'install nodeos from source should perform a "release build," which excludes'
 echo 'debugging symbols to generate faster and lighter binaries.'
 echo ''
-# environment
+# find nodeos
 [[ -z "$EOSIO_ROOT" && $(git --version) ]] && export EOSIO_ROOT="$(git rev-parse --show-toplevel)"
 [[ -z "$EOSIO_ROOT" ]] && export EOSIO_ROOT="$(echo $(pwd)/ | grep -ioe '.*/eos/' -e '.*/eosio/' -e '.*/build/' | sed 's,/build/,/,')"
 if [[ ! -f "$EOSIO_ROOT/build/bin/nodeos" && ! -f "$EOSIO_ROOT/build/programs/nodeos/nodeos" ]]; then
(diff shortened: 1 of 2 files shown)


heifner:
This information is available via the new `--extract-build-info` https://github.com/EOSIO/eos/pull/7864

```
{
  "debug": false,
  "os": "OS_LINUX",
  "arch": "ARCH_X86_64",
  "boost_version": 107000,
  "compiler": "4.2.1 Compatible Clang 7.0.0 (tags/RELEASE_700/final)"
}
```

tests/release-build.sh
```diff
@@ -0,0 +1,55 @@
...
+# check for jq
+if ! $(jq --version 1>/dev/null); then
+    echo 'ERROR: Test requires jq, but jq was not found in your PATH!'
+    echo ''
+    exit 1
+fi
+# find nodeos
+[[ -z "$EOSIO_ROOT" && $(git --version) ]] && export EOSIO_ROOT="$(git rev-parse --show-toplevel)"
+[[ -z "$EOSIO_ROOT" ]] && export EOSIO_ROOT="$(echo $(pwd)/ | grep -ioe '.*/eos/' -e '.*/eosio/' -e '.*/build/' | sed 's,/build/,/,')"
+if [[ ! -f "$EOSIO_ROOT/build/bin/nodeos" && ! -f "$EOSIO_ROOT/build/programs/nodeos/nodeos" ]]; then
```
arhag: You know that this test file needs to be located within the `tests` subdirectory of the build directory. So you can just get the path of the source file and then find nodeos relative to it (`../programs/nodeos/nodes`) rather than dealing with git commands or assuming the build folder is named `build`.

tests/release-build.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# test name and purpose
+echo ''
+echo '                         ##### Release Build Test #####'
+echo ''
+echo '    The purpose of this test is to ensure that nodeos was built with compiler'
```
arhag: To add to @swatanabe-b1's comment [here](https://github.com/EOSIO/eos/pull/7882#issuecomment-530543223),  this test isn't truly checking if nodeos was built with debug symbols nor is it checking which optimization level it was built with. It is, however, checking part of the build environment (specifically the debug flag) embedded in nodeos which determines whether nodeos is compatible with a particular Chainbase file it is attempting to open. Checking whether the released binaries are compatible with existing Chainbase files is probably the more important thing to check as part of our CICD process; the CICD sync tests do a good job of that but this test is intended to be a faster way of catching a class of problems that we know would lead to failures in our sync tests later on. However, due to correlation between the optimization setting and status of `NDEBUG`, this test also kind of acts as a check of the compiler optimization used (at least for the typical way the build would be configured with `CMAKE_BUILD_TYPE`).

tests/release-build.sh
```diff
@@ -0,0 +1,55 @@
...
+    echo 'ERROR: Build info JSON file not found!'
+    echo ''
+    echo 'Looked in the following places:'
+    echo "$ ls -la \"$(pwd)\""
+    ls -la "$(pwd)"
+    echo 'Release build test not run.'
+    exit 2
+fi
+# test state files for debug flag
+if [[ "$(cat build-info.json | jq .debug)" == 'false' ]]; then
```
arhag: We would benefit from a test very similar to this one that compares the full `build-info.json` to an expected JSON build info that we commit to the repository. However, that one should only run on pinned builds.

.cicd/platforms/amazon_linux-2-unpinned.dockerfile
```diff
@@ -5,7 +5,7 @@ RUN yum update -y && \
     yum install -y which git sudo procps-ng util-linux autoconf automake \
     libtool make bzip2 bzip2-devel openssl-devel gmp-devel libstdc++ libcurl-devel \
     libusbx-devel python3 python3-devel python-devel libedit-devel doxygen \
-    graphviz clang patch llvm-devel llvm-static
+    graphviz clang patch llvm-devel llvm-static vim-common jq
```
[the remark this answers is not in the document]
kj4ezj: I chose to include `xxd` here and `jq` on the other pull request so that these tools would be available in the container when the container is pulled or SSH'ed into for debugging. heifner and jerstern were doing this today, and I think this practice will become more common.

tests/release-build.sh
kj4ezj: I will add this to my backlog.

spoonincode:
fwiw `-frecord-gcc-switches` works _really_ well in clang (not so much gcc). If you really wanted to check it was really made with optimization turned on and/or track other flags.

PR #7882, opened 2019-09-10
Create Release Build Test

kj4ezj:
## Change Description
One problem we identified on Friday is that we have no good way of knowing whether a binary was built with debugging symbols or not. Debugging symbols enable software engineers to inspect and control a running program with a debugging tool, but they significantly slow down
performance-critical applications like `nodeos`. I created an integration test which fails when EOSIO is built with debugging symbols.  
  
Though I can write the test to only fail in a CI environment like Buildkite, it currently fails for local debugging builds as well. I know a lot of members of our community build and install EOSIO for production systems from source, and I wanted this test to be able to catch build-type errors for those users despite any inconvenience to our own developers. This test behavior is open to persuasion from any stakeholders.  
  
The test provides useful error messages when test prerequisites are not met, and when the test fails.
   
## Consensus Changes
- [ ] Consensus Changes
None.

## API Changes
- [ ] API Changes
None.

## Documentation Additions
- [ ] Documentation Additions
None.

### DIFF
--- a/.cicd/platforms/amazon_linux-2-unpinned.dockerfile
+++ b/.cicd/platforms/amazon_linux-2-unpinned.dockerfile
@@ -5,7 +5,7 @@ RUN yum update -y && \
     yum install -y which git sudo procps-ng util-linux autoconf automake \
     libtool make bzip2 bzip2-devel openssl-devel gmp-devel libstdc++ libcurl-devel \
     libusbx-devel python3 python3-devel python-devel libedit-devel doxygen \
-    graphviz clang patch
+    graphviz clang patch vim-common
 # build cmake.
 RUN curl -LO https://cmake.org/files/v3.13/cmake-3.13.2.tar.gz && \
     tar -xzf cmake-3.13.2.tar.gz && \
--- a/tests/release-build.sh
+++ b/tests/release-build.sh
@@ -34,6 +34,7 @@ if [[ ! -f data/state/shared_memory.bin ]]; then
     echo "$ ls -la \"$(pwd)/data/state\""
     ls -la "$(pwd)/data/state"
     echo 'Release build test not run.'
+    rm -rf config data
     exit 2
 fi
 # test state files for debug flag
@@ -48,4 +49,5 @@ echo "Debug Byte = 0x$DEBUG_BYTE"
 echo 'First kilobyte of shared_memory.bin:'
 echo '$ xxd -l 1024 shared_memory.bin'
 xxd -l 1024 data/state/shared_memory.bin
+rm -rf config data
 exit 3
--- a/tests/release-build.sh
+++ b/tests/release-build.sh
@@ -10,7 +10,7 @@ echo 'performance-critical applications like nodeos. Anyone intending to build a
 echo 'install nodeos from source should perform a "release build," which excludes'
 echo 'debugging symbols to generate faster and lighter binaries.'
 echo ''
-# environment
+# find nodeos
 [[ -z "$EOSIO_ROOT" && $(git --version) ]] && export EOSIO_ROOT="$(git rev-parse --show-toplevel)"
 [[ -z "$EOSIO_ROOT" ]] && export EOSIO_ROOT="$(echo $(pwd)/ | grep -ioe '.*/eos/' -e '.*/eosio/' -e '.*/build/' | sed 's,/build/,/,')"
 if [[ ! -f "$EOSIO_ROOT/build/bin/nodeos" && ! -f "$EOSIO_ROOT/build/programs/nodeos/nodeos" ]]; then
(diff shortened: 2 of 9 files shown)


swatanabe-b1:
This is not checking for debug symbols.  It's checking whether nodeos was built with -DNDEBUG.  For performance, the thing that matters most is whether nodeos was built with optimization enabled.  The three are usually correlated (Debug = -g vs Release = -DNDEBUG -O3), but they can be set independently.

kj4ezj:
> This is not checking for debug symbols. It's checking whether nodeos was built with -DNDEBUG.
Thank you, I will update the description to accurately reflect what it is doing.

.cicd/platforms/amazon_linux-2-unpinned.dockerfile
```diff
@@ -5,7 +5,7 @@ RUN yum update -y && \
     yum install -y which git sudo procps-ng util-linux autoconf automake \
     libtool make bzip2 bzip2-devel openssl-devel gmp-devel libstdc++ libcurl-devel \
     libusbx-devel python3 python3-devel python-devel libedit-devel doxygen \
-    graphviz clang patch
+    graphviz clang patch vim-common jq
```
arhag: You shouldn't need `jq` in the `release/1.8.x` version of this test doesn't use `jq`. But it is okay to keep it if you want for other reasons.

tests/release-build.sh
```diff
@@ -0,0 +1,64 @@
...
+if ! $(xxd --version 2>/dev/null); then
+    echo 'ERROR: Test requires xxd, but xxd was not found in your PATH!'
+    echo ''
+    echo 'The xxd hex dump tool can be installed as part of the vim-common package on most operating systems.'
+    exit 1
+fi
+# find nodeos
+[[ -z "$EOSIO_ROOT" && $(git --version) ]] && export EOSIO_ROOT="$(git rev-parse --show-toplevel)"
+[[ -z "$EOSIO_ROOT" ]] && export EOSIO_ROOT="$(echo $(pwd)/ | grep -ioe '.*/eos/' -e '.*/eosio/' -e '.*/build/' | sed 's,/build/,/,')"
+if [[ ! -f "$EOSIO_ROOT/build/bin/nodeos" && ! -f "$EOSIO_ROOT/build/programs/nodeos/nodeos" ]]; then
```
arhag: You know that this test file needs to be located within the `tests` subdirectory of the build directory. So you can just get the path of the source file and then find nodeos relative to it (`../programs/nodeos/nodes`) rather than dealing with git commands or assuming the build folder is named `build`.

kj4ezj:
> This is not checking for debug symbols. It's checking whether nodeos was built with -DNDEBUG.  

Thank you, I have updated the test purpose, description, and messages to accurately reflect what it is doing.

tests/release-build.sh
```diff
@@ -0,0 +1,64 @@
...
+    echo "$ ls -la \"$EOSIO_ROOT/build/bin\""
+    ls -la "$EOSIO_ROOT/build/bin"
+    echo "$ ls -la \"$EOSIO_ROOT/build/programs/nodeos\""
+    ls -la "$EOSIO_ROOT/build/programs/nodeos"
+    echo 'Release build test not run.'
+    exit 2
+fi
+[[ -f "$EOSIO_ROOT/build/bin/nodeos" ]] && cd "$EOSIO_ROOT/build/bin" || cd "$EOSIO_ROOT/build/programs/nodeos"
+# run nodeos to generate state files
+./nodeos --config-dir "$(pwd)/config" --data-dir "$(pwd)/data" 1>/dev/null 2>/dev/null &
```
arhag: I would prefer to see nodeos run with the `config` and `data` directories stored within another directory (that is clearly marked as intended for this test) within the initial current working directory of this script (which should be the build directory when this script is run by `ctest`). I'm a little concerned with potential directory conflicts between the nodeos directories this script uses and ones that may already exist for a user running ctest on their local machine.

tests/release-build.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+# test name and purpose
+echo ''
+echo '                         ##### Release Build Test #####'
+echo ''
+echo '    The purpose of this test is to ensure that nodeos was built with compiler'
```
arhag: To add to @swatanabe-b1's comment [here](https://github.com/EOSIO/eos/pull/7882#issuecomment-530543223),  this test isn't truly checking if nodeos was built with debug symbols nor is it checking which optimization level it was built with. It is, however, checking part of the build environment (specifically the debug flag) embedded in nodeos which determines whether nodeos is compatible with a particular Chainbase file it is attempting to open. Checking whether the released binaries are compatible with existing Chainbase files is probably the more important thing to check as part of our CICD process; the CICD sync tests do a good job of that but this test is intended to be a faster way of catching a class of problems that we know would lead to failures in our sync tests later on. However, due to correlation between the optimization setting and status of `NDEBUG`, this test also kind of acts as a check of the compiler optimization used (at least for the typical way the build would be configured with `CMAKE_BUILD_TYPE`).

PR #7886, opened 2019-09-11
promote read-only disablement log from net_plugin to warn level

spoonincode:
## Change Description
The message that net_plugin was disabling incoming connections was only info level. This seems a serious enough message to promote to warning level so it's easier to spot because some users have been caught by it. Plus, I've made it so if incoming connection are already disabled (`p2p-listen-endpoint` is empty) then there is no warning, so it's not necessarily going to be warning spam.

This does change the behavior so that instead of setting `p2p-max-nodes-per-host` to 0 the entire listening socket is disabled instead. I think that's fine as it's "more off".

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3311,6 +3311,13 @@ namespace eosio {
 
       my->dispatcher.reset( new dispatch_manager( my_impl->thread_pool->get_executor() ) );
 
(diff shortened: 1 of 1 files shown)


spoonincode:
yeah not sure why I mixed that up; I even tried it and didn't _think_ I saw it printed as info but yep it's there

PR #7919, opened 2019-09-15
decouple wavm runtime from being required & initial support for platform specific wasm runtimes

spoonincode:
## Change Description
eosio's code currently assumes that both wabt and wavm wasm runtimes will be built and available. This hasn't been much of a problem except for interfering with unsupported win32 builds where our wavm fork is non functional.

Looking forward to 2.0 though, there will be more diversity in what platforms different runtimes support. It's currently looking like:

|runtime|platforms operational on|
|-|-|
|wabt|anywhere|
|wavm|64-bit, POSIX (Linux & macOS)|
|EOS-VM Interpreter| POSIX (Linux & macOS)|
|EOS-VM JIT|x86-64, POSIX (Linux & macOS)|
|EOS-VM OC|x86-64, Linux only|

This PR adds some initial basic scaffolding to support central definition of what runtimes are supported. Frankly I don't care much for this approach as it just dumps some `#ifdefs` around which I'd rather not do. However there are aspects that make something cleaner more challenging: the way we've constructed the intrinsic registration, that we need parts of the wavm intrinsics even when not using wavm (because we use the wavm "linker" for validation), and some runtimes like EOS-VM OC have additional runtime options.

There is a functional issue where if you `unittests/unit_test -- --unsupported-runtime` all the tests simply fail instead of unit_test stopping early. But `ctest` is properly set up only register runtimes that are enabled.

If anyone wants to take a go at a better approach before 2.0, no beef here.

As part of the above effort, this PR also makes wavm's runtime enabled only on non-win32 builds. You can experiment with this on any platform by manually changing the `if()` in the cmakelists, and see that wavm is not allowed as a runtime nor does it show up in `ctest`

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -166,6 +168,12 @@ namespace eosio { namespace chain {
       const chainbase::database& db;
    };
 
+#define _ADD_PAREN_1(...) ((__VA_ARGS__)) _ADD_PAREN_2
+#define _ADD_PAREN_2(...) ((__VA_ARGS__)) _ADD_PAREN_1
+#define _ADD_PAREN_1_END
+#define _ADD_PAREN_2_END
+#define _WRAPPED_SEQ(SEQ) BOOST_PP_CAT(_ADD_PAREN_1 SEQ, _END)
+
```
swatanabe-b1: These are reserved identifiers.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -38,12 +38,14 @@ namespace eosio { namespace chain {
...
          if(vm == wasm_interface::vm_type::wavm)
             runtime_interface = std::make_unique<webassembly::wavm::wavm_runtime>();
-         else if(vm == wasm_interface::vm_type::wabt)
+#endif
+         if(vm == wasm_interface::vm_type::wabt)
             runtime_interface = std::make_unique<webassembly::wabt_runtime::wabt_runtime>();
-         else
-            EOS_THROW(wasm_exception, "wasm_interface_impl fall through");
+         if(!runtime_interface)
+            EOS_THROW(wasm_exception, "${r} wasm runtime not supported on this platform and/or configuration", ("r", vm));
```
swatanabe-b1: Do we want to distinguish between unsupported runtimes and totally non-existent runtimes in the error message?

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
spoonincode: not sure if @b1bart has any thoughts on why there are named that way

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
spoonincode: Unless I'm misunderstanding something, since this function works on an enum, I think the only way you get here is for an unsupported runtime. If you `--wasm banana` you'll fail much earlier when the string fails to be converted to an enum.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
b1bart: bad habit I suppose.  feel free to rename them and properly `undefine` them it is merely a nod to the idea that they are "internal".

PR #7935, opened 2019-09-17
Experimental/wb2 remove stack limit

larryk85:
## Change Description

Remove the hard limit of 8k stack elements from eos-vm, this PR also merged the latest develop and removed a flaky test that was failing every time with eos-vm because of it's fast initialization time.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


unittests/api_tests.cpp
```diff
@@ -921,66 +922,6 @@ BOOST_AUTO_TEST_CASE(checktime_fail_tests) { try {
    BOOST_REQUIRE_EQUAL( t.validate(), true );
 } FC_LOG_AND_RETHROW() }
 
-
-BOOST_FIXTURE_TEST_CASE(checktime_intrinsic, TESTER) { try {
```
swatanabe-b1: This should not be removed.  It's a necessary test case.

larryk85:
No longer needed.

PR #7947, opened 2019-09-18
Function body code size test

jeffreyssmith2nd:
## Change Description

- Adds a unit test to verify that a wasm is rejected when the function body size is too large.
- Updated several exceptions to be more specific.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/wasm_tests.cpp
+++ b/unittests/wasm_tests.cpp
@@ -1868,7 +1868,6 @@ BOOST_FIXTURE_TEST_CASE( code_size, TESTER )  try {
       code_function_body.insert(code_function_body.end(), function_body.begin(), function_body.end());
    }
 
-
    std::vector<U8> code;
    code.insert(code.end(), code_start.begin(), code_start.end());
    code.insert(code.end(), code_function_body.begin(), code_function_body.end());
--- a/unittests/wasm_tests.cpp
+++ b/unittests/wasm_tests.cpp
@@ -8,7 +8,10 @@
 #include <eosio/chain/wast_to_wasm.hpp>
 #include <eosio/testing/tester.hpp>
 
+#include <Inline/Serialization.h>
+#include <IR/Module.h>
 #include <Runtime/Runtime.h>
+#include <WASM/WASM.h>
 
 #include <boost/test/unit_test.hpp>
 #include <boost/algorithm/string/predicate.hpp>
@@ -883,7 +886,6 @@ BOOST_FIXTURE_TEST_CASE( lotso_globals, TESTER ) try {
    BOOST_CHECK_THROW(set_code(N(globals),
       string(ss.str() + "(global $z (mut i64) (i64.const -12)))")
    .c_str()), eosio::chain::wasm_execution_error);
-
 } FC_LOG_AND_RETHROW()
 
 BOOST_FIXTURE_TEST_CASE( offset_check, TESTER ) try {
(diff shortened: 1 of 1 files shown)


unittests/wasm_tests.cpp
```diff
@@ -1843,6 +1845,39 @@ BOOST_FIXTURE_TEST_CASE( depth_tests, TESTER ) try {
...
+   std::vector<U8> code_start = {
+      0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00, 0x01, 0x07, 0x01, 0x60,
+      0x03, 0x7e, 0x7e, 0x7e, 0x00, 0x03, 0x02, 0x01, 0x00, 0x07, 0x09, 0x01,
+      0x05, 0x61, 0x70, 0x70, 0x6c, 0x79, 0x00, 0x00, 0x0a, 0x8b, 0x80, 0x80,
+      0x0a, 0x01, 0x86, 0x80, 0x80, 0x0a, 0x00
+   };
+
+   std::vector<U8> function_body = {
+      0x41, 0xff, 0xff, 0xff, 0xff, 0x07, 0x1a
+   };
```
swatanabe-b1: It would be simpler (and easier to verify that the counts are correct) if you just filled the function body with nop (0x01).

unittests/wasm_tests.cpp
```diff
@@ -1843,6 +1845,39 @@ BOOST_FIXTURE_TEST_CASE( depth_tests, TESTER ) try {
 
 } FC_LOG_AND_RETHROW()
 
+// TODO: Update to use eos-vm once merged
+BOOST_FIXTURE_TEST_CASE( code_size, TESTER )  try {
```
swatanabe-b1: The TESTER is not used at all.

unittests/wasm_tests.cpp
```diff
@@ -1843,6 +1845,39 @@ BOOST_FIXTURE_TEST_CASE( depth_tests, TESTER ) try {
...
+   }
+
+   std::vector<U8> code;
+   code.insert(code.end(), code_start.begin(), code_start.end());
+   code.insert(code.end(), code_function_body.begin(), code_function_body.end());
+   code.insert(code.end(), code_end.begin(), code_end.end());
+
+   Module module;
+   Serialization::MemoryInputStream stream((const U8*)code.data(), code.size());
+   BOOST_CHECK_THROW(WASM::serialize(stream, module), FatalSerializationException);
```
swatanabe-b1: I'm not really fond of testing internal components that are not directly visible to users.  We don't really care about exactly where the restriction is enforced as long as the user-visible behavior is fixed.

unittests/wasm_tests.cpp
```diff
@@ -1843,6 +1845,39 @@ BOOST_FIXTURE_TEST_CASE( depth_tests, TESTER ) try {
 
 } FC_LOG_AND_RETHROW()
 
+// TODO: Update to use eos-vm once merged
```
swatanabe-b1: I don't want to have to update tests, just because the implementation changes.

swatanabe-b1:
> > The code size here is wrong. It's still 0x1400006, but the correct value is now 0x14000002. The size of the code segment also needs to be adjusted.
> 
> I'm not sure I understand, since I'm just using the constraint constant.

The original loop emitted more bytes than the constant.  It emitted 0x1400004 bytes, because that is the next larger multiple of 7.  The new version emits exactly 0x1400000 bytes, which is smaller by 4.

jeffreyssmith2nd:
> > > The code size here is wrong. It's still 0x1400006, but the correct value is now 0x14000002. The size of the code segment also needs to be adjusted.
> > 
> > 
> > I'm not sure I understand, since I'm just using the constraint constant.
> 
> The original loop emitted more bytes than the constant. It emitted 0x1400004 bytes, because that is the next larger multiple of 7. The new version emits exactly 0x1400000 bytes, which is smaller by 4.

Ah I see now. It was misleadingly passing, because an exception was being thrown elsewhere in the code. Changing it to `maximum_code_size` + 4 hits the appropriate exception.

PR #7956, opened 2019-09-20
Allow p2p incomming connections when in read only mode

elmato:
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -99,6 +99,7 @@ namespace eosio {
       uint32_t                         max_client_count = 0;
       uint32_t                         max_nodes_per_host = 1;
       uint32_t                         num_clients = 0;
+      bool                             allow_incomming = false;
 
       vector<string>                   supplied_peers;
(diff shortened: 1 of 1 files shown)


heifner:
Can you provide your use case for enabling this behavior. The point of a read only node is so it can feed plugins not to do any validation on transactions/blocks.

elmato:
The use case is to offer a read-only peering mode where clients can sync to the chain but they are not allowed to to push transactions using that channel.

This is useful for some sidechains that wants to let user get in sync with the chain, but they have to push transactions through over another channel (UI / RPC)

heifner:
> The use case is to offer a read-only peering mode where clients can sync to the chain but they are not allowed to to push transactions using that channel.
> 
> This is useful for some sidechains that wants to let user get in sync with the chain, but they have to push transactions through over another channel (UI / RPC)

Can't they just not load the `chain_api_plugin` ?

swatanabe-b1:
Can we close this?  It should be resolved by https://github.com/EOSIO/eos/pull/8743, right?

PR #7962, opened 2019-09-21
set immutable chain_id during construction of controller

arhag:
## Change Description

Fixes a bug introduced in EOSIO/eos#7841 in which plugins like net_plugin were incorrectly using the default chain ID.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -559,8 +559,8 @@ struct controller_impl {
    void startup(std::function<bool()> shutdown, const genesis_state& genesis) {
       EOS_ASSERT( db.revision() < 1, database_exception, "This version of controller::startup only works with a fresh state database." );
       const auto& genesis_chain_id = genesis.compute_chain_id();
-      EOS_ASSERT( genesis_chain_id == chain_id, misc_exception,
-                  "genesis state provided to startup does not match the chain ID that controller was constructed with",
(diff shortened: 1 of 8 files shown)


libraries/chain/controller.cpp
```diff
@@ -546,6 +558,11 @@ struct controller_impl {
 
    void startup(std::function<bool()> shutdown, const genesis_state& genesis) {
       EOS_ASSERT( db.revision() < 1, database_exception, "This version of controller::startup only works with a fresh state database." );
+      const auto& genesis_chain_id = genesis.compute_chain_id();
+      EOS_ASSERT( genesis_chain_id == chain_id, misc_exception,
```
heifner: There is a `chain_id_type_exception` which might be a better fit. Here and below.

libraries/chain/controller.cpp
```diff
@@ -546,6 +558,11 @@ struct controller_impl {
...
+                  "genesis state provided to startup does not match the chain ID that controller was constructed with",
+                  ("genesis_chain_id", genesis_chain_id)("controller_chain_id", chain_id)
```
heifner: I think we should get out of the habit of using captured log messages without format specifiers in the string.

PR #7967, opened 2019-09-22
support unix socket HTTP server for nodeos

spoonincode:
## Change Description
When unix socket support for the http plugin was initially added it was implicitly disabled for nodeos because it was only available to configure if the program activating http plugin had a default unix socket path set. (keosd had a default, nodeos doesn't)

Some users have requested the ability to use unix socket for nodeos because they have something in front of nodeos (like haproxy) and desire the, at least on paper, lower overhead of a unix socket.

This PR removes the check thus allowing any application using http plugin (like nodeos) to enable the unix socket by configuring it appropriately. It still is off by default for nodeos.

Enabling it for nodeos is as simple as a `unix-socket-path = nodeos.sock` for example. And to test you can `curl --unix-socket $EOSIO_DATA_DIR/nodeos.sock http://x/v1/chain/get_info`

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -392,6 +392,10 @@ namespace eosio {
          cfg.add_options()
             ("unix-socket-path", bpo::value<string>()->default_value(current_http_plugin_defaults.default_unix_socket_path),
              "The filename (relative to data-dir) to create a unix socket for HTTP RPC; set blank to disable.");
+      else
+         cfg.add_options()
+            ("unix-socket-path", bpo::value<string>(),
+             "The filename (relative to data-dir) to create a unix socket for HTTP RPC; set blank to disable.");   
 #endif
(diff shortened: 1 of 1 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -392,6 +392,10 @@ namespace eosio {
          cfg.add_options()
             ("unix-socket-path", bpo::value<string>()->default_value(current_http_plugin_defaults.default_unix_socket_path),
              "The filename (relative to data-dir) to create a unix socket for HTTP RPC; set blank to disable.");
+      else
```
spoonincode: unfortunately I had to do the else here because `default_value()` balked at an empty string 🤷‍♂

PR #7968, opened 2019-09-23
cmake policy CMP0074

an-ivannikov:
$ cmake --help-policy CMP0074
CMP0074
-------

``find_package()`` uses ``<PackageName>_ROOT`` variables.

In CMake 3.12 and above the ``find_package(<PackageName>)`` command now
searches prefixes specified by the ``<PackageName>_ROOT`` CMake
variable and the ``<PackageName>_ROOT`` environment variable.
Package roots are maintained as a stack so nested calls to all ``find_*``
commands inside find modules also search the roots as prefixes.  This policy
provides compatibility with projects that have not been updated to avoid using
``<PackageName>_ROOT`` variables for other purposes.

The ``OLD`` behavior for this policy is to ignore ``<PackageName>_ROOT``
variables.  The ``NEW`` behavior for this policy is to use
``<PackageName>_ROOT`` variables.

This policy was introduced in CMake version 3.12.  CMake version
3.15.2 warns when the policy is not set and uses ``OLD`` behavior.
Use the ``cmake_policy()`` command to set it to ``OLD`` or ``NEW``
explicitly.

.. note::
  The ``OLD`` behavior of a policy is
  ``deprecated by definition``
  and may be removed in a future version of CMake.

$ cmake --version
cmake version 3.15.2

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,5 @@
 cmake_minimum_required( VERSION 3.8 )
+cmake_policy( SET CMP0074 NEW )
 
 project( EOSIO )
 include(CTest) # suppresses DartConfiguration.tcl error


spoonincode:
Hello. PRs need to be based off the `develop` branch. But, can you describe what this is resolving? When using cmake 3.15.3 on Linux I am not seeing any unexpected warnings from cmake. If you're seeing warnings it may be due to something that needs to be fixed in the build script, not in the cmake file. Can you describe the problem and environment more?

spoonincode:
Closing due to no input from opener.

PR #7973, opened 2019-09-23
Add a unit test for the write order for aliased intrinsic arguments.

swatanabe-b1:
## Change Description

This only affects lower/upper bound for secondary keys, as they are the only intrinsics with multiple output parameters.  The order depends on the alignment of the arguments.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
>This only affects lower/upper bound for secondary keys, as they are the only intrinsics with multiple output parameters

Isn't anything that takes a reference, like get_resource_limits(), also at risk? We should probably have a test that covers reference types.

swatanabe-b1:
> 
> Isn't anything that takes multiple references, like get_resource_limits(), also at risk? 

Yes.  I thought I checked all the intrinsics, but I missed that somehow.

swatanabe-b1:
I can't find any tests for get_resource_limits or the rest of the privileged_api.

PR #7975, opened 2019-09-24
EOS-VM Optimized Compiler

spoonincode:
## Change Description
EOS-VM Optimized Compiler is a high performance WASM runtime designed for operation inside of nodeos. It is a tier up runtime that works in conjunction with the configured baseline runtime (such as wabt or EOS-VM JIT). When enabled, actions on a contract are initially dispatched to the baseline runtime while EOS-VM Optimized Compiler does an optimized compilation in the background. Once this compilation is complete, actions on that contract will then be run with the optimized compiled code from then on. This optimized compile can take a handful of seconds but remember it is performed in the background so it doesn’t cause any hiccups. Optimized compilations are also saved to a new data file so that when restarting nodeos there is no need to recompile previously compiled contracts.

### Usage
The option `--eos-vm-oc-enable` (or `eos-vm-oc-enable = true` in the config file) enables EOS-VM Optimized Compiler. There is no need to replay an existing chain. You can restart nodeos with or without this option at any time in the future without conflicts in the code cache file. The number of simultaneous background compiles is configured with the option `eos-vm-oc-compile-threads` which defaults to 1. More compile threads means EOS-VM Optimized Compiler can tier up more contracts simultaneously.

There is also an option `eos-vm-oc-cache-size-mb` that controls the maximum size of the cache. The default of 1GB is more than enough for any public EOSIO chain at the moment. Be wary of shrinking this too small as it can cause cache thrashing.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/webassembly/eos-vm-oc.cpp
```diff
@@ -0,0 +1,53 @@
...
+eosvmoc_runtime::~eosvmoc_runtime() {
+}
+
+std::unique_ptr<wasm_instantiated_module_interface> eosvmoc_runtime::instantiate_module(const char* code_bytes, size_t code_size, std::vector<uint8_t> initial_memory,
+                                                                                     const digest_type& code_hash, const uint8_t& vm_type, const uint8_t& vm_version) {
+
+   return std::make_unique<eosvmoc_instantiated_module>(code_hash, vm_type, *this);
+}
+
+void eosvmoc_runtime::immediately_exit_currently_running_module() {}
```
swatanabe-b1: This is unreachable, right?

libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
```diff
@@ -0,0 +1,1303 @@
...
+		// Load/store operators
+		//
+
+		#define EMIT_LOAD_OP(valueTypeId,name,llvmMemoryType,naturalAlignmentLog2,conversionOp) \
+			void valueTypeId##_##name(LoadOrStoreImm<naturalAlignmentLog2> imm) \
+			{ \
+				auto byteIndex = pop(); \
+				auto pointer = coerceByteIndexToPointer(byteIndex,imm.offset,llvmMemoryType); \
+				auto load = irBuilder.CreateLoad(pointer); \
+				load->setAlignment(1<<imm.alignmentLog2); \
```
swatanabe-b1: llvm load docs: "Overestimating the alignment results in undefined behavior"

WebAssembly Spec: "The alignment memarg.align in load and store instructions does
not affect the semantics... Unaligned access violating that property is still
allowed and must succeed regardless of the annotation"

libraries/chain/webassembly/eos-vm-oc/LLVMJIT.h
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include "Inline/BasicTypes.h"
+#include "IR/Module.h"
+
+#pragma push_macro("N")
+#undef N
```
swatanabe-b1: Nothing in this file uses ``N``, you can just #undef it.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -0,0 +1,251 @@
...
+         siglongjmp(*cb_in_main_segment->jmp, EOSVMOC_EXIT_CHECKTIME_FAIL);
+
+   //was the segfault within data?
+   if((uintptr_t)info->si_addr >= cb_in_main_segment->execution_thread_memory_start &&
+      (uintptr_t)info->si_addr < cb_in_main_segment->execution_thread_memory_start+cb_in_main_segment->execution_thread_memory_length)
+         siglongjmp(*cb_in_main_segment->jmp, EOSVMOC_EXIT_SEGV);
+
+notus:
+   if(chained_handler)
+      chained_handler(sig, info, ctx);
```
swatanabe-b1: This shouldn't fall through.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -0,0 +1,251 @@
...
+   __builtin_unreachable();
+}
+static intrinsic eosio_exit_intrinsic("env.eosio_exit", IR::FunctionType::get(IR::ResultType::none,{IR::ValueType::i32}), (void*)&eosio_exit,
+  boost::hana::index_if(intrinsic_table, ::boost::hana::equal.to(BOOST_HANA_STRING("env.eosio_exit"))).value()
+);
+
+static void throw_internal_exception(const std::string& s) {
+   EOSVMOC_MEMORY_PTR_cb_ptr;
+   *cb_ptr->eptr = std::make_exception_ptr(wasm_execution_error(FC_LOG_MESSAGE(error, s)));
+   siglongjmp(*cb_ptr->jmp, EOSVMOC_EXIT_EXCEPTION);
```
swatanabe-b1: This leaks s.  It may or may not actually leak memory depending on how large std::string's SBO buffer is.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -0,0 +1,251 @@
...
+      printf("full_linear_memory_start %li\n", -memory::cb_offset + offsetof(eosio::chain::eosvmoc::control_block, full_linear_memory_start));
+      printf("cb_offset %lu\n", memory::cb_offset);
+      printf("remaining call depth: %li\n", -memory::cb_offset + offsetof(eosio::chain::eosvmoc::control_block, current_call_depth_remaining));
+      printf("first intrinsic: %lu\n", memory::first_intrinsic_offset);
+      printf("current code offset: %li\n", -memory::cb_offset + offsetof(eosio::chain::eosvmoc::control_block, running_code_base));
+      once_is_enough = true;
+   }
+
+   //if we're the first executor created, go setup the signal handling. For now we'll just leave this attached forever
+   if(std::lock_guard g(inited_signal_mutex); inited_signal == false) {
```
swatanabe-b1: local static initialization is thread-safe.  You can just create a dummy variable instead of using a mutex+bool like this.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -0,0 +1,251 @@
...
+   struct stat s;
+   fstat(cc.fd(), &s);
+   code_mapping = (uint8_t*)mmap(nullptr, s.st_size, PROT_EXEC|PROT_READ, MAP_SHARED, cc.fd(), 0);
+   code_mapping_size = s.st_size;
+   mapping_is_executable = true;
+}
+
+void executor::execute(const code_descriptor& code, const memory& mem, apply_context& context) {
+   if(mapping_is_executable == false) {
+      mprotect(code_mapping, code_mapping_size, PROT_EXEC|PROT_READ);
```
swatanabe-b1: mprotect can fail for various reasons.

libraries/chain/webassembly/eos-vm-oc/ipc_helpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+   write_message_with_fds(s.native_handle(), message, fds);
+}
+
+void write_message_with_fds(int fd_to_send_to, const eosvmoc_message& message, const std::vector<wrapped_fd>& fds) {
+   struct msghdr msg = {};
+   struct cmsghdr* cmsg;
+
+   size_t sz = fc::raw::pack_size(message);
+   EOS_ASSERT(sz < max_message_size, misc_exception, "trying to send too large of message");
+   char buff[sz];
```
swatanabe-b1: Variable length arrays are a C99 feature which is not standard C++.

programs/nodeos/main.cpp
```diff
@@ -80,6 +82,7 @@ enum return_codes {
 
 int main(int argc, char** argv)
 {
+   eosio::chain::eosvmoc::start_compile_monitor();
```
swatanabe-b1: This is causing an undefined symbol on OSX.

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
```diff
@@ -0,0 +1,305 @@
...
+   fstat(fd, &st); //XXX
+   return st.st_size;
+}
+
+static void copy_memfd_contents_to_pointer(void* dst, int fd) {
+   struct stat st;
+   fstat(fd, &st); //XXX
+   void* contents = mmap(nullptr, st.st_size, PROT_READ, MAP_SHARED, fd, 0); //XXX
+   memcpy(dst, contents, st.st_size);
+   munmap(contents, st.st_size);
```
swatanabe-b1: why not just ``read``?

programs/nodeos/main.cpp
spoonincode: Right, this is just slop added in to get it work. See item #3 in review comments

libraries/chain/webassembly/eos-vm-oc.cpp
spoonincode: Yes. Needs at least a comment, but deserving of any more form of decoration?

libraries/chain/webassembly/eos-vm-oc/LLVMJIT.h
spoonincode: llvm's headers do
Just commenting out the `#undef N` gives me errors like

```
In file included from /home/spoon/eos/src/libraries/chain/webassembly/eos-vm-oc/LLVMJIT.h:8:
In file included from /usr/include/llvm/IR/Module.h:27:
In file included from /usr/include/llvm/IR/Function.h:28:
In file included from /usr/include/llvm/IR/BasicBlock.h:23:
In file included from /usr/include/llvm/IR/Instruction.h:22:
In file included from /usr/include/llvm/IR/DebugLoc.h:18:
In file included from /usr/include/llvm/IR/TrackingMDRef.h:17:
/usr/include/llvm/IR/Metadata.h:1227:48: error: member initializer 'string_to_name' does not name a non-static data member or base class
  MDTupleTypedArrayWrapper(const MDTuple *N) : N(N) {}
                                               ^~~~
/home/spoon/eos/src/libraries/chain/include/eosio/chain/name.hpp:89:28: note: expanded from macro 'N'
#define N(X) eosio::chain::string_to_name(#X)
```

swatanabe-b1: The #undef is definitely needed.  It's the push and pop that are unnecessary.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -166,6 +205,11 @@ namespace eosio { namespace chain {
       wasm_cache_index wasm_instantiation_cache;
 
       const chainbase::database& db;
+      const wasm_interface::vm_type wasm_runtime_time;
```
swatanabe-b1: s/_time/_type/

libraries/chain/webassembly/eos-vm-oc/code_cache.cpp
```diff
@@ -0,0 +1,376 @@
...
+   wait_on_compile_monitor_message();
+
+   _monitor_reply_thread = std::thread([this]() {
+      fc::set_thread_name("oc-monitor");
+      _ctx.run();
+   });
+}
+
+code_cache_async::~code_cache_async() {
+   _compile_monitor_write_socket.shutdown(local::datagram_protocol::socket::shutdown_send);
```
swatanabe-b1: I don't think it's safe to access the socket from multiple threads without synchronization.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -0,0 +1,251 @@
...
+   uint64_t current_gs;
+   syscall(SYS_arch_prctl, ARCH_GET_GS, &current_gs);
+   if(current_gs == 0)
+      goto notus;
+
+   cb_in_main_segment = reinterpret_cast<control_block*>(current_gs - memory::cb_offset);
+
+   //as a double check that the control block pointer is what we expect, look for the magic
+   if(cb_in_main_segment->magic != signal_sentinel)
+      goto notus;
```
swatanabe-b1: If this were to trigger, there's no guarantee that cb_in_main_segment points to valid memory.

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+         //either way, we are done
+         _ctx.post([this, current_compile_it]() {
+            current_compiles.erase(current_compile_it);
+         });
+      });
+
+   }
+
+   boost::signals2::signal<void()> connection_dead_signal;
```
swatanabe-b1: A signal seems like overkill.  There's always exactly one.

libraries/chain/webassembly/eos-vm-oc/ipc_helpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+   int wrote;
+   do {
+      wrote = sendmsg(fd_to_send_to, &msg, 0);
+   } while(wrote == -1 && errno == EINTR);
+}
+
+wrapped_fd memfd_for_vector(const std::vector<uint8_t>& bytes) {
+   int fd = syscall(SYS_memfd_create, "eosvmoc_code", MFD_CLOEXEC);
+   ftruncate(fd, bytes.size());
+   uint8_t* b = (uint8_t*)mmap(nullptr, bytes.size(), PROT_READ|PROT_WRITE, MAP_SHARED, fd, 0);
```
swatanabe-b1: XXX

libraries/chain/webassembly/eos-vm-oc/ipc_helpers.cpp
```diff
@@ -0,0 +1,131 @@
...
+   memcpy(b, bytes.data(), bytes.size());
+   munmap(b, bytes.size());
+   return wrapped_fd(fd);
+}
+
+wrapped_fd memfd_for_blob(const shared_blob& blob) {
+   int fd = syscall(SYS_memfd_create, "eosvmoc_code", MFD_CLOEXEC);
+   ftruncate(fd, blob.size());
+   uint8_t* b = (uint8_t*)mmap(nullptr, blob.size(), PROT_READ|PROT_WRITE, MAP_SHARED, fd, 0);
```
swatanabe-b1: XXX MAP_FAILED

libraries/chain/webassembly/eos-vm-oc/memory.cpp
```diff
@@ -0,0 +1,46 @@
...
+
+memory::memory() {
+   int fd = syscall(SYS_memfd_create, "eosvmoc_mem", MFD_CLOEXEC);
+   FC_ASSERT(fd >= 0, "Failed to create memory memfd");
+   int ret = ftruncate(fd, wasm_memory_size+memory_prologue_size);
+   FC_ASSERT(!ret, "Failed to grow memory memfd");
+
+   mapsize = total_memory_per_slice*number_slices;
+   mapbase = (uint8_t*)mmap(nullptr, mapsize, PROT_NONE, MAP_PRIVATE|MAP_ANON, 0, 0);
+   FC_ASSERT(mapbase != MAP_FAILED, "Failed to mmap memory");
```
swatanabe-b1: This and the other FC_ASSERTs do not unmap the memory that was already mapped.

swatanabe-b1: Okay, you haven't mapped anything, yet here, but it does leak fd.

libraries/chain/webassembly/eos-vm-oc/intrinsic.cpp
```diff
@@ -0,0 +1,19 @@
...
+   static intrinsic_map_t intrinsic_map;
+   return intrinsic_map;
+}
+
+const intrinsic_map_t& get_intrinsic_map() {
+   return the_intrinsic_map();
+}
+
+intrinsic::intrinsic(const char* n, const IR::FunctionType* t, void* f, size_t o) {
+   the_intrinsic_map().erase(n);
```
swatanabe-b1: Is it possible for ``n`` to already exist in the map?

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -0,0 +1,251 @@
...
+static std::mutex inited_signal_mutex;
+static bool inited_signal;
+
+static void(*chained_handler)(int,siginfo_t*,void*);
+[[noreturn]] static void segv_handler(int sig, siginfo_t* info, void* ctx)  {
+   control_block* cb_in_main_segment;
+
+   //a 0 GS value is an indicator an executor hasn't been active on this thread recently
+   uint64_t current_gs;
+   syscall(SYS_arch_prctl, ARCH_GET_GS, &current_gs);
```
swatanabe-b1: What is the rationale for using syscall over the glibc wrapper?

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
```diff
@@ -0,0 +1,309 @@
...
+wrapped_fd get_connection_to_compile_monitor(int cache_fd) {
+   FC_ASSERT(the_compile_monitor_trampoline.compile_manager_fd >= 0, "EOS-VM oop connection doesn't look active");
+
+   int socks[2]; //0: our socket to compile_manager_session, 1: socket we'll give to compile_maanger_session
+   socketpair(AF_UNIX, SOCK_SEQPACKET | SOCK_CLOEXEC, 0, socks);
+   wrapped_fd socket_to_monitor_session(socks[0]);
+
+   std::vector<wrapped_fd> fds_to_pass; 
+   fds_to_pass.emplace_back(socks[1]);
+   fds_to_pass.emplace_back(cache_fd);  //put cache_fd in to a wrapped_fd just temporarily, ick
```
swatanabe-b1: This is especially bad, because it will close cache_fd if write_message_with_fds throws, but not if any other part of the function throws.

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+wrapped_fd get_connection_to_compile_monitor(int cache_fd) {
+   FC_ASSERT(the_compile_monitor_trampoline.compile_manager_fd >= 0, "EOS-VM oop connection doesn't look active");
+
+   int socks[2]; //0: our socket to compile_manager_session, 1: socket we'll give to compile_maanger_session
+   socketpair(AF_UNIX, SOCK_SEQPACKET | SOCK_CLOEXEC, 0, socks);
+   wrapped_fd socket_to_monitor_session(socks[0]);
+
+   std::vector<wrapped_fd> fds_to_pass; 
+   fds_to_pass.emplace_back(socks[1]);
```
swatanabe-b1: This can leak socks[1].

libraries/chain/webassembly/eos-vm-oc/code_cache.cpp
```diff
@@ -0,0 +1,376 @@
...
+   }
+
+   code_cache_header* cache_header = nullptr;
+   {
+      char header[total_header_size];
+      std::ifstream hs(_cache_file_path.generic_string(), std::ifstream::binary);
+      hs.read(header, sizeof(header));
+      EOS_ASSERT(!hs.fail(), bad_database_version_exception, "failed to read code cache header");
+
+      cache_header = reinterpret_cast<code_cache_header*>(header+header_offset);
```
swatanabe-b1: I don't think this is guaranteed to be correctly aligned.

swatanabe-b1: Also, when header goes out of scope, cache_header is left dangling.

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
spoonincode: Nope, there are some unittests that create two controllers ergo two wasm_runtimes ergo two connections

libraries/chain/webassembly/eos-vm-oc/code_cache.cpp
```diff
@@ -0,0 +1,376 @@
...
+
+   bfs::create_directories(data_dir);
+
+   if(!bfs::exists(_cache_file_path)) {
+      std::ofstream ofs(_cache_file_path.generic_string(), std::ofstream::trunc);
+      EOS_ASSERT(ofs.good(), database_exception, "unable to create EOS-VM Optimized Compiler code cache");
+      bfs::resize_file(_cache_file_path, eosvmoc_config.cache_size);
+      bip::file_mapping creation_mapping(_cache_file_path.generic_string().c_str(), bip::read_write);
+      bip::mapped_region creation_region(creation_mapping, bip::read_write);
+      new (creation_region.get_address()) allocator_t(eosvmoc_config.cache_size, total_header_size);
```
swatanabe-b1: The boost::interprocess::rb_tree_best_fit constructor requires enough memory for the header and at least one block.  Is cache_size validated anywhere?

libraries/chain/webassembly/eos-vm-oc/intrinsic.cpp
spoonincode: It is. First all intrinsics are registered and then the internals of EOS-VM OC come along and overwrite eosio_exit

libraries/chain/webassembly/eos-vm-oc/code_cache.cpp
```diff
@@ -0,0 +1,376 @@
...
+   //however, if it's currently being compiled there is no way to cancel the compile,
+   //so instead set a poison boolean that indicates not to insert the code in to the cache
+   //once the compile is complete
+   const std::unordered_map<code_tuple, bool>::iterator compiling_it = _outstanding_compiles_and_poison.find({code_id, vm_version});
+   if(compiling_it != _outstanding_compiles_and_poison.end())
+      compiling_it->second = true;
+}
+
+void code_cache_base::check_eviction_threshold(size_t free_bytes) {
+   if(free_bytes < _free_bytes_eviction_threshold) {
```
swatanabe-b1: What if the cache is fragmented due to many small wasms?  Can we get in a situation where nothing is ever evicted because we have enough free bytes in total, but the wasms we want to execute don't fit in any free segment?

spoonincode: Maybe on a toofull message from the compile process, cache eviction should be more aggressive and disregard the threshold

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
swatanabe-b1: Sure there can be more than one controller, but the connection_dead_signal is only used to erase the compile_monitor_session that it is part of.  ``connect`` is only called once, right after the compile_monitor_session is emplaced into compile_monitor::_compile_sessions.

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
```diff
@@ -0,0 +1,309 @@
...
+   }
+
+   void kick_compile_off(const code_tuple& code_id, wrapped_fd&& wasm_code) {
+      //prepare a requst to go out to the trampoline
+      int socks[2];
+      socketpair(AF_UNIX, SOCK_SEQPACKET | SOCK_CLOEXEC, 0, socks);
+      local::datagram_protocol::socket response_socket(_ctx);
+      response_socket.assign(local::datagram_protocol(), socks[0]);
+      std::vector<wrapped_fd> fds_pass_to_trampoline;
+      fds_pass_to_trampoline.emplace_back(socks[1]);
```
swatanabe-b1: May leak socks[1].

libraries/chain/webassembly/eos-vm-oc/executor.cpp
spoonincode: Likewise for anything else that accesses cb_in_main_segment in here. Are you suggesting that it's a worthless check and might as well be removed?

libraries/chain/webassembly/eos-vm-oc/executor.cpp
spoonincode: That the docs don't allow `arch_prctl` & `mprotect` from a signal handling context. Yeah the docs don't allow `syscall` either but it's even less likely to do be touching some state.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
swatanabe-b1: I presumed that the intent of this check is to determine whether gs is ours.  Thus, the later checks should be safe provided this passes.  The check here is not entirely worthless, but it isn't very reliable either.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/eos-vm-oc.h
```diff
@@ -0,0 +1,38 @@
...
+   size_t execution_thread_code_length;
+   uintptr_t execution_thread_memory_start;
+   size_t execution_thread_memory_length;
+#ifdef __cplusplus
+   eosio::chain::apply_context* ctx;
+   std::exception_ptr* eptr;
+#else
+   void* ctx;
+   void* eptr;
+#endif
```
swatanabe-b1: I think it's better not to have the struct use different types for C vs. C++.  You can add a typed (non-member) accessor if needed.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc.hpp
```diff
@@ -0,0 +1,794 @@
...
+   constexpr int cb_full_linear_memory_start_segment_offset = OFFSET_OF_CONTROL_BLOCK_MEMBER(full_linear_memory_start);
+   constexpr int cb_first_invalid_memory_address_segment_offset = OFFSET_OF_CONTROL_BLOCK_MEMBER(first_invalid_memory_address);
+
+   char dumpster;
+   uint64_t scratch;
+
+   asm volatile("mov %%gs:(%[Ptr]), %[Dumpster]\n"                   //probe memory location at ptr to see if valid
+                "mov %%gs:%c[firstInvalidMemory], %[Scratch]\n"      //get first invalid memory address
+                "cmpb $0, %%gs:-1(%[Scratch])\n"                     //is last byte in valid linear memory 0?
+                "jle 2f\n"                                           //if so, this will be a null terminated string one way or another
```
swatanabe-b1: Did you mean je? I thought jle is a signed comparison.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc.hpp
```diff
@@ -0,0 +1,794 @@
...
+{
+   constexpr int cb_full_linear_memory_start_segment_offset = OFFSET_OF_CONTROL_BLOCK_MEMBER(full_linear_memory_start);
+   constexpr int cb_first_invalid_memory_address_segment_offset = OFFSET_OF_CONTROL_BLOCK_MEMBER(first_invalid_memory_address);
+
+   static_assert(sizeof(T) == 64 || sizeof(T) == 32 || sizeof(T) == 20 || sizeof(T) == 16 || sizeof(T) == 8 ||
+                 sizeof(T) == 4 || sizeof(T) == 2 || sizeof(T) == 1, "array_ptr sizeof(T) only supports limited values");
+
+   if constexpr(sizeof(T) == 20) {
+      asm("shl $2, %[Length]\n"
+          "lea (%[Length], %[Length], 4), %[Length]\n"
```
swatanabe-b1: Is this really better then imul?  We're not running i386 here.  Actually, why is this being done in inline asm at all?

libraries/chain/webassembly/eos-vm-oc/intrinsic.cpp
swatanabe-b1: Oh, so that's why you're setting the static initialization order.

swatanabe-b1: That could use some explanation.

swatanabe-b1:
I've finished reviewing modulo whatever changes you make after this.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc.hpp
```diff
@@ -0,0 +1,794 @@
...
+                : [Ptr] "+r" (ptr),
+                  [Dumpster] "=r" (dumpster),
+                  [Scratch] "=r" (scratch)
+                : [linearMemoryStart] "i" (cb_full_linear_memory_start_segment_offset),
+                  [firstInvalidMemory] "i" (cb_first_invalid_memory_address_segment_offset)
+                : "cc"
+               );
+
+   return null_terminated_ptr((char*)ptr);
+   __builtin_unreachable();
```
larryk85: Is this `__builtin_unreachable` redundant?

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc.hpp
spoonincode: that's the code clang generated for me for comparative c++. but agree, this probably shouldn't be asm.  it is holdover from an early impl that inspected overflow flags and such. the less inline asm the better

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/gs_seg_helpers.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include <stdint.h>
+#include <setjmp.h>
+#include <string.h>
+
+#include <eosio/chain/webassembly/eos-vm-oc/eos-vm-oc.h>
+
+#ifdef __clang__
+   #define GS_PTR __attribute__((address_space(256)))
```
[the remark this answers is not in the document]
spoonincode: I'm pretty sure GS segment is completely unused by Linux userland bits: TLS, SSP, all on FS. Arch actually has SSP enabled by default for gcc & clang and I haven't seen any problems; the asm I've seen in functions def uses FS.

libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
```diff
@@ -0,0 +1,1318 @@
...
+namespace LLVMJIT
+{
+	static std::string getExternalFunctionName(Uptr functionDefIndex)
+	{
+		return "wasmFunc" + std::to_string(functionDefIndex);
+	}
+
+	bool getFunctionIndexFromExternalName(const char* externalName,Uptr& outFunctionDefIndex)
+	{
+		const char wasmFuncPrefix[] = "wasmFunc";
```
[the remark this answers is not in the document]
spoonincode: This is a good point, but I will go ahead and mention that specifically for LLVMEmitIR.cpp I resisted making any more changes than I had to at this point (even left it all tabs). This facilitates doing a diff between EOS-VM OC's LLVMEmitIR.cpp and WAVM's LLVMEmitIR.cpp

libraries/chain/webassembly/eos-vm-oc/executor.cpp
swatanabe-b1: Setting errno is an issue, so I suppose I can live with this in the signal handler.  You're using syscall in other places, too, though.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc.hpp
```diff
@@ -0,0 +1,818 @@
...
+
+using namespace eosio::chain::eosvmoc;
+
+class eosvmoc_instantiated_module;
+
+class eosvmoc_runtime : public eosio::chain::wasm_runtime_interface {
+   public:
+      eosvmoc_runtime(const boost::filesystem::path data_dir, const eosvmoc::config& eosvmoc_config, const chainbase::database& db);
+      ~eosvmoc_runtime();
+      bool inject_module(IR::Module&) override { return false; }
```
swatanabe-b1: Doesn't this need injection?

spoonincode: Injection is done in run_compile(), this avoids the post-injection/re-injection function size differences

libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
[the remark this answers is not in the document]
larryk85: Yeah, I noticed that a lot of this code was the same as WAVM's after adding this comment.  I thought that I had removed it for now and added to a list of changes we should do for the WAVM portion of this at some point.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/gs_seg_helpers.h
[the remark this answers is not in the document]
larryk85:
```c++
/* SPDX-License-Identifier: GPL-2.0 */
/*
 * GCC stack protector support.
 *
 * Stack protector works by putting predefined pattern at the start of
 * the stack frame and verifying that it hasn't been overwritten when
 * returning from the function.  The pattern is called stack canary
 * and unfortunately gcc requires it to be at a fixed offset from %gs.
 * On x86_64, the offset is 40 bytes and on x86_32 20 bytes.  x86_64
 * and x86_32 use segment registers differently and thus handles this
 * requirement differently.
 *
 * On x86_64, %gs is shared by percpu area and stack canary.  All
 * percpu symbols are zero based and %gs points to the base of percpu
 * area.  The first occupant of the percpu area is always
 * fixed_percpu_data which contains stack_canary at offset 40.  Userland
 * %gs is always saved and restored on kernel entry and exit using
 * swapgs, so stack protector doesn't add any complexity there.
 *
 * On x86_32, it's slightly more complicated.  As in x86_64, %gs is
 * used for userland TLS.  Unfortunately, some processors are much
 * slower at loading segment registers with different value when
 * entering and leaving the kernel, so the kernel uses %fs for percpu
 * area and manages %gs lazily so that %gs is switched only when
 * necessary, usually during task switch.
 *
 * As gcc requires the stack canary at %gs:20, %gs can't be managed
 * lazily if stack protector is enabled, so the kernel saves and
 * restores userland %gs on kernel entry and exit.  This behavior is
 * controlled by CONFIG_X86_32_LAZY_GS and accessors are defined in
 * system.h to hide the details.
 */
```
Also, Clang `ShadowCallStack` uses the gs register for this same kind of purpose.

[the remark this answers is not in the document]
spoonincode: Your quoted blob is from kernel sources. Kernel has its own GS value; it even says so in that blob
> Userland %gs is always saved and restored on kernel entry and exit using swapgs

If you compile with `-fstack-protector-strong` you'll see code like
```
00000000000011f0 <fooz>:
fooz():
    11f0:       41 54                   push   %r12
    11f2:       48 89 fe                mov    %rdi,%rsi
    11f5:       48 83 ec 50             sub    $0x50,%rsp
    11f9:       64 48 8b 04 25 28 00    mov    %fs:0x28,%rax
...
    122c:       48 8b 44 24 48          mov    0x48(%rsp),%rax
    1231:       64 48 33 04 25 28 00    xor    %fs:0x28,%rax
    1238:       00 00 
    123a:       75 07                   jne    1243 <fooz+0x53>
    123c:       48 83 c4 50             add    $0x50,%rsp
    1240:       41 5c                   pop    %r12
    1242:       c3                      retq   
    1243:       e8 08 fe ff ff          callq  1050 <__stack_chk_fail@plt>
```
it's def using FS.

Having trouble finding applicably of ShadowCallStack, docs say it's ARM8 only
http://releases.llvm.org/9.0.0/tools/clang/docs/ShadowCallStack.html

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
spoonincode: ah okay, I dunno that whole thing is just a pattern I typically use for asio server + asio clients so that I avoid shared_ptrs and client objects are completely decoupled

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
spoonincode: this is N/A after merging from develop

libraries/chain/webassembly/eos-vm-oc/LLVMJIT.h
spoonincode: This is one of those cases where I prefer to leave things how I found them. Maybe some day it'll save someone a few minutes trying to figure out where `N()` is after LLVMJIT.h somehow gets included in their source file 3 layers deep

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/memory.hpp
```diff
@@ -0,0 +1,67 @@
...
+      static constexpr uint64_t wasm_memory_size                  = eosio::chain::wasm_constraints::maximum_linear_memory;
+      static constexpr uint64_t intrinsic_count                   = boost::hana::length(intrinsic_table);
+      //warning: changing the following 3 params will invalidate existing PIC
+      static constexpr uint64_t mutable_global_size               = 8u  * eosio::chain::wasm_constraints::maximum_mutable_globals/4u;
+      static constexpr uint64_t table_size                        = 16u * eosio::chain::wasm_constraints::maximum_table_elements;
+      static constexpr size_t   wcb_allowance                     = 512u;
+      static_assert(sizeof(control_block) <= wcb_allowance, "EOS-VM OC memory doesn't set aside enough memory for control block");
+
+      //round up the prologue to multiple of 4K page
+      static constexpr uint64_t memory_prologue_size = ((memory::wcb_allowance + mutable_global_size + table_size + intrinsic_count*UINT64_C(8))+UINT64_C(4095))/UINT64_C(4096)*UINT64_C(4096);
```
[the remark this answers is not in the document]
spoonincode: unfortunately querying _SC_PAGESIZE won't allow me to fold this in to a constexpr. is there anything else similar in a #define you're aware of?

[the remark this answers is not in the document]
larryk85: Right, sadly not.  We could add a constexpr variable in eos-vm in the next release and in some initializer place an assert that the two values match.  It won't give us any compile time protection, but it would at least ensure that the page size is correct with the system.

PR #7991, opened 2019-09-26
missing block log chain id unit tests

brianjohnson5972:
## Change Description
#7988 
Added test cases for restarting chain with existing state but with and without an existing block_log

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/restart_chain_tests.cpp
+++ b/unittests/restart_chain_tests.cpp
@@ -1,0 +1,59 @@
+#include <sstream>
+
+#include <eosio/chain/block_log.hpp>
+#include <eosio/chain/global_property_object.hpp>
+#include <eosio/chain/snapshot.hpp>
+#include <eosio/testing/tester.hpp>
+
+#include <boost/mpl/list.hpp>
+#include <boost/test/unit_test.hpp>
+
+#include <contracts.hpp>
+#include <snapshots.hpp>
+
+using namespace eosio;
+using namespace testing;
+using namespace chain;
+
+void remove_existing_blocks(controller::config& config) {
+   auto block_log_path = config.blocks_dir / "blocks.log";
+   remove(block_log_path);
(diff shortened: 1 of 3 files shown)


unittests/restart_chain_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+   tester other;
+   for (auto new_block : blocks) {
+      other.push_block(new_block);
+   }
+   blocks.clear();
+
+   other.close();
+   genesis_state genesis;
+   genesis.initial_timestamp = fc::time_point::from_iso_string("2020-01-01T00:00:00.999");
```
arhag: Better to use "2020-01-01T00:00:01.000" so that the time lands on a valid slot time.

unittests/restart_chain_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+   remove(block_index_path);
+}
+
+BOOST_AUTO_TEST_SUITE(restart_chain_tests)
+
+BOOST_AUTO_TEST_CASE(test_existing_state_without_block_log)
+{
+   tester chain;
+
+   chain.create_account(N(snapshot));
```
arhag: No need to create this account.

unittests/restart_chain_tests.cpp
```diff
@@ -0,0 +1,83 @@
...
+   for (auto new_block : blocks) {
+      other.push_block(new_block);
+   }
+}
+
+BOOST_AUTO_TEST_CASE(test_restart_with_different_chain_id)
+{
+   tester chain;
+
+   chain.create_account(N(snapshot));
```
arhag: No need to create this account.

Issue #7939, opened 2019-09-17
Remove genesis_state from block log header for trimmed block log


brianjohnson5972:
When a block log has version 3 or newer, and the block log is trimmed from starting at block 1, the header needs to be re-written and the chain_id should be written out instead.

Completes #7646

PR #8012, opened 2019-09-30
7939 trim block log v3 support

brianjohnson5972:
## Change Description
#7939 
- Added support in eosio-blocklog to trim the front of a blocklog and create the version 3 blocklog format
- Added invariant logic for existing version 3 blocklog logic
- cleanup

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/block_log.hpp
```diff
@@ -78,11 +78,40 @@ namespace eosio { namespace chain {
...
+      uint32_t version = 0;                              //blocklog version
+      uint32_t first_block = 0;                          //first block in blocks.log
+      uint32_t last_block = 0;                          //last block in blocks.log
+      FILE* blk_in = nullptr;                            //C style files for reading blocks.log and blocks.index
+      FILE* ind_in = nullptr;                            //C style files for reading blocks.log and blocks.index
+      //we use low level file IO because it is distinctly faster than C++ filebuf or iostream
+      uint64_t index_pos = 0;                            //filepos in blocks.index for block n, +8 for block n+1
+      uint64_t fpos0 = 0;                                //filepos in blocks.log for block n and block n+1
+      uint64_t fpos1 = 0;                                //filepos in blocks.log for block n and block n+1
+      uint64_t first_block_pos = 0;                      //file position in blocks.log for the first block in the log
```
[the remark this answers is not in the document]
brianjohnson5972: I'll refactor to only track first block num, first block position, and last block num and have an accessor to retrieve a specific blocks position.

libraries/chain/block_log.cpp
```diff
@@ -906,4 +953,230 @@ namespace eosio { namespace chain {
...
+      trim_data original_block_log(block_dir);
+      if (truncate_at_block <= original_block_log.first_block) {
+         ilog("There are no blocks before block ${n} so do nothing.", ("n", truncate_at_block));
+         return false;
+      }
+      if (truncate_at_block > original_block_log.last_block) {
+         ilog("All blocks are before block ${n} so do nothing (trim front would delete entire blocks.log).", ("n", truncate_at_block));
+         return false;
+      }
+      original_block_log.find_block_pos(truncate_at_block);
```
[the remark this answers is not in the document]
brianjohnson5972: find_block_pos was the original code that setup fpos0 and fpos1 (the start of the next block) and index_pos, I added block_pos to just return the location of the block, so this method will go away with the refactor.

libraries/chain/block_log.cpp
```diff
@@ -157,15 +157,17 @@ namespace eosio { namespace chain {
          explicit fileptr_datastream( FILE* file, const std::string& filename ) : _file(file), _filename(filename) {}
 
          void skip( size_t s ) {
-            std::vector<char> d( s );
-            read( &d[0], s );
+            auto status = fseek(_file, s, SEEK_SET);
```
arhag: Should be `SEEK_CUR`.

libraries/chain/block_log.cpp
```diff
@@ -1133,50 +1136,44 @@ namespace eosio { namespace chain {
          fclose(ind_in);
    }
 
+   uint64_t trim_data::block_index(uint32_t n) const {
+      using namespace std;
+      EOS_ASSERT( n <= last_block, block_log_exception,
```
arhag: Should also have a check to make sure `first_block <= n`.

libraries/chain/block_log.cpp
```diff
@@ -1133,50 +1136,44 @@ namespace eosio { namespace chain {
...
+   }
+
    uint64_t trim_data::block_pos(uint32_t n) {
       using namespace std;
-      index_pos = sizeof(uint64_t) * (n - first_block);
+      // can indicate the location of the block after the last block
+      if (n == last_block + 1) {
+         return ftell(blk_in);
+      }
+      const uint64_t index_pos = sizeof(uint64_t) * (n - first_block);
```
arhag: Why not use `block_index(n)` instead?

libraries/chain/block_log.cpp
```diff
@@ -1133,50 +1136,44 @@ namespace eosio { namespace chain {
...
-
-      if (n != last_block)
-         cout << " starts at position " << fpos1 << '\n';
-      else
-         cout << " is past end\n";
-
-      //read blocks.log and verify block number n is found at file position fpos0
-      auto status = fseek(blk_in, fpos0 + blknum_offset, SEEK_SET);
-      EOS_ASSERT( status == 0, block_log_exception, "cannot seek to ${file} ${pos} from beginning of file", ("file", block_file_name.string())("pos", fpos0 + blknum_offset) );
+      //read blocks.log and verify block number n is found at the determined file position
```
arhag: Why is any of the rest of this necessary? Isn't it safe to assume that block.index hasn't been corrupted?

libraries/chain/block_log.cpp
```diff
@@ -1023,7 +1028,7 @@ namespace eosio { namespace chain {
          }
 
          // read in the previous contiguous memory into the read buffer
```
arhag: Isn't the code within this for loop going to write each buffer in the wrong order? It seems like it would be fine if `to_write <= detail::reverse_iterator::_buf_len`. But if it is larger, then the later sections in the original block log would be written out in the `new_block_file` before the earlier sections. Or am I missing something?

programs/eosio-blocklog/main.cpp
```diff
@@ -262,7 +262,7 @@ void smoke_test(bfs::path block_dir) {
    for (uint32_t n = td.first_block; ; n += delta) {
       if (n > td.last_block)
          n = td.last_block;
-      td.find_block_pos(n);                                 //check block 'n' is where blocks.index says
+      td.block_pos(n);                                 //check block 'n' is where blocks.index says
```
arhag: Is this the only reason why the extra code at the end of `block_pos` is needed? It seems like that extra code could just go here.

libraries/chain/block_log.cpp
```diff
@@ -547,6 +588,12 @@ namespace eosio { namespace chain {
 
          new_block_stream << chain_id;
       }
+      else {
+         EOS_ASSERT( false, block_log_exception,
```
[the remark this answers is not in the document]
arhag: The other one (which I missed in the initial review) was changed in the recent commits, but this one still remains.

libraries/chain/block_log.cpp
brianjohnson5972: because I created the method and then forgot to swap out the code here :(

libraries/chain/block_log.cpp
brianjohnson5972: I was just keeping the same logic as before.  If I remove it here, I need to add it to smoke_test in eosio-blocklog.  This may not be the most useful validation, but not sure that we really have a scenario where we can say that we have a reason to trust the index is correct.  Without this check the only validation we are doing is that the index position falls in a valid range in the block log file.  My vote is to leave this so that we are at least doing this and revisit index validation when we actually take a step back and think about what we want to be doing in this code.

arhag: Sure, that's fine. We can keep it here.

libraries/chain/block_log.cpp
brianjohnson5972: read_buffer = to_write

brianjohnson5972: also, wasn't writing to the new file in the correct location, fixed that and the mode the new cfile was opened in.

PR #8014, opened 2019-10-01
Retain persisted trx until expired on speculative nodes

heifner:
## Change Description

- Keep persisted trx until expired on speculative nodes. Retains 1.8.x behavior.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
We explicitly decided to go with the 2.0 behavior of not re-applying the speculative api node transaction since it will be naturally handled by abort/fork/in-coming-block behavior.

heifner:
abort/forked un-applied transactions are not re-applied on non-producing nodes, so this PR is needed to maintain 1.8.x behavior.

PR #8028, opened 2019-10-03
Make subjective_cpu_leeway a config option - 1.7

heifner:
## Change Description

- Add `config.ini` and command line option for `subjective_cpu_leeway`
- Already was configurable via `producer_api_plugin` `update_runtime_options`
- Changed default from 3ms to 31ms

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions
- New producer_plugin option:
         `subjective-cpu-leeway-us` defaults to 31000us
          "Time in microseconds allowed for a transaction that starts with insufficient CPU quota to complete and cover its CPU usage."

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -708,6 +708,7 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
    my->_max_scheduled_transaction_time_per_block_ms = options.at("max-scheduled-transaction-time-per-block-ms").as<int32_t>();
 
    if( options.at( "subjective-cpu-leeway-us" ).as<int32_t>() != config::default_subjective_cpu_leeway_us ) {
+      chain::controller& chain = my->chain_plug->chain();
       chain.set_subjective_cpu_leeway( fc::microseconds( options.at( "subjective-cpu-leeway-us" ).as<int32_t>() ) );
    }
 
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -573,6 +573,8 @@ void producer_plugin::set_program_options(
           "offset of last block producing time in microseconds. Negative number results in blocks to go out sooner, and positive number results in blocks to go out later")
          ("max-scheduled-transaction-time-per-block-ms", boost::program_options::value<int32_t>()->default_value(100),
           "Maximum wall-clock time, in milliseconds, spent retiring scheduled transactions in any block before returning to normal transaction processing.")
+         ("subjective-cpu-leeway-us", boost::program_options::value<int32_t>()->default_value( config::default_subjective_cpu_leeway_us ),
```
larryk85: You could use `bpo` instead of `boost::program_options`.  Also, shouldn't this be `uint32_t` and not `int32_t`?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -705,6 +707,10 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
 
    my->_max_scheduled_transaction_time_per_block_ms = options.at("max-scheduled-transaction-time-per-block-ms").as<int32_t>();
 
+   if( options.at( "subjective-cpu-leeway-us" ).as<int32_t>() != config::default_subjective_cpu_leeway_us ) {
```
larryk85: Same as above (`uint32_t` vs. `int32_t`).

plugins/producer_plugin/producer_plugin.cpp
heifner: `bpo` <- I just copied the one above it.

It is `int32_t` because the existing `producer_plugin` `runtime_options` is 
```
fc::optional<int32_t> subjective_cpu_leeway_us;
```

plugins/producer_plugin/producer_plugin.cpp
heifner: See above.

PR #8044, opened 2019-10-09
Add greylist limit - develop

arhag:
## Change Description

This PR adds support for a greylist that applies to all accounts and makes the accounts' resource limits behave (as far as deciding on whether to subjectively reject the transaction) as if the elastic resource multiplier (which can vary from 1 to 1000 according to dynamics that are part of consensus) is limited to a locally configured ceiling. The value for this limit can be specified through the new `greylist-limit` configuration option in the producer plugin. It can also be updated during runtime by specifying the new optional field `greylist_limit` within the `runtime_options` provided to the existing `update_runtime_options` RPC of the producer API.

Consider a transaction that is billed to an account that is not on any explicit greylist and that hits a timeout deadline (or NET limit) driven by the quotas of the billed account (rather than other limits). When this greylist limit is not the limiting factor (e.g. the `greylist-limit` is set to 500 but the network is used to the point where the multiplier is at 200), the system will continue to throw the usual exception: `tx_cpu_usage_exceeded` (for hitting the CPU limit) or `tx_net_usage_exceeded` (for hitting the NET limit). But when the greylist limit is the limiting factor, the greylist versions of the exceptions will be thrown instead: `greylist_cpu_usage_exceeded` or `greylist_net_usage_exceeded`. In the case of hitting NET limits, this can be the difference between the subjective exception `greylist_net_usage_exceeded` (which would, for example, cause a deferred transaction to be ignored and not retired) and the objective exception `tx_net_usage_exceeded` (which could cause a deferred transaction to retire with `soft_fail` or `hard_fail`).

This effectively acts as a subjective means of capping the maximum resource multiplier to a value lower than 1000. It is not possible to change that maximum value of 1000 in an objective way without a change to the consensus protocol rules, which this PR does not attempt to do. 

This PR also:

- uses the new `eosio::chain::config::maximum_elastic_resource_multiplier` constant rather than hardcoding the number 1000 in various places through the code;
- improves the `resource_limits_test/sanity_check` test;
- and, adds the `whitelist_blacklist_tests/greylist_limit_tests` test which tests the new greylist feature.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

Added new configuration option `greylist-limit` to the producer plugin which is an integer ranging from 1 to 1000 that determines the limit for the new greylist feature. The maximum value of 1000 (which is also the default) means that greylist is not the limiting factor and therefore basically has no effect.

Added a new optional integer field `greylist_limit` to the `options` argument of the `update_runtime_options` RPC of the producer API which allows changing the greylist limit during runtime.

## Documentation Additions
- [ ] Documentation Additions


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -646,6 +646,8 @@ void producer_plugin::set_program_options(
           "Limits the maximum time (in milliseconds) that is allowed for sending blocks to a keosd provider for signing")
          ("greylist-account", boost::program_options::value<vector<string>>()->composing()->multitoken(),
           "account that can not access to extended CPU/NET virtual resources")
+         ("global-greylist-limit", boost::program_options::value<uint32_t>()->default_value(1000),
+          "Limit (between 1 and 1000) on the multiple that CPU/NET virtual resources can extend during low congestion (only enforced subjectively; use 1000 to not enforce any limit)")
          ("produce-time-offset-us", boost::program_options::value<int32_t>()->default_value(0),
```
heifner: To me, `global` doesn't provide anything useful. I think `greylist-limit` is as descriptive.

heifner: `usage` instead of `congestion` I think is more descriptive.

libraries/chain/controller.cpp
```diff
@@ -3131,6 +3131,20 @@ void controller::set_subjective_cpu_leeway(fc::microseconds leeway) {
...
+               "Invalid limit (${limit}) passed into set_global_greylist_limit. "
+               "Must be between 1 and ${max}.",
+               ("limit", limit)("max", chain::config::maximum_elastic_resource_multiplier)
+   );
+   my->global_greylist_limit = limit;
+}
+
+uint32_t controller::get_global_greylist_limit()const {
+   return my->global_greylist_limit;
+}
```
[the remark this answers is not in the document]
arhag: We agreed to instead store it in `controller::config` and keep the exception as `misc_exception`.

PR #8046, opened 2019-10-09
#7658: modified code to handle new db_runtime_exception

ndcgundlach:
## Change Description

Added handling for new db_runtime_exception from chainbase.  In the case of incompatible db, exit with error code 2.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


programs/nodeos/main.cpp
```diff
@@ -112,9 +112,21 @@ int main(int argc, char** argv)
...
+            return DATABASE_DIRTY;
+         }
+      } catch (...) { } 
+
+      elog( "${e}", ("e",e.what()));
+      return OTHER_FAIL;
    } catch( const fc::exception& e ) {
       if( e.code() == fc::std_exception_code ) {
-         if( e.top_message().find( "database dirty flag set" ) != std::string::npos ) {
+         if( e.top_message().find( "Database dirty flag set" ) != std::string::npos ) {
```
heifner: You should add a catch of std::system_error and check error code instead of searching for string.

programs/nodeos/main.cpp
```diff
@@ -127,13 +139,8 @@ int main(int argc, char** argv)
...
-   } catch( const std::runtime_error& e ) {
-      if( std::string(e.what()).find("database dirty flag set") != std::string::npos ) {
-         elog( "database dirty flag set (likely due to unclean shutdown): replay required" );
-         return DATABASE_DIRTY;
-      } else {
-         elog( "${e}", ("e",e.what()));
-      }
+   }  catch( const std::runtime_error& e ) {
+      elog( "${e}", ("e",e.what()));
       return OTHER_FAIL;
```
heifner: This is the same as std::exception catch and therefore can be removed.

programs/nodeos/main.cpp
```diff
@@ -112,9 +112,21 @@ int main(int argc, char** argv)
       return FIXED_REVERSIBLE;
```
heifner: Need to refresh to latest fc and chainbase once those PRs are merged.

programs/nodeos/main.cpp
```diff
@@ -112,9 +112,21 @@ int main(int argc, char** argv)
...
+      try {
+         std::rethrow_exception(e.get_inner_exception());
+      } catch (const std::system_error&i) {
+         if (chainbase::db_error_code::dirty == i.code().value()) {
+            elog("Database dirty flag set (likely due to unclean shutdown): replay required" );
+            return DATABASE_DIRTY;
+         }
+      } catch (...) { } 
+
+      elog( "${e}", ("e",e.what()));
```
heifner: Use `to_detail_string()` instead of `what()`

programs/nodeos/main.cpp
ndcgundlach: This is a relic at this point and should be removed.

tests/validate-dirty-db.py
```diff
@@ -103,7 +103,7 @@ def runNodeosAndGetOutput(myTimeout=3):
         expectedRetCode=2
         if retCode != expectedRetCode:
             errorExit("Expected return code to be %d, but instead received %d." % (expectedRetCode, retCode))
-        db_dirty_msg="database dirty flag set"
+        db_dirty_msg="atabase dirty flag set"
```
brianjohnson5972: This looks like an accidental deletion

ndcgundlach: It might look like it, but it isn't. 😳 I would actually like to get rid of all the display text checks and stick with something more programatic like return codes, etc.

programs/eosio-blocklog/main.cpp
```diff
@@ -81,6 +81,7 @@ void blocklog::read_log() {
 
    optional<chainbase::database> reversible_blocks;
    try {
+      elog("opening reversible db");
```
heifner: This should be an `ilog` not an `elog`

PR #8118, opened 2019-10-22
remove gettext/libintl dependency

spoonincode:
## Change Description

gettext is only used for cleos localization support. But, cleos has not had any effort to translate it nor is that effort on any short term roadmap AFAIK. Remove this dependency for now as it's an LGPL dep and causing some grief. When we get back around to localizing cleos we can reevaluate this need; it's possible something like boost::locale would be a better fit.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
AFAICT .cicd didn't need updating -- for macos it didn't specify installing `gettext` because it was already getting that package indirectly through another package (possibly `graphviz` as it requires `gettext`). For Linux we never needed to install anything as these functions were provided by libc. Did you see something in .cicd that needs fix up?

PR #8147, opened 2019-10-30
Optimize push_transaction

heifner:
## Change Description

- Optimize `push_transaction` by using deque instead of vector to avoid vector having to reallocate and move transaction receipts and max transaction metas in block.
- `std::deque` used for boost < 1.71 now since configurable `boost::container::deque` is not available until boost 1.71. boost deque used if boost >= 1.71 available.
- Calculate `action_receipt` digest at action execution time to avoid copying of `action_receipts` and to bill for hash.
- Calculate `transaction_receipt` digest at transaction execution time to bill for hash.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -118,15 +118,16 @@ struct building_block {
    optional<producer_authority_schedule> _new_pending_producer_schedule;
    vector<digest_type>                   _new_protocol_feature_activations;
    size_t                                _num_new_protocol_features_that_have_activated = 0;
-   vector<transaction_metadata_ptr>      _pending_trx_metas;
-   vector<transaction_receipt>           _pending_trx_receipts;
-   vector<digest_type>                   _action_receipt_digests;
+   deque<transaction_metadata_ptr>       _pending_trx_metas;
+   deque<transaction_receipt>            _pending_trx_receipts; // boost deque in 1.71 with 1024 elements performs better
+   deque<digest_type>                    _pending_trx_receipt_digests;
+   deque<digest_type>                    _action_receipt_digests;
 };
 
 struct assembled_block {
    block_id_type                     _id;
    pending_block_header_state        _pending_block_header_state;
-   vector<transaction_metadata_ptr>  _trx_metas;
+   deque<transaction_metadata_ptr>   _trx_metas;
    signed_block_ptr                  _unsigned_block;
 
    // if the _unsigned_block pre-dates block-signing authorities this may be present.
@@ -161,7 +162,7 @@ struct pending_state {
       return _block_stage.get<assembled_block>()._pending_block_header_state;
    }
 
-   const vector<transaction_receipt>& get_trx_receipts()const {
+   const deque<transaction_receipt>& get_trx_receipts()const {
       if( _block_stage.contains<building_block>() )
          return _block_stage.get<building_block>()._pending_trx_receipts;
 
@@ -171,7 +172,7 @@ struct pending_state {
       return _block_stage.get<completed_block>()._block_state->block->transactions;
    }
 
-   vector<transaction_metadata_ptr> extract_trx_metas() {
+   deque<transaction_metadata_ptr> extract_trx_metas() {
       if( _block_stage.contains<building_block>() )
          return std::move( _block_stage.get<building_block>()._pending_trx_metas );
 
@@ -1372,6 +1373,7 @@ struct controller_impl {
       r.cpu_usage_us         = cpu_usage_us;
       r.net_usage_words      = net_usage_words;
       r.status               = status;
+      pending->_block_stage.get<building_block>()._pending_trx_receipt_digests.emplace_back( r.digest() );
       return r;
    }
 
@@ -1523,11 +1525,6 @@ struct controller_impl {
 
       const auto& gpo = self.get_global_properties();
 
-      const auto size = gpo.configuration.max_block_cpu_usage /
-            ( gpo.configuration.min_transaction_cpu_usage ? gpo.configuration.min_transaction_cpu_usage : 1 );
-      bb._pending_trx_receipts.reserve( size );
-      bb._pending_trx_metas.reserve( size );
-
       // modify state of speculative block only if we are in speculative read mode (otherwise we need clean state for head or read-only modes)
       if ( read_mode == db_read_mode::SPECULATIVE || pending->_block_status != controller::block_status::incomplete )
       {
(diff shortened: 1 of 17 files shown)


libraries/chain/controller.cpp
```diff
@@ -2104,13 +2114,7 @@ struct controller_impl {
...
 
    checksum256_type calculate_action_merkle() {
-      vector<digest_type> action_digests;
-      const auto& actions = pending->_block_stage.get<building_block>()._actions;
-      action_digests.reserve( actions.size() );
-      for( const auto& a : actions )
-         action_digests.emplace_back( a.digest() );
-
-      return merkle( move(action_digests) );
+      return merkle( move(pending->_block_stage.get<building_block>()._action_receipt_digests) );
```
swatanabe-b1: It seems risky to clobber _action_receipt_digests here.

heifner: It is a bit risky. Could add a `clear()` after which would make it a bit less risky. Or could create a version of` merkle` that takes by `const&` instead of by value.

libraries/chain/controller.cpp
```diff
@@ -1518,6 +1521,13 @@ struct controller_impl {
       auto& bb = pending->_block_stage.get<building_block>();
       const auto& pbhs = bb._pending_block_header_state;
 
+      const auto& gpo = self.get_global_properties();
+
+      const auto size = gpo.configuration.max_block_cpu_usage /
```
arhag: This could potentially be very large. How about putting a cap on it, like 50000?

libraries/chain/controller.cpp
arhag: Let's just remove `calculate_action_merkle` and call `merkle( std::move( bb._action_receipt_digests ) )` from `finalize_block`.

heifner: Nice, I like that, we are doing other moves there.

libraries/chain/controller.cpp
heifner: Lets try a deque and see how that goes.

libraries/chain/controller.cpp
heifner: Moved the `move` to finalize block as suggested by Arhag.

libraries/chain/controller.cpp
heifner: `std::deque` is not nearly as good `boost::container::queue` configured to 1024 elements per block. However, the configurable boost queue is not available until boost 1.71. Going with std::queue for now as it is an improvement and we can switch to boost queue later.

libraries/chain/controller.cpp
heifner: Removed the `calculate_*_merkle` methods and call `merkle` directly in `finalize_block`.

swatanabe-b1:
> * `std::queue` used for now since configurable `boost::container::queue` is not available until boost 1.71 and don't want to require minimum boost version of 1.71 at this time. boost queue configured with 1024 elements per block was much better performing than std::queue.

Is it worthwhile to define an alias for deque and/or #if BOOST_VERSION >= 107100 to make it easy to update when we do require Boost 1.71?

heifner:
> Is it worthwhile to define an alias for deque and/or #if BOOST_VERSION >= 107100 to make it easy to update when we do require Boost 1.71?

I thought about that and figured it was a bit much, but with some of the changes I already made I don't think it would be too bad now. I'll give it a shot.

libraries/chain/controller.cpp
```diff
@@ -1518,6 +1523,8 @@ struct controller_impl {
       auto& bb = pending->_block_stage.get<building_block>();
       const auto& pbhs = bb._pending_block_header_state;
 
+      const auto& gpo = self.get_global_properties();
+
```
swatanabe-b1: Is there a reason for moving this declaration earlier?  The only new use is still in scope from the old location.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -82,6 +81,16 @@ namespace eosio { namespace chain {
    using private_key_type = fc::crypto::private_key;
    using signature_type   = fc::crypto::signature;
 
+#if BOOST_VERSION >= 107100
```
swatanabe-b1: Defined in boost/version.hpp.

libraries/chain/controller.cpp
heifner: No idea why, fixed.

libraries/chain/include/eosio/chain/types.hpp
heifner: Added include and include of `<boost/container/deque.hpp>`

libraries/chain/controller.cpp
```diff
@@ -118,15 +118,16 @@ struct building_block {
    optional<producer_authority_schedule> _new_pending_producer_schedule;
    vector<digest_type>                   _new_protocol_feature_activations;
    size_t                                _num_new_protocol_features_that_have_activated = 0;
-   vector<transaction_metadata_ptr>      _pending_trx_metas;
-   vector<transaction_receipt>           _pending_trx_receipts;
-   vector<digest_type>                   _action_receipt_digests;
+   deque<transaction_metadata_ptr>       _pending_trx_metas;
+   deque<transaction_receipt>            _pending_trx_receipts; // boost deque in 1.71 with 1024 elements performs better
+   deque<digest_type>                    _pending_trx_receipt_digests;
```
arhag: Make sure to reset the size of this new container in `make_block_restore_point`.

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -324,6 +324,24 @@ namespace impl {
          mvo(name, std::move(array));
       }
 
+      /**
+       * template which overloads add for deques of types which contain ABI information in their trees
+       * for these members we call ::add in order to trigger further processing
+       */
+      template<typename M, typename Resolver, require_abi_t<M> = 1>
+      static void add( mutable_variant_object &mvo, const char* name, const deque<M>& v, Resolver resolver, abi_traverse_context& ctx )
```
arhag: Also add the `extract` for the `deque`.

libraries/chain/controller.cpp
heifner: Added.

libraries/chain/include/eosio/chain/abi_serializer.hpp
heifner: Added

PR #8151, opened 2019-10-31
Debian Package: Make sure root is owner/group when building dpkg.

pnx:
## Change Description

Just passing `--root-owner-group` to `dpkg` in `scripts/generate_deb.sh`. Without this flag files in the .deb files are owned by the user running the script. 

The official .deb files from this repo has root as owner group. so this change should just make it easier to create the same packages as the official ones.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/generate_deb.sh
+++ b/scripts/generate_deb.sh
@@ -53,6 +53,6 @@ export SSUBPREFIX
 . ./generate_tarball.sh ${NAME}
 echo "Unpacking tarball: ${NAME}.tar.gz..."
 tar -xzvf ${NAME}.tar.gz -C ${PROJECT} || exit 1
-dpkg-deb --build ${PROJECT} || exit 1
+dpkg-deb --build --root-owner-group ${PROJECT} || exit 1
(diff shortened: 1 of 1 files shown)


NorseGaud:
Regarding the Travis failure: Mac tests failed for some memory reasons. Likely a transient problem, so I'm restarting the job.

pnx:
> Regarding the Travis failure: Mac tests failed for some memory reasons. Likely a transient problem, so I'm restarting the job.
> 
> Update after second run: Failed again with what looks like the same problem.

Just to be sure. this should not be related to this PR right?

spoonincode:
@pnx I merged this but we're going to have to back it out since it appears 16.04 does not support `--root-owner-group`. If you can come up with something that works on both we can reconsider it; I think it's a good fix.

pnx:
> @pnx I merged this but we're going to have to back it out since it appears 16.04 does not support `--root-owner-group`. If you can come up with something that works on both we can reconsider it; I think it's a good fix.

@spoonincode See #8408

PR #8165, opened 2019-11-04
[Develop] Prevent buildkite clone to speedup pipeline

NorseGaud:
## Change Description

Problem: Several steps don't need to do a clone locally on the host as they do the clone inside of the VM (like on mac). Others don't need to obtain the submodules as we only need to execute a script under the .cicd directory (test-metrics for example).

Solution: Use the skip-checkout plugin to stop buildkite from cloning and doing a submodule update on the host and just skip right to running commands.

+ We also added DISABLE_CONCURRENCY ENV so we can prevent concurrency on scheduled builds.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


NorseGaud:
Original Develop branch - 
Screen Shot 2019-11-06 at 9 28 43 AM

New changes -
Screen Shot 2019-11-06 at 9 28 52 AM

Issue #8066, opened 2019-10-11
Add simple test for State History Plugin


brianjohnson5972:
Add an integration test that sets up a nodeos with a state history plugin and runs this simple javascript example:
https://github.com/EOSIO/eos/blob/state-history-docs/docs/state-history-plugin/js-example.md

heifner:
Resolved by #8198

PR #8198, opened 2019-11-07
State History Plugin Integration Test

brianjohnson5972:
## Change Description
Added an integration test that starts a node instance with a javascript client that interacts with a nodeos with the state history plugin.
#8066 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/Node.py
```diff
@@ -1581,3 +1581,40 @@ def modifyBuiltinPFSubjRestrictions(self, nodeId, featureCodename, subjectiveRes
...
+            for configLine in configLines:
+                match = re.match(r'\s*p2p-server-address\s*=\s*([^\:\s]+)\:(\d+)', configLine)
+                if match:
+                    host = match.group(1)
+                    port = match.group(2)
+                    response = self.disconnectP2P(host, port)
+                    Utils.Print("Attempted disconnect from %s:%s.  Response: %s" % (host, port, json.dumps(response, indent=2)))
+                    assert not exitOnError or "connection removed" in response, Print("ERROR: Tried to disconnect from host: %s, port: %s.  Response was: %s" % (host, port, response))
+                    found = True
+                    break
```
jgiszczak: This loop seems quite fragile.  Can we at least document which topology this is expected to work with?

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
+#!/usr/bin/env python3
+
+from testUtils import Utils
+import testUtils
```
jgiszczak: Is this import necessary?  I don't see any use of anything from testUtils that isn't in `class Utils`, which is already imported.

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+        filePair[0].close()
+        filePair[1].close()
+        index += 1
+
+    Print("Shutdown state_history_plugin nodeos")
+    shipNode.kill(signal.SIGTERM)
+
+    files = None
+
+    maxFirstBN = None
```
jgiszczak: `maxFirstBN = -1`
`minLastBN = sys.maxsize`

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+            except json.decoder.JSONDecodeError as er:
+                Utils.errorExit("javascript client output was malformed in %s. Exception: %s" % (shipClientErrorFile, er))
+
+            for status in statuses:
+                statusDesc = status["status"]
+                if statusDesc == "done":
+                    done = True
+                    firstBlockNum = status["first_block_num"]
+                    lastBlockNum = status["last_block_num"]
+                    if maxFirstBN is None:
```
jgiszczak: No need for the conditional with the above initializations.  Initial values above will never be seen or used since an `error` status will result in `errorExit()` or `assert done` with `done` still equal to `False`.

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+            match = re.search(r'info\s+([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{3})\s.+Received\sblock\s+.+\s#([0-9]+)\s@\s([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{3})',
+                              line)
+            if match:
+                rcvTimeStr = match.group(1)
+                prodTimeStr = match.group(3)
+                blockNum = int(match.group(2))
+                if blockNum > maxFirstBN:
+                    # ship requests can only affect time after clients started
+                    timeFmt = '%Y-%m-%dT%H:%M:%S.%f'
+                    rcvTime = datetime.strptime(rcvTimeStr, '%Y-%m-%dT%H:%M:%S.%f')
```
jgiszczak: Use `timeFmt` here as well.

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+                    break
+                if statusDesc == "error":
+                    Utils.errorExit("javascript client reporting error see: %s." % (shipClientErrorFile))
+
+        assert done, Print("ERROR: Did not find a \"done\" status for client %d" % (i))
+
+    Print("All clients active from block num: %s to block_num: %s." % (maxFirstBN, minLastBN))
+
+    stderrFile=Utils.getNodeDataDir(shipNodeNum, "stderr.txt")
+    biggestDeltaSeconds = 0.0
```
jgiszczak: `biggestDeltaSeconds = timedelta(seconds=0)`

jgiszczak:
```
biggestDelta = timedelta(seconds=0)
totalDelta = timedelta(seconds=0)
```

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+                rcvTimeStr = match.group(1)
+                prodTimeStr = match.group(3)
+                blockNum = int(match.group(2))
+                if blockNum > maxFirstBN:
+                    # ship requests can only affect time after clients started
+                    timeFmt = '%Y-%m-%dT%H:%M:%S.%f'
+                    rcvTime = datetime.strptime(rcvTimeStr, '%Y-%m-%dT%H:%M:%S.%f')
+                    prodTime = datetime.strptime(prodTimeStr, timeFmt)
+                    delta = rcvTime - prodTime
+                    deltaSeconds = delta.total_seconds()
```
jgiszczak: Together with initializations above, just use `biggestDelta = max(delta, biggestDelta)`

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+                    if deltaSeconds > biggestDeltaSeconds:
+                        biggestDeltaSeconds = deltaSeconds
+
+                    totalDeltaSeconds += deltaSeconds
+                    timeCount += 1
+                    assert deltaSeconds < 0.100, Print("ERROR: block_num: %s took more than %.3f seconds to be received." % (blockNum, deltaSeconds))
```
jgiszczak: `assert delta < timedelta(seconds=0.100)`

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+
+            line = f.readline()
+
+    avg = totalDeltaSeconds / timeCount if timeCount > 0 else 0.0
```
jgiszczak: `totalDelta / timeCount` will work as expected.  `avg` will be a `timedelta` object which will print in the format `0:00:00.000000` or `0:00:00,000000` depending on locale so the word 'seconds' can be omitted in the message below.

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+finally:
+    TestHelper.shutdown(cluster, walletMgr, testSuccessful=testSuccessful, killEosInstances=killEosInstances, killWallet=killWallet, keepLogs=keepLogs, cleanRun=killAll, dumpErrorDetails=dumpErrorDetails)
+    if shipTempDir is not None:
+        if dumpErrorDetails and not testSuccessful:
+            for index in range(0, args.num_clients):
+                Print(Utils.FileDivider)
+                shipClientErrorFile = "%s%d.err" % (shipClientFilePrefix, i)
+                with open(shipClientErrorFile, "r") as f:
+                    Print("Contents of %s" % (shipClientErrorFile))
+                    line = f.readline()
```
jgiszczak: Since there's no check on a maximum line count here, just `Print(f.read())` and dispense with manually iterating through the lines of the file.

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+                    while line:
+                        Print(line)
+                        line = f.readline()
+                Print(Utils.FileDivider)
+                lineCount = 0
+                shipClientOutputFile = "%s%d.out" % (shipClientFilePrefix, i)
+                with open(shipClientOutputFile, "r") as f:
+                    Print("Contents of %s" % (shipClientOutputFile))
+                    line = f.readline()
```
jgiszczak:
```
maxLines = 20000
for lineCount, line in zip(range(1, maxLines+1), f.readlines()):
    Print(line)
if lineCount == maxLines:
    Print("...       CONTENT TRUNCATED AT %d lines" % (maxLines))
```

tests/Node.py
brianjohnson5972: removed unused methods

tests/ship_test.py
brianjohnson5972: removed

tests/ship_test.py
```diff
@@ -0,0 +1,235 @@
...
+        popen=Utils.delayedCheckOutput(cmd, stdout=outFile, stderr=errFile)
+        starts.append(time.perf_counter())
+        clients.append((popen, cmd))
+        files.append((outFile, errFile))
+        Print("Client %d started, Ship node head is: %s" % (i, shipNode.getBlockNum()))
+
+    Print("Stopping all %d clients" % (args.num_clients))
+
+    index = 0
+    for popen, _ in clients:
```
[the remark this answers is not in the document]
brianjohnson5972: cleaned up the use of files, but still using index for reporting status.

tests/ship_test.py
brianjohnson5972: using timeDelta.get_seconds() since 0:00:xx.xxxxxx is really overkill

tests/ship_test.py
brianjohnson5972: Changed this to limit the number of lines.  Added function to do that.

tests/ship_test.py
brianjohnson5972: we discussed this and keeping the original done

PR #8223, opened 2019-11-15
kv database

tbfleming:
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -132,6 +132,52 @@ namespace eosio { namespace chain {
...
+   template<typename A, typename B>
+   inline int compare_blob(const A& a, const B& b) {
+      static_assert(
+         std::is_same_v<std::decay_t<decltype(*a.data())>, char> ||
+         std::is_same_v<std::decay_t<decltype(*a.data())>, unsigned char>);
+      static_assert(
+         std::is_same_v<std::decay_t<decltype(*b.data())>, char> ||
+         std::is_same_v<std::decay_t<decltype(*b.data())>, unsigned char>);
+
+      auto r = std::char_traits<unsigned char>::compare(
```
swatanabe-b1: The standard only defines specializations char_traits for char, [char8_t], char16_t, char32_t, and wchar_t.

libraries/chain/kv_context_chainbase.cpp
```diff
@@ -0,0 +1,225 @@
...
+         }
+         return move_to(idx.lower_bound(boost::make_tuple(database_id, contract, prefix)));
+      }
+
+      kv_it_stat kv_it_prev() override {
+         std::decay_t<decltype(idx.end())> it;
+         if (kv_key)
+            it = idx.lower_bound(boost::make_tuple(database_id, contract, *kv_key));
+         else if (next_prefix)
+            it = idx.lower_bound(boost::make_tuple(database_id, contract, *next_prefix));
```
swatanabe-b1: This isn't quite right.  Consider:
```
keys = { "\x00\xFF\xFF", "\x01", "\x01\x00" }
prefix = "\x00\xFF\xFF";
next_prefix = "\x01\x00\x00"
```

Then ``lower_bound="\x01\x00"``, decrementing it gives ``"\x01"``, which is outside the prefix, so ``move_to`` returns ``iterator_end``.

libraries/chain/kv_context_chainbase.cpp
```diff
@@ -0,0 +1,225 @@
...
+         actual_size = kv->kv_value.size();
+         return kv_it_stat::iterator_ok;
+      }
+   }; // kv_iterator_chainbase
+
+   struct kv_context_chainbase : kv_context {
+      chainbase::database&       db;
+      name                       database_id;
+      name                       receiver;
+      std::optional<shared_blob> temp_data_buffer;
```
swatanabe-b1: I think optional is pointless.  There's no difference in behavior between an empty buffer and a buffer that holds an empty string, right.

tbfleming: It prevents me from having to construct a shared_blob, which needs an allocator

libraries/chain/kv_context_chainbase.cpp
tbfleming: `lower_bound=end`

tbfleming: Oops. Even with that it's not right.

libraries/chain/include/eosio/chain/types.hpp
tbfleming: Switched to `memcmp`

libraries/chain/kv_context_chainbase.cpp
tbfleming: Switched `next_prefix` generation.

libraries/chain/resource_limits.cpp
```diff
@@ -271,20 +310,37 @@ bool resource_limits_manager::set_account_limits( const account_name& account, i
...
+   auto& limits = get_or_create_pending_account_limits( account );
+   bool decreased_limit = disk_bytes >= 0 && (limits.disk_bytes < 0 || disk_bytes < limits.disk_bytes);
+   _db.modify(limits, [disk_bytes]( resource_limits_object& pending_limits ) {
+      pending_limits.disk_bytes = disk_bytes;
+   });
+   return decreased_limit;
+}
+
+int64_t resource_limits_manager::get_account_disk_limit( const account_name& account ) const {
+   const auto* pending_buo = _db.find<resource_limits_object,by_owner>( boost::make_tuple(true, account) );
```
arhag: Remove this line.

libraries/chain/apply_context.cpp
```diff
@@ -882,6 +983,10 @@ void apply_context::add_ram_usage( account_name account, int64_t ram_delta ) {
    }
 }
 
+void apply_context::add_disk_usage( account_name account, int64_t disk_delta ) {
+   trx_context.add_disk_usage( account, disk_delta );
```
arhag: Are we not interested in tracking DISK deltas per account in the traces just like with RAM?

libraries/chain/include/eosio/chain/kv_config.hpp
```diff
@@ -0,0 +1,42 @@
...
+namespace eosio { namespace chain {
+
+   /**
+    * @brief limits for a kv database.
+    *
+    * Each database (ram or disk, currently) has its own limits for these parameters.
+    * The key and value limits apply when adding or modifying elements.  They may be reduced
+    * below existing database entries.
+    */
+   struct kv_database_config {
```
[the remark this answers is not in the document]
swatanabe-b1: It won't quite disallow usage.  You could still have an empty key with an empty value.

libraries/chain/controller.cpp
```diff
@@ -960,13 +966,31 @@ struct controller_impl {
                   section.read_row(legacy_global_properties, db);
 
                   db.create<global_property_object>([&legacy_global_properties,&gs_chain_id](auto& gpo ){
-                     gpo.initalize_from(legacy_global_properties, gs_chain_id);
+                     gpo.initalize_from(legacy_global_properties, gs_chain_id, genesis_state{}.initial_kv_configuration);
```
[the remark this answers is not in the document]
swatanabe-b1: I'm completely redoing the `genesis_state` and including a version.

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -83,6 +83,10 @@ const static uint32_t   default_block_cpu_effort_pct           = 80 * percent_1;
 const static uint16_t   default_controller_thread_pool_size    = 2;
 const static uint32_t   default_max_variable_signature_length  = 16384u;
 
+const static uint32_t   default_max_kv_key_size                = 1024;
+const static uint32_t   default_max_kv_value_size              = 256*1024; // Large enough to hold most contracts
```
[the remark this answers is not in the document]
swatanabe-b1: I changed it to 1MiB.  It's unused, right now though.

libraries/chain/include/eosio/chain/kv_config.hpp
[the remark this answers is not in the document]
arhag: Okay. Can you explicitly value initialize the fields of `kv_database_config` so that it is robust to modifications of the structs in ways that seem like they shouldn't change things but do in often subtle ways? (Example someone throwing an explicit default constructor on `kv_config` and forgetting to initialize `kvram` and `kvdisk` explicitly.)

libraries/chain/apply_context.cpp
arhag: I saw you added the `account_disk_deltas` to `action_trace`, but it will also need to be included into a SHiP `action_trace_v1` struct. Note that we will need to introduce an `action_trace_v1` struct here but it will also need to be included for the changes planned for action return values (so watch out for conflicts with that upcoming PR). The current implementation of action return values in develop (which this PR branch includes) introduced `action_receipt_v1`, because the entire return value was added into the receipt (which we don't want to do anymore), but did not need to modify the `action_trace`. The new approach will interestingly not require a change to the `action_receipt` (there is no need for `action_receipt_v1` because there is no new field added to the receipt, instead the way `act_digest` gets computed changes when the `ACTION_RETURN_VALUE` protocol feature is activated) but will require an `action_trace_v1` to hold the `return_value` field.

libraries/chain/kv_context_chainbase.cpp
```diff
@@ -0,0 +1,244 @@
...
+      name                       contract;
+      std::vector<char>          prefix;
+      std::vector<char>          next_prefix;
+      const kv_object*           current = nullptr;
+
+      kv_iterator_chainbase(chainbase::database& db, tracker_type& tracker, uint32_t& itr_count, name database_id, name contract, std::vector<char> prefix)
+         : db{ db }, tracker{ tracker }, itr_count(itr_count), database_id{ database_id }, contract{ contract }, prefix{ std::move(prefix) } {
+
+         ++itr_count;
+         next_prefix = this->prefix;
```
arhag: If we were willing to create a special `prefix_supremum` type that contained a single `string_view` (into `prefix`) and created overloads of `operator()` within `unsigned_blob_less` to handle `prefix_supremum` as one of its arguments, then we could still use `next_prefix` in the implementation of `kv_it_prev ` but avoid this extra copy here and also avoid having to potentially iterate up to `prefix.size()` times below.

arhag: In fact, if we are willing to redesign the intrinsics `kv_it_create` to not retain the `prefix` in native memory (and essentially create an iterator that acts as if an empty prefix was provided) and redesign the intrinsics `kv_it_next`, `kv_it_prev`, and `kv_it_lower_bound` to all take a `prefix` which defines the range of keys the iterator can cover as a restrictive context under which that operation acts, then we could avoid needing to copy a prefix at all and dramatically lower the maximum native-side memory usage a contract could impose.

libraries/chain/include/eosio/chain/kv_context.hpp
```diff
@@ -0,0 +1,65 @@
...
+      virtual kv_it_stat kv_it_prev()                                                                   = 0;
+      virtual kv_it_stat kv_it_lower_bound(const char* key, uint32_t size)                              = 0;
+      virtual kv_it_stat kv_it_key(uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size)   = 0;
+      virtual kv_it_stat kv_it_value(uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size) = 0;
+   };
+
+   struct kv_resource_manager {
+      void     update_table_usage(int64_t delta) { return _update_table_usage(*_context, delta); }
+      apply_context* _context;
+      uint64_t       billable_size;
```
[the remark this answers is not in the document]
swatanabe-b1: It can be const, but should it be?  It doesn't make a big difference either way, but I usually don't make non-static data members const, as it prevents assignment.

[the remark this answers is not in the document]
arhag: No, never mind. Just keep as is.

libraries/chain/kv_context_chainbase.cpp
```diff
@@ -0,0 +1,244 @@
...
+         while (!next_prefix.empty()) {
+            if (++next_prefix.back())
```
tbfleming: I forgot to do the add as uint8_t

libraries/chain/kv_context_chainbase.cpp
arhag: We decided to go another (simpler) way of just disallowing prefixes longer than the current maximum key limit.

libraries/chain/kv_context_chainbase.cpp
tbfleming: Correction: it's ok.

unittests/test-contracts/kv_test/kv_test.cpp
```diff
@@ -0,0 +1,398 @@
...
+            for(uint32_t i = 0; i < itop.count; ++i) {
+               kv_it_create(itop.db, ""_n, nullptr, 0);
+            }
+         }
+      } else {
+         std::map<name, std::vector<int>> iterators;
+         for(auto itop : params) {
+            auto& iters = iterators[itop.db];
+            if(itop.erase) {
+               auto& iters = iterators[itop.db];
```
arhag: Redundant.

unittests/test-contracts/kv_bios/kv_bios.cpp
```diff
@@ -0,0 +1,44 @@
...
+   }
+   void kvlimits_impl(name db, uint32_t k, uint32_t v, uint32_t i) {
+      uint32_t limits[4];
+      limits[0] = 0;
+      limits[1] = k;
+      limits[2] = v;
+      limits[3] = i;
+      set_kv_parameters_packed(db.value, limits, sizeof(limits));
+      int sz = get_kv_parameters_packed(db.value, nullptr, 0, 0);
+      limits[0] = limits[1] = 0xFFFFFFFF;
```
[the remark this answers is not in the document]
swatanabe-b1: There were only two fields originally.

Issue #8224, opened 2019-11-16
Create integration test with identical transactions sent to different nodes


brianjohnson5972:
Will send identical transactions to multiple nodes and verify that only 1 gets received.

heifner:
Resolved by #8232

PR #8232, opened 2019-11-20
Create integration test for sending copies of the same transaction into the network

brianjohnson5972:
## Change Description
Added flags and made changes to allow previous test to construct transfer transactions and send them to all nodes and verify that all transaction make it into a block exactly once.
#8224

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -1031,7 +1031,7 @@ class Node(object):
             return (False, msg)
 
     # returns tuple with indication if transaction was successfully sent and either the transaction or else the exception output
-    def pushTransaction(self, trans, opts = "--skip-sign", silentErrors=False):
+    def pushTransaction(self, trans, opts="--skip-sign", silentErrors=False):
         assert(isinstance(trans, dict))
         cmd="%s %s push transaction -j" % (Utils.EosClientPath, self.eosClientArgs())
         cmdArr=cmd.split()
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -715,15 +715,26 @@ class Node(object):
         return self.waitForBlock(blockNum, timeout=timeout, blockType=blockType)
 
     # Trasfer funds. Returns "transfer" json return object
-    def transferFunds(self, source, destination, amountStr, memo="memo", force=False, waitForTransBlock=False, exitOnError=True, reportStatus=True, sign=False):
+    def transferFunds(self, source, destination, amountStr, memo="memo", force=False, waitForTransBlock=False, exitOnError=True, reportStatus=True, sign=False, dontSend=False, expiration=None):
         assert isinstance(amountStr, str)
         assert(source)
         assert(isinstance(source, Account))
         assert(destination)
         assert(isinstance(destination, Account))
+        assert(expiration is None or isinstance(expiration, int))
 
-        cmd="%s %s -v transfer -j" % (
-            Utils.EosClientPath, self.eosClientArgs())
+        dontSendStr = ""
+        if dontSend:
+            dontSendStr = "--dont-broadcast "
+            if expiration is None:
+                # default transaction expiration to be 4 minutes in the future
+                expiration = 240
+
+        expirationStr = ""
+        if expiration is not None:
+            expirationStr = "--expiration %d " % (expiration)
+        cmd="%s %s -v transfer -j %s %s" % (
+            Utils.EosClientPath, self.eosClientArgs(), dontSendStr, expirationStr)
         cmdArr=cmd.split()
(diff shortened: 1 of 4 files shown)


tests/Node.py
```diff
@@ -1019,6 +1030,33 @@ def pushMessage(self, account, action, data, opts, silentErrors=False, signature
                 Utils.Print("ERROR: Exception during push message.  cmd Duration=%.3f sec.  %s" % (end - start, msg))
             return (False, msg)
 
+    # returns tuple with indication if transaction was successfully sent and either the transaction or else the exception output
+    def pushTransaction(self, trans, opts = "--skip-sign", silentErrors=False):
```
jgiszczak: Whitespace inconsistency with default value of argument.

tests/nodeos_high_transaction_test.py
```diff
@@ -237,22 +255,46 @@ def findTransInBlock(transId, transToBlock, node):
...
+                trans=node.transferFunds(fromAccount, toAccount, transferAmount, "transfer round %d" % (round), exitOnError=False, reportStatus=False, sign = True, dontSend = args.send_duplicates)
+                attempts += 1
+
+            if args.send_duplicates:
+                sendTrans = trans
+                trans = None
+                numAccepted = 0
+                attempts = 0
+                while trans is None and attempts < 2:
+                    for nodeNum in nodeOrder:
```
jgiszczak: This and the following line can be `for node in map(lambda ordinal: allNodes[ordinal], nodeOrder):`.  It's a little more obvious that it's `allNodes` being iterated.  Maybe.  I'm not insisting.

tests/nodeos_high_transaction_test.py
```diff
@@ -218,6 +222,20 @@ def findTransInBlock(transId, transToBlock, node):
     #verify nodes are in sync and advancing
     cluster.waitOnClusterSync(blockAdvancing=5)
 
+    nodeOrder = None
```
jgiszczak: While iteration of `nodeOrder` on line 283 is protected with another `if args.send_duplicates:`, better to initialize it to `[]` rather than `None`.  Later logic changes that might remove the check would then degrade gracefully to doing nothing in the loop rather than throwing `TypeError`.

tests/nodeos_high_transaction_test.py
```diff
@@ -237,22 +255,46 @@ def findTransInBlock(transId, transToBlock, node):
...
             node = nonProdNodes[accountIndex % nonProdNodeCount]
-            trans=node.transferFunds(fromAccount, toAccount, transferAmount, "transfer round %d" % (round), exitOnError=False, reportStatus=False, sign = True)
-            if trans is None:
-                # delay and see if transfer is accepted now
-                Utils.Print("Transfer rejected, delay 1 second and see if it is then accepted")
-                time.sleep(1)
-                trans=node.transferFunds(fromAccount, toAccount, transferAmount, "transfer round %d" % (round), exitOnError=False, reportStatus=False, sign = True)
+            trans = None
+            attempts = 0
+            max_attempts = 3 if not args.send_duplicates else 1  # for send_duplicates we are just constructing a transaction, so should never require a second attempt
```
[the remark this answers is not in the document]
jgiszczak: Not a class member.  It's a `try` block, not a class.  So just a constant at the beginning of the file with the other initializations.

tests/nodeos_high_transaction_test.py
```diff
@@ -237,22 +255,46 @@ def findTransInBlock(transId, transToBlock, node):
...
+                    time.sleep(1)
+                trans=node.transferFunds(fromAccount, toAccount, transferAmount, "transfer round %d" % (round), exitOnError=False, reportStatus=False, sign = True, dontSend = args.send_duplicates)
+                attempts += 1
+
+            if args.send_duplicates:
+                sendTrans = trans
+                trans = None
+                numAccepted = 0
+                attempts = 0
+                while trans is None and attempts < 2:
```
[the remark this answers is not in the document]
brianjohnson5972: Well I guessed at 3 to allow us to catch up if we somehow exceeded the net staking for each account, but for the duplicates I think we could waste a lot of time trying to send it 3 times when likely it is because the p2p transaction got there before the duplicate did.  But decided to make both of them 2 and use the one constant.

tests/nodeos_high_transaction_test.py
brianjohnson5972: Actually, I don't use nodeNum, so I will just iterate over allNodes.

Issue #8235, opened 2019-11-22
Unable to run Cleos after eosio install via brew on Catalina


ismyhc:
Trying to install eosio and use Cleos on Catalina is giving the following error:

OS: Catalina
Fresh install of brew

brew tap eosio/eosio
brew install eosio

then when running Cleos

```
dyld: Library not loaded: /usr/local/opt/openssl/lib/libssl.1.0.0.dylib
  Referenced from: /usr/local/bin/cleos
  Reason: image not found
```

There is a dylib at that location, but the version is different.
what I see is /usr/local/opt/openssl/lib/libssl.1.1.1.dylib

Tried 
brew update && brew upgrade
brew install openssl

brew remove openssl
brew install openssl

Any ideas?

chesterliliang:
same issue

bigonenice:
same isuue

hkabesa:
+1

orbachar:
+1

spoonincode:
This is ultimately due to homebrew removing openssl 1.0. We'll have a quick fix to rebuild our binaries with openssl 1.1; long term we need to investigate static linking openssl again.

duongph223:
@spoonincode Please help me to fix this.

parcool:
```
$ dyld: Library not loaded: /usr/local/opt/openssl/lib/libcrypto.1.0.0.dylib
  Referenced from: /usr/local/bin/keosd
  Reason: image not found
```
macOS Majave.

hoang-dao:
Homebrew changed its version of Openssl to 1.1 days ago (https://github.com/Homebrew/homebrew-core/commit/0349a7ca76f483483c6d5c1d4cfe6b458dee2665 and https://github.com/Homebrew/homebrew-core/commit/b4b9c1cd84772e99432b5d2679fb0068da1e1e64).
We can downgrade Openssl to the older version to temporarily fix this issue while waiting for the new rebuild to come (or do it manually):

```
brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/30fd2b68feb458656c2da2b91e577960b11c42f4/Formula/openssl.rb
```

arezfuzin:
@hoang-dao your solution is working for mongodb but it will put issue for another package like postgres or telepresence.

spoonincode:
This is fixed in 1.8.7's build scripts as it explicitly seeks out version 1.1

aiavci:
[...]

PR #8245, opened 2019-11-25
[Develop] explictly use openssl 1.1 via brew on macos

spoonincode:
## Change Description

At the moment our macos binaries are not working for fresh mac installs, see #8235.

homebrew has removed openssl 1.0 from their repositories. See some discussion here
https://github.com/Homebrew/homebrew-core/pull/46876
Additionally, they decided to alias `openssl` to install and point to the 1.1 version, see discussion here
https://github.com/Homebrew/homebrew-core/pull/47034

The last bit is somewhat interesting: if on a fresh macos you do a `brew install openssl` you'll end up installing openssl 1.1 and also `/usr/local/opt/openssl` is symlinked over to `openssl@1.1`. So to fix our binary problem all we need to do is reset our CI image and let it install openssl 1.1 now instead.

However, that won't (easily) cause builds on macos installations that already have openssl 1.0 installed to use 1.1 as `/usr/local/opt/openssl` could still be pointing to openssl 1.0. At least for me, just doing a `brew upgrade` won't fix up this directory to be 1.1, and `brew upgrade openssl` won't do anything since openssl 1.1 is already installed. One does wonder if another openssl 1.1 upgrade is pushed out in homebrew if the upgrade will stomp on 1.0 or not?

Anyways, that long winded train of thought is there just to lead in to the simple solution: let's just be more explicit and ask for 1.1 directly. It ensures everyone using the build scripts will indeed be using it and we won't get odd surprises.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/platforms/pinned/macos-10.14-pinned.sh
+++ b/.cicd/platforms/pinned/macos-10.14-pinned.sh
@@ -2,7 +2,7 @@
 set -eo pipefail
 VERSION=1
 brew update
-brew install git cmake python@2 python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl jq || :
+brew install git cmake python@2 python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl@1.1 jq || :
 # install clang from source
 git clone --single-branch --branch release_80 https://git.llvm.org/git/llvm.git clang8
 cd clang8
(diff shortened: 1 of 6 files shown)


NorseGaud:
1.8 is covered in https://github.com/EOSIO/eos/pull/8246

PR #8246, opened 2019-11-25
Install llvm@4 from homebrew -> Install llvm release_40 from github/source + openssl@1.1 fix

NorseGaud:
## Change Description

Homebrew removed llvm@4, so we need to install it from the source as we do for Linux.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
technically you can still install it from brew by referencing it like
`brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/1d91e94e8ecdf6877ad2c24a7cda1114e50f2a14/Formula/llvm%404.rb`
Not really sure that's worth doing

NorseGaud:
> technically you can still install it from brew by referencing it like
> `brew install https://raw.githubusercontent.com/Homebrew/homebrew-core/1d91e94e8ecdf6877ad2c24a7cda1114e50f2a14/Formula/llvm%404.rb`
> Not really sure that's worth doing

This is what I'm thinking:
1. We've seen this multiple times now from brew that it makes everyone uncomfortable.
2. Personally, I've advocated for code consistency between Linux and mac so things are easier to maintain; if we install from source for one, why not install from source for the other.

I'd like to use this as an opportunity to move darwin away from brew.

spoonincode:
> I'd like to use this as an opportunity to move darwin away from brew.

Indeed, as I've said multiple times I would prefer binaries that have all deps beyond the basics static linked so the user could just untar & run. For both linux & mac. Unfortunately gmp & libusb make this problematic because we can't static link them and our legal team does not want us distributing bundled binaries with them (which would be cheesy anyways).

I have private branches that remove the gmp & libusb deps (with some caveats) but I'm having trouble rallying enough support to take this further. I believe with low to medium effort we could take our pinned build script and massively simplify it such that a single script builds all deps and works anywhere... linux/mac/freebsd/mingw, whatever. (python, needed for some unittests, an exception here)

I actually may need to take that on anyways: our requirement for a hardened macos nodeos for "pass project" would prohibit linking to any unsigned libraries (like gmp, etc) as macos hardened runtime does not allow that.

spoonincode:
can you cherry pick in 3b54f79 ?

NorseGaud:
> > I'd like to use this as an opportunity to move darwin away from brew.
> 
> Indeed, as I've said multiple times I would prefer binaries that have all deps beyond the basics static linked so the user could just untar & run. For both linux & mac. Unfortunately gmp & libusb make this problematic because we can't static link them and our legal team does not want us distributing bundled binaries with them (which would be cheesy anyways).
> 
> I have private branches that remove the gmp & libusb deps (with some caveats) but I'm having trouble rallying enough support to take this further. I believe with low to medium effort we could take our pinned build script and massively simplify it such that a single script builds all deps and works anywhere... linux/mac/freebsd/mingw, whatever. (python, needed for some unittests, an exception here)
> 
> I actually may need to take that on anyways: our requirement for a hardened macos nodeos for "pass project" would prohibit linking to any unsigned libraries (like gmp, etc) as macos hardened runtime does not allow that.

Apologies, I meant move LLVM away from brew.

> can you cherry pick in 3b54f79 ?

Sure, I'll handle that first thing tomorrow

CMakeLists.txt
```diff
@@ -71,7 +71,7 @@ if ("${OPENSSL_ROOT_DIR}" STREQUAL "")
       set(OPENSSL_ROOT_DIR $ENV{OPENSSL_ROOT_DIR})
       set(OPENSSL_INCLUDE_DIR ${OPENSSL_ROOT_DIR}/include)
    elseif (APPLE)
-      set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl")
+      set(OPENSSL_ROOT_DIR "/usr/local/opt/openssl@1.1")
       set(OPENSSL_INCLUDE_DIR "/usr/local/opt/openssl/include")
```
spoonincode: @1.1 here too

NorseGaud:
I tested the build script for each distro and it works: https://buildkite.com/EOSIO/eosio-scripts/builds/6

NorseGaud:
Tested the build scripts on each distro and it seems to work: https://buildkite.com/EOSIO/eosio-scripts/builds/8

PR #8306, opened 2019-12-10
Extend shutdown allowed time in under min available resources test

brianjohnson5972:
## Change Description
Fixed corner case for failures and added debug logging. And increase time allowed to shutdown.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -1363,7 +1363,9 @@ class Node(object):
         self.killed=True
 
     def verifyAlive(self, silent=False):
-        if not silent and Utils.Debug: Utils.Print("Checking if node(pid=%s) is alive(killed=%s): %s" % (self.pid, self.killed, self.cmd))
+        logStatus=not silent and Utils.Debug
+        pid=self.pid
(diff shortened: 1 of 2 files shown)


tests/Node.py
```diff
@@ -1375,10 +1377,13 @@ def verifyAlive(self, silent=False):
             # mark node as killed
             self.pid=None
             self.killed=True
+            if logStatus: Utils.Print("Determined node(formerly pid=%s) is killed" % (pid))
             return False
         except PermissionError as ex:
+            if logStatus: Utils.Print("Determined node(formerly pid=%s) is killed" % (pid))
```
heifner: This should say "is alive", right?

tests/Node.py
```diff
@@ -1375,10 +1377,13 @@ def verifyAlive(self, silent=False):
...
             return True
 
+        if logStatus: Utils.Print("Determined node(pid=%s) is alive (kill command executed)" % (self.pid))
```
heifner: This didn't call `kill`, well technically it did, but the kill is kill 0 which is just check the status. I think you should remove "(kill command executed)"

PR #8307, opened 2019-12-10
reset the new handler

larryk85:
## Change Description

Because LLVM sets the new_handler to abort when new cannot allocate any more memory, we've lost the ability to exit cleanly in some instances and not corrupt the DB.  This fix should reset the new_handler to throw `std::bad_alloc` and use the old behavior of EOSIO.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -32,6 +32,8 @@
 #include <fc/scoped_exit.hpp>
 #include <fc/variant_object.hpp>
 
+#include <new>
+
 namespace eosio { namespace chain {
 
 using resource_limits::resource_limits_manager;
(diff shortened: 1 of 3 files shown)


larryk85:
Want to see if the tests fail with the handler unset and ensure that the tests pass when I uncomment the `throw std::bad_alloc` line.

libraries/chain/controller.cpp
```diff
@@ -304,6 +315,8 @@ struct controller_impl {
     read_mode( cfg.read_mode ),
     thread_pool( "chain", cfg.thread_pool_size )
    {
+
+      // reset the
```
heifner: Looks like this was added unintentionally and also the newline below.

larryk85: Gotcha

PR #8309, opened 2019-12-11
Extend shutdown allowed time in under min available resources test - 2.0

brianjohnson5972:
## Change Description
Fixed corner case for failures and added debug logging. And increase time allowed to shutdown.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


brianjohnson5972:
The fact that this PR was initially created with devlop as the base seems to have caused problems in the buildkite pipeline, so closing and opening a new PR.

PR #8332, opened 2019-12-16
Docs as Code for source installation

NorseGaud:
## Change Description

**Problem:** The source install procedure documentation is typically updated last, after changes to CI and EOSIO have happened.

**Solution:** Refactor CI scripts and pipelines to pull commands from documentation so we ensure updates are made to the docs before CI can pass. This causes all teams involved to be responsible for updating the docs, instead of only Developer Relations.

What these changes do:

1. Generation of tags (docker & anka) is done using every code block EXCEPT for the TEST block and any others wrapped in an "IGNORE" comment.
    - Why? We use CTEST and not `make test` in our scripts so we can split up the tests into individual jobs/steps.
    - We need to test everything that a community member would run. This is an improvement over how we currently generate tags since we only run the dependency install and don't test if they actually work or not. **We can avoid filling up the registries with bad images.**
2. `$HOME/eosio` is the new root. The commands will clone into `$HOME/eosio/eos`. However, all dependencies and eos binaries will end up in `$HOME/eosio/install/(bin|lib|etc)`.
    - If users want to have more than one eosio installation on their system (common for BPs), they can modify the $EOSIO_LOCATION env from the instructions.
    - CDT and Contracts will also use `$HOME/eosio/cdt` & `$HOME/eosio/install`, etc. This allows for the creation of a new build script that will pull the install commands from the docs in a specific branch of each product/repo (eosio, cdt, contracts) and then install them in the right order for community members.

## Code block rules/conversions:

- The following are removed from all code blocks: `grep -v -e '<!--' -e '-->' -e '```' -e '\#.*' -e '^$'`
- All code blocks in the MD files are wrapped with a comment like `<!-- DAC CLONE -->` or `<!-- DAC BUILD -->`, allowing our scripts to pull specific sections.
- Anything inside of `<!-- DAC IGNORE` comments are ignored (like the test block)
- Indented lines are executed together with the last non-indented line. This allows you to execute multiple commands in the same docker layer.

#### Dockerfile
1. Lines matching `^[ ].*` (a space at the start of the line) are printed without any modification.
2. Lines starting with `^export.*=([^\s]+)$` are converted to have ENV before them so that they're available in the docker tag creation process and not a specific layer.

#### Mac SH script
1. Nothing :)
---

We tested that eosio.contracts works with the image generated from this: https://buildkite.com/EOSIO/eosio-dot-contracts/builds/991#_

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


swatanabe-b1:
Renaming scripts from .sh to .md doesn't make them documentation.

NorseGaud:
> Renaming scripts from .sh to .md doesn't make them documentation.

Would you be able to elaborate for me? I'm not sure I understand what you mean.

NorseGaud:
> Still evaluating how we want to proceed with this.

No worries. Though, I want to remind everyone that this PR does not mean that the docs we're pulling commands from are going to be included in Thorka and suggested to the community. Here is what I see happening:

1. Community uses packages.
2. If they don't want packages, they use the build scripts.
3. IF they don't want to use those scripts, they can copy and paste from the docs we have available.

At some point, we can create a wrapper script that replaces the build script and instead pulls what it needs to execute for the community member from the docs of each repo/project and install them in the right order.

NorseGaud:
Github actions are all passing: https://github.com/EOSIO/eos/pull/8449

lparisc:
@NorseGaud @arhag d0e4df4 removes the glue that maps`.cicd/docs` to eos `docs` from where actual documentation is rendered. Once other pending items are addressed, if any, this PR should be able to proceed.

heifner:
Decided to go a different direction.

PR #8335, opened 2019-12-16
Add `_n` operator to the `name` type

johndebord:


libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -77,6 +77,19 @@ namespace eosio::chain {
...
 
+   namespace detail {
+      template <char... Str>
+      struct to_const_char_arr {
+         static constexpr const char value[] = {Str...};
+      };
+   } // namespace detail
+
+   template <typename T, T... Str>
+   inline name operator""_n() {
```
larryk85: This should be `constexpr`, I'm not sure why this is different than what is in eosio.cdt.  One thing is that you need to make sure the `name` constructors are `constexpr`.

brianjohnson5972:
@johndebord is this really still a work in progress, or has it been abandoned?  If I don't hear back from you I'm going to assume it has been abandoned.

jeffreyssmith2nd:
Closing due to lack of response/activity. Can be reopened if this is still needed.

PR #8360, opened 2019-12-19
Make wasm constraints configurable

swatanabe-b1:
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


swatanabe-b1:
Depends on #8223 and EOSIO/eos-vm#144

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -147,20 +148,42 @@ executor::executor(const code_cache_base& cc) {
...
 
+   uint64_t max_call_depth = eosio::chain::wasm_constraints::maximum_call_depth+1;
+   uint64_t max_pages = eosio::chain::wasm_constraints::maximum_linear_memory/eosio::chain::wasm_constraints::wasm_page_size;
+   if(context.control.is_builtin_activated(builtin_protocol_feature_t::configurable_wasm_limits)) {
+      const wasm_config& config = context.control.get_global_properties().wasm_configuration;
+      max_call_depth = config.max_call_depth;
+      max_pages = config.max_pages;
+   }
+   mem.reset(max_pages);
+   EOS_ASSERT(code.starting_memory_pages <= (int)max_pages, wasm_execution_error, "Initial memory out of range");
```
spoonincode: Is this line trying to detect when `config.max_pages` has been shrunk since the wasm was instantiated?

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -147,20 +148,42 @@ executor::executor(const code_cache_base& cc) {
...
    }
 
+   uint64_t max_call_depth = eosio::chain::wasm_constraints::maximum_call_depth+1;
+   uint64_t max_pages = eosio::chain::wasm_constraints::maximum_linear_memory/eosio::chain::wasm_constraints::wasm_page_size;
+   if(context.control.is_builtin_activated(builtin_protocol_feature_t::configurable_wasm_limits)) {
+      const wasm_config& config = context.control.get_global_properties().wasm_configuration;
+      max_call_depth = config.max_call_depth;
+      max_pages = config.max_pages;
+   }
+   mem.reset(max_pages);
```
spoonincode: Far too dangerous IMO. Could fail to acquire the resources. Also I think if the limit is large (like 2GB or something) it won't even work as-is.
Probably beyond a certain threshold (could be 33MB, could be more, could be less) OC should switch to using `mprotect()`.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
swatanabe-b1: Yes.

libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
```diff
@@ -1258,12 +1281,11 @@ namespace LLVMJIT
 		}
 
 		if(module.tables.size()) {
-			current_prologue -= 8; //now pointing to LAST element
-			current_prologue -= 16*(module.tables.defs[0].type.size.min-1); //now pointing to FIRST element
 			auto tableElementType = llvm::StructType::get(context,{llvmI8PtrType, llvmI64Type});
-			defaultTablePointer = emitLiteralPointer((void*)current_prologue,tableElementType->getPointerTo(256));
+			llvm::Type* tableArrayTy = llvm::ArrayType::get(tableElementType, module.tables.defs[0].type.size.min);
+			defaultTablePointer = new llvm::GlobalVariable(*llvmModule, tableArrayTy, false, llvm::GlobalValue::ExternalLinkage, llvm::ConstantAggregateZero::get(tableArrayTy), getTableSymbolName());
```
spoonincode: does this need to be marked ExternallyInitialized?

libraries/chain/webassembly/eos-vm-oc/LLVMJIT.cpp
```diff
@@ -133,7 +133,7 @@ namespace LLVMJIT
 			if(SectionName == ".stack_sizes") {
 				return stack_sizes.emplace_back(numBytes).data();
 			}
-			WAVM_ASSERT_THROW(isReadOnly);
+			//WAVM_ASSERT_THROW(isReadOnly);
```
spoonincode: what's going on here?

libraries/chain/webassembly/eos-vm-oc/executor.cpp
spoonincode: Why does that check need to happen here instead of a central location?

spoonincode:
Can you describe what the behavior is when shrinking a limit after a wasm has been accepted? For example if the initial data max is set to 128KB, my wasm has 96KB worth of data, and then the limit is changed to 64KB.

swatanabe-b1:
The intent is that max_pages and max_call_depth are enforced on apply.  The rest are enforced at set_code.

libraries/chain/webassembly/eos-vm-oc/LLVMJIT.cpp
swatanabe-b1: The wasmTable is not in a read-only section.

spoonincode: ah okay. This is a safety check that llvm didn't generate a data section that it expects to write to. The table should be placed in a read only section, and the assert should go back in. AFAIK just setting the GlobalVariable const param to true and using private linkage works.

libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
swatanabe-b1: I don't think so.  The external initialization should run after the the module is fully loaded.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
swatanabe-b1: There isn't a convenient central location to put it.

libraries/chain/webassembly/eos-vm-oc/LLVMJIT.cpp
swatanabe-b1: I think I tried that, but eventually gave up.  In order to make it const, the initialization can't be handled externally.

spoonincode: a35f580 has an implementation that works.

spoonincode:
It seems like wasm memory initialization & memory_api need to be checktime'd now.

spoonincode:
actually, I wonder if some other intrinsics are abusable at large memory sizes, 256MB, or such.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
```diff
@@ -171,14 +194,16 @@ void executor::execute(const code_descriptor& code, const memory& mem, apply_con
...
+   cb->current_call_depth_remaining = max_call_depth + 1;
    cb->current_linear_memory_pages = code.starting_memory_pages;
+   cb->max_linear_memory_pages = max_pages;
    cb->first_invalid_memory_address = code.starting_memory_pages*64*1024;
    cb->full_linear_memory_start = (char*)mem.full_page_memory_base();
    cb->jmp = &executors_sigjmp_buf;
    cb->bounce_buffers = &executors_bounce_buffers;
    cb->running_code_base = (uintptr_t)(code_mapping + code.code_begin);
    cb->is_running = true;
+   cb->globals = globals;
```
spoonincode: I'm having trouble finding where this gets used.

spoonincode:
We need to bump OC's codegen version. I think it's probably possible to keep original code running in this new version by doing things like moving `max_linear_memory_pages` to the end of the control block but those changes could be done at a later time. At the moment we can just drop all version 0 on startup and start marking new ones version 1.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
swatanabe-b1: If I remember correctly, it's accessed from LLVMEmitIR.cpp when setting up the pointers for globals.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
spoonincode: It seems like the check could be `wasm_interface::apply()` (by way of placing starting memory pages in the `wasm_cache_entry`). But this would incur an additional lookup every execution which would be unfortunate.

spoonincode:
For example, something like this will run way over a 30ms deadline on EOS VM JIT, and it's only double the current memory size.

```wat
(module
 (import "env" "eosio_assert" (func $eosio_assert (param i32) (param i32)))
 (import "env" "memset" (func $memset (param i32) (param i32) (param i32) (result i32)))
 (memory 1024)
 (export "apply" (func $apply))
 (func $apply (param $0 i64) (param $1 i64) (param $2 i64)
   (drop (call $memset (i32.const 0) (i32.const 0xff) (i32.const 67108863)))
   (i32.store offset=67108860 (i32.const 0) (i32.const 0xffffffff))
   (call $eosio_assert (i32.const 1) (i32.const 0))
 )
)
```

Can't help but wonder if allowing a knob to control memory limits opens pandora's box on new deadline overrun attacks. Is anyone even asking for more memory?

swatanabe-b1:
I'm somewhat less concerned about deadline overruns, since this is primarily intended for private and permissioned chains.  These parameters will definitely cause problems if they are set too high on a public chain.

libraries/chain/webassembly/eos-vm-oc/executor.cpp
swatanabe-b1: I changed it.  The threshold for switching to mprotect can be configured at construction.

libraries/chain/webassembly/eos-vm-oc/switch_stack_linux.s
```diff
@@ -0,0 +1,13 @@
+.file	"switch_stack_linux.s"
```
spoonincode: Is this really linux only?

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/stack.hpp
```diff
@@ -0,0 +1,42 @@
...
+   execution_stack() = default;
+   execution_stack(const execution_stack&) = delete;
+   execution_stack& operator=(const execution_stack&) = delete;
+
+   void reset(std::size_t max_call_depth);
+   void reset();
+
+   template<typename F>
+   void run(F&& f) {
+      if (stack_top) {
```
spoonincode: why conditionally gate this? Why not have the altstack used all the time?

libraries/chain/webassembly/eos-vm-oc/switch_stack_linux.s
swatanabe-b1: It's not strictly linux only, but it won't work on OSX.  The code is fine on OSX, but the file format is not.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/stack.hpp
swatanabe-b1: I don't have a strong opinion either way.  It's mostly just to avoid an extra mmap when it isn't actually necessary.

spoonincode:
Increasing `max_call_depth` high enough causes nodeos crashes; increasing `max_linear_memory_init` or `max_pages` will cause deadline overruns. Public networks will be eager to bump some limits so we will need to quantify the threshold at which a limit becomes unsupported. Even private networks need guidance as values too high can cause crashes.

And since we're going to have to declare safe limits to use anyways, my potentially unpopular opinion is that we should not offer adjustable limits but rather a simple protocol feature that bumps limits to safe and tested values that we have full confidence don't introduce problems.

We know what limits users are hitting regularly. We know what limits are needlessly too strict. We know the scale of limits large private chain contracts need. I believe we are fully capable to derive limits that safely satisfy both public and private networks for another couple years.

Also, "supporting" academic limits like 4GB of wasm memory or having a max call depth of 1 just damage future optimization techniques for no gain.

swatanabe-b1:
Deadline overruns due to max_pages and max_linear_memory_init can be fixed.  I just don't consider fixing them a priority, right now.  There also isn't a hard cutoff for what is safe.  It's not like 9.9 ms is perfectly fine, while 10 ms is completely unacceptable.  It depends on how much slack you're willing to accept.  max_call_depth can be set ridiculously high before it actually crashes nodeos.  Probably much higher than we really want to set it.  It's really limited by how much hardware you're willing to throw at it.

I don't think this will really impact optimization.  It shouldn't be a problem to optimize by assuming the current limits and just wipe the cache whenever they change.

spoonincode:
>Deadline overruns due to max_pages and max_linear_memory_init can be fixed

I wonder how invasive that gets though. e.g. how well do the db functions behave with a 256MB payload?

> There also isn't a hard cutoff for what is safe. It's not like 9.9 ms is perfectly fine, while 10 ms is completely unacceptable

10ms is where we start paying out so indeed seems like a pretty good baseline for what really is completely unacceptable. I hope we are concerned with anything that approaches that.

> max_call_depth can be set ridiculously high before it actually crashes nodeos

It will crash in under 5ms. So we're at least going to have to explicitly provide a supported number for this knob to stave off a bug bounty. Regardless, I believe we shouldn't be placing "traps" for the user like this – I could imagine a private chain setting it high because hey, it's private, and then a contract going off the rails and causing a crash.

>   Probably much higher than we really want to set it

Yeah, so.. exactly my point. Nobody is ever going to care to fiddle with this knob at such extreme highs or lows. But by offering it, now we allow crashes. If we really feel this limit needs to be increased then a protocol feature that bumps it (along with other limits) to 2500 or whatever means we _know_ we didn't introduce any problems.

> I don't think this will really impact optimization. It shouldn't be a problem to optimize by assuming the current limits and just wipe the cache whenever they change.

Not all performance damage (like the increased VM usage in OC) can be resolved. Most others probably can with added complexity. And that’s my general point: nobody will be setting call depth to 1, and yet complexity must be added to support those values.

swatanabe-b1:
What does it even mean to be safe?  The point at which nodeos crashes is widely variable depending on the hardware it's running on.  Even below the point where it crashes, it will be using way too much memory.  How much is too much?  I don't know.  We can recommend values for public networks based on a guess about the least common denominator, but I don't really believe that will be a sufficient one-size-fits-all solution.

libraries/chain/wasm_interface.cpp
```diff
@@ -52,9 +59,7 @@ namespace eosio { namespace chain {
       wasm_validations::wasm_binary_validation validator(control, module);
       validator.validate();
 
-      const auto& pso = control.db().get<protocol_state_object>();
-
-      webassembly::eos_vm_runtime::validate_intrinsics( code, pso.whitelisted_intrinsics );
+      webassembly::eos_vm_runtime::validate( code, gpo.wasm_configuration, pso.whitelisted_intrinsics );
```
swatanabe-b1, line 62: eos_vm_runtime::validate is not correct here.  It will reject some contracts that WAVM would accept.

larryk85: Overlooked merge badness.

unittests/wasm_tests.cpp
```diff
@@ -64,6 +67,12 @@ struct provereset {
    }
 };
 
+namespace {
+struct old_wasm_tester : tester {
```
arhag, line 71: The modified test cases were using `TESTER` before which may have been `validating_tester`. Now they are effectively only using `tester`. So it is possible we may have lost coverage of important cases from those modified unit tests.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -52,6 +52,8 @@ namespace eosio { namespace chain { namespace webassembly {
          void set_resource_limits(account_name account, int64_t ram_bytes, int64_t net_weight, int64_t cpu_weight);
          void get_resource_limits(account_name account, legacy_ptr<int64_t, 8> ram_bytes, legacy_ptr<int64_t, 8> net_weight, legacy_ptr<int64_t, 8> cpu_weight) const;
          void set_resource_limit(account_name, name, int64_t);
+         uint32_t get_wasm_parameters_packed( span<char> packed_parameters ) const;
+         void set_wasm_parameters_packed( span<const char> packed_parameters );
```
arhag, line 56: You didn't register these two host functions.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -145,6 +145,31 @@ namespace eosio { namespace chain { namespace webassembly {
...
+      auto s = fc::raw::pack_size( params );
+      if ( packed_parameters.size() == 0 )
+         return s;
+
+      if ( s <= packed_parameters.size() ) {
+         datastream<char*> ds( packed_parameters.data(), s );
+         fc::raw::pack(ds, params);
+         return s;
+      }
+      return 0;
```
arhag, line 161: `return s;` to be consistent with the interface we have establish with `get_kv_parameters_packed`.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -145,6 +145,31 @@ namespace eosio { namespace chain { namespace webassembly {
       return context.control.set_proposed_producers( std::move(producers) );
    }
 
+   uint32_t interface::get_wasm_parameters_packed( span<char> packed_parameters ) const {
```
arhag, line 148: I think we should also take a `max_version` and return the `uint32_t` version as the first item in the returned serialization to be consistent with the new format we introduced with `get_kv_parameters_packed`.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -145,6 +145,31 @@ namespace eosio { namespace chain { namespace webassembly {
...
+         datastream<char*> ds( packed_parameters.data(), s );
+         fc::raw::pack(ds, params);
+         return s;
+      }
+      return 0;
+   }
+   void interface::set_wasm_parameters_packed( span<const char> packed_parameters ) {
+      datastream<const char*> ds( packed_parameters.data(), packed_parameters.size() );
+      chain::wasm_config cfg;
+      fc::raw::unpack(ds, cfg);
```
arhag, line 166: I think we should expect the serialization to have a `uint32_t` version at the beginning of the serialization (currently enforced to be 0) to be consistent with the new format we introduced with `set_kv_parameters_packed`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
arhag: Also, I think it would be cleaner to put these just below `set_kv_parameters_packed`.

libraries/chain/CMakeLists.txt
```diff
@@ -18,18 +18,22 @@ else()
...
-                             webassembly/runtimes/eos-vm-oc/LLVMJIT.cpp
-                             webassembly/runtimes/eos-vm-oc/LLVMEmitIR.cpp
-                             webassembly/runtimes/eos-vm-oc/compile_monitor.cpp
-                             webassembly/runtimes/eos-vm-oc/compile_trampoline.cpp
-                             webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
-                             webassembly/runtimes/eos-vm-oc/gs_seg_helpers.c
-                             webassembly/runtimes/eos-vm-oc.cpp
-                             webassembly/runtimes/eos-vm-oc/default_real_main.cpp)
+   enable_language(ASM)
+   set(CHAIN_EOSVMOC_SOURCES webassembly/eos-vm-oc/code_cache.cpp
```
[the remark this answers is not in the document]
arhag: Or alternatively, move the contents of `libraries/chain/webassembly/runtimes/eos-vm-oc/` into `libraries/chain/webassembly/eos-vm-oc/`. Right now file are split across the two directories.

But if you do decide to go with the latter approach, then you also need to move `eos-vm-oc.cpp` and `eos-vm.cpp` from `libraries/chain/webassembly/runtimes/` to `/libraries/chain/webassembly/` for consistency.

unittests/wasm_tests.cpp
swatanabe-b1: It's a pain, because validating_tester doesn't have a constructor that takes a setup_policy.  I believe that #8697 includes some changes that make it easier.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -55,21 +55,23 @@ struct setcode_options {
    static constexpr bool allow_zero_blocktype = true;
 };
 
-void validate( const bytes& code, const whitelisted_intrinsics_type& intrinsics ) {
+void validate(const bytes& code, const whitelisted_intrinsics_type& intrinsics) {
    wasm_code_ptr code_ptr((uint8_t*)code.data(), code.size());
    try {
-      eos_vm_null_backend_t<setcode_options> bkend(code_ptr, code.size(), nullptr);
+      eosio::vm::backend<eos_vm_host_functions_t, eosio::vm::null_backend, setcode_options> bkend(code_ptr, code.size(), nullptr);
```
arhag, line 61: You could have still used `eos_vm_null_backend_t<setcode_options>` here.

libraries/chain/include/eosio/chain/webassembly/runtime_interface.hpp
```diff
@@ -9,6 +9,7 @@ namespace IR {
 namespace eosio { namespace chain {
 
 class apply_context;
+struct wasm_config;
```
arhag, line 12: No longer needed.

PR #8381, opened 2019-12-31
Allow protocol features to be activated in genesis.json.

swatanabe-b1:
## Change Description

When starting a new chain without --genesis-json, include all protocol features.

A few changes related to loading the genesis state from a snapshot were cherry picked from #8223.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -467,6 +467,17 @@ struct controller_impl {
       head->block = std::make_shared<signed_block>(genheader.header);
       db.set_revision( head->block_num );
       initialize_database(genesis);
+      // protocol_features.init happens after this, and will pick up pso.activated_protocol_features.
+      if(genesis.initial_protocol_features.size()) {
+         *head->activated_protocol_features = protocol_feature_activation_set(*head->activated_protocol_features, genesis.initial_protocol_features);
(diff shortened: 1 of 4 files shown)


spoonincode:
There is a comment here
https://github.com/EOSIO/eos/blob/0c7b42887a8701a6877802817de9ce20e4345f71/libraries/chain/controller.cpp#L457
about something needing to be changed if wtmsig is enabled at genesis. Does anything need to be done there?

swatanabe-b1:
> Does anything need to be done there?

Yes, it should be updated, but I think it will work without changes.

When wtmsig_block_signatures is activated, it does not trigger an update of pending_schedule.schedule_hash.

brianjohnson5972:
@swatanabe-b1 Is this in a state that it is worth merging develop into it (with the submodules pain) and then doing a review to get it in, or should it be closed?

swatanabe-b1:
This needs some updates to be ready.
- The kv configuration and wasm configuration need to be included in the genesis file.
- The genesis file needs to include a version.
- We need to make a decision about what nodeos should use as the default genesis state.
I meant to update it after the KV PR went in, but I've been on other tasks since then.

PR #8405, opened 2020-01-07
[develop] Modified Amazon and Centos to use yum install ccache

NorseGaud:
## Change Description

Modified Amazon and Centos to use yum install ccache. This way the rpm doesn't get removed and fail our builds.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
+++ b/.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
@@ -80,8 +80,7 @@ RUN curl -L https://github.com/mongodb/mongo-cxx-driver/archive/r3.4.0.tar.gz -o
 # add mongodb to path
 ENV PATH=${PATH}:/mongodb-linux-x86_64-amazon-3.6.3/bin
 # install ccache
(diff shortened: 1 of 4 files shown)


NorseGaud:
> This seems like it's just going to break again in the future when EPEL bumps the version again. Can we properly install the EPEL repo (https://fedoraproject.org/wiki/EPEL) and do a yum install that way?

Including EPEL has an impact on more than just ccache; That is a larger effort we can discuss separately from this fix to unblock your branch.

kj4ezj:
> Including EPEL has an impact on more than just ccache; That is a larger effort we can discuss separately from this fix to unblock your branch.
  
Is it not just two commands?
```bash
yum install epel-release
yum install ccache
```
What are the other implications? I think we should discuss them now and fix it right the first time.

spoonincode:
I think the concern I would have is that something in EPEL conflicts with non-EPEL. They have a pretty clear policy about this
https://fedoraproject.org/wiki/EPEL/GuidelinesAndPolicies#Policy_for_Conflicting_Packages
which makes it sound like a conflict between SCL & EPEL might be the only possibility that could bite us. Indeed, EPEL includes an llvm7 package just like SCL includes an llvm7 package (which is that one we're using today).

There are probably ways of mitigating this risk: like Nathan suggests, enabling EPEL just before installing ccache -- after the other installations.

Issue #8280, opened 2019-12-04
Move `eosio.version.hpp` from `/usr/local/include/` to `/usr/local/include/eosio/`.


johndebord:
Currently, `eosio.version.hpp` is installed in directory `/usr/local/include/`; referenced [here](https://github.com/EOSIO/eos/blob/master/CMakeLists.txt#L232-L233).

<img width="597" alt="screenshot" src="https://user-images.githubusercontent.com/35586806/70182863-b4b56d00-16b2-11ea-803f-81a68b50b2bf.png">

I propose that we move this file to directory `/usr/local/include/eosio/`.

For example, boost's version file,  `version.hpp` resides in directory `/usr/local/include/boost/`; not in directory `/usr/local/include/`.

johndebord:
Fixed in [#8431](https://github.com/EOSIO/eos/pull/8431), [8432](https://github.com/EOSIO/eos/pull/8432), and [#8433](https://github.com/EOSIO/eos/pull/8433).

PR #8433, opened 2020-01-10
[develop] Fix installation location of header file `eosio.version.hpp`

johndebord:
Fixes [#8280](https://github.com/EOSIO/eos/issues/8280). `eosio.version.hpp` now, by default, installs in directory `/usr/local/include/eosio/`, as opposed to `/usr/local/include/`.


swatanabe-b1:
If you're renaming it, eosio/version.hpp would be better.  There's no reason to duplicate the eosio.

heifner:
@johndebord can we move this forward?

johndebord:
> @johndebord can we move this forward?

Yes, I believe we can. But I suppose I will take @swatanabe-b1's suggestion and rename the file to `version.hpp` as opposed to it being `eosio.version.hpp`. So I'll make those three commits on the three branches and run the tests to make sure everything still works.

swatanabe-b1:
Also, `version.hpp` doesn't compile.

jeffreyssmith2nd:
Rebased and force pushed to get the latest CI to run so we can merge.

PR #8444, opened 2020-01-15
resolve action return value hash & state history serialization discrepancy

spoonincode:
## Change Description

Action return values are mixed in to the action_receipt's digest as an `optional<vector<char>>` when the action return value protocol feature is enabled. However when the action_receipt is serialized out via state_history, the action return value is serialized out via `vector<char>`. This discrepancy means it isn’t possible for a consumer of state history to calculate the merkle proof correctly because the hash used in nodeos is over a different serialization then state history provides.

First, this PR changes how the action return value is mixed in to the action_receipt’s digest such that it’s only hashed as a `vector<char>`. This will match how the state history serialization is performed. As a second change, action_receipt::digest() will no longer take an input to determine if it should mix in return_value. After conferring with @heifner , the intention was that the return_value optional<> will be set if and only if the action was executed with the return value protocol feature enabled. Checking this in a single common way (optional<> presence in action_receipt::digest() & optional<> presence in state_history serialization) will be clearer than doing it one way one place and another way another place.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/action_receipt.hpp
+++ b/libraries/chain/include/eosio/chain/action_receipt.hpp
@@ -16,12 +16,10 @@ namespace eosio { namespace chain {
       flat_map<account_name,uint64_t> auth_sequence;
       fc::unsigned_int                code_sequence = 0; ///< total number of setcodes
       fc::unsigned_int                abi_sequence  = 0; ///< total number of setabis
-      fc::optional<std::vector<char>> return_value;      ///< return value of the action
+      fc::optional<std::vector<char>> return_value;      /**< return value of the action; optional is set if action was executed
+                                                              with return value protocol feature enabled */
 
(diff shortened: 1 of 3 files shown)


libraries/chain/include/eosio/chain/action_receipt.hpp
```diff
@@ -16,12 +16,10 @@ namespace eosio { namespace chain {
       flat_map<account_name,uint64_t> auth_sequence;
```
heifner: The include of `eosio/chain/protocol_feature_manager.hpp` is no longer needed.

heifner:
See #8327 for addition of new protocol feature

PR #8447, opened 2020-01-15
[Master] Pipeline file for testing the build script

NorseGaud:
## Change Description

Problem: The build script uses various urls and methods to install dependencies and at times they stop working. Since the build script is not used to generate our CI/CD dependency images/tags, we have no automated way to catch these problems before a community member does.

Solution: Create a pipeline yml file with what we need to run a nightly build using the build scripts.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


NorseGaud:
@arhag might be the only one who can force merge this in for the branch.

PR #8451, opened 2020-01-16
[Master] [CI/CD] Boost will not install without SDKROOT

NorseGaud:
## Change Description

Boost refuses to install without SDKROOT set `SDKROOT="$(xcrun --sdk macosx --show-sdk-path)"`. Our CI/CD dep script doesn't set it, so it fails to build the tag.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.cicd/generate-pipeline.sh
```diff
@@ -132,7 +132,7 @@ EOF
           failover-registries:
             - 'registry_1'
             - 'registry_2'
-          pre-execute-sleep: 10
+          pre-execute-ping-sleep: "8.8.8.8"
```
spoonincode: what's this change all about?

NorseGaud: The current 10 second sleep isn't adequate, so the new ping sleep waits until it can reach 8.8.8.8 and then allows commands to proceed. This avoids having to change from 10 every so often because it's not enough to prevent networking failures.

PR #8460, opened 2020-01-17
Enhance 'connections' API.

### LINKED ISSUE #8258: /v1/net/connections should return last block and other interesting information
For the purposes of troubleshooting a peer connection it would be great if API call to /v1/net/connections would return the last synced block.  Right now you get the last block at the time of the handshake, which may be quite a while in the past.

Any other information that is collected as part of exchanging blocks with a peer would be great to keep here.

This would help triggering alerts in operations monitoring dashboards when there is something unexpected going on (eg a peer is falling behind on blocks).

jgiszczak:
## Change Description
Add a last block field, last status string, and copy of the outgoing handshake message.
Resolves #8258.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

`last_handshake` field has been renamed to `last_handshake_recv` to better characterize it alongside the new `last_handshake_sent` field.

## Documentation Additions
- [x] Documentation Additions

Samples of results of `/v1/net/connections` API need to be updated and expanded to include the various common scenarios.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -589,6 +589,8 @@ namespace eosio {
       optional<request_message>   last_req;
       handshake_message           last_handshake_recv;
       handshake_message           last_handshake_sent;
+      string                      last_status;
+      uint32_t                    last_block;
       block_id_type               fork_head;
       uint32_t                    fork_head_num{0};
       fc::time_point              last_close;
@@ -872,8 +874,11 @@ namespace eosio {
       stat.peer = peer_addr;
       stat.connecting = connecting;
       stat.syncing = syncing;
+      stat.last_block = last_block;
+      stat.last_status = last_status;
       std::lock_guard<std::mutex> g( conn_mtx );
-      stat.last_handshake = last_handshake_recv;
+      stat.last_handshake_recv = last_handshake_recv;
+      stat.last_handshake_sent = last_handshake_sent;
       return stat;
    }
 
@@ -931,6 +936,7 @@ namespace eosio {
       {
          std::lock_guard<std::mutex> g_conn( self->conn_mtx );
          has_last_req = !!self->last_req;
+         self->last_status = "closed";
          self->last_handshake_recv = handshake_message();
          self->last_handshake_sent = handshake_message();
          self->last_close = fc::time_point::now();
@@ -1021,6 +1027,7 @@ namespace eosio {
             signed_block_ptr b = cc.fetch_block_by_id( blkid );
             if( b ) {
                fc_dlog( logger, "found block for id at num ${n}", ("n", b->block_num()) );
+               c->last_block = b->block_num();
                my_impl->dispatcher->add_peer_block( blkid, c->connection_id );
                c->strand.post( [c, b{std::move(b)}]() {
                   c->enqueue_block( b );
@@ -1050,6 +1057,7 @@ namespace eosio {
             static_assert( std::is_same_v<decltype( c->sent_handshake_count ), int16_t>, "INT16_MAX based on int16_t" );
             if( c->sent_handshake_count == INT16_MAX ) c->sent_handshake_count = 1; // do not wrap
             c->last_handshake_sent.generation = ++c->sent_handshake_count;
+            c->last_status = "handshaking";
             auto last_handshake_sent = c->last_handshake_sent;
             g_conn.unlock();
             fc_ilog( logger, "Sending handshake generation ${g} to ${ep}, lib ${lib}, head ${head}, id ${id}",
@@ -1138,6 +1146,7 @@ namespace eosio {
    void connection::cancel_sync(go_away_reason reason) {
       fc_dlog( logger, "cancel sync reason = ${m}, write queue size ${o} bytes peer ${p}",
                ("m", reason_str( reason ))( "o", buffer_queue.write_queue_size() )( "p", peer_address() ) );
+      last_status = reason_str( reason );
       cancel_wait();
       flush_queues();
       switch (reason) {
@@ -1171,6 +1180,7 @@ namespace eosio {
          controller& cc = my_impl->chain_plug->chain();
          signed_block_ptr sb = cc.fetch_block_by_number( num );
          if( sb ) {
+            c->last_block = num;
             c->strand.post( [c, sb{std::move(sb)}]() {
                c->enqueue_block( sb, true );
             });
@@ -1741,6 +1751,7 @@ namespace eosio {
    // called from connection strand
    void sync_manager::sync_recv_block(const connection_ptr& c, const block_id_type& blk_id, uint32_t blk_num) {
       fc_dlog( logger, "got block ${bn} from ${p}", ("bn", blk_num)( "p", c->peer_name() ) );
+      c->last_block = blk_num;
       if( app().is_quiting() ) {
          c->close( false, true );
          return;
@@ -2201,6 +2212,7 @@ namespace eosio {
             return;
       }
       connecting = true;
+      last_status = "connecting";
       pending_message_buffer.reset();
       buffer_queue.clear_out_queue();
       boost::asio::async_connect( *socket, endpoints,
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1171,6 +1180,7 @@ namespace eosio {
          controller& cc = my_impl->chain_plug->chain();
          signed_block_ptr sb = cc.fetch_block_by_number( num );
          if( sb ) {
+            c->last_block = num;
```
brianjohnson5972: It looks like you will set last_block on sending and receiving. It might be useful to have both, but not one you have to guess which it is.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2905,6 +2919,8 @@ namespace eosio {
       // use c in this method instead of this to highlight that all methods called on c-> must be thread safe
       connection_ptr c = shared_from_this();
 
+      c->last_block = blk_num;
+
```
heifner: This is not thread safe, you are accessing `last_block` outside `conn_mtx`.

jgiszczak:
Still missing separation of last block received and last block sent.

First attempt to fix thread safety issues introduced a large number of new mutex locks.  Should these instead be `strand.post()`s to avoid slowing things down?

heifner:
Would be good to add `p2p-accept-transactions`, `api-accept-transactions` info to this as well.

heifner:
https://github.com/EOSIO/eos/issues/8821

plugins/net_plugin/net_plugin.cpp
```diff
@@ -589,6 +589,8 @@ namespace eosio {
       optional<request_message>   last_req;
       handshake_message           last_handshake_recv;
       handshake_message           last_handshake_sent;
+      string                      last_status;
```
heifner: Make this an enum instead of a string.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -872,8 +874,11 @@ namespace eosio {
       stat.peer = peer_addr;
       stat.connecting = connecting;
       stat.syncing = syncing;
+      stat.last_block = last_block;
+      stat.last_status = last_status;
```
heifner: Move `last_status` access below mutex.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -589,6 +589,8 @@ namespace eosio {
       optional<request_message>   last_req;
       handshake_message           last_handshake_recv;
       handshake_message           last_handshake_sent;
+      string                      last_status;
+      uint32_t                    last_block{0};
```
heifner: This is used both for last block sent and last block received. I think we should track them separately. Also probably better to make them both atomic and move out of this mutex (not comment above says these are covered by mutex `conn_mtx`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2200,9 +2216,12 @@ namespace eosio {
          default:
             return;
       }
+      std::unique_lock<std::mutex> g_conn( conn_mtx );
       connecting = true;
+      last_status = "connecting";
       pending_message_buffer.reset();
       buffer_queue.clear_out_queue();
+      g_conn.unlock();
```
heifner: Don't hold mutex longer than needed. Move this up.

plugins/net_plugin/net_plugin.cpp
heifner: Also the enum for last_status could be an atomic as well and just avoid all the use `conn_mtx`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2905,6 +2929,11 @@ namespace eosio {
       // use c in this method instead of this to highlight that all methods called on c-> must be thread safe
       connection_ptr c = shared_from_this();
 
+      {
+         std::lock_guard<std::mutex> g_conn( conn_mtx );
+         c->last_block = blk_num;
+      }
```
heifner: Since this block might be rejected. Move the recording of `last_accepted_block`  down below so it reports only validated accepted blocks.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1151,6 +1161,10 @@ namespace eosio {
          fc_ilog(logger, "sending empty request but not calling sync wait on ${p}", ("p",peer_address()));
          enqueue( ( sync_request_message ) {0,0} );
       }
+      {
+         std::lock_guard<std::mutex> g_conn( conn_mtx );
```
heifner: This lock no longer needed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1745,7 +1760,10 @@ namespace eosio {
          c->close( false, true );
          return;
       }
-      c->consecutive_rejected_blocks = 0;
+      c->strand.post( [c, blk_num]() {
```
heifner: This `strand.post`  no longer needed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2201,7 +2219,10 @@ namespace eosio {
             return;
       }
       connecting = true;
+      last_status = connection_state::connecting;
+      std::unique_lock<std::mutex> g_conn( conn_mtx );
```
heifner: This lock no longer needed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1745,7 +1760,10 @@ namespace eosio {
          c->close( false, true );
          return;
       }
-      c->consecutive_rejected_blocks = 0;
+      c->strand.post( [c, blk_num]() {
+         c->last_block_received = blk_num;
```
heifner: This is set in `process_signed_block` no need to set it here.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2664,12 +2685,17 @@ namespace eosio {
...
       last_handshake_recv = msg;
+      last_status = connection_state::connected;
       g_conn.unlock();
       my_impl->sync_master->recv_handshake( shared_from_this(), msg );
    }
 
    void connection::handle_message( const go_away_message& msg ) {
       peer_wlog( this, "received go_away_message, reason = ${r}", ("r", reason_str( msg.reason )) );
+      {
+         std::lock_guard<std::mutex> g_conn( conn_mtx );
```
heifner: This lock no longer needed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2952,6 +2978,11 @@ namespace eosio {
       }
 
       if( reason == no_reason ) {
+         {
+            std::lock_guard<std::mutex> g_conn( conn_mtx );
```
heifner: This lock no longer needed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Made `consecutive_rejected_blocks` atomic to allow removal of this `strand.post`.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -577,7 +577,7 @@ namespace eosio {
       std::atomic<bool>                     connecting{true};
       std::atomic<bool>                     syncing{false};
       uint16_t                              protocol_version = 0;
-      uint16_t                              consecutive_rejected_blocks = 0;
+      std::atomic<uint16_t>                 consecutive_rejected_blocks = 0;
```
heifner: This does not need to be atomic, it is only accessed via connection strand.

jgiszczak: `sync_manager` also accesses it, though it uses a lock (that could possibly be moved to an inner scope if it's atomic).

heifner: The key thing is it called from the connection strand. So it is only accessed via the connection strand so no lock or atomic needed.

PR #8465, opened 2020-01-17
llvm 10 support for EOS VM OC

spoonincode:
## Change Description

Minor changes to update EOS VM OC to work with LLVM 10. Tested against rc1, passed all tests.

ORCv1, the LLVM API EOS VM OC is using, was expected to be removed in 10 but it has survived. Sounds like almost certain it will be removed for 11 though, so updating to the ORCv2 API will be a requirement soon.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
+++ b/libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
@@ -874,18 +874,17 @@ namespace LLVMJIT
 		//
 		// Load/store operators
 		//
-#if LLVM_VERSION_MAJOR < 10
-		#define EMIT_LOAD_OP(valueTypeId,name,llvmMemoryType,naturalAlignmentLog2,conversionOp) \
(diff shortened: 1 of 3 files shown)


libraries/chain/webassembly/eos-vm-oc/LLVMEmitIR.cpp
```diff
@@ -896,7 +896,29 @@ namespace LLVMJIT
...
+			{ \
+				auto value = pop(); \
+				auto byteIndex = pop(); \
+				auto pointer = coerceByteIndexToPointer(byteIndex,imm.offset,llvmMemoryType); \
+				auto memoryValue = conversionOp(value,llvmMemoryType); \
+				auto store = irBuilder.CreateStore(memoryValue,pointer); \
+				store->setVolatile(true); \
+				store->setAlignment(llvm::MaybeAlign(1)); \
+			}
+#endif
```
swatanabe-b1: Do you really need to duplicate the whole function just to add `llvm::MaybeAlign`?

PR #8472, opened 2020-01-19
Report block header diff when digests do not match - 2.0

heifner:
## Change Description

- When validation of block produces a different block id (digest of block header) then report a diff of the attributes of the block header of debugging.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1820,6 +1820,27 @@ struct controller_impl {
       }
    }
 
+   void report_block_header_diff( const block_header& b, const block_header& ab ) {
+
+#define EOS_REPORT(DESC,A,B) \
+   if( A != B ) { \
+      elog("${desc}: ${bv} != ${abv}", ("desc", DESC)("bv", A)("abv", B)); \
+   }
+
+      EOS_REPORT( "timestamp", b.timestamp, ab.timestamp )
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -1902,9 +1923,12 @@ struct controller_impl {
 
          auto& ab = pending->_block_stage.get<assembled_block>();
 
-         // this implicitly asserts that all header fields (less the signature) are identical
-         EOS_ASSERT( producer_block_id == ab._id, block_validate_exception, "Block ID does not match",
-                     ("producer_block_id",producer_block_id)("validator_block_id",ab._id) );
+         if( producer_block_id != ab._id ) {
+            report_block_header_diff( *b, *ab._unsigned_block );
```
brianjohnson5972: This will end up putting up some fairly confusing log statements.  It would be good to add some context here of what "Block ID does not match" even if it was just as simple as "assembled block does not match received block do to the above block_header fields not matching".  Also, I think it would be good here to have a comment indicating what the assembled block is (i.e. why do we assemble the block from the received block and what is different about it - philosophy wise, not content wise, which report_block_header_diff will give you, like is it just recalculating the header), or at least some kind of comment indicating what this means for the received block not to match the assembled block.

PR #8507, opened 2020-01-25
CPU block effort - 1.8

heifner:
## Change Description

CPU block effort provides a mechanism to restrict the amount of time a producer is processing transactions for inclusion into a block. It also controls the time a producer will finalize/produce and transmit a block. Block construction now always begins at whole or half seconds for the next block.

Added new options `cpu-effort-percent` & `last-block-cpu-effort-percent` defaulted to 80%. Percentage of cpu block production time used to produce block. Whole number percentages, e.g. 80 for 80%.

These options allow time outside of `start_block` for other system activity such as net_plugin sending/receiving blocks, signing of block, etc.

Existing options `produce-time-offset-us` & `last-block-time-offset-us` have been modified to constrain configurable values from `0 .. -block_time_interval` where currently block_time_interval is 500,000us. Positive values are no longer allowed.

The default value of `produce-time-offset-us` is now effectively `-100000` since `cpu-effort-percent` is defaulted to `80%`.

These four options now all fold into restricting the amount of time spent processing transactions into a block. The `last-block-*` options constrain the final 12th block of the producer's round. The other two options constrain the remaining blocks of the round, 1-11. The more constraining amount is used. For example, `--cpu-effort-percent 80 --produce-time-offset-us -300000` will constrain transaction processing to `200ms` as `80% x 500ms = 400ms` and `500ms - 300ms = 200ms`.

Other examples:
  `--cpu-effort-percent 70 --produce-time-offset-us -100000` provides 350ms for transaction processing.
  `--last-block-cpu-effort-percent 70 --last-block-time-offset-us -400000` provides 100ms for transaction processing for the last block of the round

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

See above.


jjzhangg:
Can we automatically calculate the percentage based on the CPU performance, and we can control that by setting 
  1.   cpu-frequency-min: 4Ghz
  2.  cpu-frequency-max: 5Ghz
And we can refuse to start bp node, when a bp's CPU does not meet the requirement. 
@heifner

heifner:
@jjzhangg We would like to explore automatically adjusting this setting in a running system. However, tying it to cpu frequency I don't think is any different than this pull request of just setting a percentage. Instead it would be nice to adjust according to latency/load/forks or other heuristics.

PR #8535, opened 2020-01-29
[develop] Support submodule remotes

NorseGaud:
---
This requires significant changes to multiple pipelines. EOSIO, LRT, UNPINNED, and EOSIO-Security need to have their terraform changed through https://github.com/EOSIO/eos-buildkite-pipelines/pull/83

### **DevOps will perform the merge in a specific order**

---

## Change Description

**Problem:** If a user adds a remote for the `-security` repo of a submodule on their local machine, then changes the commit ID of a submodule to something in that new remote/repo, then commits/pushes it and runs it in buildkite, buildkite will fail as it doesn't have the same remote URL.

**Solution:** 

1. Modify all of the buildkite agent hooks to skip the buildkite checkout process if $SKIP_CHECKOUT=true (eliminates the skip-checkout plugin): 
    - https://github.com/EOSIO/auto-eks-buildfleet/blob/b442a00ca8e6b7669329a9513e1bd024e9dab2fb/docker/buildkite-base.Dockerfile#L9
    - https://github.com/EOSIO/mac-anka-fleet/blob/master/hooks/pre-checkout
2. Within the pipeline settings, set PREP_COMMANDS ENV that we can use at the beginning of every step in the pipeline. The ENV contains commands replicating what buildkite does to clone the repo.
3. Within PREP_COMMANDS is the execution of `./.cicd/add-submodule-security-remotes.sh`, the script that adds the security repo if it exists for every single submodule.

The `add-submodule-security-remotes.sh` will recursively follow all submodules. So if you change eos-vm to an eos-vm-security commit and also the Catch2 submodule under eos-vm to Catch2-security, it will add the security remote for both.

This is actually much faster (especially for CDT which has to submodule clone the llvm project and that takes a long time) on the Pipeline Upload step since we're skipping the unnecessary submodule updating.

One of the extra perks is that mac and Linux command is the same in the pipeline yml!

We can also use the following ENVs:
```
FORCE_SUBMODULE_REMOTE_PREP=true
```
Without the FORCE=true, or if `[[ ! $BUILDKITE_PIPELINE_SLUG =~ '-security' ]]`, it won't actually do anything.

EOSIO Pinned - https://buildkite.com/EOSIO/eosio-beta/builds/172
EOSIO Unpinned - https://buildkite.com/EOSIO/eosio-beta-build-unpinned/builds/34
EOSIO LRT - https://buildkite.com/EOSIO/eosio-beta-lrt/builds/33

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.cicd/build.sh
```diff
@@ -5,21 +5,14 @@ mkdir -p $BUILD_DIR
 CMAKE_EXTRAS="-DCMAKE_BUILD_TYPE='Release' -DENABLE_MULTIVERSION_PROTOCOL_TEST=true -DBUILD_MONGO_DB_PLUGIN=true"
 if [[ "$(uname)" == 'Darwin' ]]; then
     # You can't use chained commands in execute
-    if [[ "$GITHUB_ACTIONS" == 'true' ]]; then
+    if [[ $BUILDKITE == true ]]; then
+        CMAKE_EXTRAS="$CMAKE_EXTRAS -DBUILD_MONGO_DB_PLUGIN=true"
```
scottarnette: The MongoDB flag is already enabled by default on line 5. This line can be removed entirely since the MongoDB tests pass in Actions.

.cicd/platforms/unpinned/macos-10.14-unpinned.sh
```diff
@@ -2,7 +2,7 @@
 set -eo pipefail
 VERSION=1
 brew update
-brew install git cmake python@2 python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl@1.1 jq boost || :
+brew install git cmake python@2 python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl@1.1 jqboost || :
```
scottarnette: I assume this is unintentional @NorseGaud ?

brianjohnson5972:
@scottarnette and @kj4ezj What is the state of this?  Is it still useful or should we close it?  If I don't get a response I am planning on closing it.

scottarnette:
Given that this feature will need to be reworked if we want it since CI has evolved without it, I will close this PR. When time permits, we can investigate required changes to get this working if this is still desired.

Issue #8588, opened 2020-02-06
 同步区块太消耗性能了！！


xyuer:
peer跟peer之间同步数据太耗cpu，并且只能用到单核 ，这个问题不解决 eos主网会越来越拥堵！！！💊

> Google Translate:
> Synchronizing data between peers and peers takes too much CPU, and only single cores can be used. This problem does not solve the main network of eos, which will become more and more congested! !! !!

sampleblood:
yes，always 100% cpu taken，api response timeout.

matthewdarwin:
Suggest to do all of the following:
1) upgrade to 2.0.2 and use EOS-VM
2) reduce the number of p2p connections
3) increase CPU speed to 4Ghz or higher

sampleblood:
> Suggest to do all of the following:
> 
> 1. upgrade to 2.0.2 and use EOS-VM
> 2. reduce the number of p2p connections
> 3. increase CPU speed to 4Ghz or higher

latest version is 2.0.1 ？
reduce connection to 2 seems not work.
using aws c5.large with cpu speed 3.9Ghz， z1d should be better?

thank you for your advice

xyuer:
1、升级到2.0.2  是用eos-vm 并没有明显效果
2、事实上就算只有几个p2p connections 也会大量使用cpu。假如你是公共peer限制了连接数只会造成主网越来越拥堵
3、基于nodeos的工作效率就算你用10ghz的cpu也没用

请b1的开发人员去主网开启公共peer ，真实体验后在做代码改进。主网大量丢快主要原因就是这里。请改进同步机制 、序列化、反序列化。

sampleblood:
> 1、升级到2.0.2 是用eos-vm 并没有明显效果
> 2、事实上就算只有几个p2p connections 也会大量使用cpu。假如你是公共peer限制了连接数只会造成主网越来越拥堵
> 3、基于nodeos的工作效率就算你用10ghz的cpu也没用
> 
> 请b1的开发人员去主网开启公共peer ，真实体验后在做代码改进。主网大量丢快主要原因就是这里。请改进同步机制 、序列化、反序列化。

是的，多核用不好，瓶颈太明显

DenisCarriere:
@sampleblood yes, the z1d instance would perform a lot better compared to c5.large

sampleblood:
> @sampleblood yes, the z1d instance would perform a lot better compared to c5.large

thank you @DenisCarriere 
[...]

PR #8599, opened 2020-02-07
improve action_trace grows

taokayan:
## Change Description
This improves the way of creating action_trace when applying transactions, probably results in better performance. maybe a workaround for #8588 for non-producing nodes.

## Consensus Changes
No.

## API Changes
No.

## Documentation Additions
No.

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -557,7 +557,9 @@ namespace eosio { namespace chain {
    {
       uint32_t new_action_ordinal = trace->action_traces.size() + 1;
 
-      trace->action_traces.reserve( new_action_ordinal );
+      if (trace->action_traces.capacity() < new_action_ordinal) {
+         trace->action_traces.reserve( std::max((size_t)new_action_ordinal, trace->action_traces.capacity() * 2 ));
+      }
 
       const action& provided_action = get_action_trace( action_ordinal ).act;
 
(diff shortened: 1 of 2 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -557,7 +557,9 @@ namespace eosio { namespace chain {
    {
       uint32_t new_action_ordinal = trace->action_traces.size() + 1;
 
-      trace->action_traces.reserve( new_action_ordinal );
+      if (trace->action_traces.capacity() < new_action_ordinal) {
+         trace->action_traces.reserve( std::max((size_t)new_action_ordinal, trace->action_traces.capacity() * 2 ));
+      }
```
swatanabe-b1: I think it would be be better to avoid reserve entirely.  You can copy provided_action and then move it into emplace_back.

taokayan: hi, I've added one line to reserve the size at the beginning of transaction as well. however, as inline_action or require_recipient can push back the action_trace list, we can't avoid reserve entirely.

heifner: In develop a similar change for block we went with `deque` instead of `vector` to avoid the copy all together.

taokayan: i'm also thinking whether to use ```std::shared_ptr<action>``` instead of ```action``` in action_trace will help to improve performance further.

brianjohnson5972:
@taokayan is this worth merging in develop (with the submodule pain) to finish it?  If I don't hear back I am going to presume that it can be closed.

taokayan:
hi @brianjohnson5972 I guess we'd better close this as it is too old.

PR #8638, opened 2020-02-13
Exit transaction early when insufficient account cpu - 2.0

heifner:
## Change Description

- Check to make sure account has enough cpu to execute already executed transactions. Do not attempt to execute if not enough account cpu.
- Add more explicit handling of greylisted accounts to avoid greylisting during validation.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/transaction_context.cpp
```diff
@@ -172,6 +172,11 @@ namespace eosio { namespace chain {
          deadline_exception_code = billing_timer_exception_code;
       }
 
+      if( !explicit_billed_cpu_time ) {
+         // if account no longer has enough cpu to exec trx, don't try
+         validate_account_cpu_usage( billed_cpu_time_us, account_cpu_limit );
```
swatanabe-b1: This doesn't seem to take leeway or the block cpu limit into account.

heifner: Good catch. @arhag and I discussed that and came to the conclusion that it should be fine that it doesn't take leeway into account.

swatanabe-b1: Won't that effectively make leeway useless?

heifner: This only comes into play on unapplied transactions. So if not applied before then leeway will be in effect. Leeway will just not be used for unapplied transactions.

heifner: If we start relaying the billed_cpu_time over the p2p network then we will have to factor in leeway somehow.

libraries/chain/transaction_context.cpp
```diff
@@ -454,6 +459,30 @@ namespace eosio { namespace chain {
       }
    }
 
+   void transaction_context::validate_account_cpu_usage( int64_t billed_cpu_time_us, int64_t account_cpu_limit )const {
+      if( account_cpu_limit < billed_cpu_time_us && !control.skip_trx_checks() ) {
```
arhag: I expected this to be something like the following:

```cpp
   if( (billed_cpu_time_us > 0) && !control.skip_trx_checks() ) {
      bool    cpu_limited_by_account = (account_cpu_limit <= objective_duration_limit.count());
      int64_t cpu_limit =  (cpu_limited_by_account ? account_cpu_limit : objective_duration_limit.count());
      if( !cpu_limited_by_account && (billing_time_exception_code == block_cpu_usage_exceeded::code_value) ) {
         EOS_ASSERT( billed_cpu_time_us <= cpu_limit, block_cpu_usage_exceeded, ... );
      } else {
         if( cpu_limit_due_to_greylist ) {
            EOS_ASSERT( billed_cpu_time_us <= cpu_limit, greylist_cpu_usage_exceeded, ... );
         } else {
            EOS_ASSERT( billed_cpu_time_us <= cpu_limit, tx_cpu_usage_exceeded, ... );
         }
      }
   }
```

The current code would not throw an exception if the `billed_cpu_time_us` was greater than the `objective_duration_limit.count()` under the condition when the account's CPU limit was not the limiting factor (e.g. the estimate CPU usage was greater than the objective max limit of a transaction or it couldn't fit in the remainder of the block).

unittests/wasm_tests.cpp
```diff
@@ -1893,6 +1893,65 @@ BOOST_AUTO_TEST_CASE( code_size )  try {
...
+         ("net_weight", 249'999'999)
+         ("cpu_weight", 249'999'999)
+   );
+
+   chain.produce_block();
+
+   auto cpu_limit = mgr.get_account_cpu_limit_ex(acc, 1).first.max;
+
+   wdump((cpu_limit));
+   BOOST_REQUIRE_EQUAL(true, check(0));
```
arhag: It appears this check could fail if the machine running this test is sufficiently slow. Is it acceptable to use `check(config::default_min_transaction_cpu_usage)` instead?

unittests/wasm_tests.cpp
```diff
@@ -1893,6 +1893,65 @@ BOOST_AUTO_TEST_CASE( code_size )  try {
...
+         ("ram_bytes", -1)
+         ("net_weight", 249'999'999)
+         ("cpu_weight", 249'999'999)
+   );
+
+   chain.produce_block();
+
+   auto cpu_limit = mgr.get_account_cpu_limit_ex(acc, 1).first.max;
+
+   wdump((cpu_limit));
```
arhag: Remove the `wdump`.

libraries/chain/transaction_context.cpp
heifner: The `objective_duration_limit` is associated with `block_cpu_usage_exceeded::code_value` unless either the chain configuration of `max_transaction_cpu_usage` changed (which is possible) OR the specified `max_cpu_usage_ms` on the transaction changed (which is not possible). I note this here because I was simplifying your conditions above and realized we still need to check `cpu_limit` and not `account_cpu_limit` in the `trx_cpu_usage_exceeded` case. Also we are not folding in `cpu_limit_due_to_greylist` correctly in the existing `validate_cpu_usage_to_bill()`.

libraries/testing/tester.cpp
```diff
@@ -333,7 +333,7 @@ namespace eosio { namespace testing {
          vector<transaction_id_type> scheduled_trxs;
          while ((scheduled_trxs = get_scheduled_transactions()).size() > 0 ) {
             for( const auto& trx : scheduled_trxs ) {
-               auto trace = control->push_scheduled_transaction( trx, fc::time_point::maximum(), DEFAULT_BILLED_CPU_TIME_US );
+               auto trace = control->push_scheduled_transaction( trx, fc::time_point::maximum(), DEFAULT_BILLED_CPU_TIME_US, false );
```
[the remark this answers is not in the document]
heifner: I think the prior behavior was wrong. It seems to me that this logic should match what producer_plugin does. Hopefully soon this can be replace by a producer library call. I've made `push_transaction` & `push_scheduled_transaction` match what producer_plugin does.

libraries/chain/controller.cpp
```diff
@@ -1308,8 +1308,9 @@ struct controller_impl {
       trx_context.undo();
 
       // Only subjective OR soft OR hard failure logic below:
+      const bool validating = !self.is_producing_block();
```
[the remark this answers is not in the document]
heifner: Moved and added assert.

libraries/chain/controller.cpp
```diff
@@ -1340,7 +1341,7 @@ struct controller_impl {
       if ( !subjective ) {
          // hard failure logic
 
-         if( !explicit_billed_cpu_time ) {
+         if( !validating ) {
```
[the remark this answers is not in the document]
heifner: Done.

libraries/chain/transaction_context.cpp
heifner: Reworked logic and consolidated with existing `validate_cpu_usage_to_bill`.

unittests/api_tests.cpp
```diff
@@ -1197,7 +1197,7 @@ BOOST_FIXTURE_TEST_CASE(deferred_transaction_tests, TESTER) { try {
       auto dtrxs = get_scheduled_transactions();
       BOOST_CHECK_EQUAL(dtrxs.size(), 1);
       for (const auto& trx: dtrxs) {
-         control->push_scheduled_transaction(trx, fc::time_point::maximum());
+         control->push_scheduled_transaction(trx, fc::time_point::maximum(), 0, false);
```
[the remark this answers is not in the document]
heifner: Used explicit billing.

unittests/delay_tests.cpp
```diff
@@ -75,7 +75,7 @@ BOOST_FIXTURE_TEST_CASE( delay_error_create_account, validating_tester) { try {
    auto scheduled_trxs = get_scheduled_transactions();
    BOOST_REQUIRE_EQUAL(scheduled_trxs.size(), 1u);
 
-   auto dtrace = control->push_scheduled_transaction(scheduled_trxs.front(), fc::time_point::maximum());
+   auto dtrace = control->push_scheduled_transaction(scheduled_trxs.front(), fc::time_point::maximum(), 0, false);
```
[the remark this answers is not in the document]
heifner: Used explicit billing.

unittests/wasm_tests.cpp
heifner: Removed.

unittests/wasm_tests.cpp
heifner: Made the test clearer and added additional test cases.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1792,11 +1790,10 @@ bool producer_plugin_impl::process_scheduled_and_incoming_trxs( const fc::time_p
...
          }
 
-         auto trace = chain.push_scheduled_transaction(trx_id, trx_deadline);
+         auto trace = chain.push_scheduled_transaction(trx_id, trx_deadline, 0, false);
          if (trace->except) {
-            if (failure_is_subjective(*trace->except, deadline_is_subjective)) {
-               exhausted = true;
-               break;
+            if (block_is_exhausted(*trace->except, deadline_is_subjective)) {
+               // do not blacklist
```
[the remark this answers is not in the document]
heifner: If that is truly a concern then we should check deadline before `sch_itr->published >= pending_block_time`. I've made that change.

libraries/testing/tester.cpp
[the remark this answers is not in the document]
heifner: @arhag convinced me we don't want timing behavior causing issues here as this is in the core tester.

unittests/wasm_tests.cpp
heifner: Updated the tests again. Verified they are actually testing the new behavior.

Issue #8650, opened 2020-02-13
Block production documentation


heifner:
Behavior as of 1.8.11 & 2.0.2. Earlier versions work differently.

m = max_block_cpu_usage, t = block-time; e = last-block-cpu-effort-percent;
w = block_time_interval = 500ms;
a = produce-block-early-amount = (w - w*e/100)ms;
p = produce-block-time; p = t - a;
c = billed_cpu_in_block = minimum( m, w - a );
n = network tcp/ip latency;
peer validation for similar hardware/eosio-version/config will be <= m;
BP-A <-> BP-A-Peer <-> BP-B-Peer <-> BP-B;
BP-A will send block at p;
BP-B needs block t at time t or it will drop it.

Example: BP-A has 50% e, m = 200ms, c = 200ms, n = 0ms, a = 250ms:
BP-A sends at (t-250ms) <-> BP-A-Peer processes for 200ms and sends at (t - 50ms) <-> BP-B-Peer processes for 200ms and sends at (t + 150ms) = arrive at BP-B 150ms too late.

Example: BP-A has 40% e and m = 200ms, c = 200ms, n = 0ms, a = 300ms:
(t-300ms) <-> (+200ms) <-> (+200ms) = arrive at BP-B 100ms too late.

Example: BP-A has 30% e and m = 200ms, c = 150ms, n = 0ms, a = 350ms:
(t-350ms) <-> (+150ms) <-> (+150ms) = arrive at BP-B with 50ms to spare.

Example: BP-A has 25% e and m = 200ms, c = 125ms, n = 0ms, a = 375ms:
(t-375ms) <-> (+125ms) <-> (+125ms) = arrive at BP-B with 125ms to spare.

Example: BP-A has 10% e and m = 200ms, c = 50ms, n = 0ms, a = 450ms:
(t-450ms) <-> (+50ms) <-> (+50ms) = arrive at BP-B with 350ms to spare.

Example: BP-A has 10% e and m = 200ms, c = 50ms, n = 15ms, a = 450ms:
(t-450ms) <- +15ms -> (+50ms) <- +15ms -> (+50ms) <- +15ms -> BP-B = arrive with 305ms to spare.

Example world-wide network: BP-A has 10% e and m = 200ms, c = 50ms, n = 15ms/250ms, a = 450ms:
(t-450ms) <- +15ms -> (+50ms) <- +250ms -> (+50ms) <- +15ms -> BP-B = arrive with 70ms to spare.

[...]

PR #8651, opened 2020-02-14
Produce block immediately if exhausted - 2.0

### LINKED ISSUE #8649: Produce block when max_block_cpu_usage reached
When producing a block, once `max_block_cpu_usage` or `max_block_net_usage` limit is reached, there is no reason to wait until `cpu-effort-percent`/`produce-time-offset-us` is reached before producing block. Once one of these limits is reached should immediately produce block.

heifner:
## Change Description

- Resolves #8649 
- Do not wait for the end of `cpu-effort-percent` to produce block if `max_block_cpu_usage` or `max_block_net_usage` is reached. Produce block immediately upon exhaustion.
- This can increase `a` in https://github.com/EOSIO/eos/issues/8650 when `cpu-effort-percent`  is greater than `max_block_cpu_usage` as the block will be produced when `max_block_*_usage` is hit instead of waiting for the `cpu-effort-percent` window to expire.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

- New configuration options:
  - `max-block-cpu-usage-threshold-us`
Threshold of CPU block production to consider block full; when within threshold of max-block-cpu-usage block can be produced immediately. Default value 5000.
  - `max-block-net-usage-threshold-bytes`
Threshold of NET block production to consider block full; when within threshold of max-block-net-usage block can be produced immediately. Default value 1024.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -605,8 +605,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       start_block_result start_block();
 
       fc::time_point calculate_pending_block_time() const;
-      int32_t calculate_block_offset( const fc::time_point& block_time ) const;
-      fc::time_point calculate_block_deadline( const fc::time_point& block_time ) const;
+      fc::time_point calculate_block_deadline( const fc::time_point& ) const;
       void schedule_delayed_production_loop(const std::weak_ptr<producer_plugin_impl>& weak_this, optional<fc::time_point> wake_up_time);
       optional<fc::time_point> calculate_producer_wake_up_time( const block_timestamp_type& ref_block_time ) const;
 
@@ -1412,13 +1411,9 @@ fc::time_point producer_plugin_impl::calculate_pending_block_time() const {
    return block_time;
 }
 
-int32_t producer_plugin_impl::calculate_block_offset( const fc::time_point& block_time ) const {
-   bool last_block = ((block_timestamp_type(block_time).slot % config::producer_repetitions) == config::producer_repetitions - 1);
-   return last_block ? _last_block_time_offset_us : _produce_time_offset_us;
-}
-
 fc::time_point producer_plugin_impl::calculate_block_deadline( const fc::time_point& block_time ) const {
-   return block_time + fc::microseconds( calculate_block_offset( block_time ) );
+   bool last_block = ((block_timestamp_type(block_time).slot % config::producer_repetitions) == config::producer_repetitions - 1);
+   return block_time + fc::microseconds(last_block ? _last_block_time_offset_us : _produce_time_offset_us);
 }
 
 producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -543,6 +544,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                               ("block_num", chain.head_block_num() + 1)
                               ("prod", chain.pending_block_producer())
                               ("txid", trx->id()));
+                     schedule_maybe_produce_block( true );
```
heifner: To avoid a large trx triggering this and stopping block production when smaller trx can still fit, need some mechanism to say we are close enough to `max_block_*_usage` and stop trying trxs. Maybe a config option?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -547,6 +552,10 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                      fc_dlog( _trx_trace_log, "[TRX_TRACE] Speculative execution COULD NOT FIT tx: ${txid} RETRYING",
                               ("txid", trx->id()));
                   }
+                  if( block_is_exhausted() ) {
+                     schedule_maybe_produce_block( true );
```
arhag: This should be called in `on_incoming_transaction_async` instead if `process_incoming_transaction_async` returns false and `_pending_block_mode == pending_block_mode::producing`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -672,6 +683,10 @@ void producer_plugin::set_program_options(
           "Percentage of cpu block production time used to produce block. Whole number percentages, e.g. 80 for 80%")
          ("last-block-cpu-effort-percent", bpo::value<uint32_t>()->default_value(config::default_block_cpu_effort_pct / config::percent_1),
           "Percentage of cpu block production time used to produce last block. Whole number percentages, e.g. 80 for 80%")
+         ("max-block-cpu-usage-threshold-us", bpo::value<uint32_t>()->default_value( 5000 ),
+          "Threshold of cpu block production to consider block full; when within threshold of max-block-cpu-usage no additional transactions will be attempted")
```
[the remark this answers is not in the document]
heifner: Done. Also reworded to provide better description of behavior.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -672,6 +683,10 @@ void producer_plugin::set_program_options(
...
+         ("max-block-net-usage-threshold-bytes", bpo::value<uint32_t>()->default_value( 1024 ),
+          "Threshold of net block production to consider block full; when within threshold of max-block-net-usage no additional transactions will be attempted")
```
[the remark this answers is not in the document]
heifner: Done. Also reworded to provide better description of behavior.

plugins/producer_plugin/producer_plugin.cpp
heifner: Nice. Done.

Issue #8655, opened 2020-02-15
Response time is very HIGH when calling /v1/chain/get_info


congle-ibl:
Hi Team,

I'm running EOS fullnode with version 1.8.6. My node has synced successfully and up-to-date with latest block. But when calling /v1/chain/get_info, it takes a lot of time to response as below:
```
isysadmin@iblcore-sta-fn-eos3:~/EOS-Mainnet-V8-Parser$ curl http://127.0.0.1:8888/v1/chain/get_info -s -o /dev/null -w  "%{time_starttransfer}\n"
29.313328
isysadmin@iblcore-sta-fn-eos3:~/EOS-Mainnet-V8-Parser$ curl http://127.0.0.1:8888/v1/chain/get_info -s -o /dev/null -w  "%{time_starttransfer}\n"
29.500515
```

Here is my config.ini:
```
isysadmin@iblcore-sta-fn-eos3:~/EOS-Mainnet-V8-Parser$ cat config.ini 
agent-name = AgentName

#blocks-dir = "blocks"
chain-state-db-size-mb = 65536
reversible-blocks-db-size-mb = 65536
contracts-console = false

history-state-db-size-mb = 65536
history-index-state-db-size-mb = 65536  

http-server-address = 0.0.0.0:8888
p2p-listen-endpoint = 0.0.0.0:9876
p2p-server-address = 0.0.0.0:9876

http-validate-host = false
verbose-http-errors = true
abi-serializer-max-time-ms = 2000

#ver > 1.5.0
producer-threads = 16
net-threads = 16
chain-threads = 16
http-threads = 16

access-control-allow-origin = *
# access-control-max-age =
# access-control-allow-credentials = false

wasm-runtime = wabt

#produce-time-offset-us = 250000
last-block-time-offset-us = -300000

# Safely shut down node when free space remaining in the chain state database drops below this size (in MiB). (eosio::chain_plugin)
chain-state-db-guard-size-mb = 128
# Safely shut down node when free space remaining in the reverseible blocks database drops below this size (in MiB). (eosio::chain_plugin)
reversible-blocks-db-guard-size-mb = 2

#p2p-max-nodes-per-host = 4

filter-on = *:transfer

filter-on = eosio:newaccount
filter-on = eosio:delegatebw
filter-on = eosio:undelegatebw
filter-on = eosio:bidname
filter-on = eosio:buyram
filter-on = eosio:sellram
filter-on = eosio:buyrambytes
filter-on = eosio:regproducer
[...]

PR #8664, opened 2020-02-20
configurable priority for handling http requests

taokayan:
## Change Description
This change provides flexibility for users to configure priority for handling RPC requests, possible solution to #8655 . The default priority is low. 

for example, users who are running a local nodeos can configure ```--http-request-priority=100``` for faster RPC response.

Changing to higher priority will result in shorter time of all RPC requests (including get_info, get_abi, ...) but may results in delay of syncing.

## Consensus Changes
No.

## API Changes
No.

## Documentation Additions
new nodeos parameter ```--http-request-priority```, higher value results in higher priority.


heifner:
See #8142

heifner:
Closing. Went with #8142 & #8729.

PR #8670, opened 2020-02-20
Using get raw abi in cleos

allenhan2:
## Change Description

Replace get_abi with get_raw_abi to fetch abi file, then convert the raw abi to abi before sending back to the requestor.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -445,13 +445,12 @@ auto abi_serializer_resolver = [](const name& account) -> fc::optional<abi_seria
    static unordered_map<account_name, fc::optional<abi_serializer> > abi_cache;
    auto it = abi_cache.find( account );
    if ( it == abi_cache.end() ) {
-      auto result = call(get_abi_func, fc::mutable_variant_object("account_name", account));
-      auto abi_results = result.as<eosio::chain_apis::read_only::get_abi_results>();
-
-      fc::optional<abi_serializer> abis;
-      if( abi_results.abi.valid() ) {
-         abis.emplace( *abi_results.abi, abi_serializer_max_time );
-      } else {
+       fc::optional<abi_serializer> abis;
+       auto raw_result = call(get_raw_abi_func, fc::mutable_variant_object("account_name", account));
+       if (!raw_result["abi"].is_null()) {
+           auto abi_result = fc::raw::unpack<abi_def>(raw_result["abi"].as_blob().data);
+           abis.emplace(abi_result, abi_serializer_max_time);
+       } else {
          std::cerr << "ABI for contract " << account.to_string() << " not found. Action data will be shown in hex only." << std::endl;
       }
       abi_cache.emplace( account, abis );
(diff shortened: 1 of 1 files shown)


heifner:
Note: `get_raw_abi` was added in `v1.3.0` so this will break interacting with nodeos older than that version. Since all recent supported versions should work this is just an FYI.

programs/cleos/main.cpp
```diff
@@ -2673,15 +2673,22 @@ int main( int argc, char** argv ) {
...
-         std::cerr << localized("saving abi to ${filename}", ("filename", filename)) << std::endl;
-         std::ofstream abiout( filename.c_str() );
-         abiout << abi;
-      } else {
-         std::cout << abi << "\n";
+      auto raw_abi_result = call(get_raw_abi_func, fc::mutable_variant_object("account_name", accountName));
+      auto raw_abi_blob = raw_abi_result["abi"].as_blob().data;
+      if (raw_abi_blob.size() != 0) {
+          auto abi_result = fc::raw::unpack<abi_def>(raw_abi_result["abi"].as_blob().data);
+          std::cout << "789\n";
```
jeffreyssmith2nd: This looks like debugging code that should be removed.

programs/cleos/main.cpp
```diff
@@ -445,13 +445,13 @@ auto abi_serializer_resolver = [](const name& account) -> fc::optional<abi_seria
...
-      if( abi_results.abi.valid() ) {
-         abis.emplace( *abi_results.abi, abi_serializer_max_time );
-      } else {
+       fc::optional<abi_serializer> abis;
+       auto raw_abi_result = call(get_raw_abi_func, fc::mutable_variant_object("account_name", account));
+       auto raw_abi_blob = raw_abi_result["abi"].as_blob().data;
+       if (raw_abi_blob.size() != 0) {
+           auto abi_result = fc::raw::unpack<abi_def>(raw_abi_blob);
+           abis.emplace(abi_result, abi_serializer_max_time);
+       } else {
```
jeffreyssmith2nd: It's hard to tell on Github, but it looks like you changed the indentation here. In general, most of our code uses 3 spaces.

programs/cleos/main.cpp
```diff
@@ -2673,15 +2673,22 @@ int main( int argc, char** argv ) {
...
+              std::cerr << localized("saving abi to ${filename}", ("filename", filename)) << std::endl;
+              std::ofstream abiout(filename.c_str());
+              abiout << abi;
+          } else {
+              std::cout << abi << "\n";
+          }
+      } else
+      {
+        FC_THROW_EXCEPTION(key_not_found_exception, "Key ${key}", ("key", "abi"));
       }
```
heifner: Please follow style you see used in existing code. `} else {`

programs/cleos/main.cpp
heifner: Make sure you are not using tabs.

Issue #8614, opened 2020-02-07
cleos/keosd race condition


tbfleming:
develop branch, OSX

```
$ ~/work/eos-develop/build/bin/cleos wallet unlock --password `cat ~/password`
"/Users/todd.fleming/work/eos-develop/build/bin/keosd" launched

Error 3110002: Missing Wallet API Plugin
Ensure that you have eosio::wallet_api_plugin added to your node's configuration!
Otherwise specify your wallet location with --wallet-url argument!
Error Details:
Wallet is not available

$ ~/work/eos-develop/build/bin/cleos wallet unlock --password `cat ~/password`
Unlocked: default
```

Related: #6683

allenhan2:
fixed on PR, #8671

PR #8671, opened 2020-02-20
Fix for cleos and keosd race condition

allenhan2:
## Change Description

issue: #8614
When keosd is not running, cleos will start keosd,  then immediately try to connect to its socket, then fails because wallet_api has not started yet right after http port starts taking incoming traffic.

The fix is to let http_plugin startup yield to others' (including wallet api) startup finish.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


programs/cleos/main.cpp
```diff
@@ -995,14 +996,35 @@ void ensure_keosd_running(CLI::App* app) {
...
-                                     bp::std_out > bp::null,
+                                     bp::std_out > keos_out,
                                      bp::std_err > bp::null);
         if (keos.running()) {
             std::cerr << binPath << " launched" << std::endl;
-            keos.detach();
-            try_local_port(2000);
+
+            std::string keos_msg;
+            constexpr size_t TOTOAL_RETRIES = 10;
```
swatanabe-b1: TOTAL.

programs/cleos/main.cpp
```diff
@@ -995,14 +996,35 @@ void ensure_keosd_running(CLI::App* app) {
...
+            constexpr size_t KEOS_WAIT_TIME_MS = 200;
+            size_t tries = 0;
+            for(; tries<TOTOAL_RETRIES; ++tries)
+            {
+               std::this_thread::sleep_for(std::chrono::milliseconds(KEOS_WAIT_TIME_MS));
+               std::getline(keos_out, keos_msg);
```
swatanabe-b1: getline is blocking.  There's no need to wait.

allenhan2: It behaves as syncing, if keosd returns -1 before (daemonizing) sending "keosd::config::keosd_started_msg",  pstream will make getline return immediately.  Then, cleos knows that retry fails, print out err msg, then move on, no wait.

if keosd is already there before cleos runs, cleos won't wait. Otherwise, cleos starts keosd, and wait for keosd  finishing startup  (right before daemonizing).  For any reason, if keosd hangs before startup is done, cleos will hang there too.  

If cleos should run independently from keosd, then correct, will change to asnc read instead.

allenhan2: It behaves as syncing, if keosd returns -1 before (daemonizing) sending "keosd::config::keosd_started_msg", pstream will make getline return immediately. Then, cleos knows that retry fails, print out err msg, then move on, no wait.

if keosd is already there before cleos runs, cleos won't wait. Otherwise, cleos starts keosd, and wait for keosd finishing startup (right before daemonizing). For any reason, if keosd hangs  within startup, cleos will hang there too.

If cleos should run independently from keosd, then correct, will change to asnc read instead.

allenhan2:
The problem seems to me is that the cleos does not know when keosd finishes startup (before daemonizing).  not sure why using fork and pipe helps sync the state.  

current pstream is the wrapper around pipe between parent and child.  Using async read will make cleos and keosd independent.

swatanabe-b1:
> The problem seems to me is that the cleos does not know when keosd finishes startup (before daemonizing). not sure why using fork and pipe helps sync the state.
> 

Did you look at #5705?  I meant to use fork at startup in keosd.

> current pstream is the wrapper around pipe between parent and child. Using async read will make cleos and keosd independent.

You shouldn't abuse stdout for this synchronization.  keosd may write other things to stdout and  will break if it's a pipe whose other end is closed instead of /dev/null.

allenhan2:
thanks steve, only saw matt's issue description but did not see his PR. understand what you mean now.

allenhan2:
chatted with @spoonincode, and suggestion is focusing on why it is broken and not before, could be related to some subtle asio priority issue).  At the same time, chatted with @swatanabe-b1 , and could be the issue is always there. I will do more research on this before further action.

allenhan2:
From code, the issue is cleos thinks keosd is ready when http port is pingable. But, very often wallet api has not finished loading yet when port is pingable.
Since I can not pass the logging setting from cleos -> keos ->http_plugin->wallet api loading, I can not verify my theory.
However, it helps cleos to sync with keosd state by adding signal sync.  It is minimal change to make cleos and keosd working. 
The syncing issue is application level, need to refactor and test across all applications in the future.

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
...
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(WAIT_DURATION_MS / 100));
+         }
+         return (ready_signal_received == KEOSD_READY);
+      }
+      static state ready_signal_received;
+      static constexpr uint32_t WAIT_DURATION_MS = 8000;
+   };
+}  // cleos
+}  // eosio
```
heifner: This should not go in this file. This file is copied from an open source project.

programs/cleos/main.cpp
```diff
@@ -1002,7 +999,11 @@ void ensure_keosd_running(CLI::App* app) {
         if (keos.running()) {
             std::cerr << binPath << " launched" << std::endl;
             keos.detach();
-            try_local_port(2000);
+
+           if (!cleos::KeosdSignalHndl::wait_keosd_start())
+           {
+              std::cerr << "Failed to launch " << binPath << std::endl;
+           }
         } else {
```
heifner: Follow code style found in files you modify.

programs/cleos/main.cpp
```diff
@@ -77,6 +77,7 @@ Usage: ./cleos create account [OPTIONS] creator name OwnerKey ActiveKey
 #include <fc/exception/exception.hpp>
 #include <fc/variant_object.hpp>
 #include <fc/static_variant.hpp>
+#include <thread>
```
heifner: include not needed.

programs/keosd/main.cpp
```diff
@@ -43,12 +60,23 @@ int main(int argc, char** argv)
          .default_unix_socket_path = keosd::config::key_store_executable_name + ".sock",
          .default_http_port = 0
       });
+      const int parent_id = get_parent_pid(argc, argv);
+      const int argc_no_parent_id = (parent_id > 0) ? argc - 2 : argc;
+      char** const argv_no_parent_id = (parent_id > 0) ? &argv[2] : argv;
```
heifner: This seems very fragile. What if it is not the first argument?

programs/keosd/main.cpp
```diff
@@ -31,6 +32,22 @@ bfs::path determine_home_directory()
...
 
+int get_parent_pid(int argc, char** argv) {
+   size_t idx = 1;
+   for (; idx < argc; ++idx)
+   {
+      if (!strcmp(argv[idx], "--parent_pid"))
+      {
+         break;
+      }
+   }
```
heifner: Use consistent code style found used in files you modify.

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
...
+      enum state : size_t
+      {
+         KEOSD_NOT_READY = 0,
+         KEOSD_READY = 1
+      };
+
+      static void init() {
+         ready_signal_received = KEOSD_NOT_READY;
+         std::signal(SIGUSR1, handle_keosd_start_signal);
+      }
```
heifner: We use boost signal handling in appbase. Seems like this should also use boost.

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
 } // namespace detail
 
 } // namespace CLI
+
+namespace eosio {
+   namespace cleos {
+   struct KeosdSignalHndl {
```
heifner: Use snake case for all struct/classes. Follow code style found in existing code.

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
...
+            if (duration_cast<std::chrono::milliseconds>(system_clock::now().time_since_epoch()).count() - start_time >
+                WAIT_DURATION_MS) {
+               break;
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(WAIT_DURATION_MS / 100));
+         }
+         return (ready_signal_received == KEOSD_READY);
+      }
+      static state ready_signal_received;
+      static constexpr uint32_t WAIT_DURATION_MS = 8000;
```
heifner: All caps is used for macros not constants.

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
 } // namespace detail
 
 } // namespace CLI
+
+namespace eosio {
+   namespace cleos {
+   struct KeosdSignalHndl {
+      enum state : size_t
```
heifner: Use enum class

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
...
+      {
+         KEOSD_NOT_READY = 0,
+         KEOSD_READY = 1
```
heifner: Do not use all caps for constants

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
...
+         while (ready_signal_received != KEOSD_READY) {
+            if (duration_cast<std::chrono::milliseconds>(system_clock::now().time_since_epoch()).count() - start_time >
+                WAIT_DURATION_MS) {
+               break;
+            }
+            std::this_thread::sleep_for(std::chrono::milliseconds(WAIT_DURATION_MS / 100));
+         }
+         return (ready_signal_received == KEOSD_READY);
+      }
+      static state ready_signal_received;
```
heifner: init variable

spoonincode:
The current approach, as cheesy at it is, has worked correctly in the past. From a quick code review, I am a little suspicious that the http thread pool may have introduced this problem. The thread_pool starts event handling immediately, so when the HTTP server starts listening in http_pligin::plugin_startup() it is able to service incoming requests immediately. However, it's not until wallet_api_plugin::plugin_startup() that the wallet endpoints are registered. I'm pretty sure http_plugin startup will always run before wallet_api_plugin::plugin_startup(), so it's ready to serve requests before wallet endpoints are registered.

Previously back when there was no http thread pool, the event loop http plugin was using wouldn't have been pumped until all the plugin startup()s had been called.

I wonder if we simply moved the `add_api()` calls in wallet_api's startup to wallet_api's initialize if everything would be fixed? It's a little icky because I think it means the wallet plugin may be sent requests before wallet's startup()....... but the wallet plugin doesn't have a startup() so I don't think we'd run afoul of anything at the moment?

Though, fwiw, I'm much in favor of a better startup solution. But the approach in this PR and the previous daemonization approach I did don't support win32. That's something that in my opinion we need to be mindful about.

spoonincode:
Well, if the above theory is really the problem, probably having http_plugin `post()` its startup routine to the main thread event loop is even better.

allenhan2:
@spoonincode  made a quick try, your theory works. Let me see how to make the http_plugin post()

programs/cleos/CLI11.hpp
```diff
@@ -2434,3 +2434,40 @@ struct AppFriend {
...
+         KEOSD_READY = 1
+      };
+
+      static void init() {
+         ready_signal_received = KEOSD_NOT_READY;
+         std::signal(SIGUSR1, handle_keosd_start_signal);
+      }
+
+      static void handle_keosd_start_signal(int) {
+         ready_signal_received = KEOSD_READY;
```
swatanabe-b1: This causes a data race.  See [intro.races]
"Two actions are potentially concurrent if... they are unsequenced, at least one is performed by a signal handler, and they are not both performed by the same signal handler invocation."

"The execution of a program contains a data race if it contains two potentially concurrent conflicting actions, at least one of which is not atomic, and neither happens before the other..."

programs/keosd/main.cpp
```diff
@@ -43,12 +60,23 @@ int main(int argc, char** argv)
...
-      if(!app().initialize<wallet_plugin, wallet_api_plugin, http_plugin>(argc, argv))
+      if(!app().initialize<wallet_plugin, wallet_api_plugin, http_plugin>(argc_no_parent_id, argv_no_parent_id))
          return -1;
       auto& http = app().get_plugin<http_plugin>();
       http.add_handler("/v1/" + keosd::config::key_store_executable_name + "/stop", [&a=app()](string, string, url_response_callback cb) { cb(200, fc::variant(fc::variant_object())); a.quit(); } );
       app().startup();
+
+      if (parent_id > 0)
+      {
+         const int ret_sig_que = kill(parent_id, SIGUSR1);
```
swatanabe-b1: What happens if the parent has already exited?
What happens if keosd initialization fails (e.g. someone else launched keosd concurrently).

allenhan2:
Added post to delay the threadpool and related port initialization in  http_plugin's startup. If any exception during delayed initialization, shutdown the app. Tested on my mac desktop, seems working well. Please let me know any other tests or test cases. 
thanks, @spoonincode @swatanabe-b1 @heifner

plugins/http_plugin/http_plugin.cpp
```diff
@@ -590,86 +590,91 @@ namespace eosio {
    void http_plugin::plugin_startup() {
 
       handle_sighup(); // setup logging
-
-      my->thread_pool.emplace( "http", my->thread_pool_size );
-
-      if(my->listen_endpoint) {
+      app().post(appbase::priority::high, [this] ()
+      {
+         my->thread_pool.emplace( "http", my->thread_pool_size );
```
heifner: This should be moved into the try-catch.

allenhan2: right, fixed

allenhan2: fixed

programs/cleos/CLI11.hpp
allenhan2: Cool, thanks, it is good reading.

programs/cleos/main.cpp
allenhan2: The new approach is making http_plug startup yield to wallet-api startup. So, no change here.

allenhan2:
@swatanabe-b1 could you please help to give a review of this new approach.  @heifner  asks your review on the approach (even it was for the previous one),  
BTW, dismissed your comments on the old approach though. 
thanks!

Issue #8205, opened 2019-11-08
[Feature request]: Cleos option to automatically unpack data when forming transaction


EDmitry:
Trezor hardware wallet requires all data to be unpacked in order to sign a transaction (https://github.com/trezor/trezor-firmware/issues/173). It would be great if there was an option to just make Cleos unpack data automatically instead of having to call "convert unpack_action_data" for all the data blobs in the transaction.

PR #8676, opened 2020-02-21
add cleos option to unpack data when forming transaction, useful for …

victorj8:
…example when signing the transaction with trezor wallets

## Change Description

Add an option for cleos to unpack data when forming transaction, useful for example when signing the transaction with trezor wallets #8205 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


programs/cleos/main.cpp
```diff
@@ -399,7 +423,13 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
       }
    } else {
       if (!tx_return_packed) {
-        return fc::variant(trx);
+         if (tx_unpack_data) {
+            fc::variant unpacked_data_trx;
+            abi_serializer::to_variant(trx, unpacked_data_trx, abi_serializer_resolver, abi_serializer_max_time);
```
heifner: Should add a `try-catch` around this and revert to `fc::variant(trx);`. Then I think you can just make that the behavior and remove the new `-u` option.

victorj8: Done

tests/nodeos_run_test.py
```diff
@@ -599,6 +599,33 @@
...
+
+    try:
+        assert(not trans[1]["signatures"])
+    except (AssertionError, KeyError) as _:
+        Print("ERROR: Expected signatures array to be empty due to skipping option enabled.")
+        raise
+
+    try:
+        assert("hex_data" in trans[1]["actions"][0])
+    except (AssertionError, KeyError) as _:
```
heifner: Can you add checks for `data.from, data.to, data.memo`. We would like to move away from `hex_data` only being there when `data` contains expanded form, so would prefer this test to test for expanded `data` directly.

victorj8: Done

heifner:
Need to resolve conflict

PR #8697, opened 2020-02-24
Add `deprecate_deferred_transactions` protocol feature

johndebord:


libraries/chain/wasm_interface.cpp
```diff
@@ -1491,6 +1491,12 @@ class transaction_api : public context_aware_api {
       }
 
       void send_deferred( const uint128_t& sender_id, account_name payer, array_ptr<char> data, uint32_t data_len, uint32_t replace_existing) {
+         bool stop_deferred_transactions_activated = context.control.is_builtin_activated(builtin_protocol_feature_t::stop_deferred_transactions);
+         
+         if ( stop_deferred_transactions_activated && replace_existing ) {
```
swatanabe-b1: It's a bit odd that the handling for stop_deferred_transactions is split between here and schedule_deferred_transaction.

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -194,6 +194,18 @@ may use a new `set_proposed_producers_ex` intrinsic to access extended features.
 Builtin protocol feature: ACTION_RETURN_VALUE
 
 Enables new `set_action_return_value` intrinsic which sets a value that is included in action_receipt.
+*/
+            {}
+         } )
+
+         (  builtin_protocol_feature_t::stop_deferred_transactions, builtin_protocol_feature_spec{
+            "DEPRECATE_DEFERRED_TRANSACTIONS",
+            fc::variant("8b3d91df64471bdcf4b0f14edaf4a1cf7e082e36902a1e26516c81f6c7f658d7").as<digest_type>(),
```
swatanabe-b1: The text to be hashed should start at "Builtin protocol feature:".  You only included the description.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -62,7 +62,7 @@ namespace eosio { namespace testing {
       old_bios_only,
       preactivate_feature_only,
       preactivate_feature_and_new_bios,
-      full
+      complete
```
swatanabe-b1: Any particular reason for this renaming?

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -152,14 +152,14 @@ namespace eosio { namespace testing {
...
-         void              init(const setup_policy policy = setup_policy::full, db_read_mode read_mode = db_read_mode::SPECULATIVE);
+         void              init(const setup_policy policy = setup_policy::complete, db_read_mode read_mode = db_read_mode::SPECULATIVE);
          void              init(controller::config config, const snapshot_reader_ptr& snapshot);
          void              init(controller::config config, const genesis_state& genesis);
          void              init(controller::config config);
          void              init(controller::config config, protocol_feature_set&& pfs, const snapshot_reader_ptr& snapshot);
          void              init(controller::config config, protocol_feature_set&& pfs, const genesis_state& genesis);
          void              init(controller::config config, protocol_feature_set&& pfs);
-         void              execute_setup_policy(const setup_policy policy);
+         void              execute_setup_policy(const setup_policy policy, const std::vector<digest_type>& ignored_features = {});
```
swatanabe-b1: I don't like the fact that this allows you to specify policy and ignored_features independently, but ignored_features is only meaningful for setup_policy::full.

libraries/testing/tester.cpp
```diff
@@ -1127,13 +1127,18 @@ namespace eosio { namespace testing {
...
       ( const digest_type& feature_digest ) {
          const auto& pf = pfs.get_protocol_feature( feature_digest );
          FC_ASSERT( pf.builtin_feature, "called add_digests on a non-builtin protocol feature" );
          if( !pf.enabled || pf.earliest_allowed_activation_time > current_block_time
              || pfm.is_builtin_activated( *pf.builtin_feature, current_block_num ) ) return;
 
+         const auto feature_is_ignored = std::find(ignored_features.cbegin(), ignored_features.cend(), feature_digest);
+         if( feature_is_ignored != ignored_features.cend() ) {
+            return;
+         }
```
swatanabe-b1: This won't work correctly if you ignore a feature that has other features that depend on it.

task_spec
```diff
@@ -0,0 +1,90 @@
+1st Protocol Feature `stop_deferred_transactions`
```
swatanabe-b1: This file does not belong here.  It's also inconsistent with the PR description.

libraries/chain/apply_context.cpp
```diff
@@ -512,29 +519,33 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
 
       // Use remove and create rather than modify because mutating the trx_id field in a modifier is unsafe.
       db.remove( *ptr );
```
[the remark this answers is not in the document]
johndebord: My assumption, based off of the specification, is that there are two ways to enter deferred transaction logic:

1. Through `transaction_api::send_deferred`.
2. Through a user extending his/her code via a plugin and calling `apply_context::schedule_deferred_transaction`.

Therefore two individual (seemingly redundant) checks must be made. Am I wrong in this assumption?

libraries/chain/apply_context.cpp
```diff
@@ -512,29 +519,33 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
-         gtx.sender      = receiver;
-         gtx.sender_id   = sender_id;
-         gtx.payer       = payer;
-         gtx.published   = control.pending_block_time();
-         gtx.delay_until = gtx.published + delay;
-         gtx.expiration  = gtx.delay_until + fc::seconds(control.get_global_properties().configuration.deferred_trx_expiration_window);
-
-         trx_size = gtx.set( trx );
-      } );
+      if ( !stop_deferred_transactions_activated ) {
```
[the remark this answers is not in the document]
johndebord: I need a walk-through of this function to see how this is so.

libraries/chain/wasm_interface.cpp
johndebord: This is consistent with protocol feature `builtin_protocol_feature_t::disallow_empty_producer_schedule`:
https://github.com/EOSIO/eos/blob/9cb5a2101a08d8c2d3e474409066bdfd42221a86/libraries/chain/wasm_interface.cpp#L205
https://github.com/EOSIO/eos/blob/9cb5a2101a08d8c2d3e474409066bdfd42221a86/libraries/chain/controller.cpp#L2947

libraries/testing/include/eosio/testing/tester.hpp
johndebord: I felt that the name `complete` doesn't imply all protocol features be _fully_ utilized. Do you have another suggestion?

libraries/chain/protocol_feature_manager.cpp
johndebord: Good catch. Hash has been recalculated.

libraries/testing/include/eosio/testing/tester.hpp
johndebord: I'm trying a different approach in my latest commit.

libraries/testing/tester.cpp
johndebord: I'm trying a different approach in my latest commit.

task_spec
johndebord: File has been removed.

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
swatanabe-b1: I don't think apply_context::schedule_deferred_transaction should ever be called outside of wasm.  Also, the main point that I was trying to make was that the behavior is not the same in the two cases.

libraries/chain/wasm_interface.cpp
swatanabe-b1: It's not really the same.  The way you've implemented it requires checks in both apply_context and wasm_interface to be correct.  My problem is that apply_context::schedule_deferred_transaction doesn't seem to have a clear understanding of its own preconditions.

libraries/testing/include/eosio/testing/tester.hpp
swatanabe-b1: I don't understand what you mean.

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
swatanabe-b1: Look at the call to add_ram_usage at the bottom of the function.

libraries/testing/tester.cpp
```diff
@@ -1116,24 +1116,47 @@ namespace eosio { namespace testing {
...
              || pfm.is_builtin_activated( *pf.builtin_feature, current_block_num ) ) return;
 
+         std::function<void(const digest_type&)> check_dependencies =
+         [&check_dependencies, &ignored_digests, &pf] ( const digest_type& feature ) {
+            const auto dependency_is_ignored     = std::find( ignored_digests.cbegin(), ignored_digests.cend(), feature );
+            const auto dependency_has_dependency = std::find( pf.dependencies.cbegin(),  pf.dependencies.cend(),  feature );
+            if ( dependency_is_ignored != ignored_digests.cend() && dependency_has_dependency != pf.dependencies.cend() ) {
+                check_dependencies(*dependency_has_dependency);
+            } else {
+                FC_ASSERT( pf.builtin_feature, "ignoring a protocol feature that is a dependency for another protocol feature" );
```
swatanabe-b1, line 1150: Won't this assertion alway pass?  Looks to me like a bad copy/paste.

libraries/testing/tester.cpp
```diff
@@ -1116,24 +1116,47 @@ namespace eosio { namespace testing {
...
+            }
+         };
+
+         const auto digest_is_ignored = std::find(ignored_digests.cbegin(), ignored_digests.cend(), feature_digest);
```
swatanabe-b1, line 1154: I think it would be better to put this after the check for whether we've already seen the digest.

unittests/api_tests.cpp
```diff
@@ -404,55 +404,57 @@ BOOST_FIXTURE_TEST_CASE(action_receipt_tests, TESTER) { try {
...
-	create_account( N(testapi) );
-	create_account( N(acc1) );
-	create_account( N(acc2) );
-	create_account( N(acc3) );
-	create_account( N(acc4) );
-	produce_blocks(10);
-	set_code( N(testapi), contracts::test_api_wasm() );
-	produce_blocks(1);
+BOOST_AUTO_TEST_CASE(action_receipt) { try {
+   validating_tester chain( {}, {::eosio::chain::builtin_protocol_feature_t::stop_deferred_transactions} );
```
swatanabe-b1, line 408: The original code uses TESTER.  Can you make tester define the same constructors as validating_tester?

unittests/api_tests.cpp
```diff
@@ -581,14 +583,12 @@ BOOST_FIXTURE_TEST_CASE(require_notice_tests, TESTER) { try {
...
-   fc::temp_directory tempdir;
-   validating_tester chain( tempdir, true );
-   chain.execute_setup_policy( setup_policy::preactivate_feature_and_new_bios );
+   validating_tester chain( {}, {}, ::eosio::testing::setup_policy::preactivate_feature_and_new_bios );
    const auto& pfm = chain.control->get_protocol_feature_manager();
    const auto& d = pfm.get_builtin_digest(builtin_protocol_feature_t::action_return_value); // testapi requires this
    BOOST_REQUIRE(d);
    chain.preactivate_protocol_features( {*d} );
-
+
```
swatanabe-b1, line 591: You might clean up extra whitespace.

unittests/currency_tests.cpp
```diff
@@ -25,9 +25,10 @@ using namespace eosio::chain;
 using namespace eosio::testing;
 using namespace fc;
 
+static fc::temp_directory temporary_config_dir{};
```
swatanabe-b1, line 28: Seems to be unused.

libraries/testing/include/eosio/testing/tester.hpp
swatanabe-b1: To elaborate a bit, it seems to me that if full implies the features are _fully_ utilized, then complete equally implies that they are _completely_ utilized.

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
johndebord: I believe they now have the same behavior.

libraries/chain/controller.cpp
```diff
@@ -1244,7 +1244,10 @@ struct controller_impl {
       trx->accepted = true;
 
       transaction_trace_ptr trace;
-      if( gtrx.expiration < self.pending_block_time() ) {
+
+      bool stop_deferred_transactions_activated = self.is_builtin_activated(builtin_protocol_feature_t::stop_deferred_transactions);
+
+      if( gtrx.expiration < self.pending_block_time() || stop_deferred_transactions_activated ) {
```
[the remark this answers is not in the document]
johndebord: The behavior now matches the spec; if the transaction has expired then the transaction shall expire. Then we check to see if `builtin_protocol_feature_t::stop_deferred_transactions` has been activated; if so, then the trace shall be returned, and no further operations are done on the transaction.

libraries/chain/wasm_interface.cpp
johndebord: I think I need more clarification on this if it's still a problem.

libraries/testing/include/eosio/testing/tester.hpp
johndebord: Renamed back to the original name of `complete`.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -379,7 +378,7 @@ namespace eosio { namespace testing {
 
          void schedule_protocol_features_wo_preactivation(const vector<digest_type> feature_digests);
          void preactivate_protocol_features(const vector<digest_type> feature_digests);
-         void preactivate_all_builtin_protocol_features();
+         void preactivate_builtin_protocol_features(const std::vector<builtin_protocol_feature_t>& ignored_features);
```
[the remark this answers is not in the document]
johndebord: Renamed to `preactivate_selected_protocol_features`.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -516,49 +516,30 @@ namespace eosio { namespace testing {
...
-      controller::config vcfg;
-
-      validating_tester(const flat_set<account_name>& trusted_producers = flat_set<account_name>()) {
-         auto def_conf = default_config(tempdir);
 
+      validating_tester(const ::boost::container::flat_set<::eosio::chain::account_name>& trusted_producers = ::boost::container::flat_set<::eosio::chain::account_name>(),
+                        const ::std::vector<::eosio::chain::builtin_protocol_feature_t>& ignored_features = {},
+                        const ::eosio::testing::setup_policy& policy = ::eosio::testing::setup_policy::complete) {
+         ::std::pair<::eosio::chain::controller::config, ::eosio::chain::genesis_state> def_conf = default_config(tempdir); 
+         ::eosio::testing::base_tester::ignored_features = ignored_features;
```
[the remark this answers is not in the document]
johndebord: Style has been changed to the existing style.

libraries/testing/tester.cpp
```diff
@@ -1116,24 +1116,47 @@ namespace eosio { namespace testing {
...
          FC_ASSERT( pf.builtin_feature, "called add_digests on a non-builtin protocol feature" );
          if( !pf.enabled || pf.earliest_allowed_activation_time > current_block_time
              || pfm.is_builtin_activated( *pf.builtin_feature, current_block_num ) ) return;
 
+         std::function<void(const digest_type&)> check_dependencies =
+         [&check_dependencies, &ignored_digests, &pf] ( const digest_type& feature ) {
+            const auto dependency_is_ignored     = std::find( ignored_digests.cbegin(), ignored_digests.cend(), feature );
+            const auto dependency_has_dependency = std::find( pf.dependencies.cbegin(),  pf.dependencies.cend(),  feature );
+            if ( dependency_is_ignored != ignored_digests.cend() && dependency_has_dependency != pf.dependencies.cend() ) {
+                check_dependencies(*dependency_has_dependency);
```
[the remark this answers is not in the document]
johndebord: Should now be checking protocol feature dependencies correctly.

libraries/testing/tester.cpp
johndebord: Should now be checking protocol feature dependencies correctly.

libraries/testing/tester.cpp
johndebord: Moved that logic after checking whether or not the digest exists.

unittests/api_tests.cpp
johndebord: Made matching constructors for class `tester`.

unittests/api_tests.cpp
johndebord: Cleaned up extra whitespace.

unittests/currency_tests.cpp
johndebord: Removed `static fc::temp_directory temporary_config_dir{};`.

libraries/chain/wasm_interface.cpp
swatanabe-b1: In retrospect, my complaint here was really just a restatement of the other issues I raised regarding send_deferred and schedule_deferred_transaction from a different perspective.

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
swatanabe-b1: The behavior is still different in the case of `replace_existing = true`.  `transaction_api::send_deferred` short-circuits this case to a no-op, but `apply_context::schedule_deferred_transaction` handles it in a more complex way.

libraries/chain/apply_context.cpp
```diff
@@ -522,8 +528,22 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
          gtx.expiration  = gtx.delay_until + fc::seconds(control.get_global_properties().configuration.deferred_trx_expiration_window);
 
          trx_size = gtx.set( trx );
+
```
swatanabe-b1, line 531: So this now allows deferred transactions to be replaced.  I'm not so sure that that's a good idea, since it allows the expiration to be arbitrarily extended as long as someone keeps replacing it.

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
johndebord: Yes, that is what I was told to implement.

> [XX] `send_deferred`
> [XX]   - If `replace_existing` == true in `send_deferred`; it shall be a noop

> [XX] `schedule_deferred_transaction`
> [XX]   - If there is no transaction to replace: fail
> [XX]   - If there is a transaction to replace: continue to cancel the old one but do not schedule the new one

Hence, the assertion in `apply_context::schedule_deferred_transaction`.

libraries/chain/apply_context.cpp
```diff
@@ -522,8 +528,22 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
+         if ( stop_deferred_transactions_activated ) {
+            EOS_ASSERT( ram_restrictions_activated
+               || control.is_ram_billing_in_notify_allowed()
+               || (receiver == act->account) || (receiver == payer) || privileged,
+               subjective_block_production_exception,
+               "Cannot charge RAM to other accounts during notify."
+            );
+            add_ram_usage( payer, (config::billable_size_v<generated_transaction_object> + trx_size) );
+            return;
```
swatanabe-b1, line 540: Returns from the lambda.

libraries/chain/apply_context.cpp
```diff
@@ -522,8 +528,22 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
          trx_size = gtx.set( trx );
+
+         if ( stop_deferred_transactions_activated ) {
+            EOS_ASSERT( ram_restrictions_activated
+               || control.is_ram_billing_in_notify_allowed()
+               || (receiver == act->account) || (receiver == payer) || privileged,
+               subjective_block_production_exception,
+               "Cannot charge RAM to other accounts during notify."
+            );
+            add_ram_usage( payer, (config::billable_size_v<generated_transaction_object> + trx_size) );
```
swatanabe-b1, line 539: Is there a need to duplicate this?

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
swatanabe-b1: Can you explain the motivation for this difference?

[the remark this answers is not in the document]
johndebord: The motivation is unbeknownst to me. I just took it at face-value.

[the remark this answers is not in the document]
swatanabe-b1: I'm somewhat concerned that this may be a misunderstanding.  The spec for `schedule_deferred_transaction` sounds to me like reasonable behavior for both of them.  @arhag

libraries/testing/tester.cpp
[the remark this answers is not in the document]
swatanabe-b1: Only handles the simple case of A depends on B, ignore B.  Does not handle the possibility of chaining ignores: A -> B -> C -> D, ignore D, which needs to cause all of A, B, C, and D to be ignored.

Also I'm on the fence about whether it's better to ignore additional features or whether it's better to issue an error.  On the one hand, it's bad if adding a new protocol feature breaks existing tests.  On the other hand, silently disabling more protocol features than were given may be surprising.  What do you think?

libraries/testing/include/eosio/testing/tester.hpp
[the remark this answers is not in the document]
swatanabe-b1: I don't see how the new name expresses the meaning of the function better than the old name.  I think part of the problem is that the function's behavior is odd, which makes it hard to come up with a good name.

libraries/testing/tester.cpp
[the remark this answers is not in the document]
johndebord: I definitely considered that case. But for the time being we only have the case of A depends on B.
Would it be best practice to account for future cases, such as chaining ignores? Or is the best practice only to account for what is currently in use?

And I definitely agree with you there on the last point. If it were up to me, I think `tester` is due for a re-write; to consider, from the ground up, an architecture where an arbitrary number of protocol features that may or may not depend on each other.

libraries/chain/apply_context.cpp
johndebord: Removed that logic out since lines https://github.com/EOSIO/eos/blob/deprecate-deferred-protocol-feature/libraries/chain/apply_context.cpp#L544-L546 should take care of billing correctly.

libraries/chain/apply_context.cpp
johndebord: Removed duplication.

libraries/testing/tester.cpp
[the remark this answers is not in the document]
swatanabe-b1: Best practice is to handle the public API of protocol features.  Since arbitrary protocol feature dependencies are allowed, you should support them.  Tester should not need any specific knowledge of what protocol features are actually defined.  The existing code in tester to determine the activation order of protocol features is designed to handle arbitrary dependencies correctly, so I think it's reasonable for you to do so as well.

Now, with all of that having been said, being fully general is not always worth the effort.  If you do decide to take shortcuts, you should definitely call it out in a comment, to help anyone who adds a new protocol feature and gets confused by mysterious failures.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
swatanabe-b1: Does producer_plugin know how to handle this outcome?

PR #8702, opened 2020-02-26
New options for api nodes - 2.0

### LINKED ISSUE #8685: Introduce new read-mode to nodeos: api-read-only
Enhancement request, as discussed in Mainnet BPs telegram:

Kevin Heifner, [22.02.20 22:11]
Would an option for an api-read-only be useful. read-only in that it would not speculatively execute p2p transactions but would allow for push_transaction? It would only speculatively execute transactions sent to it via api call.

Eric - sw/eden, [22.02.20 22:21]
[In reply to Kevin Heifner]
Ooh, good suggestion.
This would be useful for sure!

Aaron Cox (jesta) — Greymass, [23.02.20 02:54]
[In reply to Kevin Heifner]
Yeah, It'd second that this could be very useful.

heifner:
## Change Description

*  --p2p-accept-transactions arg (=1)    Allow transactions received over p2p 
                                        network to be evaluated and relayed if valid.
*  --api-accept-transactions arg (=1)    Allow API transactions to be evaluated 
                                        and relayed if valid.

- `read-mode = irreversible` now asserts unless:
    - `p2p-accept-transactions = false` 
    - `api-accept-transactions = false`

- Resolves #8685 by user specifying:
  - `read-mode = head`
  - `p2p-accept-transactions = false`
  - `api-accept-transactions = true`

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### Updated Descriptions: 

  **--read-mode arg (=speculative)**        
- Database read mode ("speculative", "head", "read-only", "irreversible").
In "speculative" mode: database 
contains state changes by transactions 
in the blockchain up to the head block 
as well as some transactions not yet 
included in the blockchain.

- In "head" mode: database contains state
changes by only transactions in the 
blockchain up to the head block; 
transactions received by the node are 
relayed if valid.

- In "read-only" mode: (DEPRECATED: see 
p2p-accept-transactions & 
api-accept-transactions) database 
contains state changes by only 
transactions in the blockchain up to 
the head block; transactions received 
via the P2P network are not relayed and
transactions cannot be pushed via the 
chain API.

- In "irreversible" mode: database 
contains state changes by only 
transactions in the blockchain up to 
the last irreversible block; 
transactions received via the P2P 
network are not relayed and 
transactions cannot be pushed via the chain API.
   
### New Options:
                                     
  --p2p-accept-transactions arg (=1)    Allow transactions received over p2p 
                                        network to be evaluated and relayed if valid.
  --api-accept-transactions arg (=1)    Allow API transactions to be evaluated 
                                        and relayed if valid.

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -45,8 +45,8 @@ std::ostream& operator<<(std::ostream& osm, eosio::chain::db_read_mode m) {
       osm << "head";
    } else if ( m == eosio::chain::db_read_mode::READ_ONLY ) {
       osm << "read-only";
-   } else if ( m == eosio::chain::db_read_mode::API_READ_ONLY ) {
-      osm << "api-read-only";
+   } else if ( m == eosio::chain::db_read_mode::API ) {
+      osm << "api";
    } else if ( m == eosio::chain::db_read_mode::IRREVERSIBLE ) {
       osm << "irreversible";
    }
@@ -74,8 +74,8 @@ void validate(boost::any& v,
      v = boost::any(eosio::chain::db_read_mode::HEAD);
   } else if ( s == "read-only" ) {
      v = boost::any(eosio::chain::db_read_mode::READ_ONLY);
-  } else if ( s == "api-read-only" ) {
-     v = boost::any(eosio::chain::db_read_mode::API_READ_ONLY);
+  } else if ( s == "api" ) {
+     v = boost::any(eosio::chain::db_read_mode::API);
   } else if ( s == "irreversible" ) {
      v = boost::any(eosio::chain::db_read_mode::IRREVERSIBLE);
   } else {
@@ -242,7 +242,7 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "In \"speculative\" mode database contains changes done up to the head block plus changes made by transactions not yet included to the blockchain.\n"
           "In \"head\" mode database contains changes done up to the current head block.\n"
           "In \"read-only\" mode database contains changes done up to the current head block and transactions cannot be pushed to the chain API.\n"
-          "In \"api-read-only\" mode database contains changes done up to the head block; only api transactions are speculatively executed.\n"
+          "In \"api\" mode database contains changes done up to the head block; only api transactions are speculatively executed.\n"
           "In \"irreversible\" mode database contains changes done up to the last irreversible block and transactions cannot be pushed to the chain API.\n"
           )
          ("validation-mode", boost::program_options::value<eosio::chain::validation_mode>()->default_value(eosio::chain::validation_mode::FULL),
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -242,7 +242,7 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "In \"speculative\" mode database contains changes done up to the head block plus changes made by transactions not yet included to the blockchain.\n"
           "In \"head\" mode database contains changes done up to the current head block.\n"
           "In \"read-only\" mode database contains changes done up to the current head block and transactions cannot be pushed to the chain API.\n"
-          "In \"api-read-only\" mode database contains changes done up to the head block plus changes made by api transactions but not p2p transactions. \n"
+          "In \"api-read-only\" mode database contains changes done up to the head block; only api transactions are speculatively executed.\n"
           "In \"irreversible\" mode database contains changes done up to the last irreversible block and transactions cannot be pushed to the chain API.\n"
           )
          ("validation-mode", boost::program_options::value<eosio::chain::validation_mode>()->default_value(eosio::chain::validation_mode::FULL),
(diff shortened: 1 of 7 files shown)


swatanabe-b1:
This could use a better name.  I don't see how the api's are read-only in this mode.

heifner:
@swatanabe-b1 point well taken, we decided on `api`

heifner:
We have discussed this some more and have decided that `api` and the existing `read-only` are not really database `read-mode`s. Plan on adding `p2p-accept-transactions` and  `api-accept-transactions` both defaulted to `true`.  The existing `read-only` will be deprecated in favor of specifying: `read-mode = head`, `p2p-accept-transactions = false`, and `api-accept-transactions = false`. For the `v2.0.x` & `v1.8.x` versions of this PR `read-mode` will indicate a warning message of deprecation and will automatically set `p2p-accept-transactions = false`, `api-accept-transactions = false`, and `read-mode = head`. The `develop` version of this PR will remove `read-only` mode.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -235,11 +238,13 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
...
+          "In \"speculative\" mode: database contains state changes by transactions in the blockchain up to the head block as well as some transactions not yet included in the blockchain.\n"
+          "In \"head\" mode: database contains state changes by only transactions in the blockchain up to the head block; transactions received by the node are relayed if valid.\n"
+          "In \"read-only\" mode: (DEPRECATED: see p2p-accept-transactions & api-accept-transactions) database contains state changes by only transactions in the blockchain up to the head block; transactions received via the P2P network are not relayed and transactions cannot be pushed via the chain API.\n"
+          "In \"irreversible\" mode: database contains state changes by only transactions in the blockchain up to the last irreversible block; transactions received via the P2P network are not relayed and transactions cannot be pushed via the chain API.\n"
           )
+         ( "p2p-accept-transactions", bpo::value<bool>()->default_value(true), "Allow transactions received over p2p network to be evaluated and relayed if valid.")
```
arhag: Even though this complicates things, I really think `p2p-accept-transactions` belongs in the net_plugin and not here.

I think what you could do is add another boolean (defaults to false) to the chain_plugin that tracks whether there is any source of transactions (this, or rather its negation, could be accessed via the `no_transactions` accessor). There would then need to be a function that other plugins could call which causes chain_plugin to consider possible other sources of transactions (and update that boolean accordingly). So if `api_accept_transactions` is true, `no_transactions` would always return false. But if `api_accept_transactions` is false, then `no_transactions` would return true by the end of the initialization stage of chain_plugin, but then later switch to false as other plugins (like the net_plugin) finish initialization.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -991,6 +996,25 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
...
          my->chain_config->read_mode = options.at("read-mode").as<db_read_mode>();
       }
+      my->p2p_accept_transactions = options.at( "p2p-accept-transactions" ).as<bool>();
+      my->api_accept_transactions = options.at( "api-accept-transactions" ).as<bool>();
+
+      if( my->chain_config->read_mode == db_read_mode::IRREVERSIBLE || my->chain_config->read_mode == db_read_mode::READ_ONLY ) {
+         if( my->chain_config->read_mode == db_read_mode::READ_ONLY ) {
+            wlog( "read-mode = read-only is deprecated use p2p-accept-transactions = false, api-accept-transactions = false instead." );
+         }
+         if( my->p2p_accept_transactions ) {
```
arhag: With the changes described above, this adjustment would have to be moved to the net_plugin.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -991,6 +996,25 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
...
+            my->p2p_accept_transactions = false;
+            std::stringstream ss; ss << my->chain_config->read_mode;
+            wlog( "p2p-accept-transactions set to false due to read-mode: ${m}", ("m", ss.str()) );
+         }
+         if( my->api_accept_transactions ) {
+            my->api_accept_transactions = false;
+            std::stringstream ss; ss << my->chain_config->read_mode;
+            wlog( "api-accept-transactions set to false due to read-mode: ${m}", ("m", ss.str()) );
+         }
+         EOS_ASSERT( no_transactions(), plugin_config_exception, "IRREVERSIBLE should configure no_transactions" );
```
arhag: This assertion could be moved to the beginning of `chain_plugin::plugin_startup` since by then all plugins had a chance to initialize and so the return value of `no_transactions()` should have finalized and therefore remain stable for the remainder of the process.

plugins/chain_plugin/chain_plugin.cpp
heifner: Done. Went with `accept_transactions` instead of `no_transactions`

plugins/chain_plugin/chain_plugin.cpp
heifner: Done.

plugins/chain_plugin/chain_plugin.cpp
heifner: Done.

Issue #8662, opened 2020-02-19
New rpc: /v1/chain/get_block_info


tbfleming:
This should behave like `get_block`, but return a smaller subset of data.

e.g. output:

```
{
  "block_num": 12345,
  "id": "00003039C04C45767B191E47A1C77F326AA8FE41AB26B7897036D725CA6D4DFD",
  "timestamp": "2020-02-18T17:46:37.000",
  "producer": "eosio",
  "confirmed": 0,
  "previous": "000030389B1290E805D57D94527C077A94ED0DAE33C91AA62F8FDD488C70A2FD",
  "transaction_mroot": "0000000000000000000000000000000000000000000000000000000000000000",
  "action_mroot": "2B5A5AB9FC4FD220D8928B673A889981144FA88C09AF2E2946737861D1905A91",
  "schedule_version": 0,
  "producer_signature": "SIG_K1_K2CneKYGj6YrYusp9ickseeRF5YGYLincE678zddLteGRd8rpqSWNZijcTws8MkjmXeePSWkA96cBGekWUML2CpgwssaUF",
  "ref_block_prefix": 1383860090
}
```

Also add to cleos.

matthewdarwin:
Nice.

I do hope that going forward when APIs change a new version number is assigned,

And ideally this is /v1/chain/get_block_header/123344 or similar. There is no need to POST for API calls that don't make changes.

Velua:
Any thoughts on implementing GraphQL?

tbfleming:
JSON handling in nodeos is problematic; we hope to eventually get rid of it. GraphQL inside nodeos is even worse. We have considered GraphQL inside history-tools, but that's a long way off if we ever go with it.

tbfleming:
New changes:
* Rename to `get_block_info`
* Add `ref_block_num`; this is a `uint16_t` containing the 16 LSBs of `block_num`

tbfleming:
More new changes:
* instead of `block_num_or_id`, accept a `uint32_t block_num`

PR #8704, opened 2020-02-26
add get_block_info

allenhan2:
## Change Description

issue #8662
Added get_block_info to chain plugin API and cleos

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

Tested with cul against nodeos and "cleos get block --info <block_num>" with block num only

output sample:
{
  "block_num": 10,
  "ref_block_num": 10,
  "id": "0000000ad55bc11c9037b3fe89540a6b872d04627613820fe4b661234d8b8570",
  "timestamp": "2020-02-27T01:42:12.000",
  "producer": "eosio",
  "confirmed": 0,
  "previous": "000000090acc78ad6813c89ddd592592ac449a4752953373632e6b9daaf5afe7",
  "transaction_mroot": "0000000000000000000000000000000000000000000000000000000000000000",
  "action_mroot": "8cdb8ca1d110f2a8d93605de68fae0f02ea20c226d9b8988d56c6775c81e9bac",
  "schedule_version": 0,
  "producer_signature": "SIG_K1_K15gK8oUdo7GhsP8B3D6NcBQKKjwabfCZRZKtgb83VZviy5KeKgExnvTAGhfz2ux5kdqY7CtLtrbpxppjiiNj7M4gfPNcL",
  "ref_block_prefix": 4273158032
}

### DIFF
--- a/libraries/chain/include/eosio/chain/exceptions.hpp
+++ b/libraries/chain/include/eosio/chain/exceptions.hpp
@@ -198,8 +198,6 @@ namespace eosio { namespace chain {
                                     3010015, "Key type is not a currently activated type" )
       FC_DECLARE_DERIVED_EXCEPTION( unactivated_signature_type,        chain_type_exception,
                                     3010016, "Signature type is not a currently activated type" )
-      FC_DECLARE_DERIVED_EXCEPTION( block_num_type_exception,           chain_type_exception,
-                                    3010017, "Invalid block num" )
 
 
    FC_DECLARE_DERIVED_EXCEPTION( fork_database_exception, chain_exception,
--- a/libraries/chain/include/eosio/chain/exceptions.hpp
+++ b/libraries/chain/include/eosio/chain/exceptions.hpp
@@ -198,6 +198,8 @@ namespace eosio { namespace chain {
                                     3010015, "Key type is not a currently activated type" )
       FC_DECLARE_DERIVED_EXCEPTION( unactivated_signature_type,        chain_type_exception,
                                     3010016, "Signature type is not a currently activated type" )
+      FC_DECLARE_DERIVED_EXCEPTION( block_num_type_exception,           chain_type_exception,
+                                    3010017, "Invalid block num" )
 
 
    FC_DECLARE_DERIVED_EXCEPTION( fork_database_exception, chain_exception,
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -2047,6 +2047,43 @@ fc::variant read_only::get_block(const read_only::get_block_params& params) cons
            ("ref_block_prefix", ref_block_prefix);
 }
 
+fc::variant read_only::get_block_header(const read_only::get_block_header_params& params) const {
+   signed_block_ptr block;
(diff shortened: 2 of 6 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2047,6 +2047,43 @@ fc::variant read_only::get_block(const read_only::get_block_params& params) cons
...
+         ("block_num", block->block_num())
+         ("id", block->id())
+         ("timestamp", block->timestamp)
+         ("producer", block->producer)
+         ("confirmed", block->confirmed)
+         ("previous", block->previous)
+         ("transaction_mroot", block->transaction_mroot)
+         ("action_mroot", block->action_mroot)
+         ("schedule_version", block->schedule_version)
+         ("producer_signature", block->producer_signature)
```
swatanabe-b1: Is there a reason for writing these out instead of upcasting the block to signed_block_header?  You're not including `header_extensions` or `new_producers`

allenhan2: according to the slack chat with todd, it is the expected output. Will double-check to clarify the requirements.

tbfleming: Don't want `header_extensions` since it could be large in the future. Don't need `new_producers` and it simplifies the history-tools implementation to not have that.

swatanabe-b1: I strongly disagree with this reasoning.  An api called get_block_header should return the whole block header.

swatanabe-b1: Also, I don't buy the argument that the header extensions "could be large in the future."  We shouldn't put anything into the block header extensions unless it really needs to be part of the header.

allenhan2:
Will leave it not merge now

allenhan2:
Please do not delete the branch.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -280,6 +280,12 @@ class read_only {
 
    fc::variant get_block(const get_block_params& params) const;
 
+   struct get_block_info_params {
+      string block_num;
+   };
```
heifner: This can just be `uint32_t`. No need for it to be a string.

libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -198,6 +198,8 @@ namespace eosio { namespace chain {
                                     3010015, "Key type is not a currently activated type" )
       FC_DECLARE_DERIVED_EXCEPTION( unactivated_signature_type,        chain_type_exception,
                                     3010016, "Signature type is not a currently activated type" )
+      FC_DECLARE_DERIVED_EXCEPTION( block_num_type_exception,           chain_type_exception,
+                                    3010017, "Invalid block num" )
```
heifner: I don't think we need this new exception now.

allenhan2:
@lparisc could you please check the necessary documentation updates.

PR #8717, opened 2020-02-27
Fix mutliple version protocol test intermittent failure

brianjohnson5972:
## Change Description
Fixed error in test due to 1.7 version producer producing block just as next block starts it production window and thus rejecting that block, just as production for all producers is paused.  Added logging to indicate producer is paused or resumed. Also other test cleanup and updated error message of forked_chain_test.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -963,7 +963,7 @@ void producer_plugin::handle_sighup() {
 }
 
 void producer_plugin::pause() {
-   ilog("Producer paused.")
+   ilog("Producer paused.");
    my->_pause_production = true;
 }
 
@@ -975,10 +975,10 @@ void producer_plugin::resume() {
    if (my->_pending_block_mode == pending_block_mode::speculating) {
       chain::controller& chain = my->chain_plug->chain();
       my->_unapplied_transactions.add_aborted( chain.abort_block() );
-      ilog("Producer resumed. Scheduling production.")
+      ilog("Producer resumed. Scheduling production.");
       my->schedule_production_loop();
    } else {
-      ilog("Producer resumed.")
+      ilog("Producer resumed.");
    }
 }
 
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -974,7 +975,10 @@ void producer_plugin::resume() {
    if (my->_pending_block_mode == pending_block_mode::speculating) {
       chain::controller& chain = my->chain_plug->chain();
       my->_unapplied_transactions.add_aborted( chain.abort_block() );
+      ilog("Producer resumed. Scheduling production.")
       my->schedule_production_loop();
+   } else {
+      ilog("Producer resumed.")
    }
```
heifner: These should be `fc_ilog(_log`

tests/nodeos_forked_chain_test.py
```diff
@@ -290,7 +290,7 @@ def getMinHeadAndLib(prodNodes):
 
         if producerToSlot[lastBlockProducer]["count"]!=inRowCountPerProducer:
             Utils.errorExit("Producer %s, in slot %d, expected to produce %d blocks but produced %d blocks.  At block number %d." %
-                            (blockProducer, slot, inRowCountPerProducer, producerToSlot[lastBlockProducer]["count"], blockNum))
+                            (lastBlockProducer, slot, inRowCountPerProducer, producerToSlot[lastBlockProducer]["count"], blockNum-1))
```
heifner: Add to the PR description that you also updated error message of forked_chain_test.

tests/nodeos_multiple_version_protocol_feature_test.py
```diff
@@ -97,7 +97,8 @@ def hasBlockBecomeIrr():
                           specificExtraNodeosArgs={
                              0:"--http-max-response-time-ms 990000",
                              1:"--http-max-response-time-ms 990000",
-                             2:"--http-max-response-time-ms 990000"},
+                             2:"--http-max-response-time-ms 990000",
+                             3:"--last-block-time-offset-us -200000"},
                           onlySetProds=True,
```
heifner: Maybe add a comment that 1.7 does not have the default of 1.8+ of -200000, so this is needed to avoid 'drop late blocks'.

PR #8720, opened 2020-02-28
Add unit test to verify that the description digests of protocol feat…

swatanabe-b1:
…ures are correct.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/CMakeLists.txt
+++ b/unittests/CMakeLists.txt
@@ -28,9 +28,11 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/contracts.hpp.in ${CMAKE_CURRENT_BINA
 add_subdirectory(snapshots)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/snapshots.hpp.in ${CMAKE_CURRENT_BINARY_DIR}/include/snapshots.hpp ESCAPE_QUOTES)
 
+find_package(Python COMPONENTS Interpreter)
+
 add_custom_command(
    OUTPUT protocol_feature_digest_tests.cpp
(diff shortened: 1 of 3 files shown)


unittests/CMakeLists.txt
```diff
@@ -28,9 +28,17 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/contracts.hpp.in ${CMAKE_CURRENT_BINA
 add_subdirectory(snapshots)
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/snapshots.hpp.in ${CMAKE_CURRENT_BINARY_DIR}/include/snapshots.hpp ESCAPE_QUOTES)
 
+find_package(Python COMPONENTS Interpreter)
```
spoonincode: I believe this was introduced in cmake 3.12. We're still claiming support for 3.8 at this point. Since none of our other python scripts do this, do we really need it? Ubuntu 18.04 is shipping with cmake 3.10 which would make a bump to 3.12 kind of unfortunate.

swatanabe-b1: This was the only way I found to make it work in the CICD CentOS environment.  The difference seems to be that this test needs to run python at build time instead of in the tests.

spoonincode: Okay I think I can fix that.
> run python at build time

This is new too I think... I don't believe we needed python to build previously. (I don't think this is a problem)

spoonincode: see what you think about 1b39154760. This did seem to pass tests (https://buildkite.com/EOSIO/eosio/builds/21230)

swatanabe-b1: It looks good to me.

swatanabe-b1:
I think we do want to hard-code the hash, as it's less prone to inadvertent changes than the description.

PR #8737, opened 2020-03-02
Fix/action results

larryk85:
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -53,6 +53,8 @@ void apply_context::exec_one()
 {
    auto start = fc::time_point::now();
 
+   digest_type act_digest;
+
    const auto& cfg = control.get_global_properties().configuration;
    const account_metadata_object* receiver_account = nullptr;
    try {
@@ -108,6 +110,18 @@ void apply_context::exec_one()
             }
          }
       } FC_RETHROW_EXCEPTIONS( warn, "pending console output: ${console}", ("console", _pending_console_output) )
+
+      if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
+         act_digest =   generate_action_digest(
+                           [this](const char* data, uint32_t datalen) {
+                              return trx_context.hash_with_checktime<digest_type>(data, datalen);
+                           },
+                           *act,
+                           action_return_value
+                        );
+      } else {
+         act_digest = digest_type::hash(*act);
+      }
    } catch( const fc::exception& e ) {
       action_trace& trace = trx_context.get_action_trace( action_ordinal );
       trace.error_code = controller::convert_exception_to_error_code( e );
@@ -122,23 +136,13 @@ void apply_context::exec_one()
    //    * and, the *receiver_account object itself cannot be removed because accounts cannot be deleted in EOSIO.
 
    action_trace& trace = trx_context.get_action_trace( action_ordinal );
+   trace.return_value  = std::move(action_return_value);
    trace.receipt.emplace();
 
    action_receipt& r = *trace.receipt;
    r.receiver        = receiver;
+   r.act_digest      = act_digest;
 
-   if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
-      trace.return_value = std::move(action_return_value);
-      r.act_digest = generate_action_digest(
-                        [this](const char* bytes, uint32_t len) {
-                           return trx_context.hash_with_checktime<digest_type>(bytes, len);
-                        },
-                        *act,
-                        trace.return_value
-                     );
-   } else {
-      r.act_digest = digest_type::hash(*act);
-   }
 
    r.global_sequence  = next_global_sequence();
    r.recv_sequence    = next_recv_sequence( *receiver_account );
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -144,8 +144,13 @@ void apply_context::exec_one()
       r.auth_sequence[auth.actor] = next_auth_sequence( auth.actor );
    }
 
+   auto input_hash = digest_type::hash(act->data);
    if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
-      r.return_value.emplace( std::move( action_return_value ) );
+      auto output_hash = digest_type::hash(action_return_value);
+      r.inout_digest = merkle({input_hash, output_hash});
+      trace.return_value.emplace( std::move( action_return_value ) );
+   } else {
+      r.inout_digest = std::move(input_hash);
    }
 
    trx_context.executed_action_receipt_digests.emplace_back( r.digest() );
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -120,44 +120,21 @@ void apply_context::exec_one()
    //    * a pointer to an object in a chainbase index is not invalidated if other objects in that index are modified, removed, or added;
    //    * a pointer to an object in a chainbase index is not invalidated if the fields of that object are modified;
    //    * and, the *receiver_account object itself cannot be removed because accounts cannot be deleted in EOSIO.
-   const auto& encode = [&](char* bytes, uint32_t len) {
-      const size_t bs = eosio::chain::config::hashing_checktime_block_size;
-      sha256::encoder enc;
-      while ( len > bs ) {
-         enc.write( bytes, bs );
-         bytes += bs;
-         len  -= bs;
-         trx_context.checktime();
-      }
-      enc.write( bytes, len );
-      return enc.result();
-   };
 
-   const auto& encode_action = [&](const action& act, action_receipt& r) {
-      if (control.is_builtin_activated( builtin_protocol_feature_t::action_return_value )) {
-         const action_base* base = &act; // downcast to get only the action_base packed size
-         const auto act_ret_val_size = fc::raw::pack_size(action_return_value) + 1; // + 1 for optional
-         const auto act_pack_size    = fc::raw::pack_size(act);
-         std::vector<char> buff(act_pack_size + act_ret_val_size);
-         datastream<char*> ds(buff.data(), act_pack_size + act_ret_val_size);
-         fc::raw::pack(ds, act);
-         fc::raw::pack(ds, (uint8_t)1);
-         fc::raw::pack(ds, action_return_value);
-         sha256 hashes[2];
-         const size_t lhs_size = fc::raw::pack_size(*base);
-         hashes[0] = encode(buff.data(), lhs_size);
-         hashes[1] = encode(buff.data() + lhs_size, fc::raw::pack_size(act.data) + act_ret_val_size);
-         r.act_digest = encode((char*)&hashes[0], sizeof(hashes));
-      } else {
-         r.act_digest = digest_type::hash(act);
-      }
-   };
+   // checktime encode
+   const auto& encode = [&](char* bytes, uint32_t len) { return trx_context.encode_with_checktime<digest_type>(bytes, len); };
 
    action_trace& trace = trx_context.get_action_trace( action_ordinal );
    trace.receipt.emplace();
+
+   if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
+      trace.return_value.emplace( std::move( action_return_value ) );
+   }
+
    action_receipt& r = *trace.receipt;
    r.receiver         = receiver;
-   encode_action(*act, r);
+
+   r.generate_action_digest(encode, *act, trace.return_value, !control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ));
 
    r.global_sequence  = next_global_sequence();
    r.recv_sequence    = next_recv_sequence( *receiver_account );
@@ -176,9 +153,6 @@ void apply_context::exec_one()
       r.auth_sequence[auth.actor] = next_auth_sequence( auth.actor );
    }
 
-   if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
-      trace.return_value.emplace( std::move( action_return_value ) );
-   }
    trx_context.executed_action_receipt_digests.emplace_back( r.digest() );
 
    finalize_trace( trace, start );
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -120,12 +120,43 @@ void apply_context::exec_one()
    //    * a pointer to an object in a chainbase index is not invalidated if other objects in that index are modified, removed, or added;
    //    * a pointer to an object in a chainbase index is not invalidated if the fields of that object are modified;
    //    * and, the *receiver_account object itself cannot be removed because accounts cannot be deleted in EOSIO.
+   const auto& encode = [&](char* bytes, uint32_t len) {
+      const size_t bs = eosio::chain::config::hashing_checktime_block_size;
+      sha256::encoder enc;
+      while ( len > bs ) {
+         enc.write( bytes, bs );
+         bytes += bs;
+         len  -= bs;
+         trx_context.checktime();
+      }
+      enc.write( bytes, len );
+      return enc.result();
+   };
+
+   const auto& encode_action = [&](const action& act, action_receipt& r) {
+      if (control.is_builtin_activated( builtin_protocol_feature_t::action_return_value )) {
+         const action_base* base = &act; // downcast to get only the action_base packed size
+         auto act_ret_val_size = fc::raw::pack_size(action_return_value);
+         auto act_pack_size    = fc::raw::pack_size(act);
+         std::vector<char> buff(act_pack_size + act_ret_val_size);
+         datastream<char*> ds(buff.data(), act_pack_size + act_ret_val_size);
+         fc::raw::pack(ds, act);
+         fc::raw::pack(ds, action_return_value);
+         sha256 hashes[2];
+         const size_t lhs_size = fc::raw::pack_size(*base);
+         hashes[0] = encode(buff.data(), lhs_size);
+         hashes[1] = encode(buff.data() + lhs_size, fc::raw::pack_size(act.data) + act_ret_val_size);
+         r.act_digest = encode((char*)&hashes[0], sizeof(hashes));
+      } else {
+         r.act_digest = digest_type::hash(act);
+      }
+   };
 
    action_trace& trace = trx_context.get_action_trace( action_ordinal );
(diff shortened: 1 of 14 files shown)


libraries/chain/apply_context.cpp
```diff
@@ -144,10 +144,12 @@ void apply_context::exec_one()
       r.auth_sequence[auth.actor] = next_auth_sequence( auth.actor );
    }
 
+   auto input_hash = digest_type::hash(act->data);
    if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
-      r.return_value.emplace( std::move( action_return_value ) );
+      auto output_hash = digest_type::hash(action_return_value);
```
heifner: Both of these `digest_type::hash` should use instead `encode` method of `wasm_interface` which includes `checktime` calls.

libraries/chain/apply_context.cpp
```diff
@@ -144,10 +144,12 @@ void apply_context::exec_one()
...
+      r.act_digest = merkle({r.act_digest, output_hash});
```
heifner: An `else` could be added so `r.act_digest` is not computed above if not needed.

heifner: Oh, wait I read that as `{input_hash, output_hash}`. `input_hash` is currently not used.

libraries/chain/apply_context.cpp
larryk85: Forgot to add that.  Thanks

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_serialization.hpp
```diff
@@ -574,35 +566,43 @@ inline fc::optional<uint64_t> cap_error_code( const fc::optional<uint64_t>& erro
...
 template <typename ST>
 datastream<ST>& operator<<(datastream<ST>& ds, const history_context_wrapper<bool, eosio::chain::action_trace>& obj) {
+   const auto& trace = obj.obj;
+   if (!trace.return_value) {
+      fc::raw::pack( ds, fc::unsigned_int( 0 ));
+   } else {
+      fc::raw::pack( ds, fc::unsigned_int( 1 ));
+   }
    bool  debug_mode = obj.context;
    fc::raw::pack(ds, fc::unsigned_int(0));
```
swatanabe-b1: You only need one version field.

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_serialization.hpp
```diff
@@ -531,21 +531,13 @@ datastream<ST>& operator<<(datastream<ST>& ds, const history_serial_wrapper<eosi
 
 template <typename ST>
 datastream<ST>& operator<<(datastream<ST>& ds, const history_serial_wrapper<eosio::chain::action_receipt>& obj) {
-   if (!obj.obj.return_value) {
```
tbfleming: Need version field (0)

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_serialization.hpp
tbfleming: Also, the version flipping creates a minefield for adding a _v2 in the future. It seems cleaner to always generate a _v1.

tbfleming: Actually, it created a minefield when merging with another PR. _v1 in that branch includes `account_disk_deltas`, which doesn't exist in _v0.

tbfleming: Here's how I resolved conflicts with that merge:

```
template <typename ST>
datastream<ST>& operator<<(datastream<ST>& ds, const history_context_wrapper<bool, eosio::chain::action_trace>& obj) {
   bool debug_mode = obj.context;
   fc::raw::pack(ds, fc::unsigned_int(1));
   fc::raw::pack(ds, as_type<fc::unsigned_int>(obj.obj.action_ordinal));
   fc::raw::pack(ds, as_type<fc::unsigned_int>(obj.obj.creator_action_ordinal));
   fc::raw::pack(ds, bool(obj.obj.receipt));
   if (obj.obj.receipt) {
      fc::raw::pack(ds, make_history_serial_wrapper(obj.db, as_type<eosio::chain::action_receipt>(*obj.obj.receipt)));
   }
   fc::raw::pack(ds, as_type<uint64_t>(obj.obj.receiver.to_uint64_t()));
   fc::raw::pack(ds, make_history_serial_wrapper(obj.db, as_type<eosio::chain::action>(obj.obj.act)));
   fc::raw::pack(ds, as_type<bool>(obj.obj.context_free));
   fc::raw::pack(ds, as_type<int64_t>(debug_mode ? obj.obj.elapsed.count() : 0));
   if (debug_mode)
      fc::raw::pack(ds, as_type<std::string>(obj.obj.console));
   else
      fc::raw::pack(ds, std::string{});
   history_serialize_container(ds, obj.db, as_type<flat_set<eosio::chain::account_delta>>(obj.obj.account_ram_deltas));
   history_serialize_container(ds, obj.db, as_type<flat_set<eosio::chain::account_delta>>(obj.obj.account_disk_deltas));

   fc::optional<std::string> e;
   if (obj.obj.except) {
      if (debug_mode)
         e = obj.obj.except->to_string();
      else
         e = "Y";
   }
   fc::raw::pack(ds, as_type<fc::optional<std::string>>(e));
   fc::raw::pack(ds,
                 as_type<fc::optional<uint64_t>>(debug_mode ? obj.obj.error_code : cap_error_code(obj.obj.error_code)));
   fc::raw::pack(ds, as_type<rc::optional<eosio::bytes>>(obj.obj.return_value));
   return ds;
}
```

tbfleming: (I undid the `obj.obj => trace` since that made the merge more difficult)

libraries/chain/apply_context.cpp
larryk85: Resolved.

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_serialization.hpp
larryk85: Gotcha, went back to the way it was and added the other fixes.

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_serialization.hpp
larryk85: Fixed.

libraries/chain/apply_context.cpp
```diff
@@ -120,12 +120,44 @@ void apply_context::exec_one()
...
+         fc::raw::pack(ds, action_return_value);
+         sha256 hashes[2];
+         const size_t lhs_size = fc::raw::pack_size(*base);
+         hashes[0] = encode(buff.data(), lhs_size);
+         hashes[1] = encode(buff.data() + lhs_size, fc::raw::pack_size(act.data) + act_ret_val_size);
+         r.act_digest = encode((char*)&hashes[0], sizeof(hashes));
+      } else {
+         r.act_digest = digest_type::hash(act);
+      }
+   };
```
heifner: I think this should be lifted out of here and put somewhere else. Maybe `action_receipt`. Then it could be tested instead duplicating the code in the tests. `encode` could be passed in so tests could use a non-checktime version.

larryk85: Sure, I waffled on this a lot and went with the path of least impact first.  And then made a bunch of changes elsewhere that negated that small impact, so will do make it a method of action_receipt and have it parameterized to either checktime or not.

libraries/chain/include/eosio/chain/action_receipt.hpp
```diff
@@ -15,8 +16,37 @@ namespace eosio { namespace chain {
...
       fc::unsigned_int                code_sequence = 0; ///< total number of setcodes
       fc::unsigned_int                abi_sequence  = 0; ///< total number of setabis
-      fc::optional<std::vector<char>> return_value;      /**< return value of the action; optional is set if action was executed
-                                                              with return value protocol feature enabled */
+
+      static digest_type encode(char* bytes, uint32_t len) {
+         digest_type::encoder enc;
+         enc.write(bytes, len);
+         return enc.result();
+      }
```
swatanabe-b1: Does this really belong here?

libraries/chain/include/eosio/chain/action_receipt.hpp
```diff
@@ -15,8 +16,37 @@ namespace eosio { namespace chain {
...
+            hashes[0]  = enc(buff.data(), lhs_size);
+            hashes[1]  = enc(buff.data() + lhs_size, fc::raw::pack_size(act.data) + extra_size);
+            act_digest = enc((char*)&hashes[0], sizeof(hashes));
+         }
+      }
+
+      template <typename Extra>
+      void generate_action_digest(const action& act, const Extra& extra, bool legacy=true) {
+         generate_action_digest(encode, act, extra, legacy);
+      }
```
swatanabe-b1: Is there any point to having this, when it's completely unused and untested?

libraries/chain/wasm_interface.cpp
```diff
@@ -875,39 +875,39 @@ class crypto_api : public context_aware_api {
       }
 
       void assert_sha256(array_ptr<char> data, uint32_t datalen, const fc::sha256& hash_val) {
-         auto result = encode<fc::sha256::encoder>( data, datalen );
```
swatanabe-b1: `encode` is now unused.

unittests/misc_tests.cpp
```diff
@@ -1110,19 +1110,22 @@ BOOST_AUTO_TEST_CASE(stable_priority_queue_test) {
...
-                         .abi_sequence = 6 };
+                         .abi_sequence = 6,
+                         .inout_digest = fc::sha256("1010")};
       auto d = ar.digest();
       ar.return_value.emplace();
       BOOST_REQUIRE_NE( ar.digest(), d );
 
    } FC_LOG_AND_RETHROW()
 }
+*/
```
arhag: Remove this test if no longer necessary.

unittests/api_tests.cpp
```diff
@@ -2372,6 +2372,38 @@ BOOST_FIXTURE_TEST_CASE(action_ordinal_test, TESTER) { try {
    set_code( N(erin), contracts::test_api_wasm() );
    produce_blocks(1);
 
+   // generate test hash
+   auto gth = [&](auto ac) {
+      action testact;
+      testact.account = ac.account;
+      testact.name    = ac.name;
+      testact.authorization = ac.authorization;
+      std::vector<char> buff(fc::raw::pack_size(testact));
```
arhag: Why copy into `testact` when you can just serialize the `action_base` portion of `ac`? I think it would also make this code a little easier to follow.

unittests/api_tests.cpp
```diff
@@ -2372,6 +2372,38 @@ BOOST_FIXTURE_TEST_CASE(action_ordinal_test, TESTER) { try {
...
+      enc.write(buff.data(), fc::raw::pack_size(*(action_base*)(&testact)));
+      auto ret = enc.result();
+      return ret;
+   };
+
+   // prove act digest
+   auto pad = [&](const auto& action_hash, const auto& test_hash, auto payload, auto value) {
+      std::vector<char> buff(1 + fc::raw::pack_size(value));
+      datastream<char*> ds(buff.data(), buff.size());
+      fc::raw::pack(ds, std::vector<char>{});
```
arhag: Why is `payload` not used? And how do these tests even pass? Are all the action payloads empty in the portion of the tests below that use `pad`?

unittests/api_tests.cpp
```diff
@@ -2498,8 +2540,8 @@ BOOST_FIXTURE_TEST_CASE(action_ordinal_test, TESTER) { try {
    BOOST_REQUIRE_EQUAL(atrace[10].act.account, N(testapi));
    BOOST_REQUIRE_EQUAL(atrace[10].act.name, TEST_METHOD("test_action", "test_action_ordinal4"));
    BOOST_REQUIRE_EQUAL(atrace[10].receipt.valid(), true);
-   BOOST_REQUIRE_EQUAL(atrace[10].receipt->return_value.valid(), true); // return value not set is still a return value, it is just empty
-   BOOST_REQUIRE_EQUAL(atrace[10].receipt->return_value->size(), 0 );   // state_history_plugin keys off presence of return_value for version of receipt
+   BOOST_REQUIRE_EQUAL(atrace[10].return_value.valid(), true); // return value not set is still a return value, it is just empty
+   BOOST_REQUIRE_EQUAL(atrace[10].return_value->size(), 0 );   // state_history_plugin keys off presence of return_value for version of receipt
```
arhag: Comment needs to be updated.

unittests/api_tests.cpp
```diff
@@ -2372,6 +2372,38 @@ BOOST_FIXTURE_TEST_CASE(action_ordinal_test, TESTER) { try {
...
+      datastream<char*> ds(buff.data(), buff.size());
+      fc::raw::pack(ds, std::vector<char>{});
+      fc::raw::pack(ds, value);
+      sha256 hashes[2];
+      hashes[0] = test_hash;
+      sha256::encoder enc;
+      enc.write(buff.data(), ds.tellp());
+      hashes[1] = enc.result();
+      enc.reset();
+      enc.write((char*)&hashes[0], sizeof(hashes));
```
arhag: Again, recommend properly serializing the hashes rather than making assumptions about their internal representation.

unittests/api_tests.cpp
```diff
@@ -2372,6 +2372,38 @@ BOOST_FIXTURE_TEST_CASE(action_ordinal_test, TESTER) { try {
...
+      auto ret = enc.result();
+      return ret;
+   };
+
+   // prove act digest
+   auto pad = [&](const auto& action_hash, const auto& test_hash, auto payload, auto value) {
+      std::vector<char> buff(1 + fc::raw::pack_size(value));
+      datastream<char*> ds(buff.data(), buff.size());
+      fc::raw::pack(ds, std::vector<char>{});
+      fc::raw::pack(ds, value);
```
arhag: Wait, isn't this packing an `optional<vector<char>>`? Again, how do these tests even pass?

libraries/chain/include/eosio/chain/action_receipt.hpp
```diff
@@ -15,8 +16,37 @@ namespace eosio { namespace chain {
...
+            const auto action_size = fc::raw::pack_size(act);
+            const auto extra_size  = fc::raw::pack_size(extra);
+            std::vector<char> buff(action_size + extra_size);
+            fc::datastream<char*> ds(buff.data(), action_size + extra_size);
+            fc::raw::pack(ds, act);
+            fc::raw::pack(ds, extra);
+            digest_type hashes[2];
+            const size_t lhs_size = fc::raw::pack_size(*base);
+            hashes[0]  = enc(buff.data(), lhs_size);
+            hashes[1]  = enc(buff.data() + lhs_size, fc::raw::pack_size(act.data) + extra_size);
```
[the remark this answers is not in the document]
arhag: Also, I just noticed that your code was taking an `fc::optional<std::vector<char>>` type as the `Extra` type and therefore adding an extra 0x01 byte before the serialization of the action return value. The above implementation also avoids that bug by being explicit about the type of the `action_output` argument.

unittests/api_tests.cpp
arhag: Okay, I see why they pass: https://github.com/EOSIO/eos/pull/8737#discussion_r390699583

This test function is also reproducing the buggy hash calculation.

libraries/chain/apply_context.cpp
```diff
@@ -123,9 +123,22 @@ void apply_context::exec_one()
...
+      r.act_digest = generate_action_digest(
+                        [this](const char* bytes, uint32_t len) {
+                           return trx_context.hash_with_checktime<digest_type>(bytes, len);
+                        },
+                        *act,
+                        trace.return_value
+                     );
+   } else {
+      r.act_digest = digest_type::hash(*act);
+   }
```
heifner: Seems like this should be moved up into the try-catch.

libraries/testing/tester.cpp
```diff
@@ -86,7 +99,14 @@ namespace eosio { namespace testing {
       return fc::json::from_string( read_gzipped_snapshot(fn) );
    }
 
-   const fc::microseconds base_tester::abi_serializer_max_time{1000*1000}; // 1s for slow test machines
+   void write_binary_snapshot( const char* fn, const std::string& snapshot ) {
+      return write_gzipped_snapshot( fn, snapshot );
+   }
+
+   void write_json_snapshot( const char* fn, const fc::variant& snapshot) {
+      return write_gzipped_snapshot( fn, fc::json::to_string(snapshot, fc::time_point{fc::microseconds{1000*1000}}) );
```
heifner: Use `abi_serializer_max_time`

unittests/api_tests.cpp
```diff
@@ -2372,6 +2372,36 @@ BOOST_FIXTURE_TEST_CASE(action_ordinal_test, TESTER) { try {
...
+         fc::raw::pack(ds2, act.data);
+         fc::raw::pack(ds2, act_output);
+         fc::raw::pack(ds, sha256::hash(buf2.data(), buf2.size()));
+      }
+
+      digest_type computed_act_digest = sha256::hash(buf1.data(), ds.tellp());
+
+      return expected_act_digest == computed_act_digest;
+   };
+
```
heifner: How does duplicating logic of `generate_action_digest` test anything? Why not just call `generate_action_digest` ?

arhag: `generate_action_digest` is technically a different implementation (more general) and I'm thinking we may have plans to further generalize it in the future. This more specific implementation acts as a safeguard to check for regressions/bugs as the code is further refactored.

libraries/testing/tester.cpp
arhag: I'm just going to remove the write snapshot functions introduced in this PR because they don't seem to serve any purpose currently and in fact aren't even used.

libraries/chain/apply_context.cpp
heifner: From Areg: "A light client following the blockchain will know when to switch the rules by detecting activation of the relevant protocol feature."

heifner: Ok.

PR #8739, opened 2020-03-02
Tests and basic implementation of Responses for Trace API

b1bart:
In this PR for review:
- [ ] basic response variant formatting w/o ABI considerations
- [ ] basic tests framework for "response" black box
- [ ] basic tests
- [ ] edge test for missing data/metadata
- [ ] edge test for corrupt data/metadata
- [ ] test for deadline exceptions

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/request_handler.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/request_handler.hpp
@@ -1,0 +1,118 @@
+#pragma once
+
+#include <fc/variant.hpp>
+#include <eosio/trace_api_plugin/metadata_log.hpp>
+#include <eosio/trace_api_plugin/data_log.hpp>
+
+namespace eosio::trace_api_plugin {
+
+   using now_function = std::function<fc::time_point()>;
+
+   namespace detail {
+      class response_formatter {
+      public:
+         static fc::variant process_block( const block_trace_v0& trace, const  std::optional<lib_entry_v0>& best_lib, const now_function& now, const fc::time_point& deadline );
+      };
+   }
+
+   class deadline_exceeded : public std::runtime_error {
+   public:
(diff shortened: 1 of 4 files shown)


plugins/trace_api_plugin/include/eosio/trace_api_plugin/request_handler.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+         // scan for the block offset and latest LIB
+         try {
+            logfile_provider.scan_metadata_log_from(block_height, 0, [&](const metadata_log_entry& e) -> bool {
+               if (e.contains<block_entry_v0>()) {
+                  const auto& block = e.get<block_entry_v0>();
+                  if (block.number == block_height) {
+                     block_offset = block.offset;
+                  }
+               } else if (e.contains<lib_entry_v0>()) {
```
brianjohnson5972: I am not sure how scan_metadata_log_from would return a lib_entry_v0 based on just a block_height value?  I was thinking the lib_entry_v0 was just for the readers understanding of the file data, but now thinking we need to work through the use case that causes the reader to return it.

b1bart: The logfile reader should just be reading sequentially with no inspection on what its passing.   As long as the slice file for a given block height contains all `lib_entry_v0` structures relevant to blocks in that slice then we should be able to find the "Best" one from that scan and it should ideally be an entry which is at or past the end of the slice once that slice is final.

b1bart: Will accept the PR and move this Discussion off comments

brianjohnson5972: ignore me, forgot that my reader just keeps reading till you tell it to stop.

Issue #8703, opened 2020-02-26
History_Plugin data pruning


aclark-b1:
Add an RPC call to history_plugin that prunes all data before a specified block height from the table

heifner:
history_plugin is deprecated. The community is welcome to grab the associated PRs above, but these will not be merged.

PR #8740, opened 2020-03-02
Add prune history_plugin feature - develop

victorj8:
## Change Description

This change adds an endpoint which can be used to free memory (prune) used by the history plugin. #8703 

Does *not* prune data associated with `/v1/history/get_key_accounts` or `/v1/history/get_controlled_accounts`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [X ] API Changes
The next endpoint was added

/v1/history/prune_history

It has a parameter called "height" which represents the maximum height of the block which we want to prune data from history plugin.

- New option that controls timeout of `prune_history` and existing `get_actions`:
  - `history-api-timeout-us` Timeout for history API calls (microseconds). Default 100000.

Note: API call only available when running in `read-mode = irreversible`.

WARNING: If `history_api_plugin` is deployed to a public API node then external proxy settings should be put into place to restrict `/v1/history/prune_history`. The intended use of `history_api_plugin` is only for internal nodeos not exposed to the public.

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/plugins/history_plugin/history_plugin.cpp
+++ b/plugins/history_plugin/history_plugin.cpp
@@ -15,7 +15,6 @@ namespace eosio {
    using boost::signals2::scoped_connection;
 
    static appbase::abstract_plugin& _history_plugin = app().register_plugin<history_plugin>();
-   static int32_t prune_timeout = 100000;
 
 
    struct account_history_object : public chainbase::object<account_history_object_type, account_history_object>  {
@@ -69,6 +68,12 @@ namespace eosio {
             composite_key< account_history_object,
                member<account_history_object, account_name, &account_history_object::account >,
                member<account_history_object, int32_t, &account_history_object::account_sequence_num >
+            >
+         >,
+         ordered_unique<tag<by_action_sequence_num>,
+            composite_key< account_history_object,
+               member<account_history_object, uint64_t, &account_history_object::action_sequence_num >,
+               member<account_history_object, account_history_object::id_type, &account_history_object::id>
             >
          >
       >
@@ -143,6 +148,7 @@ namespace eosio {
          std::set<filter_entry> filter_out;
          chain_plugin*          chain_plug = nullptr;
          fc::optional<scoped_connection> applied_transaction_connection;
+         fc::microseconds history_api_timeout_us = fc::microseconds(100000);
          int64_t last_id_pruned = 0;
 
           bool filter(const action_trace& act) {
@@ -309,8 +315,8 @@ namespace eosio {
              "Do not track actions which match receiver:action:actor. Action and Actor both blank excludes all from Reciever. Actor blank excludes all from reciever:action. Receiver may not be blank.")
             ;
       cfg.add_options()
-            ("history-plugin-prune-timeout-us", bpo::value<int32_t>()->composing(),
-             "Timeout for prune history request (microseconds).")
+            ("history-api-timeout-us", bpo::value<int32_t>()->composing(),
+             "Timeout for history API calls (microseconds).")
             ;
    }
 
@@ -346,8 +352,8 @@ namespace eosio {
             }
          }
 
-        if( options.count( "history-plugin-prune-timeout-us" )) {
-            prune_timeout = options.at( "history-plugin-prune-timeout-us" ).as<int32_t>();
+        if( options.count( "history-api-timeout-us" )) {
+            my->history_api_timeout_us = fc::microseconds(options.at( "history-api-timeout-us" ).as<int32_t>());
         }
 
          my->chain_plug = app().find_plugin<chain_plugin>();
@@ -439,7 +445,7 @@ namespace eosio {
                                  });
 
            end_time = fc::time_point::now();
-           if( end_time - start_time > fc::microseconds(100000) ) {
+           if( end_time - start_time > history->history_api_timeout_us ) {
               result.time_limit_exceeded_error = true;
               break;
            }
@@ -450,36 +456,50 @@ namespace eosio {
 
       read_write::prune_history_result read_write::prune_history( const read_write::prune_history_params& params ) const {
          auto& chain = history->chain_plug->chain();
+
+         EOS_ASSERT( chain.get_read_mode() == db_read_mode::IRREVERSIBLE, chain::plugin_exception, "attempting to call prune history while NOT in irreversible read mode" );
+
          chainbase::database& db = const_cast<chainbase::database&>( chain.db() ); // Override read-only access to state DB (highly unrecommended practice!)
 
-         prune_history_result result{-1, false};
+         prune_history_result result{0, false};
 
-         if(chain.irreversible_read_mode()) {
-            auto start_time = fc::time_point::now();
+         auto start_time = fc::time_point::now();
 
-            const auto& idx = db.get_index<action_history_index, by_id>();
-            auto itr = idx.lower_bound(history->last_id_pruned);
+         const auto& action_idx = db.get_index<action_history_index, by_id>();
+         auto itr_action = action_idx.lower_bound(history->last_id_pruned);
 
-            while( itr != idx.end() &&  itr->block_num <= params.height ) {
-               fc::datastream<const char*> ds( itr->packed_action_trace.data(), itr->packed_action_trace.size() );
-               action_trace t;
-               fc::raw::unpack( ds, t );
+         while( itr_action != action_idx.end() ) {
+           if(itr_action->block_num <= params.height) {
+             const auto& account_idx = db.get_index<account_history_index, by_action_sequence_num>();
+             auto itr_account = account_idx.lower_bound( boost::make_tuple( itr_action->action_sequence_num, 0 ) );
 
-               auto &a = db.get<account_history_object, by_account_action_seq>(boost::make_tuple(t.receipt->receiver, t.receipt->recv_sequence));
-               db.remove(a);
+             while(itr_account != account_idx.end() && itr_account->action_sequence_num == itr_action->action_sequence_num) {
+               auto itr_account_copy = itr_account;
+               itr_account++;
+               db.remove(*itr_account_copy);
 
-               auto itr_copy = itr;
-               itr++;
-               db.remove(*itr_copy);
+               result.records_removed++;
+             }
 
-               result.last_id = itr->id._id;
+             auto itr_action_copy = itr_action;
+             itr_action++;
+             db.remove(*itr_action_copy);
 
-               auto now = fc::time_point::now();
-               if( now - start_time > fc::microseconds(prune_timeout) ) {
-                  result.time_limit_exceeded_error = true;
-                  break;
-               }
-            }
+             result.records_removed++;
+             history->last_id_pruned = itr_action->id._id;
+           } else {
+             itr_action++;
+           }
+
+           auto now = fc::time_point::now();
+           if( now - start_time > history->history_api_timeout_us ) {
+             result.time_limit_exceeded_error = true;
+             break;
+           }
+         }
+
+         if (!result.time_limit_exceeded_error) {
+           history->last_id_pruned = 0;
          }
 
          return result;
--- a/plugins/history_plugin/history_plugin.cpp
+++ b/plugins/history_plugin/history_plugin.cpp
@@ -466,29 +466,35 @@ namespace eosio {
          const auto& action_idx = db.get_index<action_history_index, by_id>();
          auto itr_action = action_idx.lower_bound(history->last_id_pruned);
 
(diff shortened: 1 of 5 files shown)


plugins/history_plugin/history_plugin.cpp
```diff
@@ -15,6 +15,7 @@ namespace eosio {
    using boost::signals2::scoped_connection;
 
    static appbase::abstract_plugin& _history_plugin = app().register_plugin<history_plugin>();
+   static int32_t prune_timeout = 100000;
```
heifner: Should be member of `history_plugin_impl` not a static.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -438,6 +448,42 @@ namespace eosio {
...
+         auto& chain = history->chain_plug->chain();
+         chainbase::database& db = const_cast<chainbase::database&>( chain.db() ); // Override read-only access to state DB (highly unrecommended practice!)
+
+         prune_history_result result{-1, false};
+
+         if(chain.irreversible_read_mode()) {
+            auto start_time = fc::time_point::now();
+
+            const auto& idx = db.get_index<action_history_index, by_id>();
+            auto itr = idx.lower_bound(history->last_id_pruned);
```
heifner: `last_id_pruned` is not updated. It should be updated instead of `result.last_id` below.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -438,6 +448,42 @@ namespace eosio {
...
+
+            while( itr != idx.end() &&  itr->block_num <= params.height ) {
```
heifner: This exits on the first block that is found that is larger that `params.height`. Since you are iterating `by_id` there is no order you can assume for `block_num`.

plugins/history_plugin/history_plugin.cpp
heifner: You will also need to reset `last_id_pruned` if you don't timeout.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -438,6 +448,42 @@ namespace eosio {
...
+               fc::datastream<const char*> ds( itr->packed_action_trace.data(), itr->packed_action_trace.size() );
+               action_trace t;
+               fc::raw::unpack( ds, t );
+
+               auto &a = db.get<account_history_object, by_account_action_seq>(boost::make_tuple(t.receipt->receiver, t.receipt->recv_sequence));
```
heifner: Looks like you need to add an index to `account_history_object` for `account_sequence_num` and use that to find and remove. I know we talked about not adding an index, but I don't see any way around it. Anyone not running with the greymass patch for history_plugin will need to do a replay or start from a given block via a snapshot anyway. The `account_sequence_num` index on `account_history_object` will need to fold in `id` so that it is `ordered_unique`. See large comment in `chain/types.hpp` for details.

The good news about that is unpack of the `packed_action_trace` will no longer be needed.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -438,6 +448,42 @@ namespace eosio {
         return result;
       }
 
+      read_write::prune_history_result read_write::prune_history( const read_write::prune_history_params& params ) const {
+         auto& chain = history->chain_plug->chain();
+         chainbase::database& db = const_cast<chainbase::database&>( chain.db() ); // Override read-only access to state DB (highly unrecommended practice!)
+
+         prune_history_result result{-1, false};
+
+         if(chain.irreversible_read_mode()) {
```
heifner: This should be an `EOS_ASSERT` not an `if`.

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
...
+   public:
+      read_write(history_const_ptr&& history)
+        : history(history) {}
+
+      struct prune_history_params {
+        int32_t             height;     // height of the last block we want to prune data of
+      };
+
+      struct prune_history_result {
+        int64_t             last_id;   // last id pruned
```
heifner: Since `prune` doesn't take an `id` this is not helpful. How about just returning the # of records removed.

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
 
 namespace history_apis {
 
+class read_write {
+   history_const_ptr history;
+
+   public:
+      read_write(history_const_ptr&& history)
```
heifner: Make `explicit` and accept `history_ptr`

plugins/history_plugin/history_plugin.cpp
```diff
@@ -340,6 +346,10 @@ namespace eosio {
             }
          }
 
+        if( options.count( "history-plugin-prune-timeout-us" )) {
+            prune_timeout = options.at( "history-plugin-prune-timeout-us" ).as<int32_t>();
```
heifner: Change to `history-api-timeout-us` and also use for value in `get_actions` instead of hard-coded `100000`

heifner: Rename `prune_timeout` to `history_api_timeout_us` and make it a `fc::microseconds` instead of `int32_t`

plugins/history_plugin/history_plugin.cpp
```diff
@@ -306,6 +308,10 @@ namespace eosio {
             ("filter-out,F", bpo::value<vector<string>>()->composing(),
              "Do not track actions which match receiver:action:actor. Action and Actor both blank excludes all from Reciever. Actor blank excludes all from reciever:action. Receiver may not be blank.")
             ;
+      cfg.add_options()
+            ("history-plugin-prune-timeout-us", bpo::value<int32_t>()->composing(),
+             "Timeout for prune history request (microseconds).")
```
heifner: See below. Change description to "Timeout for history API calls (microseconds)."

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
 
 namespace history_apis {
 
+class read_write {
+   history_const_ptr history;
```
heifner: Should be `history_ptr`

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
...
+
+   public:
+      read_write(history_const_ptr&& history)
+        : history(history) {}
```
heifner: I realize you are only following example, but this should use `std::forward` or just take by value and use `std::move`.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -306,6 +314,10 @@ namespace eosio {
             ("filter-out,F", bpo::value<vector<string>>()->composing(),
              "Do not track actions which match receiver:action:actor. Action and Actor both blank excludes all from Reciever. Actor blank excludes all from reciever:action. Receiver may not be blank.")
             ;
+      cfg.add_options()
+            ("history-api-timeout-us", bpo::value<int32_t>()->composing(),
```
heifner: Remove the `composing`

heifner: Add a `->default( history_api_timeout_us.count() )`

plugins/history_plugin/history_plugin.cpp
```diff
@@ -340,6 +352,10 @@ namespace eosio {
             }
          }
 
+        if( options.count( "history-api-timeout-us" )) {
+            my->history_api_timeout_us = fc::microseconds(options.at( "history-api-timeout-us" ).as<int32_t>());
+        }
```
heifner: With a default you will not need the `options.count`

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
...
+      struct prune_history_params {
+        int32_t             height;            // height of the last block we want to prune data of
+      };
+
+      struct prune_history_result {
+        int64_t             records_removed;   // # of records removed
+        bool                time_limit_exceeded_error;
+      };
+
+      prune_history_result prune_history( const prune_history_params& )const;
```
heifner: This method should not be const.

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -126,6 +145,7 @@ class history_plugin : public plugin<history_plugin> {
       void plugin_shutdown();
 
       history_apis::read_only  get_read_only_api()const { return history_apis::read_only(history_const_ptr(my)); }
+      history_apis::read_write get_read_write_api() { return history_apis::read_write(std::move(my)); }
```
heifner: You can't move `my` here. You need `my`.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -142,6 +148,8 @@ namespace eosio {
          std::set<filter_entry> filter_out;
          chain_plugin*          chain_plug = nullptr;
          fc::optional<scoped_connection> applied_transaction_connection;
+         fc::microseconds history_api_timeout_us = fc::microseconds(100000);
+         int64_t last_id_pruned = 0;
```
heifner: As you have it, this is not last id pruned but rather next_id_to_prune. Either rename to something more descriptive or update to value before remove so that it is actually last_id_pruned.

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
...
 
+class read_write {
+   history_ptr history;
+
+   public:
+      explicit read_write(history_ptr&& history)
+        : history(history) {}
+
+      struct prune_history_params {
+        int32_t             height;            // height of the last block we want to prune data of
```
heifner: Make this an `uint32_t`

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
...
+      };
+
+      struct prune_history_result {
+        int64_t             records_removed;   // # of records removed
```
heifner: Make this an `uint64_t`

plugins/history_plugin/history_plugin.cpp
```diff
@@ -438,6 +452,57 @@ namespace eosio {
...
+
+         prune_history_result result{0, false};
+
+         auto start_time = fc::time_point::now();
+
+         const auto& action_idx = db.get_index<action_history_index, by_id>();
+         auto itr_action = action_idx.lower_bound(history->last_id_pruned);
+
+         while( itr_action != action_idx.end() ) {
+           if(itr_action->block_num <= params.height) {
```
heifner: Since `id` is incrementing with `block_num`, you can break out of loop once `block_num` > `params.height`.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -438,6 +452,57 @@ namespace eosio {
...
+             const auto& account_idx = db.get_index<account_history_index, by_action_sequence_num>();
+             auto itr_account = account_idx.lower_bound( boost::make_tuple( itr_action->action_sequence_num, 0 ) );
+
+             while(itr_account != account_idx.end() && itr_account->action_sequence_num == itr_action->action_sequence_num) {
+               auto itr_account_copy = itr_account;
+               itr_account++;
+               db.remove(*itr_account_copy);
+
+               result.records_removed++;
+             }
```
heifner: Add deadline check to this loop.

plugins/history_plugin/include/eosio/history_plugin/history_plugin.hpp
```diff
@@ -18,6 +18,25 @@ namespace eosio {
 
 namespace history_apis {
 
+class read_write {
+   history_ptr history;
+
+   public:
+      explicit read_write(history_ptr&& history)
+        : history(history) {}
+
```
heifner: This should be `: history(std::move(history))`

victorj8:
Still pending to implement to add a way to restrict the prune_history call, one suggestion is to add an option to history plugin called history-enable-prune which will control this behavior (and default it to false).

PR #8745, opened 2020-03-03
Added concept of data handlers to trace_api_plugin

b1bart:
data handlers are used to parse action data and are distinct from the response subsystem to maintain separation of concerns.  In this PR:

- [ ] : mock data handler for the existing response tests
- [ ] : base64 data handler for reference
- [ ] : abi data handler with support for a dictionary of abi's shared amongst multiple response handlers 

the abi data handler supports an injected logger function to separate the logging of exceptions that cause it to revert to hex encoding from logic of producing data

Unfortunately, abi_serializer has a native concept of "now" which makes it difficult to test easily.  As such, there is no deadline provided to the abi serialization process inside a single action.   The response processor will enforce deadlines between actions and there is no descent into recursive ABI'd structures (like `eosio.msig`) so this should be a minimal attack surface.

### DIFF
--- a/plugins/trace_api_plugin/abi_data_handler.cpp
+++ b/plugins/trace_api_plugin/abi_data_handler.cpp
@@ -9,12 +9,19 @@ namespace eosio::trace_api_plugin {
    }
 
    fc::variant abi_data_handler::process_data(const action_trace_v0& action, const fc::time_point& deadline) {
-      if (abi_serializer_by_account.count(action.account) == 0) {
-         return fc::to_hex(action.data.data(), action.data.size());
+      if (abi_serializer_by_account.count(action.account) > 0) {
+         const auto& serializer_p = abi_serializer_by_account.at(action.account);
+         auto type_name = serializer_p->get_action_type(action.action);
+
+         if (!type_name.empty()) {
+            try {
+               return serializer_p->binary_to_variant(type_name, action.data, fc::microseconds::maximum());
+            } catch (...) {
+               if (log) log(std::current_exception());
+            }
+         }
       }
 
-      const auto& serializer_p = abi_serializer_by_account.at(action.account);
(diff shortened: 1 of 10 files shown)


plugins/trace_api_plugin/abi_data_handler.cpp
```diff
@@ -0,0 +1,27 @@
...
+   }
+
+   fc::variant abi_data_handler::process_data(const action_trace_v0& action, const fc::time_point& deadline) {
+      if (abi_serializer_by_account.count(action.account) > 0) {
+         const auto& serializer_p = abi_serializer_by_account.at(action.account);
+         auto type_name = serializer_p->get_action_type(action.action);
+
+         if (!type_name.empty()) {
+            try {
+               return serializer_p->binary_to_variant(type_name, action.data, fc::microseconds::maximum());
```
heifner: Why are you not using `deadline` here?

b1bart: adding resolution from the live discussion here:

Since the operators will directly control available ABIs for the first deliverable, we can leave this as `fc::microseconds::maximum()` .  In the future version where ABIs are not implicitly trusted (eg they are derived from the running chain) we will need to enforce a deadline. 

At that time, we will likely make intrusive changes to `abi_serializer` to support the concept of a caller defined `yield` function replacing the internally plumbed time checks.  This allows separation of concerns over why a stalled/long-running call needs to terminate from the actual logic of `abi_serializer`.  This does a few important things:
* it allows the concept of "time" to be mocked in tests for determinism instead of being deeply coupled to the system clock etc
* it allows interruption for reasons other than time, for instance, a critical priority task needs to interrupt ABI processing occasionally and restricting _all_ ABI processing to short time slices to service that response time guarantee is too aggressive.

PR #8746, opened 2020-03-04
Update helper message for CPU

jafri:
A common problem for EOSIO users is the cryptic message when out of CPU, this auxiliary update should clarify the issue for users.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -459,7 +459,7 @@ namespace eosio { namespace chain {
                const int64_t cpu_limit = (cpu_limited_by_account ? account_cpu_limit : objective_duration_limit.count());
                EOS_ASSERT( billed_us <= cpu_limit,
                            tx_cpu_usage_exceeded,
-                           "${desc} CPU time (${billed} us) is greater than the maximum billable CPU time for the transaction (${billable} us)",
+                           "${desc} Transaction CPU time (${billed} us) is greater than the maximum available (${billable} us). Please stake more resources to CPU.",
                            ("desc", (estimate ? "estimated" : "billed"))("billed", billed_us)( "billable", cpu_limit )
                );
             }


libraries/chain/transaction_context.cpp
```diff
@@ -459,7 +459,7 @@ namespace eosio { namespace chain {
                const int64_t cpu_limit = (cpu_limited_by_account ? account_cpu_limit : objective_duration_limit.count());
                EOS_ASSERT( billed_us <= cpu_limit,
                            tx_cpu_usage_exceeded,
-                           "${desc} CPU time (${billed} us) is greater than the maximum billable CPU time for the transaction (${billable} us)",
+                           "${desc} Transaction CPU time (${billed} us) is greater than the maximum available (${billable} us). Please stake more resources to CPU.",
```
swatanabe-b1: This isn't alway accurate.  This message will also appear when the cpu time exceeds the chain's configured maximum transaction CPU usage.  Also, advice of how to fix the error doesn't really belong here, as it depends on the system contract.
- "Transaction" should not be capitalized.
- The other messages from this function should be consistent.

Soleone:
> Also, advice of how to fix the error doesn't really belong here, as it depends on the system contract.

This is a very common error, and I have literally seen a hundred confused users by now that don't know how to interpret this message. It is paramount that this needs to be addressed, and I applaud Syed for finally just stepping up and doing something about it.

A suggested action would be incredibly helpful, but of course it is hard to get something generic enough that works perfect in all set ups. Defaulting to current most popular EOSIO chains behavior, i.e. more tokens need to be staked does not seem unreasonable. But users need to know the next step on what to look for next, and hopefully they have heard already about the concept of staking. But I don't see a way on how leaving a user without any next suggested action here to resolve the error can ever work.

This is one of the few interfaces the user has directly with the system, there is barely any other direct messaging to the user, let's ensure it's good.

I would love if we could keep this as simple as possible and either just ship a small improvement now and something better later (e.g. configurable error message per chain). Just anything please to stop users having this terribly confusing image of EOS when in reality it is kind of elegant how the staking mechanism works.

Thanks :)

libraries/chain/transaction_context.cpp
arhag: @jafri:

`cpu_limited_by_account` can be used to determine whether the limit was reached because the account's quota is too low (and therefore it may help to recommend increasing the account's CPU quota) or if it is due to other objective reasons (e.g. the limit was determined by either the maximum transaction CPU consensus parameter or by the maximum specified in the transaction header).

And as @swatanabe-b1 mentioned, the language here should not be specific to any particular system contract. So even the word "stake" isn't  really appropriate. Personally, I think a phrase like "increasing CPU resources may help" would be acceptable, but only in the case where we know the limit was reached due to a billed account's quota (i.e. if `cpu_limited_by_account` is true).

Soleone:
What can we do here to drive this forward? It looks small, but I get questions from users multiple times a week on what that cryptic error message is supposed to mean. Would love to make this better just a bit without too much bike shedding.

arhag:
@Soleone:

See https://github.com/EOSIO/eos/pull/8746#discussion_r397480163.

The changes needed to address the mentioned concerns are:

- Use `cpu_limited_by_account` to distinguish between two different cases: the first case is where the limit is reached because the billed account(s) quota is too low; the second case is where the limit is determined by other objective reasons (maximum consensus limit on transaction CPU or maximum per-transaction limit set in the header). An appropriate error message must be given for each of the two cases. It is not appropriate to suggest increasing CPU resources in the latter case since that isn't the limiting factor.
- Avoid using "stake" in the error message. My suggested alternative for the error message in the first case described above is `"Transaction's ${desc} CPU time (${billed} us) is greater than the maximum available (${billable} us). Increasing CPU resources may help."`.
- Consistency in language and grammar for all the error messages of all the exceptions that are explicitly thrown from within `validate_account_cpu_usage`.
- Ensure all tests continue to pass.

PR #8747, opened 2020-03-04
history_plugin overflow fix - 2.0

victorj8:
## Change Description
Cherry picked some useful changes to history plugin from
https://github.com/greymass/eos/commit/e58a0af9199d0474c33987e9622bac218d0fe77c

Co-Authored-By: Scott Sallinen <scotts@ece.ubc.ca>

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -380,6 +380,12 @@ struct controller_impl {
       } catch ( std::exception& e ) {
          wlog( "std::exception: ${details}", ("details", e.what()) );
       } catch ( ... ) {
+         auto expPtr = std::current_exception();
+         try {
+           if(expPtr) std::rethrow_exception(expPtr);
+         } catch(const std::exception& e) {
+           wlog( "${details}", ("details", e.what()) );
+         }
          wlog( "signal handler threw exception" );
       }
    }
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -380,12 +380,6 @@ struct controller_impl {
       } catch ( std::exception& e ) {
          wlog( "std::exception: ${details}", ("details", e.what()) );
       } catch ( ... ) {
-         auto expPtr = std::current_exception();
-         try {
-           if(expPtr) std::rethrow_exception(expPtr);
-         } catch(const std::exception& e) {
-           wlog( "${details}", ("details", e.what()) );
-         }
          wlog( "signal handler threw exception" );
       }
    }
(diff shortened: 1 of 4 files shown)


libraries/chain/controller.cpp
```diff
@@ -380,6 +380,12 @@ struct controller_impl {
       } catch ( std::exception& e ) {
          wlog( "std::exception: ${details}", ("details", e.what()) );
       } catch ( ... ) {
+         auto expPtr = std::current_exception();
+         try {
+           if(expPtr) std::rethrow_exception(expPtr);
+         } catch(const std::exception& e) {
+           wlog( "${details}", ("details", e.what()) );
+         }
```
heifner: You can remove all this, it is not needed because `std::exception` handling was added above since this patch was created.

plugins/history_plugin/history_plugin.cpp
```diff
@@ -112,7 +112,9 @@ namespace eosio {
 
    static void add(chainbase::database& db, const vector<permission_level_weight>& controlling_accounts, const account_name& account_name, const permission_name& permission)
    {
-      for (auto controlling_account : controlling_accounts ) {
+      std::set<permission_level_weight> controlling_accounts_set(controlling_accounts.begin(), controlling_accounts.end());
+
+      for (auto controlling_account : controlling_accounts_set ) {
```
heifner: Sorry for the bit of misdirection here. The idea is to avoid uniqueness constraint on `account_control_history_object`. I think the most direct and easiest approach is to first find by   `[controlled_account, controlled_permission, controlling_account]` and only `create` if not found.

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -84,6 +84,10 @@ struct permission_level_weight {
    friend bool operator == ( const permission_level_weight& lhs, const permission_level_weight& rhs ) {
       return tie( lhs.permission, lhs.weight ) == tie( rhs.permission, rhs.weight );
    }
+
+   friend bool operator < ( const permission_level_weight& lhs, const permission_level_weight& rhs ) {
+      return tie( lhs.permission, lhs.weight ) < tie( rhs.permission, rhs.weight );
+   }
 };
```
heifner: Go ahead and remove this as not needed.

victorj8:
This PR will be closed, a description of the problem was referenced in #8772

PR #8748, opened 2020-03-04
Add prune history_plugin feature - 2.0

victorj8:
## Change Description
This change adds an endpoint which can be used to free memory (prune) used by the history plugin. #8703 

Does *not* prune data associated with `/v1/history/get_key_accounts` or `/v1/history/get_controlled_accounts`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [X ] API Changes
The next endpoint was added

/v1/history/prune_history

It has a parameter called "height" which represents the maximum height of the block which we want to prune data from history plugin.

- New option that controls timeout of `prune_history` and existing `get_actions`:
  - `history-api-timeout-us` Timeout for history API calls (microseconds). Default 100000.

Note: API call only available when running in `read-mode = irreversible`.

WARNING: If `history_api_plugin` is deployed to a public API node then external proxy settings should be put into place to restrict `/v1/history/prune_history`. The intended use of `history_api_plugin` is only for internal nodeos not exposed to the public.

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2834,10 +2834,6 @@ block_id_type controller::last_irreversible_block_id() const {
    return get_block_id_for_num( lib_num );
 }
 
-bool controller::irreversible_read_mode() const {
-   return my->read_mode == db_read_mode::IRREVERSIBLE;
-}
-
 const dynamic_global_property_object& controller::get_dynamic_global_properties()const {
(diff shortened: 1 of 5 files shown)


victorj8:
Still pending to implement to add a way to restrict the prune_history call, one suggestion is to add an option to history plugin called history-enable-prune which will control this behavior (and default it to false).

PR #8750, opened 2020-03-04
Move parts of state-history-plugin to libraries/state_history

tbfleming:
## Change Description

This moves some parts of ship's logic into a library so the newest new CDT tester may include it.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/state_history/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
...
+             trace_converter.cpp
+             ${HEADERS}
+           )
+
+target_link_libraries( state_history 
+                       PUBLIC eosio_chain fc chainbase softfloat
+                     )
+
+target_include_directories( state_history
+                            PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_SOURCE_DIR}/../wasm-jit/Include"
```
arhag, line 16: It doesn't seem you actually need `"${CMAKE_CURRENT_SOURCE_DIR}/../wasm-jit/Include"` here.

libraries/state_history/trace_converter.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <eosio/state_history/trace_converter.hpp>
+
+namespace eosio {
+namespace state_history {
+
+using eosio::chain::packed_transaction;
+using eosio::chain::plugin_exception;
+
+bool is_onblock(const transaction_trace_ptr& p) {
+   if (p->action_traces.size() != 1)
```
arhag, line 11: Should be:
```cpp
if (p->action_traces.empty())
```

The `is_onblock` in `trace_api/chain_extraction.hpp` also needs a similar correction.

gleehokie:
@tbfleming should the new directory be named libraries/state-history instead of libraries/state_history (dash instead of underscore)?  We use "-" as the separator for directory names in the eos repo currently.

tbfleming:
@gleehokie `state-history` isn't a valid namespace name and I thought it'd get confusing to have a directory name subtly different.

libraries/state_history/CMakeLists.txt
tbfleming: removed

libraries/state_history/trace_converter.cpp
tbfleming: changed

PR #8753, opened 2020-03-05
Trace api extraction 2.0 - first test passing

heifner:
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/chain_extraction.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/chain_extraction.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <eosio/trace_api_plugin/trace.hpp>
+#include <eosio/trace_api_plugin/extract_util.hpp>
 #include <exception>
 #include <functional>
 #include <map>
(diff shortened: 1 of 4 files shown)


plugins/trace_api_plugin/include/eosio/trace_api_plugin/chain_extraction.hpp
```diff
@@ -19,8 +20,8 @@ class chain_extraction_impl_type {
     * @param except_handler called on exceptions, logging if any is left to the user
     * @param shutdown called when this class determines application should shutdown because of fatal error
     */
-   chain_extraction_impl_type( StoreProvider& store, std::function<void(std::exception_ptr)> except_handler )
-   : store(store)
+   chain_extraction_impl_type( StoreProvider store, std::function<void(std::exception_ptr)> except_handler )
+   : store(std::move(store))
```
b1bart: As this is a templated parameter, I think it should be `StoreProvider&& store` and `std::forward<StoreProvider>(store)`.  I've made the same situation in my code come to think of it 😇

heifner: Not for class template type. If the constructor was templated then yes.

b1bart: right... :). ☕️

PR #8760, opened 2020-03-05
Trace API Plugin and configuration scaffolding 

b1bart:
In this PR:
- [ ] basic plugin scaffolding for (2) plugins
  - a full system plugin `_trace_api_plugin` (temp name until the namespace is renamed)
  - a plugin that only serves the RPC api `trace_api_rpc_plugin` 
  - only the full system will be linked in and available to `nodeos`
  - the rpc plugin is laying the ground work for an independent process  and to force separation of internal logic
- [ ] new header and test with what little testable logic fell out of configuration code
- [ ] new concept for adding `context` to exceptions so that logging appropriately correlates messages to the code that produced them.  There is an example of how this is wired in through the `abi_data_handler` so that the throwing code can provide the context to the injected log function.

### DIFF
--- a/plugins/trace_api_plugin/trace_api_plugin.cpp
+++ b/plugins/trace_api_plugin/trace_api_plugin.cpp
@@ -1,11 +1,12 @@
 #include <eosio/trace_api_plugin/trace_api_plugin.hpp>
 
-#include <regex>
-#include <fc/io/json.hpp>
 #include <eosio/trace_api_plugin/abi_data_handler.hpp>
 #include <eosio/trace_api_plugin/request_handler.hpp>
 
+#include <eosio/trace_api_plugin/configuration_utils.hpp>
+
 using namespace eosio::trace_api_plugin;
+using namespace eosio::trace_api_plugin::configuration_utils;
 
 namespace {
    const std::string logger_name("trace_api");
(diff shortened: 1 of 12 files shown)


plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -0,0 +1,229 @@
...
+                  "  \"w\" : the value is in weeks\n"
+                  "a value of \"-1\" will disable automatic maintenance of the trace slice files\n"
+                  );
+   }
+
+   void plugin_initialize(const appbase::variables_map& options) {
+      if( options.count("minimum-irreversible-trace-history") ) {
+         auto value = options.at("minimum-irreversible-trace-history").as<std::string>();
+         try {
+            minimum_irreversible_trace_history = parse_microseconds(value);
```
heifner: `parse_microseconds` is nice. However, no other existing option supports this capability. I think this needs to be lifted out into the main code base and used for all options that take microseconds (or really any time) or removed for now.

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -0,0 +1,229 @@
...
+         } catch (...) {
+            elog("Malformed minimum-irreversible-trace-history: \"${val}\"!", ("val", value));
```
heifner: I would remove the `"!"`.

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -0,0 +1,229 @@
...
+      if( options.count("trace-rpc-abi") ) {
+         const std::vector<std::string> key_value_pairs = options["trace-rpc-abi"].as<std::vector<std::string>>();
+         for (const auto& entry : key_value_pairs) {
+            try {
+               auto kv = parse_kv_pairs(entry);
+               auto account = chain::name(kv.first);
+               auto abi = abi_def_from_file_or_str(kv.second);
+               data_handler->add_abi(account, abi);
+            } catch (...) {
+               elog("Malformed trace-rpc-abi provider: \"${val}\", ignoring!", ("val", entry));
```
heifner: I don't think this should be ignored. I think log and throw, so they don't run with malformed abi because they didn't notice.

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -0,0 +1,229 @@
...
+                  "   a relative path from `data-dir` to a file containing a valid JSON-encoded ABI\n"
+                  );
+   }
+
+   void plugin_initialize(const appbase::variables_map& options) {
+      data_handler = std::make_shared<abi_data_handler>([](const exception_with_context& e){
+         log_exception(e, fc::log_level::debug);
+      });
+
+      if( options.count("trace-rpc-abi") ) {
```
heifner: If not found, this should log a BIG warning. Or maybe we even have another option that explicitly indicates you are not providing abis and assert if not specified.

PR #8761, opened 2020-03-05
Initial store implementation

heifner:
## Change Description

Created for Brian as he ran out the door.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
@@ -10,34 +10,4 @@ namespace eosio { namespace trace_api_plugin {
       block_trace_v0
    >;
 
-   /**
-    * append an entry to the end of the slice
-    *
-    * @param entry : the entry to append
-    * @param slice : the slice to append entry to
-    * @return the offset in the slice where that entry is written
-    */
-   template<typename DataEntry, typename Slice>
-   uint64_t append_data_log(const DataEntry &entry, Slice &slice) {
-      auto data = fc::raw::pack(entry);
-      const auto offset = slice.tellp();
-      slice.write(data.data(), data.size());
-      slice.flush();
-      slice.sync();
-      return offset;
-   }
-
-   /**
-    * extract an entry from the data log
-    *
-    * @param slice : the slice to extract entry from
-    * @return the extracted data log
-    */
-   template<typename DataEntry, typename Slice>
-   DataEntry extract_data_log( Slice& slice ) {
-      DataEntry entry;
-      auto ds = slice.create_datastream();
-      fc::raw::unpack(ds, entry);
-      return entry;
-   }
 }}
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
@@ -10,4 +10,34 @@ namespace eosio { namespace trace_api_plugin {
       block_trace_v0
    >;
 
+   /**
+    * append an entry to the end of the slice
+    *
+    * @param entry : the entry to append
+    * @param slice : the slice to append entry to
+    * @return the offset in the slice where that entry is written
+    */
+   template<typename DataEntry, typename Slice>
+   uint64_t append_data_log(const DataEntry &entry, Slice &slice) {
+      auto data = fc::raw::pack(entry);
+      const auto offset = slice.tellp();
+      slice.write(data.data(), data.size());
+      slice.flush();
+      slice.sync();
+      return offset;
+   }
+
+   /**
+    * extract an entry from the data log
+    *
+    * @param slice : the slice to extract entry from
+    * @return the extracted data log
+    */
+   template<typename DataEntry, typename Slice>
+   DataEntry extract_data_log( Slice& slice ) {
+      DataEntry entry;
+      auto ds = slice.create_datastream();
+      fc::raw::unpack(ds, entry);
+      return entry;
+   }
 }}
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
@@ -73,14 +73,6 @@ namespace eosio::trace_api_plugin {
 
       const boost::filesystem::path _slice_dir;
       const uint32_t _width;
-
-      static constexpr uint32_t _current_version = 1;
-
-      static constexpr uint _slice_size = 10;
-      static constexpr uint _null_term_size = 1;
-      static constexpr const char* _trace_prefix = "trace_";
-      static constexpr const char* _trace_index_prefix = "trace_index_";
-      static constexpr uint _max_file_size = 12 + 10 + 1 + 10 + 4 + 1; // "trace_index_" + 10-digits + '-' + 10-digits + ".log" + null-char
    };
 
    class store_provider {
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
@@ -94,15 +94,27 @@ namespace eosio::trace_api_plugin {
       bool find_index_slice(uint32_t slice_number, bool append, fc::cfile& index_file) const;
 
       /**
-       * Find or create the index file associated with the indicated slice_number
+       * Find or create the trace file associated with the indicated slice_number
        *
        * @param slice_number : slice number of the requested slice file
        * @param append : indicate if the file is going to be appended (or read)
-       * @param index_file : the cfile that will be set to the appropriate slice filename
+       * @param trace_file : the cfile that will be set to the appropriate slice filename
        *                     and opened to that file
        * @return the true if file was found (i.e. already existed)
        */
       bool find_or_create_trace_slice(uint32_t slice_number, bool append, fc::cfile& trace_file) const;
+
+      /**
+       * Find the trace file associated with the indicated slice_number
+       *
+       * @param slice_number : slice number of the requested slice file
+       * @param append : indicate if the file is going to be appended (or read)
+       * @param trace_file : the cfile that will be set to the appropriate slice filename (always)
+       *                     and opened to that file (if it was found)
+       * @return the true if file was found (i.e. already existed), if not found index_file
+       *         is set to the appropriate file, but not open
+       */
+      bool find_trace_slice(uint32_t slice_number, bool append, fc::cfile& trace_file) const;
 
    private:
       // returns true if slice is found, slice_file will always be set to the appropriate path for
@@ -170,13 +182,13 @@ namespace eosio::trace_api_plugin {
       std::optional<data_log_entry> read_data_log( uint32_t block_height, uint64_t offset ) {
          const uint32_t slice_number = _slice_provider.slice_number(block_height);
          fc::cfile trace;
-         if (_slice_provider.find_or_create_trace_slice(slice_number, false, trace)) {
+         if( !_slice_provider.find_trace_slice(slice_number, false, trace) ) {
             const std::string offset_str = boost::lexical_cast<std::string>(offset);
             const std::string bh_str = boost::lexical_cast<std::string>(block_height);
             throw malformed_slice_file("Requested offset: " + offset_str + " to retrieve block number: " + bh_str + " but this trace file is new, so there are no traces present.");
          }
          const uint64_t end = file_size(trace.get_file_path());
-         if (offset >= end) {
+         if( offset >= end ) {
             const std::string offset_str = boost::lexical_cast<std::string>(offset);
             const std::string bh_str = boost::lexical_cast<std::string>(block_height);
             const std::string end_str = boost::lexical_cast<std::string>(end);
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
@@ -145,10 +145,11 @@ namespace eosio::trace_api_plugin {
          const uint64_t end = file_size(index.get_file_path());
          offset = index.tellp();
          uint64_t last_read_offset = offset;
-         bool cont = true;
-         while (offset < end && cont) {
+         while (offset < end) {
             const auto metadata = extract_store<metadata_log_entry>(index);
-            cont = fn(metadata);
+            if(! fn(metadata)) {
+               break;
+            }
             last_read_offset = offset;
             offset = index.tellp();
          }
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
@@ -5,7 +5,7 @@
 #include <fc/variant.hpp>
 #include <eosio/trace_api_plugin/metadata_log.hpp>
 #include <eosio/trace_api_plugin/data_log.hpp>
-#include <stdio.h>
+#include <ios>
 
 namespace eosio::trace_api_plugin {
    using namespace boost::filesystem;
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
@@ -1,0 +1,186 @@
+#pragma once
+
+#include <fc/io/cfile.hpp>
+#include <boost/filesystem.hpp>
+#include <fc/variant.hpp>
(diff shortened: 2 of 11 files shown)


plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+      const boost::filesystem::path _slice_dir;
+      const uint32_t _width;
+
+      static constexpr uint32_t _current_version = 1;
+
+      static constexpr uint _slice_size = 10;
+      static constexpr uint _null_term_size = 1;
+      static constexpr const char* _trace_prefix = "trace_";
+      static constexpr const char* _trace_index_prefix = "trace_index_";
+      static constexpr uint _max_file_size = 12 + 10 + 1 + 10 + 4 + 1; // "trace_index_" + 10-digits + '-' + 10-digits + ".log" + null-char
```
heifner: Rename to `_max_filename_size`

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+       * @param offset : the offset in the datalog to read
+       * @return empty optional if the data log does not exist, data otherwise
+       * @throws std::exception : when the data is not the correct type or if the log is corrupt in some way
+       *
+       */
+      std::optional<data_log_entry> read_data_log( uint32_t block_height, uint64_t offset ) {
+         const uint32_t slice_number = _slice_provider.slice_number(block_height);
+         fc::cfile trace;
+         if (_slice_provider.find_trace_slice(slice_number, false, trace)) {
+            const std::string offset_str = boost::lexical_cast<std::string>(offset);
```
heifner: Replace all `boost::lexical_cast<std::string>` with `std::to_string`

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
heifner: We are using C++17 so you can do: 
`static constexpr uint _max_filename_size = std::char_traits<char>::length(_trace_index_prefix) + 10 + 1 + 10 + std::char_traits<char>::length(_ext) + 1;`

heifner: Add a constexpr for `_trace_ext = ".log"`

heifner: `_slice_size` and `_null_term_size` are not used.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
```diff
@@ -10,4 +10,34 @@ namespace eosio { namespace trace_api_plugin {
...
+    * @param slice : the slice to extract entry from
+    * @return the extracted data log
+    */
+   template<typename DataEntry, typename Slice>
+   DataEntry extract_data_log( Slice& slice ) {
+      DataEntry entry;
+      auto ds = slice.create_datastream();
+      fc::raw::unpack(ds, entry);
+      return entry;
+   }
```
heifner: I think it would be less confusing to name these template params `File` instead `Slice`. Also the template method `extract_metadata_log`.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+       * @param fn : the functor/lambda
+       * @return the highest offset read during this scan
+       */
+      template<typename Fn>
+      uint64_t scan_metadata_log_from( uint32_t block_height, uint64_t offset, Fn&& fn ) {
+         // ignoring offset
+         offset = 0;
+         fc::cfile index;
+         const uint32_t slice_number = _slice_provider.slice_number(block_height);
+         _slice_provider.find_index_slice(slice_number, false, index);
```
heifner: A find method should never create. Maybe rename to `find_or_create_index_slice`

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+         const uint64_t end = file_size(index.get_file_path());
+         offset = index.tellp();
+         uint64_t last_read_offset = offset;
+         bool cont = true;
+         while (offset < end && cont) {
+            const auto metadata = extract_store<metadata_log_entry>(index);
+            cont = fn(metadata);
+            last_read_offset = offset;
+            offset = index.tellp();
+         }
```
heifner: If this was `if( !fn(metadata) ) break;` you don't need `cont` or `last_read_offset`

plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
```diff
@@ -10,4 +10,34 @@ namespace eosio { namespace trace_api_plugin {
...
 
+   /**
+    * append an entry to the end of the slice
+    *
+    * @param entry : the entry to append
+    * @param slice : the slice to append entry to
+    * @return the offset in the slice where that entry is written
+    */
+   template<typename DataEntry, typename Slice>
+   uint64_t append_data_log(const DataEntry &entry, Slice &slice) {
```
b1bart: This seems like a duplicate of 
```
class store_provider {
  template<typename DataEntry, typename Slice>
  static uint64_t append_store(const DataEntry &entry, Slice &slice);
}
```

Is there a need to have this duplicate and in a plugin-wide header vs something more specific to the Store sub-system?

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+            const std::string offset_str = boost::lexical_cast<std::string>(offset);
+            const std::string bh_str = boost::lexical_cast<std::string>(block_height);
+            const std::string end_str = boost::lexical_cast<std::string>(end);
+            throw malformed_slice_file("Requested offset: " + offset_str + " to retrieve block number: " + bh_str + " but this trace file only goes to offset: " + end_str);
+         }
+         trace.seek(offset);
+         return extract_store<data_log_entry>(trace);
+      }
+
+      /**
```
heifner: I think `append_store` & `extract_store` should be private.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
b1bart: Moreover, the `scan_metadata_log_from` is a reader function and the intent is that the reading process _never_ mutates the filesystem.  I would prefer a hard line in the code where we can reasonably assure ourselves that this will never create a file.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+      {}
+      explicit malformed_slice_file(const std::string& what_arg)
+         :std::runtime_error(what_arg)
+      {}
+   };
+
+   class store_provider;
+
+
+   class slice_provider {
```
b1bart: This seems like an unused abstraction, was the intent to allow `slice_provider` to be swapped for some other implementation at some point?  

Even then it doesn't seem like a clean abstraction to me.  Either
- The `slice_provider` is abstracting *only* writes to/from the filesystem (though it seems like `store_provider` is tightly coupled to) in which case `slice_number()` seems out of place.  The division of data into buckets/slices is owned by the caller
- The `slice_provider` is abstracting the division of data *and* the filesystem in which case the caller should not need to be aware of the `slice_number` it should just push data down and let that next system divide it up and own the edge cases.  

My recommendation is to remove this layer of abstraction and roll it into `store_provider` as it seems 1:1 coupled to that class.

As you already have `vslice` for testing, I could see this becoming the non-mock implementation of a slice that is backed by `fc::cfile` but that would require that `store_provider` be completely abstracted from `fc::cfile`.  That seems like more work at this point.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
b1bart: Many of these things are only used in `store_provider::find_slice` or as a direct input (based on whether this is an index or data log).  Can we move all of them into the implementation of a `private` method like `make_slice_filename` with a `bool` or `enum` indicating the type of log?

I think that increase in locality will help the next engineer when they are reading the code.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
```diff
@@ -0,0 +1,186 @@
...
+       * @throws std::exception : when the data is not the correct type or if the log is corrupt in some way
+       *
+       */
+      std::optional<data_log_entry> read_data_log( uint32_t block_height, uint64_t offset ) {
+         const uint32_t slice_number = _slice_provider.slice_number(block_height);
+         fc::cfile trace;
+         if (_slice_provider.find_trace_slice(slice_number, false, trace)) {
+            const std::string offset_str = boost::lexical_cast<std::string>(offset);
+            const std::string bh_str = boost::lexical_cast<std::string>(block_height);
+            throw malformed_slice_file("Requested offset: " + offset_str + " to retrieve block number: " + bh_str + " but this trace file is new, so there are no traces present.");
```
b1bart: Pending a discussion, we need to decide whether this should return an empty `optional` or `throw`. 

If it remains as a `throw` the message should be that the "trace file does not exist" right?  This reader should not create a new one.

plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -0,0 +1,83 @@
...
+   }
+
+   void store_provider::find_slice_pair(uint32_t block_height, bool append, fc::cfile& trace, fc::cfile& index) {
+      const uint32_t slice_number = _slice_provider.slice_number(block_height);
+      const bool trace_created = _slice_provider.find_trace_slice(slice_number, append, trace);
+      const bool index_created = _slice_provider.find_index_slice(slice_number, append, index);
+      if (trace_created != index_created) {
+         const std::string trace_status = trace_created ? "new" : "existing";
+         const std::string index_status = index_created ? "new" : "existing";
+         elog("Trace file is ${ts}, but it's metadata file is ${is}. This means the files are not consistent.", ("ts", trace_status)("is", index_status));
```
b1bart: More of an integration note for later.   In the extraction and response system we have been feeding in a `std::function<void(const std::exception_ptr&)>` (or if we adopt it a `std::function<void(const exception_with_context&)>`) and letting the plugin code decide what it wants to log and at what level. 

For consistency we will want to do that here as well.  The intent is to not have to worry about custom `fc::logger` s, log levels etc in code this deep

plugins/trace_api_plugin/test/test_trace_file.cpp
```diff
@@ -0,0 +1,405 @@
...
+BOOST_AUTO_TEST_SUITE(slice_tests)
+   BOOST_FIXTURE_TEST_CASE(write_data_trace, test_fixture)
+   {
+      vslice vs;
+      const auto offset = append_data_log( bt, vs );
+      BOOST_REQUIRE_EQUAL(offset,0);
+
+      const auto expected_offset = vs._pos;
+      const auto offset2 = append_data_log( bt2, vs );
+      BOOST_REQUIRE_EQUAL(offset2, expected_offset);
```
b1bart: Is this a necessary invariant to test?  If, for example, in the future we introduce some alignment/padding or internal data in between entries this test will break but I don't think the overall system will be broken.  As long as it returns an offset which can be used to retrieve the exact data (as you test on the lines below)

plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
brianjohnson5972: it can extract anything since it is templated, so metadata would be too restrictive

plugins/trace_api_plugin/test/test_trace_file.cpp
```diff
@@ -0,0 +1,405 @@
...
+      const bool read_file = false;
+      const bool append_file = true;
+      // create trace slices
+      for (uint i = 0; i < 9; ++i) {
+         bool new_file = sp.find_trace_slice(i, append_file, slice);
+         BOOST_REQUIRE(new_file);
+         bfs::path fp = slice.get_file_path();
+         BOOST_REQUIRE_EQUAL(fp.parent_path().generic_string(), tempdir.path().generic_string());
+         const std::string expected_filename = "trace_0000000" + boost::lexical_cast<std::string>(i) + "00-0000000" + boost::lexical_cast<std::string>(i+1) + "00.log";
+         BOOST_REQUIRE_EQUAL(fp.filename().generic_string(), expected_filename);
```
b1bart: I'm on the fence as to whether this is a necessary test.  On the one hand, we expect manual maintenance so, the filesystem structure is an API of sorts.  

On the other hand, this seems like a pedantic test that requires anyone changing that filename to do so in two places.  On top of that it only guards the filename, if someone changes the path structure this test won't catch the breakage of our contract with operators.

Maybe this is something we test for outside of unit tests? where we can actually guarantee the operator-level contract that given a config you will see the following files... ?

plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
heifner: I'm only saying rename `Slice` => `File` for `extract_data_log` & `extract_metadata_log`

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
brianjohnson5972: I moved the variables to an anonymous namespace, does that do it?

b1bart: that helps the visibility.  putting them in the function (while some find it abhorrent) means that you can see the definition and only uses of those variables in the same window/terminal.  I'm not going to press it just clarifying the ask.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
brianjohnson5972: yes on cont, but we need to cache offset before we move it to the end and determine we are at the end of the file.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
brianjohnson5972: it is used in slice_provider and unit test.  I'd like to make this a follow on task if you really think that a comment isn't enough to allow someone to maintain this.

plugins/trace_api_plugin/include/eosio/trace_api_plugin/data_log.hpp
brianjohnson5972: forgot to cleanup, it is gone now

plugins/trace_api_plugin/include/eosio/trace_api_plugin/store_provider.hpp
brianjohnson5972: This is really for testing, if we pull it into store_provider, my preference would still be to have it as a inner class, but if these are private I'll need to add a test_friend class to allow testing.  I think putting a comment for slice_provider indicating that it is only intended to be used by store_provider and unit tests.  I will admit that having them both use *_provider will lead to confusion.  Would prefer to make that my next task (after this gets merged).

plugins/trace_api_plugin/store_provider.cpp
brianjohnson5972: This sounds like a plugin level design that we need, so treating this like a next step (i.e. after PR is merged)

plugins/trace_api_plugin/test/test_trace_file.cpp
brianjohnson5972: not necessary, mostly providing training wheels at the outset, can remove now

plugins/trace_api_plugin/test/test_trace_file.cpp
brianjohnson5972: I don't think useful tests need to stand the test of time and live on forever.  If someone is changing the format, they can decide if this should be removed.  If they don't think about anything, then it is good that we have this here.

brianjohnson5972: and having this here, makes that higher level test easier, you just need to make sure that files are read from the correct spot

PR #8770, opened 2020-03-06
Trace api extraction - onerror test

heifner:
## Change Description

- Do not report context free actions.
- Only report executed transactions.
  - soft_fail `onerror` transaction reported with the id of the input transaction since that is the id the user knows about.
- Remove `transaction_trace_v0` status since they are all `executed` except for the `soft_fail` which would be confusing to users.
 
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/test/test_extraction.cpp
+++ b/plugins/trace_api_plugin/test/test_extraction.cpp
@@ -1,6 +1,8 @@
 #define BOOST_TEST_MODULE trace_data_extraction
 #include <boost/test/included/unit_test.hpp>
 
+#include <eosio/chain/types.hpp>
+#include <eosio/chain/contract_types.hpp>
 #include <eosio/chain/trace.hpp>
 #include <eosio/chain/transaction.hpp>
 #include <eosio/chain/block.hpp>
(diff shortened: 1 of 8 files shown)


plugins/trace_api_plugin/test/test_extraction.cpp
```diff
@@ -354,4 +373,54 @@ BOOST_AUTO_TEST_SUITE(block_extraction)
...
+            { chain::packed_transaction(transfer_trx) } );
+      signal_accepted_block( bsp1 );
+
+      const uint64_t expected_lib = 0;
+      const block_entry_v0 expected_entry { bsp1->id, 1, 0 };
+
+      const block_trace_v0 expected_trace { bsp1->id, 1, bsp1->prev(), chain::block_timestamp_type(1), "bp.one"_n,
+         {
+            {
+               ptrx1.id(),
```
b1bart: I'm missing something I think.  `ptrx.id()` is the id of a "virtual" packed transaction that has the `onerror` action right?  `transfer_trx.id()` would be the ID of the transaction that "failed".   So, I would have expected this to produce a trace with that ID and a status of `soft_failed`

PR #8774, opened 2020-03-06
Change action data handling and output

b1bart:
In this PR:
- [ ] `data` is now always the hex encoded payload of the action
- [ ] `params` is a new optional field that represents the translated version of `data` (eg via ABIs)  if one is successfully created.   This field is not present if there is no such translation.

### DIFF
--- a/plugins/trace_api_plugin/request_handler.cpp
+++ b/plugins/trace_api_plugin/request_handler.cpp
@@ -51,13 +51,19 @@ namespace {
       for ( const auto& a: actions) {
          yield();
 
-         result.emplace_back(fc::mutable_variant_object()
-            ("receiver", a.receiver.to_string())
-            ("account", a.account.to_string())
-            ("action", a.action.to_string())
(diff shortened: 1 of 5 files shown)


plugins/trace_api_plugin/request_handler.cpp
```diff
@@ -51,13 +51,19 @@ namespace {
...
+         auto action_variant = fc::mutable_variant_object()
+               ("receiver", a.receiver.to_string())
+               ("account", a.account.to_string())
+               ("action", a.action.to_string())
+               ("authorization", process_authorizations(a.authorization, std::forward<Yield>(yield)))
+               ("data", fc::to_hex(a.data.data(), a.data.size()));
+
+         auto params = data_handler(a);
+         if (!params.is_null()) {
+            action_variant("params", params);
```
heifner: I think this should be `parameters` ;)

nksanthosh: :-)

PR #8778, opened 2020-03-08
Trace API plugin - yield pattern

b1bart:
In this PR:
- [ ] Move away from deadlines and concepts of time to a "yield pattern"
  - a `yield_function` takes no parameters, returns no values, and may sometimes throw a `yield_exception` which should bubble out of a long-running task to the top-level caller once the task gracefully terminates itself.  In most cases, this can just throw all the way out.  

### Why?

There are many reasons why the caller of a potentially long running task may have set a deadline before.  Some of them did not need to be destructive, such as processing a "critical" priority task queue with guaranteed maximum latency when the queue rarely has a task.   In this case, a yield function can intelligently check for the presence of a critical task before terminating the cooperative task instead of always terminating it so that the caller can potentially service the queue. 

Additionally, even in use-cases where time is the only consideration this pattern allows the caller to define what the source of that time is.  This is useful for unit tests when having a hard-coded dependency on the system clock can cause non-determinism or difficulty constructing tests.

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api_plugin/common.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api_plugin/common.hpp
@@ -3,19 +3,37 @@
 #include <tuple>
 
 namespace eosio::trace_api_plugin {
+   template<typename R, typename ...Args>
+   class optional_delegate : private std::function<R(Args...)> {
+   public:
+      using std::function<R(Args...)>::function;
+
+      /**
+       * overloaded call operator to ignore unset functions
+       */
+      template<typename U = R>
+      auto operator()( Args... args ) const -> std::enable_if_t<!std::is_void_v<U>, R> {
+         if (static_cast<bool>(*this)) {
+            return std::function<R(Args...)>::operator()(std::move(args)...);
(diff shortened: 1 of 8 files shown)


plugins/trace_api_plugin/include/eosio/trace_api_plugin/common.hpp
```diff
@@ -4,18 +4,30 @@
...
+    * calling code should expect that this can throw yield_exception and gracefully unwind if it does
+    * @throws yield_exception if the provided yield needs to terminate the long running process for any reason
+    */
+   using yield_function = std::function<void()>;
+
+   template<typename F, typename ...Args>
+   void call_if_set(F&& f, Args&&  ...args ) {
+      if (static_cast<bool>(f))
+         std::forward<F>(f)(std::forward<Args>(args)...);
+   }
```
heifner: I think `yeild_function` should be a type that supports call if set instead of making the `yeild` invoker responsible. Something like `NoExceptDelegate`.
 
https://stackoverflow.com/questions/31602903/enforce-noexcept-on-stdfunction

PR #8788, opened 2020-03-09
dfuse Deep Mind changes

maoueh:
## Change Description
Discussed last week, here the dfuse Deep Mind patchset rebased on develop. 

It contains low-level dfuse instrumentation of the nodeos binary for post-consumption through a console reader.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


maoueh:
I'll rebase all this after the first wave of comments.

abourget:
Things I'm thinking we might want to look:

* the `action_id`, and its alignment (or mis-alignment) with the action ordinals.

libraries/chain/apply_context.cpp
```diff
@@ -172,9 +176,11 @@ void apply_context::exec()
 {
    _notified.emplace_back( receiver, action_ordinal );
    exec_one();
+   trx_context.action_id.increment();
```
gleehokie: Why is `action_id` being incremented?  Is this directly related to Deep Mind?  If not should it be pulled out into a separate PR?

libraries/chain/apply_context.cpp
```diff
@@ -172,9 +176,11 @@ void apply_context::exec()
...
    for( uint32_t i = 1; i < _notified.size(); ++i ) {
       std::tie( receiver, action_ordinal ) = _notified[i];
       exec_one();
+      trx_context.action_id.increment();
```
gleehokie: Same comment.  Why is `action_id` being incremented? Is this directly related to Deep Mind? If not should it be pulled out into a separate PR?

libraries/chain/apply_context.cpp
```diff
@@ -237,6 +243,12 @@ void apply_context::require_recipient( account_name recipient ) {
          recipient,
          schedule_action( action_ordinal, recipient, false )
       );
+
+      if (eosio::chain::chain_config::deep_mind_enabled) {
```
gleehokie: I notice that each time `dmlog` is called it is prefaced with this check.  Wouldn't it be clearer to move this check into `dmlog` itself?

libraries/chain/apply_context.cpp
```diff
@@ -471,21 +495,37 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
       EOS_ASSERT( replace_existing, deferred_tx_duplicate, "deferred transaction with the same sender_id and payer already exists" );
 
       bool replace_deferred_activated = control.is_builtin_activated(builtin_protocol_feature_t::replace_deferred);
 
-      EOS_ASSERT( replace_deferred_activated || !control.is_producing_block()
-                     || control.all_subjective_mitigations_disabled(),
-                  subjective_block_production_exception,
-                  "Replacing a deferred transaction is temporarily disabled." );
+      // For the sake of testing Deep Mind, we would still like to replace a deferred
+      // transaction so we can later test that ram correction is correclty handled through
```
gleehokie: Spelling correction: s/correclty/correctly.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -604,6 +610,10 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
          throw;
       }
 
+      eosio::chain::chain_config::deep_mind_enabled = options.at( "deep-mind" ).as<bool>();
```
arhag: Instead of this I suggest the following:
- Create an `fc::logger _deep_mind_log` in `chain_plugin.cpp` in a similar way that `_trx_trace_log` was created in `producer_plugin.cpp` (including adding a `handle_sighup` to `chain_plugin`). 
- Add a `fc::logger* deep_mind_log = nullptr;` member variable to `controller_impl`. 
- Add an inline member function `fc::logger* controller_impl::get_deep_mind_logger()const` which simply returns `deep_mind_log`.
- Add a member function `fc::logger* controller::get_deep_mind_logger()const` which simply returns `my->get_deep_mind_logger()`.
- Add a member function `void controller::enable_deep_mind( fc::logger* log )` which sets the `my->deep_mind_log` pointer to the provided value if it is currently `nullptr` (you could assert if wasn't `nullptr`).
- Finally, within `chain_plugin::plugin_initialize`, if and only if `options.at( "deep-mind" ).as<bool>()` is true, call `my->chain->enable_deep_mind( &_deep_mind_log )`. 

Then everywhere you current do:
```cpp
if (eosio::chain::chain_config::deep_mind_enabled) {
```
could be replaced with:
```cpp
if (auto dm_log = get_deep_mind_logger()) {
```
or:
```cpp
if (auto dm_log = control.get_deep_mind_logger()) {
```
or something similar depending on where it is at (`controller`, `apply_context`, `transaction_context`, etc.). You will probably need to add a reference to `controller` into `resource_limits_manager` though. Also instead of including the one log within `protocol_feature_manager.cpp`, I would instead modify `protocol_feature_manager::activate_feature` to return `protocol_feature_manager::const_iterator` pointing to the newly activated protocol feature and allow the caller in `controller_impl::start_block` to do the logging.

Then within the if statement you can log by doing something like the following:
```cpp
   fc_dlog(*dm_log, "message", ...);
```

libraries/chain/wasm_interface.cpp
```diff
@@ -1144,7 +1144,7 @@ class console_api : public context_aware_api {
    public:
       console_api( apply_context& ctx )
       : context_aware_api(ctx,true)
-      , ignore(!ctx.control.contracts_console()) {}
+      , ignore(!eosio::chain::chain_config::deep_mind_console_enabled) {}
```
arhag: Revert back to `ignore(!ctx.control.contracts_console())`.

libraries/chain/apply_context.cpp
```diff
@@ -17,6 +17,10 @@ using boost::container::flat_set;
 namespace eosio { namespace chain {
 
 static inline void print_debug(account_name receiver, const action_trace& ar) {
+   if (eosio::chain::chain_config::deep_mind_console_enabled) {
+      return;
```
arhag: Remove. This is unnecessary because if `contracts-console` is disabled, then `ar.console.empty()` will always be true and this function will do nothing.

libraries/chain/apply_context.cpp
abourget: Yes it's related.. it's how actions are stitched together, the hierarchy is created.  It was there before the action ordinals were introduced.  Without action IDs, the stream of data cannot be reassembled.

maoueh: It's something we would prefer to keep as-is for now as it requires upstream changes and all our data models if removed.

We are open to switch to the `action_ordinal` in a future revision of the deep mind patch, it will require some planning on our side to refactor our data model and related code paths.

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -13,6 +14,10 @@ namespace eosio { namespace chain {
  * values specified by the producers.
  */
 struct chain_config {
+   static bool deep_mind_enabled;
+   static bool deep_mind_console_enabled;
+   static bool deep_mind_subjective_mitigations_disabled;
```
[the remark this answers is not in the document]
maoueh: > deep_mind_console_enabled

We will move to using it instead, now that `contracts_console` does not print by default anymore in `nodeos` log output.

> deep_mind_subjective_mitigations_disabled

We will remove that. Like discussed privately, a compile flag option that enabled a config option `subjective-mitigations-disabled` will be introduced.

I would follow pattern used for `EOSIO_EOS_VM_OC_DEVELOPER` and CMake option `EOSVMOC_ENABLE_DEVELOPER_OPTIONS`, is that correct?

What about:

```
EOSIO_ENABLE_DEVELOPER_OPTIONS (CMake)
EOSIO_DEVELOPER (Define)
```

> deep_mind_enabled

I'll follow instructions on comment https://github.com/EOSIO/eos/pull/8788#discussion_r390044578

libraries/chain/apply_context.cpp
maoueh: Fixed thanks.

libraries/chain/include/eosio/chain/chain_config.hpp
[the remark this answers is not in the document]
arhag: The
```
EOSIO_ENABLE_DEVELOPER_OPTIONS (CMake)
EOSIO_DEVELOPER (Define)
```
pattern sounds good.

plugins/chain_plugin/chain_plugin.cpp
maoueh: All applied. 

I decided on not implementing the suggested `activate_feature` change. The `activate_feature` is used on other code paths than just `start_block`, so it seemed more safe to stay in there so the log print is output for all paths.

programs/nodeos/main.cpp
```diff
@@ -60,9 +63,31 @@ void logging_conf_handler()
 
 void initialize_logging()
 {
+   auto success = fc::log_config::register_appender<dfuse::dm::log_appender>( "deep-mind" );
```
maoueh: @arhag Suggestions here about registering the appender, and the default config changeset below?

In the first version(s) of our patch, we had a forked `fc` sub-module, and we hijacked the default config to register the deep mind appender + logger.

We later move to a new library inside `eos` codebase directly, since it was much easier to work with (and to improve build performance). But at that point, we needed to also move registration of the appender + default logger config in the `nodeos` binary directly.

It never felt "good" to my and I'm wondering if there is something else that could be done here, or a better location where I could put that.

plugins/chain_plugin/chain_plugin.cpp
maoueh: Also, instead of using`fc_dlog(*dm_log, "message", ...);`, I adjusted the `dmlog` macro to use the logger and write directly, making it as lean as possible, and more readable (IMO).

maoueh:
This is ready for a new round of review.

libraries/dm/include/dfuse/dm/log_appender.hpp
```diff
@@ -0,0 +1,25 @@
...
+
+   class log_appender : public fc::appender
+   {
+       public:
+            log_appender( const fc::variant& args );
+            log_appender();
+
+            virtual ~log_appender();
+            void initialize( boost::asio::io_service& io_service ) {
+                 setbuf(stdout, NULL);
```
heifner: I don't think you should hijack `stdout` like this. Can you use a file instead?

libraries/dm/log_appender.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <unistd.h>
+#endif
+#include <boost/thread/mutex.hpp>
+#include <fc/exception/exception.hpp>
+#include <iomanip>
+#include <mutex>
+#include <sstream>
+
+namespace dfuse { namespace dm {
+   class log_appender::impl {
```
heifner: I think this could be moved into `fc` as a type of registered appender like `console` and `gelf`. Then you can configure it via a `logging.json`. Call it say `dmlog` type since the `DMLOG` is hard coded.

We hope to move to a more full featured logging framework in the future so this hopefully would only be temporary.

programs/nodeos/main.cpp
heifner: I believe this should be registered via a `logging.json`. Add a new appender type to `fc` so that is possible. That new appender can be a new type of console appender (called say `dmlog`) or even a file appender without too much work.

programs/nodeos/main.cpp
```diff
@@ -60,9 +63,31 @@ void logging_conf_handler()
...
+
+      fc::logger_config dmlc;
+      dmlc.name = "deep-mind";
+      dmlc.level = log_level::debug;
+      dmlc.enabled = true;
+      dmlc.appenders.push_back("deep-mind");
+
+      cfg.loggers.push_back( dmlc );
+
+      fc::configure_logging(cfg);
```
heifner: If configured by a `logging.json` then this `else` is not needed.

libraries/dm/include/dfuse/dm/log_appender.hpp
maoueh: While I agree the `stdout` is not always a good thing to hijack, in this specific case, someone is expected to have `deep-mind = true` which is opt-in.

All the underlying reading tools are currently built around reading `stdout`. The `stdout` was a good choice because by default, `nodeos` outputs to `stderr`. So it was making sense for us.

Using a file based is definitely not something we want to use, we don't want to deal with its constraints (need to truncate it, disk access, disk full, etc).

I know there was some discussion about using a `fifo` instead, which would be a good compromise. But it seemed to be ok for now to still push through `stdout` for the initial version.

libraries/dm/log_appender.cpp
maoueh: It was in `fc` before, but it was harder for us to maintained, since we had two private forked submodules, which is not easy to deal with.

So we moved it back directly in EOSIO. I don't have strong disagreement here, I'm willing to move it back to `fc` again.

`dmlog_appender` would be correct with you?

libraries/dm/include/dfuse/dm/log_appender.hpp
heifner: Can you just add a `fflush` call to `out`?

programs/nodeos/main.cpp
maoueh: Answering here for https://github.com/EOSIO/eos/pull/8788#discussion_r391109310 also (was unable to comment there for unknown reason).

Ideally, we would have all the settings in the codebase directly, and only `deep-mind = true` would be needed to drive the dmlog, without further manually intervention.

Operation wise, it's easier for user to only activate a single config, without the need to configure a specialized `logging.json` that would enable the logger. 

If you agree, I would enable dmlog "deep-mind" as a valid logger in `fc` submodule that would work out of the box in nodeos, without the need for `logging.json`.

programs/nodeos/main.cpp
maoueh: Moving the logger to `fc` will resolve all this, all this code would go away.

libraries/dm/log_appender.cpp
heifner: yes.

libraries/chain/include/eosio/chain/protocol_feature_manager.hpp
```diff
@@ -245,7 +247,7 @@ class protocol_feature_set {
 class protocol_feature_manager {
 public:
 
-   protocol_feature_manager( protocol_feature_set&& pfs );
+   protocol_feature_manager( controller& c, protocol_feature_set&& pfs );
```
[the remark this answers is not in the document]
maoueh: Sure, will do

unittests/resource_limits_test.cpp
```diff
@@ -4,19 +4,21 @@
...
-class resource_limits_fixture: private chainbase_fixture<1024*1024>, public resource_limits_manager
+class resource_limits_fixture: private tester, private chainbase_fixture<1024*1024>, public resource_limits_manager
 {
    public:
       resource_limits_fixture()
-      :chainbase_fixture()
-      ,resource_limits_manager(*chainbase_fixture::_db)
+      :tester()
+      ,chainbase_fixture()
+      ,resource_limits_manager(*tester::control, *chainbase_fixture::_db)
```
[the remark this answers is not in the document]
maoueh: Yeah no problem.

libraries/chain/apply_context.cpp
```diff
@@ -495,8 +521,27 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
+         dmlog(logger, "DTRX_OP MODIFY_CANCEL ${action_id} ${sender} ${sender_id} ${payer} ${published} ${delay} ${expiration} ${trx_id} ${trx}",
+            ("action_id", trx_context.action_id.current())
+            ("sender", receiver)
+            ("sender_id", sender_id)
+            ("payer", ptr->payer)
+            ("published", ptr->published)
+            ("delay", ptr->delay_until)
+            ("expiration", ptr->expiration)
+            ("trx_id", dtrx.id())
+            ("trx", control.to_variant_with_abi(dtrx, fc::microseconds(5000000)))
```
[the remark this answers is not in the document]
maoueh: Yeah make sense

unittests/resource_limits_test.cpp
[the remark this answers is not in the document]
maoueh: Since the logger instance was later in the lifecycle via the `enable_deep_mind` call, I had to pass to a lambda that knows how to retrieve the deepmind logger instance.

I'll push what I have so far, missing piece will be the RAM event context object.

maoueh:
This is ready for another review round.

libraries/dm/include/dfuse/dm/log_appender.hpp
maoueh: Discussion now happening on https://github.com/EOSIO/fc/pull/148

programs/nodeos/main.cpp
maoueh: Discussion now happening on https://github.com/EOSIO/fc/pull/148

maoueh:
Pushed with latest changes @arhag @heifner

maoueh:
Updated to now merged deep mind branch on `fc`.

libraries/chain/controller.cpp
```diff
@@ -3231,23 +3292,70 @@ const flat_set<account_name> &controller::get_resource_greylist() const {
...
+   // Changing the standard `stdout` behavior from buffered to unbuffered can is disruptive
+   // and can lead to weird scenarios in the logging process if `stdout` is used there too.
+   //
+   // In a future version, the `fc::dmlog_appender` implementation will switch to a `FIFO` file
+   // approach, which will remove the dependency on `stdout` and hence this call.
+   //
+   // For the time being, when `deep-mind = true` is activated, we set `stdout` here to
+   // be an unbuffered I/O stream.
+   setbuf(stdout, NULL);
+}
```
heifner: Move this to `dmlog.hpp` as a standalone function.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -26,6 +26,9 @@
 #include <deque>
 #include <cstdint>
 
+#define RAM_EVENT_ID( FORMAT, ... ) \
+   fc::format_string( FORMAT, fc::mutable_variant_object()__VA_ARGS__ )
+
```
heifner: This can be moved to `dmlog.hpp`

programs/nodeos/main.cpp
```diff
@@ -63,6 +64,23 @@ void initialize_logging()
...
+      fc::logger_config dmlc;
+      dmlc.name = "deep-mind";
+      dmlc.level = fc::log_level::debug;
+      dmlc.enabled = true;
+      dmlc.appenders.push_back("deep-mind");
+
+      cfg.loggers.push_back( dmlc );
+
+      fc::configure_logging(cfg);
+   }
```
heifner: Could be a function in `dmlog.hpp` that does this to keep all the `"deep-mind"` config together.

libraries/chain/apply_context.cpp
```diff
@@ -250,6 +252,12 @@ void apply_context::require_recipient( account_name recipient ) {
          recipient,
          schedule_action( action_ordinal, recipient, false )
       );
+
+      if (auto dm_logger = control.get_deep_mind_logger()) {
+         fc_dlog(*dm_logger, "CREATION_OP NOTIFY ${action_id}",
```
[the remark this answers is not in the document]
heifner: @arhag Your thoughts on this?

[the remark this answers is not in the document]
arhag: This is what they had before. I suggested to move to the pointer version so that each controller can be configured to not due logging (more modular) and also so that it is more efficient (no lookup by string in the critical path).

libraries/chain/controller.cpp
maoueh: The whole `enable_...` function or just setting `stdout` unbuffered?

programs/nodeos/main.cpp
maoueh: Will do

libraries/chain/include/eosio/chain/types.hpp
maoueh: Will do

libraries/chain/controller.cpp
heifner: Well, now that controller still has a logger pointer, just the large comment and `setbuf`

maoueh: Added `set_dmlog_appender_stdout_unbuffered` in a `deep_mind.hpp` file.

libraries/chain/include/eosio/chain/types.hpp
maoueh: Moved to `deep_mind.hpp`

programs/nodeos/main.cpp
maoueh: Added `add_deep_mind_logger` to `deep_mind.hpp` to take care of adding deep mind logger to default config.

maoueh:
@heifner @arhag New comments addressed.

maoueh:
Anything left to be done/changed here. Would like to get that out in the wild.

libraries/chain/CMakeLists.txt
```diff
@@ -41,6 +41,7 @@ if("eos-vm-oc" IN_LIST EOSIO_WASM_RUNTIMES)
    include_directories(${LLVM_INCLUDE_DIRS})
    add_definitions(${LLVM_DEFINITIONS})
 
+   option(EOSIO_ENABLE_DEVELOPER_OPTIONS "enable developer options for EOSIO" OFF)
```
maoueh: Oups, this is at wrong location, will need to be moved.

libraries/chain/resource_limits.cpp
```diff
@@ -200,7 +238,22 @@ void resource_limits_manager::add_pending_ram_usage( const account_name account,
               "Ram usage delta would underflow UINT64_MAX");
 
    _db.modify( usage, [&]( auto& u ) {
-     u.ram_usage += ram_delta;
+      u.ram_usage += ram_delta;
+
+      if (auto dm_logger = _get_deep_mind_logger()) {
+         EOS_ASSERT(!ram_trace.is_generic(), misc_exception, "deep mind does not accept generic ram event");
```
maoueh: Left-over, will remove.

unittests/resource_limits_test.cpp
```diff
@@ -4,6 +4,7 @@
 #include <eosio/chain/resource_limits.hpp>
 #include <eosio/chain/config.hpp>
 #include <eosio/testing/chainbase_fixture.hpp>
+#include <eosio/testing/tester.hpp>
```
maoueh: Not required AFAIK, will remove.

libraries/chain/controller.cpp
```diff
@@ -1321,6 +1341,12 @@ struct controller_impl {
          trace->except = e;
          trace->except_ptr = std::current_exception();
          trace->elapsed = fc::time_point::now() - trx_context.start;
+
+         if (auto dm_logger = get_deep_mind_logger()) {
+            fc_dlog(*dm_logger, "DTRX_OP FAILED ${action_id}",
+               ("action_id", trx_context.action_id.current())
```
maoueh: Gonna use `trx_context.get_action_id()`.

libraries/chain/include/eosio/chain/transaction_context.hpp
```diff
@@ -83,12 +99,14 @@ namespace eosio { namespace chain {
...
 
+         uint32_t get_action_id() const { return action_id.current(); }
+
       private:
 
          friend struct controller_impl;
          friend class apply_context;
 
-         void add_ram_usage( account_name account, int64_t ram_delta );
+         void add_ram_usage( account_name account, int64_t ram_delta, const ram_trace&& ram_trace );
```
arhag: Should be `const ram_trace&`. And again consider renaming to avoid same name for type and argument.

libraries/chain/transaction_context.cpp
```diff
@@ -467,9 +467,9 @@ namespace eosio { namespace chain {
       }
    }
 
-   void transaction_context::add_ram_usage( account_name account, int64_t ram_delta ) {
+   void transaction_context::add_ram_usage( account_name account, int64_t ram_delta, const ram_trace&& ram_trace ) {
       auto& rl = control.get_mutable_resource_limits_manager();
-      rl.add_pending_ram_usage( account, ram_delta );
+      rl.add_pending_ram_usage( account, ram_delta, std::move(ram_trace) );
```
arhag: Do not `move`.

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -553,6 +585,9 @@ class apply_context {
 
       action_name get_sender() const;
 
+      uint32_t get_action_id() const;
+      void increment_action_id();
```
arhag: Put the implementation of these two simple functions here so they can be inlined.

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -543,7 +575,7 @@ class apply_context {
       uint64_t next_recv_sequence( const account_metadata_object& receiver_account );
       uint64_t next_auth_sequence( account_name actor );
 
-      void add_ram_usage( account_name account, int64_t ram_delta );
+      void add_ram_usage( account_name account, int64_t ram_delta, const ram_trace&& ram_trace );
```
arhag: Once again `const ram_trace&` both here and in `update_db_usage` above.

libraries/chain/apply_context.cpp
```diff
@@ -886,8 +1098,8 @@ uint64_t apply_context::next_auth_sequence( account_name actor ) {
    return amo.auth_sequence;
 }
 
-void apply_context::add_ram_usage( account_name account, int64_t ram_delta ) {
-   trx_context.add_ram_usage( account, ram_delta );
+void apply_context::add_ram_usage( account_name account, int64_t ram_delta, const ram_trace&& ram_trace ) {
+   trx_context.add_ram_usage( account, ram_delta, std::move(ram_trace) );
```
arhag: Do not `move`.

libraries/chain/apply_context.cpp
```diff
@@ -624,7 +767,7 @@ void apply_context::update_db_usage( const account_name& payer, int64_t delta )
          require_authorization( payer );
       }
    }
-   add_ram_usage(payer, delta);
+   add_ram_usage(payer, delta, std::move(ram_trace));
```
arhag: Do not `move`.

libraries/chain/apply_context.cpp
```diff
@@ -484,6 +504,8 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
    }
 
    uint32_t trx_size = 0;
+   fc::string event_id;
```
arhag: Just use `std::string` everywhere instead of `fc::string`.

libraries/chain/apply_context.cpp
```diff
@@ -508,8 +534,27 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
          trx_id_for_new_obj = ptr->trx_id;
       }
 
+      if (auto dm_logger = control.get_deep_mind_logger()) {
+         fc::datastream<const char*> ds( ptr->packed_trx.data(), ptr->packed_trx.size() );
+         transaction dtrx;
+         fc::raw::unpack(ds, static_cast<transaction&>(dtrx) );
```
arhag: The `static_cast` is redundant.

libraries/chain/apply_context.cpp
```diff
@@ -541,14 +620,39 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
+
+
    auto& generated_transaction_idx = db.get_mutable_index<generated_transaction_multi_index>();
    const auto* gto = db.find<generated_transaction_object,by_sender_id>(boost::make_tuple(sender, sender_id));
    if ( gto ) {
-      add_ram_usage( gto->payer, -(config::billable_size_v<generated_transaction_object> + gto->packed_trx.size()) );
+      fc::string event_id;
+      if (auto dm_logger = control.get_deep_mind_logger()) {
+         // unpack gtx->packed_trx into `dtrx` REVISE THIS!
+         auto gtx = generated_transaction(*gto);
```
arhag: Why are you copying from `gto` rather than just using `gto` directly? All usage of `gto` for logging should be done prior to it being removed.

libraries/chain/apply_context.cpp
```diff
@@ -541,14 +620,39 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
...
+         fc::datastream<const char*> ds( gtx.packed_trx.data(), gtx.packed_trx.size() );
+         transaction dtrx;
+         fc::raw::unpack(ds, static_cast<transaction&>(dtrx) );
```
arhag: Again redundant `static_cast`.

maoueh:
@arhag Yeah indeed, I'm used to rebase my own branches, never realized it breaks this review flow. Will follow that as well as the `Resolve Conversation` comment. I'm sorry to give you more work.

Those are good comments.

arhag:
@maoueh: No worries. Also, I forgot to mention: make sure to bring in the latest changes from the `develop` branch and resolve the existing merge conflicts.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 #include <eosio/chain/block_state.hpp>
+#include <eosio/chain/deep_mind.hpp>
```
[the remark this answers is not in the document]
arhag: Or actually, following the suggestion in [this comment](https://github.com/EOSIO/eos/pull/8788#discussion_r398953350), `deep_mind.hpp` and `deep_mind.cpp` files can be entirely removed.

libraries/chain/resource_limits.cpp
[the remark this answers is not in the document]
maoueh: Yeah, the idea was to be notified fast that some generic ram trace were used and that we need to update those.

I removed the assertion and I will still log the ram op. But instead, I will change the reading code to fail if it detects a RAM_OP of an invalid type.

libraries/chain/include/eosio/chain/deep_mind.hpp
```diff
@@ -0,0 +1,15 @@
+#pragma once
+#include <fc/string.hpp>
+#include <fc/variant_object.hpp>
+#include <fc/log/logger_config.hpp>
+
+#define RAM_EVENT_ID( FORMAT, ... ) \
```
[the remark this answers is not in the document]
maoueh: The `deep_mind.{hpp,cpp}` files have been removed.

libraries/chain/trace.cpp
```diff
@@ -34,4 +35,29 @@ action_trace::action_trace(
...
+   const char* operation,
+   const char* legacy_tag
+)
+:action_id( action_id )
+,event_id( event_id )
+,family( family )
+,operation( operation )
+,legacy_tag( legacy_tag )
+{
+   EOS_ASSERT(!is_generic(), misc_exception, "ram trace from constructor cannot be generic");
```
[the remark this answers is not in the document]
maoueh: The idea was indeed to prevent programmer error that would see a `ram_trace` param and would try to construct an empty `ram_trace`, with default values like `ram_trace(0, nullptr, nullptr, nullptr, nullptr)`. That being said, there is probably a bunch of other possibilities someone would use here like empty strings, and the assertions could not cover all cases.

The idea behind `generic_ram_trace` was to improve readability, usage and searchability (for me). Like if you don't know how to construct a full `ram_trace`, use a `generic_ram_trace` instead.

That's what I kept for now, but I could switch back to a single `ram_trace` struct.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -65,6 +65,35 @@ namespace eosio { namespace chain {
       std::exception_ptr                         except_ptr;
    };
 
+   struct ram_trace {
```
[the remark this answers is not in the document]
maoueh: I removed `is_generic` but kept `generic_ram_trace` for now (see other comment about that).

I thought that `generic_ram_trace` would be easier for you guys to use and I could take care about the `action_id` and the rest, specially if not available where calls for RAM I made, I would make those changes at the same time.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -294,9 +296,12 @@ namespace eosio { namespace chain {
          void set_greylist_limit( uint32_t limit );
          uint32_t get_greylist_limit()const;
 
-         void add_to_ram_correction( account_name account, uint64_t ram_bytes );
+         void add_to_ram_correction( account_name account, uint64_t ram_bytes, uint32_t action_id, const char* event_id );
```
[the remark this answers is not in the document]
maoueh: Since ram correction are generating a different OP (`RAM_CORRECTION_OP`) than normal ram operation (`RAM_OP`), most of the fields would not be used here. Hence why I decided to not use a `ram_trace` for this one.

libraries/chain/include/eosio/chain/apply_context.hpp
maoueh: The `trx_context` type is incomplete in the header, and I would need to also forward declare the `action_id` class to fix it.

I think it's not worth it, what do you think?

libraries/chain/apply_context.cpp
maoueh: Not sure why we had this part, I removed it, did not see any problems from our test suite.

libraries/chain/authorization_manager.cpp
```diff
@@ -187,28 +196,55 @@ namespace eosio { namespace chain {
...
    }
 
-   void authorization_manager::modify_permission( const permission_object& permission, const authority& auth ) {
+   void authorization_manager::modify_permission( const permission_object& permission, const authority& auth, uint32_t action_id ) {
       for(const key_weight& k: auth.keys)
          EOS_ASSERT(k.key.which() < _db.get<protocol_state_object>().num_supported_key_types, unactivated_key_type,
            "Unactivated key type used when modifying permission");
 
       _db.modify( permission, [&](permission_object& po) {
+         const permission_object old_permission(po);
```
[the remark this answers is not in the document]
maoueh: Changed, but I did not move, don't think it's worth it.

libraries/chain/include/eosio/chain/transaction_context.hpp
maoueh: I've changed all occurrences to `const ram_trace& trace`. Not that much better but are least, there are different.

maoueh:
@arhag Updated PR against your latest comments.

libraries/chain/include/eosio/chain/apply_context.hpp
arhag: Ah right. Fair point. Then it is fine as is.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1077,6 +1120,13 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
             } );
 
       my->accepted_block_connection = my->chain->accepted_block.connect( [this]( const block_state_ptr& blk ) {
+         if (auto dm_logger = my->chain->get_deep_mind_logger()) {
+            fc_dlog(*dm_logger, "ACCEPTED_BLOCK ${num} ${blk}",
+               ("num", blk->block_num)
+               ("blk", chain().to_variant_with_abi(blk, abi_serializer::create_yield_function(fc::microseconds(config::dmlog_abi_serializer_max_time_us))))
```
maoueh: ABI serializer is not needed here, the block state does not contain any ABI serializable data, so I will remove that.

libraries/chain/include/eosio/chain/trace.hpp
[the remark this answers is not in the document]
arhag: Here is my preferred way of handling it:
```cpp
struct ram_trace {
   uint32_t     action_id  = 0;
   const char*  event_id   = "generic";
   const char*  family     = "generic";
   const char*  operation  = "generic";
   const char*  legacy_tag = "generic";
};
```

And that is all that is needed.

Then it can be used throughout the code like (for example):
```cpp
update_db_usage(payer, config::billable_size_v<table_id_object>, ram_trace(get_action_id(), event_id.c_str(), "table", "add", "create_table"));
```
and
```cpp
      if (auto dm_logger = _get_deep_mind_logger()) {
         fc_dlog(*dm_logger, "RAM_OP ${action_id} ${event_id} ${family} ${operation} ${legacy_tag} ${payer} ${new_usage} ${delta}",
            ("action_id", trace.action_id)
            ("event_id", trace.event_id)
            ("family", trace.family)
            ("operation", trace.operation)
            ("legacy_tag", trace.legacy_tag)
            ("payer", account)
            ("new_usage", u.ram_usage)
            ("delta", ram_delta)
         );
      }
```
And we could use it for a generic RAM event like so:
```cpp
ram_trace{get_action_id()}
```

But I can understand if you want to make the generic one explicit. Also, I imagine you would like to avoid making changes everywhere you currently use `ram_trace` and `generic_ram_trace`. So how about the following?
```cpp
struct ram_trace {
public:
   ram_trace(uint32_t action_id, const char* event_id, const char* family, const char* operation, const char* legacy_tag)
   :action_id(action_id),event_id(event_id),family(family),operation(operation),legacy_tag(legacy_tag)
   {}

   inline uint32_t    get_action_id() const { return action_id; }	
   inline const char* get_event_id() const { return event_id; }
   inline const char* get_family() const { return family; }
   inline const char* get_operation() const { return operation; }
   inline const char* get_legacy_tag() const { return legacy_tag; }

private:
   ram_trace(uint32_t action_id)
   :action_id(action_id)
   {}

   friend ram_trace generic_ram_trace(uint32_t);

private:
   uint32_t     action_id  = 0;
   const char*  event_id   = "generic";
   const char*  family     = "generic";
   const char*  operation  = "generic";
   const char*  legacy_tag = "generic";
};

ram_trace generic_ram_trace(uint32_t action_id = 0) {
   return {action_id};
}
```
Or better yet avoid the default argument for `generic_ram_trace` (but then that would require some changes to the `resource_limits_test.cpp` file.

Once again that is all you would need to add to the header file and nothing would need to be added to the `trace.cpp` file.

libraries/chain/CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ file(GLOB HEADERS "include/eosio/chain/*.hpp"
                   "include/eosio/chain/webassembly/*.hpp"
                   "${CMAKE_CURRENT_BINARY_DIR}/include/eosio/chain/core_symbol.hpp" )
 
+option(EOSIO_ENABLE_DEVELOPER_OPTIONS "enable developer options for EOSIO" OFF)
```
arhag: I thought when you said you were going to move this, you were going to put it in the root CMakeLists.txt.

This doesn't belong in the chain folder because it doesn't actually affect libchain and instead affects the nodeos as a whole (specifically the `chain_plugin`). The corresponding `target_compile_definitions` will also need to moved to `plugins/chain_plugin/CMakeLists.txt` and changed to affect the `chain_plugin` target.

maoueh: Ahah sorry. It was wrongly placed before because it was wrapped within the conditional `if("eos-vm-oc" IN_LIST EOSIO_WASM_RUNTIMES` which was obviously completely wrong! So I moved it to be free on the conditional.

Sure, I'll move that to `chain_plugin`.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -222,6 +225,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "Number of worker threads in controller thread pool")
          ("contracts-console", bpo::bool_switch()->default_value(false),
           "print contract's output to console")
+         ("deep-mind", bpo::bool_switch()->default_value(false),
+          "print deeper information about eosio software")
```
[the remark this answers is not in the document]
maoueh: I was sure I had changed it. I backported the patch and your comments to your 2.0.x-dm to measure the performance impact, so I must have done it there instead only.

I picked your message and will make the necessary change.

libraries/chain/CMakeLists.txt
maoueh: So just to be sure I get this right.

The `option` definition should go on the root `CMakeLists.txt` and the `target_compile_definitions` in `plugins/chain_plugin/CMakeLists.txt`, correct?

libraries/chain/CMakeLists.txt
maoueh: Now tracked by another conversation.

unittests/resource_limits_test.cpp
maoueh: Removed.

libraries/chain/controller.cpp
maoueh: Replaced.

libraries/chain/CMakeLists.txt
arhag: Yes.

maoueh: Done.

plugins/chain_plugin/chain_plugin.cpp
maoueh: Done.

libraries/chain/include/eosio/chain/trace.hpp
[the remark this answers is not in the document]
maoueh: I went with an hybrid between your two version. I forced `action_id` in `generic_ram_trace(uint32_t action_id)` and made the fields public without any getter.

Tell me if alright with you.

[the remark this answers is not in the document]
arhag: That's fine with me. Make sure to clean up the commented out code.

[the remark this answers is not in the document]
maoueh: Yeah, oups. Pushed fixed version.

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -99,6 +99,8 @@ const static uint32_t   hashing_checktime_block_size       = 10*1024;  /// call
 const static eosio::chain::wasm_interface::vm_type default_wasm_runtime = eosio::chain::wasm_interface::vm_type::wabt;
 const static uint32_t   default_abi_serializer_max_time_us = 15*1000; ///< default deadline for abi serialization methods
 
+const static uint32_t   dmlog_abi_serializer_max_time_us = 5*1000*1000; ///< default deadline for abi serialization methods performed for deep mind
```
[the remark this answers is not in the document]
maoueh: Yeah sure. You're talking about the comment correct?

As for the 5s, it's really arbitrary, just to give plenty of time to the serializer to do its job. Not sure what would be a correct value.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -250,6 +255,10 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "In \"light\" mode all incoming blocks headers will be fully validated; transactions in those validated blocks will be trusted \n")
          ("disable-ram-billing-notify-checks", bpo::bool_switch()->default_value(false),
           "Disable the check which subjectively fails a transaction if a contract bills more RAM to another account within the context of a notification handler (i.e. when the receiver is not the code of the action).")
+#ifdef EOSIO_DEVELOPER
+         ("disable-all-subjective-mitigations", bpo::bool_switch()->default_value(false),
+          "Disable all subjective mitigations checks in the entire codebase.")
+#endif
```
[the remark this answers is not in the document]
maoueh: Here?:

```
            bool                     disable_all_subjective_mitigations = false; //< for testing purposes only
```

If here, what should I write exactly? 

Draft:

```
//< for developer & testing purposes, can be configured using `disable-all-subjective-mitigations` when `EOSIO_DEVELOPER` build option is provided
```

Otherwise, I'm not sure where...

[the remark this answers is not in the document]
heifner: Yes, That works.

libraries/chain/include/eosio/chain/config.hpp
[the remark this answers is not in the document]
heifner: Yes, the comment.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1077,6 +1120,13 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
             } );
 
       my->accepted_block_connection = my->chain->accepted_block.connect( [this]( const block_state_ptr& blk ) {
+         if (auto dm_logger = my->chain->get_deep_mind_logger()) {
+            fc_dlog(*dm_logger, "ACCEPTED_BLOCK ${num} ${blk}",
+               ("num", blk->block_num)
+               ("blk", blk)
```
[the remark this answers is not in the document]
maoueh: We log the full block state to avoid having to reconstruct it so we reduce the chance of differences, specially with the deferred handling that needs to be made correctly. It's much easier to log the full block state than taking the chance. And we also only log the transaction_trace for now in applied_transaction, so there is info missing like the transaction signatures, which are present in the `block_state`.

It's something that we could change in the future to reduce the amount of data transferred. I'll take not of this for a future improvement, but for now, I would prefer to avoid changing it if you agree.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1091,6 +1141,13 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
 
       my->applied_transaction_connection = my->chain->applied_transaction.connect(
             [this]( std::tuple<const transaction_trace_ptr&, const signed_transaction&> t ) {
+               if (auto dm_logger = my->chain->get_deep_mind_logger()) {
+                  fc_dlog(*dm_logger, "APPLIED_TRANSACTION ${block} ${traces}",
+                     ("block", chain().head_block_num() + 1)
+                     ("traces", chain().to_variant_with_abi(std::get<0>(t), abi_serializer::create_yield_function(fc::microseconds(config::dmlog_abi_serializer_max_time_us))))
```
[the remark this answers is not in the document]
maoueh: So, we never really bothered with this problem since are motto as always been: "let it crash, and deal with it when it happens". In fact, in never happen yet that we got an exception being thrown here. Not saying I would not change this.

I would need to think, doing it without ABI might be a solution, I just fear it goes unnoticed  this way. At the same time when the ABI is wrong or inexistent, it already keep it in hex form. I'll think of that today.

As for the 5s, like mentioned in a comment below, it's really arbitrary, enough in our eye to let the ABI serializer finish its job no matter what. Do you have a suggestion?

plugins/chain_plugin/chain_plugin.cpp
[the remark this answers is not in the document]
heifner: It does seem like something you would want to address in the future. It is fine for now.

plugins/chain_plugin/chain_plugin.cpp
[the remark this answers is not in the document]
heifner: It will not crash in all cases. In a number of these places it will be silently ignored. If you want a crash when this happens then I would add a try-catch, log why, and stop/crash. In this particular location it will ignore the exception and will not publish to the channel nor will any other signal slots be signaled. So it will just silently be ignored.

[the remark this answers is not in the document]
heifner: You say it has "never happened yet" I wonder if it might have and you have not noticed.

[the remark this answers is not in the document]
maoueh: Well, you are right that it’s possible that it went unnoticed already. However, that would mean we are missing some transactions, or some data in general, which I assume would have been reported. But I agree, it’s a possibility. I think in our reading code, we should assert that transaction traces match with what we have in the block. That could make a good safe guard.

Now, about the serializer throwing exceptions, I’ll go with a try catch and write the transaction without ABI serializer work as well as logging an error. I think it’s good compromise, will make life easier.

Now, are you ok if I create an helper for that, there is around 8 such call right now, probably in Controller? Or do you prefer that I avoid that and duplicate the code.

Once we agree here, I’ll make the necessary changes this afternoon as well as taking care of your other comments.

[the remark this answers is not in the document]
heifner: We don't want the duplicate code. I hate that controller becomes the dumping ground for everything because it is convenient, however, the `to_variant_with_abi` is already there so it makes sense to add a `maybe_to_variant_with_abi` or something like that, which does the same thing with a try-catch that falls back to copying `obj` to `pretty_output` if exception. Note that even that copy could be expensive.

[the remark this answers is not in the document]
maoueh: Yeah, the controller is such a beast, that's why I fear adding new stuff to it :)

Did the changes with `maybe_to_variant_with_abi`. I think we can afford the expensive copy, assuming this should not be triggered often and that it costs less than doing the ABI serialiazation to variant anyway.

Will see with the new log if this is really happening or not it seems. Also update all occurrences as well as the comments.

PR #8798, opened 2020-03-11
Trace api Automatic Slice File cleanup

brianjohnson5972:
## Change Description
- Added feature to automatically cleanup slices when data is no longer required.
- Added config parameter to indicate how many files at a minimum to keep after lib.
- Minor fixes.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/store_provider.cpp
+++ b/plugins/trace_api_plugin/store_provider.cpp
@@ -136,6 +136,9 @@ namespace eosio::trace_api {
       if( append ) {
          trace_file.seek_end(0);
       }
+      else {
+         trace_file.seek(0); // ensure we are at the start of the file
+      }
       return true;
    }
 
--- a/plugins/trace_api_plugin/store_provider.cpp
+++ b/plugins/trace_api_plugin/store_provider.cpp
@@ -12,14 +12,15 @@ namespace {
 
 namespace eosio::trace_api {
    namespace bfs = boost::filesystem;
-   store_provider::store_provider(const bfs::path& slice_dir, uint32_t width)
-   : _slice_directory(slice_dir, width) {
+   store_provider::store_provider(const bfs::path& slice_dir, uint32_t stride_width, std::optional<uint32_t> minimum_irreversible_history_blocks)
+   : _slice_directory(slice_dir, stride_width, minimum_irreversible_history_blocks) {
    }
 
    void store_provider::append(const block_trace_v0& bt) {
       fc::cfile trace;
       fc::cfile index;
-      find_or_create_slice_pair(bt.number, true, trace, index);
+      const uint32_t slice_number = _slice_directory.slice_number(bt.number);
+      _slice_directory.find_or_create_slice_pair(slice_number, true, trace, index);
       // storing as static_variant to allow adding other data types to the trace file in the future
       const uint64_t offset = append_store(data_log_entry { bt }, trace);
 
@@ -33,6 +34,7 @@ namespace eosio::trace_api {
       _slice_directory.find_or_create_index_slice(slice_number, true, index);
       auto le = metadata_log_entry { lib_entry_v0 { .lib = lib }};
       append_store(le, index);
+      _slice_directory.cleanup_old_slices(lib);
    }
 
    get_block_t store_provider::get_block(uint32_t block_height, const yield_function& yield) {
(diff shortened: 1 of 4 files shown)


plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -166,4 +159,47 @@ namespace eosio::trace_api {
...
+      const uint32_t lib_slice_number = slice_number( lib );
+      if (lib_slice_number < 1 || (_last_cleaned_up_slice && _last_cleaned_up_slice >= lib_slice_number))
+         return;
+
+      // can only cleanup a slice once our last needed history block (lib - *_minimum_irreversible_history_blocks)
+      // is out of that slice (... - width)
+      const int64_t cleanup_block_number = static_cast<int64_t>(lib) - static_cast<int64_t>(*_minimum_irreversible_history_blocks) - _width;
+      if (cleanup_block_number > 0) {
+         uint32_t cleanup_slice_num = slice_number(static_cast<uint32_t>(cleanup_block_number));
+         // since we subtracted width, we are guaranteed cleanup_slice_num is not
```
b1bart: this comment seems cut off, maybe should end with "the slice that contains LIB"

plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -166,4 +159,47 @@ namespace eosio::trace_api {
...
+         while (!_last_cleaned_up_slice || *_last_cleaned_up_slice < cleanup_slice_num) {
+            fc::cfile trace;
+            fc::cfile index;
+            const uint32_t slice_to_clean = _last_cleaned_up_slice ? *_last_cleaned_up_slice + 1 : 0;
+            const bool trace_found = find_trace_slice(slice_to_clean, false, trace);
+            if (trace_found) {
+               trace.close();
```
b1bart: Is it necessary to have opened and closed the file for consistency?  or is this an artifact of `find_trace_slice` opening the file?

plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -166,4 +159,47 @@ namespace eosio::trace_api {
...
+               bfs::remove(trace.get_file_path());
+            }
+            const bool index_found = find_index_slice(slice_to_clean, false, index);
```
b1bart: I think we should delete the index and then the data.  I'm not sure if the OS will maintain that order but if it does our reader code will have a better chance at returning a `404` if the index disappears first rather than a temporary `500` because the index was still present but the data was gone.   I realize re-ordering this won't completely close that gap but we should attempt to be consistent here.

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -100,6 +100,9 @@ struct trace_api_common_impl {
                   "the location of the trace directory (absolute path or relative to application data dir)");
       cfg_options("trace-slice-stride", bpo::value<uint32_t>()->default_value(10'000),
                   "the number of blocks each \"slice\" of trace data will contain on the filesystem");
+      cfg_options("trace-minimum-irreversible-history-blocks", boost::program_options::value<int32_t>()->default_value(-1),
```
b1bart: need to remove `trace-minimum-irreversible-history-us` below however, I would argue that this parameter should not be part of the common config options in the future but for now it is fine.  If we separate into a reader-only plugin, that plugin would still have the "common" configs but this one would be irrelevant as it would be ignored.

plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -33,6 +34,7 @@ namespace eosio::trace_api {
       _slice_directory.find_or_create_index_slice(slice_number, true, index);
       auto le = metadata_log_entry { lib_entry_v0 { .lib = lib }};
       append_store(le, index);
+      _slice_directory.cleanup_old_slices(lib);
```
b1bart: I think this is fine for now however, we should add a backlog task to process this on a low priority timer instead of in the critical path of the writing process.  I don't think it will do work often enough to warrant a change prior to the initial release.

plugins/trace_api_plugin/store_provider.cpp
brianjohnson5972: Since resetting the unique_file pointer calls fclose, this isn't needed.

plugins/trace_api_plugin/store_provider.cpp
brianjohnson5972: yep

plugins/trace_api_plugin/store_provider.cpp
brianjohnson5972: I had thought about that, but figured that this is only on lib write and since I cache the last deleted it is only on one lib write per stride, all other ones will just do a simple logic check.

plugins/trace_api_plugin/store_provider.cpp
brianjohnson5972: yes

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -111,13 +114,24 @@ struct trace_api_common_impl {
 
       slice_stride = options.at("trace-slice-stride").as<uint32_t>();
 
-      store = std::make_shared<store_provider>(trace_dir, slice_stride);
+      const char* trace_minimum_irreversible_history_blocks = "trace-minimum-irreversible-history-blocks";
+      const int32_t blocks = options.at(trace_minimum_irreversible_history_blocks).as<int32_t>();
+      EOS_ASSERT(blocks >= -1, chain::plugin_config_exception,
+                 "\"${blocks}\" must be greater to or equal to -1.",("blocks",trace_minimum_irreversible_history_blocks));
```
b1bart: I think `trace_minimum_irreversible_history_blocks` is the `const char *` and this assert message's value should be `blocks` right? 

or, is the message just meant to deduplicate the constant string?   maybe change the "key" to `${opt}` if so?

as is its confusing.

heifner: I find the `const char *` of same name as string not helpful. I say just duplicate the string.

brianjohnson5972: ok, but I hate making 2 strings always match, let along 3.

PR #8799, opened 2020-03-11
Issue 8357 sig handling

ndcgundlach:
## Change Description

Added child signal handling to prevent zombie processes

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


programs/cleos/main.cpp
```diff
@@ -2421,7 +2421,25 @@ CLI::callback_t header_opt_callback = [](CLI::results_t res) {
    return true;
 };
 
+
+// TODO remove this when eos-vm-oc is broken out into a separate lib and not linked in
```
swatanabe-b1: Since this is for eos-vm-oc, it would be better to put it in compile_monitor_trampoline::start.

spoonincode:
What happened to the approach we talked about previously, where we only `-wrap=main` nodeos & unittests?

programs/cleos/main.cpp
ndcgundlach: I considered that but I didn't want to affect signal handling in other processes and just limit it to cleos.

ndcgundlach:
> What happened to the approach we talked about previously, where we only `-wrap=main` nodeos & unittests?

I was going to need to define a _real_main in the other processes and thought that was a bit hokey, esp given that you were planning on changing the wrap to support OS X.  If you want, I can do that.

swatanabe-b1:
It should be safe to put a definition of `__real_main` in eos-vm-oc as long as it's in a different translation unit from `__wrap_main`.

libraries/chain/webassembly/eos-vm-oc/compile_monitor.cpp
```diff
@@ -278,6 +278,11 @@ void launch_compile_monitor(int nodeos_fd) {
 
 struct compile_monitor_trampoline {
    void start() {
+      struct sigaction sa;
+      sa.sa_handler = SIG_DFL;
+      sa.sa_flags = SA_NOCLDWAIT;
+      sigaction(SIGCHLD, &sa, nullptr);
```
swatanabe-b1: This seems risky.  Have you verified that nodeos (or any other program including the unit tests that might use OC) never waits for any subprocess?

ndcgundlach: I felt it was risky also that is why I had it in the cleos process instead of here.  Your comment above was for me to move it.  I did not see where nodeos was waiting, however.

swatanabe-b1: Is it still needed at all?  I thought that cleos would no longer start up OC at all after 9317d70.

ndcgundlach: Good point!  Not really.

swatanabe-b1:
The tests are failing because the fc submodule is out of sync.

ndcgundlach:
> The tests are failing because the fc submodule is out of sync.

I messed up a commit.  I am actually going to kill the PR.  I have a new branch where I cherry-picked the non messed up commits.  I will do a PR on that one.

PR #8801, opened 2020-03-11
Expire blacklisted scheduled transactions by LIB time - develop

heifner:
## Change Description

- Use LIB time for removal of blacklisted scheduled transactions instead of `now`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2830,6 +2830,11 @@ block_id_type controller::last_irreversible_block_id() const {
    return get_block_id_for_num( lib_num );
 }
 
+block_timestamp_type controller::last_irreversible_block_timestamp() const {
+   return my->fork_db.root()->header.timestamp;
+}
+
+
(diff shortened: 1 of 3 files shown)


libraries/chain/controller.cpp
```diff
@@ -2830,6 +2830,11 @@ block_id_type controller::last_irreversible_block_id() const {
    return get_block_id_for_num( lib_num );
 }
 
+block_timestamp_type controller::last_irreversible_block_timestamp() const {
```
arhag: Call this `last_irreversible_block_time` and return a `time_point` to be consistent with the other ones.

heifner: Done. I used `.to_time_point()` since I dislike auto type conversion. Maybe my days with Ada programming language coming out, but really because I've been burned too many times.

PR #8834, opened 2020-03-17
trace_api_plugin yield timeout - develop

heifner:
## Change Description

- `trace_api_plugin` now honors `http-max-response-time-ms` as a specification of retrieval and ABI processing time. `http-max-response-time-ms` is also the time allowed for `http_plugin` conversion to JSON. Maximum time allowed for `trace_api_plugin` `get_block` is now 2x `http-max-response-time-ms`
- Modified `http_plugin` logging to not error log on `timeout_exception`.
- Future work: add yield capability to `http_plugin` `url_handler` so `http-max-response-ms` can be used as a specification of complete time allowed for api calls e.g. `get_table_rows` allowing the removal of hard-coded 10ms limit.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/test/test_data_handlers.cpp
+++ b/plugins/trace_api_plugin/test/test_data_handlers.cpp
@@ -15,10 +15,10 @@ BOOST_AUTO_TEST_SUITE(abi_data_handler_tests)
       auto action = action_trace_v0 {
          0, "alice"_n, "alice"_n, "foo"_n, {}, {}
       };
-      abi_data_handler handler;
+      abi_data_handler handler(exception_handler{});
 
       auto expected = fc::variant();
-      auto actual = handler.process_data(action);
+      auto actual = handler.process_data(action, [](){});
 
       BOOST_TEST(to_kv(expected) == to_kv(actual), boost::test_tools::per_element());
    }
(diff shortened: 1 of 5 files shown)


plugins/trace_api_plugin/test/test_data_handlers.cpp
```diff
@@ -15,10 +15,10 @@ BOOST_AUTO_TEST_SUITE(abi_data_handler_tests)
       auto action = action_trace_v0 {
          0, "alice"_n, "alice"_n, "foo"_n, {}, {}
       };
-      abi_data_handler handler;
+      abi_data_handler handler(exception_handler{});
 
       auto expected = fc::variant();
-      auto actual = handler.process_data(action);
+      auto actual = handler.process_data(action, [](){});
```
b1bart: I didn't catch this in the last review (for 2.0.x) but is there a reason why these are no-op functions instead of `{}` or `nullptr`?  The `optional_delegate` should have a default constructor inherited from `std::function` which will prevent the call right?

heifner: No, just habit from working with `std::function`

PR #8837, opened 2020-03-17
Trace API compressed data log support

b1bart:
## Change Description

Initial testing with the EOS Public Network showed that 10k block slice files averaged about 3GiB during peak blocks.  These traces compressed down to ~120MiB with `gzip -9` (4% of their original size).   This indicated that there was a lot of redundancy in the data stream that could be compressed. 

This PR adds:
- [ ] Support for a `clog` or compressed log format when reading the trace data from the filesystem.
- [ ] Configurations for a background thread that will optionally compress irreversible sections of the log without delaying the critical operations of the writer (this thread also takes over the job of pruning data files).
- [ ] Logging from this background thread output to the `debug` level of the Trace API specific logger.
- [ ] a command line utility that can be invoked to compress logs from an outside process.

### `clog` format

The `clog` is a generic compressed file with an index of seek-able decompression points appended to the end.  The layout looks like this:

```
 /====================\ file offset 0
|                    |
|  Compressed Data   |
|  with seek points  |
|                    |
|--------------------|  file offset END - 2 - (16 * seek point count)
|                    |
|  mapping of        |
|    orig offset to  |
|    seek pt offset  |
|                    |
|--------------------|  file offset END - 2
|  seek pt count     |
\====================/  file offset END
```

The data is compressed as raw zlib with "seek points" in the form of full compressor flushes placed at regular intervals.  A decompressor can start from any of these "seek points" without reading previous data and it can traverse a seek point if data straddles one without issue.

As the file is being compressed, the seek point index records the original uncompressed offset with the new compressed offset creating a mapping so that the original index values (uncompressed offsets) can be mapped to the nearest seek point _before_ the uncompressed offset.   This dramatically reduces the time to seek to parts of the uncompressed file that were later in the stream

With 512 seek points, the compression is still comparable to `gzip -9` on the EOS public network test data set and the data is still immediately available via the RPC.  This reduces the growth of the trace directory from ~50GiB/day to ~2.5GiB/day for full data. 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [X] Documentation Additions

There are a few new configuration options and a new tool:
### `trace-minimum-uncompressed-irreversible-history-blocks`
Number of blocks to ensure are uncompressed past LIB. Compressed "slice" files are still accessible but may carry a performance loss on retrieval.

A value of `-1` indicates that automatic compression of "slice" files will be turned off.

### `trace_api_util` command

This is a command line util in `plugins/trace_api_plugin/utils` that will perform all future command line tasks associated with the Trace API.  For now, it provides the `log` to `clog` conversion for manual operations.  This is for integrating with existing maintenance process or use-cases where manual control is desirable.

```
Usage: trace_api_util <options> command ...

Commands:
  compress              Compress a trace file to into the "clog" format

Options:
  -h [ --help ]         show usage help message
```

```
Usage: trace_api_util compress <options> input-path [output-path]

Compress a trace file to into the "clog" format.  By default the name of the
of the compressed file will the the same as the input-path with a changing the
extension to "clog".

Positional Options:
  input-path                      path to the file to compress
  output-path                     [Optional] output file or directory path.

Options:
  -h [ --help ]                         show usage help message
  -s [ --seek-point-stride ] arg (=512) the number of bytes between seek points
                                        in a compressed trace.  A smaller 
                                        stride may degrade compression 
                                        efficiency but increase read efficiency
```

Example:
```
./trace_api_util compress ~/eosio/traces/trace_0000000000_0000010000.log
```
This will compress `~/eosio/traces/trace_0000000000_0000010000.log` and produce the file `~/eosio/traces/trace_0000000000_0000010000.clog`.  It will not automatically remove the original file.


plugins/trace_api_plugin/utils/compress_trace.cpp
```diff
@@ -0,0 +1,102 @@
...
+   auto opts = vis_desc.add_options();
+   opts("help,h", "show usage help message");
+   opts("seek-points,s", bpo::value<uint32_t>()->default_value(512),
+        "the number of seek points to create when compressing the trace.  "
+        "More seek points will degrade compression efficiency but increase read efficiency");
+
+   bpo::options_description hidden_desc("hidden");
+   auto hidden_opts = hidden_desc.add_options();
+   hidden_opts("input-path,i", bpo::value< std::string >()->required(), "input path");
+   hidden_opts("output-path,o", bpo::value< std::string >(), "output path");
```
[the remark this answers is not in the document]
heifner: Also the PR could use an example of usage of this util.

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -103,6 +103,11 @@ struct trace_api_common_impl {
       cfg_options("trace-minimum-irreversible-history-blocks", boost::program_options::value<int32_t>()->default_value(-1),
                   "Number of blocks to ensure are kept past LIB for retrieval before \"slice\" files can be automatically removed.\n"
                   "A value of -1 indicates that automatic removal of \"slice\" files will be turned off.");
+      cfg_options("trace-minimum-uncompressed-irreversible-history-blocks", boost::program_options::value<int32_t>()->default_value(-1),
+                  "Number of blocks to ensure are uncompressed past LIB. Compressed \"slice\" files are still accessible but may carry a performance loss on retrieval\n"
+                  "A value of -1 indicates that automatic compression of \"slice\" files will be turned off.");
+      cfg_options("trace-compression-seek-points", boost::program_options::value<uint32_t>()->default_value(512),
+                  "Number of seek points to use when compressing \"slice\" files.  A higher number of seek points can increase read performance at the expense of compression efficiency\n");
```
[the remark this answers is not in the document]
b1bart: I will remove it, it is exposed in the manual process so, if you really want control we have a route for you to do so, just not through the manual process.

plugins/trace_api_plugin/include/eosio/trace_api/common.hpp
```diff
@@ -63,6 +63,8 @@ namespace eosio::trace_api {
    using exception_with_context = std::tuple<const std::exception_ptr&, char const *, uint64_t, char const *>;
    using exception_handler = optional_delegate<void, const exception_with_context&>;
 
+   using log_handler = optional_delegate<void, const std::string&>;
```
heifner: `optional_delegate` was moved to `fc` in `release/2.0.x`. Also it now supports syntax: `optional_delegate<void(const std::string&)>`

heifner: I'm not sure how you have `optional_delegate` here without having a conflict on merge.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <fc/io/cfile.hpp>
+
+namespace eosio::trace_api {
+
+   class compressed_file_datastream;
+   struct compressed_file_impl;
+   /**
+    * wrapper for read-only access to a compressed file.
+    * compressed files support seeking though the efficiency of random access is lower than
+    * an uncompressed file in trade for smaller on-disk resource usage
```
heifner: I think this sentence could use some work.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,87 @@
...
+    */
+   class compressed_file {
+   public:
+      compressed_file( fc::path file_path );
```
heifner: `explicit`

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+      compressed_file_datastream create_datastream();
+      
+      static bool process( const fc::path& input_path, const fc::path& output_path, uint16_t seek_point_count );
+
+      /**
+       * Provide default move construction/assignment
+       */
+      compressed_file( compressed_file&& );
+      compressed_file& operator= ( compressed_file&& );
```
heifner: I would move these up under other constructor.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
heifner: "Compressed files support seeking. The efficiency of random access is lower than an uncompressed file. The trade off is smaller on-disk resource usage."

plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
```diff
@@ -255,6 +310,14 @@ namespace eosio::trace_api {
        */
       std::optional<data_log_entry> read_data_log( uint32_t block_height, uint64_t offset ) {
          const uint32_t slice_number = _slice_directory.slice_number(block_height);
+
+         // Prefer a compressed trace if one exists
```
heifner: Shouldn't you prefer an uncompressed over a compressed?

plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -175,35 +203,117 @@ namespace eosio::trace_api {
...
-      if (!_minimum_irreversible_history_blocks)
-         return;
+   void slice_directory::set_lib(uint32_t lib) {
+      _best_known_lib = lib;
+      _maintenance_condition.notify_one();
+   }
+
+   void slice_directory::start_maintenance_thread(log_handler log) {
+      _maintenance_thread = std::thread([this, log=std::move(log)](){
+         fc::set_os_thread_name( "TraceAPI-Maintenance" );
```
heifner: `console_appender` uses 9 chars for thread name. How about `trace-mx`

plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
```diff
@@ -184,10 +217,23 @@ namespace eosio::trace_api {
...
       std::optional<uint32_t> _last_cleaned_up_slice;
+      const std::optional<uint32_t> _minimum_uncompressed_irreversible_history_blocks;
+      std::optional<uint32_t> _last_compressed_slice;
+      uint32_t _compression_seek_points;
+
+      std::atomic<uint32_t> _best_known_lib{0};
+      std::mutex _maintenance_mtx;
+      std::condition_variable _maintenance_condition;
+      std::thread _maintenance_thread;
+      std::atomic_bool _maintenance_shutdown{false};
```
heifner: This looks fine, but I think it would have been simpler to use `named_thread_pool` with size of 1 and posted jobs using `async_thread_pool`.

plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
b1bart: I could see that making sense, will change.

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            if (written < n && ret == Z_STREAM_END) {
+               throw std::runtime_error("Attempting to read past the end of a compressed file");
+            }
+         } while (strm.avail_out == 0 && written < n);
+      }
+   }
+
+   void seek( long loc, const std::unique_ptr<fc::cfile>& file_ptr  ) {
```
heifner: Minor thing, but I would make this take a `fc::cfile&` instead of an `unique_ptr` and `const` here is a bit misleading.

plugins/trace_api_plugin/include/eosio/trace_api/common.hpp
b1bart: yeah that was odd but its fixed now

plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
b1bart: I'm on the fence here.  If you are fine I will leave it.   The idea of a thread-pool implies that you could have more than one thread in the pool and as this is not safe for parallel background tasks I would want to do a little more work making it resilient to a future engineer making a reasonable assumption that they could parallelize the compression tasks by upping the number of threads in the pool.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,87 @@
...
+         file_ptr.reset();
+      }
+
+      auto get_file_path() const {
+         return file_path;
+      }
+
+      compressed_file_datastream create_datastream();
+      
+      static bool process( const fc::path& input_path, const fc::path& output_path, uint16_t seek_point_count );
```
brianjohnson5972: Needs a comment indicating this converts the input_path file to a compressed file at output_path, and maybe a little more info with it.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace eosio::trace_api {
+
+   class compressed_file_datastream;
+   struct compressed_file_impl;
+   /**
+    * wrapper for read-only access to a compressed file.
+    * compressed files support seeking though the efficiency of random access is lower than
+    * an uncompressed file in trade for smaller on-disk resource usage
+    */
```
brianjohnson5972: If the cfile read interface is matched for a code reason (like the unpack requirements below on the stream) then add that to the comment (if it is just because it is better to use the same pattern than to not, then no need for comment)

plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp
heifner: I'm fine with leaving it as is.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,95 @@
...
+
+      void open() {
+         file_ptr = std::make_unique<fc::cfile>();
+         file_ptr->set_file_path(file_path);
+         file_ptr->open("rb");
+      }
+
+      bool is_open() const { return (bool)file_ptr; }
+
+      void seek( long loc );
```
heifner: Precondition is that the file is not empty. Would be nice for this method to fail better so that `compressed_file` could be more general purpose.

heifner: Actually I guess the `cfile::seek` will throw a `std::ios_base::failure` if unable to seek to location.

plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -121,15 +126,44 @@ struct trace_api_common_impl {
          minimum_irreversible_history_blocks = blocks;
       }
 
-      store = std::make_shared<store_provider>(trace_dir, slice_stride, minimum_irreversible_history_blocks);
+      const int32_t uncompressed_blocks = options.at("trace-minimum-uncompressed-irreversible-history-blocks").as<int32_t>();
+      EOS_ASSERT(uncompressed_blocks >= -1, chain::plugin_config_exception,
```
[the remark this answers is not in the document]
b1bart: Im going to pass on this for now.  it its set higher then we are not incorrectly handling it there just wont be anything to compress.  If we see an uptick of users who are confused by this not being a warning/error we can consider changing.

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+      } else {
+         bytes_remaining_before_sync -= read_size;
+      }
+   }
+
+   deflateEnd(&strm);
+   input_file.close();
+
+   // write out the seek point table
+   static_assert(sizeof(seek_point_entry) == 16, "unexpected size for seek point");
```
brianjohnson5972: Make 16 a constexpr static var and use it here and elsewhere.

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+   output_file.write(reinterpret_cast<const char*>(seek_point_map.data()), seek_point_map.size() * sizeof(seek_point_entry));
+
+   // write out the seek point count
+   static_assert(sizeof(seek_point_count) == 2, "Unexpected size for seek point count");
```
brianjohnson5972: make 2 a constexpr static var and use here and elsewhere.

plugins/trace_api_plugin/include/eosio/trace_api/compressed_file.hpp
```diff
@@ -0,0 +1,145 @@
...
+    * compressed files support seeking and reading
+    *
+    * the efficiency of seeking is lower than that of an uncompressed file as each seek translates to
+    *  - 2 seeks + 1 read to load and process the seek-point-mapping
+    *  - potentially a read/decompress/discard of the data between the seek point and the requested offset
+    *
+    * More seek points can lower the average amount of data that must be read/decompressed/discarded in order
+    * to seek to any given offset.  However, each seek point has some effect on the file size as it represents a
+    * flush of the compressor which can degrade compression performance.
+    */
```
brianjohnson5972: I'd like either a reference here to the compressed file format being documented before the process method or that moved here.  Advantage of having it there is it is more likely to be maintained if it changes, but I kept thinking it would be good to have that documented somewhere and then finally found it when I got to the process method.

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+         return;
+      }
+
+      long remaining = loc;
+
+      // special case when this is before the first seek point
+      if ( iter == seek_point_map.begin() ) {
+         file.seek(0);
+      } else {
+         // if lower bound wasn't exact it will be one past the seek point we need
```
brianjohnson5972: got confused on this, I think changing it to "iter" instead of "it" will make it clearer.

heifner: What I wasn't sure about here is if it is safe not to check for `seek_point_map.end()`. I think I convinced myself that is not possible but I wouldn't stake my life on it.

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+ * @param seek_point_count
+ * @return
+ */
+bool compressed_file::process( const fc::path& input_path, const fc::path& output_path, uint16_t seek_point_count ) {
+   if (!fc::exists(input_path)) {
+      throw std::ios_base::failure(std::string("Attempting to create compressed_file from file that does not exist: ") + input_path.generic_string());
+   }
+   std::vector<seek_point_entry> seek_point_map(seek_point_count);
+
+   size_t input_size = fc::file_size(input_path);
```
brianjohnson5972: Make input_size and ideal_seek_stride const

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+   if (deflateInit2(&strm, Z_BEST_COMPRESSION, Z_DEFLATED, -15, 8, Z_DEFAULT_STRATEGY) != Z_OK) {
+      return false;
+   }
+
+   constexpr size_t buffer_size = 64*1024;
+   auto input_buffer = std::vector<uint8_t>(buffer_size);
+   auto output_buffer = std::vector<uint8_t>(buffer_size);
+
+   auto bytes_remaining_before_sync = ideal_seek_stride;
+   int next_sync_point = 0;
```
brianjohnson5972: verify/assert ideal_seek_stride is less than buffer_size, std::min check below can drop checking buffer_size

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+   size_t read_offset = 0;
+   while (read_offset < input_size) {
+      auto bytes_remaining = input_size - read_offset;
+      auto read_size = std::min({ buffer_size, bytes_remaining, bytes_remaining_before_sync });
```
brianjohnson5972: doesn't change so make it const

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,311 @@
...
+      input_file.read(reinterpret_cast<char*>(input_buffer.data()), read_size);
+
+      strm.avail_in = read_size;
+      strm.next_in = input_buffer.data();
+
+      do {
```
brianjohnson5972: Can we make this a method, or a lambda and pass in Z_NO_FLUSH/Z_FINISH/Z_FULL_FLUSH?  Should method assert str.avail_in == 0 for Z_NO_FLUSH? (or set next_in to input_buffer and set avail_in also)

spoonincode:
How does this compare to just letting a modern filesystem do the compression transparently? I worry we're going to end up with implementing compression code multiple times over (we already compress the state history log).

plugins/trace_api_plugin/compressed_file.cpp
brianjohnson5972: I was confused on this, and thus resolving it now that I am not confused.

b1bart:
> How does this compare to just letting a modern filesystem do the compression transparently? 

As this is an optional part, not enabled by default (unlike the state history plugin), users who are running on a compressed filesystem may not be the target users for the feature.  

For other users, where they do not have a modern compressed filesystem, are billed by size regardless of the underlying filesystem (networked file storage, S3, etc), or are moving these portable files around between systems with divergent filesystems, it may be useful.

spoonincode:
Understood, it just seems anti-KISS. And we may end up with every plugin doing its own compression mechanism.

plugins/trace_api_plugin/utils/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_executable( compress_trace compress_trace.cpp )
```
spoonincode: Should this be installed and/or part of binaries?

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -0,0 +1,315 @@
...
+      throw std::ios_base::failure(std::string("Attempting to create compressed_file from file that does not exist: ") + input_path.generic_string());
+   }
+
+   const size_t input_size = fc::file_size(input_path);
+   if (input_size == 0) {
+      throw std::ios_base::failure(std::string("Attempting to create compressed_file from file that is empty: ") + input_path.generic_string());
+   }
+
+   // subtract 1 to make sure that the truncated division will only create a seek point if there is at least one byte
+   // in the next stride.  So, a file size of N and a stride <= N results in 0 seek points.  N + 1 will have a seek
```
brianjohnson5972: wouldn't this be "and a stride >= N results in 0 seep points."?

b1bart: ... yes

PR #8838, opened 2020-03-18
set_action_return_value prohibited for context free actions

heifner:
## Change Description

- `set_action_return_value` should not be allowed on context-free actions
- The return value has non-trivial side-effects since it is committed to in the `action_mroot` via the `action_receipt`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/wasm_interface.cpp
```diff
@@ -1122,12 +1122,19 @@ class action_api : public context_aware_api {
...
+      using context_aware_api::context_aware_api;
+
       void set_action_return_value( array_ptr<char> packed_blob, uint32_t datalen ) {
          context.action_return_value.assign( packed_blob.value, packed_blob.value + datalen );
       }
+
 };
 
-class console_api : public context_aware_api {
+   class console_api : public context_aware_api {
```
arhag: Unnecessary indent.

PR #8841, opened 2020-03-18
Removed H1 title in the plugin file

bobgt:
## Change Description
Removed **H1** title in the markdown file as front matter metadata already exists. 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

## Documentation Formatting and Syntax
- [x] Syntax Deletion

### DIFF
--- a/docs/01_nodeos/03_plugins/index.md
+++ b/docs/01_nodeos/03_plugins/index.md
@@ -23,6 +23,7 @@ For information on specific plugins, just select from the list below:
 * [`state_history_plugin`](state_history_plugin/index.md)
 * [`test_control_api_plugin`](test_control_api_plugin/index.md)
 * [`test_control_plugin`](test_control_plugin/index.md)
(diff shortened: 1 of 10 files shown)


dskvr:
The frontmatter should probably be manually defined because the `name_of_plugin` as a signature would likely be better than an auto-formatted variant such as `Name Of Plugin`. Additionally, it would need to be applied to all plugins, not just this one.

bobgt:
> The frontmatter should probably be manually defined because the `name_of_plugin` as a signature would likely be better than an auto-formatted variant such as `Name Of Plugin`. Additionally, this fix would probably need to be applied to all plugins, not just this one.

Thanks. Yes, we have to do it for all the plugin files.

bobgt:
Closing this PR as it the changes here are addressed in PR# 8842.

PR #8846, opened 2020-03-18
skip context free actions during light validation

heifner:
## Change Description

- Skip context_free_action execution during light validation.
- Since context free actions do not have any side-effects (except for possible console output) and can only influence execution via asserting, the fact that they exist in a signed block is indication that the producer executed them and they didn't assert. Therefore, they can be skipped during light validation where producers are implicitly trusted.
- Any node that wishes to include console output in action traces for context free actions will need to run with `--disable-replay-opts`

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

- Any node that wishes to include console output in action traces for context free actions will need to run with `--disable-replay-opts`


libraries/chain/transaction_context.cpp
```diff
@@ -638,12 +638,14 @@ namespace eosio { namespace chain {
       const auto& db = control.db();
       const auto& auth_manager = control.get_authorization_manager();
 
-      for( const auto& a : trx.context_free_actions ) {
-         auto* code = db.find<account_object, by_name>(a.account);
-         EOS_ASSERT( code != nullptr, transaction_exception,
-                     "action's code account '${account}' does not exist", ("account", a.account) );
-         EOS_ASSERT( a.authorization.size() == 0, transaction_exception,
-                     "context-free actions cannot have authorizations" );
+      if( trx.context_free_actions.size() > 0 && !control.skip_auth_check() ) {
```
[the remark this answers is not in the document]
heifner: Optimization. Figured that it is often empty and quicker to check than the other.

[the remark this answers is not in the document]
heifner: How would you feel about `skip_trx_checks()` which is probably what I should have used?

libraries/chain/apply_context.cpp
```diff
@@ -61,51 +61,53 @@ void apply_context::exec_one()
       try {
          action_return_value.clear();
          receiver_account = &db.get<account_metadata_object,by_name>( receiver );
-         privileged = receiver_account->is_privileged();
-         auto native = control.find_apply_handler( receiver, act->account, act->name );
-         if( native ) {
-            if( trx_context.enforce_whiteblacklist && control.is_producing_block() ) {
-               control.check_contract_list( receiver );
-               control.check_action_list( act->account, act->name );
+         if( !context_free || (context_free && !control.skip_auth_check() ) ) {
```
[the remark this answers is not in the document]
heifner: How would you feel about skip_trx_checks() which is probably what I should have used?

libraries/chain/apply_context.cpp
```diff
@@ -61,51 +61,53 @@ void apply_context::exec_one()
       try {
          action_return_value.clear();
          receiver_account = &db.get<account_metadata_object,by_name>( receiver );
-         privileged = receiver_account->is_privileged();
-         auto native = control.find_apply_handler( receiver, act->account, act->name );
-         if( native ) {
-            if( trx_context.enforce_whiteblacklist && control.is_producing_block() ) {
-               control.check_contract_list( receiver );
-               control.check_action_list( act->account, act->name );
+         if( !context_free || (context_free && !control.skip_trx_checks() ) ) {
```
arhag: The second `context_free` is redundant.

Maybe for clarity you can rewrite as:
```cpp
if( !(context_free && control.skip_trx_checks()) ) {
```

unittests/api_tests.cpp
```diff
@@ -845,6 +845,75 @@ BOOST_FIXTURE_TEST_CASE(deferred_cfa_success, TESTER)  try {
...
+   chain.set_transaction_headers(trx);
+   // run normal passing case
+   auto sigs = trx.sign(chain.get_private_key(N(testapi), "active"), chain.control->get_chain_id());
+   auto trace = chain.push_transaction(trx);
+   blocks.push_back(chain.produce_block());
+
+   BOOST_REQUIRE(trace->receipt);
+   BOOST_CHECK_EQUAL(trace->receipt->status, transaction_receipt::executed);
+   BOOST_CHECK_EQUAL(2, trace->action_traces.size());
+   BOOST_CHECK_EQUAL("test\n", trace->action_traces.at(0).console); // cfa executed
```
arhag: Please also add a check on the `context_free` field here and for `other` below. Also grab and store the `trace->receipt->global_sequence` to compare it the one returned from `other` below.

heifner: Added compare of `digest()` since that is important. I also added the `global_sequence` since that will provide a better error message if this fails.

libraries/chain/apply_context.cpp
heifner: done

PR #8848, opened 2020-03-19
Skip checks 

heifner:
## Change Description

The implementation of the following options were changed to more closely reflect their descriptions.
 - `force-all-checks`
          do not skip any validation checks while replaying blocks (useful for replaying blocks from untrusted source)
 - `disable-replay-opts`
          disable optimizations that specifically target replay

Previously `disable-replay-opts` was tied to skipping transaction validation checks during replay. However, the existing option `force-all-checks` is a better indication of this functionality. Now `disable-replay-opts` only disables **optimizations** during replay; specifically the creation of database undo sessions. Now `force-all-checks` enforces the additional transaction validation checks that can be skipped when blocks are from a trusted source.

`controller::skip_trx_checks()` and `controller::skip_auth_check()` both now respect the `force-all-checks` option.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

- `force-all-checks` help text was updated.
- Additional information above on the use of these options.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3011,7 +3011,7 @@ optional<producer_authority_schedule> controller::proposed_producers()const {
    return producer_authority_schedule::from_shared(gpo.proposed_schedule);
 }
 
-bool controller::light_validation_allowed(bool replay_opts_disabled_by_policy) const {
+bool controller::light_validation_allowed() const {
    if (!my->pending || my->in_trx_requiring_checks) {
       return false;
    }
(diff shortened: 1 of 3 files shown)


libraries/chain/controller.cpp
```diff
@@ -3049,7 +3049,7 @@ bool controller::skip_db_sessions( ) const {
 }
 
 bool controller::skip_trx_checks() const {
-   return light_validation_allowed(my->conf.disable_replay_opts);
+   return light_validation_allowed(my->conf.force_all_checks);
```
arhag: Can we just remove the arguments for `light_validation_allowed` now?

heifner: Done.

PR #8851, opened 2020-03-23
add light validation sync test

huangminghuang:
## Change Description
Add an integration test for messaging syncing between a production node and a light validation node.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/light_validation_sync_test.py
+++ b/tests/light_validation_sync_test.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env python3
+
+from testUtils import Account
+from testUtils import Utils
+from Cluster import Cluster
+from WalletMgr import WalletMgr
+from Node import Node
+from Node import ReturnType
+from TestHelper import TestHelper
+from TestHelper import AppArgs
+import json
+
+###############################################################
+# nodeos_chainbase_allocation_test
+#
+# Test snapshot creation and restarting from snapshot
+#
+###############################################################
+
+# Parse command line arguments
+args = TestHelper.parse_args({"-v","--clean-run","--dump-error-details","--leave-running","--keep-logs"})
+Utils.Debug = args.v
+killAll=args.clean_run
+dumpErrorDetails=args.dump_error_details
+dontKill=args.leave_running
+killEosInstances=not dontKill
+killWallet=not dontKill
+keepLogs=args.keep_logs
+
+walletMgr=WalletMgr(True)
+cluster=Cluster(walletd=True)
+cluster.setWalletMgr(walletMgr)
+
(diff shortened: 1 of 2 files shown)


tests/light_validation_sync_test.py
```diff
@@ -0,0 +1,112 @@
...
+from Node import Node
+from Node import ReturnType
+from TestHelper import TestHelper
+from TestHelper import AppArgs
+import json
+
+###############################################################
+# nodeos_chainbase_allocation_test
+#
+# Test snapshot creation and restarting from snapshot
```
heifner: Update comment for this test.

tests/light_validation_sync_test.py
```diff
@@ -0,0 +1,112 @@
...
+    # - account_object (bootstrap)
+    # - code_object (bootstrap)
+    # - generated_transaction_object
+    # - global_property_object
+    # - key_value_object (bootstrap)
+    # - protocol_state_object (bootstrap)
+    # - permission_object (bootstrap)
+    # The bootstrap process has created account_object and code_object (by uploading the bios contract),
+    # key_value_object (token creation), protocol_state_object (preactivation feature), and permission_object
+    # (automatically taken care by the automatically generated eosio account)
```
heifner: This comment has nothing to do with this test.

tests/light_validation_sync_test.py
```diff
@@ -0,0 +1,112 @@
...
+    trx = {
+        "actions": [{"account": "payloadless", "name": "doit", "authorization": [{
+          "actor": "payloadless", "permission": "active"}], "data": ""}],
+        "context_free_actions": [{"account": "payloadless", "name": "doit", "data": ""}],
+        "context_free_data": ["a1b2c3", "1a2b3c"],
+    } 
+
+    cmd = "push transaction '{}' -p payloadless".format(json.dumps(trx))
+    trans = producerNode.processCleosCmd(cmd, cmd, silentErrors=False)
+    assert trans
```
heifner: Other python tests use `errorExit` shouldn't that be used instead of `assert`?

huangminghuang: I followed the example from node_chainbase_allocation_test.py which also uses assert. If you like more message to the assertion, I can use the expression "assert condition, message".  The problem of using errorExit() is that you need to use explicit if statement to check the condition which make it more verbose.

heifner: A message to the assert I think would be helpful.

PR #8859, opened 2020-03-24
Fix spurious HTTP related test failure [2.0]

b1bart:
Do not enforce an `http-max-response-time-ms` for the ignition wallet in `testnet.template`.  These responses are not large and sometimes burps happen


b1bart:
#8862 indicates that a `-1` here is not a good choice

PR #8861, opened 2020-03-24
HTTP Plugin async APIs [2.0]

b1bart:
## Change Description

**This is a Draft PR while I make sure the API is usable in an integrating system with a parallel PR**

### Async HTTP API Registration

This PR introduces the concept of an "Async" API in the `http-plugin`.  Internal systems providing API handlers that do not need to execute on the primary `appbase` thread for safety reasons can register as "async" and the handlers will be called from the HTTP thread pool instead of the main application thread.  

As before, passing a `(code, response)` to the continuation will schedule response JSON-ification for the HTTP thread pool regardless of what thread the handler was called from so that the systems can feel free to asynchronously call that continuation.

The only other difference between the two types of dispatching from the integrating system point of view is that presently the HTTP thread pool does not have the concept of `priority` so, async APIs do not need to register their handlers with an explicit `priority`.

This PR does not include migrating any potentially async-able APIs to the new dispatch system as each will need to be carefully evaluated for thread-safety.  However, the Trace API has told me it has its eye on this. 

### Clean-up

This PR also includes a fair amount of refactoring to clean up http request handling.  A few things to note are:
 - Pushed `priority` for regular handlers into the handler rather than in a `std::pair`. 
 - created `internal_url_handler` which is explicit about the contents of the `url_handler` map's value
 - Created abstractions to manage different connection types inside of the non-templated `url_handlers`. 
 - Created `in_flight<T>` abstraction to manage `bytes_in_flight` with RAII. 
 - Flattened some of the nesting spam by making methods which generate lambdas.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -127,6 +127,48 @@ namespace eosio {
           static const long timeout_open_handshake = 0;
       };
 #endif
+      /**
+       * virtualized wrapper for the various underlying connection functions needed in req/resp processng
+       */
+      struct abstract_conn {
+         virtual ~abstract_conn() {}
+         virtual bool verify_max_bytes_in_flight() = 0;
+         virtual void send_http_response() = 0;
+         virtual void handle_exception() = 0;
+      };
+
+      using abstract_conn_ptr = std::shared_ptr<abstract_conn>;
+
+      template<typename T>
+      using connection_ptr = typename websocketpp::server<T>::connection_ptr;
+
+      /**
+       * internal url handler that contains more parameters than the handlers provided by external systems
+       */
+      using internal_url_handler = std::function<void(abstract_conn_ptr, string, string, url_response_callback)>;
+
+      /**
+       * Helper method to calculate the "in flight" size of a string
+       * @param s - the string
+       * @return in flight size of s
+       */
+      static size_t in_flight_sizeof( const string& s ) {
+         return s.size();
+      }
+
+      /**
+       * Helper method to calculate the "in flight" size of a fc::variant
+       * This is an estimate based on fc::raw::pack if that process can be successfully executed
+       *
+       * @param v - the fc::variant
+       * @return in flight size of v
+       */
+      static size_t in_flight_sizeof( const fc::variant& v ) {
+         try {
+            return fc::raw::pack_size( v );
+         } catch(...) {}
+         return 0;
+      }
    }
 
    using websocket_server_type = websocketpp::server<detail::asio_with_stub_log<websocketpp::transport::asio::basic_socket::endpoint>>;
@@ -141,7 +183,7 @@ namespace eosio {
    class http_plugin_impl {
       public:
          // key -> priority, url_handler
-         map<string,std::pair<int,url_handler>>  url_handlers;
+         map<string,detail::internal_url_handler>  url_handlers;
          optional<tcp::endpoint>  listen_endpoint;
          string                   access_control_allow_origin;
          string                   access_control_allow_headers;
@@ -227,7 +269,7 @@ namespace eosio {
          }
 
          template<class T>
-         static void handle_exception(typename websocketpp::server<T>::connection_ptr con) {
+         static void handle_exception(detail::connection_ptr<T> con) {
             string err = "Internal Service error, http: ";
             const auto deadline = fc::time_point::now() + fc::exception::format_time_limit;
             try {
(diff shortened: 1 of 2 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -287,13 +328,201 @@ namespace eosio {
                error_results results{websocketpp::http::status_code::too_many_requests, "Busy", error_results::error_info()};
                con->set_body( fc::json::to_string( results, fc::time_point::maximum() ));
                con->set_status( websocketpp::http::status_code::too_many_requests );
+               con->send_http_response();
```
heifner: `send_http_response` can only be called if `defer_http_response` has first been called.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -287,13 +328,201 @@ namespace eosio {
...
+               return object;
+            }
+
+            /**
+             * mutable accessor (can be moved frmo)
+             * @return mutable reference to the contained object
+             */
+            T& operator* () {
+               return object;
+            }
```
heifner: What is the point of these `operator*` ? You can just `std::move` the object without them which you actually do.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -287,13 +328,201 @@ namespace eosio {
...
+               // post  back to an HTTP thread to to allow the response handler to be called from any thread
+               boost::asio::post( thread_pool->get_executor(), [this, con, code, tracked_response=std::move(tracked_response)]() {
+                  try {
+                     std::string json = fc::json::to_string( *tracked_response, fc::time_point::now() + max_response_time );
+                     auto tracked_json = make_in_flight(std::move(json), *this);
+                     con->set_body( std::move( *tracked_json ) );
+                     con->set_status( websocketpp::http::status_code::value( code ) );
+                     con->send_http_response();
+                  } catch( ... ) {
+                     handle_exception<T>( con );
```
heifner: Use `con->handle_exception()`

plugins/http_plugin/http_plugin.cpp
heifner: It looks like it is safe to call `defer_http_response` more than once, so you call it before calling `send_http_response`

plugins/http_plugin/http_plugin.cpp
```diff
@@ -287,13 +328,201 @@ namespace eosio {
...
+          */
+         detail::internal_url_handler make_app_thread_url_handler( int priority, url_handler next ) {
+            return [this, priority, next=std::move(next)]( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) mutable {
+               auto tracked_b = make_in_flight(std::move(b), *this);
+               if (!conn->verify_max_bytes_in_flight()) {
+                  return;
+               }
+
+               // post to the app thread capturing the body_scope
+               app().post( priority, [next=std::move(next), conn=std::move(conn), r=std::move(r), tracked_b=std::move(tracked_b), then=std::move(then)]() mutable {
```
heifner: Do not move `next` as it is captured by this lambda that is used multiple times. Also add comment why it is not being moved and update existing comment as it is not clear what is meant by `body_scope`.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -319,62 +549,15 @@ namespace eosio {
                }
 
                con->append_header( "Content-type", "application/json" );
+               con->defer_http_response();
```
heifner: With this set here, you need to make sure all exits from function call `send_http_response`. If no url_handler found it is not called nor is it called on exception.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -287,13 +328,202 @@ namespace eosio {
...
+          * return to the http thread pool for response processing
+          *
+          * @pre b.size() has been added to bytes_in_flight by caller
+          * @param priority - priority to post to the app thread at
+          * @param next - the next handler for responses
+          * @return the constructed internal_url_handler
+          */
+         detail::internal_url_handler make_app_thread_url_handler( int priority, url_handler next ) {
+            auto next_ptr = std::make_shared<url_handler>(std::move(next));
+            return [this, priority, next_ptr]( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) mutable {
```
heifner: Minor thing: you could `std::move(next_ptr)` here, but this is only called once on startup so no big deal.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -463,8 +463,9 @@ namespace eosio {
                   return;
                }
 
-               // post to the app thread capturing the body_scope
-               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), tracked_b=std::move(tracked_b), then=std::move(then)]() mutable {
+               // post to the app thread taking shared ownership of next (via std::shared_ptr),
+               // sole ownership of the tracked body and the passed in parameters
+               app().post( priority, [next_ptr=std::move(next_ptr), conn=std::move(conn), r=std::move(r), tracked_b=std::move(tracked_b), then=std::move(then)]() mutable {
```
heifner: This one can't be a move. The other one could be, but not this one.

b1bart: 🤦‍♂

PR #8865, opened 2020-03-25
Add possibility to run .cicd scripts from different environments

maoueh:
## Change Description

I know that everything under `.cicd` folder is not supported and could change at any moment, and it's for internal usages.

I've used them internally to build the base pinned images as well as the packages and it much more convenient to use your scripts than build mine from scratch.

I understand this could be declined, but I hope you will consider them as with those small changes, it's possible to use them pretty much as-is (if you know what you do).

#### With FORCE_LINUX=true, you can run the Docker version in `build.sh` and `package.sh` scripts

When running the scripts from Mac OS X, it was always picking up the Darwin conditional
paths in the script. `FORCE_LINUX=true` can be used on a Mac OS X environment to run
the scripts.

#### Fix shell expansion in `package.sh`

In the packaging script, there is a Docker pre command that add executable bit permission
on *.sh scripts in the `build/packages` folder. However, without the escaping, the shell
expansion was performed in the `package.sh` script execution directly instead of being
delayed so that the expansion is performed inside the `docker run` call!

I suspect this worked in your CI system because execution of `package.sh` in performed in
such way that filesystem layout of `package.sh` and filesystem layout of `docker run` were
the same and as such, the expanded `chmod 755 file1.sh file2.sh` command was correctly
working inside the container.

The shell expansion via the `*` is now escaped which correctly delays the expansion until
it's in the `docker run` execution environment.

#### Conditional buildkite-agent upload

Now, the upload task is performed only  if BUILDKITE env is set to `true`, so it's possible
to run the script without having to upload the actual package.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/helpers/file-hash.sh
+++ b/.cicd/helpers/file-hash.sh
@@ -4,4 +4,4 @@ set -eo pipefail
 FILE_NAME=$(basename $1 | awk '{split($0,a,/\.(d|s)/); print a[1] }')
 export DETERMINED_HASH=$(sha1sum $1 | awk '{ print $1 }')
 export HASHED_IMAGE_TAG="eos-${FILE_NAME}-${DETERMINED_HASH}"
-export FULL_TAG="eosio/ci:$HASHED_IMAGE_TAG"
+export FULL_TAG="${IMAGE_NAME:-"eosio/ci"}:$HASHED_IMAGE_TAG"
--- a/.cicd/package.sh
+++ b/.cicd/package.sh
@@ -2,12 +2,14 @@
 set -eo pipefail
 . ./.cicd/helpers/general.sh
 mkdir -p $BUILD_DIR
-if [[ $(uname) == 'Darwin' ]]; then
+if [[ $(uname) == 'Darwin' && $FORCE_LINUX != true ]]; then
     bash -c "cd build/packages && chmod 755 ./*.sh && ./generate_package.sh brew"
     ARTIFACT='*.rb;*.tar.gz'
     cd build/packages
     [[ -d x86_64 ]] && cd 'x86_64' # backwards-compatibility with release/1.6.x
-    buildkite-agent artifact upload "./$ARTIFACT" --agent-access-token $BUILDKITE_AGENT_ACCESS_TOKEN
+    if [[ "$BUILDKITE" == 'true' ]]; then
+      buildkite-agent artifact upload "./$ARTIFACT" --agent-access-token $BUILDKITE_AGENT_ACCESS_TOKEN
+    fi
     for A in $(echo $ARTIFACT | tr ';' ' '); do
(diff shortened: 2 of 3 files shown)


.cicd/package.sh
```diff
@@ -2,12 +2,14 @@
...
 . ./.cicd/helpers/general.sh
 mkdir -p $BUILD_DIR
-if [[ $(uname) == 'Darwin' ]]; then
+if [[ $(uname) == 'Darwin' && $FORCE_LINUX != true ]]; then
     bash -c "cd build/packages && chmod 755 ./*.sh && ./generate_package.sh brew"
     ARTIFACT='*.rb;*.tar.gz'
     cd build/packages
     [[ -d x86_64 ]] && cd 'x86_64' # backwards-compatibility with release/1.6.x
-    buildkite-agent artifact upload "./$ARTIFACT" --agent-access-token $BUILDKITE_AGENT_ACCESS_TOKEN
+    if [[ "$BUILDKITE" == 'true' ]]; then
```
maoueh: Other script uses `BUILDKITE` to gate some buildkite features, so I used it here also. I had initially a check for the `buildkite-agent` command directly, I could revert to that if you prefer.

.cicd/helpers/file-hash.sh
```diff
@@ -4,4 +4,4 @@ set -eo pipefail
 FILE_NAME=$(basename $1 | awk '{split($0,a,/\.(d|s)/); print a[1] }')
 export DETERMINED_HASH=$(sha1sum $1 | awk '{ print $1 }')
 export HASHED_IMAGE_TAG="eos-${FILE_NAME}-${DETERMINED_HASH}"
-export FULL_TAG="eosio/ci:$HASHED_IMAGE_TAG"
\ No newline at end of file
+export FULL_TAG="${IMAGE_NAME:-"eosio/ci"}:$HASHED_IMAGE_TAG"
```
maoueh: This enable usage of a different identifier for the base image, so they can be stored in a different distribution system.

.cicd/package.sh
scottarnette: 👍

.cicd/helpers/file-hash.sh
scottarnette: 👍

scottarnette:
Note that the failed checks listed are due to:
1. Executing this on an old pipeline accidentally.
2. A bug in a CI step related to branch names. This has since been fixed on `develop`.

maoueh:
@scottarnette Would you accept a backport of this patch against `release/2.0.x` branch?

heifner:
@maoueh yes

PR #8868, opened 2020-03-25
Feature/new host function system

larryk85:
## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.github/workflows/main.yml
```diff
@@ -6,8 +6,23 @@ jobs:
     name: Start Job
     runs-on: ubuntu-latest
     steps:
-      - name: Start Job
-        run: echo "Start Job."
+      - name: Start Job.
```
spoonincode: what's up with the various build/ci changes?

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -96,7 +96,7 @@ const static uint32_t   setcode_ram_bytes_multiplier       = 10;     ///< multip
 
 const static uint32_t   hashing_checktime_block_size       = 10*1024;  /// call checktime from hashing intrinsic once per this number of bytes
 
-const static eosio::chain::wasm_interface::vm_type default_wasm_runtime = eosio::chain::wasm_interface::vm_type::wabt;
+const static eosio::chain::wasm_interface::vm_type default_wasm_runtime = eosio::chain::wasm_interface::vm_type::eos_vm_jit;
```
spoonincode: Should be based on the availability of EOS VM JIT -- should default to EOS VM when JIT isn't available (ARM builds, etc)

libraries/chain/webassembly/console.cpp
```diff
@@ -0,0 +1,108 @@
...
+   // TODO predicate these for ignore
+   void interface::prints(null_terminated_ptr str) {
+      context.console_append( static_cast<const char*>(str) );
+   }
+
+   void interface::prints_l(legacy_array_ptr<const char> str, uint32_t str_len ) {
+      context.console_append(string(str, str_len));
+   }
+
+   void interface::printi(int64_t val) {
```
spoonincode: where is the check to skip this stuff if console isn't enabled now?

libraries/chain/include/eosio/chain/config.hpp
spoonincode: :man_facepalming: sorry I remember now that neither EOS VM JIT nor EOS VM work on anything but x86_64, so academic for now

.github/workflows/main.yml
swatanabe-b1: I'm not sure what branch this was built of of originally.  I rebased it.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,589 @@
...
+   inline static host_function_registrator<&interface::NAME, alias_check, early_validate_pointers, legacy_static_check_wl_args, context_free_check, ##__VA_ARGS__> NAME ## _registrator = {"env", #NAME};
+
+   class interface {
+      public:
+         interface(apply_context& ctx) : context(ctx) {}
+
+         inline apply_context& get_context() { return context; }
+         inline const apply_context& get_context() const { return context; }
+
+         // checktime api
```
spoonincode, line 44: with wabt gone for good, seems like checktime() and call_depth_assert() can be removed?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+#define REGISTER_CF_HOST_FUNCTION(NAME, ...) \
+   inline static host_function_registrator<&interface::NAME, alias_check, early_validate_pointers, static_check_wl_args, ##__VA_ARGS__> NAME ## _registrator = {"env", #NAME};
+
+#define REGISTER_LEGACY_HOST_FUNCTION(NAME, ...) \
+   inline static host_function_registrator<&interface::NAME, alias_check, early_validate_pointers, legacy_static_check_wl_args, context_aware_check, ##__VA_ARGS__> NAME ## _registrator = {"env", #NAME};
+
+#define REGISTER_LEGACY_CF_HOST_FUNCTION(NAME, ...) \
+   inline static host_function_registrator<&interface::NAME, alias_check, early_validate_pointers, legacy_static_check_wl_args, ##__VA_ARGS__> NAME ## _registrator = {"env", #NAME};
+
+#define REGISTER_LEGACY_CF_ONLY_HOST_FUNCTION(NAME, ...) \
```
arhag, line 34: There is no point in adding an `alias_check` precondition to these three macros to register legacy host functions.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         int64_t set_proposed_producers(legacy_array_ptr<char> packed_producer_schedule);
+         int64_t set_proposed_producers_ex(uint64_t packed_producer_format, legacy_array_ptr<char> packed_producer_schedule);
+         uint32_t get_blockchain_parameters_packed(legacy_array_ptr<char> packed_blockchain_parameters) const;
+         void set_blockchain_parameters_packed(legacy_array_ptr<char> packed_blockchain_parameters);
+         bool is_privileged(account_name account) const;
+         void set_privileged(account_name account, bool is_priv);
+
+         REGISTER_HOST_FUNCTION(is_feature_active, privileged_check);
+         REGISTER_HOST_FUNCTION(activate_feature, privileged_check);
+         REGISTER_HOST_FUNCTION(preactivate_feature, privileged_check);
```
arhag, line 63: This should be legacy. We don't want to support `const digest_type&` as an argument for host functions going forward. 

Also, I'm concerned about the possible changed behavior of references. I think we may need a new `legacy_ref` wrapper type and make this host function's argument type into `legacy_ref<const digest_type>`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+
+         // system api
+         uint64_t current_time() const;
+         uint64_t publication_time() const;
+         bool is_feature_activated(const digest_type&) const;
+         name get_sender() const;
+
+         REGISTER_HOST_FUNCTION(current_time);
+         REGISTER_HOST_FUNCTION(publication_time);
+         REGISTER_HOST_FUNCTION(is_feature_activated);
```
arhag, line 251: This should be legacy. We don't want to support `const digest_type&` as an argument for host functions going forward.

Also, I'm concerned about the possible changed behavior of references. I think we may need a new `legacy_ref` wrapper type and make this host function's argument type into `legacy_ref<const digest_type>`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         void printsf(float);
+         void printdf(double);
+         void printqf(const float128_t&);
+         void printn(name);
+         void printhex(legacy_array_ptr<const char>);
+
+         REGISTER_LEGACY_CF_HOST_FUNCTION(prints);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(prints_l);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printi);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printui);
```
arhag, line 294: Why are `printi` and `printui` legacy?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_CF_HOST_FUNCTION(printi128);
+         REGISTER_CF_HOST_FUNCTION(printui128);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printsf);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printdf);
```
arhag, line 298: Why are `printsf` and `printdf` legacy?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_CF_HOST_FUNCTION(printqf);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printn);
```
arhag, line 300: Why is `printn` legacy?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         int32_t db_next_i64(int32_t, uint64_t&);
+         int32_t db_previous_i64(int32_t, legacy_ptr<uint64_t>);
+         int32_t db_find_i64(uint64_t, uint64_t, uint64_t, uint64_t);
+         int32_t db_lowerbound_i64(uint64_t, uint64_t, uint64_t, uint64_t);
+         int32_t db_upperbound_i64(uint64_t, uint64_t, uint64_t, uint64_t);
+         int32_t db_end_i64(uint64_t, uint64_t, uint64_t);
+
+         REGISTER_LEGACY_HOST_FUNCTION(db_store_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_update_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_remove_i64);
```
arhag, line 318: Why is `db_remove_i64` legacy?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_LEGACY_HOST_FUNCTION(db_get_i64);
+         REGISTER_HOST_FUNCTION(db_next_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_previous_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_find_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_lowerbound_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_upperbound_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_end_i64);
```
arhag, line 325: Why are `db_find_i64`, `db_lowerbound_i64`, `db_upperbound_i64`, and `db_end_i64` legacy?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         int32_t db_find_i64(uint64_t, uint64_t, uint64_t, uint64_t);
+         int32_t db_lowerbound_i64(uint64_t, uint64_t, uint64_t, uint64_t);
+         int32_t db_upperbound_i64(uint64_t, uint64_t, uint64_t, uint64_t);
+         int32_t db_end_i64(uint64_t, uint64_t, uint64_t);
+
+         REGISTER_LEGACY_HOST_FUNCTION(db_store_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_update_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_remove_i64);
+         REGISTER_LEGACY_HOST_FUNCTION(db_get_i64);
+         REGISTER_HOST_FUNCTION(db_next_i64);
```
arhag, line 320: I'm concerned that references like `uint64_t&` may not be treated the same under the new system as they were under the old system. There may be a need for a `legacy_ref` wrapper type.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         void printqf(const float128_t&);
+         void printn(name);
+         void printhex(legacy_array_ptr<const char>);
+
+         REGISTER_LEGACY_CF_HOST_FUNCTION(prints);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(prints_l);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printi);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printui);
+         REGISTER_CF_HOST_FUNCTION(printi128);
+         REGISTER_CF_HOST_FUNCTION(printui128);
```
arhag, line 296: Again, I think we will need to treat the reference types in `printi128` and `printui128` differently.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printsf);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(printdf);
+         REGISTER_CF_HOST_FUNCTION(printqf);
```
arhag, line 299: Again, I think we will need to treat the reference type in `printqf` differently.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         int32_t db_idx64_previous(int32_t, uint64_t&);
+
+         REGISTER_HOST_FUNCTION(db_idx64_store);
+         REGISTER_HOST_FUNCTION(db_idx64_update);
+         REGISTER_LEGACY_HOST_FUNCTION(db_idx64_remove);
+         REGISTER_LEGACY_HOST_FUNCTION(db_idx64_find_secondary);
+         REGISTER_HOST_FUNCTION(db_idx64_find_primary);
+         REGISTER_LEGACY_HOST_FUNCTION(db_idx64_lowerbound);
+         REGISTER_LEGACY_HOST_FUNCTION(db_idx64_upperbound);
+         REGISTER_LEGACY_HOST_FUNCTION(db_idx64_end);
```
arhag, line 347: Why is `db_idx64_end` legacy?

Similarly for the other db end host functions below.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_HOST_FUNCTION(db_idx64_next);
+         REGISTER_HOST_FUNCTION(db_idx64_previous);
```
arhag, line 349: Again, concerns about reference types apply to the relevant functions above.

They also apply to the other db host functions below.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_CF_HOST_FUNCTION(__fixdfti);
+         REGISTER_CF_HOST_FUNCTION(__fixunssfti);
+         REGISTER_CF_HOST_FUNCTION(__fixunsdfti);
+         REGISTER_CF_HOST_FUNCTION(__floatsidf);
+         REGISTER_CF_HOST_FUNCTION(__floatsitf);
+         REGISTER_CF_HOST_FUNCTION(__floatditf);
+         REGISTER_CF_HOST_FUNCTION(__floatunsitf);
+         REGISTER_CF_HOST_FUNCTION(__floatunditf);
+         REGISTER_CF_HOST_FUNCTION(__floattidf);
+         REGISTER_CF_HOST_FUNCTION(__floatuntidf);
```
arhag, line 560: Again, concerns about reference types apply to most of the compiler builtin host functions above.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_LEGACY_CF_HOST_FUNCTION(memcpy);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(memmove);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(memcmp);
+         REGISTER_LEGACY_CF_HOST_FUNCTION(memset);
+
+         // transaction api
+         void send_inline(legacy_array_ptr<char>);
+         void send_context_free_inline(legacy_array_ptr<char>);
+         void send_deferred(legacy_ptr<const uint128_t>, account_name, legacy_array_ptr<char>, uint32_t);
+         bool cancel_deferred(legacy_ptr<const uint128_t>);
```
arhag, line 459: The first argument of `send_deferred` and `cancel_deferred` used to be a a `const uint128_t&` but is now a pointer to a `const uint128_t`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         // context free api
+         int32_t get_context_free_data(uint32_t index, legacy_array_ptr<char> buffer) const;
+         REGISTER_LEGACY_CF_ONLY_HOST_FUNCTION(get_context_free_data)
+
+         // privileged api
+         int32_t is_feature_active(int64_t feature_name) const;
+         void activate_feature(int64_t feature_name) const;
+         void preactivate_feature(const digest_type&);
+         void set_resource_limits(account_name account, int64_t ram_bytes, int64_t net_weight, int64_t cpu_weight);
+         void get_resource_limits(account_name account, legacy_ptr<int64_t, 8> ram_bytes, legacy_ptr<int64_t, 8> net_weight, legacy_ptr<int64_t, 8> cpu_weight) const;
```
arhag, line 53: These last three arguments used to be `int64_t&` but are pointers to `int64_t`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         REGISTER_INJECTED_HOST_FUNCTION(_eosio_i64_to_f64);
+         REGISTER_INJECTED_HOST_FUNCTION(_eosio_ui32_to_f64);
+         REGISTER_INJECTED_HOST_FUNCTION(_eosio_ui64_to_f64);
+
+         // producer api
+         int32_t get_active_producers(legacy_array_ptr<account_name>) const;
+         REGISTER_LEGACY_HOST_FUNCTION(get_active_producers);
+
+         // crypto api
+         void assert_recover_key(legacy_ptr<const fc::sha256>, legacy_array_ptr<char>, legacy_array_ptr<char>) const;
```
arhag, line 197: First argument type changed from `const fc::sha256&` to pointer to `const fc::sha256`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         int32_t recover_key(legacy_ptr<const fc::sha256>, legacy_array_ptr<char>, legacy_array_ptr<char>) const;
```
arhag, line 198: First argument type changed from `const fc::sha256&` to pointer to `const fc::sha256`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -0,0 +1,581 @@
...
+         void assert_sha256(legacy_array_ptr<char>, legacy_ptr<const fc::sha256>) const;
+         void assert_sha1(legacy_array_ptr<char>, legacy_ptr<const fc::sha1>) const;
+         void assert_sha512(legacy_array_ptr<char>, legacy_ptr<const fc::sha512>) const;
+         void assert_ripemd160(legacy_array_ptr<char>, legacy_ptr<const fc::ripemd160>) const;
+         void sha256(legacy_array_ptr<char>, legacy_ptr<fc::sha256>) const;
+         void sha1(legacy_array_ptr<char>, legacy_ptr<fc::sha1>) const;
+         void sha512(legacy_array_ptr<char>, legacy_ptr<fc::sha512>) const;
+         void ripemd160(legacy_array_ptr<char>, legacy_ptr<fc::ripemd160>) const;
```
arhag, line 206: Last argument of the above 8 host functions changed from reference to pointer.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
swatanabe-b1: This doesn't change the behavior.  I can just rename legacy_ptr to legacy_ref.

libraries/chain/include/eosio/chain/wasm_interface.hpp
```diff
@@ -20,63 +20,13 @@ namespace eosio { namespace chain {
...
-                      const string& export_name,
-                      IR::ObjectType type,
-                      Runtime::ObjectInstance*& out) override
-         { try {
-            bool fail = false;
-
-            if( whitelisted_intrinsics != nullptr ) {
-               // Protect access to "private" injected functions; so for now just simply allow "env" since injected
-               // functions are in a different module.
-               EOS_ASSERT( mod_name == "env", wasm_exception,
```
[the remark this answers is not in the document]
swatanabe-b1: - If the name is not whitelisted, it will be rejected by the whitelist check
- If it is whitelisted, it will be rejected by the link check because the imports available in "env" are disjoint from the imports available in eosio_injected_module_name.

I'll add an explicit check for env in eos-vm.cpp to make it more future proof.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+            if constexpr (is_reference_proxy_type_v<arg_t>) {
+               if constexpr (is_span_type_v<dependent_type_t<arg_t>>) {
+                  using dep_t = dependent_type_t<arg_t>;
+                  const dep_t& s = arg;
+                  EOS_ASSERT( s.size() <= std::numeric_limits<wasm_size_t>::max() / (wasm_size_t)sizeof(dependent_type_t<dep_t>),
+                        wasm_execution_error, "length will overflow" );
+                  volatile auto check = *(reinterpret_cast<const char*>(arg.original_ptr) + s.size_bytes() - 1);
+                  ignore_unused_variable_warning(check);
+               } else {
+                  EOS_ASSERT(arg.original_ptr != ctx.get_interface().get_memory(), wasm_execution_error, "references cannot be created for null pointers");
```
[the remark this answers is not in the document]
swatanabe-b1: Even though there was a specialization to handle pointer arguments in the old intrinsic system, no intrinsics actually used it.  All legacy_ptr's correspond to what were reference arguments, for which this behavior is correct.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+               }
+            } else if constexpr (vm::is_reference_type_v<arg_t>) {
```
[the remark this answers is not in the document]
swatanabe-b1: Correct.  This branch is obsolete.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+   struct remove_reference_proxy {
+      using type = T;
+   };
+   template<typename T, std::size_t A>
+   struct remove_reference_proxy<vm::reference_proxy<T, A>> {
+      using type = T;
+   };
+   template<typename T>
+   struct remove_reference_proxy<vm::reference<T>> {
+      using type = T;
```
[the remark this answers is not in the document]
swatanabe-b1: I'm not sure what a reference<reference_proxy<>> would mean.  It's definitely not used now.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+                                       std::is_same_v<decltype(is_unvalidated_ptr(std::declval<T>())), std::true_type> ||
+                                       std::is_same_v<name, T> ||
+                                       std::is_arithmetic_v<T>;
+      };
+      template <typename T>
+      struct is_whitelisted_type {
+         static constexpr bool value = std::is_arithmetic_v<std::decay_t<T>> ||
+                                       std::is_same_v<std::decay_t<T>, name> ||
+                                       std::is_pointer_v<T>    ||
+                                       std::is_lvalue_reference_v<T> ||
```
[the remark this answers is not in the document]
swatanabe-b1: Instead of arithmetic types, I would make it integral types and softfloat types.

libraries/chain/include/eosio/chain/webassembly/common.hpp
```diff
@@ -2,98 +2,115 @@
...
 
 using namespace fc;
 
+namespace eosio { namespace chain { namespace webassembly {
+   // forward declaration
+   class interface;
+}}} // ns eosio::chain::webassembly
+
+// TODO need to fix up wasm_tests to not use this macro
+#define EOSIO_INJECTED_MODULE_NAME "eosio.injection"
```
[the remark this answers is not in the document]
swatanabe-b1: No idea.  It doesn't really matter, so I can just change it back.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+                       "${code} does not have permission to call this API", ("code", ctx.get_host().get_context().get_receiver()));
+         }));
+
+   EOS_VM_PRECONDITION(alias_check,
+         EOS_VM_INVOKE_ON_ALL(([&](auto&& arg, auto&&... rest) {
+            using namespace eosio::vm;
+            using arg_t = std::decay_t<decltype(arg)>;
+            if constexpr (is_span_type_v<arg_t>) {
+               // check alignment while we are here
+               EOS_ASSERT( reinterpret_cast<std::uintptr_t>(arg.data()) % alignof(dependent_type_t<arg_t>) == 0,
```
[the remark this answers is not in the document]
swatanabe-b1: As far as I know all compilers optimize division by a constant.  There are quite a few tricks beyond just powers of 2, as well.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+   inline static constexpr bool are_whitelisted_types_v = true; //(... && detail::is_whitelisted_type<Ts>::value);
+
+   template <typename... Ts>
+   inline static constexpr bool are_whitelisted_legacy_types_v = (... && detail::is_whitelisted_legacy_type<Ts>::value);
+
+   template <typename T, typename U>
+   inline static bool is_aliasing(const T& a, const U& b) {
+      std::uintptr_t a_ui = reinterpret_cast<std::uintptr_t>(a.data());
+      std::uintptr_t b_ui = reinterpret_cast<std::uintptr_t>(b.data());
+      return a_ui < b_ui ? a_ui + a.size() >= b_ui : b_ui + b.size() >= a_ui;
```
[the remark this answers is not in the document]
swatanabe-b1: This will also return true if they are adjacent, but not overlapping.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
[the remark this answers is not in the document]
arhag: So the conclusion we reached is that the `early_validate_pointers` precondition can be replaced by a `null_pointer_check` precondition that validates for any `legacy_ptr` that it does point to the start of the WASM linear memory (to maintain same behavior will all the reference arguments in the old intrinsic system).

Furthermore, we don't care to impose such an unnecessary null check on new host functions going forward that want to be able to modify the values (stored in WASM linear memory) passed to them. Rather than using references to handle such cases (for example out variables), we want to enforce new host functions to be explicit about it by using a pointer directly (even though the pointer will always be pointing to a valid region in WASM linear memory). And for these new pointers we don't want to impose validation like in `null_pointer_check`.

This does mean that some of the functions mentioned below which are registered as non-legacy but still use a reference need to be converted into using a `legacy_ptr` instead and be registered using the legacy version of the macros.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+                                       std::is_same_v<decltype(is_legacy_array_ptr(std::declval<T>())), std::true_type> ||
+                                       std::is_same_v<decltype(is_unvalidated_ptr(std::declval<T>())), std::true_type> ||
+                                       std::is_same_v<name, T> ||
+                                       std::is_arithmetic_v<T>;
+      };
+      template <typename T>
+      struct is_whitelisted_type {
+         static constexpr bool value = std::is_arithmetic_v<std::decay_t<T>> ||
+                                       std::is_same_v<std::decay_t<T>, name> ||
+                                       std::is_pointer_v<T>    ||
```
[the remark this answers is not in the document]
arhag: Following from the conclusion [here](https://github.com/EOSIO/eos/pull/8868#discussion_r411549794), pointers are allowed as a non-legacy type, but references should now be banned from use in non-legacy host functions.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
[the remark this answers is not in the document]
arhag: Similarly, a correction to this following the conclusion [here](https://github.com/EOSIO/eos/pull/8868#discussion_r411549794) is that we should ban all references. Furthermore, pointers are allowed but they should be only to integral types or softfloat types.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -0,0 +1,159 @@
...
+                                       std::is_lvalue_reference_v<T> ||
+                                       std::is_same_v<std::decay_t<T>, float128_t> ||
+                                       eosio::vm::is_span_type_v<std::decay_t<T>>;
```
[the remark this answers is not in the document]
arhag: Or more precisely: spans of integral types or of softfloat types.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
arhag: No need for `legacy_ref`. This argument would type would instead be `legacy_ptr<const digest_type>`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
arhag: No need for `legacy_ref`. This argument would type would instead be `legacy_ptr<const digest_type>`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
arhag: No need for `legacy_ref`. Just need to be consistent about using `legacy_ptr<uint64_t>` like in `db_previous_i64` rather than using `uint64_t&`.

libraries/chain/webassembly/console.cpp
```diff
@@ -39,7 +39,7 @@ namespace eosio { namespace chain { namespace webassembly {
       });
    }
 
-   void interface::printi128(const __int128& val) {
+   void interface::printi128(legacy_ptr<const __int128> val) {
 		predicated_print(context,
       [&]() {
 			bool is_negative = (val < 0);
```
arhag, line 45: The fact that no code changes were needed here despite the change in function signature is a little concerning. I don't like that a `reference_proxy` type can implicitly convert to both a pointer and a reference. This is why I suggest removing the implicit conversion to references from `reference_proxy` (which I suggest renaming to `proxy_pointer`).

libraries/chain/webassembly/console.cpp
```diff
@@ -62,39 +62,39 @@ namespace eosio { namespace chain { namespace webassembly {
       });
    }
 
-   void interface::printui128(const unsigned __int128& val) {
+   void interface::printui128(legacy_ptr<const unsigned __int128> val) {
 		predicated_print(context,
       [&]() {
 			fc::uint128_t v(val>>64, static_cast<uint64_t>(val) );
```
arhag, line 68: Again, this is concerning. What is actually being cast into a `uint64_t` here after the change to the function signature? If we just treat `val` as a pointer (rather maybe a pointer or maybe a reference), it would be very clear that we must write `fc::uint128_t v(*val >> 64, static_cast<uint64_t>(*val) );` here.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -315,61 +315,61 @@ namespace eosio { namespace chain { namespace webassembly {
...
-         int32_t db_idx64_store(uint64_t, uint64_t, uint64_t, uint64_t, const uint64_t&);
-         void db_idx64_update(int32_t, uint64_t, const uint64_t&);
+         int32_t db_idx64_store(uint64_t, uint64_t, uint64_t, uint64_t, legacy_ptr<const uint64_t>);
+         void db_idx64_update(int32_t, uint64_t, legacy_ptr<const uint64_t>);
          void db_idx64_remove(int32_t);
          int32_t db_idx64_find_secondary(uint64_t, uint64_t, uint64_t, legacy_ptr<const uint64_t>, legacy_ptr<uint64_t>);
-         int32_t db_idx64_find_primary(uint64_t, uint64_t, uint64_t, uint64_t&, uint64_t);
+         int32_t db_idx64_find_primary(uint64_t, uint64_t, uint64_t, legacy_ptr<uint64_t>, uint64_t);
          int32_t db_idx64_lowerbound(uint64_t, uint64_t, uint64_t, legacy_ptr<uint64_t, 8>, legacy_ptr<uint64_t, 8>);
          int32_t db_idx64_upperbound(uint64_t, uint64_t, uint64_t, legacy_ptr<uint64_t, 8>, legacy_ptr<uint64_t, 8>);
```
arhag, line 335: Explicit alignments are redundant.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -315,61 +315,61 @@ namespace eosio { namespace chain { namespace webassembly {
...
-         int32_t db_idx128_store(uint64_t, uint64_t, uint64_t, uint64_t, const uint128_t&);
-         void db_idx128_update(int32_t, uint64_t, const uint128_t&);
+         int32_t db_idx128_store(uint64_t, uint64_t, uint64_t, uint64_t, legacy_ptr<const uint128_t>);
+         void db_idx128_update(int32_t, uint64_t, legacy_ptr<const uint128_t>);
          void db_idx128_remove(int32_t);
          int32_t db_idx128_find_secondary(uint64_t, uint64_t, uint64_t, legacy_ptr<const uint128_t>, legacy_ptr<uint64_t>);
-         int32_t db_idx128_find_primary(uint64_t, uint64_t, uint64_t, uint128_t&, uint64_t);
+         int32_t db_idx128_find_primary(uint64_t, uint64_t, uint64_t, legacy_ptr<uint128_t>, uint64_t);
          int32_t db_idx128_lowerbound(uint64_t, uint64_t, uint64_t, legacy_ptr<uint128_t, 16>, legacy_ptr<uint64_t, 8>);
          int32_t db_idx128_upperbound(uint64_t, uint64_t, uint64_t, legacy_ptr<uint128_t, 16>, legacy_ptr<uint64_t, 8>);
```
arhag, line 358: Explicit alignments are redundant.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -380,66 +380,66 @@ namespace eosio { namespace chain { namespace webassembly {
          int32_t db_idx256_lowerbound(uint64_t, uint64_t, uint64_t, legacy_array_ptr<uint128_t, 16>, legacy_ptr<uint64_t, 8>);
          int32_t db_idx256_upperbound(uint64_t, uint64_t, uint64_t, legacy_array_ptr<uint128_t, 16>, legacy_ptr<uint64_t, 8>);
```
arhag, line 381: Explicit alignments are redundant.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -380,66 +380,66 @@ namespace eosio { namespace chain { namespace webassembly {
...
-         int32_t db_idx_double_store(uint64_t, uint64_t, uint64_t, uint64_t, const float64_t&);
-         void db_idx_double_update(int32_t, uint64_t, const float64_t&);
+         int32_t db_idx_double_store(uint64_t, uint64_t, uint64_t, uint64_t, legacy_ptr<const float64_t>);
+         void db_idx_double_update(int32_t, uint64_t, legacy_ptr<const float64_t>);
          void db_idx_double_remove(int32_t);
          int32_t db_idx_double_find_secondary(uint64_t, uint64_t, uint64_t, legacy_ptr<const float64_t>, legacy_ptr<uint64_t>);
-         int32_t db_idx_double_find_primary(uint64_t, uint64_t, uint64_t, float64_t&, uint64_t);
+         int32_t db_idx_double_find_primary(uint64_t, uint64_t, uint64_t, legacy_ptr<float64_t>, uint64_t);
          int32_t db_idx_double_lowerbound(uint64_t, uint64_t, uint64_t, legacy_ptr<float64_t, 8>, legacy_ptr<uint64_t, 8>);
          int32_t db_idx_double_upperbound(uint64_t, uint64_t, uint64_t, legacy_ptr<float64_t, 8>, legacy_ptr<uint64_t, 8>);
```
arhag, line 404: Explicit alignments are redundant.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -380,66 +380,66 @@ namespace eosio { namespace chain { namespace webassembly {
...
-         int32_t db_idx_long_double_store(uint64_t, uint64_t, uint64_t, uint64_t, const float128_t&);
-         void db_idx_long_double_update(int32_t, uint64_t, const float128_t&);
+         int32_t db_idx_long_double_store(uint64_t, uint64_t, uint64_t, uint64_t, legacy_ptr<const float128_t>);
+         void db_idx_long_double_update(int32_t, uint64_t, legacy_ptr<const float128_t>);
          void db_idx_long_double_remove(int32_t);
          int32_t db_idx_long_double_find_secondary(uint64_t, uint64_t, uint64_t, legacy_ptr<const float128_t>, legacy_ptr<uint64_t>);
-         int32_t db_idx_long_double_find_primary(uint64_t, uint64_t, uint64_t, float128_t&, uint64_t);
+         int32_t db_idx_long_double_find_primary(uint64_t, uint64_t, uint64_t, legacy_ptr<float128_t>, uint64_t);
          int32_t db_idx_long_double_lowerbound(uint64_t, uint64_t, uint64_t, legacy_ptr<float128_t, 8>, legacy_ptr<uint64_t, 8>);
          int32_t db_idx_long_double_upperbound(uint64_t, uint64_t, uint64_t, legacy_ptr<float128_t, 8>,  legacy_ptr<uint64_t, 8>);
```
arhag, line 428: Explicit alignments are redundant.

libraries/chain/include/eosio/chain/webassembly/common.hpp
```diff
@@ -104,6 +104,8 @@ namespace eosio { namespace chain {
       }
       EOS_VM_FROM_WASM(name, (uint64_t e)) { return name{e}; }
       uint64_t to_wasm(name&& n) { return n.to_uint64_t(); }
+      EOS_VM_FROM_WASM(float32_t, (float f)) { return ::to_softfloat32(f); }
+      EOS_VM_FROM_WASM(float64_t, (double f)) { return ::to_softfloat64(f); }
```
arhag, line 108: I think this (and its conversion in the other direction which is missing here) should be moved into EOS VM (`as_value` and `as_result`). Also, `as_result` should be a lot more strict. It should only allow the return types: `int32_t`, `uint32_t`, `int64_t`, `uint64_t`, `float32_t`, `float64_t`, and `void*`. This will make it symmetric with what `as_value` allows (note that it would also need to be updated to take a `float32_t` rather than a `float` and a `float64_t` rather than a `double`).

larryk85: I don’t agree with this argument. Eos-vm is not eosio. I don’t want to enforce limitation on the base implementation unless it absolutely makes sense to do so. Most implementations that are going use eos-vm are not going to use softfloat (I would assume) and shouldn’t have the data types restricted at the eos-vm layer. Hardening of which types are useable is a per project decision.

larryk85: I don’t agree with this argument. Eos-vm is not eosio. I don’t want to enforce limitation on the base implementation unless it absolutely makes sense to do so. Most implementations that are going use eos-vm are not going to use softfloat (I would assume) and shouldn’t have the data types restricted at the eos-vm layer. Hardening of which types are useable is a per project decision.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -71,21 +84,12 @@ namespace eosio { namespace chain { namespace webassembly {
...
-                  using dep_t = dependent_type_t<arg_t>;
-                  const dep_t& s = arg;
-                  EOS_ASSERT( s.size() <= std::numeric_limits<wasm_size_t>::max() / (wasm_size_t)sizeof(dependent_type_t<dep_t>),
-                        wasm_execution_error, "length will overflow" );
-                  volatile auto check = *(reinterpret_cast<const char*>(arg.original_ptr) + s.size_bytes() - 1);
-                  ignore_unused_variable_warning(check);
-               } else {
-                  EOS_ASSERT(arg.original_ptr != ctx.get_interface().get_memory(), wasm_execution_error, "references cannot be created for null pointers");
-               }
+            if constexpr (decltype(detail::is_legacy_ptr(std::declval<arg_t>()))::value && !is_span_type_v<dependent_type_t<arg_t>>) {
```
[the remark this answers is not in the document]
larryk85: I don’t agree with this one fully either. The reason it is in eos-vm is because we already have multiple targets that will utilize that type of copying back functionality. If we move it out specifically for eos, then we need to create and maintain separate ones for the other projects. The reference_proxy is not just for the legacy stuff, it’s also for pointers and references. If you’d like a separate pointer_proxy in eos, that is fine. I don’t see any harm in that.

libraries/chain/include/eosio/chain/webassembly/common.hpp
arhag: EOS VM still allows adding support for any return types desired by adding new `from_wasm` functions to a derived type converter class.

Requiring the user to harden the return types (rather than starting with a minimal set and allowing the user to expand as necessary) is an alternative approach. But that will require a lot more changes because preconditions do not have access to the return type.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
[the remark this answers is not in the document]
arhag: Good point on others benefiting from an array copying wrapper. 

How about the following?

```cpp
template <typename T, std::size_t LegacyAlign=0>
struct reference_proxy;
```

becomes

```cpp
template <typename T, std::size_t Extent = dynamic_extent, std::size_t LegacyAlign=0>
struct proxy_span;
```

where we can again, for now, only support `Extent == dynamic_extent` or `Extent == 1`. 

A `proxy_span<T, 1, LegacyAlign>` would be a specialization of the template that has a similar structure and behavior as `reference_proxy<T, LegacyAlign>` where `T` is not a `span`. EOS VM would know to expand this out as a `void*` within the host function that WASM code calls.

 A `proxy_span<T, dynamic_extent, LegacyAlign>` would behave like a `reference_proxy<span<T>, LegacyAlign>`. EOS VM would know to expand this out as `void*, wasm_size_t` withn the host function that WASM code calls.

The `proxy_span` would only be able to implicitly convert to the appropriate `span` type (no implicit conversion to a pointer or to a reference) and it would not have a `ref()` member function.

In EOSIO, `legacy_ptr<T>` would be a type alias to `eosio::vm::proxy_span<T, 1, alignof(T)>` and `legacy_array_ptr<T>` (which I still think should be renamed to `legacy_span`) would be a type alias to `eosio::vm::proxy_span<T, dynamic_extent, alignof(T)>`.

And then the condition in the `if constexpr` here gets to be simplified to just `decltype(detail::is_legacy_ptr(std::declval<arg_t>()))::value`.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
swatanabe-b1: I made it explicit to avoid depending on implementation defined behavior.  I only used explicit alignment for the intrinsics where the alignment is observable from wasm.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -105,6 +109,13 @@ namespace eosio { namespace chain { namespace webassembly {
                        "${code} does not have permission to call this API", ("code", ctx.get_host().get_context().get_receiver()));
          }));
 
+   namespace detail {
+      template<typename T>
+      vm::span<const T> to_span(const vm::reference_proxy<T>& val) { return &val.ref(); }
```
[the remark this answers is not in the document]
swatanabe-b1: I thought I fixed that.  There's also another problem: there's no const overload of get right now.

libraries/chain/include/eosio/chain/webassembly/common.hpp
swatanabe-b1: Intrinsics that return pointers and references are relatively rare.  They must point into linear memory, and can therefore only really return their arguments safely.

Using void* as the return type is not a significant burden:
  - Any intrinsic that returns a reference can be changed to return a pointer with some adjustments.
  - Any intrinsic that returns a pointer can trivially be changed to return [const] void*.
  - If using typed pointers or references as results is needed for some reason, they can easily be implemented in terms of void*.

Therefore, I do not believe that there is a strong reason to support T* and T& in the base implementation.

T* and T& are not supported for as_value.  For the sake of consistency, they should not be supported in as_result either.

In addition, for an extensible system, I generally favor keeping the base implementation minimal.

Overall, I think the balance is in favor of no support for T* and T&.

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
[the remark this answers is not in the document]
swatanabe-b1: Currently `reference_proxy<T*>` and `reference_proxy<T&>` are both equivalent to `reference_proxy<T>`, which is strange.  I think we need a total of three templates.  reference_proxy, pointer_proxy, and span_proxy.  We can either make them separate templates or three specializations of reference_proxy (and leave the primary template undefined).

libraries/chain/include/eosio/chain/webassembly/preconditions.hpp
```diff
@@ -111,7 +111,7 @@ namespace eosio { namespace chain { namespace webassembly {
 
    namespace detail {
       template<typename T>
-      vm::span<const T> to_span(const vm::reference_proxy<T>& val) { return &val.ref(); }
+      vm::span<const T> to_span(const vm::argument_proxy<T*>& val) { return {val.get(), 1}; }
```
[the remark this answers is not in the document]
swatanabe-b1: argument_proxy currently propagates const on accessors.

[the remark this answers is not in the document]
swatanabe-b1: The pointer returned by get() may point to the internal storage.

PR #8880, opened 2020-03-25
add pruned_block to signed_block conversion

huangminghuang:
## Change Description
Add pruned_block to signed_block conversion.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/block.cpp
+++ b/libraries/chain/block.cpp
@@ -1,4 +1,5 @@
 #include <eosio/chain/block.hpp>
+#include <eosio/chain/types.hpp>
 
 namespace eosio { namespace chain {
    void additional_block_signatures_extension::reflector_init() {
@@ -111,4 +112,22 @@ namespace eosio { namespace chain {
       return validate_and_extract_block_extensions( block_extensions );
    }
 
+   signed_block_ptr pruned_block::to_signed_block() const {
+      if (prune_state != prune_state_type::complete_legacy)
+         return signed_block_ptr{};
+
+      auto result = std::make_shared<signed_block>(*this);
+      result->block_extensions = this->block_extensions;
+
+      auto visitor = overloaded{
+         [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
+         [](const pruned_transaction &trx) -> transaction_receipt::trx_type { 
+            return packed_transaction{trx.get_signed_transaction(), trx.get_compression()}; 
+      }};
+
+      for (const pruned_transaction_receipt &r : transactions){
+         result->transactions.emplace_back(transaction_receipt{r, r.trx.visit(visitor)});
+      }
+      return result;
+   }
 } } /// namespace eosio::chain
--- a/libraries/chain/block.cpp
+++ b/libraries/chain/block.cpp
@@ -122,9 +122,7 @@ namespace eosio { namespace chain {
       auto visitor = overloaded{
           [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
           [](const pruned_transaction &trx) -> transaction_receipt::trx_type {
-             auto& prunable_data = trx.get_prunable_data().prunable_data;
-             EOS_ASSERT(prunable_data.contains<prunable_transaction_data::full_legacy>(), tx_not_full_legacy, "pruned_transaction does not contain the full legacy data");
-             auto& legacy = prunable_data.get<prunable_transaction_data::full_legacy>();
+             const auto& legacy = trx.get_prunable_data().prunable_data.get<prunable_transaction_data::full_legacy>();
              return packed_transaction(trx.get_packed_transaction(), legacy.signatures, legacy.packed_context_free_data, trx.get_compression());
           }};
 
--- a/libraries/chain/block.cpp
+++ b/libraries/chain/block.cpp
@@ -116,14 +116,17 @@ namespace eosio { namespace chain {
       if (prune_state != prune_state_type::complete_legacy)
          return signed_block_ptr{};
 
-      auto result = std::make_shared<signed_block>(*this);
+      auto result = std::make_shared<signed_block>(*static_cast<const signed_block_header*>(this));
       result->block_extensions = this->block_extensions;
 
       auto visitor = overloaded{
-         [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
-         [](const pruned_transaction &trx) -> transaction_receipt::trx_type { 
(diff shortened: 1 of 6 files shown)


libraries/chain/block.cpp
```diff
@@ -111,4 +112,22 @@ namespace eosio { namespace chain {
       return validate_and_extract_block_extensions( block_extensions );
    }
 
+   signed_block_ptr pruned_block::to_signed_block() const {
+      if (prune_state != prune_state_type::complete_legacy)
+         return signed_block_ptr{};
+
+      auto result = std::make_shared<signed_block>(*this);
```
arhag: I think it would be clearer about what was going on here if you were explicit about the cast to the base class:
```cpp
auto result = std::make_shared<signed_block>(*static_cast<const signed_block_header*>(this));
```

libraries/chain/block.cpp
```diff
@@ -111,4 +112,22 @@ namespace eosio { namespace chain {
...
+      result->block_extensions = this->block_extensions;
+
+      auto visitor = overloaded{
+         [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
+         [](const pruned_transaction &trx) -> transaction_receipt::trx_type { 
+            return packed_transaction{trx.get_signed_transaction(), trx.get_compression()};
```
arhag: This would recompress and possibly give different a different `packed_trx` or `packed_context_free_data` than the original. We need to get the original packed forms of those, which should be possible for all `pruned_transaction` transactions within the block since we know that this block is `complete_legacy`.

Note that this will mean adding a way to access the `packed_context_free_data` within a `pruned_transaction` when it contains `full_legacy`.

libraries/chain/block.cpp
heifner: I would make this `auto result = std::make_shared<signed_block>(static_cast<const signed_block_header&>(*this));` so the slice is obvious.

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -31,12 +31,13 @@ namespace eosio { namespace chain {
    };
 
    struct transaction_receipt : public transaction_receipt_header {
-
-      transaction_receipt():transaction_receipt_header(){}
+      using trx_type = fc::static_variant<transaction_id_type, packed_transaction>;
+      transaction_receipt() : transaction_receipt_header() {}
+      transaction_receipt(const transaction_receipt_header& header, trx_type&& t): transaction_receipt_header(header), trx(t){}
```
arhag: `trx(std::move(t))`

libraries/chain/block.cpp
```diff
@@ -111,4 +112,22 @@ namespace eosio { namespace chain {
...
+
+      auto visitor = overloaded{
+         [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
+         [](const pruned_transaction &trx) -> transaction_receipt::trx_type { 
+            return packed_transaction{trx.get_signed_transaction(), trx.get_compression()}; 
+      }};
+
+      for (const pruned_transaction_receipt &r : transactions){
+         result->transactions.emplace_back(transaction_receipt{r, r.trx.visit(visitor)});
+      }
```
heifner: Can we add a `to_transaction_receipt` to `pruned_transaction_receipt` for this.

heifner: .. and a `to_packed_transaction` to `pruned_transaction`

libraries/chain/include/eosio/chain/block.hpp
heifner: Should be `trx( std::move( t ) )`

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -161,6 +162,8 @@ namespace eosio { namespace chain {
       }
 
       flat_multimap<uint16_t, block_extension> validate_and_extract_extensions()const;
+
+      signed_block_ptr to_signed_block() const;
```
heifner: I would prefer this to be right under `clone()`

libraries/chain/block.cpp
```diff
@@ -111,4 +112,25 @@ namespace eosio { namespace chain {
...
+         return signed_block_ptr{};
+
+      auto result = std::make_shared<signed_block>(*static_cast<const signed_block_header*>(this));
+      result->block_extensions = this->block_extensions;
+
+      auto visitor = overloaded{
+          [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
+          [](const pruned_transaction &trx) -> transaction_receipt::trx_type {
+             auto& prunable_data = trx.get_prunable_data().prunable_data;
+             EOS_ASSERT(prunable_data.contains<prunable_transaction_data::full_legacy>(), tx_not_full_legacy, "pruned_transaction does not contain the full legacy data");
```
arhag: There is no need for this assertion or the new `tx_not_full_legacy` exception. Assuming the invariants of the `pruned_block` have not been violated, then this assertion should never fail because you already checked that `prune_state` must be `prune_state_type::complete_legacy` by this point. And even if there was a programming error that caused this invariant to fail to hold, the `get` function on the `static_variant` would throw an exception anyway.

libraries/chain/block.cpp
```diff
@@ -111,4 +112,25 @@ namespace eosio { namespace chain {
...
+      if (prune_state != prune_state_type::complete_legacy)
+         return signed_block_ptr{};
+
+      auto result = std::make_shared<signed_block>(*static_cast<const signed_block_header*>(this));
+      result->block_extensions = this->block_extensions;
+
+      auto visitor = overloaded{
+          [](const transaction_id_type &id) -> transaction_receipt::trx_type { return id; },
+          [](const pruned_transaction &trx) -> transaction_receipt::trx_type {
+             auto& prunable_data = trx.get_prunable_data().prunable_data;
```
arhag: Make this `const auto&` for clarity. Similar with `legacy` below.

libraries/chain/block.cpp
arhag: `to_transaction_receipt` would have to return an `optional<transaction_receipt>` in that case (or throw an exception if called on a `pruned_transaction_receipt` that contained a `pruned_transaction` with prunable data that was not `full_legacy`). By only implementing `to_signed_block` as a public member function, we can handle the case where the `pruned_block` contains unconvertible data really cleanly: just return `signed_block_ptr{}`.

Issue #8887, opened 2020-03-26
Nodeos rejects sync request from the other nodeos being started with the same snapshot


allenhan2:
Both nodeoses are started with the same snapshot, the primary was started about 12hours earlier and synced with mainnet, then secondary started with the same snapshot and tried to sync with the primary (w/o syncing with mainnet). The secondary receives the error messages like below:
info  2020-03-26T15:09:09.386 thread-10 net_plugin.cpp:1474           operator()           ] requesting range 111866391 to 111868890, from 127.0.0.1:10076
info  2020-03-26T15:09:09.387 thread-10 net_plugin.cpp:1588           recv_handshake       ] handshake from 127.0.0.1:10066 - ab3770b, lib 112197911, head 112198241, head id 377e2341a4acb9b8.. sync 1
warn  2020-03-26T15:09:09.387 thread-10 net_plugin.cpp:2663           handle_message       ] ["127.0.0.1:10066 - ab3770b" 127.0.0.1:10076]received go_away_message, reason = some other non-fatal condition, possibly unknown block
error 2020-03-26T15:09:09.387 thread-10 net_plugin.cpp:1455           request_next_chunk   ] Unable to continue syncing at this time

Branch is "release/2.0.x", (pulled the latest commit, ffa8190b5391985d7f9fc00dfd5a1803db74f069)
The primary nodoes is a release build,  the same error msg for both release and debug version of secondary nodeos.
snapshot :
snapshot-06aaf216455bd0b4090c7e8211f4de09e0a2cd9238ef3ad2b0d66bfc1d930d38.bin on https://snapshots-main.eossweden.org/

The syncing is fine after the secondary uses a newer version of snapshot.

heifner:
Lets add an integration test that tests this.

heifner:
Fixed in #8889

kobrien:
Currently deploying from 2.0.6 and I think I might be seeing this error?

`info  2020-07-10T12:52:58.379 net-1     net_plugin.cpp:3090           connection_monitor   ] p2p client connections: 0/25, peer connections: 16/16
info  2020-07-10T12:52:58.379 net-0     net_plugin.cpp:2141           operator()           ] host: mars.peer.eosbixin.com port: 643
info  2020-07-10T12:52:58.592 net-1     net_plugin.cpp:1070           operator()           ] Sending handshake generation 1 to mars.peer.eosbixin.com:643, lib 128732200, head 128732530, id 2301b7baaad3d567
[...]

PR #8889, opened 2020-03-27
Sync from snapshot - 2.0

heifner:
## Change Description

- Fix for case where a second node is started from the same snapshot as the node being synced from. Delay check for unknown block until block is actually requested.
- Resolves https://github.com/EOSIO/eos/issues/8887

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1185,7 +1185,7 @@ namespace eosio {
          signed_block_ptr sb;
          try {
             sb = cc.fetch_block_by_number( num );
-         } catch(...) {}
+         } FC_LOG_AND_DROP();
          if( sb ) {
             c->strand.post( [c, sb{std::move(sb)}]() {
                c->enqueue_block( sb, true );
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1168,7 +1182,10 @@ namespace eosio {
          connection_ptr c = weak.lock();
          if( !c ) return;
          controller& cc = my_impl->chain_plug->chain();
-         signed_block_ptr sb = cc.fetch_block_by_number( num );
+         signed_block_ptr sb;
+         try {
+            sb = cc.fetch_block_by_number( num );
+         } catch(...) {}
```
brianjohnson5972: Would be good to capture reason this failed (at least other than if it is because it is not available in block log) for the ilog below.

heifner: Yue Han had the same comment I think which I "ignored", so two against one, I'll add something.

PR #8891, opened 2020-03-27
Added block log v4

huangminghuang:
## Change Description
This PR create fro block log version 4 which changes each block entry from the serialization of signed_block to the tuple of offset to the next entry, compression_status and pruned_block.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

Some member functions uses signed_block_ptr are marked as deprecated. The usages of those should be replaced by their counterparts with pruned_block_ptr in the future. 

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -7,6 +7,7 @@
 #include <boost/iostreams/device/mapped_file.hpp>
 #include <boost/filesystem.hpp>
 #include <variant>
+#include <algorithm>
 
 #define LOG_READ  (std::ios::in | std::ios::binary)
 #define LOG_WRITE (std::ios::out | std::ios::binary | std::ios::app)
@@ -53,10 +54,14 @@ namespace eosio { namespace chain {
          uint32_t offset = 0;
       };
 
+      size_t get_stream_pos(fc::cfile_datastream& ds) { return ds.tellp(); }
+      size_t get_stream_pos(std::ifstream& ds) { return ds.tellg();  }
+      void   skip_streamm_pos(fc::cfile_datastream& ds, size_t n) { ds.skip(n);  }
+      void   skip_streamm_pos(std::ifstream& ds, size_t n) { ds.seekg(n, ds.cur);  }
 
       template <typename Stream>
       void unpack(Stream& ds, log_entry_v4& entry){
-         auto start_pos = ds.tellp();
+         auto start_pos = get_stream_pos(ds);
          fc::raw::unpack(ds, entry.offset);
          uint8_t compression;
          fc::raw::unpack(ds, compression);
@@ -64,8 +69,12 @@ namespace eosio { namespace chain {
          EOS_ASSERT(entry.compression == packed_transaction::cf_compression_type::none, block_log_exception,
                   "Only support compression_type none");
          fc::raw::unpack(ds, static_cast<signed_block&>(entry));
-         EOS_ASSERT(ds.tellp() - start_pos + sizeof(uint64_t) == entry.offset , block_log_exception,
-                  "Invalid block log entry offset");
+         uint64_t current_stream_offset = get_stream_pos(ds) - start_pos;
+
+         int64_t bytes_to_skip = static_cast<int64_t>(entry.offset) - sizeof(uint64_t) - current_stream_offset;
+         EOS_ASSERT(bytes_to_skip >= 0, block_log_exception,
+                    "Invalid block log entry offset");
+         skip_streamm_pos(ds, bytes_to_skip);
       }
 
       std::vector<char> pack(const signed_block& block, packed_transaction::cf_compression_type compression) {
@@ -648,7 +657,7 @@ namespace eosio { namespace chain {
 
       ilog( "Reconstructing '${new_block_log}' from backed up block log", ("new_block_log", block_log_path) );
 
-      std::fstream  old_block_stream;
+      std::ifstream  old_block_stream;
       std::fstream  new_block_stream;
 
       old_block_stream.open( (backup_dir / "blocks.log").generic_string().c_str(), LOG_READ );
@@ -893,9 +902,14 @@ namespace eosio { namespace chain {
          }
 
          if (pruned) {
-            my->block_file.seek(pos);
-            std::vector<char> buffer = detail::pack(entry);
-            EOS_ASSERT(buffer.size() <= entry.offset, block_log_exception, "Not enough space reserved in block log entry to serialize pruned block.");
+            // we don't want to rewrite entire entry, just the block data itself.
+            const auto block_offset = detail::offset_to_block_start(my->version);
+            my->block_file.seek(pos + block_offset);
+            const uint32_t max_block_size = entry.offset - block_offset - sizeof(uint64_t);
+            std::vector<char> buffer(max_block_size);
+            fc::datastream<char*> stream(buffer.data(), buffer.size());
+            static_cast<signed_block&>(entry).pack(stream, entry.compression);
+            EOS_ASSERT(buffer.size() <= max_block_size, block_log_exception, "Not enough space reserved in block log entry to serialize pruned block.");
             my->block_file.write(buffer.data(), buffer.size());
             my->block_file.flush();
          }
@@ -1120,6 +1134,7 @@ namespace eosio { namespace chain {
       FILE* ind_in = nullptr;                            //C style files for reading blocks.log and blocks.index
       //we use low level file IO because it is distinctly faster than C++ filebuf or iostream
       uint64_t first_block_pos = 0;                      //file position in blocks.log for the first block in the log
+      uint64_t block_file_size = 0;
       chain_id_type chain_id;
    };
 
@@ -1276,6 +1291,7 @@ namespace eosio { namespace chain {
          const auto expected_totem = block_log::npos;
          std::decay_t<decltype(block_log::npos)> actual_totem;
          size = fread ( (char*)&actual_totem, sizeof(actual_totem), 1, blk_in);
+         block_file_size = fc::file_size(block_file_name);
 
          EOS_ASSERT(size == 1, block_log_exception,
                     "Expected to read ${size} bytes, but did not read any bytes", ("size", sizeof(actual_totem)));
@@ -1337,6 +1353,24 @@ namespace eosio { namespace chain {
       auto size = fread((void*)&block_n_pos, sizeof(block_n_pos), 1, ind_in);                   //filepos of block n
       EOS_ASSERT( size == 1, block_log_exception, "cannot read ${file} entry for block ${b}", ("file", index_file_name.string())("b",n) );
 
+      if (version >= 4) {
+         uint64_t block_n_plus_1_pos = block_file_size;
+         if (n < last_block) {
+            size = fread((void*)&block_n_plus_1_pos, sizeof(block_n_plus_1_pos), 1, ind_in);
+            EOS_ASSERT( size == 1, block_log_exception, "cannot read ${file} entry for block ${b}", ("file", index_file_name.string())("b",n+1) );
+         }
+
+         uint32_t entry_offset = 0;
+         status = fseek(blk_in, block_n_pos, SEEK_SET);
+         EOS_ASSERT( status == 0, block_log_exception, "cannot seek to ${file} ${pos} from beginning of file for block ${b}", ("file", block_file_name.string())("pos", block_n_pos)("b",n) );
+         size = fread((void*)&entry_offset, sizeof(entry_offset), 1, blk_in);
+         EOS_ASSERT( size == 1, block_log_exception, "cannot read ${file} entry for block ${b}", ("file", block_file_name.string())("b",n) );
+
+         EOS_ASSERT((entry_offset == (block_n_plus_1_pos - block_n_pos)), block_log_exception, 
+            "The entry offset in block ${n} is ${entry_offset} does not match the difference of block positions in index file", 
+               ("n", n)("entry_offset", entry_offset)("pos_diff", block_n_plus_1_pos - block_n_pos));
+      }
+
       //read blocks.log and verify block number n is found at the determined file position
       const auto calc_blknum_pos = block_n_pos + detail::block_log_impl::blknum_offset_from_block_entry(version);
       status = fseek(blk_in, calc_blknum_pos, SEEK_SET);
@@ -1376,7 +1410,7 @@ namespace eosio { namespace chain {
       return 0;
    }
 
-   void block_log::smoke_test(fc::path block_dir) {
+   void block_log::smoke_test(fc::path block_dir, uint32_t interval) {
       trim_data td(block_dir);
       auto status = fseek(td.blk_in, -sizeof(uint64_t), SEEK_END);             //get last_block from blocks.log, compare to from blocks.index
       EOS_ASSERT( status == 0, block_log_exception, "cannot seek to ${file} ${pos} from beginning of file", ("file", td.block_file_name.string())("pos", sizeof(uint64_t)) );
@@ -1392,10 +1426,10 @@ namespace eosio { namespace chain {
       bnum = fc::endian_reverse_u32(bnum) + 1;                       //convert from big endian to little endian and add 1
       EOS_ASSERT( td.last_block == bnum, block_log_exception, "blocks.log says last block is ${lb} which disagrees with blocks.index", ("lb", bnum) );
       ilog("blocks.log and blocks.index agree on number of blocks");
-      uint32_t delta = (td.last_block + 8 - td.first_block) >> 3;
-      if (delta < 1)
-         delta = 1;
-      for (uint32_t n = td.first_block; ; n += delta) {
+      if (interval == 0) {
+         interval = std::max((td.last_block + 8 - td.first_block) >> 3, 1U);
+      }
+      for (uint32_t n = td.first_block; ; n += interval) {
          if (n > td.last_block)
             n = td.last_block;
          td.block_pos(n);                                 //check block 'n' is where blocks.index says
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -36,7 +36,7 @@ namespace eosio { namespace chain {
     *            compression_status and pruned_block.
     */
 
-   enum Versions {
+   enum versions {
       initial_version = 1,
       block_x_start_version = 2,
       genesis_state_or_chain_id_version = 3,
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -68,7 +68,7 @@ namespace eosio { namespace chain {
          entry.compression = static_cast<packed_transaction::cf_compression_type>(compression);
          EOS_ASSERT(entry.compression == packed_transaction::cf_compression_type::none, block_log_exception,
                   "Only support compression_type none");
-         fc::raw::unpack(ds, static_cast<signed_block&>(entry));
+         static_cast<signed_block&>(entry).unpack(ds, entry.compression);
          uint64_t current_stream_offset = get_stream_pos(ds) - start_pos;
 
          int64_t bytes_to_skip = static_cast<int64_t>(entry.offset) - sizeof(uint64_t) - current_stream_offset;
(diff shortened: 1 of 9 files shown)


libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
...
+      //    2. An uint32_t indicating the compression status for the serialization of the pruned_block following this.
+      //    3. The serialization of a pruned_block representation of the block for the entry including padding.
+
+      std::size_t padded_size = b.maximum_pruned_pack_size(segment_compression);
+      std::vector<char> buffer(padded_size + 2* sizeof(uint32_t));
+      fc::datastream<char*> stream(buffer.data(), buffer.size());
+
+      uint32_t offset      = buffer.size();
+      uint32_t compression = static_cast<uint32_t>(segment_compression);
+      stream.write((char*)&offset, sizeof(offset));
```
swatanabe-b1: offset here doesn't take into account its own size or the size of compression.

libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
...
+   }
+
+   uint64_t detail::block_log_impl::write_log_entry(const pruned_block& b, pruned_transaction::cf_compression_type segment_compression) {
+      uint64_t pos = block_file.tellp();
+      std::vector<char> buffer;
+     
+      if (version >= 4)  {
+         buffer = detail::block_log_impl::pack(b, segment_compression);
+      } else {
+         auto block_ptr = b.to_signed_block();
```
arhag: I think it is unfortunate that we need to do a costly conversion to a `signed_block` (and the associated heap allocations) to write out the old blocks to disk. This performance hit isn't as critical as the reads (which can be done multiple times per block rather than just once), but it would still be nice if we did better here. Of course we could do a custom serialization as if it was a signed_block, but I suppose it isn't important to work on right now (and possibly ever because we will encourage people to switch to version 4 block logs). Still, I think we should at least add a `TODO` message here to remind us to go back and improve the performance of this.

libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
       }
    }
 
+   std::vector<char>  detail::block_log_impl::pack(const pruned_block& b, pruned_transaction::cf_compression_type segment_compression) {
+
+      // In version 4 of the irreversible blocks log format, these log entries consists of the following in order:
+      //    1. An uint32_t offset from the start of this log entry to the start of the next log entry.
+      //    2. An uint32_t indicating the compression status for the serialization of the pruned_block following this.
```
arhag: Use a `fc::unsigned_int` to represent the compression status, not `uint32_t`.

libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
...
+      //    3. The serialization of a pruned_block representation of the block for the entry including padding.
+
+      std::size_t padded_size = b.maximum_pruned_pack_size(segment_compression);
+      std::vector<char> buffer(padded_size + 2* sizeof(uint32_t));
```
arhag: This would be adding the `sizeof(uint32_t)` once and then also the `pack_size` of the `segment_compression` value as a `fc::unsigned_int` (currently should always be 1 but don't hard code).

libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
...
+      }
+      block_file.write(buffer.data(), buffer.size());
+      block_file.write((char*)&pos, sizeof(pos));
+      index_file.write((char*)&pos, sizeof(pos));
+      return pos;
+   }
+
+   void detail::block_log_impl::unpack(fc::cfile_datastream& ds, pruned_block& block) { 
+      uint32_t len;
+      uint32_t compression;
```
arhag: Again this would be an `fc::unsigned_int`.

libraries/chain/block_log.cpp
```diff
@@ -388,27 +451,44 @@ namespace eosio { namespace chain {
...
-      fc::raw::unpack(ds, *result);
-      return result;
+      return my->read_log_entry();
    }
 
-   void block_log::read_block_header(block_header& bh, uint64_t pos)const {
+   void block_log::read_block_header(block_header& bh, uint64_t pos) const {
       my->check_open_files();
 
+      if (my->version == 4)
```
arhag: Should be `>=`.

libraries/chain/block_log.cpp
```diff
@@ -388,27 +451,44 @@ namespace eosio { namespace chain {
...
+         pos += 2 * sizeof(uint32_t);
```
arhag: Instead of just skipping over the appropriate amount, actually unpack the length and the compression status. Then assert that the compression status is `compression_type::none`, because this implementation doesn't yet support compressed block log entries yet.

libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
...
+
+   uint64_t detail::block_log_impl::write_log_entry(const pruned_block& b, pruned_transaction::cf_compression_type segment_compression) {
+      uint64_t pos = block_file.tellp();
+      std::vector<char> buffer;
+     
+      if (version >= 4)  {
+         buffer = detail::block_log_impl::pack(b, segment_compression);
+      } else {
+         auto block_ptr = b.to_signed_block();
+         EOS_ASSERT(block_ptr, block_log_append_fail, "Unable to convert block to legacy format");
```
arhag: Also assert that `segment_compression` is `none` when `version < 4`, because old versions of block logs do not support compressed log entries.

libraries/chain/block_log.cpp
```diff
@@ -286,28 +296,76 @@ namespace eosio { namespace chain {
...
+      block_file.write((char*)&pos, sizeof(pos));
+      index_file.write((char*)&pos, sizeof(pos));
+      return pos;
+   }
+
+   void detail::block_log_impl::unpack(fc::cfile_datastream& ds, pruned_block& block) { 
+      uint32_t len;
+      uint32_t compression;
+      fc::raw::unpack(ds, len);
+      fc::raw::unpack(ds, compression);
```
arhag: Assert that `compression` is `none` because other compression types aren't supported yet.

libraries/chain/block_log.cpp
```diff
@@ -834,6 +857,56 @@ namespace eosio { namespace chain {
...
+
+         auto entry = my->read_v4_log_entry(pos);
+
+         EOS_ASSERT(entry.block->block_num() == block_num, block_log_exception,
+                     "Wrong block was read from block log.");
+
+         auto pruner = overloaded{[](transaction_id_type&) { return false; },
+                                  [&id](pruned_transaction& ptx) { return ptx.id() == id && prune(ptx); }};
+
+         for (auto trx : entry.block->transactions) {
```
arhag, line 880: `auto&`

libraries/chain/block_log.cpp
```diff
@@ -834,6 +857,56 @@ namespace eosio { namespace chain {
...
+            if (trx.trx.visit(pruner)) {
+               my->block_file.seek(pos);
+               std::vector<char> buffer = detail::block_log_impl::pack(*entry.block, static_cast<pruned_transaction::cf_compression_type>(entry.compression.value));
+               my->block_file.write(buffer.data(), buffer.size());
+               my->block_file.flush();
+               return;
+            }
+         }
+
+         EOS_ASSERT(true, block_log_exception,
```
arhag, line 890: I think you meant to write `false` here.  Also, don't use that pattern. Instead use `EOS_THROW`.

arhag: Better yet return a boolean from this function rather than throwing.

libraries/chain/block_log.cpp
```diff
@@ -39,16 +39,21 @@ namespace eosio { namespace chain {
    namespace detail {
       using unique_file = std::unique_ptr<FILE, decltype(&fclose)>;
 
+      struct log_entry {
+         uint32_t         offset;
+         fc::unsigned_int compression;
```
arhag, line 44: Make this a `pruned_transaction::cf_compression_type` type.

libraries/chain/block_log.cpp
```diff
@@ -834,6 +857,56 @@ namespace eosio { namespace chain {
...
+                     "Wrong block was read from block log.");
+
+         auto pruner = overloaded{[](transaction_id_type&) { return false; },
+                                  [&id](pruned_transaction& ptx) { return ptx.id() == id && prune(ptx); }};
+
+         for (auto trx : entry.block->transactions) {
+            if (trx.trx.visit(pruner)) {
+               my->block_file.seek(pos);
+               std::vector<char> buffer = detail::block_log_impl::pack(*entry.block, static_cast<pruned_transaction::cf_compression_type>(entry.compression.value));
+               my->block_file.write(buffer.data(), buffer.size());
```
arhag, line 884: Assert that `buffer.size() <= entry.offset`.

libraries/chain/block_log.cpp
huangminghuang: auto& doesn't work. This is an initialization statement, not an assignment statement.

swatanabe-b1: I don't follow.  auto& does work in initialization and does not work in assignment.

huangminghuang: Never mind, I was looking at the wrong line. I thought you were talking the auto& pruner.

heifner: We discussed. He thought you meant line 877, not line 880.

libraries/chain/block.cpp
```diff
@@ -21,16 +21,27 @@ namespace eosio { namespace chain {
...
+   };
 
    pruned_transaction_receipt::pruned_transaction_receipt(const transaction_receipt& other, bool legacy)
      : transaction_receipt_header(static_cast<const transaction_receipt_header&>(other)),
-       trx(other.trx.visit([&](const auto& obj) { return translate_transaction_receipt(obj, legacy); }))
+       trx(other.trx.visit(transaction_receipt_translator{legacy}))
+   {}
+
+   pruned_transaction_receipt::pruned_transaction_receipt(transaction_receipt&& other, bool legacy)
+     : transaction_receipt_header(static_cast<const transaction_receipt_header&>(other)),
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/block.cpp
```diff
@@ -88,6 +99,16 @@ namespace eosio { namespace chain {
       }
    }
 
+    pruned_block::pruned_block( signed_block&& other, bool legacy )
+     : signed_block_header(std::move(static_cast<const signed_block_header&>(other))),
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -206,7 +208,7 @@ namespace eosio { namespace chain {
    };
 
    struct prunable_transaction_data {
-      enum class compression_type {
+      enum class compression_type : uint8_t {
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -251,6 +253,11 @@ namespace eosio { namespace chain {
       std::size_t maximum_pruned_pack_size( compression_type segment_compression ) const;
 
       prunable_data_type  prunable_data;
+
+      static prunable_transaction_data make(bool legacy, const signed_transaction& t,
+                                            packed_transaction::compression_type _compression);
+      static prunable_transaction_data make(bool legacy, const packed_transaction& t);
+      static prunable_transaction_data make(bool legacy, packed_transaction&& t);
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/block.cpp
```diff
@@ -21,16 +21,27 @@ namespace eosio { namespace chain {
...
+      fc::static_variant<transaction_id_type, pruned_transaction> operator()(const transaction_id_type& tid) const {
+         return tid;
+      }
+      fc::static_variant<transaction_id_type, pruned_transaction> operator()(const packed_transaction& ptrx) const {
+         return pruned_transaction(ptrx, legacy);
+      }
+      fc::static_variant<transaction_id_type, pruned_transaction> operator()(packed_transaction&& ptrx) const {
+         return pruned_transaction(std::move(ptrx), legacy);
+      }
+   };
```
heifner, line 35: Added to #8900

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -151,7 +152,7 @@ namespace eosio { namespace chain {
       // Returns the maximum_pruned_padded_size.  It is the caller's responsibility to
       // reserve enough space after the end if in-place pruning is desired.
       template<typename Stream>
-      std::size_t pack(Stream& stream, pruned_transaction::cf_compression_type segment_compression) {
+      std::size_t pack(Stream& stream, pruned_transaction::cf_compression_type segment_compression) const {
          std::size_t padded_size = maximum_pruned_pack_size( segment_compression );
```
heifner, line 156: block.hpp changes added to #8900

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -194,10 +194,12 @@ namespace eosio { namespace chain {
...
-      vector<signature_type>                  signatures;
-      fc::enum_type<uint8_t,compression_type> compression;
-      bytes                                   packed_context_free_data;
-      bytes                                   packed_trx;
+     friend struct pruned_transaction;
+     friend struct prunable_transaction_data;
+     vector<signature_type>                   signatures;
+     fc::enum_type<uint8_t, compression_type> compression;
+     bytes                                    packed_context_free_data;
+     bytes                                    packed_trx;
```
heifner, line 203: Added to #8900

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -263,8 +270,10 @@ namespace eosio { namespace chain {
       pruned_transaction& operator=(const pruned_transaction&) = delete;
       pruned_transaction& operator=(pruned_transaction&&) = default;
 
-      pruned_transaction(const packed_transaction& other, bool legacy) : pruned_transaction(other.get_signed_transaction(), legacy, other.get_compression()) {}
-      explicit pruned_transaction(const signed_transaction& t, bool legacy, compression_type _compression = compression_type::none);
+      pruned_transaction(const packed_transaction& other, bool legacy);
+      pruned_transaction(packed_transaction&& other, bool legacy);
+      explicit pruned_transaction(const signed_transaction& t, bool legacy,
+                                  compression_type _compression = compression_type::none);
       explicit pruned_transaction(signed_transaction&& t, bool legacy, compression_type _compression = compression_type::none);
```
heifner, line 277: Added to #8900

libraries/chain/transaction.cpp
```diff
@@ -471,25 +471,58 @@ std::size_t prunable_transaction_data::maximum_pruned_pack_size(prunable_transac
    return 1 + prunable_data.visit([&](const auto& t){ return padded_pack_size(t, compression); });
 }
 
-static prunable_transaction_data make_prunable_transaction_data( bool legacy, const signed_transaction& t, pruned_transaction::compression_type _compression ) {
+prunable_transaction_data prunable_transaction_data::make( bool legacy, const signed_transaction& t, packed_transaction::compression_type _compression ) {
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/transaction.cpp
```diff
@@ -471,25 +471,58 @@ std::size_t prunable_transaction_data::maximum_pruned_pack_size(prunable_transac
...
+   } else {
+      return { prunable_transaction_data::full{ t.signatures, t.get_context_free_data() } };
+   }
+}
+
+prunable_transaction_data prunable_transaction_data::make( bool legacy, packed_transaction&& t) {
+   if(legacy) {
+      return { prunable_transaction_data::full_legacy{ std::move(t.signatures), std::move(t.packed_context_free_data) } };
+   } else {
+      return { prunable_transaction_data::full{ std::move(t.signatures), t.get_context_free_data() } };
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/transaction.cpp
```diff
@@ -471,25 +471,58 @@ std::size_t prunable_transaction_data::maximum_pruned_pack_size(prunable_transac
...
+   }
+}
```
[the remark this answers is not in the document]
heifner: Done in #8900

libraries/chain/block_log.cpp
```diff
@@ -300,18 +312,18 @@ namespace eosio { namespace chain {
 
       // In version 4 of the irreversible blocks log format, these log entries consists of the following in order:
       //    1. An uint32_t offset from the start of this log entry to the start of the next log entry.
-      //    2. An uint32_t indicating the compression status for the serialization of the pruned_block following this.
+      //    2. An fc::unsigned_int indicating the compression status for the serialization of the pruned_block following this.
       //    3. The serialization of a pruned_block representation of the block for the entry including padding.
 
       std::size_t padded_size = b.maximum_pruned_pack_size(segment_compression);
-      std::vector<char> buffer(padded_size + 2* sizeof(uint32_t));
+      std::vector<char> buffer(padded_size + offset_to_block_start(4));
```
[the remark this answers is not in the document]
huangminghuang: This is static function, there's no access to version data member here. The function name has been changed to reflect this is for version 4 only.

libraries/chain/block_log.cpp
```diff
@@ -84,24 +97,34 @@ namespace eosio { namespace chain {
...
+                                                       pruned_transaction::cf_compression_type segment_compression);
 
             uint64_t write_log_entry(const pruned_block&                     b,
                                      pruned_transaction::cf_compression_type segment_compression);
 
             void             read_block_header(block_header& bh, uint64_t file_pos);
             pruned_block_ptr read_block(uint64_t pos);
             void             read_head();
-            log_entry        read_v4_log_entry(uint64_t pos);
+            void             read_v4_log_entry(uint64_t pos, log_entry&);
```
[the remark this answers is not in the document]
huangminghuang: I mad the change because all but one unpack log entry usage needs the heap allocated pruned_block. Most of the usage use it in a loop and discard the value every iteration. I consider this is private to block_log.cpp, so the efficiency took precedence.

libraries/chain/include/eosio/chain/block_log.hpp
```diff
@@ -59,11 +56,12 @@ namespace eosio { namespace chain {
           * Return offset of block in file, or block_log::npos if it does not exist.
           */
          uint64_t                       get_block_pos(uint32_t block_num) const;
-         signed_block_header_ptr        read_head() const;
-         const signed_block_header_ptr& head() const;
-         const block_id_type&           head_id() const;
+         signed_block_header*           head() const;
```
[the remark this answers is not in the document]
heifner: Addressed in merge.

libraries/chain/include/eosio/chain/block_log.hpp
```diff
@@ -59,11 +56,12 @@ namespace eosio { namespace chain {
...
+         [[deprecated]] block_id_type   head_id() const; // use head()->id() instead
```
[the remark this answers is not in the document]
heifner: Removed in #8900

libraries/chain/block_log.cpp
```diff
@@ -741,6 +874,51 @@ namespace eosio { namespace chain {
       }));
    }
 
+   bool prune(packed_transaction& ptx) {
+      ptx.prune_all();
+      return true;
+   }
+   
+   void block_log::prune_transactions(uint32_t block_num, const std::vector<transaction_id_type>& ids) {
```
heifner, line 882: I think this should return how many were pruned, so the caller can verify if desired. Also remove the `prune` method above and just add the extra logic to your overloaded lambda.

PR #8927, opened 2020-04-09
Fix SHIP block delay - develop

heifner:
## Change Description

- Fix for `state_history_plugin`  sending blocks out on a one block delay (500ms).
- Use `block_state` of `accepted_block` signal instead of `chain.head_block_num()` as head is not updated at the time of the signal.
- Result `last_irreversible.block_num` & `last_irreversible.block_id` still lags by 500ms even with this change. See #8951.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tbfleming:
Is there any possibility of reentering ship (e.g. a socket event) between on_accepted_block and the head being updated?

PR #8928, opened 2020-04-09
replace boost::bind with std::bind, fixing boost 1.73beta builds

spoonincode:
## Change Description

As of boost 1.73, boost bind no longer places `_1`, `_2`, etc in the global namespace causing build failures for the handful of locations we expect it.

While it's possible to get those back in to the global namespace with a preprocessor definition that is considered depreciated behavior. Besides, in general we've replaced boostisms with standard C++ when appropriate. So, rewrite these as std::bind instead.

Normally I would have desired to change these to
```c++
using namespace std::placeholders;
std::bind(..., _1);
```
However there is a problem with that pre-1.73: it is ambiguous between std::placeholders and boost::placeholders. It is possible to disable these earlier boost bind versions from placing `_1`, `_2`, etc in the global namespace via a preprocessor definition but some other boost libraries (at least unit test) expect those placeholders to be in the global namespace; can't just globally apply said preprocessor definition it would have to be applied more judiciously.

Since we only use the placeholders in a handful of locations, I'm just going to be explicit with `std::placeholders::_1`. It might be a little bit of a trap for the next developer who expects using ... and _1 to simply work though, but as mentioned, we rarely use this.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/authority_checker.hpp
+++ b/libraries/chain/include/eosio/chain/authority_checker.hpp
@@ -173,9 +173,9 @@ namespace detail {
             };
 
             permissions.reserve(authority.waits.size() + authority.keys.size() + authority.accounts.size());
-            std::for_each(authority.accounts.begin(), authority.accounts.end(), boost::bind<void>(emplace_permission, 1, _1));
-            std::for_each(authority.keys.begin(), authority.keys.end(), boost::bind<void>(emplace_permission, 2, _1));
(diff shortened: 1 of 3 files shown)


heifner:
I think I would rather of seen these just converted to lambdas.

spoonincode:
Yeah.. crossed my mind. But I guess I had tunnel vision to make the change as small as possible since probably want to back port to 2.0. Can always change it later.

PR #8929, opened 2020-04-10
Chainlib support for replacing keys

tbfleming:
## Change Description

This adds minimal support to chainlib for replacing keys. The initial use case is in `eosio-tester`, which exposes these new functions. Unlike prior PRs, this PR:
* Doesn't assume the need to replace production keys implies the need to replace owner and active
* Doesn't assume a particular system contract
* Allows replacing owner and active on arbitrary accounts
* Doesn't provide either command-line options or new utilities for accessing this functionality

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3407,6 +3407,38 @@ fc::optional<chain_id_type> controller::extract_chain_id_from_db( const path& st
    return {};
 }
 
+void controller::replace_producer_keys( const public_key_type& key ) {
+   ilog("Replace producer keys with ${k}", ("k", key));
+   mutable_db().modify( db().get<global_property_object>(), [&]( auto& gp ) {
+      gp.proposed_schedule_block_num = {};
+      gp.proposed_schedule.version = 0;
+      gp.proposed_schedule.producers.clear();
+   });
+   my->head->pending_schedule = {};
+   for (auto& prod: my->head->active_schedule.producers ) {
+      ilog("${n}", ("n", prod.producer_name));
+      prod.authority.visit([&](auto &auth) {
+         auth.threshold = 1;
+         auth.keys = {key_weight{key, 1}};
+      });
+   }
+}
+
+void controller::replace_account_keys( name account, const public_key_type& key ) {
+   auto* owner = db().find<permission_object, by_owner>(boost::make_tuple(account, N(owner)));
+   if (owner) {
+      mutable_db().modify(*owner, [&](auto& p) {
+         p.auth = authority(key);
+      });
+   }
+   auto* active = db().find<permission_object, by_owner>(boost::make_tuple(account, N(active)));
+   if (active) {
+      mutable_db().modify(*active, [&](auto& p) {
+         p.auth = authority(key);
+      });
+   }
+}
+
 /// Protocol feature activation handlers:
 
 template<>
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3414,7 +3414,9 @@ void controller::replace_producer_keys( const public_key_type& key ) {
       gp.proposed_schedule.version = 0;
       gp.proposed_schedule.producers.clear();
    });
+   auto version = my->head->pending_schedule.schedule.version;
    my->head->pending_schedule = {};
+   my->head->pending_schedule.schedule.version = version;
    for (auto& prod: my->head->active_schedule.producers ) {
       ilog("${n}", ("n", prod.producer_name));
       prod.authority.visit([&](auto &auth) {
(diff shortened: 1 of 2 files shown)


heifner:
Related: #7413

libraries/chain/controller.cpp
```diff
@@ -3407,6 +3407,38 @@ fc::optional<chain_id_type> controller::extract_chain_id_from_db( const path& st
...
+         auth.threshold = 1;
+         auth.keys = {key_weight{key, 1}};
+      });
+   }
+}
+
+void controller::replace_account_keys( name account, const public_key_type& key ) {
+   auto* owner = db().find<permission_object, by_owner>(boost::make_tuple(account, N(owner)));
+   if (owner) {
+      mutable_db().modify(*owner, [&](auto& p) {
```
arhag, line 3430: The RAM usage for `account` will be out-of-sync and can lead to problems down the line. Please correct the RAM usage using the same calculation done in `apply_eosio_updateauth`. You can bypass `add_ram_usage` and go directly to `resource_limits_manager::add_pending_ram_usage` and then call `resource_limits_manager::verify_account_ram_usage` afterwards to ensure nothing went wrong.  

However, if you want this function to only make the changes if all are possible without violating the invariants (i.e. if the account has sufficient RAM quota). then you will need to ensure ahead of time (before doing any modification to the database in this function) that the account has sufficient RAM quota to support all of the changes (it is possible to have a valid `authority` that takes up less space than a single key one).

libraries/chain/controller.cpp
```diff
@@ -3407,6 +3407,38 @@ fc::optional<chain_id_type> controller::extract_chain_id_from_db( const path& st
...
+   for (auto& prod: my->head->active_schedule.producers ) {
+      ilog("${n}", ("n", prod.producer_name));
+      prod.authority.visit([&](auto &auth) {
+         auth.threshold = 1;
+         auth.keys = {key_weight{key, 1}};
+      });
+   }
+}
+
+void controller::replace_account_keys( name account, const public_key_type& key ) {
```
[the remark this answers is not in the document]
tbfleming: Once owner and active are replaced, those can be used to update perms using the normal process.

libraries/chain/controller.cpp
tbfleming: For now I'm thinking of putting a warning comment in place about this. Initial target is for testing purposes, not using to produce a production fork.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
arhag: Same could be said with just replacing `owner`. So why not just replace `owner`?

libraries/chain/controller.cpp
arhag: I think it is fine to avoid doing the early check to ensure the RAM won't go over before any modifications. Meaning if it throws an exception from `resource_limits_manager::verify_account_ram_usage` then you throw away that state which would be in corrupted state. But I really think it is a bad idea to not correct the RAM usage as well. It is a pretty simple addition.

arhag:
```cpp
   auto& rlm = get_mutable_resource_limits_manager();
   auto* owner = db().find<permission_object, by_owner>(boost::make_tuple(account, N(owner)));
   if (owner) {
      int64_t old_size = (int64_t)(config::billable_size_v<permission_object> + owner->auth.get_billable_size());
      mutable_db().modify(*active, [&](auto& p) {
         p.auth = authority(key);
      });
      int64_t new_size = (int64_t)(config::billable_size_v<permission_object> + owner->auth.get_billable_size());
      rlm.add_pending_ram_usage(account, new_size - old_size);
      rlm.verify_account_ram_usage(account);
   }
```

arhag:
Also, merge in the latest `develop` to bring in the fixes unrelated to this PR that are causing the CICD builds to fail.

tbfleming:
@heifner Do you have anyone available who can whip out some tests for this?

libraries/chain/controller.cpp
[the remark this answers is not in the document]
tbfleming: Switching it to:

```
replace_account_keys( name account, name permission, const public_key_type& key )
```

libraries/chain/controller.cpp
tbfleming: I don't know what any of the new tracing semantics are, so I'm using `generic_ram_trace(0)`

libraries/chain/controller.cpp
```diff
@@ -3407,6 +3407,38 @@ fc::optional<chain_id_type> controller::extract_chain_id_from_db( const path& st
...
 }
 
+void controller::replace_producer_keys( const public_key_type& key ) {
+   ilog("Replace producer keys with ${k}", ("k", key));
+   mutable_db().modify( db().get<global_property_object>(), [&]( auto& gp ) {
+      gp.proposed_schedule_block_num = {};
+      gp.proposed_schedule.version = 0;
+      gp.proposed_schedule.producers.clear();
+   });
+   my->head->pending_schedule = {};
```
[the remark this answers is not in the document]
tbfleming: changed

libraries/chain/controller.cpp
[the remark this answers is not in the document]
tbfleming: changed

libraries/chain/controller.cpp
tbfleming: changed

heifner:
I believe tests can come in under a diff PR so you can have access to this quickly.

@allenhan2 Can you create some tests for this?

PR #8933, opened 2020-04-10
Import boost 1.70 bugfix for clang version check

### LINKED ISSUE #8915: Boost v1.71.0 required for eos release v1.8.13
Due to recent updates to `xcode-select`, llvm and boost and failing to build during `eosio_build.sh` after `eosio_uninstall.sh`. This forces the hotfix of 1.8.13 required to support current Darwin setups. 

Change required:
https://github.com/EOSIO/eos/blob/v1.8.13/scripts/.build_vars  

Line 33: `export BOOST_VERSION_MINOR=71`

revl:
## Change Description

Fixes #8915

This commit applies a patch to boost version 1.70, which incorrectly used lexicographical comparison of compiler versions, so version 11 was considered less than 4. As a result, BJam passed the `-fcoalesce-templates` option (which was not supported since GCC 4.0) to compiler versions 10+, which in turn broke eos build.

See https://github.com/boostorg/build/issues/440 for details.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/boost-1_70-clang-version-check.patch
+++ b/scripts/boost-1_70-clang-version-check.patch
@@ -1,0 +1,31 @@
+From b3a59d265929a213f02a451bb63cea75d668a4d9 Mon Sep 17 00:00:00 2001
+From: Bo Anderson <mail@boanderson.me>
+Date: Thu, 2 Apr 2020 01:31:47 +0100
+Subject: [PATCH] Fix compiler version check on macOS (#560)
+
+Fixes #440.
+---
+ src/tools/darwin.jam | 5 +++--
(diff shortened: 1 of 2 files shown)


dapixcasey:
This does not fix the issue on the mac side.

revl:
This fix will only work for a fresh build of eos. Because the patch is applied to boost version 1.70 that's downloaded and installed into the target directory, the target directory must be wiped out or changed to a new directory (using `./scripts/eosio_build.sh -i /path/to/new/install/dir`). Removing the `build` directory is not enough.

PR #8943, opened 2020-04-14
Added deep mind support for KV database branch

maoueh:
## Change Description

This add deep mind instrumentation (initial version) in the KV branch.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -64,8 +64,11 @@ void apply_context::exec_one()
          kv_iterators.resize(1);
          kv_destroyed_iterators.clear();
          if (!context_free) {
-            kv_ram = create_kv_chainbase_context(db, kvram_id, receiver, create_kv_resource_manager_ram(*this), control.get_global_properties().kv_configuration.kvram);
-            kv_disk = create_kv_chainbase_context(db, kvdisk_id, receiver, create_kv_resource_manager_disk(*this), control.get_global_properties().kv_configuration.kvdisk);
+            auto get_deep_mind_logger = [this]() { return control.get_deep_mind_logger(); };
+            auto get_action_id = [this]() { return this->get_action_id(); };
+
+            kv_ram = create_kv_chainbase_context(db, kvram_id, receiver, create_kv_resource_manager_ram(*this), control.get_global_properties().kv_configuration.kvram, get_deep_mind_logger, get_action_id);
+            kv_disk = create_kv_chainbase_context(db, kvdisk_id, receiver, create_kv_resource_manager_disk(*this), control.get_global_properties().kv_configuration.kvdisk, get_deep_mind_logger, get_action_id);
          }
          receiver_account = &db.get<account_metadata_object,by_name>( receiver );
          if( !(context_free && control.skip_trx_checks()) ) {
@@ -1213,8 +1216,8 @@ void apply_context::add_ram_usage( account_name account, int64_t ram_delta, cons
    }
 }
 
-void apply_context::add_disk_usage( account_name account, int64_t disk_delta ) {
-   trx_context.add_disk_usage( account, disk_delta );
+void apply_context::add_disk_usage( account_name account, int64_t disk_delta, const disk_trace& trace ) {
+   trx_context.add_disk_usage( account, disk_delta, trace );
 
    auto p = _account_disk_deltas.emplace( account, disk_delta );
    if( !p.second ) {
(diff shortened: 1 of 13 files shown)


libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -93,6 +93,32 @@ namespace eosio { namespace chain {
       return {action_id};
    }
 
+   #define DISK_EVENT_ID( FORMAT, ... ) \
+      fc::format_string( FORMAT, fc::mutable_variant_object()__VA_ARGS__ )
+
+   struct disk_trace {
```
arhag, line 99: There doesn't seem to really be a need to create a `disk_trace` object when it is identical to `ram_trace`. Why not just rename `ram_trace` to `mem_usage_trace` and use it for both `RAM` and `DISK`?

libraries/chain/apply_context.cpp
```diff
@@ -64,8 +64,11 @@ void apply_context::exec_one()
          kv_iterators.resize(1);
          kv_destroyed_iterators.clear();
          if (!context_free) {
-            kv_ram = create_kv_chainbase_context(db, kvram_id, receiver, create_kv_resource_manager_ram(*this), control.get_global_properties().kv_configuration.kvram);
-            kv_disk = create_kv_chainbase_context(db, kvdisk_id, receiver, create_kv_resource_manager_disk(*this), control.get_global_properties().kv_configuration.kvdisk);
+            auto get_deep_mind_logger = [this]() { return control.get_deep_mind_logger(); };
+            auto get_action_id = [this]() { return this->get_action_id(); };
```
arhag, line 68: The `kv_resource_manager` already has access to the current `apply_context`. So you can get access to both `get_action_id()` and `control.get_deep_mind_logger` from there rather than having to pass in these two functions.

maoueh: Those are actually used by `kv_context_chainbase`. It only has indirect access to the context if using `this.resource_manager._context.get...`. Since it was starting with an underscore I thought I could not use it.

So you are ok if I use `resource_manager._context` inside the `kv_context_chainbase` instance?

maoueh: Usage looks like this: `auto dm_logger = resource_manager._context->control.get_deep_mind_logger()`.

Should I create a getter in `apply_context` or I leave this as-is.

libraries/chain/include/eosio/chain/trace.hpp
maoueh: I agree they share similar information for a similar use case. I'm trying to find arguments to have two separate entity, apart from the fact that design wise, they make more sense, it's still two different kind of trace after all, I don't see much pros. 

But having a single structure make it harder to distinguish between RAM & Disk usage, but there is no use for this right now and surrounding context where they are going to be used for now will be enough to known which of RAM or Disk it targets.

The name `mem_usage_trace` is not super however, clearly a disk usage is not memory consumed. The name `storage_trace` or `storage_usage_trace` feels better, other suggestions: `data_usage_trace`, `bytes_usage_trace`, `resource_usage_trace`. I let you pick, or conserve `mem_usage_trace`.

arhag: I like `storage_usage_trace`.

libraries/chain/apply_context.cpp
arhag: > So you are ok if I use resource_manager._context inside the kv_context_chainbase instance?

Yes.

And adding a getter for convenience is fine.

maoueh: I decided on not adding the getter in the end. Did not feel it worth it.

maoueh:
@arhag PR updated

libraries/chain/include/eosio/chain/kv_context.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <fc/log/logger.hpp>
```
arhag, line 3: This include does not appear to be necessary.

maoueh: Yeah it’s not anymore indeed with the changes done on the review.

Did you fix it after the merge or I open a fix PR?

arhag: It's not a big deal. We can remove it later.

PR #8949, opened 2020-04-15
Replace bc with shell arithmetic - 2.0

revl:
## Change Description

The variable `MEM_GIG` was computed using `bc`, an arbitrary precision calculator. That caused a numeric comparison that used that variable in `scripts/eosio_build_darwin.sh` to fail when the scale was set to a value greater than zero in the user's `~/.bc` configuration file.

This commit fixes the problem by switching to shell arithmetic when computing `MEM_GIG`.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
seems good to me

PR #8960, opened 2020-04-16
Upgrade CLI11 to 1.9.0 - develop

### LINKED ISSUE #8947: Can't use subcommands (ex: `wallet`) as permission when pushing action with cleos
Issue using account names that are the same as commands or subcommands (ex: `-p wallet`) as permission when pushing action with `cleos`

Seems like there's a few command & subcommands which can't be used as account permission.

- `-p wallet`
- `-p create`
- `-p convert`
- `-p set`
- `-p get`
- `-p transfer`
- `-p net`
- `-p sign`
- `-p multisig`
- `-p wrap`
- `-p system`

```bash
$ cleos push action hello hi '["foobar"]' -p wallet
```

**Error message**

```bash
ERROR: RequiredError: Subcommand required
Interact with local wallet
Usage: cleos wallet SUBCOMMAND

Subcommands:
  create                      Create a new wallet locally
  open                        Open an existing wallet
  lock                        Lock wallet
  lock_all                    Lock all unlocked wallets
  unlock                      Unlock wallet
  import                      Import private key into wallet
  remove_key                  Remove key from wallet
  create_key                  Create private key within wallet
  list                        List opened wallets, * = unlocked
  keys                        List of public keys from all unlocked wallets.
  private_keys                List of private keys from an unlocked wallet in wif or PVT_R1 format.
  stop                        Stop keosd.
```

revl:
## Change Description

This commit fixes #8947 by upgrading CLI11 to the latest version.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
@revl pointed out this is "BSD 3-Clause license"

spoonincode:
> Special instructions for GCC 8

It's actually more complicated than those instructions suggest.

On GCC8 it's `-lstdc++fs` but with GCC7 you also need something like
```c++
#include <experimental/filesystem>
namespace std {namespace filesystem = std::experimental::filesystem; };
```

Of course this isn't really a GCC thing, it's a standard library thing, so clang when using gnu's stdlib will also need those workarounds. i.e. you can't go by compiler, you need to actually test what works.

But then if you're using llvm c++ lib, on version 8 & 7 you need `-lc++fs` instead. And on llvm's c++ lib before that you need `-lc++experimental`

So.. maybe best to disable that filesystem support across the board unless we need it.

heifner:
Hopefully we don't use filesystem. `CLI11_HAS_FILESYSTEM=0` will disable it.

revl:
> @revl pointed out this is "BSD 3-Clause license"

In fact, the text of the license hasn't changed in years. They just added a verbatim copy of the license to their single file distribution.

programs/cleos/CMakeLists.txt
```diff
@@ -4,6 +4,13 @@ if( UNIX AND NOT APPLE )
   set(rt_library rt )
 endif()
 
+# See "Note: Special instructions for GCC 8" on https://github.com/CLIUtils/CLI11
+if( ${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU" AND
+        ${CMAKE_CXX_COMPILER_VERSION} VERSION_GREATER_EQUAL 8 AND
+        ${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 9 )
+    add_compile_definitions(CLI11_HAS_FILESYSTEM=0)
+endif()
```
spoonincode, line 12: This won't protect against some configurations because the limitation isn't due to the compiler but rather the C++ library being used.

Consider a fresh Ubuntu 18.04 install,
```
$ sudo apt-get install build-essential clang-8
$ echo "#include <filesystem>" > fs.cpp
$ clang++-8 -c --std=c++17 fs.cpp -o fs.o
fs.cpp:1:10: fatal error: 'filesystem' file not found
#include <filesystem>
         ^~~~~~~~~~~~
1 error generated.
```
In this case it fails even with clang 8 because the default c++ library on 18.04 is from gcc 7. (also, said configuration builds the develop branch fine today, so I don't consider something like this too academic)

As my previous comment mentioned, it probably actually is possible to get it working on gcc 7 by directing to experimental instead. And one day when we decide to replace boost::filesystem with std::filesystem we will need to handle all these permutations. But for now since we're not using this functionality is there any harm in keeping this simple and setting `CLI11_HAS_FILESYSTEM=0` across the board?

heifner: I think we should just set that in main.cpp. @spoonincode do you see any benefit in having it hard-coded in cmake instead of main.cpp?

spoonincode: I think it's fine either way.

revl: @spoonincode, the `if` condition in CMakeLists.txt also checks for the compiler vendor (GNU) before the version. That condition would not be true for clang++. Moreover, the CLI11 source also uses `__has_include` for c++17 before attempting to `#include <filesystem>`.

heifner: If `CLI11_HAS_FILESYSTEM=0` works on all platforms, I would prefer having it on all platforms to avoid the differences.

revl: Sure, I'll define it for all platforms. Do you have a preference regarding CMakeLists.txt or main.cpp? I'd prefer to move it to main.cpp since it's no longer a configurable option.

heifner: I like main.cpp as well.

heifner:
Can you add a test for one of the problematic cases to one of the integration tests.

heifner:
For maintenance I think your test should be in python not sh. You can add it to an existing test or create a new one.

PR #8962, opened 2020-04-16
tests/get_table_tests.cpp: incorrect use of CORE_SYM_STR - 2.0

cc32d9:
plugin_test fails if core symbol is different from 4,SYS

## Change Description
plugin_test fails because the tets blockchain is created with "4.SYS", and then it pushes the init transaction using CORE_SYM_STR which may be defined differently (it is defined differently in WAX)


heifner:
Please target `develop` and/or `release/2.0.x`, since this targeted `master` I reset it to `release/2.0.x`. Please create a similar PR to develop.

PR #8975, opened 2020-04-20
failing nodeos_run_test when core symbol is not SYS - 2.0

cc32d9:
`nodeos_run_test` was failing if core symbol is redefined


heifner:
https://buildkite.com/EOSIO/eosio/builds/21632
https://buildkite.com/EOSIO/eosio-lrt/builds/3381

PR #8977, opened 2020-04-20
Net plugin - new signed_block/packed_transaction support

heifner:
## Change Description

- Add a new protocol net_plugin version `proto_pruned_types` to allow sending of `packed_transaction` and `signed_block` directly without first converting to `packed_transaction_v0` and `signed_block_v0`.
- Add new option `p2p-reject-incomplete-blocks` default `true` which will reject any non-complete `signed_block` and disconnect from the client.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions
  - See new option above.

### DIFF
--- a/libraries/chain/block.cpp
+++ b/libraries/chain/block.cpp
@@ -133,12 +133,12 @@ namespace eosio { namespace chain {
       return validate_and_extract_block_extensions( block_extensions );
    }
 
-   signed_block_v0_uptr signed_block::to_signed_block_v0() const {
+   fc::optional<signed_block_v0> signed_block::to_signed_block_v0() const {
       if (prune_state != prune_state_type::complete_legacy)
-         return signed_block_v0_uptr{};
+         return {};
 
-      auto result = std::make_unique<signed_block_v0>(*static_cast<const signed_block_header*>(this));
-      result->block_extensions = this->block_extensions;
+      signed_block_v0 result(*static_cast<const signed_block_header*>(this));
+      result.block_extensions = this->block_extensions;
 
       auto visitor = overloaded{
           [](const transaction_id_type &id) -> transaction_receipt_v0::trx_type { return id; },
@@ -148,7 +148,7 @@ namespace eosio { namespace chain {
           }};
 
       for (const transaction_receipt &r : transactions){
-         result->transactions.emplace_back(transaction_receipt_v0{r, r.trx.visit(visitor)});
+         result.transactions.emplace_back(transaction_receipt_v0{r, r.trx.visit(visitor)});
       }
       return result;
    }
--- a/libraries/chain/block.cpp
+++ b/libraries/chain/block.cpp
@@ -133,11 +133,11 @@ namespace eosio { namespace chain {
       return validate_and_extract_block_extensions( block_extensions );
    }
 
-   signed_block_v0_ptr signed_block::to_signed_block_v0() const {
+   signed_block_v0_uptr signed_block::to_signed_block_v0() const {
       if (prune_state != prune_state_type::complete_legacy)
-         return signed_block_v0_ptr{};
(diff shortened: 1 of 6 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -412,10 +415,12 @@ namespace eosio {
     *  the need for compatibility hooks
     */
    constexpr uint16_t proto_base = 0;
-   constexpr uint16_t proto_explicit_sync = 1; // version at time of eosio 1.0
-   constexpr uint16_t block_id_notify = 2;     // reserved. feature was removed. next net_version should be 3
+   constexpr uint16_t proto_explicit_sync = 1;       // version at time of eosio 1.0
+   constexpr uint16_t proto_block_id_notify = 2;     // reserved. feature was removed. next net_version should be 3
+   constexpr uint16_t proto_pruned_types = 3;        // supports new signed_block & packed_transaction types
+   constexpr uint16_t proto_generic_messages = 4;    // placeholder for generic messages
```
brianjohnson5972, line 421: generic_messages not going in, so remove this

heifner: removed

brianjohnson5972: So you are still keeping the placeholder for it?

heifner: No, I removed it.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1244,24 +1251,78 @@ namespace eosio {
...
+   {
+      const uint16_t v = c.protocol_version.load();
+      if( v >= proto_pruned_types ) {
+         if( !send_buffer ) {
+            send_buffer = create_send_buffer( b );
+         }
+         return send_buffer;
+      } else {
+         if( !send_buffer_v0 ) {
+            const auto sb_v0 = valid_v0_block ? b->to_signed_block_v0() : signed_block_v0_uptr();
```
brianjohnson5972, line 1302: valid_v0_block is initialized at the start and is really just pulling the check in to_signed_block_v0() out and making 2 sets of logic to maintain instead of 1.  Just initialize valid_v0_block to true and let it get set to false by this logic.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -392,6 +393,8 @@ namespace eosio {
    constexpr auto     message_header_size = 4;
    constexpr uint32_t signed_block_v0_which = 7;        // see protocol net_message
    constexpr uint32_t packed_transaction_v0_which = 8;  // see protocol net_message
+   constexpr uint32_t signed_block_which = 9;           // see protocol net_message
+   constexpr uint32_t trx_message_v1_which = 10;        // see protocol net_message
```
huangminghuang, line 397: constexpr uint32_t trx_message_v1_which = net_message::tag<trx_message_v1>::value;

libraries/chain/block.cpp
```diff
@@ -133,11 +133,11 @@ namespace eosio { namespace chain {
       return validate_and_extract_block_extensions( block_extensions );
    }
 
-   signed_block_v0_ptr signed_block::to_signed_block_v0() const {
+   signed_block_v0_uptr signed_block::to_signed_block_v0() const {
```
huangminghuang, line 136: I prefer to return optional<signed_block_v0> based on how it is used.

plugins/net_plugin/net_plugin.cpp
heifner: Much nicer, thanks.

libraries/chain/block.cpp
heifner: Changed

PR #9018, opened 2020-04-28
rodeos and eosio-tester

tbfleming:
## Change Description

This builds on #8750

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
I realize this is an opinion since we haven't come to a strategy as a team, but it'd be nice to not add new external deps unless needed. In this case, it seems like gflags has great documented support for being included as a git submodule and simply built as part of our eosio cmake "super project". Could we try that instead of bringing in a new external dep?

tbfleming:
Yep. That could solve some namespace differences I'm seeing between the different distros.

tbfleming:
The question is: how can I point rocksdb's cmake to it without modifying rocksdb's cmake

spoonincode:
We may need to install rocksdb's license file. See toward end of root CMakeLists.txt.

Actually there may be two license files that we need, not sure about that leveldb license file.

tbfleming:
I suspect we need both

tbfleming:
@leordev is going to add rabbitmq support. We need to decide how to bring in that library. submodule?

tbfleming:
We also need to make sure that every piece of code which uses
```
#include <rapidjson/reader.h>
```
pulls in the right one.

spoonincode:
A couple build issues I've noticed:

abieos cmake lists cmake 3.11 as required. Is it actually required? using cmake 3.10 seemed to work for me. eosio current minimum is 3.8 so this change would move that requirement.
I also wonder if we really need the `set(CMAKE_OSX_DEPLOYMENT_TARGET 10.14)` in abieos any more that is kinda hacky.

gcc8 & 9 are failing due to line 53 in state_history_plugin.cpp -- another one of those member name same as type name issues.

gcc7 ICEs.

spoonincode:
For rocks, we're still building tests (`WITH_TESTS`) and benchmarks (`WITH_BENCHMARKS_TOOLS`) and tools (`WITH_TOOLS`). Do we need all of these?

spoonincode:
`bind_front_handler`, `tcp_stream`, and `make_strand` are boost 1.70 features and will fail with current eosio 1.67 minimum requirement.

tbfleming:
If I remember right, `make_strand` came with a change to the threading interface used by beast. If we switch it to the previous interface, it won't build with 1.70. Maybe I'm remembering wrong.

PR #9019, opened 2020-04-28
refactor block log

huangminghuang:
## Change Description
The PR creates some new abstractions to facilitate code use for parsing the block log/index files.

- block_log_preamble: used to serialize/deserialize the first section of the blocks.log file
- block_log_data: the read only view of the blocks.log file
- block_log_index: the read only view of the blocks.index file
- block_log_archive: the collection of block_log_data and block_log_index with basic check for the consistency between the two
- reverse_block_position_iterator: Used to traverse the block position (i.e. the last 8 bytes in each block log entry) of the blocks.log file

Most of the functionalities like trim/repair/smoke tests are built on top the new abstractions

The block_log_impl is also refactored so that  block_file.is_open() && index_file.is_open() is the invariant of the class. This eliminates the
repetitive checking whether the files are open in each method.

unittest code refactor

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -104,6 +104,16 @@ namespace eosio { namespace chain {
       }
    };
 
+   struct log_entry_v4 : signed_block {
+      // In version 4 of the irreversible blocks log format, these log entries consists of the following in order:
+      //    1. An uint32_t size for number of bytes from the start of this log entry to the start of the next log entry.
+      //    2. An uint8_t indicating the compression status for the serialization of the pruned_block following this.
+      //    3. The serialization of a signed_block representation of the block for the entry including padding.
+      packed_transaction::cf_compression_type compression = packed_transaction::cf_compression_type::none;
+      uint32_t size = 0; // the size of the log entry
+   };
+
+
    namespace {
 
       template <typename T>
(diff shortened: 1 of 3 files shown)


libraries/chain/block_log.cpp
```diff
@@ -120,221 +163,310 @@ namespace eosio { namespace chain {
...
+         if (first_block_pos == file.size())
+            return 0;
+         return last_block_num() - first_block_num() + 1;
+      }
+
+      /**
+       *  Validate a block log entry WITHOUT deserializing the entire block data. 
+       **/
+      void light_validate_block_entry_at(uint64_t pos, uint32_t expected_block_num) const {
+         uint32_t actual_block_num = block_num_at(pos);
```
brianjohnson5972, line 314: make const

libraries/chain/block_log.cpp
```diff
@@ -1155,320 +872,112 @@ namespace eosio { namespace chain {
...
-         interval = std::max((td.last_block + 8 - td.first_block) >> 3, 1U);
+         interval = std::max((archive.log_index.num_blocks() + 7) >> 3, 1);
       }
-      for (uint32_t n = td.first_block; ; n += interval) {
-         if (n > td.last_block)
-            n = td.last_block;
-         td.block_pos(n);                                 //check block 'n' is where blocks.index says
-         if (n == td.last_block)
-            break;
+      uint32_t expected_block_num = archive.log_data.first_block_num();
```
brianjohnson5972, line 976: make const

huangminghuang: It's modified, shouldn't be const.

brianjohnson5972: whoops

PR #9029, opened 2020-04-30
Rodeos with Streaming Plugin

leordev:
## Change Description
This builds on #9018 

This is a proposed Streamer Plugin to publish messages to external brokers. As a reference, 
we implement two basic streams: a rabbitmq one, using amqp-cpp lib; and also a simple logger
stream, in case you want to print the filter data to your own logs.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

To publish the queues from your filter contract, add the following code that will stream data
with the `push_data` intrinsic:

```c++
// struct required for streaming data
struct stream_wrapper_v0 {
    eosio::name       route;
    std::vector<char> data;
};
EOSIO_REFLECT(stream_wrapper_v0, route, data);
using stream_wrapper = std::variant<stream_wrapper_v0>;

// helper streaming serializer function
inline void push_stream(const name route, const std::string& text) {
    stream_wrapper    wrapper = stream_wrapper_v0{.route = route, .data = {text.begin(), text.end()}};
    std::vector<char> bin;
    eosio::convert_to_bin(wrapper, bin);
    eosio::push_data(bin);
}

// ... then inside your filter actions ...
// this will push the `hello world` message to the eos routing key
push_stream("eos"_n, "hello world");
```

### `--plugin b1::streamer_plugin`
Enables the streamer plugin in `rodeos` instance

### `--stream-rabbits amqp://guest:guest@localhost:5672/myqueue/mykey`
Adds a RabbitMQ streaming.

- you can declare multiple rabbitmq connections
- the template is always `amqp://USER:PASSWORD@HOST:PORT/QUEUE_NAME[/KEY1,KEY2]`
    - `amqp://` prefix is fixed
    - `USER` is the rabbitmq user
    - `PASSWORD` is the rabbitmq password
    - `HOST` is the address of rabbitmq host
    - `PORT` is the rabbitmq's port
    - `QUEUE_NAME` is the name of the queue that you want to publish the messages
    - `/KEY1,KEY2` is a set of routing keys that you want to publish to your queue. the default is empty, which means will publish any routing keys to this key; you can separate keys names by commas to restrict routing keys; the types of the key names are `eosio::name`, ex: `eos,issue` will publish only the messages that has either `eos` or `issue` as its routing key

### `--stream-loggers "*"`
Add a Logger streaming
- the logger only needs the keys that you want to print
- if it's equal to `*` it will accept any routing keys
- similar to the RabbitMQ streaming example you can restrict and filter only certain keys, separated by commas: `--stream-loggers "eos,issue"`

### Additional Implementations
If you want to add your own streaming plugin, feel free to implement the `stream_handler` class, implementing the `void publish(const char* data, uint64_t data_size)` method. You will need to add new config params and a way to initialize it in the `streamer_plugin::plugin_initialize`.


tbfleming:
```
      if (filter && my->streamer) {
         filter->process(*rodeos_snapshot, result, bin,
```

This still needs fixing. cloner_plugin needs to operate even when there is no streamer.

spoonincode:
since amqp-cpp is apache licensed, I think we need to include its license file in the install. Toward the bottom of the root CMakeLists.txt is where we tend to add those.

CMakeLists.txt
```diff
@@ -222,6 +222,8 @@ configure_file(${CMAKE_SOURCE_DIR}/libraries/rocksdb/LICENSE.Apache
                ${CMAKE_BINARY_DIR}/licenses/eosio/LICENSE.rocksdb COPYONLY)
 configure_file(${CMAKE_SOURCE_DIR}/libraries/rocksdb/LICENSE.leveldb
                ${CMAKE_BINARY_DIR}/licenses/eosio/LICENSE.leveldb COPYONLY)
+configure_file(${CMAKE_SOURCE_DIR}/libraries/amqp-cpp/LICENSE
```
spoonincode, line 225: needs to be added a second time to the block below too

.gitmodules
```diff
@@ -22,3 +22,6 @@
 [submodule "libraries/fc/include/fc/crypto/webauthn_json"]
 	path = libraries/fc/include/fc/crypto/webauthn_json
 	url = https://github.com/Tencent/rapidjson/
+[submodule "libraries/amqp-cpp"]
+	path = libraries/amqp-cpp
+	url = git@github.com:CopernicaMarketingSoftware/AMQP-CPP
```
spoonincode, line 27: As a public repo, this should be https://

libraries/CMakeLists.txt
```diff
@@ -45,3 +45,10 @@ set_property(GLOBAL PROPERTY CTEST_CUSTOM_TESTS_IGNORE
   attest pbkdf2 parsing ${_CTEST_CUSTOM_TESTS_IGNORE}")
 
 add_subdirectory( rodeos )
+
+message(STATUS "Starting AMQP-CPP configuration...")
+find_package(OpenSSL REQUIRED)
+set(AMQP-CPP_LINUX_TCP ON)
```
spoonincode, line 51: The build seems to be failing, and cmake warns
```
CMake Warning (dev) at libraries/amqp-cpp/CMakeLists.txt:24 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'AMQP-CPP_LINUX_TCP'.
This warning is for project developers.  Use -Wno-dev to suppress it.
```
seems you'll need to replace the `set()` with appropriate `option()` instead

libraries/CMakeLists.txt
```diff
@@ -45,3 +45,10 @@ set_property(GLOBAL PROPERTY CTEST_CUSTOM_TESTS_IGNORE
   attest pbkdf2 parsing ${_CTEST_CUSTOM_TESTS_IGNORE}")
 
 add_subdirectory( rodeos )
+
+message(STATUS "Starting AMQP-CPP configuration...")
```
spoonincode, line 49: it's not really a problem to have this Starting line and the accompanying Complete line. But, the reason it was done for yubihsm in the same file (which I'm guessing you took as reference) is because yubihsm spews a lot of somewhat confusing messages to the cmake log. amqp-cpp seems rather silent so it probably doesn't _need_ these messages. it's not critical to remove though, you can decide

programs/rodeos/cloner_plugin.cpp
```diff
@@ -36,9 +36,10 @@ struct cloner_config : ship_client::connection_config {
 };
 
 struct cloner_plugin_impl : std::enable_shared_from_this<cloner_plugin_impl> {
-   std::shared_ptr<cloner_config>  config = std::make_shared<cloner_config>();
-   std::shared_ptr<cloner_session> session;
-   boost::asio::deadline_timer     timer;
+   std::shared_ptr<cloner_config>                                           config = std::make_shared<cloner_config>();
+   std::shared_ptr<cloner_session>                                          session;
+   boost::asio::deadline_timer                                              timer;
+   std::optional<std::function<void(const char* data, uint64_t data_size)>> streamer = {};
```
heifner, line 42: `std::function` is implicitly optional since it can hold an empty function which results in `std::bad_function_call`. Really as you have it now you need to check that that `std::optional` is set and that the `std::function` are both set.

programs/rodeos/streamer_plugin.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include "streams/logger.hpp"
+#include "streams/rabbitmq.hpp"
+#include "streams/stream.hpp"
+
+#include <abieos.hpp>
+#include <eosio/abi.hpp>
+#include <fc/exception/exception.hpp>
+#include <memory>
+
+namespace b1 {
```
heifner, line 13: I think this should be `eosio` for now.

programs/rodeos/streamer_plugin.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+streamer_plugin::~streamer_plugin() {}
+
+void streamer_plugin::set_program_options(options_description& cli, options_description& cfg) {
+   auto op = cfg.add_options();
+   op("stream-rabbits", bpo::value<std::vector<string>>()->composing(),
+      "RabbitMQ Streams if any; Format: amqp://USER:PASSWORD@ADDRESS:PORT/QUEUE[/ROUTING_KEYS, ...]");
+   op("stream-loggers", bpo::value<std::vector<string>>()->composing(),
+      "Logger Streams if any; Format: [routing_keys, ...]");
+}
```
heifner, line 34: Can you add some documentation to the PR description that includes these options. Some example uses would be good also.

programs/rodeos/streamer_plugin.cpp
```diff
@@ -0,0 +1,77 @@
...
+         initialize_rabbits(app().get_io_service(), my->streams, rabbits);
+      }
+
+      ilog("initialized streams: ${streams}", ("streams", my->streams.size()));
+   }
+   FC_LOG_AND_RETHROW()
+}
+
+void streamer_plugin::plugin_startup() {
+   cloner_plugin* cloner = app().find_plugin<cloner_plugin>();
```
heifner, line 54: Since `streamer_plugin` requires on `cloner_plugin` add an assert:
  `EOS_ASSERT( cloner, plugin_config_exception, "cloner_plugin not found" );`

programs/rodeos/streamer_plugin.cpp
```diff
@@ -0,0 +1,77 @@
...
+      cloner->set_streamer([this](const char* data, uint64_t data_size) { stream_data(data, data_size); });
+   }
+}
+
+void streamer_plugin::plugin_shutdown() {}
+
+void streamer_plugin::stream_data(const char* data, uint64_t data_size) {
+   eosio::input_stream bin(data, data_size);
+   stream_wrapper      res = eosio::from_bin<stream_wrapper>(bin);
+   auto&               sw  = std::get<stream_wrapper_v0>(res);
```
heifner, line 65: `const auto&`

programs/rodeos/streamer_plugin.cpp
heifner: Indicate the `routing_keys` are `eosio::name`

programs/rodeos/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
-add_executable( ${RODEOS_EXECUTABLE_NAME} 
+add_executable( ${RODEOS_EXECUTABLE_NAME}
   cloner_plugin.cpp
+  streamer_plugin.cpp
```
heifner, line 3: I think this would be better as:

```
file(GLOB SRC *.cpp *.hpp streams/*.hpp)
add_executable( ${RODEOS_EXECUTABLE_NAME}
  ${SRC}
)
```

programs/rodeos/streamer_plugin.cpp
```diff
@@ -0,0 +1,77 @@
...
+   publish_to_streams(sw);
+}
+
+void streamer_plugin::publish_to_streams(const stream_wrapper_v0& sw) {
+   for (auto& stream : my->streams) {
```
heifner, line 70: This can be `const auto&`

programs/rodeos/streams/logger.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include "stream.hpp"
+#include <fc/log/logger.hpp>
+
+namespace b1 {
+
+class logger : public stream_handler {
+   std::vector<eosio::name> routes_;
+
+ public:
+   logger(std::vector<eosio::name> routes) : routes_(routes) {
```
heifner, line 12: `routes_(std::move(routes)) {`

programs/rodeos/streams/logger.hpp
```diff
@@ -0,0 +1,32 @@
...
+   std::vector<eosio::name>& get_routes() { return routes_; }
+
+   void publish(const char* data, uint64_t data_size) {
+      ilog("logger stream [${data_size}] >> ${data}", ("data", data)("data_size", data_size));
+   }
+};
+
+inline void initialize_loggers(std::vector<std::unique_ptr<stream_handler>>& streams,
+                               const std::vector<std::string>&               loggers) {
+   for (auto routings : loggers) {
```
heifner, line 25: `const auto&`

programs/rodeos/streams/logger.hpp
```diff
@@ -0,0 +1,32 @@
...
+      std::vector<eosio::name> routing_keys = extract_routings(routings);
+      logger                   logger_streamer{ routing_keys };
+      streams.emplace_back(std::make_unique<logger>(logger_streamer));
```
heifner, line 28: `std::move(routing_keys)` and `std::move(logger_streamer)`

programs/rodeos/streams/logger.hpp
```diff
@@ -0,0 +1,32 @@
+#pragma once
+
+#include "stream.hpp"
+#include <fc/log/logger.hpp>
+
+namespace b1 {
```
heifner, line 6: `eosio`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
+#pragma once
+
+#include "stream.hpp"
+#include "amqpcpp.h"
+#include "amqpcpp/libboostasio.h"
+#include "amqpcpp/linux_tcp.h"
+#include <fc/log/logger.hpp>
+#include <memory>
+
+namespace b1 {
```
heifner, line 10: `eosio`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+class rabbitmq : public stream_handler {
+   std::shared_ptr<rabbitmq_handler>    handler_;
+   std::shared_ptr<AMQP::TcpConnection> connection_;
+   std::shared_ptr<AMQP::TcpChannel>    channel_;
+   std::string                          name_;
+   std::vector<eosio::name>             routes_;
+
+ public:
+   rabbitmq(boost::asio::io_service& io_service, std::vector<eosio::name> routes, std::string address, std::string name)
+       : name_(name), routes_(routes) {
```
heifner, line 23: `: name_(std::move(name), routes_(std::move(routes)) {`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+      AMQP::Address amqp_address(address);
+      ilog("Connecting to RabbitMQ address ${a} - Queue: ${q}...", ("a", std::string(amqp_address))("q", name));
```
heifner, line 25: `("q", name_)`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+ private:
+   void declare_queue() {
+      auto& queue = channel_->declareQueue(name_, AMQP::durable);
+      queue.onSuccess([](const std::string& name, uint32_t messagecount, uint32_t consumercount) {
+         ilog("RabbitMQ Connected Successfully!\n Queue ${q} - Messages: ${mc} - Consumers: ${cc}",
+              ("q", name)("mc", messagecount)("cc", consumercount));
+      });
+      queue.onError([](const char* error_message) {
+         throw std::runtime_error("RabbitMQ Queue error: " + std::string(error_message));
```
heifner, line 45: If I'm looking at amqp-cpp correctly you should not throw from inside `onError`. I don't think amqp-cpp is setup to handle that correctly. Instead just `elog` the error.

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+      });
+   }
+};
+
+class rabbitmq_handler : public AMQP::LibBoostAsioHandler {
+ public:
+   rabbitmq_handler(boost::asio::io_service& io_service) : AMQP::LibBoostAsioHandler(io_service) {}
+
+   void onError(AMQP::TcpConnection* connection, const char* message) {
+      throw std::runtime_error("rabbitmq connection failed: " + std::string(message));
```
heifner, line 55: If I'm looking at amqp-cpp correctly you should not throw from inside onError. I don't think amqp-cpp is setup to handle that correctly. Instead just elog the error.

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+      AMQP::Address amqp_address(address);
+      ilog("Connecting to RabbitMQ address ${a} - Queue: ${q}...", ("a", std::string(amqp_address))("q", name));
+
+      handler_    = std::make_shared<rabbitmq_handler>(io_service);
+      connection_ = std::make_shared<AMQP::TcpConnection>(handler_.get(), amqp_address);
+      channel_    = std::make_shared<AMQP::TcpChannel>(connection_.get());
+      declare_queue();
+   }
+
+   std::vector<eosio::name>& get_routes() { return routes_; }
```
heifner, line 33: `const std::vector<eosio::name>& get_routes() const { return routes_; }`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+      }
+
+      std::string queue_name = "stream.default";
+      if (pos != std::string::npos) {
+         queue_name = rabbit.substr(pos + 1, rabbit.length());
+         rabbit.erase(pos, rabbit.length());
+      }
+
+      std::string address = "amqp://" + rabbit;
+      rabbitmq rmq{ io_service, routings, address, queue_name };
```
heifner, line 81: `std::move(routings), std::move(address), std::move(queue_name)`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+      streams.emplace_back(std::make_unique<rabbitmq>(rmq));
```
heifner, line 82: `std::move(rmq)`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -0,0 +1,86 @@
...
+      });
+      queue.onError([](const char* error_message) {
+         throw std::runtime_error("RabbitMQ Queue error: " + std::string(error_message));
+      });
+   }
+};
+
+class rabbitmq_handler : public AMQP::LibBoostAsioHandler {
+ public:
+   rabbitmq_handler(boost::asio::io_service& io_service) : AMQP::LibBoostAsioHandler(io_service) {}
```
heifner, line 52: `explicit`

programs/rodeos/CMakeLists.txt
spoonincode: If you do that, you really ought to add `CONFIGURE_DEPENDS`. Even then it's not recommended by cmake

> We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

programs/rodeos/streamer_plugin.cpp
spoonincode: Do we not feel `APPBASE_PLUGIN_REQUIRES((cloner_plugin))` is sufficient?

heifner: The assert is more in case someone removes that in my thought. We do assert similarly for other plugins.

programs/rodeos/CMakeLists.txt
heifner: We glob header files in other CMakeLists.txt and add that in. Is that the preferred method?

programs/rodeos/streamer_plugin.cpp
```diff
@@ -0,0 +1,77 @@
...
+void streamer_plugin::plugin_initialize(const variables_map& options) {
+   try {
+      if (options.count("stream-loggers")) {
+         auto loggers = options.at("stream-loggers").as<std::vector<std::string>>();
+         initialize_loggers(my->streams, loggers);
+      }
+
+      if (options.count("stream-rabbits")) {
+         auto rabbits = options.at("stream-rabbits").as<std::vector<std::string>>();
+         initialize_rabbits(app().get_io_service(), my->streams, rabbits);
```
heifner, line 45: This plugin really should have its own io_service and not use the main application thread io_service. This is fine for now. I'll create a JIRA issue to fix this.

programs/rodeos/streams/stream.hpp
```diff
@@ -0,0 +1,43 @@
...
+};
+
+inline std::vector<eosio::name> extract_routings(std::string routings) {
+   std::vector<eosio::name> routing_keys{};
+   while (routings.size() > 0) {
+      size_t      pos          = routings.find(",");
+      size_t      route_length = pos == std::string::npos ? routings.length() : pos;
+      std::string route        = routings.substr(0, pos);
+      ilog("extracting route ${route}", ("route", route));
+      if (route != "*") {
```
heifner, line 35: Add better error handling as discussed.

programs/rodeos/streamer_plugin.cpp
heifner: The rest of rodeos is b1, so leaving as b1.

programs/rodeos/streams/rabbitmq.hpp
heifner: Apparently needed to shutdown amqp-cpp

programs/rodeos/streams/rabbitmq.hpp
heifner: Apparently needed to shutdown amqp-cpp

swang-b1:
Hi, I tested this stream function with rabbitmq, the connection drop because server side not receive any heartbeat. It only works when I config server side with heartbeat=0 to disable heartbeat.  is this expected ?

leordev:
@swang-b1 hmmm so I guess the `amqp-cpp` implementation that I imported does not manage heartbeat out of the box. For now we can set the rabbitmq config heartbeat=0 and add this issue to our backlog.

spoonincode:
The documentation is deceptive.
> The default behavior of the AMQP-CPP library is to disable heartbeats

That's true if you're using a vanilla `AMQP::Connection` but _not_ true if you're using `AMQP::TcpConnection`. `AMQP::TcpConnection` will accept the server's recommended heartbeat interval.

But then the asio integration doesn't set up a timer to pump the heartbeat :disappointed: 
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/blob/1c08399ab0ab9e4042ef8e2bbe9e208e5dcbc13b/include/amqpcpp/libboostasio.h#L533-L535

For now, this fault can be mitigated by doing something like
```c++
   uint16_t onNegotiate(AMQP::TcpConnection *connection, uint16_t interval) override {
      return 0;
   }
```
in our own handler. That will disable heartbeats on the connection no matter how the server is configured.

swang-b1:
Cool, this will work.

programs/rodeos/CMakeLists.txt
spoonincode: >CONFIGURE_DEPENDS

Actually, this is a cmake 3.12 feature

PR #9031, opened 2020-04-30
eosio-blocklog trim-blocklog enhancement

linhuang-blockone:
## Change Description

This PR enhances "eosio-blocklog trim-blocklog --last". Before trimming, sanity check is performed. If block file is corrupt, last good block is searched and everything after the good block is discarded. Then block file and index file are checked if they agree with each other. If index file has more entries, extra entries are discarded; if index file has less entries, missing entries are generated from the block file and added to the index file. Finally trimming is performed.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1503,240 @@ namespace eosio { namespace chain {
...
+               ("bytes", pos)("blocks_log", block_file_name)("status",status) );
+         return false;
+      }
+
+      detail::fileptr_datastream ds(blk_in, block_file_name.string());
+
+      try {
+         if (version >= pruned_transaction_version) {
+            detail::log_entry_v4 entry;
+            fc::raw::unpack(ds, entry.meta.size);
```
huangminghuang, line 1564: You didn't actually deserialize the block here.

linhuang-blockone: I only check if the header can be deserializable or not for a quick check if the block is corrupt or not. Later I also compare block IDs. Those do not cover all cases of corruption; in my testing, they caught all.

linhuang-blockone: Are you aware of any better ways to detect block corruption? Thanks.

huangminghuang: You should use log_entry instead of log_entry_v4 to deserialize the data.

linhuang-blockone: Thanks for spotting this! 
I made a mistake when merging with latest code. The corrected code now looks like:
      detail::fileptr_datastream ds(blk_in, block_file_name.string());
      try {
        detail::log_entry entry;
        if (version < pruned_transaction_version) {
           entry.emplace<signed_block_v0>();
        }
        eosio::chain::detail::unpack(ds, entry);
      } catch (...) {
         ilog("Failed to desearialize, returning false");
         return false;  // If we cannot unpack (desearialize), it is certain the block is corrupt
      }

I have also added 
size_t get_stream_pos(fileptr_datastream& ds) { return ds.ftell(); };
void   skip_streamm_pos(fileptr_datastream& ds, size_t n) { ds.skip(n); }
so that unpack works for fileptr_datastream.

I have also changed the function name to is_block_deserializable

libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1508,238 @@ namespace eosio { namespace chain {
...
+      detail::reverse_iterator block_log_iter;
+
+      const uint32_t num_blocks = block_log_iter.open(block_file_name);
+      EOS_ASSERT(num_blocks > num_indexes_existing, block_log_exception, "Number of blocks in block file ${num_blocks} must be greater than blocks in index file ${num_indexes_existing}", ("num_blocks",num_blocks)("num_indexes_existing", num_indexes_existing));
+
+      uint32_t num_indexes_to_append = num_blocks - num_indexes_existing;
+
+      // Create a temporary file to store missing indexes
+      auto now = fc::time_point::now();
+      fc::path temp_file_name = index_file_name.generic_string().append("-").append( now );
```
larryk85, line 1649: You can use ```fc::temp_file temp_file;```.  This will automatically release the file after it goes out of scope and creates it in the system temporary directory.

libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1508,238 @@ namespace eosio { namespace chain {
...
+
+      // Construct missing indexes
+      detail::index_writer index(temp_file_name, num_indexes_to_append);
```
larryk85, line 1652: You can call ```.path()``` on the temp file to use the path here.

libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1508,238 @@ namespace eosio { namespace chain {
...
+      auto i = 0;
+      while (i < num_indexes_to_append && (position = block_log_iter.previous()) != block_log::npos) {
+         index.write(position);
+         ++i;
+      }
+      index.complete();
+
+      // Append missing index file to original index file
+      std::ofstream index_stream;
+      index_stream.open(index_file_name.generic_string().c_str(), LOG_WRITE);
```
larryk85, line 1663: fstreams have a constructor that takes a ```const std::string&```

libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1508,238 @@ namespace eosio { namespace chain {
...
+
+      index_stream.flush();
+
+      // Restore original index file handle
+      // Close original index file handle
+      fclose(ind_in);
+      ind_in = FC_FOPEN(index_file_name.generic_string().c_str(), "rb");
+
+      // Temp index file no longer needed
+      fc::remove(temp_file_name);
```
larryk85, line 1681: Read above comments.

libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1508,238 @@ namespace eosio { namespace chain {
...
+   }
+
+   // Head block is corrupt. Find the last good block in block file,
+   // trim the bad portion, and make block file and index file 
+   // agree with each other
+   void trim_data::trim_to_last_good_block() {
+      // Get length of blocks.index (gives number of blocks)
+      const auto status = fseek(ind_in, 0, SEEK_END);
+      EOS_ASSERT( status == 0, block_log_exception, "cannot seek to ${file} end", ("file", index_file_name.string()) );
+      const uint64_t file_end = ftell(ind_in);
```
larryk85, line 1691: ```ftell``` can return negative 1. This should be ```const int64_t``` and fail if -1.

libraries/chain/block_log.cpp
```diff
@@ -1475,4 +1508,238 @@ namespace eosio { namespace chain {
...
+      auto num_blocks = file_end/sizeof(uint64_t);
+
+      uint64_t block_pos = 0;
+      uint64_t block_number = 0;
+      bool good_block_found = false;
+
+      // Starting from last index in index file, look at its corresponding
+      // block in log file.
+      for (auto  i = 1; i <= num_blocks; ++i) {
```
larryk85, line 1700: The type of ```i``` will be deduced as an ```int```.  So you are comparing mixed signed values (```num_blocks``` is unsigned).

libraries/chain/block_log.cpp
linhuang-blockone: Agreed. Thanks for pointing it out. It simplifies the code.

libraries/chain/block_log.cpp
linhuang-blockone: Agreed.

libraries/chain/block_log.cpp
linhuang-blockone: Agreed.

libraries/chain/block_log.cpp
linhuang-blockone: Agreed. This could be a potential problem if ftell returned -1. Thanks.

libraries/chain/block_log.cpp
linhuang-blockone: Agreed.

libraries/chain/block_log.cpp
linhuang-blockone: Agreed. This makes code cleaner.

larryk85:
Lin created a different PR for this.

PR #9032, opened 2020-04-30
add options in fc for not using GMP and for static linking GMP

spoonincode:
## Change Description

This adds two new options for configuring GMP usage in the libsecp256k1 library used by eosio. The defaults are previous behavior: GMP is required and dynamic linked.

The tricky bit with libgmp is that using it markedly improves key recovery performance, but the library is LGPL so public binaries cannot static link to it.

There are two new potentially useful configurations with these options:

`GMP_USAGE=Off` : disable usage of GMP. This may be useful for a "standalone" binary that depends on no other system libraries to be present. Performance of key recovery would suffer a bit but the binary would work whether or not the user has GMP on their box.

`GMP_USAGE=On`, `GMP_STATIC=On`. On a macos hardened build, unsigned dylib usage is not allowed. For users who need to do a macos hardened build, they can still retain the performance benefits of libgmp and not violate LGPL as long as they do not distribute the resulting binary.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
This branch is damaged more than I want to bother with due to submodule smashing. Redoing as #9054

Issue #8970, opened 2020-04-17
APIs should return 400s if the provided input is not parseable


b1bart:
## Overview

In #8899 with commits 954f4bc35 and e03a5d2d5 the HTTP API endpoint added no longer results in a `500` when the parameters passed in as the `POST` body are not parseable as the expected data structure.  Instead, those failures are caught and respond with a `400` to indicate to the user that there is something wrong with their request and not falsely implicate the server application.  

For that PR, we confined the scope of the change to the features added in the PR, however, there are many other APIs that have this same basic issue which should be corrected.

Simply put, if the failure is caused by malformed input by the user then the HTTP error code should be `400`

All existing HTTP APIs should be audited and corrected.

allenhan2:
This one has been fixed

PR #9033, opened 2020-05-01
Adding message body check (400) for http calls

allenhan2:
## Change Description

For Issue: #8970
Add parameters check for all http requests. 

Changes:
apply parameters check for every http calls.
   a) if empty body is not allowed, then throw exception for empty body
   b) if empty body is required, then set empty body as "{}" for string and default for other types.
   c) if empty body is allowed, then returns same as empty body is required for empty body, and returns same as empty is not allowed for non-empty body.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


allenhan2:
@b1bart, just uses chain API for example, could you please help to review it before applying it to other APIs, thanks,

plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
```diff
@@ -157,6 +158,26 @@ namespace eosio {
 
       error_info error;
    };
+
+   enum class http_params_types {
+      no_params_required = 0,
+      params_required = 1
+   };
+
+   template<typename T, http_params_types require_params>
```
b1bart, line 167: I am unconvinced that its appropriate to pass this in as a template parameter.  However, if there is good reason for that then the conditional should use `if constexpr` as the outer conditional with the body check inside that to clearly delineate where the compile time behavior change(s) are expected.

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -99,12 +82,12 @@ namespace {
    }\
 }
 
-#define CHAIN_RO_CALL(call_name, http_response_code) CALL(chain, ro_api, chain_apis::read_only, call_name, http_response_code)
-#define CHAIN_RW_CALL(call_name, http_response_code) CALL(chain, rw_api, chain_apis::read_write, call_name, http_response_code)
+#define CHAIN_RO_CALL(call_name, http_response_code, require_params) CALL_WITH_400(chain, ro_api, chain_apis::read_only, call_name, http_response_code, require_params)
+#define CHAIN_RW_CALL(call_name, http_response_code) CALL(chain, rw_api, chain_apis::read_write, call_name, http_response_code, false)
```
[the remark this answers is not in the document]
allenhan2: Yeah, I was back and forth on using a parameter or template. Eventually decide to use template, so I will remove the parameter.

plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
allenhan2: The template enforce the type and can do compile time optimization. Good point to make it constexpr. Is that okay, if I continue with template?

tests/CMakeLists.txt
```diff
@@ -162,6 +163,10 @@ set_property(TEST nodeos_repeat_transaction_lr_test PROPERTY LABELS long_running
 
 add_test(NAME cli_test COMMAND tests/cli_test.py WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 
+add_test(NAME plugin_http_api_test COMMAND tests/plugin_http_api_test.py -v --clean-run --dump-error-detail WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+set_tests_properties(plugin_http_api_test PROPERTIES TIMEOUT 3000)
+set_property(TEST plugin_http_api_test PROPERTY LABELS plugin_http_api_tests)
```
kj4ezj, line 168: Applying this label will create a new test group called `plugin_http_api_tests`, which the CI system will have to be modified to know to look for. Instead of creating a new test group, please add your test to one of the existing test groups:
- No label for unit/parallelizable tests
- `nonparallelizable_tests` for tests that run fast (<7 min) but have conflicts with other tests (ports, files, PIDs, etc) and cannot be parallelized
- `long_running_tests` for tests that take a long time to run
As a general rule of thumb, new tests should never be the longest-running test in their test group.

tests/CMakeLists.txt
```diff
@@ -162,6 +163,10 @@ set_property(TEST nodeos_repeat_transaction_lr_test PROPERTY LABELS long_running
 
 add_test(NAME cli_test COMMAND tests/cli_test.py WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 
+add_test(NAME plugin_http_api_test COMMAND tests/plugin_http_api_test.py -v --clean-run --dump-error-detail WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+set_tests_properties(plugin_http_api_test PROPERTIES TIMEOUT 3000)
```
kj4ezj, line 167: You mentioned in an IM that your test takes about 20 seconds to run.  
A timeout of 3,000 seems excessive for a test which takes 20 seconds to run. When choosing a timeout, consider that you have exceptional hardware and it might take longer for a community contributor to run the same test.

unittests/CMakeLists.txt
```diff
@@ -47,7 +47,8 @@ target_include_directories( unit_test PUBLIC
                             ${CMAKE_BINARY_DIR}/contracts
                             ${CMAKE_CURRENT_SOURCE_DIR}/contracts
                             ${CMAKE_CURRENT_BINARY_DIR}/contracts
-                            ${CMAKE_CURRENT_BINARY_DIR}/include )
+                            ${CMAKE_CURRENT_BINARY_DIR}/include
+                            ${CMAKE_SOURCE_DIR}/plugins/http_plugin/include )
```
kj4ezj, line 51: Should this line go [here](https://github.com/EOSIO/eos/blob/e1b4f6d5218b8fb72476951af2c02728cbe50da0/tests/CMakeLists.txt#L13)?

allenhan2: yes, http_plugin/include  is required for unit-testing function "parse_params" in http_plugin.hpp

tests/CMakeLists.txt
allenhan2: changed to 40 seconds,

tests/CMakeLists.txt
allenhan2: set as "nonparallelizable_tests"

allenhan2:
@b1bart, could you please review the fix, thanks!

PR #9041, opened 2020-05-05
move minimum boost from 1.67->1.70; gcc 7->8

spoonincode:
## Change Description

Accepted fallout from #9018

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,8 +38,8 @@ set( TESTER_EXECUTABLE_NAME eosio-tester )
 
 # http://stackoverflow.com/a/18369825
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
-        message(FATAL_ERROR "GCC version must be at least 7.0!")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -98,7 +98,7 @@ endif()
 set( Boost_USE_STATIC_LIBS ON CACHE STRING "ON or OFF" )
 # Most boost deps get implictly picked up via fc, as just about everything links to fc. In addition we pick up
 # the pthread dependency through fc.
-find_package(Boost 1.67 REQUIRED COMPONENTS program_options unit_test_framework)
+find_package(Boost 1.70 REQUIRED COMPONENTS program_options unit_test_framework)
```
heifner, line 101: What about `EosioTester.cmake.in` & `EosioTesterBuild.cmake.in`

spoonincode: Yeah, I guess those should be "fixed", though I was probably subconsciously ignoring them because #9042. But I'm not sure how eagerly we can take that.

heifner: I have no idea. Maybe input from @larryk85 ?

PR #9020, opened 2020-04-28
add help text to wasm-runtime - develop

kimjh2005:
#8945 Help text for `--wasm-runtime` does not show available
## Change Description

This change adds available wasm-runtime options in nodeos help.
nodeos --help

   Old help text
  --wasm-runtime runtime                Override default WASM runtime

 New help text
  --wasm-runtime runtime (=eos-vm-jit) Override default WASM runtime (
"eos-vm-jit", "eos-vm", ["eos-vm-oc"]) "eos-vm-jit" : A WebAssembly runtime
that compiles WebAssembly code to native x86 code prior to execution.
"eos-vm" : A WebAssembly interpreter. 
 "eos-vm-oc" :  Unsupported. Instead, use one of the other runtimes along with the option enable-eos-vm-oc

*Note: The option and the  text for "eos-vm-oc" only appears in help text when EOSIO_EOS_VM_DEVELOPER is defined.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

PR #9063, opened 2020-05-07
add help text to wasm-runtime - 2.0.x

kimjh2005:
Created PR for release/2.0.x from  PR(develop)
https://github.com/EOSIO/eos/pull/9020

## Change Description

nodeos --help

Old help text
--wasm-runtime runtime Override default WASM runtime

New help text
--wasm-runtime runtime (=wabt)        Override default WASM runtime ( "wabt",
                                          "eos-vm-jit", "eos-vm"[, "eos-vm-oc"])
                                        "wabt" : The WebAssembly Binary 
                                        Toolkit.
                                        "eos-vm-jit" : A WebAssembly runtime 
                                        that compiles WebAssembly code to 
                                        native x86 code prior to execution.
                                        "eos-vm" : A WebAssembly interpreter."eos-vm-oc" : Unsupported. Instead, use one of the other runtimes along with the option enable-eos-vm-oc

*Note: The text for "eos-vm-oc" only appears in help text when EOSIO_EOS_VM_DEVELOPER is defnend.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -189,12 +189,35 @@ chain_plugin::chain_plugin()
    app().register_config_type<eosio::chain::db_read_mode>();
    app().register_config_type<eosio::chain::validation_mode>();
    app().register_config_type<chainbase::pinnable_mapped_file::map_mode>();
+   app().register_config_type<eosio::chain::wasm_interface::vm_type>();
 }
 
 chain_plugin::~chain_plugin(){}
 
 void chain_plugin::set_program_options(options_description& cli, options_description& cfg)
 {
+   // build wasm_runtime help text
+   std::string wasm_runtime_opt = "Override default WASM runtime (";
+   std::string wasm_runtime_desc;
+   std::string delim;
+#ifdef EOSIO_EOS_VM_JIT_RUNTIME_ENABLED
+   wasm_runtime_opt += " \"eos-vm-jit\"";
+   wasm_runtime_desc += "\"eos-vm-jit\" : A WebAssembly runtime that compiles WebAssembly code to native x86 code prior to execution.\n";
+   delim = ", ";
+#endif
+
+#ifdef EOSIO_EOS_VM_RUNTIME_ENABLED
+   wasm_runtime_opt += delim + "\"eos-vm\"";
+   wasm_runtime_desc += "\"eos-vm\" : A WebAssembly interpreter.\n";
+   delim = ", ";
+#endif
+
+#ifdef EOSIO_EOS_VM_OC_DEVELOPER
+   wasm_runtime_opt += delim + "\"eos-vm-oc\"";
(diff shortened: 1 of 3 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -209,7 +232,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
                EOS_ASSERT(false, plugin_exception, "");
             }
 #endif
-         }), "Override default WASM runtime")
+         })->default_value(eosio::chain::wasm_interface::vm_type::eos_vm_jit, "eos-vm-jit"), wasm_runtime_opt.c_str()
```
spoonincode, line 235: For 2.0 the default wasm runtime needs to remain wabt. Actually that's curious, why can't we use  `chain::config::default_wasm_runtime` here to keep the default defined in a single location?

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -189,12 +189,35 @@ chain_plugin::chain_plugin()
...
    app().register_config_type<eosio::chain::validation_mode>();
    app().register_config_type<chainbase::pinnable_mapped_file::map_mode>();
+   app().register_config_type<eosio::chain::wasm_interface::vm_type>();
 }
 
 chain_plugin::~chain_plugin(){}
 
 void chain_plugin::set_program_options(options_description& cli, options_description& cfg)
 {
+   // build wasm_runtime help text
```
spoonincode, line 199: in 2.0 there is another run time, wabt. We'll need to come up with proper description for it :innocent:

plugins/chain_plugin/chain_plugin.cpp
kimjh2005: used chain::config::default_wasm_runtime.

plugins/chain_plugin/chain_plugin.cpp
kimjh2005: changed it as wabt.
"wabt" : The WebAssembly Binary Toolkit.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -209,7 +237,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
                EOS_ASSERT(false, plugin_exception, "");
             }
 #endif
-         }), "Override default WASM runtime")
+         })->default_value(eosio::chain::config::default_wasm_runtime, default_wasm_runtime_str), wasm_runtime_opt.c_str()
```
spoonincode, line 240: I suspect the reason you need a `default_wasm_runtime_str` here is because only `operator>>` is defined for vm_type and not a `operator<<` (see libraries/chain/wasm_interface.cpp). I suspect if you were to move the `vm_type_string()` logic over to a `operator<<` there would be no need for `default_wasm_runtime_str`

If so, this is still just a minor issue and not something I think absolutely must be done.

PR #9066, opened 2020-05-08
separate out signature provider from producer plugin

spoonincode:
## Change Description

Currently the only time nodeos needs to perform a signature is when signing a block. Thus defining keys and where to find the private key for those keys has been in producer plugin.

Going forward though, this will not be the only use case for signing in nodeos. For example, a witness needs to sign blocks as well.

Move the signature provider code out of producer plugin to its own plugin so that any plugin can leverage signing through a single impl.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -720,7 +720,7 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
       const std::vector<std::string> key_spec_pairs = options["signature-provider"].as<std::vector<std::string>>();
       for (const auto& key_spec_pair : key_spec_pairs) {
          try {
-            auto [pubkey, provider] = app().get_plugin<signature_provider_plugin>().signature_provider_for_specification(key_spec_pair);
+            const auto& [pubkey, provider] = app().get_plugin<signature_provider_plugin>().signature_provider_for_specification(key_spec_pair);
             my->_signature_providers[pubkey] = provider;
          } catch(secure_enclave_exception& e) {
             elog("Error with Secure Enclave signature provider: ${e}; ignoring ${val}", ("e", e.top_message())("val", key_spec_pair));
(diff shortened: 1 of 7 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -781,28 +720,8 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
...
-               my->_signature_providers[pubkey] = make_key_signature_provider(private_key_type(spec_data));
-            } else if (spec_type_str == "KEOSD") {
-               my->_signature_providers[pubkey] = make_keosd_signature_provider(my, spec_data, pubkey);
-            }
-#ifdef __APPLE__
-            else if (spec_type_str == "SE") {
-               my->_signature_providers[pubkey] = make_se_signature_provider(pubkey);
-            }
-#endif
+            auto [pubkey, provider] = app().get_plugin<signature_provider_plugin>().signature_provider_for_specification(key_spec_pair);
```
larryk85, line 723: I would make this ```const auto&```.  The compiler should optimize what you have, but if not then it is two non-trivial copies, and one more copy for the provider.

plugins/signature_provider_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,10 @@
+file(GLOB HEADERS "include/eosio/signature_provider_plugin/*.hpp")
+add_library( signature_provider_plugin
+             signature_provider_plugin.cpp
+             ${HEADERS} )
+
+target_link_libraries( signature_provider_plugin appbase fc http_client_plugin )
+target_include_directories( signature_provider_plugin PUBLIC include )
```
larryk85, line 7: Shouldn't this be PRIVATE and not PUBLIC.

spoonincode: I think it needs to be PUBLIC because, for example, producer_plugin that uses this has to `#include <eosio/signature_provider_plugin/signature_provider_plugin.hpp>`

larryk85: Yeah, you are right. Sorry knee jerk.  I've been looking at the EOS code base over the last week and looking at all of the places of the over use of PUBLIC.  It's the reason why we can't simply use CMake PACKAGE without it barfing all of the libs and includes everywhere.

larryk85: It is one of the reasons, let me put it that way.

PR #9072, opened 2020-05-11
CFD pruning integration test

huangminghuang:
## Change Description
This test creates a producer node and 2 verification nodes. 
The producer node is configured with state history plugin.
A transaction with context free data is pushed to the producer node.
Both validation nodes are stop before the context free data has been pushed. 
Next, it uses eosio-blocklog to prune the transaction with context free
data, restart the producer node and check if the context free data in the block 
log has actually been pruned. 

Afterwords, both validation nodes are restarted and checks whether the light validation
node keeps syncing after the block with context free data while the full validation node
stops syncing right at the block.  

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


nksanthosh:
Very nice work Huang-Ming. Thanks

PR #9091, opened 2020-05-14
[docs] Add cleos validate signatures command reference

lparisc:
## Change Description

Adds documentation additions referenced in PR #9068 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/docs/02_cleos/03_command-reference/validate/signatures.md
+++ b/docs/02_cleos/03_command-reference/validate/signatures.md
@@ -2,7 +2,7 @@
 Validate signatures and recover public keys
 
 ## Positional Parameters
-- `transaction` _TEXT_ - The JSON string or filename defining the transaction to validate
+- `transaction` _TEXT_ - The JSON string or filename defining the signed transaction to validate
 
 ## Options
 - `-c,--chain-id` - The chain id that will be used in signature verification
--- a/docs/02_cleos/03_command-reference/validate/signatures.md
+++ b/docs/02_cleos/03_command-reference/validate/signatures.md
@@ -1,0 +1,34 @@
+## Description
+Validate signatures and recover public keys
+
+## Positional Parameters
+- `transaction` _TEXT_ - The JSON string or filename defining the transaction to validate
+
+## Options
+- `-c,--chain-id` - The chain id that will be used in signature verification
+
+## Example
+```sh
+cleos validate signatures '{ \
+  "expiration": "2018-08-02T20:24:36", \
+  "ref_block_num": 14207, \
+  "ref_block_prefix": 1438248607, \
+  "max_net_usage_words": 0, \
+  "max_cpu_usage_ms": 0, \
+  "delay_sec": 0, \
+  "context_free_actions": [], \
+  "actions": [{ \
+      "account": "eosio", \
+      "name": "newaccount", \
+      "authorization": [{ \
+          "actor": "eosio", \
+          "permission": "active" \
+        } \
+      ], \
+      "data": \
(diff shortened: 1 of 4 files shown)


docs/02_cleos/03_command-reference/validate/signatures.md
```diff
@@ -0,0 +1,46 @@
...
+    "SIG_K1_Jy81u5yWSE4vGET1cm9TChKrzhAz4QE2hB2pWnUsHQExGafqhVwXtg7a7mbLZwXcon8bVQJ3J5jtZuecJQADTiz2kwcm7c"
+  ],
+  "context_free_data": []
+}
+-->
+```sh
+cleos -u https://api.testnet.eos.io validate signatures '{ "expiration": "2020-04-23T04:47:23", "ref_block_num": 20, "ref_block_prefix": 3872940040, 
+"max_net_usage_words": 0, "max_cpu_usage_ms": 0, "delay_sec": 0, "context_free_actions": [], "actions": [ { "account": "eosio", "name": "voteproducer", "authorization": [ { "actor": "initb", "permission": "active" } ], "data": "000000008093dd74000000000000000001000000008093dd74" } ], "transaction_extensions": [], "signatures": [ "SIG_K1_Jy81u5yWSE4vGET1cm9TChKrzhAz4QE2hB2pWnUsHQExGafqhVwXtg7a7mbLZwXcon8bVQJ3J5jtZuecJQADTiz2kwcm7c" ], "context_free_data": [] }'
+```
+
```
arhag, line 44: Do you not want to put the JSON output of this example here?
```json
[
  "EOS7pCywBCz5zw2bc7teCVcT7MEWUr9s749qnYDNPEsBoH32vGqqN"
]
```

lparisc: Yes indeed, I just added the output. I also missed the transaction keyword and removed the url, which was unnecessary.

arhag: It isn't a `transaction` keyword. That is just a placeholder for the actual transaction. Now that you added the word `transaction` the command no longer works.

Also, the a URL to a valid nodeos API node is necessary unless you specify the optional `--chain-id` parameter.

So I would recommend either keeping that URL or alternatively adding `--chain-id cf057bbfb72640471fd910bcb67639c22df9f92470936cddc1ade0e2f2e7dc4f` to the example command.

docs/02_cleos/03_command-reference/validate/validate-signatures.md
```diff
@@ -49,3 +49,10 @@ cleos validate signatures transaction '{ "expiration": "2020-04-23T04:47:23", "r
 
 ## See Also
 - [How to submit a transaction](../../02_how-to-guides/how-to-submit-a-transaction.md)
+
+## Output
+```console
```
arhag, line 54: By the way, when do you use `console` here and when do you use `json`. I notice some other examples use `json` when the output is JSON.

docs/02_cleos/03_command-reference/validate/signatures.md
lparisc: Ah sorry I will fix it back - just realized `transaction` is used as an option not a subcommand as in `cleos sign transaction` or `cleos push transaction`. I will add both examples, w/ and w/o the chain id.

lparisc: Done

docs/02_cleos/03_command-reference/validate/validate-signatures.md
lparisc: When it is the output of a command we should use `console` since it could mix JSON with regular output. In this case we could use either `json` or `console`.

arhag: Probably a good idea to keep `console` here then. Because if no `--chain-id` is provided it will print the following to stderr:
```console
info  2020-05-14T23:27:08.309 thread-0  main.cpp:2647                 operator()           ] grabbing chain_id from nodeos
```

Issue #9056, opened 2020-05-07
In building develop branch, make -j16 is too demanding for an 8-core i9 CPU with a 16GB memory


b1aafulei:
https://github.com/EOSIO/eos/blob/44e216b484ad2fdf3367a8ff5bb9f604ff7048c2/scripts/.build_vars#L43

https://github.com/EOSIO/eos/blob/44e216b484ad2fdf3367a8ff5bb9f604ff7048c2/scripts/helpers/general.sh#L75

The current `JOBS` as in the `make -j${JOBS}` in the build script is determined by the code listed above.

In short, `JOBS` is the lower of 

1. the number of logical CPU cores
2. the size of main memory in GB

However, in building the current `develop` branch, this configuration is too demanding, at least for my case. My Macbook Pro has an 8-core i9-9880H CPU with a 16GB memory, so it counts as having 16 logical cores, and thus `JOBS=16`.

Now the `eosio_build.sh` **consistently** freezes my Mac (for four times) when it arrives at 

```
[ 77%] Building CXX object libraries/chain/CMakeFiles/eosio_chain.dir/webassembly/privileged.cpp.o
```

I am attaching a screenshot of the Activity Monitor at the freezing moment.

![clang](https://user-images.githubusercontent.com/51936236/81271594-e4600b00-907e-11ea-8d59-c5b171cd1a50.png)

## Workaround

The workaround is to terminate the build script before it comes to the compilation, go to the `build` folder, and manually run `make -j8`.

## Suggestion

Consider changing

https://github.com/EOSIO/eos/blob/44e216b484ad2fdf3367a8ff5bb9f604ff7048c2/scripts/.build_vars#L43

into

```
 export CPU_CORES=$(grep -c ^processor /proc/cpuinfo 2>/dev/null || sysctl -n hw.physicalcpu) 
```

which counts the number of physical CPU cores.

Or maybe other ways to avoid the freezing in the compilation.

dixia:
I second this. The memory consumption is too much that ends up to freeze the compilation

TimothyBanks:
Changes have been merged from the PRs referenced in this issue.

b1aafulei:
> Changes have been merged from the PRs referenced in this issue.

Thanks for the fix. The `developer` branch is fine now, but it seems the `release/2.1.x` branch still experience the problem.

PR #9097, opened 2020-05-15
Update algorithm for determining number of parallel jobs.

TimothyBanks:
For issue:  https://github.com/EOSIO/eos/issues/9056

## Change Description
Updated the heuristic for determining the number of GNU make parallel jobs.  It will now require at least 2 GB per job.

### DIFF
--- a/scripts/.build_vars
+++ b/scripts/.build_vars
@@ -64,7 +64,7 @@ export BUILD_CLANG=${BUILD_CLANG:-false}
 
 export CORE_SYMBOL_NAME=${CORE_SYMBOL_NAME:-SYS}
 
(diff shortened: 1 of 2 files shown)


spoonincode:
If we're going to adjust this on macos builds seems we should adjust it on linux builds too? But doesn't this harm compile performance for users who do have plenty of memory?

It seems like maybe the number should be set based on amount of RAM, capped to the number of processors. i.e. `min(RAM/3GB, ncpu)` or such.

TimothyBanks:
I had misread the /proc/cpuinfo about the process being a logical core.   I can definitely do something based on available memory.

PR #9098, opened 2020-05-15
Initial checkin of resource monior plugin

linhuang-blockone:
## Change Description

This PR resolves https://blockone.atlassian.net/projects/EPE/issues/EPE-43.

**Background**

It was reported from a customer that when file system which `data/blocks` belongs to is
 running out of space, the producer continued to produce blocks and update state but
`blocks.log` was *corrupted* in a sense that it no longer contained all the irreversible blocks.
It was also observed that when file system which `data/state` belongs is running out of space, *nodeos* will crash with SIGBUS as the state file is unable to acquire new pages.

**Design**

The solution is to have a dedicated plugin to monitor resource usages (file system space for now,
CPU, memory, and networking bandwidth in the future). The plugin uses a thread to periodically
check space usage of file systems of directories being monitored. If space used
is over a predefined threshold, a graceful shutdown is initiated.

The directories to be monitored are configured. They can be any paths for of blocks, state,  trace, state history, and others.

**Alternative Solution**

Alternative solution to blocks directory space exhaustion was considered: starting from
`block_log::append`, make sure all exceptions was handled properly. But it was deemed to be
too tedious and error prone to review all possible exceptions, and might likely break consensus. The plugin approach is more generic, monitoring any file systems and without modifying
existing code.

**Additional Bug Fixed**

This PR also fixes a bug in `log_irreversible` of `libraries/chain/controller.cpp`.
`db.commit` was done before `blog.append`. This could be a problem if `blog.append` threw.
Now `db.commit` is done after `blog.append`. Even if `blog.append` throws, DB can be rolled back.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

*resource_monitor_plugin* is a new plugin. It is not must have but recommended to be used.

**Configurable Parameters**

*resource-monitor-directory*
The directory whose file system to be monitored. This is a mandatory  option. It can appear multiple times for different directories. If two directories are on the same file system, the file system will be checked only once.

*resource-monitor-interval-seconds*
Time in seconds between two consecutive checks of resource usage. It should be greater than 1 and less than 60 (one minute). This is optional; the default value is 2 seconds.

*resource-monitor-space-threshold*
Threshold in terms of percentage of used space vs total space. If  used space is over the threshold, a graceful shutdown is initiated. The value should be greater than 1 to less than 99
This is optional; the default value 95 (95%).


plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,228 @@
...
+      for ( auto& dir: dirs ) {
+         add_file_system( dir );
+      }
+   }
+
+   // Start main thread
+   void plugin_startup() {
+      ilog("Creating and starting monitor thread");
+      monitor_thread = std::thread( [this] {
+         fc::set_os_thread_name( "resource_monitor" );
```
heifner, line 91: `console_appender` uses 9 chars for thread name reporting. Recommend "resmon".

matthewdarwin:
I'm not sure if 95% is a good default value for file system space.  By default some file systems reserve 5% for superuser.  So if nodeos is running as a non-privileged user, then writes will fail before the new check will fail. (eg https://ma.ttias.be/change-reserved-blocks-ext3-ext4-filesystem-linux/)

Of course the check is customizable, but maybe more conservative would be better as a default?

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
linhuang-blockone: Thanks Kevin for pointing out this! I will change the name to "resmon".

linhuang-blockone: Thanks Matthew for your comments. Somehow there is no reply under your comment. I was struggling when deciding the default value. Maybe 90% is more reasonable.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,228 @@
...
+class resource_monitor_plugin_impl {
+public:
+
+   void set_program_options(options_description&, options_description& cfg) {
+      cfg.add_options()
+         ( "resource-monitor-directory",  bpo::value<vector<bfs::path>>()->composing(),
+           "The directory whose file system to be monitored. This is a mandatory option. It can appear multiple times for different directories")
+         ( "resource-monitor-interval-seconds", bpo::value<uint32_t>()->default_value(2),
+           "Time in seconds between two consecutive checks of resource usage. Should be greater than 1 and less than 60 (one minute)" )
+         ( "resource-monitor-space-threshold", bpo::value<uint32_t>()->default_value(95),
```
spoonincode, line 57: default is still 95 here, which I think will overwrite the 90 that is intended

linhuang-blockone: Thanks Matt! Sorry I should have used constants for those. Will change them.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,231 @@
...
+   void monitor_task() {
+      ilog("Entering monitor_task");
+
+      while ( true ) {
+         if ( is_any_filesystem_space_below_threshold() ) {
+            appbase::app().quit(); // This will gracefully stop Nodeos
+            break;
+         }
+         
+         std::this_thread::sleep_for( std::chrono::milliseconds( sleep_time_in_ms ));
```
spoonincode, line 138: Won't this cause the thread to block for up to `sleep_time_in_ms` even during shutdown? (I haven't tried so maybe I'm reading it wrong) It seems like if configured for say, 60 seconds, shutdown could be blocked up to 60 seconds?

If so, may want to explore driving this via an asio io_context & a deadline_timer. That will allow you to stop the thread immediately on shutdown.

spoonincode:
I'm a bit concerned this isn't easy enough for users to make use of: it requires them to manually configure the directories to monitor.

IMO we should explore having other plugins register their directories. For example, maybe trace plugin does something like

```c++
if(auto resmon_plugin = app().find_plugin<resource_monitor_pligin>())
  resmon_plugin->monitor_directory(trace_dir);
```

This way the directories that are used are automatically monitored (we could maybe even make resource_monitor_plugin default enabled).

Be aware in the above example we would need to be mindful about plugin initialization and startup order -- trace_api_plugin configures trace_dir during its initialize() but that may be before resource_monitor_plugin's initialize().

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
linhuang-blockone: You are right. If someone configures sleep_time_in_ms to 60 seconds, shutdown would potentially be blocked up to 60 seconds. I am thinking of not allowing sleep time to be configurable and set it to 1 second to simplify code. But I will check out asio io_context and deadline_timer.

linhuang-blockone:
I was thinking of always monitoring anything under "data" directory but then thought the user
might have better idea about file system space configurations. If we choose auto registration by other plugins, is the order of plugins initialization the same as they are specified on the command line? Or can we make the initialization of resource monitor the last so it can get "data_dir", "trace_dir", and any other directories we are interested in? This way we don't need to modify other plugin's code.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+         if ( done ) {
+            break;
+         }
+      }
+   }
+
+   bool is_any_filesystem_space_below_threshold() {
```
b1bart, line 146: Can this be `const`?  Can we also separate this such that downstream effects of running out of space (eg, print a message and quit) are colocated and distinct from the method which checks the thresholds?

This would allow easier changes to how we handle this in the future and more precise testing efforts.

linhuang-blockone: Thanks Bart! I will print "quit" at the place I initiate quit, and remove "quit" print in is_any_filesystem_space_below_threshold. I will also think about ways to improve testability for future additions.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,231 @@
...
+            wlog( "Unable to get space info for ${path_name}: [code: ${ec}] ${message}",
+               ("path_name", fs.path_name.string())
+               ("ec", ec.value())
+               ("message", ec.message()));
+
+            continue;
+         }
+
+         if (info.available < fs.min_available_bytes) {
+            wlog("Space usage on ${path}'s file system exceeded configured threshold ${threshold}%, initiate shuting down!!! available: ${available}, Capacity: ${capacity}, min_available_bytes: ${min_available_bytes}", ("path", fs.path_name.string()) ("threshold", used_space_threshold_in_percentage) ("available", info.available) ("capacity", info.capacity) ("min_available_bytes", fs.min_available_bytes));
```
larryk85, line 164: This should be something like "initiating shut down", or simply "shutting down".

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,231 @@
...
+         "Unable to get space info for ${path_name}: [code: ${ec}] ${message}",
+         ("path_name", path_name.string())
+         ("ec", ec.value())
+         ("message", ec.message()));
+
+      auto min_available_bytes = (100 - used_space_threshold_in_percentage) * (info.capacity / 100); // (100 - used_space_threshold_in_percentage) is the percentage of minimum number of available bytes the file system must maintain  
+
+      // Add to the list
+      filesystem_info fs = { statbuf.st_dev, min_available_bytes, path_name };
+      filesystems.push_back(fs);
```
larryk85, line 205: I would use ```filesystems.emplace_back(statbuf.st_dev, min_available_bytes, path_name);``` this would ensure that no copies are made.

linhuang-blockone: Thanks Bucky for your comments! I will incorporate them together with restructuring to address Bart's comments.

spoonincode:
>If we choose auto registration by other plugins, is the order of plugins initialization the same as they are specified on the command line? Or can we make the initialization of resource monitor the last so it can get "data_dir", "trace_dir", and any other directories we are interested in?

The specifics will have to be investigated and studied. I believe the plugin object gets instantiated as a global constructor so _all_ plugins -- even plugins not enabled -- will be available via `get_plugin<>()` as soon as `main()` starts. That is a little bizarre and maybe not what appbase originally intended. Regardless, it means something like

```c++
if(auto resmon_plugin = app().find_plugin<resource_monitor_pligin>())
  resmon_plugin->monitor_directory(trace_dir);
```
may call in to resource_monitor_plugin before `resource_monitor_plugin::initialize()`. Or it might call in to it after  `resource_monitor_plugin::initialize()`. resource_monitor_plugin would have to be prepared to handle both cases

heifner:
What you do know is at time of `plugin_startup`, all `plugin_initialize` have been called.

linhuang-blockone:
I think I can add a function to be called by resource monitor's plugin_startup.
The function will read other's plugin's configured directories we want to monitor plus
app().data_dir(). This way we don't need to change other plugin's code.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -13,45 +13,41 @@
...
 #include <sys/stat.h>
 
+using namespace eosio::resource_monitor;
+
 namespace bfs = boost::filesystem;
 
 namespace eosio {
    static appbase::abstract_plugin& _resource_monitor_plugin = app().register_plugin<resource_monitor_plugin>();
 
+   static vector<bfs::path> directories_monitored;
```
heifner, line 40: This should be a member of `resource_monitor_plugin_impl` and not a static.
`monitor_directory()` can find the plugin and if available register the directory.

plugins/resource_monitor_plugin/system_file_space_provider.cpp
```diff
@@ -0,0 +1,19 @@
+#include <eosio/resource_monitor_plugin/system_file_space_provider.hpp>
+
+//#include <boost/filesystem.hpp>
+
+//#include <sys/stat.h>
```
heifner, line 5: Remove commented out code

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,143 @@
...
+      return 3;
+   }
+
+   private:
+      Space_provider space_provider;
+
+      boost::asio::deadline_timer timer;
+   
+      uint32_t sleep_time_in_secs;
+      uint32_t threshold;
```
heifner, line 123: init these to 0 here.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+         // Add to the list
+         filesystems.emplace_back(statbuf.st_dev, available_required, path_name);
+         
+         ilog("${path_name}'s file system monitored. available_required: ${available_required}, capacity: ${capacity}, threshold: ${threshold}", ("path_name", path_name.string()) ("available_required", available_required) ("capacity", info.capacity) ("threshold", threshold) );
+      }
+   
+   int space_monitor_loop() {
+      if ( is_threshold_exceeded() ) {
+         wlog("Shut down");
```
heifner, line 94: Question: Should stopping be made an option of the plugin?
Also should it start logging a warning at some threshold before shutdown threshold?

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,143 @@
...
+            return 1;
+         } else {
+           // Loop over
+           space_monitor_loop();
+         }
+
+         return 2;
+      });
+
+      return 3;
```
heifner, line 114: What do these return values mean? It doesn't look like they are ever used.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+      space_handler.set_sleep_time(sleep_time);
+      ilog("sleep_time = ${sleep_time}", ("sleep_time", sleep_time));
+
+      auto threshold = options.at("resource-monitor-space-threshold").as<uint32_t>();
+      EOS_ASSERT(threshold > 0 && threshold < 100, chain::plugin_config_exception,
+         "\"resource-monitor-space-threshold\" must be greater than 0 and less than 100");
+
+      space_handler.set_threshold(threshold);
+      ilog("threshold = ${threshold}", ("threshold", threshold));
```
heifner, line 73: This should be more descriptive if an info level log or made debug level.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,157 @@
...
+      cfg.add_options()
+         ( "resource-monitor-interval-seconds", bpo::value<uint32_t>()->default_value(def_sleep_time_in_secs),
+           "Time in seconds between two consecutive checks of resource usage. Should be greater than 1 and less than 60 (one minute)" )
+         ( "resource-monitor-space-threshold", bpo::value<uint32_t>()->default_value(def_used_space_threshold_in_percentage),
+             "Threshold in terms of percentage of used space vs total space. If used space is over the threshold, a graceful shutdown is initiated. The value should be greater than 1 to less than 99" )
+         ;
+   }
+
+   void plugin_initialize(const appbase::variables_map& options) {
+      ilog("plugin_initialize");
```
heifner, line 59: I think this should be debug level logging.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,157 @@
...
+         "\"resource-monitor-interval-seconds\" must be greater than 0 and less than 60 (1 minute)");
+
+      space_handler.set_sleep_time(sleep_time);
+      ilog("sleep_time = ${sleep_time}", ("sleep_time", sleep_time));
+
+      auto threshold = options.at("resource-monitor-space-threshold").as<uint32_t>();
+      EOS_ASSERT(threshold > 0 && threshold < 100, chain::plugin_config_exception,
+         "\"resource-monitor-space-threshold\" must be greater than 0 and less than 100");
+
+      space_handler.set_threshold(threshold);
```
heifner, line 72: Consider moving the assert into the `set_threshold` method so you don't have to guard against overflow in your calculation.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -689,6 +691,7 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
             my->blocks_dir = app().data_dir() / bld;
          else
             my->blocks_dir = bld;
+         resource_monitor_plugin::monitor_directory(my->blocks_dir);
```
heifner, line 694: `blocks_dir` and `state_dir` need to be added even if not explicitly specified.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -699,6 +702,7 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
             protocol_features_dir = app().config_dir() / pfd;
          else
             protocol_features_dir = pfd;
+         resource_monitor_plugin::monitor_directory(protocol_features_dir);
```
heifner, line 705: Do we write anything to `protocol_features_dir` after startup? I don't think we need to monitor it.

spoonincode:
Should the plugin be default enabled? Seems like now that it is monitoring all the directories automatically that would be desirable?

programs/nodeos/CMakeLists.txt
```diff
@@ -61,6 +61,7 @@ target_link_libraries( ${NODE_EXECUTABLE_NAME}
         PRIVATE -Wl,${whole_archive_flag} txn_test_gen_plugin        -Wl,${no_whole_archive_flag}
         PRIVATE -Wl,${whole_archive_flag} db_size_api_plugin         -Wl,${no_whole_archive_flag}
         PRIVATE -Wl,${whole_archive_flag} producer_api_plugin        -Wl,${no_whole_archive_flag}
+        PRIVATE -Wl,${whole_archive_flag} resource_monitor_plugin    -Wl,${no_whole_archive_flag}
```
heifner, line 64: You don't need `whole_archive_flag` for `resource_monitor_plugin` since you include and use it in main.

plugins/resource_monitor_plugin/test/test_plugin.sh
```diff
@@ -0,0 +1,105 @@
+#!/bin/bash
```
heifner, line 1: Please use python for tests.

plugins/resource_monitor_plugin/test/test_monitor_loop.cpp
```diff
@@ -0,0 +1,430 @@
...
+
+      BOOST_TEST(finished_in_time == true);
+   }
+
+   BOOST_FIXTURE_TEST_CASE(one_loop_59_sec_interval, space_handler_fixture)
+   {
+      auto constexpr num_loops = 1;
+      auto constexpr interval  = 59;
+
+      mock_get_space = [ i = 0 ]( const bfs::path& p, boost::system::error_code& ec) mutable -> bfs::space_info {
```
heifner, line 386: There is a lot of duplication of code for `mock_get_space` pull out into one reusable lambda.

plugins/resource_monitor_plugin/test/test_monitor_loop.cpp
```diff
@@ -0,0 +1,430 @@
...
+         } else {
+            rc.available = 100000;
+         }
+
+         i++;
+
+         return rc;
+      };
+
+      mock_get_stat = []( const char *path, struct stat *buf ) -> int {
```
heifner, line 403: This is repeated in each test. Please factor out all this copy/pasted code into something easier to maintain.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,157 @@
...
+   // System is shutting down.
+   void plugin_shutdown() {
+      ilog("shutdown...");
+
+      ctx.stop();
+
+      // Wait for the thread to end
+      monitor_thread.join();
+
+      ilog("exit shutdown");
```
heifner, line 113: Make these debug level.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
linhuang-blockone: Will put out a warning at each percentage level 5% before shutdown.

Will make shutdown an option.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
linhuang-blockone: Will do.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
linhuang-blockone: Thank you so much Kevin for detailed comments! I will incorporate them.

Will use find_plugin to get the API.

plugins/resource_monitor_plugin/system_file_space_provider.cpp
linhuang-blockone: Will do.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
linhuang-blockone: I use those return values in unit testing so the test cases know where the function returns.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
linhuang-blockone: Will make it more descriptive.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
linhuang-blockone: Agreed.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
linhuang-blockone: Agreed.

plugins/chain_plugin/chain_plugin.cpp
linhuang-blockone: Agreed.

programs/nodeos/CMakeLists.txt
linhuang-blockone: I don't understand whole_archive_flag here. Should I change it to no_whole_archive_flag?

plugins/chain_plugin/chain_plugin.cpp
linhuang-blockone: Will remove.

plugins/resource_monitor_plugin/test/test_monitor_loop.cpp
linhuang-blockone: Agreed. Refactored.

plugins/resource_monitor_plugin/test/test_plugin.sh
linhuang-blockone: Will do.

plugins/resource_monitor_plugin/test/test_monitor_loop.cpp
```diff
@@ -0,0 +1,163 @@
...
+   std::function<bfs::space_info(const bfs::path& p, boost::system::error_code& ec)> mock_get_space;
+   std::function<int(const char *path, struct stat *buf)> mock_get_stat;
+
+   file_space_handler_t space_handler;
+};
+
+BOOST_AUTO_TEST_SUITE(monitor_loop_tests)
+   BOOST_FIXTURE_TEST_CASE(zero_loop, space_handler_fixture)
+   {
+      BOOST_TEST(test_loop_common(0, 1)  == true);
```
larryk85, line 120: No need for the ```== true```.  BOOST_TEST is assuming a boolean value.

plugins/resource_monitor_plugin/test/test_threshold.cpp
```diff
@@ -0,0 +1,285 @@
...
+   file_space_handler_t space_handler;
+};
+
+BOOST_AUTO_TEST_SUITE(threshol_tests)
+   BOOST_FIXTURE_TEST_CASE(equal_to_threshold, threshold_fixture)
+   {
+      std::map<bfs::path, uintmax_t> availables {{"/test0", 200000}};
+      std::map<bfs::path, int>       devs       {{"/test0", 0}};
+
+      BOOST_TEST(test_threshold_common(availables, devs) == false);
```
larryk85, line 95: You just negate the expression.

plugins/resource_monitor_plugin/test/test_add_file_system.cpp
```diff
@@ -0,0 +1,144 @@
...
+      BOOST_REQUIRE_THROW(add_file_system("/test"), chain::plugin_config_exception);
+   }
+
+   BOOST_FIXTURE_TEST_CASE(different_file_systems, add_file_system_fixture)
+   {
+      std::vector<int> capacity  {1000000, 2000000, 3000000, 4000000};
+      std::vector<int> available {500000,  1500000, 2500000, 3500000};
+      std::vector<int> devs      {0,       1,       2,       3};
+
+      test_add_file_systems_common(capacity, available, devs);
```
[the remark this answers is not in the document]
linhuang-blockone: Agreed. Thanks for pointing this out.

plugins/resource_monitor_plugin/test/test_add_file_system.cpp
```diff
@@ -0,0 +1,144 @@
...
+   }
+
+   BOOST_FIXTURE_TEST_CASE(same_file_system, add_file_system_fixture)
+   {
+      std::vector<int> capacity  {1000000, 2000000, 3000000, 4000000};
+      std::vector<int> available {500000,  1500000, 2500000, 3500000};
+      std::vector<int> devs      {0,       0,       0,       0};
+
+      // As long as no exceptions, it is considered success.
+      test_add_file_systems_common(capacity, available, devs);
```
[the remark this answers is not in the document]
linhuang-blockone: Will change.

plugins/resource_monitor_plugin/test/test_add_file_system.cpp
```diff
@@ -0,0 +1,144 @@
...
+   }
+
+   BOOST_FIXTURE_TEST_CASE(mixed_file_systems, add_file_system_fixture)
+   {
+      std::vector<int> capacity  {1000000, 2000000, 3000000, 4000000, 50000};
+      std::vector<int> available {500000,  1500000, 2500000, 3500000, 20000};
+      std::vector<int> devs      {0,       2,       2,       0,       3};
+
+      // As long as no exceptions, it is considered success.
+      test_add_file_systems_common(capacity, available, devs);
```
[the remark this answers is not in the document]
linhuang-blockone: Will change.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,165 @@
...
+      bool is_threshold_exceeded() const {
+         // Go over each monitored file system
+         for (auto& fs: filesystems) {
+            boost::system::error_code ec;
+            auto info = space_provider.get_space(fs.path_name, ec);
+            if ( ec ) {
+               // As the system is running and this plugin is not a critical
+               // part of the system, we should not exit.
+               // Just report the failure and continue;
+               wlog( "Unable to get space info for ${path_name}: [code: ${ec}] ${message}. Ignire this failure.",
```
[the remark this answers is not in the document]
linhuang-blockone: Will change.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,165 @@
...
+         struct stat statbuf;
+         auto status = space_provider.get_stat(path_name.string().c_str(), &statbuf);
+         EOS_ASSERT(status == 0, chain::plugin_config_exception,
+                    "Failed to run stat on ${path} with status ${status}", ("path", path_name.string())("status", status));
+
+         // If the file system containing the path is already
+         // in the filesystem list, do not add it again
+         for (auto& fs: filesystems) {
+            if (statbuf.st_dev == fs.st_dev) { // Two files belong to the same file system if their device IDs are the same.
+               ilog("${path_name}'s file system already monitored", ("path_name", path_name.string()));
```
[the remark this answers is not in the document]
linhuang-blockone: Will change.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,165 @@
...
+#include <boost/filesystem.hpp>
+#include <boost/asio.hpp>
+
+#include <appbase/application.hpp>
+#include <eosio/chain/exceptions.hpp>
+
+namespace bfs = boost::filesystem;
+
+namespace eosio::resource_monitor {
+   template<typename Space_provider>
```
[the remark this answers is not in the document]
linhuang-blockone: Will Change.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,164 @@
...
+public:
+   resource_monitor_plugin_impl()
+      :space_handler(system_file_space_provider(), ctx)
+      {
+      }
+
+   void set_program_options(options_description&, options_description& cfg) {
+      cfg.add_options()
+         ( "resource-monitor-interval-seconds", bpo::value<uint32_t>()->default_value(def_sleep_time_in_secs),
+           "Time in seconds between two consecutive checks of resource usage. Should be greater than 1 and less than 300 (five minutes)" )
```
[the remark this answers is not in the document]
linhuang-blockone: Thanks for spotting this.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -0,0 +1,164 @@
...
+         ( "resource-monitor-space-threshold", bpo::value<uint32_t>()->default_value(def_used_space_threshold_in_percentage),
+           "Threshold in terms of percentage of used space vs total space. If used space is over the threshold, a graceful shutdown is initiated. The value should be greater than 5 to less than 99" )
```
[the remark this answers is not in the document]
linhuang-blockone: Sorry about this. Use constant and fix them all.

plugins/resource_monitor_plugin/test/test_monitor_loop.cpp
linhuang-blockone: Wanted to signify actual value and expected value. Will change.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -0,0 +1,165 @@
...
+   
+   int space_monitor_loop() {
+      if ( is_threshold_exceeded() && shutdown_on_exceeded ) {
+         wlog("Shut down");
+         appbase::app().quit(); // This will gracefully stop Nodeos
+         return 0;
+      }
+
+      timer.expires_from_now( boost::posix_time::seconds( sleep_time_in_secs ));
+      timer.async_wait([this](auto ec) {
```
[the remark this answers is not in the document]
linhuang-blockone: Agreed.

programs/nodeos/CMakeLists.txt
larryk85: His won't need the flag, but it's probably good to keep the flag for that plugin in case we change how and what directories it grabs.  If someone removes it from main, then they will spend probably way too much time trying figure out why it no longer works.  It can potentially make the binary slightly bigger, but I wouldn't mark that as a concern of ours.

PR #9101, opened 2020-05-16
[2.0.x] epe-37 fixed test-contracts build

kimjh2005:
Since the removal of those deprecated headers in eosio.cdt v1.7.x,  it is not possible to build EOSIO with -DEOSIO_COMPILE_TEST_CONTRACTS=true with an installed version of eosio.cdt of 1.7.x or later. 
https://blockone.atlassian.net/browse/EPE-37
## Change Description

#1. Changed include path from eosiolib/* to eosiolib/capi/, eosiolib/core/, eosiolib/contracts/*.
#2. Added namespace if call to a function is ambiguous
#3. Handled function arguments changes.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/test-contracts/test_api/test_action.cpp
+++ b/unittests/test-contracts/test_api/test_action.cpp
@@ -1,14 +1,32 @@
-#include <eosiolib/action.hpp>
-#include <eosiolib/chain.h>
-#include <eosiolib/crypto.h>
-#include <eosiolib/datastream.hpp>
-#include <eosiolib/db.h>
-#include <eosiolib/eosio.hpp>
-#include <eosiolib/print.hpp>
-#include <eosiolib/privileged.h>
-#include <eosiolib/transaction.hpp>
+#include <eosiolib/contracts/eosio/action.hpp>
+#include <eosiolib/capi/eosio/chain.h>
+#include <eosiolib/capi/eosio/crypto.h>
+#include <eosiolib/core/eosio/datastream.hpp>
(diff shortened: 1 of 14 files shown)


unittests/test-contracts/test_api/test_action.cpp
```diff
@@ -1,15 +1,33 @@
-#include <eosiolib/action.hpp>
-#include <eosiolib/chain.h>
-#include <eosiolib/crypto.h>
-#include <eosiolib/datastream.hpp>
-#include <eosiolib/db.h>
-#include <eosiolib/eosio.hpp>
-#include <eosiolib/print.hpp>
-#include <eosiolib/privileged.h>
-#include <eosiolib/transaction.hpp>
+#include <eosiolib/contracts/eosio/action.hpp>
```
larryk85, line 1: The way the default includes work is that you can use #include <eosio/action.hpp>.  You don't need the eosiolib/[contracts|core|capi] part.  Also, we shouldn't be including the capi headers, use the c++ equivalent headers.  These will include the capi headers for you.

heifner:
This seems like a lot for `release/2.0.x` are you sure you want to back port this? @nksanthosh @b1bart

kimjh2005:
It doesn't have to be back ported to release/2.0.x.  I will close the PR.

kimjh2005:
I will close this PR as it is too late to merge it to 2.0.x.

PR #9104, opened 2020-05-18
Ship v1

huangminghuang:
## Change Description

* Add get_blocks_result_v1 so that the block field can support new signed_block/trace data structure
* Add partial_transaction_v1 to incorporate pruable_data_type instead of signatures/CFDs
* Change the context free data compression in trace log from always zlib to user configurable and thus adding a compression tag in the trace log disk format.
* Change state history ABI string to support new protocol and add unit test for it
* Modify rodeos to support ship v1
* Add opaque type to describe a skippable field during deserialization.
* Add unit test to make sure the data serialized from state_history library can be deserialized with abieos ship_protocols types.
* Add new fc::datastream specializations for fc::cfile, std::streambuf, std::vector/std::deque.
* Add integration test for rodeos 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


nksanthosh:
Documentation for this functionality needs to be created.

PR #9111, opened 2020-05-18
Update algorithm for determining number of parallel jobs - develop

TimothyBanks:
For issue:  https://github.com/EOSIO/eos/issues/9056

## Change Description
Updated the heuristic for determining the number of GNU make parallel jobs.  It will now require at least 2 GB per job.


TimothyBanks:
@johndebord The builds succeed on this one now.

PR #9117, opened 2020-05-19
Exit on rodeos filter wasm error

heifner:
## Change Description

- Report filter wasm exceptions as errors so it is clear exception is from the filter wasm.
- Add option `clone-exit-on-filter-wasm-error` to `cloner_plugin` which is part of `rodeos` defaults to false. If enabled, it will shutdown rodeos on error.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

- New rodeos option `clone-exit-on-filter-wasm-error` - `Shutdown application if filter wasm throws an exception`

### DIFF
--- a/libraries/rodeos/rodeos.cpp
+++ b/libraries/rodeos/rodeos.cpp
@@ -245,14 +245,24 @@ void rodeos_filter::process(rodeos_db_snapshot& snapshot, const ship_protocol::g
    backend->initialize(&cb);
    try {
       (*backend)(cb, "env", "apply", uint64_t(0), uint64_t(0), uint64_t(0));
+
+      if (!filter_state->console.empty())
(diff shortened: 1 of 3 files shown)


b1bart:
Is retry logic scoped for rodeos before a public alpha/beta release?  or is the long-term expectation that failure equates to "halt and catch fire"

heifner:
> Is retry logic scoped for rodeos before a public alpha/beta release? or is the long-term expectation that failure equates to "halt and catch fire"

@tbfleming

tbfleming:
If a wasm fails, retrying it would get the same result, so it doesn't. Unless there's a determinism bug...

It retries when there's a ship connection failure.

PR #9121, opened 2020-05-21
reliable_amqp_publisher: encapsulate some AMQP reliability needs

spoonincode:
## Change Description

As EOSIO gains more support for AMQP it gains the need in some use cases to reliably publish messages to AMQP exchanges. reliable_amqp_publisher is a small class for publishing messages to an exchange that:

* Reconnects to the AMQP server on connection failure
* Reconnects to the AMQP server on channel failure (such as when publishing to a non-existing exchange)
* Publishes messages in an AMQP transaction for positive confirmation the broker received them
* Queues unconfirmed messages while not connected
* Saves unconfirmed messages on disk on exit, and restores on relaunch

reliable_amqp_queue expects that the exchange it is configured to publish to will exist. It will not attempt to create it.

I am expecting and fully accept grief due to the lack of integration tests. I would like to explore with the team how we can best add integration tests for AMQP.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
```diff
@@ -0,0 +1,236 @@
...
+   my(new reliable_amqp_publisher_impl(url, exchange, unconfirmed_path)) {}
+
+void reliable_amqp_publisher::publish_message(std::vector<char>&& data) {
+   my->ctx.dispatch([this, d=std::move(data)]() mutable {
+      constexpr unsigned max_queued_messages = 1u<<20u;
+      if(my->message_deque.size() > max_queued_messages) {
+         if(my->logged_exceeded_max_depth == false)
+            elog("AMQP connection ${a} publishing to \"${e}\" has reached ${max} unconfirmed messages; further messages will be dropped",
+                 ("a", (std::string)*my->amqp_address)("e", my->exchange)("max", max_queued_messages));
+         my->logged_exceeded_max_depth = true;
```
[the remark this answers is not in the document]
heifner: I could see wiring in an `on_max_depth_error` `std::function` which would allow a user of this to do something different; `app().quit()` for example. We don't necessarily need to over engineer this now, so that could be added later if needed.

libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
```diff
@@ -0,0 +1,236 @@
...
+      if(boost::filesystem::exists(data_file_path)) {
+         try {
+            fc::datastream<fc::cfile> file;
+            file.set_file_path(data_file_path);
+            file.open("rb");
+            fc::raw::unpack(file, message_deque);
+         } FC_RETHROW_EXCEPTIONS(error, "Failed to load previously cached AMQP messages from ${f}", ("f", (fc::path)data_file_path));
+      }
+      boost::system::error_code ec;
+      boost::filesystem::remove(data_file_path, ec);
```
heifner, line 78: I think you should verify that you can create the `data_file_path` before starting so that you are less likely to fail on shutdown because of file permissions or invalid path.

libraries/reliable_amqp_publisher/include/eosio/reliable_amqp_publisher/reliable_amqp_publisher.hpp
```diff
@@ -0,0 +1,52 @@
...
+ * reliable_amqp_publisher's queue until the exchange exists because the queue will be unable to make progress.
+ *
+ * At a large enough uncomfirmed queue depth (currently on the order of 1 million entries) reliable_amqp_publisher
+ * will start dropping messages after logging an error.
+ */
+
+class reliable_amqp_publisher {
+   public:
+      /// Create a reliable queue to the given server. "name" is used as filename for persisting messages
+      ///  so should be unique to usage
```
heifner, line 26: I either don't understand or this comment is not correct. `unconfirmed_path` unmolested is used for the file name. Not sure what "name" refers to.

libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
```diff
@@ -0,0 +1,236 @@
...
+   void pump_queue() {
+      if(stopping || in_flight || !connected || message_deque.empty())
+         return;
+
+      constexpr size_t max_msg_single_transaction = 16;
+      const unsigned msgs_this_transaction = std::min(message_deque.size(), max_msg_single_transaction);
+
+      channel->startTransaction();
+      std::for_each_n(message_deque.begin(), msgs_this_transaction, [this](const std::vector<char>& msg) {
+         channel->publish(exchange, "", msg.data(), msg.size());
```
heifner, line 170: Seems like you should provide an option to setting a routing_key to be used. Leo's publish uses "" for exchange and a configured routing_key.

libraries/reliable_amqp_publisher/include/eosio/reliable_amqp_publisher/reliable_amqp_publisher.hpp
spoonincode: thanks you are right this comment didn't get changed after some refactoring

libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
spoonincode: IMO configuration of the exchange is more useful than the routing_key, but I have no qualm adding support for both.

libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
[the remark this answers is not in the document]
spoonincode: This is a tough one. I can see it argued a number different ways :woman_shrugging: In general, I feel log spam is a bad thing because it'll make it harder to see the trees through the forest. Since this is actually at the error level (not many logs in EOSIO are), it should _really_ stick out in some sort of monitoring solution I hope. Though maybe amqp logs should be on a separate logger so a monitoring system can flag an alarm on _any_ warns & errors from that logger -- really you should never see neither warns nor errors from this class on a healthy running infra.

But yeah, exiting is also a worthy option too.

spoonincode:
Closing this in favor of #9183

PR #9122, opened 2020-05-21
amqp_witness_plugin: sign blocks as non-producer and publish over AMQP

spoonincode:
## Change Description

A _witness_ is, conceptually, a node that signs a block attesting that the entire contents of the block are valid. A producer is a witness. This plugin extends the concept to a non-producing node. When the plugin is activated irreversible blocks are signed and those signatures are published to an AMQP exchange.

An AMQP server and existing exchange must be specified. The plugin is able to use any existing signature provider: a soft key specified in the config, keosd, or a Secure Enclave key. Blocks older than a specified age will not be signed -- useful for sync or replay.

Either the block’s signature digest or simply its action merkle root is reported in the message and what is signed. I believe this is appropriate to identify the block for both applications considered. A stateful application, like a light validator, will know the block signature digest it is seeking witness signatures for. A stateless application, like an application that is only inspecting action receipts, only wishes to know that the action receipt merkle proof for a block satisfies a merkle root that a witness has signed.

This plugin communicates to the AMQP server via the reliable_amqp_publisher for increased publishing reliability.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [X] Documentation Additions

Yes big time; a new plugin and the start of more AMQP support throughout nodeos.

### DIFF
--- a/plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
+++ b/plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
@@ -23,7 +23,7 @@ struct amqp_witness_plugin_impl {
    std::unique_ptr<reliable_amqp_publisher> rqueue;
 
    struct amqp_witness_msg {
-      bool is_sig_digest;
+      bool is_sig_digest = false;
       chain::digest_type digest;
       chain::signature_type sig;
    };
--- a/plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
+++ b/plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
@@ -13,10 +13,11 @@ struct amqp_witness_plugin_impl {
    };
 
    signature_provider_plugin::signature_provider_type signature_provider;
-   unsigned staleness_limit;
-   std::string amqp_server;
-   std::string exchange;
-   witness_signature_over witness_signature_type;
+   unsigned staleness_limit = 600;
+   std::string amqp_server = "amqp://";
+   std::string exchange = "witness_signatures";
+   std::optional<std::string> routing_key;
+   witness_signature_over witness_signature_type = witness_signature_over::action_mroot;
    bool delete_previous = false;
 
    std::unique_ptr<reliable_amqp_publisher> rqueue;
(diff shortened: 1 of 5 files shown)


plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
```diff
@@ -0,0 +1,122 @@
...
+   signature_provider_plugin::signature_provider_type signature_provider;
+   unsigned staleness_limit;
+   std::string amqp_server;
+   std::string exchange;
+   witness_signature_over witness_signature_type;
+   bool delete_previous = false;
+
+   std::unique_ptr<reliable_amqp_publisher> rqueue;
+
+   struct amqp_witness_msg {
```
spoonincode, line 24: I really didn't want to do it because the message is so simple.... but it's probably best practice to version any struct sent

plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
```diff
@@ -0,0 +1,122 @@
...
+static appbase::abstract_plugin& _amqp_witness_plugin = app().register_plugin<amqp_witness_plugin>();
+
+struct amqp_witness_plugin_impl {
+   enum class witness_signature_over {
+      action_mroot,
+      sig_digest
+   };
+
+   signature_provider_plugin::signature_provider_type signature_provider;
+   unsigned staleness_limit;
```
heifner, line 16: initialize `staleness_limit` & `witness_signature_type`

plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
```diff
@@ -0,0 +1,122 @@
...
+   std::string amqp_server;
+   std::string exchange;
+   witness_signature_over witness_signature_type;
+   bool delete_previous = false;
+
+   std::unique_ptr<reliable_amqp_publisher> rqueue;
+
+   struct amqp_witness_msg {
+      bool is_sig_digest;
```
heifner, line 25: initialize `is_sig_digest`

plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
swatanabe-b1: You could always encode future extensions in the first byte, replacing `is_sig_digest`.

samstickkz:
What does it exactly do ?

spoonincode:
Simply put, allows a fully validating node to sign the action merkle root of a block it fully validated and then publish that signature over AMQP for other components to consume and act upon. See #9379 for a version that places the signatures on chain instead.

This is a small building block in a larger system.

PR #9125, opened 2020-05-21
Add unity build support for some targets

huangminghuang:
## Change Description
This PR add support to enable cmake unity build for certain targets. Enabling unity build can significantly improve the build time for the project. However, it's not possible enable it globally with CMAKE_UNITY_BUILD because some of the libraries/submodules requires modification to make it work. This PR uses ENABLE_UNITY_BUILD flag to enable unity build for the targets don't have problem with it. 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/build.sh
+++ b/.cicd/build.sh
@@ -2,7 +2,7 @@
 set -eo pipefail
 . ./.cicd/helpers/general.sh
 mkdir -p $BUILD_DIR
-CMAKE_EXTRAS="-DCMAKE_BUILD_TYPE='Release' -DENABLE_MULTIVERSION_PROTOCOL_TEST=true"
+CMAKE_EXTRAS="-DCMAKE_BUILD_TYPE='Release' -DENABLE_MULTIVERSION_PROTOCOL_TEST=true -DENABLE_UNITY_BUILD=ON"
 if [[ "$(uname)" == 'Darwin' && $FORCE_LINUX != true ]]; then
     # You can't use chained commands in execute
     if [[ "$GITHUB_ACTIONS" == 'true' ]]; then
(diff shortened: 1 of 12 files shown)


heifner:
@spoonincode do you see any issues with providing this option?

huangminghuang:
> I played with it some and I guess not. It does feel a little hokey needing to sprinkle these all through various cmakefiles. Unless someone is actively using the feature it risks becoming wrong/broken like a handful of other cmake stuff we have sprinkled about still.

IMO, the option should be enabled by default for CI build. This is build times I got from comparing the CI build with the ENABLE_UNITY_BUILD ON and OFF. 

|                          | ENABLE_UNITY_BUILD=OFF | ENABLE_UNITY_BUILD=ON |
|---------------|-----------------------------|-----------------------------|
| Amazon Linux | 26.35 | 18.15 |
| CentOS          |  28.03 | 17.34 |
| macOS.           | 23.14  | 18.34 |
| Ubuntu 16.04 | 27.58  | 17.10  |
| Ubuntu 18.04 | 24.38 | 15.19 |
| Docker            | 21.59 | 16.45 |

Personally, I don't use build script in the EOS repo. I prefer to supply the cmake options I need and use ninja. Without unity build, using "ninja" for clean build without limiting jobs would freeze my Mac with 32 GB ram, I need to explicit limit the jobs to 4 to avoid it. With unity build, I don't need to specify the number of jobs and it also compiles a lot faster. I can even go as far as 18 jobs on my 8-core machine without any issues.

spoonincode:
Should we be keeping CI & build script in parity?

PR #9132, opened 2020-05-26
Rodeos streamer exchanges

TaraTritt:
## Change Description
This builds on #9029

This adds the non breaking changes in the Streamer plugin to create RabbitMQ exchanges, versus the current implementation which only allows directly pushing to RabbitMQ queues. An additional parameter was added to the streamer interface for the `publish` function:  `routing_key`. 

Also renamed the use of the internal `routing_keys` to a more appropriate `stream_routes` as to not confuse this with the RabbitMQ term routing keys. 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### `--stream-rabbits-exchange amqp://guest:guest@localhost:5672/myexchange::direct,mystreamingroute`
Adds a RabbitMQ exchange streamer.

- you can declare multiple rabbitmq exchange connections
- the template is always `amqp://USER:PASSWORD@HOST:PORT/EXCHANGE_NAME[::EXCHANGE_TYPE][/STREAMING_ROUTE1,STREAMING_ROUTE2]`
    - `amqp://` prefix is fixed
    - `USER` is the rabbitmq user
    - `PASSWORD` is the rabbitmq password
    - `HOST` is the address of rabbitmq host
    - `PORT` is the rabbitmq's port
    - `EXCHANGE_NAME` is the name of the exchange that you want to publish the messages
    - `::EXCHANGE_TYPE` is the type of exchange you want to create. By default a `direct` exchange is created. Currently only `direct` and `fanout` are supported.
    - `/STREAMING_ROUTE1,STREAMING_ROUTE2` is a set of streaming routes that you want to determine which exchanges to push messages to.  
        - `direct` exchanges: the default is empty, which means it will publish all messages to all exchanges and pass along the routing key specified in the message. This is useful if you have specified a `direct` exchange where the binding key of the queue to the exchanges specifies which routing keys it cares about therefore filtering at the exchange level via RabbitMQ routing keys.
        - `fanout` exchanges: you can separate routes by commas to specify multiple streaming routes. This is useful if you have specified multiple `fanout` exchanges for each of your event types and want to forward messages to only certain exchanges therefore filtering prior to publishing a message to an exchange.

### DIFF
--- a/programs/rodeos/streams/rabbitmq.hpp
+++ b/programs/rodeos/streams/rabbitmq.hpp
@@ -120,7 +120,6 @@ inline void initialize_rabbits(boost::asio::io_service&                      io_
       }
 
       std::string queue_name = "stream.default";
-
       if (pos != std::string::npos) {
          queue_name = rabbit.substr(pos + 1, rabbit.length());
          rabbit.erase(pos, rabbit.length());
@@ -129,7 +128,6 @@ inline void initialize_rabbits(boost::asio::io_service&                      io_
       std::string address = "amqp://" + rabbit;
       rabbitmq rmq{ io_service, std::move(routings), std::move(address), std::move(queue_name) };
       streams.emplace_back(std::make_unique<rabbitmq>(std::move(rmq)));
-
    }
 }
 
--- a/programs/rodeos/streams/rabbitmq.hpp
+++ b/programs/rodeos/streams/rabbitmq.hpp
@@ -32,7 +32,7 @@ class rabbitmq : public stream_handler {
    }
 
    rabbitmq(boost::asio::io_service& io_service, std::vector<eosio::name> routes, std::string address, std::string exchange_name, std::string exchange_type)
-       : exchangeName_(std::move(exchange_name), routes_(std::move(routes)) {
+       : exchangeName_(std::move(exchange_name)), routes_(std::move(routes)) {
       AMQP::Address amqp_address(address);
       ilog("Connecting to RabbitMQ address ${a} - Exchange: ${e}...", ("a", std::string(amqp_address))("e", exchangeName_));
 
--- a/programs/rodeos/streams/rabbitmq.hpp
+++ b/programs/rodeos/streams/rabbitmq.hpp
@@ -43,11 +43,11 @@ class rabbitmq : public stream_handler {
 
    const std::vector<eosio::name>& get_routes() const override { return routes_; }
 
-   void publish(const char* data, uint64_t data_size, const std::string& route) override {
+   void publish(const char* data, uint64_t data_size, const std::string& routing_key) override {
       if (DEFAULT_EXCHANGE == exchangeName_) {
          channel_->publish(exchangeName_, queueName_, data, data_size, 0);
       } else {
-         channel_->publish(exchangeName_, route, data, data_size, 0);
+         channel_->publish(exchangeName_, routing_key, data, data_size, 0);
       }
    }
 
@@ -77,8 +77,8 @@ class rabbitmq : public stream_handler {
          case "fanout"  :
             type = AMQP::fanout;
             break;
-         case "topic"  :
-            type = AMQP::topic;
+         case "direct"  :
+            type = AMQP::direct;
             break;
       }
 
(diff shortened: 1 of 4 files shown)


programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -45,6 +70,24 @@ class rabbitmq : public stream_handler {
          throw std::runtime_error("RabbitMQ Queue error: " + std::string(error_message));
       });
    }
+
+   void declare_exchange(std::string& exchange_type) {
+      auto type = AMQP::direct;
+      if (exchange_type == "fanout") {
+         type = AMQP::fanout;
+      } else if (exchange_type == "direct") {
```
TaraTritt, line 78: probs will remove this since be default its `AMQP::direct`

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -15,28 +15,53 @@ class rabbitmq : public stream_handler {
...
+   rabbitmq(boost::asio::io_service& io_service, std::vector<eosio::name> routes, std::string address, std::string queue_name)
+       : queueName_(std::move(queue_name)), routes_(std::move(routes)) {
       AMQP::Address amqp_address(address);
-      ilog("Connecting to RabbitMQ address ${a} - Queue: ${q}...", ("a", std::string(amqp_address))("q", name_));
+      ilog("Connecting to RabbitMQ address ${a} - Queue: ${q}...", ("a", std::string(amqp_address))("q", queueName_));
 
-      handler_    = std::make_shared<rabbitmq_handler>(io_service);
-      connection_ = std::make_shared<AMQP::TcpConnection>(handler_.get(), amqp_address);
-      channel_    = std::make_shared<AMQP::TcpChannel>(connection_.get());
+      init(io_service, amqp_address);
```
leordev, line 28: std::move(amqp_address)

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -15,28 +15,53 @@ class rabbitmq : public stream_handler {
...
+      exchangeName_ = DEFAULT_EXCHANGE;
       declare_queue();
    }
 
+   rabbitmq(boost::asio::io_service& io_service, std::vector<eosio::name> routes, std::string address, std::string exchange_name, std::string exchange_type)
+       : exchangeName_(std::move(exchange_name)), routes_(std::move(routes)) {
+      AMQP::Address amqp_address(address);
+      ilog("Connecting to RabbitMQ address ${a} - Exchange: ${e}...", ("a", std::string(amqp_address))("e", exchangeName_));
+
+      init(io_service, amqp_address);
```
leordev, line 39: move addr again

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -15,28 +15,53 @@ class rabbitmq : public stream_handler {
...
+         channel_->publish(exchangeName_, queueName_, data, data_size, 0);
+      } else {
+         channel_->publish(exchangeName_, routing_key.to_string(), data, data_size, 0);
+      }
+   }
 
  private:
+   inline static const std::string DEFAULT_EXCHANGE = "";
+
+   void init(boost::asio::io_service& io_service, AMQP::Address& amqp_address) {
```
leordev, line 57: here you can get the value, since you moved above

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -15,28 +15,53 @@ class rabbitmq : public stream_handler {
...
+   }
+
    const std::vector<eosio::name>& get_routes() const override { return routes_; }
 
-   void publish(const char* data, uint64_t data_size) override { channel_->publish("", name_, data, data_size, 0); }
+   void publish(const char* data, uint64_t data_size, const eosio::name& routing_key) override {
+      if (DEFAULT_EXCHANGE == exchangeName_) {
+         channel_->publish(exchangeName_, queueName_, data, data_size, 0);
+      } else {
+         channel_->publish(exchangeName_, routing_key.to_string(), data, data_size, 0);
```
leordev, line 50: I feel like you should store as string in our constructor instead of converting from name -> string for every message that is being published

TaraTritt: ah very good point, will do

leordev: ohhhh.... nevermind it actually comes from the filter

PR #9135, opened 2020-05-27
amqp_compressed_proof_plugin: Send compressed action merkle proofs over AMQP

spoonincode:
## Change Description

The amqp_compressed_proof_plugin creates “compressed” action merkle proofs over a list of filtered action receivers and their actions once they become irreversible. The proofs are compressed in that they contain as few leafs of the merkle tree as necessary to match the filtered actions with the action_mroot. Additionally, the input stream is constructed in a manner that attempts to be efficient in validation within a smart contract — hashing can be done with minimal copying and working through the proof requires very little temporary memory.

Configuration of the plugin is a mouthful to say the least. amqp_compressed_proof_plugin can generate multiple proofs per block sending them to different AMQP servers and/or exchanges. This is desirable if different actions need to be handled by different downstream consumers on the message bus. But there is no easy way in appbase’s configuration system to define a “group” other then a repeating string parsed in a particular manner. That is what `compressed-proof` option is here, a repeating string with 4 fields describing a single proof to be created and sent in the form of

`NAME=FILTER=SERVER=EXCHANGE`

Where,
* **NAME**: A unique name for this proof. Used for naming the persistence files created for this publisher. I elected to make this manually configurable instead of auto-generated so that users could tweak the exchange, filter list, or server without losing the persistent unconfirmed AMQP queue that goes with each entry.
* **FILTER**: A comma separated list of receivers to include in the proof.
* **SERVER**: URI for the AMQP server
* **EXCHANGE**: exchange name to publish to

An example valid configuration may be, for example
```
# pusblish all actions on eosio and eosio.token to the exchange named 'wheelhouse' on the local amqp server
compressed-proof = proof1=eosio,eosio.token=amqp://=wheelhouse
# Additionally publish actions on eosio via another exchange named 'stack'
compressed-proof = justSendEOSIOHere=eosio=amqp://=stack
```

Parts of the implementation refer to a “command stream” when generating or parsing the proof. This terminology comes about because the serialized data structure is effectively three commands that operate a stack. After the final input from the command stream, the stack should have a single entry that is the same as the action_mroot.

At the front of the serialized proof is a boolean indicating if the proof needs to be computed with action return value semantics or not. I believe there may be a desire to include the block header as well. Downstream components consuming the proof may want to generate additional EOSIO transactions based on the actions in the proof. Block header will be required in that use case for tapos fields. I’m not really sure how to handle this yet… maybe the block header should simply be slapped on the front too.

There may also be a desire to increase the filter granularity by filtering on action name.

Typically you would combine this plugin with #9122: downstream consumers will only operate on the compressed action merkle proof if other witnesses have attested to the action_mroot being correct.

As part of the unit tests, a validator that can parse the compressed command stream is implemented via abieos as an example of a reasonably efficient implementation that could work in a smart contract.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
+++ b/plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
@@ -274,8 +274,6 @@ void compressed_proof_generator::add_result_callback(action_filter_func&& filter
 }
 
 struct amqp_compressed_proof_plugin_impl {
-   bool delete_data = false;
-
    std::unique_ptr<compressed_proof_generator> proof_generator;
    std::list<reliable_amqp_publisher>          publishers;
 };
@@ -298,7 +296,7 @@ void amqp_compressed_proof_plugin::set_program_options(options_description& cli,
          ;
 
    cli.add_options()
-         ("compressed-proof-delete", bpo::bool_switch(&my->delete_data),
+         ("compressed-proof-delete", bpo::bool_switch(),
           "Delete all compressed proof data files for unconfirmed messages and blocks");
 }
 
@@ -312,10 +310,10 @@ void amqp_compressed_proof_plugin::plugin_initialize(const variables_map& option
 
       chain::controller& controller = app().get_plugin<chain_plugin>().chain();
 
-      for(const auto& p : boost::filesystem::directory_iterator(dir)) {
-         if(boost::starts_with(p.path().filename().generic_string(), file_prefix))
-            boost::filesystem::remove(p.path(), ec);
-      }
+      if(options.count("compressed-proof-delete") && options.at("compressed-proof-delete").as<bool>())
+         for(const auto& p : boost::filesystem::directory_iterator(dir))
+            if(boost::starts_with(p.path().filename().generic_string(), file_prefix))
+               boost::filesystem::remove(p.path(), ec);
 
       const boost::filesystem::path reversible_blocks_file = dir / (file_prefix + "-reversible.bin"s);
       my->proof_generator = std::make_unique<compressed_proof_generator>(reversible_blocks_file);
--- a/plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
+++ b/plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
@@ -1,0 +1,397 @@
+#include <eosio/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.hpp>
+
+#include <eosio/chain/protocol_feature_manager.hpp>
+#include <eosio/chain/exceptions.hpp>
+
+#include <eosio/reliable_amqp_publisher/reliable_amqp_publisher.hpp>
+
+#include <boost/multi_index_container.hpp>
+#include <boost/multi_index/key.hpp>
+#include <boost/multi_index/ordered_index.hpp>
+#include <boost/multi_index/hashed_index.hpp>
+#include <boost/algorithm/string/predicate.hpp>
+#include <boost/algorithm/string.hpp>
+
+#include <fc/io/cfile.hpp>
+
+//using boost::multi_index_container;
+using namespace boost::multi_index;
+using namespace std::string_literals;
+
+namespace eosio {
+static appbase::abstract_plugin& _amqp_compressed_proof_plugin = app().register_plugin<amqp_compressed_proof_plugin>();
+
+struct action_entry {
+   chain::action         action;
+   std::vector<char>     return_value;
+   chain::action_receipt action_receipt;
+
+   action_entry() = default;
+
+   action_entry(const chain::action_trace& atrace) :
+     action(atrace.act), return_value(atrace.return_value), action_receipt(*atrace.receipt) {}
+
+   bool operator<(const action_entry& other) const {
+      return action_receipt.global_sequence < other.action_receipt.global_sequence;
+   }
+};
+
+struct combine_nodes{};
+using merkle_cmd = fc::static_variant<chain::digest_type, action_entry, combine_nodes>;
+
+struct merkle_creator_node {
+   chain::checksum_type  hash;
+   std::list<merkle_cmd> cmd_stream;
+};
+
+class merkle_cmd_creator {
+public:
+   void add_noninterested_action(const action_entry& ae) {
+      handle_input_action(merkle_creator_node{chain::digest_type::hash(ae.action_receipt)});
+   }
+
+   void add_interested_action(const action_entry& ae) {
+      handle_input_action(merkle_creator_node{chain::digest_type::hash(ae.action_receipt), {ae}});
+   }
+
+   std::vector<char> generate_serialized_proof(bool arv_activated) {
+      if(!first_input_pair && curent_level.empty())
+         return std::vector<char>();
+      if(first_input_pair && curent_level.empty())
+         return fc::raw::pack<bool, std::list<merkle_cmd>>((bool)arv_activated, first_input_pair->cmd_stream);
+
+
+      if(first_input_pair) {
+         merkle_creator_node even_pair{first_input_pair->hash};
+         curent_level.emplace_back(combine(*first_input_pair, even_pair));
+      }
+
+      while(curent_level.size() > 1) {
+         if(curent_level.size() % 2)
+            curent_level.emplace_back(merkle_creator_node{curent_level.back().hash});
+         for(unsigned int i = 0; i < curent_level.size(); i+=2)
+            curent_level[i/2] = combine(curent_level[i], curent_level[i+1]);
+         curent_level.resize(curent_level.size()/2);
+      }
+
+      return fc::raw::pack<bool, std::list<merkle_cmd>>((bool)arv_activated, curent_level[0].cmd_stream);
+   }
+
+private:
+   void handle_input_action(merkle_creator_node input) {
+      if(!first_input_pair) {
+         first_input_pair = input;
+         return;
+      }
+
+      curent_level.emplace_back(combine(*first_input_pair, input));
+      first_input_pair.reset();
+   }
+
+   merkle_creator_node combine(merkle_creator_node& left, merkle_creator_node& right) {
+      chain::digest_type::encoder digest_enc;
+      fc::raw::pack(digest_enc, chain::make_canonical_left(left.hash), chain::make_canonical_right(right.hash));
+
+      merkle_creator_node new_node = {digest_enc.result()};
+
+      if(left.cmd_stream.empty() && right.cmd_stream.empty()) {
+         //do nothing, this is an uninteresting part of the tree
+      }
+      else if(left.cmd_stream.empty()) {
+         new_node.cmd_stream.emplace_front(left.hash);
+         new_node.cmd_stream.splice(new_node.cmd_stream.end(), right.cmd_stream);
+         new_node.cmd_stream.emplace_back(combine_nodes());
+      }
+      else if(right.cmd_stream.empty()) {
+         new_node.cmd_stream.splice(new_node.cmd_stream.end(), left.cmd_stream);
+         new_node.cmd_stream.emplace_back(right.hash);
+         new_node.cmd_stream.emplace_back(combine_nodes());
+      }
+      else {
+         new_node.cmd_stream.splice(new_node.cmd_stream.end(), left.cmd_stream);
+         new_node.cmd_stream.splice(new_node.cmd_stream.end(), right.cmd_stream);
+         new_node.cmd_stream.emplace_back(combine_nodes());
+      }
+
+      return new_node;
+   }
+
+   std::optional<merkle_creator_node> first_input_pair;
+   std::vector<merkle_creator_node>   curent_level;
+};
+
+struct compressed_proof_generator_impl {
+   struct reversible_action_entries {
+      chain::block_num_type  block_num;
+      chain::block_id_type   block_id;
+      std::set<action_entry> action_entries;
+   };
+
+   struct by_block_id;
+   struct by_block_num;
+
+   typedef multi_index_container<
+      reversible_action_entries,
+      indexed_by<
+         hashed_unique<tag<by_block_id>,   key<&reversible_action_entries::block_id>, std::hash<fc::sha256>>,
+         ordered_unique<tag<by_block_num>, key<&reversible_action_entries::block_num>>
+      >
+   > reversible_action_entries_index_type;
+   reversible_action_entries_index_type reversible_action_entries_index;
+
+   static bool is_onblock(const chain::transaction_trace_ptr& p) {
+      if (p->action_traces.size() != 1)
+         return false;
+      auto& act = p->action_traces[0].act;
+      if (act.account != eosio::chain::config::system_account_name || act.name != N(onblock) ||
+          act.authorization.size() != 1)
+         return false;
+      auto& auth = act.authorization[0];
+      return auth.actor == eosio::chain::config::system_account_name &&
+             auth.permission == eosio::chain::config::active_name;
+   }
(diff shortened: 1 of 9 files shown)


plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,397 @@
...
+void compressed_proof_generator::on_applied_transaction(const chain::transaction_trace_ptr& trace) {
+   if(!trace->receipt)
+      return;
+   //only executed & delayed transaction traces would make it in to action_mroot
+   if(trace->receipt->status != chain::transaction_receipt::status_enum::executed &&
+      trace->receipt->status != chain::transaction_receipt::status_enum::delayed)
+      return;
+
+   if(my->is_onblock(trace))
+      my->onblock_trace.emplace(trace);
```
swatanabe-b1, line 214: What if we abort a block in which onblock suceeded and start a new block in which onblock failed?

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,397 @@
...
+      indexed_by<
+         hashed_unique<tag<by_block_id>,   key<&reversible_action_entries::block_id>, std::hash<fc::sha256>>,
+         ordered_unique<tag<by_block_num>, key<&reversible_action_entries::block_num>>
+      >
+   > reversible_action_entries_index_type;
+   reversible_action_entries_index_type reversible_action_entries_index;
+
+   static bool is_onblock(const chain::transaction_trace_ptr& p) {
+      if (p->action_traces.size() != 1)
+         return false;
```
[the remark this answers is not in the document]
spoonincode: Good spot. This is copy pasta from state history. state history plugin in release/2.0.x still has the `p->action_traces.size() != 1` check, but state history in develop has that condition fixed after it was spotted here https://github.com/EOSIO/eos/pull/8750#discussion_r417622287. This is a defect that ought to have the fix backported to 2.0.x 

This is also code that is now copy pasted 3 places: state history, trace api, and now here. Pretty lousy.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: I believe state history & trace api have the same semantics? Honestly I didn't give it much thought since I figured those had been noodled through.

swatanabe-b1: I think this is a problem here and in state history and trace as well, but only when running the chain in speculative mode (which we probably don't want to do anyway?).  I suppose it could also be a problem if we receive a bad block in head mode.  If we get a bad block in irreversible mode, we have bigger problems.

heifner: @spoonincode I will create a fix for this here, state history, and trace.

heifner: Fix pushed up to this branch.

heifner: This is turning out to be more complicated than I first thought, still looking into it. `block_num` is available but could be the same across an aborted block. id is not available unless block is from an external producer. There is no start_block or abort_block signal.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -219,8 +219,11 @@ void compressed_proof_generator::on_applied_transaction(const chain::transaction
 void compressed_proof_generator::on_accepted_block(const chain::block_state_ptr& bsp) {
    std::set<action_entry> action_entries_this_block;
 
-   if(my->onblock_trace)
-      action_entries_this_block.emplace((*my->onblock_trace)->action_traces.front());
+   // verify block_num of onblock_trace because last block could have been aborted & new block have no onblock
+   if(my->onblock_trace && (*my->onblock_trace)->block_num == bsp->block_num) {
```
swatanabe-b1: This looks correct to me, but it's tricky.  If the previous block is the same, then onblock will consistently either succeed or fail, since it's always starting from the same state.  If the previous block is different, then it must be an ancestor of the previous block corresponding to the onblock trace, because pushing any block will reset onblock_trace.

heifner: block_header timestamp is a slot and everything else should be consistent as well... hmm, ok, nice thanks!

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
heifner: Added verification of block_num see discussion below.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,393 @@
...
+      for(const chain::action_trace& at : it->second->action_traces)
+         action_entries_this_block.emplace(at);
+   }
+   my->cached_traces.clear();
+   my->onblock_trace.reset();
+
+   my->reversible_action_entries_index.emplace(generator_impl::reversible_action_entries{bsp->block_num, bsp->id, std::move(action_entries_this_block)});
+}
+
+void compressed_proof_generator::on_irreversible_block(const chain::block_state_ptr& bsp, const chain::controller& controller) {
```
swatanabe-b1, line 244: Do we always need to wait for irreversibility or can we just follow the database read mode and let the user decide?

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,393 @@
...
+      chain::controller& controller = app().get_plugin<chain_plugin>().chain();
+
+      if(options.count("compressed-proof-delete") && options.at("compressed-proof-delete").as<bool>())
+         for(const auto& p : boost::filesystem::directory_iterator(dir))
+            if(boost::starts_with(p.path().filename().generic_string(), file_prefix))
+               boost::filesystem::remove(p.path(), ec);
+
+      const boost::filesystem::path reversible_blocks_file = dir / (file_prefix + "-reversible.bin"s);
+      my->proof_generator = std::make_unique<compressed_proof_generator>(reversible_blocks_file);
+      controller.irreversible_block.connect([&generator=*my->proof_generator,&controller](const chain::block_state_ptr& bsp) {
```
swatanabe-b1, line 323: Doesn't this need to keep a connection, so we can disconnect on shutdown?

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,393 @@
...
+   }
+
+private:
+   void handle_input_action(merkle_creator_node input) {
+      if(!first_input_pair) {
+         first_input_pair = input;
+         return;
+      }
+
+      curent_level.emplace_back(combine(*first_input_pair, input));
```
swatanabe-b1, line 87: Is there a reason for combining the leaf nodes in `handle_input_actions`?  It would simplify the code a bit if `generate_serialized_proof` handled all the combining.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
swatanabe-b1: No.  I didn't consider the timestamp.  The block timestamp can skip time and is observable from within a contract.

swatanabe-b1: There's also the case of activating `preactive_feature`, which happens before `onblock`.

heifner: Hmm, back to the drawing board then.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: We don't here and neither does the witness plugin either. Is relying on the DB read mode really a good indication vs a separate config item though?

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: By the time shutdown() is called the event loop is stopped so controller will never pump out more blocks (or other signals) so the connection to the now destroyed amqp_compressed_proof_plugin will never be called.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
swatanabe-b1: The witness plugin is somewhat different.  Changing the point where it signs changes the meaning of the signature.  Signing before a block becomes irreversible is potentially dangerous.

swatanabe-b1: For the compressed proof plugin, decoupling the proof generation from on_accepted_block requires you to maintain significant extra state.  I don't thing it's worthwhile to make it configurable separately from the database read mode.

unittests/compressed_merkle_tests.cpp
```diff
@@ -0,0 +1,210 @@
...
+   conns.emplace_back(ctrl.applied_transaction.connect([&](std::tuple<const transaction_trace_ptr&, const packed_transaction_ptr&> t) {
+      generator.on_applied_transaction(std::get<0>(t));
+   }));
+   conns.emplace_back(ctrl.accepted_block.connect([&](const block_state_ptr& p) {
+      generator.on_accepted_block(p);
+   }));
+   return conns;
+}
+
+BOOST_AUTO_TEST_SUITE(test_compressed_proof)
```
swatanabe-b1, line 49: There should probably be at least one test case where the action schedule order is different from the execution order.

spoonincode: Probably something with deferred too since they're special

spoonincode:
This branch was rebased to `reliable_amqp_publisher-dev` and force updated. There were no changes to any commit during the rebase.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: I think there was a reason in a previous impl. But indeed now none of that is needed, can just be chopped out.

swatanabe-b1:
It looks like this needs to be rebased.

spoonincode:
yes, was debating doing that before or after merging #9183. I can do it now

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,391 @@
...
+         else
+            id = r.trx.get<chain::packed_transaction>().id();
+         auto it = cached_traces.find(id);
+         EOS_ASSERT(it != cached_traces.end(), chain::misc_exception, "missing trace for transaction ${id}", ("id", id));
+         for(const chain::action_trace& at : it->second->action_traces)
+            action_entries_this_block.emplace(at);
+      }
+      clear_caches();
+
+      if(irreversible)
```
swatanabe-b1, line 154: If the `db_read_mode` is `IRREVERSIBLE`, will there ever be a situation where `on_accepted_block` is not immediately followed by `on_irreversible_block`?

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,391 @@
...
+            const boost::filesystem::path amqp_unconfimed_file   = dir / (file_prefix + "-unconfirmed-"s + name + ".bin"s);
+            reliable_amqp_publisher& publisher = my->publishers.emplace_back(tokens[2], tokens[3], "", amqp_unconfimed_file, "eosio.node.compressed_proof_v0");
+
+            std::set<chain::name> filter_on_names;
+            for(const std::string& name_string : filtered_receivers)
+               filter_on_names.emplace(name_string);
+
+            callbacks.emplace_back(std::make_pair(
+               [filter_on_names](const chain::action& act) {
+                  return filter_on_names.find(act.account) != filter_on_names.end();
```
swatanabe-b1, line 352: I think we do need the ability to filter on the action name as well.  Even if we put the actions we're watching in a separate contract, this contract will likely need some administrative actions that we won't want to include.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -0,0 +1,391 @@
...
+         run_callbacks_on_block(bsp, action_entries, controller);
+
+         reversible_action_entries_index.get<by_block_num>().erase(
+                 reversible_action_entries_index.get<by_block_num>().begin(),
+                 reversible_action_entries_index.get<by_block_num>().upper_bound(bsp->block_num)
+         );
+      }
+   }
+
+   void run_callbacks_on_block(const chain::block_state_ptr& bsp, const std::set<action_entry>& action_entries, const chain::controller& controller) {
```
swatanabe-b1, line 176: It would be nice if there were a way to limit the size of the proof, splitting it into several proofs as needed.

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: 100% agree; I need this for the bananas and the banana contract. It was something I was going to add later but I'm fine doing it now too

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: I wasn't able to quite convince myself that removing all that is safe. I couldn't find a good way of testing all the cases in unit tests because, for example, I couldn't find a good way to switch between speculative & irreversible mode while the proof generator was attached to controller's signals (because all of tester's ctors init the controller).

spoonincode:
Branch has been rebased & force pushed. Not something we typically like to do but the base branch had been rebased to completely eliminate some commits that may have muddled boxed vs develop differences. No changes to previous commits before this comment were made but one commit was dropped because it was completely empty now. (github base switched back and forth to update github display)

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: After playing with it some more, I can't find any problems just driving it via `on_accepted_block`

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
spoonincode: Is that something that should be rectified ASAP or can it wait until later?

swatanabe-b1: It can wait.

PR #9143, opened 2020-05-28
Prune data integration test fix

brianjohnson5972:
## Change Description
The eosio_blocklog_prune_test test was not sufficiently waiting long enough for the nodes to advance their irreversible blocks, so the timeout was changed to calculate how many block time slots needed to occur.  Also cleaned up test to use helper methods on Node and added better reporting for when tests identify that the head/lib is not advancing.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -489,8 +489,7 @@ class Node(object):
         ret=Utils.waitForBool(lam, timeout)
         return ret
 
-    def waitForBlock(self, blockNum, timeout=None, blockType=BlockType.head, reportInterval=None):
-        lam = lambda: self.getBlockNum(blockType=blockType) > blockNum
+    def waitForBlock(self, blockNum, timeout=None, blockType=BlockType.head, reportInterval=None, errorContext=None):
         blockDesc = "head" if blockType == BlockType.head else "LIB"
         count = 0
 
@@ -506,8 +505,41 @@ class Node(object):
                     info = self.node.getInfo()
                     Utils.Print("Waiting on %s block num %d, get info = {\n%s\n}" % (blockDesc, blockNum, info))
 
-        reporter = WaitReporter(self, reportInterval) if reportInterval is not None else None
-        ret=Utils.waitForBool(lam, timeout, reporter=reporter)
+        class RequireBlockNum:
+            def __init__(self, node, blockNum):
+                self.node = node
+                self.blockNum = blockNum
+                self.lastBlockNum = None
+                self.passed = False
+                self.advanced = None
+
+            def __call__(self):
+                currentBlockNum = self.node.getBlockNum(blockType=blockType)
+                self.advanced = False
+                if self.lastBlockNum is None or self.lastBlockNum < currentBlockNum:
+                    self.advanced = True
+                elif self.lastBlockNum > currentBlockNum:
+                    Utils.Print("waitForBlock is waiting to reach block number: %d and the block number has rolled back from %d to %d." %
+                                (self.blockNum, self.lastBlockNum, currentBlockNum))
+                self.lastBlockNum = currentBlockNum
+                self.passed = self.lastBlockNum > self.blockNum
+                return self.passed
+
+            def __enter__(self):
+                return self
+
+            def __exit__(self, exc_type, exc_value, exc_traceback):
+                if not self.passed:
+                    notAdvanceStr="(but has not change since last sleep)" if self.advanced else ""
+                    Utils.Print("waitForBlock never reached block number: %d.  It started at: %d and had progressed%s to: %d after %d seconds." %
+                                (blockNum, currentBlockNum, notAdvanceStr, self.lastBlockNum, time.time()-currentTime))
+
+        with RequireBlockNum(self, blockNum) as lam:
+
+            reporter = WaitReporter(self, reportInterval) if reportInterval is not None else None
+            ret=Utils.waitForBool(lam, timeout, reporter=reporter)
+
+        assert ret is not None or errorContext is None, Utils.errorExit("%s." % (errorContext))
         return ret
 
     def waitForIrreversibleBlock(self, blockNum, timeout=None, blockType=BlockType.head):
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -484,7 +484,7 @@ class Node(object):
         ret=Utils.waitForTruth(lam, timeout)
         return ret
 
-    def waitForNextBlock(self, timeout=WaitSpec.default, blockType=BlockType.head):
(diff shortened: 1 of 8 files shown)


tests/Node.py
```diff
@@ -483,14 +484,21 @@ def waitForTransFinalization(self, transId, timeout=None):
...
+        if isinstance(timeout, Timeout):
+            timeout = timeout.seconds(num, num+1)
         lam = lambda: self.getHeadBlockNum() > num
         ret=Utils.waitForBool(lam, timeout)
         return ret
 
-    def waitForBlock(self, blockNum, timeout=None, blockType=BlockType.head, reportInterval=None):
-        lam = lambda: self.getBlockNum(blockType=blockType) > blockNum
+    def waitForBlock(self, blockNum, timeout=None, blockType=BlockType.head, reportInterval=None, errorContext=None):
+#    def waitForBlock(self, blockNum, timeout=Timeout.default, blockType=BlockType.head, reportInterval=None, errorContext=None):
```
jgiszczak, line 496: Commented code.

tests/Node.py
```diff
@@ -506,11 +514,44 @@ def __call__(self):
...
+                self.lastBlockNum = currentBlockNum
+                self.passed = self.lastBlockNum > self.blockNum
+                return self.passed
+
+            def __enter__(self):
+                return self
+
+            def __exit__(self, exc_type, exc_value, exc_traceback):
+                if not self.passed:
+                    notAdvanceStr="(but has not change since last sleep)" if self.advanced else ""
```
jgiszczak, line 542: "change" -> "changed"

tests/testUtils.py
```diff
@@ -50,6 +51,65 @@ class BlockLogAction(EnumType):
...
+        self.value = timeout
+        if self.sleep_interval is None:
+            self.sleep_interval = max(timeout / 10, 1)
+
+    def sleepTime(self, default=None):
+        time = self.sleep_interval if self.sleep_interval is not None else default
+        return time if time is not None else Timeout.default_interval_seconds
+
+    def asSeconds(self):
+        assert self.value != -1, "Timeout.calculate was used, but calculate method never called."
```
jgiszczak, line 101: There is no `calculate` method.  Perhaps this should read: "Timeout automatic calculation was used but convert method was never called."

tests/testUtils.py
```diff
@@ -50,6 +51,65 @@ class BlockLogAction(EnumType):
 addEnum(BlockLogAction, "prune_transactions")
 
 ###########################################################################################
+
+class Timeout:
```
huangminghuang, line 55: It's not timeout anymore, I would change the name to WaitSpec. Have an const object called defaultWaitSpec, use defaultWaitSpec for the default value of  functions  that require it.

tests/testUtils.py
```diff
@@ -50,6 +51,65 @@ class BlockLogAction(EnumType):
...
+            blockWindowsToWait = (endBlockNum - startBlockNum + blocksPerWindow - 1) / blocksPerWindow
+            secondsPerWindow = blocksPerWindow / 2
+            timeout += blockWindowsToWait * secondsPerWindow
+
+        self.value = timeout
+        if self.sleep_interval is None:
+            self.sleep_interval = max(timeout / 10, 1)
+
+    def sleepTime(self, default=None):
+        time = self.sleep_interval if self.sleep_interval is not None else default
```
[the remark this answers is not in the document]
brianjohnson5972: In the init, it is checking that if the sleep_interval is not None, that the value is appropriate for a positive amount of time.  Here it is letting you either:
indicate that your instance fall determines the sleep_interval (by passing to init)
calculate it based off of your calculated wait time
or else use the passed in default value

yes it is complicated, and I don't think we really even need a sleep_interval (I think it is always 3 seconds), but figured if I didn't put this in here, that you end up having people not use things because they need x and y, and z, and some methods give you x and y and some give you y and z.  I'm fine with whatever, but would rather do a separate cleanup/refactor task, but I don't think we will be given time to actually do it.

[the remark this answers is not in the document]
brianjohnson5972: Added a comment, since if we do leave it in, it at least needs that.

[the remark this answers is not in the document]
huangminghuang: If I read it right, the function could be rewritten as
```
def sleepTime(self, default=Timeout.default_interval_seconds):
    return self.sleep_interval  if self.sleep_interval else default
```

This makes the code a lot less convoluted.

tests/testUtils.py
```diff
@@ -244,6 +312,8 @@ def waitForObj(lam, timeout=None, sleepTime=3, reporter=None):
 
     @staticmethod
     def waitForBool(lam, timeout=None, sleepTime=3, reporter=None):
```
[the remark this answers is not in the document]
brianjohnson5972: changed waitForObj(...) to waitForTruth (since python docs refer to __bool__ as returning the truth value) and replacing all calls to waitForBool with that

tests/testUtils.py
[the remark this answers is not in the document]
brianjohnson5972: Or, if not, I say we just pull all of them out of everything and just use 3 seconds (or less if remaining time is less than that)

tests/Node.py
brianjohnson5972: Commented code was what I wanted actually, so removed comment and replaced the executed line.

tests/Node.py
brianjohnson5972: Fixed both

tests/testUtils.py
brianjohnson5972: fixed

tests/testUtils.py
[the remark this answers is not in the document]
brianjohnson5972: Removed sleepTime from WaitSpec.

tests/testUtils.py
brianjohnson5972: Changed to WaitSpec

tests/testUtils.py
jgiszczak: This still says `but calculate method never called`.

I'm not sure how useful saying `Timeout.calculate was used` is, either.  I misinterpreted it the first time, not realizing it refers to the attribute set below.  It may make sense once it stops talking about a nonexistent `calculate` method.

brianjohnson5972: Yeah, fixed it to say convert and added more information.  This is intended to tell the test writer that either somewhere in the chain of methods it needs to call convert or else they incorrectly used WaitSpec.calculate(...)

PR #9164, opened 2020-06-01
Fix for connection cycle not being in sync with test startup.

brianjohnson5972:
## Change Description
Test restarts several nodes at the same time, which occasionally leads to connection cycle being delayed to the next pass, which is 30 seconds later, so test was failing.  Added extra time to wait for lib to start moving. 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/eosio_blocklog_prune_test.py
+++ b/tests/eosio_blocklog_prune_test.py
@@ -156,7 +156,8 @@ try:
     #
     # Check lightValidationNode keeping sync while the full fullValidationNode stop syncing after the cfd block
     #
-    lightValidationNode.waitForBlock(cfTrxBlockNum, blockType=BlockType.lib, timeout=WaitSpec.calculate(), errorContext="ligtValidationNode did not advance")
+    timeForNodesToWorkOutReconnect=30 + 10    # 30 is reconnect cycle since all nodes are restarting and giving time for a little churn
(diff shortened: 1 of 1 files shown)


tests/eosio_blocklog_prune_test.py
```diff
@@ -156,7 +156,8 @@
     #
     # Check lightValidationNode keeping sync while the full fullValidationNode stop syncing after the cfd block
     #
-    lightValidationNode.waitForBlock(cfTrxBlockNum, blockType=BlockType.lib, timeout=WaitSpec.calculate(), errorContext="ligtValidationNode did not advance")
+    timeForNodesToWorkOutReconnect=30 + 10    # 30 is reconnect cycle since all nodes are restarting and giving time for a little churn
+    lightValidationNode.waitForBlock(cfTrxBlockNum, blockType=BlockType.lib, timeout=WaitSpec.calculate(leeway=timeForNodesToWorkOutReconnect), errorContext="ligtValidationNode did not advance")
```
jgiszczak, line 160: Let's correct the spelling mistake while we're here.

brianjohnson5972: fixed

PR #9169, opened 2020-06-03
Fix onblock trace tracking - develop

heifner:
## Change Description

- Fix issue identified with current scheme for capturing `onblock` transaction traces. See https://github.com/EOSIO/eos/pull/9135#discussion_r432711431 & https://github.com/EOSIO/eos/pull/9135#discussion_r433423912
  - The previous scheme would report an `onblock` trace for the case where a block is aborted with a valid `onblock` followed by a non-aborted block with an invalid `onblock`.
- Added new controller signal `block_start` for plugins to use to know when a new block starts.
- Added an `is_onblock` to `trace.hpp` so this code does not have to be duplicated across various plugins.
- `develop` version of #9170 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api/chain_extraction.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api/chain_extraction.hpp
@@ -85,8 +85,7 @@ private:
 
          std::vector<transaction_trace_v1>& traces = bt.transactions_v1;
          traces.reserve( block_state->block->transactions.size() + 1 );
-         // verify block_num of onblock_trace because last block could have been aborted & new block have no onblock
-         if( onblock_trace && onblock_trace->trace->block_num == block_state->block_num )
+         if( onblock_trace )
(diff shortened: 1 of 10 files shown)


plugins/trace_api_plugin/include/eosio/trace_api/chain_extraction.hpp
```diff
@@ -77,13 +70,19 @@ class chain_extraction_impl_type {
       store_lib( block_state );
    }
 
+   void on_block_start( uint32_t block_num ) {
+      cached_traces.clear();
```
b1bart, line 74: I like the pattern used in `state_history_plugin` where we have a method that clears all caches called from the `block_start` handler and after processing a `block_state_ptr` on lines 99-100

heifner: Done.

PR #9175, opened 2020-06-03
Ship delay error fix

brianjohnson5972:
## Change Description
Added Node methods to analyze production delays, added reporting these at the end of tests, and changed ship_test to only identify as an error when a block is reported as received in the ship node significantly later than when it would have been produced in the producing node (it originally assumed that it was produced at the slot time, which resulted in VM hiccups delaying block production and triggering this error). Also added storing nodeId in the Node so it would not need to be passed in to so many methods.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Cluster.py
+++ b/tests/Cluster.py
@@ -119,6 +119,7 @@ class Cluster(object):
         self.useBiosBootFile=False
         self.filesToCleanup=[]
         self.alternateVersionLabels=Cluster.__defaultAlternateVersionLabels()
+        self.biosNode = None
 
 
     def setChainStrategy(self, chainSyncStrategy=Utils.SyncReplayTag):
@@ -496,7 +497,8 @@ class Cluster(object):
     def initializeNodes(self, defproduceraPrvtKey=None, defproducerbPrvtKey=None, onlyBios=False):
         port=Cluster.__BiosPort if onlyBios else self.port
         host=Cluster.__BiosHost if onlyBios else self.host
-        node=Node(host, port, walletMgr=self.walletMgr)
+        nodeNum="bios" if onlyBios else 0
+        node=Node(host, port, nodeNum, walletMgr=self.walletMgr)
         if Utils.Debug: Utils.Print("Node: %s", str(node))
 
         node.checkPulse(exitOnError=True)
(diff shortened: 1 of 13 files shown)


tests/Cluster.py
```diff
@@ -707,7 +710,17 @@ def getNode(self, nodeId=0, exitOnError=True):
         return self.nodes[nodeId]
 
     def getNodes(self):
-        return self.nodes
+        nodes = []
```
jgiszczak, line 713: If the purpose of this convoluted code is to return a copy of the list of nodes, use a one-liner: `return self.nodes[:]`.  This uses slice notation to compactly copy a list.  `self.nodes` should have been set to an empty list in the class's `__init__()` method so the additional checks for existence and emptiness are not required.

tests/Node.py
```diff
@@ -1486,3 +1487,72 @@ def modifyBuiltinPFSubjRestrictions(self, nodeId, featureCodename, subjectiveRes
     def createSnapshot(self):
         param = { }
         return self.processCurlCmd("producer", "create_snapshot", json.dumps(param))
+
+    @staticmethod
+    def findFiles(path):
```
jgiszczak, line 1492: Since we're moving this, can we also clarify its name?  It's `findStderrFiles`.

tests/Node.py
```diff
@@ -1486,3 +1487,72 @@ def modifyBuiltinPFSubjRestrictions(self, nodeId, featureCodename, subjectiveRes
...
+                            if matchBlockActuallyProduced:
+                                prodTimeStr = matchBlockActuallyProduced.group(1)
+                                break
+
+                            line = f.readline()
+
+                        prodTime = datetime.strptime(prodTimeStr, Utils.TimeFmt)
+                        slotTime = datetime.strptime(slotTimeStr, Utils.TimeFmt)
+                        delta = prodTime - slotTime
+                        millisecondsPerSecond=1000
```
jgiszczak, line 1542: No need for this.  Just use the `milliseconds=` keyword parameter in the next line rather than the `seconds=` keyword parameter.

PR #9181, opened 2020-06-04
amqp_trx_plugin - transaction in, transaction_trace out

heifner:
## Change Description

- `amqp_trx_plugin`

Binary `fc::pack` format:
```
// consume message types
```

- `amqp_trace_plugin`

Binary `fc::pack` format:
```
// publish message types
   std::string error_message;
};
```

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

`eosio::amqp_trx_plugin` options

- `amqp-trx-address`
  - AMQP address: Format: amqp://USER:PASSWORD@ADDRESS:PORT
      Will consume from 'trx' queue

`eosio::amqp_trace_plugin` options

- `amqp-trace-address`
  - AMQP address: Format: amqp://USER:PASSWORD@ADDRESS:PORT
      Will publish to 'trace' queue.
- `amqp-trace-exchange`
  - Existing AMQP exchange to send transaction trace messages.


spoonincode:
I wonder if the more AMQPy way of doing this would be instead of the `amqp-trx-exchange` config, leveraging the `reply-to` field of a message. If you have multiple callers this allows you to send the responses back to the individual caller. I'm guessing it's not a huge problem for us in this case though since a submitter likely knows if a response was intended for itself or someone else because it's expecting a particular trxid.

swatanabe-b1:
If we use `reply-to` then we'd still need somewhere to publish transaction traces when `amqp-trx-publish-all-traces` is on.

programs/cleos/main.cpp
```diff
@@ -436,12 +431,52 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+               } );
+               consumer.onReceived( [&]( const AMQP::Message& message, uint64_t delivery_tag, bool redelivered ) {
+                  if( message.hasCorrelationID() && message.correlationID() == id ) {
+                     fc::datastream<const char*> ds( message.body(), message.bodySize() );
+                     fc::raw::unpack( ds, trace_msg );
+                     qp_trace.ack( delivery_tag );
+                     received_trace.set_value();
+                  }
+               } );
+               received_trace.get_future().wait();
```
swatanabe-b1, line 461: Should this have a timeout?

.cicd/build.sh
```diff
@@ -3,6 +3,11 @@ set -eo pipefail
 . ./.cicd/helpers/general.sh
 mkdir -p $BUILD_DIR
 CMAKE_EXTRAS="-DCMAKE_BUILD_TYPE='Release' -DENABLE_MULTIVERSION_PROTOCOL_TEST=true -DENABLE_UNITY_BUILD=ON"
+if [[ "$BUILDKITE" == 'true' ]]; then
+    SECRETS_DIR='/System/Volumes/Data/Network/NAS/MAC_FLEET/ANKA/secrets'
+    AMQP_CONNECTION_STRING=${AMQP_CONNECTION_STRING:-"$(cat $SECRETS_DIR/amqp-connection-string.txt)"}
+    CMAKE_EXTRAS="$CMAKE_EXTRAS -DAMQP_CONN_STR='"$AMQP_CONNECTION_STRING"'"
```
swatanabe-b1, line 9: Is this a different queue for each test run, or is it safe to have multiple tests connecting to the same queue?

programs/cleos/main.cpp
```diff
@@ -436,12 +431,52 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+               consumer.onError( []( const char* message ) {
+                  std::cerr << "consume failed: " << message << std::endl;
+               } );
+               consumer.onReceived( [&]( const AMQP::Message& message, uint64_t delivery_tag, bool redelivered ) {
+                  if( message.hasCorrelationID() && message.correlationID() == id ) {
+                     fc::datastream<const char*> ds( message.body(), message.bodySize() );
+                     fc::raw::unpack( ds, trace_msg );
+                     qp_trace.ack( delivery_tag );
+                     received_trace.set_value();
+                  }
```
swatanabe-b1, line 459: Do we need to reject other messages?  What if there are multiple instances of cleos running in parallel?

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,241 @@
...
+                        const fc::static_variant<fc::exception_ptr, chain::transaction_trace_ptr>& result ) {
+
+      try {
+         if( result.contains<fc::exception_ptr>() ) {
+            auto& ex = *result.get<fc::exception_ptr>();
+            std::string err = ex.to_string();
+            fc_dlog( logger, "bad packed_transaction : ${e}", ("e", err) );
+            transaction_trace_exception tex{ ex.code() };
+            fc::unsigned_int which = transaction_trace_msg::tag<transaction_trace_exception>::value;
+            // TODO; use fc::datastream<std::vector<char>> when available
```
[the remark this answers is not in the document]
spoonincode: oh hah is that the trick, I could not figure that one out

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,241 @@
...
+   }
+}
+
+void amqp_trx_plugin::plugin_shutdown() {
+   try {
+      fc_dlog( logger, "shutdown.." );
+      my->shutting_down = true; // stop receiving transactions to consume
+      my->applied_transaction_connection.reset();
+      // drain queue so all traces are published
+      app().post( priority::lowest, [me = my](){} );
```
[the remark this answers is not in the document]
heifner: The first one is so all trx in progress are published to the trace queue. However, I see that this needs to be up one line.

.cicd/build.sh
heifner: For this particular test it is safe for all tests to share the same queue. If we have more complicated setups in the future a diff scheme will likely be needed.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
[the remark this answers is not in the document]
heifner: That variadic function just does what is done here.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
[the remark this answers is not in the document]
swatanabe-b1: Does it matter where you put it?  It'll just sit in the queue until after `plugin_shutdown` returns, won't it?

programs/cleos/main.cpp
heifner: Probably safes to reject.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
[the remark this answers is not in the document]
heifner: Duh, yeah. What I need to do is post the `my->applied_transaction_connection.reset()`

programs/cleos/main.cpp
```diff
@@ -436,12 +431,52 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+            eosio::chain::named_thread_pool thread_pool( "amqp", 1 );
+            transaction_trace_msg trace_msg;
+            eosio::transaction_msg msg{packed_transaction( std::move( trx ), true, compression )};
+            auto buf = fc::raw::pack( msg );
+            const auto& tid = msg.get<packed_transaction>().id();
+            const string id = tid.str();
+            {
+               eosio::amqp qp_trx( fc::logger::get( DEFAULT_LOGGER ), thread_pool.get_executor(), amqp_address, "trx" );
+               eosio::amqp qp_trace( fc::logger::get( DEFAULT_LOGGER ), thread_pool.get_executor(), amqp_address,
+                                     "trace" );
```
swatanabe-b1, line 444: All instances of cleos are reading from the same queue, which means that the messages will only be dispatched to one of them, right?  They'll only be dispatched to another consumer after the first consumer fails to ack them.

spoonincode: That's correct. And if no one ACKs that message it keeps getting requeued forever unless you attach ttl rules to the queue, etc.
Really for an RPC use case like cleos, I think the AMQP thing is to use a server named exclusive queue that gets set with `reply-to`

heifner: Does RabbitMQ have expiration of messages? Could move to a `reply-to`  but I don't think that is the actual use case in that others are likely to be interested in the trace. Thinking about it I'm not sure even if the message should be ack as others might want to see the trace.

programs/cleos/main.cpp
heifner: I figured the use can ctrl-c if they give up on it.

programs/cleos/main.cpp
heifner: I removed the ack as other consumers may be interested in the trace.

programs/cleos/main.cpp
swatanabe-b1: Dropping acks is not a good solution, because the message will still only be delivered to one client at a time.  It will only be re-queued after you close the connection.

swatanabe-b1: If you want multicast, you need a queue per recipient.

heifner: That can be done in rabbitmq right? Setup a fanout or whatever? So here should we assume we are the only consumer and ack, reject appropriately?

swatanabe-b1: We shouldn't assume that we're the only consumer.  As Matt suggested, we should use a server named exclusive queue, which enforces uniqueness.  The fanout is set up by telling the exchange to send the message to multiple queues.  It depends on the exchange type, the message's routing key, and the queue's binding key.  The exchange type determines how the routing key is matched against the binding keys of all the queues.

swatanabe-b1: The two main options are to set the routing key to the transaction ID or to set it to a unique ID specified by the client (i.e. reply-to).

swatanabe-b1: I suppose that we could set it up to broadcast all transaction traces, but that's kind of wasteful when cleos only cares about one specific transaction.

spoonincode: >I don't think that is the actual use case in that others are likely to be interested in the trace

To me, the plugin seems to be trying to do two different use cases: a textbook RPC use case (which makes sense to use reply-to). But then a different use case which is a 1-to-many broadcast (which makes sense to spam everything to a single exchange and let the exchange and bound queues decide their semantics)

I'm not entirely sure I understand the full _ultimate_ use cases though. Like in the 1-to-many broadcast case at that point the transactions aren't in blocks yet, is that deceptive for the use case? Or, say, if you're going to run multiple instances of this plugin all broadcasting, will the downstream listeners understand they can get duplicates.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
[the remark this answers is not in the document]
swatanabe-b1: Well I guess it depends on why you want `datastream<vector<char>>`. I assumed that it was to reduce verbosity, in which case the variadic version works just as well.

programs/cleos/main.cpp
swatanabe-b1: Okay.  Looks like that matches what cleos does normally.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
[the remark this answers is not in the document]
heifner: I was thinking more for optimization.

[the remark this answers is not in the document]
swatanabe-b1: Reallocation also increases the cost.  The extra cost should be bounded at 2x for both, so I have no idea which way is actually faster.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,249 @@
...
+      // always want plugin_shutdown even on exception
+      plugin_shutdown();
+      throw;
+   }
+}
+
+void amqp_trx_plugin::plugin_shutdown() {
+   try {
+      fc_dlog( logger, "shutdown.." );
+      my->shutting_down = true; // stop receiving transactions to consume
```
swatanabe-b1, line 227: The worker thread does not synchronize with anything later in this function, so it's entirely possible that we could return from `plugin_shutdown` and continue on to shutdown the `chain_plugin` all before `shutting_down` becomes visible to the worker thread.

heifner: `shutting_down` is atomic bool

swatanabe-b1: Initial state: `chain_plugin` and `amqp_trx_plugin` are running.
Threads: `app`, `amqp`

I believe that the following order is consistent:

1. app: enter app().shutdown()
2. app: enter amqp_trx_plugin.plugin_shutdown()
3. amqp: read my->shutting_down == true
4. app:  shutting_down = false
5. app: shutdown and destroy other plugins
6. amqp: handle_message
7. app: `my->chain_plug->accept_transaction` (posted from 6)
8. app: `chain_plug` is a dangling pointer.  KABOOM

Note that it appears that this is only possible if `http_plugin` fails to initialize, as everywhere else uses `app().quit()` instead of `app.shutdown()`.

`app.quit()` has a different problem: `plugin_shutdown` is run after the event loop has already exited, so the handler passed to `post` is never run.  This avoids blowing up in `accept_transaction`, but means that the `amqp` thread is joined in the destructor of `amqp_trx_plugin_impl`.  All the members of `amqp_trx_plugin_impl` declared after `thread_pool` will have been destroyed at this point.  KABOOM.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,249 @@
...
+      my->amqp_trace.emplace( logger, my->thread_pool->get_executor(), my->amqp_trx_address, "trace" );
+
+      auto& consumer = my->amqp_trx->consume();
+      consumer.onSuccess( []( const std::string& consumer_tag ) {
+         fc_dlog( logger, "consume started: ${tag}", ("tag", consumer_tag) );
+      } );
+      consumer.onError( []( const char* message ) {
+         fc_wlog( logger, "consume failed: ${e}", ("e", message) );
+      } );
+      consumer.onReceived( [my=my](const AMQP::Message& message, uint64_t delivery_tag, bool redelivered) {
```
swatanabe-b1, line 196: This looks like a `shared_ptr` cycle.

programs/cleos/main.cpp
heifner: I'm not opposed to creating two plugins. One that publishes all traces to an exchange and another that serves as an RPC endpoint using an explicit reply-to or the current implicit reply-to and correlation-id approach. I thought the current plugin served both those by providing the publish all as an option but it might be cleaner to have them as separate plugins which would allow for different configuration and running them on different nodes.

As for using reply-to, see discussion on Correlation id here: https://www.rabbitmq.com/tutorials/tutorial-six-python.html I was just using an implicit reply-to of "trace". That "trace" could come in via the request if you think that is cleaner.

The original request for this plugin as I understood it was that it would run on the block signing node so that the "speculative execution" would be the actual execution into the block. It would not be run on other nodes.

swatanabe-b1: Correlation id is intended to distinguish multiple requests from a single client.  Even that tutorial recommends that each client should have its own queue and use reply-to.

heifner: It implies that there is overhead in the reply-to approach. RabbitMQ even has a proprietary way of doing reply-to to offset the overhead (https://www.rabbitmq.com/direct-reply-to.html). Also I don't think the use case is for a blocking client. I would imagine different parts of the system are going to be interested in transactions.

Proposal:
- Create a `amqp_trace_plugin` which will do what `amqp-trx-publish-all-traces` does now and nothing more.
- `amqp_trx_plugin` will be incoming only. I don't see a need for a blocking client. cleos will be blocking as it is now, but cleos is the the intended user. I only added support to cleos so it could be easily tested.

Thoughts? If the desire is for a reply-to, I can certainly add it.

swatanabe-b1: The overhead of reply-to is not because of reply-to per se, but is rather a consequence of using short-lived clients.  Making cleos listen to the broadcast would impose the same overhead (and more).  With that being said, I'm fine with not returning a trace from `amqp_trx_plugin`.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,169 @@
...
+   void handle_message( chain::packed_transaction_ptr trx ) {
+      const auto& tid = trx->id();
+      dlog( "received packed_transaction ${id}", ("id", tid) );
+
+      auto trx_in_progress = trx_in_progress_size.load();
+      if( trx_in_progress > def_max_trx_in_progress_size ) {
+         wlog( "Dropping trx, too many trx in progress ${s} bytes", ("s", trx_in_progress) );
+         if( trace_plug ) {
+            std::string err = "Dropped trx, too many trx in progress " + std::to_string( trx_in_progress ) + " bytes";
+            trace_plug->publish_error( trx->id().str(), chain::tx_resource_exhaustion::code_enum::code_value, std::move(err) );
```
swatanabe-b1, line 70: Is there anything that forces the `amqp_trace_plugin` to be initialized before `amqp_trx_plugin`?

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,169 @@
...
+      });
+
+      auto& consumer = my->amqp_trx->consume();
+      consumer.onSuccess( []( const std::string& consumer_tag ) {
+         dlog( "consume started: ${tag}", ("tag", consumer_tag) );
+      } );
+      consumer.onError( []( const char* message ) {
+         wlog( "consume failed: ${e}", ("e", message) );
+      } );
+      consumer.onReceived( [this](const AMQP::Message& message, uint64_t delivery_tag, bool redelivered) {
```
swatanabe-b1, line 134: `this` is somewhat scary as the callback can survive the destruction of `amqp_trx_shutdown` (It can be preserved by the `post` from `handle_message`).  I _think_ it's still safe, because the callback will only be executed by the thread pool, which is joined in `plugin_shutdown`.

heifner: Agree. I was counting on the thread_pool shutdown protection.

programs/cleos/main.cpp
```diff
@@ -436,12 +432,59 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+                  if( message.hasCorrelationID() && message.correlationID() == id && !received) {
+                     fc::datastream<const char*> ds( message.body(), message.bodySize() );
+                     fc::raw::unpack( ds, trace_msg );
+                     // don't ack as other consumers may be interested in trace: qp_trace.ack( delivery_tag );
+                     received = true;
+                     received_trace.set_value();
+                  }
+               } );
+               received_trace.get_future().wait();
+               thread_pool.stop();
```
swatanabe-b1, line 470: Is it possible to leave this scope with an exception?

.cicd/build.sh
heifner: Working on running these tests differently. One rabbitmq per test.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,175 @@
...
+      }
+
+      trx_in_progress_size += trx->get_estimated_size();
+      app().post( priority::medium_low, [my=shared_from_this(), trx{std::move(trx)}]() {
+         my->chain_plug->accept_transaction( trx,
+            [my, trx](const fc::static_variant<fc::exception_ptr, chain::transaction_trace_ptr>& result) mutable {
+               // publish to trace plugin as execptions are not reported via controller signal applied_transaction
+               if( result.contains<chain::exception_ptr>() ) {
+                  auto& eptr = result.get<chain::exception_ptr>();
+                  if( my->trace_plug ) {
```
swatanabe-b1, line 82: I get a segfault if the `amqp_trace_plugin` was not enabled.  I think the reason is that `my->trace_plug` is never null even if `amqp_trace_plugin` skipped initialization.

plugins/amqp_trace_plugin/amqp_trace_plugin.cpp
```diff
@@ -0,0 +1,151 @@
...
+      my->amqp_trace_exchange = options.at("amqp-trace-exchange").as<std::string>();
+   }
+   FC_LOG_AND_RETHROW()
+}
+
+void amqp_trace_plugin::plugin_startup() {
+   handle_sighup();
+   try {
+
+      if( my->amqp_trace_address.empty() ) return;
```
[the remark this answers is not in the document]
heifner: Switched this to check state of plugin instead of setting up a dependency.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: Interesting the segfault seems to be a macos only thing. Anyway, now check if initialized.

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/amqp_handler.hpp
```diff
@@ -0,0 +1,75 @@
+#pragma once
```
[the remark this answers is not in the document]
heifner: Sorry, thought I had cleaned that up.

programs/cleos/main.cpp
```diff
@@ -436,12 +432,61 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+            const auto& tid = msg.get<packed_transaction>().id();
+            const string id = tid.str();
+            {
+               eosio::chain::named_thread_pool trx_thread_pool( "mqtrx", 1 );
+               eosio::chain::named_thread_pool trace_thread_pool( "mqtrac", 1 );
+               eosio::amqp qp_trx( trx_thread_pool.get_executor(), amqp_address, "trx", [](const std::string& err){
+                  std::cerr << "AMQP trx error: " << err << std::endl;
+                  exit(1);
+               } );
+               eosio::amqp qp_trace( trace_thread_pool.get_executor(), amqp_address, "trace", [](const std::string& err){
```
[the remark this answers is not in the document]
heifner: Removed

programs/cleos/main.cpp
```diff
@@ -436,12 +432,61 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+               std::atomic_bool received = false;
+               auto& consumer = qp_trace.consume();
+               consumer.onError( []( const char* message ) {
+                  std::cerr << "consume failed: " << message << std::endl;
+               } );
+               consumer.onReceived( [&]( const AMQP::Message& message, uint64_t delivery_tag, bool redelivered ) {
+                  if( message.hasCorrelationID() && message.correlationID() == id && !received) {
+                     fc::datastream<const char*> ds( message.body(), message.bodySize() );
+                     fc::raw::unpack( ds, trace_msg );
+                     // don't ack as other consumers may be interested in trace: qp_trace.ack( delivery_tag );
```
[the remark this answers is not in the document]
heifner: Since the goal is for amqp trx to `nodeos` to be async with possible multiple readers, modified `cleos` to only submit trx and not try to consume traces.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -0,0 +1,83 @@
...
+   }
+
+   void reject( uint64_t delivery_tag ) {
+      channel_->reject( delivery_tag );
+   }
+
+private:
+   void declare_queue() {
+      std::promise<void> connected;
+      auto& queue = channel_->declareQueue( name_, AMQP::durable );
```
swatanabe-b1, line 59: `declare_queue` is executing in the application thread, right?  I'm not sure that this is safe.  The `io_context` is already running, so the operation may complete at any time, and has a race with setting the handler.  It looks to me as though accessing a `Channel` from multiple threads is unsafe, in general.  (I wasn't really sure from the amqp-cpp documentation, so I poked around the implementation a bit).

spoonincode: Yeah I'm surprised there is nothing explicit in the docs about threading, but I've seen some comments from Copernica devs that indicate as such
https://github.com/CopernicaMarketingSoftware/AMQP-CPP/issues/238#issuecomment-525279268

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
swatanabe-b1: I don't like the fact that the behavior changes depending on the order that the plugins are specified on the command line.  Would there be any problem with checking `my->trace_plug->get_state() == started` here instead of in `plugin_startup`?

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
heifner: Changed to call `declare_queue` on amqp thread.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -0,0 +1,99 @@
...
+      } );
+   }
+
+   void wait() {
+      if( connected_ ) {
+         auto r = connected_->get_future().wait_for( std::chrono::seconds( 10 ) );
+         if( r == std::future_status::timeout ) {
+            on_error( "AMQP timeout declaring queue" );
+         }
+         connected_.reset();
```
swatanabe-b1, line 64: Data race with `connected->set_value()` in `declare_queue` when the timeout expired.

heifner: Yep, I was trying to save promise memory by using a `unique_ptr`, I'll just go back to a raw promise.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
swatanabe-b1: I think this still leaves a window on shutdown where the trx plugin is running, but the trace plugin is not.  This looks mostly harmless to me.
- The trace will be dropped silently (which can already happen at startup and shutdown, anyway)
- The trx_plugin_impl object will never be destroyed, because the callback holds a `shared_ptr` to it.  (We're already shutting down, so memory leaks are not critical)

heifner: Seems like that kind of issue should be handled in `chain_plugin`  and/or `application::exec()` as all plugins which hook up to signals have that issue.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,186 @@
...
+         if( my->consume_message( message.body(), message.bodySize() ) ) {
+            my->amqp_trx->ack( delivery_tag );
+         } else {
+            my->amqp_trx->reject( delivery_tag );
+         }
+      } );
+
+   } catch( ... ) {
+      // always want plugin_shutdown even on exception
+      plugin_shutdown();
```
swatanabe-b1, line 165: What if we started `my->trace_plug` in this function?  It seems like the trace plugin won't be shut down in this case, because appbase doesn't know that it was started.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: My point was all plugins detach from the signals but the application thread can still be processing so they can miss the signal call.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: Added call to shutdown

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: I was proposing that since every plugin depends on `chain_plugin` and `chain_plugin` is soon to be a library instead of a "plugin" that `chain_plugin` should hold the invariant that by the time it shuts down the application thread has processed all queued tasks and that no controller signals will be emitted. Then all plugins can safely disconnect from signals knowing they will not miss anything. 
There would still be a race on posting to `app().post()` for plugins like `amqp_trx_plugin` and `net_plugin` and `http_plugin`. However, not having your trx processed seems like less of a problem then it being processed and put into a block but missing the signals.

swatanabe-b1: Right.  I'm not really concerned about the lost notifications right now, as it's a more general problem and we're not making it worse.  I'm somewhat more concerned about the `shared_ptr` created in `amqp_trx_plugin_impl::publish_error` preventing the `amqp_trx_plugin_impl` from being destroyed.

heifner: Not sure how that could be since: "Uninvoked handler objects that were scheduled for deferred invocation on the io_context, or any associated strand, are destroyed."

swatanabe-b1: I take it you're looking at https://www.boost.org/doc/libs/1_73_0/doc/html/boost_asio/reference/io_context/_io_context.html.  The thing is, the relevant io_context is not destroyed either, because it's a member of the `amqp_trace_plugin_impl` (I meant `trace_plugin` not `trx_plugin` above ).

heifner: Ah, ok, I was looking at the wrong thing. I see what  you mean now.

programs/cleos/main.cpp
heifner: The destructor of `named_thread_pool` calls `stop()`, not sure what else could be done.

swatanabe-b1: `trx_thread_pool` is destroyed after `qp_trx`.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -0,0 +1,95 @@
...
+   amqp( boost::asio::io_service& io_service, const std::string& address, std::string name, on_error_t on_err )
+         : name_( std::move( name ) )
+         , on_error( std::move(on_err) )
+   {
+      AMQP::Address amqp_address( address );
+      ilog( "Connecting to AMQP address ${a} - Queue: ${q}...", ("a", std::string( amqp_address ))( "q", name_ ) );
+
+      handler_ = std::make_unique<amqp_handler>( *this, io_service );
+      connection_ = std::make_unique<AMQP::TcpConnection>( handler_.get(), amqp_address );
+      channel_ = std::make_unique<AMQP::TcpChannel>( connection_.get() );
```
swatanabe-b1, line 23: Is this safe to run in the application thead?  I'd guess that it isn't, but I'm not sure.

programs/cleos/main.cpp
heifner: Added a try-catch

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -0,0 +1,188 @@
...
+      ilog( "Starting amqp_trx_plugin" );
+
+      my->thread_pool.emplace( "amqp_t", 1 );
+
+      my->amqp_trx.emplace( my->thread_pool->get_executor(), my->amqp_trx_address, "trx", [](const std::string& err) {
+         elog( "amqp error: ${e}", ("e", err) );
+         app().quit();
+      });
+
+      auto& consumer = my->amqp_trx->consume();
```
swatanabe-b1, line 144: This needs to be run in the amqp thread.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
heifner: Refactored to run in amqp thread.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: Refactored to run in amqp thread.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -0,0 +1,178 @@
...
+         : thread_pool_( "ampq", 1) // amqp is not thread safe, use only one thread
+         , name_( std::move( name ) )
+         , on_error_( std::move( on_err ) )
+         , on_consume_( std::move( on_consume ) )
+   {
+      AMQP::Address amqp_address( address );
+      ilog( "Connecting to AMQP address ${a} - Queue: ${q}...", ("a", std::string( amqp_address ))( "q", name_ ) );
+
+      handler_ = std::make_unique<amqp_handler>( *this, thread_pool_.get_executor() );
+      connection_ = std::make_unique<AMQP::TcpConnection>( handler_.get(), amqp_address );
```
swatanabe-b1, line 38: The handler is safe to construct in the main thread (at least as long as it happens before the construction of the connection).  The connection is definitely not safe to construct here.  `AMQP::TcpConnection` begins issuing async operations before it is fully constructed.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -1,66 +1,142 @@
...
+               std::vector<char> buf = f();
+               AMQP::Envelope env( buf.data(), buf.size() );
+               env.setCorrelationID( cid );
+               my->channel_->publish( exchange, my->name_, env, 0 );
+      } );
    }
 
-   void reject( uint64_t delivery_tag ) {
-      channel_->reject( delivery_tag );
+   /// close connection and explicitly stop thread pool execution
```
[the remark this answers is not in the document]
heifner: Right, I removed the `close()`, I'll update the comment.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -1,66 +1,142 @@
...
+   /// do not call from lambda's passed to publish or constructor e.g. on_error
+   void stop() {
+      // drain amqp queue
+      std::promise<void> stop_promise;
+      boost::asio::post( *handler_->amqp_strand(), [&]() {
+         stop_promise.set_value();
+      } );
```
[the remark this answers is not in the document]
heifner: In practice it does. I think because the I/O is scheduled before the shutdown.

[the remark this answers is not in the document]
heifner: In practice it does. I believe because the async  I/O is scheduled which allows it to finish.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
heifner: Moved to amqp thread.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -1,66 +1,142 @@
...
+/// All public interface methods can be called from any thread.
 class amqp {
 public:
+   // called from amqp thread on errors
    using on_error_t = std::function<void(const std::string& err)>;
-   amqp( boost::asio::io_service& io_service, const std::string& address, std::string name, on_error_t on_err )
-         : name_( std::move( name ) )
-         , on_error( std::move(on_err) )
+   // called from amqp thread on consume of message
+   // return empty optional for no ack/reject of message
```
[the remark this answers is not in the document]
heifner: Updated.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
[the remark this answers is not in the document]
swatanabe-b1: So, this should be taken as "best effort" and doesn't make a hard guarantee, then?  I suppose that's consistent with the general level of reliability that we for messages.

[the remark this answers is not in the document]
heifner: For this plugin, yes.

plugins/amqp_trace_plugin/include/eosio/amqp_trace_plugin/amqp_handler.hpp
```diff
@@ -0,0 +1,181 @@
...
+      if( handler_ ) {
+         // drain amqp queue
+         std::promise<void> stop_promise;
+         boost::asio::post( *handler_->amqp_strand(), [&]() {
+            stop_promise.set_value();
+         } );
+         stop_promise.get_future().wait();
+
+         thread_pool_.stop();
+         handler_.reset();
```
swatanabe-b1, line 83: This is scary, because it leaves the `TcpConnection` holding a dangling pointer, but it's safe (for now) because the first thing that `~TcpConnection` does is to clear its handler.

Issue #8971, opened 2020-04-17
Better error messaging on names with invalid characters


b1bart:
## Overview

In the following code, `string_to_uint64_t(...)` is used to convert a `std::string_view` at runtime to the proper value for an `eosio::name`:
https://github.com/EOSIO/eos/blob/4652e39ef346f51d0ff3c3c92f2dae78e2fc4acd/libraries/chain/name.cpp#L9-L16

However, `string_to_uint64_t(...)` does not validate that the input characters are part of the valid character set for `eosio::name` s, instead any invalid character is turned into a `0`:

https://github.com/EOSIO/eos/blob/4652e39ef346f51d0ff3c3c92f2dae78e2fc4acd/libraries/chain/include/eosio/chain/name.hpp#L21

Invalid characters are "detected" and an error is thrown however, it is thrown by the canonical check and presents an incorrect cause of the issue for instance:

```
Name not properly normalized (name: aaaa9, normalized: aaaa) 
```

PR #9186, opened 2020-06-05
struct name fix check #8971

dimas1185:
## Change Description

github issue:
https://github.com/EOSIO/eos/issues/8971

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/name.hpp
+++ b/libraries/chain/include/eosio/chain/name.hpp
@@ -22,9 +22,9 @@ namespace eosio::chain {
          return (c - '1') + 1;
       else if( c == '.')
          return 0;
-      else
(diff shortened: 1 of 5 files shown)


libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -18,24 +20,36 @@ namespace eosio::chain {
          return (c - 'a') + 6;
       if( c >= '1' && c <= '5' )
          return (c - '1') + 1;
+      else if( c == '.')
+         return 0;
+      else
```
allenhan2, line 25: no need to else here, directly call assert, and remove the unreachable return value

PR #9192, opened 2020-06-08
small cleanups to cfile/datastream changes in ship-v1

spoonincode:
## Change Description

These are some fixups to cfile & datastream changes in the ship-v1 branch that were already picked in to #9183 and received feedback there.

Getting the cfile eof check right is particularly tricky and I don't claim to have solved all the interesting corner cases one can concoct.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/fc/include/fc/io/cfile.hpp
+++ b/libraries/fc/include/fc/io/cfile.hpp
@@ -65,6 +65,11 @@ public:
    }
 
    void seek( long loc ) {
+      seek_end( 0 );
+      size_t sz = tellp();
+      if( loc > sz )
+         throw std::ios_base::failure( "cfile: " + _file_path.generic_string() +
+                                       " unable to seek to: " + std::to_string(loc) );
       if( 0 != fseek( _file.get(), loc, SEEK_SET ) ) {
          throw std::ios_base::failure( "cfile: " + _file_path.generic_string() +
                                        " unable to SEEK_SET to: " + std::to_string(loc) );
@@ -79,9 +84,17 @@ public:
    }
 
    void skip( long loc) {
-      if( 0 != fseek( _file.get(), loc, SEEK_CUR ) ) {
-         throw std::ios_base::failure( "cfile: " + _file_path.generic_string() +
-                                       " unable to SEEK_CUR to: " + std::to_string(loc) );
+      if( loc < 0 ) {
+         if( 0 != fseek( _file.get(), loc, SEEK_CUR ) ) {
+            throw std::ios_base::failure( "cfile: " + _file_path.generic_string() +
+                                          " unable to SEEK_CUR to: " + std::to_string(loc) );
+         }
+      }
+      else while( loc > 0 ) {
+         char bit_bucket[4096];
+         auto read_max = std::min( long(sizeof(bit_bucket)), loc );
+         read( bit_bucket, read_max );
+         loc -= read_max;
       }
    }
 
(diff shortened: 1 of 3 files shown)


libraries/fc/include/fc/io/cfile.hpp
```diff
@@ -79,9 +84,17 @@ class cfile {
...
+         if( 0 != fseek( _file.get(), loc, SEEK_CUR ) ) {
+            throw std::ios_base::failure( "cfile: " + _file_path.generic_string() +
+                                          " unable to SEEK_CUR to: " + std::to_string(loc) );
+         }
+      }
+      else while( loc > 0 ) {
+         char bit_bucket[4096];
+         auto read_max = std::min( long(sizeof(bit_bucket)), loc );
+         read( bit_bucket, read_max );
+         loc -= read_max;
```
heifner, line 97: Does this mean `fseek` just doesn't work? Why is this needed?

spoonincode: Seeking beyond the end of the file isn't an error, you need to actually read from it to see an EOF error. I thought maybe just do the `fseek` and call `eof()` but the problem with that is that if someone deliberately tries to `skip()` to the end _because they then want to append at the end_, there shouldn't be an error in the case.

swatanabe-b1: Can't you also seek past the end of the file and start writing as well?  Anyway, seeking to the end should definitely not be an error.

libraries/fc/include/fc/io/datastream.hpp
```diff
@@ -118,7 +119,7 @@ class datastream<Streambuf, typename std::enable_if_t<std::is_base_of_v<std::str
       buf.pubseekoff(off, std::ios::beg);
       return true;
    }
-   bool remaining() { return buf.in_avail(); }
+   bool remaining() { return buf.in_avail() >= 0; }
```
swatanabe-b1, line 122: This may still be wrong if `in_avail` returns 0.  Use `buf.sgetc() != EOF`.

libraries/fc/include/fc/io/cfile.hpp
spoonincode: Instead of the bit bucket, we could do something like `seek()` that probes the file size instead. Without knowing average size of `skip()` it's hard to know which will be more efficient (I expect each time probing the file size is a syscall).

libraries/fc/include/fc/io/datastream.hpp
spoonincode: Taking your word for it, I didn't investigate this one

libraries/fc/include/fc/io/cfile.hpp
huangminghuang: Why do you need to detect whether the file position passed the end of file exactly at seek() and skip()? In normal C programing, it's detected at the next read() operation and it's generally good enough. In detecting at seek()/skip(), you are adding a lot of overhead for normal usage of these two operations.

spoonincode: okay, if that's what you all want, I will remove it

PR #9194, opened 2020-06-10
Add trace plugin  API test

allenhan2:
## Change Description

Testing:
1. Start Nodeos and Keosd
2. send transaction
3. get transaction id and block number
4. call trace API to get block, and verify transaction is in the block data

Jira: https://blockone.atlassian.net/browse/EPE-104

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+        for idx in range(len(self.account_names)):
+            self.accounts[idx].name =  self.account_names[idx]
+            self.walletMgr.importKey(self.accounts[idx], testWallet)
+        for account in self.accounts:
+            node.createInitializeAccount(account, self.cluster.eosioAccount, buyRAM=1000000, stakedDeposit=5000000, waitForTransBlock=True, exitOnError=True)
+        time.sleep(self.sleep_s)
+
+    def test_TraceApi(self) :
+        node = self.cluster.getNode(0)
+        for account in self.accounts:
```
brianjohnson5972, line 45: this should be in startEnv in the parent class

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+            self.assertIsNotNone(node.verifyAccount(account))
+
+        expectedAmount = Node.currencyIntToStr(5000000, CORE_SYMBOL)
```
brianjohnson5972, line 48: the portion of this that is verifying the balance, should go in startEnv in the parent class

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+
+        xferAmount = Node.currencyIntToStr(123456, CORE_SYMBOL)
+        trans = node.transferFunds(self.accounts[0], self.accounts[1], xferAmount, "test transfer a->b")
+        transId = Node.getTransId(trans)
+        blockNum = Node.getTransBlockNum(trans)
+
+        self.assertEqual(node.getAccountEosBalanceStr(self.accounts[0].name), Utils.deduceAmount(expectedAmount, xferAmount))
+        self.assertEqual(node.getAccountEosBalanceStr(self.accounts[1].name), Utils.addAmount(expectedAmount, xferAmount))
+        time.sleep(self.sleep_s)
+        base_cmd_str = ("curl http://%s:%s/v1/") % (TestHelper.LOCAL_HOST, node.port)
```
brianjohnson5972, line 63: make a method Node.get_trace_api_block, that uses Node.processCurlCmd so you just need to pass in the block num here, and if we add more trace_api testing, we can reuse this.

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+    @classmethod
+    def setUpClass(self):
+        self.cleanEnv(self)
+        self.startEnv(self)
+
+    @classmethod
+    def tearDownClass(self):
+        self.cleanEnv(self)
+
+if __name__ == "__main__":
```
brianjohnson5972, line 84: add a static method run_tests() to SingleNodeTestCase for if check and unittest.main(), I'm guessing you will have to pass in __name__

tests/Cluster.py
```diff
@@ -166,7 +166,7 @@ def setAlternateVersionLabels(self, file):
     # pylint: disable=too-many-statements
     def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="mesh", delay=1, onlyBios=False, dontBootstrap=False,
                totalProducers=None, sharedProducers=0, extraNodeosArgs=" --http-max-response-time-ms 990000 ", useBiosBootFile=True, specificExtraNodeosArgs=None, onlySetProds=False,
-               pfSetupPolicy=PFSetupPolicy.FULL, alternateVersionLabelsFile=None, associatedNodeLabels=None, loadSystemContract=True):
+               pfSetupPolicy=PFSetupPolicy.FULL, alternateVersionLabelsFile=None, associatedNodeLabels=None, loadSystemContract=True, enableTrace=False):
```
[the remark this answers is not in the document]
allenhan2: fixed. move the trace outside of launch

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+import unittest
+
+from testUtils import Utils
+from Cluster import Cluster
+from TestHelper import TestHelper
+from Node import Node
+from WalletMgr import WalletMgr
+from core_symbol import CORE_SYMBOL
+
+class TraceApiPluginTest(unittest.TestCase):
```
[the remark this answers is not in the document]
allenhan2: Originally tried with SingleNode, did not work, there was issue of creating account except system account `eosio`, so eventually switch to cluster instead.

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+    sleep_s = 1
+    cluster=Cluster(walletd=True, defproduceraPrvtKey=None)
+    walletMgr=WalletMgr(True)
+    accounts = []
+    account_names = ["alice", "bob", "charlie"]
```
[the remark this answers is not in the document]
allenhan2: Moved to startEnv

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,85 @@
...
+    # kill nodeos and keosd and clean up dir
+    def cleanEnv(self) :
+        self.cluster.killall(allInstances=True)
+        self.cluster.cleanup()
+        self.walletMgr.killall(allInstances=True)
+        self.walletMgr.cleanup()
+
+    # start keosd and nodeos
+    def startEnv(self) :
+        self.cluster.launch(totalNodes=1, enableTrace=True)
```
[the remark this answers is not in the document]
allenhan2: This will takes more time, let's create a separated Jira to address the python test frame work issue.

tests/trace_plugin_test.py
allenhan2: StartEnv is for every test case, not for test suite, rather wants to keep here

tests/trace_plugin_test.py
allenhan2: It is better to stay in the test case, otherwise, it will go to the global variable or to pass in as parameter, not flexible, better to let test case define the test data

tests/trace_plugin_test.py
allenhan2: You mean Node.py, right? It is very generic function, better to make it specific for specific plugin. But I can create a local function.

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,91 @@
...
+        self.assertTrue(isTrxInBlock)
+
+    @classmethod
+    def setUpClass(self):
+        self.cleanEnv(self)
+        self.startEnv(self)
+
+    @classmethod
+    def tearDownClass(self):
+        self.cleanEnv(self)
```
brianjohnson5972, line 88: either need to check if test passed or just don't call cleanup() when tearing down

tests/trace_plugin_test.py
```diff
@@ -0,0 +1,91 @@
...
+            account_balances.append(amount)
+
+        xferAmount = Node.currencyIntToStr(123456, CORE_SYMBOL)
+        trans = node.transferFunds(self.accounts[0], self.accounts[1], xferAmount, "test transfer a->b")
+        transId = Node.getTransId(trans)
+        blockNum = Node.getTransBlockNum(trans)
+
+        self.assertEqual(node.getAccountEosBalanceStr(self.accounts[0].name), Utils.deduceAmount(expectedAmount, xferAmount))
+        self.assertEqual(node.getAccountEosBalanceStr(self.accounts[1].name), Utils.addAmount(expectedAmount, xferAmount))
+        ret_json = self.get_block(blockNum, node)
```
brianjohnson5972, line 70: before this, call node.getBlock(blockNum) to retrieve the block and verify that you can find the transaction id in the block, this way if this test case ever fails, we will know whether the problem is in trace_api or before it.

PR #9196, opened 2020-06-10
Resource monitor plugin -- develop branch

linhuang-blockone:
## Change Description
This is a replicated PR of https://github.com/EOSIO/eos/pull/9098 for develop branch.
Only differences are enabling resource monitor plugin by default for future releases, creating
new test cases for it, and moving  resource_monitor_impl back to resource_monitor_plugin.cpp
to be consistent with other plugins.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [X] Documentation Additions

**Configurable Parameters**

*resource-monitor-interval-seconds*
Time in seconds between two consecutive checks of resource usage. It should be between 1 and 300 (5 minutes). This is optional; the default value is 2 seconds.

*resource-monitor-space-threshold*
Threshold in terms of percentage of used space vs total space.  If used space is above (threshold - 5%), a warning is generated.  If used  space is above the threshold and resource-monitor-not-shutdown-on- threshold-exceeded is enabled, a graceful shutdown is initiated. The value   should be between 6 and 99

*resource-monitor-not-shutdown-on-threshold-exceeded*
Used to indicate nodeos will not shutdown when threshold is exceeded.


plugins/resource_monitor_plugin/test/test_resmon_plugin.cpp
```diff
@@ -0,0 +1,144 @@
...
+   void set_program_options() {
+      options_description dummy;
+      _my.set_program_options(dummy, _cfg);
+   }
+
+   void initialize(const std::vector<std::string>& args){
+      // We only have at most 3 arguments. OK to hardcodied in test
+      // programs.
+      const char* argv[10];
+      // argv[0] is program name, no need to fill in
```
larryk85, line 29: You should have some kind of assert here that the argument size is what you expect.  If not then it could silently overflow argv if in the future if a lot more args are added to resmon.

linhuang-blockone: Thanks for pointing out this! Sorry for the overlook. Will change.

PR #9198, opened 2020-06-11
Reenable OC and update it to the new intrinsic wrappers.

swatanabe-b1:
## Change Description

- Remove unvalidated_ptr and pass the parameters for memcpy and friends in a
  struct.  Checking the memory range inside the intrinsics does not work
  with OC.
- Minor adjustments where OC was outdated.
- Add test for memcpy with a range that crosses the end of memory.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/webassembly/eos-vm-oc.hpp
```diff
@@ -329,471 +183,194 @@ struct wasm_function_type_provider<Ret(Args...)> {
...
-         : [applyContextOffset] "i" (cb_ctx_ptr_offset)
-         );
-      (class_from_wasm<Cls>::value(*ctx).*Method)(params...);
-      return void_type();
+          : [cPtr] "=r" (ctx)
+          : [applyContextOffset] "i" (cb_ctx_ptr_offset)
+          );
+      Interface host(*ctx);
+      eos_vm_oc_type_converter tc{&host, eos_vm_oc_execution_interface{stack + sizeof...(A)}};
+      return result_resolver{tc}, eosio::vm::invoke_with_host<F, Preconditions, native_args>(tc, &host, std::make_index_sequence<sizeof...(A)>());
```
spoonincode, line 342: with clang 10 there are 70+ warnings on this `result_resolver` about `expression result unused`

swatanabe-b1: That's unfortunate.

libraries/chain/include/eosio/chain/webassembly/common.hpp
```diff
@@ -79,6 +94,17 @@ namespace eosio { namespace chain {
          return {ptr};
       }
 
+      template <typename T>
+      auto from_wasm(void* ptr, vm::tag<T> = {}) const
+         -> std::enable_if_t< vm::is_argument_proxy_type_v<T> &&
+                              std::is_pointer_v<typename T::proxy_type>, T> {
+         if constexpr(T::is_legacy()) {
+            EOS_ASSERT(ptr != this->get_interface().get_memory(), wasm_execution_error, "references cannot be created for null pointers");
```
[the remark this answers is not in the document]
swatanabe-b1: Actually, it does—from the contract's POV.  The pointer we see here is the wasm pointer + the base of memory.  If ptr is equal to the start of linear memory, that implies that the offset was 0, which implies that the contract passed the integer value 0 which is interpreted as the null pointer.

PR #9203, opened 2020-06-12
Use new block_start signal for clearing onblock cached value

heifner:
## Change Description

- Update to use new scheme for tracking `onblock`

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
+++ b/plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
@@ -138,24 +138,17 @@ struct compressed_proof_generator_impl {
    > reversible_action_entries_index_type;
    reversible_action_entries_index_type reversible_action_entries_index;
 
-   static bool is_onblock(const chain::transaction_trace_ptr& p) {
-      if(p->action_traces.empty())
-         return false;
-      auto& act = p->action_traces[0].act;
-      if(act.account != eosio::chain::config::system_account_name || act.name != N(onblock) ||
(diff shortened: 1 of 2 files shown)


spoonincode:
It really would have been nice to come up with a unit test for this. I spent a few minutes on it and couldn't come up with something easily though

plugins/amqp_compressed_proof_plugin/amqp_compressed_proof_plugin.cpp
```diff
@@ -328,6 +323,9 @@ void amqp_compressed_proof_plugin::plugin_initialize(const variables_map& option
       controller.accepted_block.connect([&generator=*my->proof_generator](const chain::block_state_ptr& p) {
          generator.on_accepted_block(p);
       });
+      controller.block_start.connect([&generator=*my->proof_generator](uint32_t block_num) {
```
spoonincode, line 326: we should also wire up the signal in the unit test file

spoonincode: actually you know.. that's just code duplication. probably what should be done is add a metohd  to `compressed_proof_generator` that takes a controller and wires it up.

spoonincode: (I'll do that in the base PR)

PR #9207, opened 2020-06-14
call boost program option notifiers before plugin initialize

spoonincode:
## Change Description

This PR changes boost program option notifiers to fire before plugin initialization instead of after initialization but before startup. Current behavior has bitten me an embarrassing number of times and I’m not immediately sure of any drawback moving the notifiers earlier given our typical plugin pattern.

Using notifiers, IMO, cleans up option handling. For example, without using the notifiers our typical pattern is something like

```c++
struct my_plugin_impl {
   std::string server_address;
}

void my_plugin::set_program_options(options_description&, options_description& cfg) {
   cfg.add_options()("https-server-address", bpo::value<string>(), "set the address");
}

void my_plugin::plugin_initialize(const variables_map& options) {
   if( options.count( "https-server-address" )
     my->server_address = options.at("https-server-address").as<string>();
   if(my->server_address.size())
      dothedew();
   //...
```

But using the notifiers it can be more like
```c++
struct my_plugin_impl {
   std::string server_address;
}

void my_plugin::set_program_options(options_description&, options_description& cfg) {
   cfg.add_options()("https-server-address", bpo::value<string>()->notifier([this](const string& a) {
      my->server_address = a;
   }), "set the address");
}

void my_plugin::plugin_initialize(const variables_map& options) {
   if(my->server_address.size())
      dothedew();
   //...
```
In particular notice that the `https-server-address` string doesn't have to be copy pastaed around, and that there is no need for the `options.at("https-server-address").as<string>();` mouthful.

But the code above won't work today because the notifiers are fired _after_ plugin_initialize() (before startup()). This PR moves the notifiers to occur before plugin_initialize() allowing plugin_initialize() to make use of values set in notifiers.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/appbase/application.cpp
+++ b/libraries/appbase/application.cpp
@@ -340,6 +340,13 @@ bool application::initialize_impl(int argc, char** argv, vector<abstract_plugin*
       std::cerr << "         removing these items." << std::endl;
    }
 
+   try {
+      bpo::notify(options);
+   } catch(...) {
+      std::cerr << "Failed to notify plugins of options\n";
+      return false;
+   }
+
    if(options.count("plugin") > 0)
    {
       auto plugins = options.at("plugin").as<std::vector<std::string>>();
@@ -355,8 +362,6 @@ bool application::initialize_impl(int argc, char** argv, vector<abstract_plugin*
       for (auto plugin : autostart_plugins)
          if (plugin != nullptr && plugin->get_state() == abstract_plugin::registered)
             plugin->initialize(options);
-
-      bpo::notify(options);
    } catch (...) {
       std::cerr << "Failed to initialize\n";
       return false;
--- a/libraries/appbase/application.cpp
+++ b/libraries/appbase/application.cpp
@@ -343,8 +343,8 @@ bool application::initialize_impl(int argc, char** argv, vector<abstract_plugin*
    try {
       bpo::notify(options);
    } catch(...) {
-      std::cerr << "Failed to notify plugins of options\n";
-      return false;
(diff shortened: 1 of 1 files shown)


libraries/appbase/application.cpp
```diff
@@ -340,6 +340,13 @@ bool application::initialize_impl(int argc, char** argv, vector<abstract_plugin*
       std::cerr << "         removing these items." << std::endl;
    }
 
+   try {
+      bpo::notify(options);
+   } catch(...) {
```
revl, line 345: I would catch `std::exception` instead of or in addition to the ellipsis and then print the exception message to cerr.

spoonincode: I wonder if it makes any sense to make appbase depend on fc so it can use fc's exception helpers. The only reason to catch here is to make a useful error message, otherwise the exception still just makes it up to the handler in main() AFAICT. So maybe it should just print the message it currently is and rethrow?

revl: I'm concerned when error messages are suppressed like that, not so much in this particular case, but there's another catch-everything statement at the bottom of that method.

The method already uses a mixture of returning false and throwing exceptions for error handling, so it doesn't really matter how to report the error. I'd probably do something like `std::cerr << "Failed to notify plugins of options: " << e.what() << std::endl;` because it seems to be consistent with the majority of error handling cases in this method.

PR #9209, opened 2020-06-15
add empty content http request handling

allenhan2:
## Change Description

related to PR:
https://github.com/EOSIO/eos/pull/9033

The fix: 
if empty content string (such as "{}", space or tab is allowed in between curly braces) will be treated same as no input.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
@@ -4,6 +4,7 @@
 #include <fc/reflect/reflect.hpp>
 #include <fc/io/json.hpp>
 #include <eosio/chain/exceptions.hpp>
+#include<boost/algorithm/string.hpp>
 
 namespace eosio {
    using namespace appbase;
@@ -159,6 +160,27 @@ namespace eosio {
       error_info error;
    };
 
+   static inline bool isEmptyContent(std::string& body) {
+      boost::algorithm::trim(body);
+
+      if (body.empty()) {
+         return true;
+      }
+      const size_t body_size = body.size();
+      if ((body_size > 1) && (body.at(0) == '{') && (body.at(body_size - 1) == '}')) {
+         for(size_t idx=1; idx<body_size-1; ++idx)
+         {
+            if ((body.at(idx) != ' ') && (body.at(idx) != '\t'))
+            {
+               return false;
+            }
+         }
+      } else {
+         return false;
+      }
+      return true;
+   }
+
    enum class http_params_types {
       no_params_required = 0,
       params_required = 1,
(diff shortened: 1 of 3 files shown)


plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
```diff
@@ -159,30 +160,52 @@ namespace eosio {
       error_info error;
    };
 
+   inline bool isEmptyContent(std::string& body) {
```
heifner, line 163: Should be snake case not camel case.
I don't expect an `is` method to modify the string.
Seems like this could be easily implemented with a `string_view` instead.

allenhan2: I was wondering whether need to implement trim, for handling broader scenarios, just added in case that leading or trailing spaces being sent. 
If there is no need to trim here, rather like to change to const string& string_view

allenhan2: Or, I can pull the trim outside of checking.

allenhan2: Move trim out side of checking, so is method won't modify. pushed change

plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
```diff
@@ -166,23 +186,25 @@ namespace eosio {
    };
 
    template<typename T, http_params_types params_type>
-   T parse_params(const std::string& body) {
+   T parse_params(std::string& body) {  // needs to trim space,
```
b1bart, line 189: While I don't think this will have issues, I don't like the practice of mutating inputs in a way that is not clear to the caller.  

Please restore the `const` qualifier on the input and either change to `boost::algorithm::trim_copy` or create a `std::string_view` of the equivalent sub-string.  The former is fine unless we are really concerned with the extra performance impact of the copy. 

Unfortunately, I don't know of a clean algorithm to do so and neither `std::string_view` or `boost::string_view` is compatible with `boost::algorithm::trim(_copy)`

allenhan2: I don't see a clean way to trim w/o copy or modifying, or implement a string_view trim.

allenhan2: Using a trimmed string_view, so no change to user input data

PR #9210, opened 2020-06-15
Fix eosio-blocklog trim front

huangminghuang:
## Change Description

Fix trim block front for older version of block log.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -87,20 +87,25 @@ namespace eosio { namespace chain {
 
       template <typename Stream>
       void write_to(Stream& ds) const {
-         EOS_ASSERT(version >= block_x_start_version, block_log_exception, "this method does not support writeing block log ${version}",
-                    ("version", version));
          ds.write(reinterpret_cast<const char*>(&version), sizeof(version));
-         ds.write(reinterpret_cast<const char*>(&first_block_num), sizeof(first_block_num));
+         if (version != initial_version) {
+            ds.write(reinterpret_cast<const char*>(&first_block_num), sizeof(first_block_num));
 
-         std::visit(overloaded{[&ds](const chain_id_type& id) { ds << id; },
-                               [&ds](const genesis_state& state) {
-                                  auto data = fc::raw::pack(state);
-                                  ds.write(data.data(), data.size());
-                               }},
-                    chain_context);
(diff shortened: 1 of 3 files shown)


spoonincode:
Looks like it resolved my problem

unittests/restart_chain_tests.cpp
```diff
@@ -232,7 +235,8 @@ BOOST_FIXTURE_TEST_CASE(test_light_validation_restart_from_block_log_with_pruned
    BOOST_REQUIRE_NO_THROW(block_log::repair_log(blocks_dir));
 }
 
-BOOST_AUTO_TEST_CASE(test_trim_blocklog_front) {
+
+void trim_blocklog_front() {
```
brianjohnson5972, line 239: I think this should pass in an optional<uint32_t> initial_default_version, optional<uint32_t> pre_trim_default_version, uint32_t expected_trim_version and call set_version before creating the initial block log and then before calling trim_block_log_front and then verify the version for the trimmed block log.

libraries/chain/block_log.cpp
```diff
@@ -87,20 +87,25 @@ namespace eosio { namespace chain {
...
+                                 }},
+                     chain_context);
 
-         auto totem = block_log::npos;
-         ds.write(reinterpret_cast<const char*>(&totem), sizeof(totem));
+            auto totem = block_log::npos;
+            ds.write(reinterpret_cast<const char*>(&totem), sizeof(totem));
+         }
+         else {
+            const auto& state = std::get<genesis_state>(chain_context);
```
brianjohnson5972, line 105: (Missed this) If we are opening this up to supporting initial_version, then we need to add some logic to reset to change the version so that the header information remains consistent to keep the block log valid, unless we can see that reset is not allowed to be called with version == 1.

unittests/restart_chain_tests.cpp
brianjohnson5972: this is already verified by calling smoke_test, so resolving this

libraries/chain/block_log.cpp
brianjohnson5972: This only happens if reset is called after calling set_default, which is a test method, so it won't.

Issue #9176, opened 2020-06-03
Make Nodeos fatally stop when both history and state_history plugins are loaded


aclark-b1:
Today I ran across someone else running history and state_history on the same nodeos instance.  I referred them to this issue and advised against it.

It would be great to make nodeos fatally stop when both plugins are loaded.

_Originally posted by @matthewdarwin in https://github.com/EOSIO/eos/issues/6377#issuecomment-560463018_

There is a longer term solution we're evaluating here, but in the meantime Nodeos should not allow both history and state_history plugins to be enabled at the same time.

tbfleming:
An additional change to consider: remove history plugin support from eosjs

matthewdarwin:
Removing the old history plugin seems like a much better idea to solve this issue.

PR #9215, opened 2020-06-17
Stop Nodeos or issue warning if both state history plugin and history plugin are enabled

linhuang-blockone:
## Change Description

This PR resolves Jira https://blockone.atlassian.net/browse/EPE-135 (https://github.com/EOSIO/eos/issues/9176).
When both state history plugin and history plugin are enabled,
if `state-history-allow-legacy-history-plugin` option is set, a warning is issued but *Nodeos* is let to continue; if `state-history-allow-legacy-history-plugin` option is not set, *Nodeos* will gracefully shut down with a message.

## Change Type
**Select ONE**
- [X] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [X] Documentation Additions

A new configuration option `state-history-allow-legacy-history-plugin`  is added to state history plugin. When both state history plugin and history plugin are enabled,
if `state-history-allow-legacy-history-plugin` is set, a warning is issued but *Nodeos* is let to continue; if it is is not set, *Nodeos* will gracefully shut down.

### DIFF
--- a/libraries/appbase/include/appbase/application.hpp
+++ b/libraries/appbase/include/appbase/application.hpp
@@ -156,6 +156,18 @@ namespace appbase {
             return *ptr;
          }
 
+         template<typename Plugin>
+         bool is_plugin_initialized()const {
+            string name = boost::core::demangle(typeid(Plugin).name());
+            for ( const auto& plugin: initialized_plugins ) {
+               if ( plugin->name() == name ) {
+                  return true;
+               }
+            }
+
+            return false;
+         }
+
          /**
           * Fetch a reference to the method declared by the passed in type.  This will construct the method
           * on first access.  This allows loose and deferred binding between plugins
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -3,6 +3,7 @@
 #include <eosio/state_history/log.hpp>
 #include <eosio/state_history/serialization.hpp>
 #include <eosio/state_history_plugin/state_history_plugin.hpp>
+#include <eosio/history_plugin/history_plugin.hpp>
 
 #include <boost/asio/bind_executor.hpp>
 #include <boost/asio/ip/host_name.hpp>
@@ -42,6 +43,7 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
    fc::optional<state_history_traces_log>                     trace_log;
    fc::optional<state_history_chain_state_log>                chain_state_log;
    bool                                                       stopping = false;
+   bool                                                       allow_legacy_history_plugin = false;
    fc::optional<scoped_connection>                            applied_transaction_connection;
    fc::optional<scoped_connection>                            block_start_connection;
    fc::optional<scoped_connection>                            accepted_block_connection;
@@ -278,6 +280,16 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
    std::map<session*, std::shared_ptr<session>> sessions;
 
    void listen() {
+      if ( app().is_plugin_initialized<history_plugin>() ) {
+         wlog( "Legacy history plugin is running; this will cause random crashes." );
+         if ( allow_legacy_history_plugin ) {
+            wlog( "With state-history-allow-legacy-history-plugin set, proceed at your own risk" );
+         } else {
+            wlog( "With state-history-allow-legacy-history-plugin not set, shutting down." );
+            appbase::app().quit();
+         }
+      }
+
       boost::system::error_code ec;
 
       auto address  = boost::asio::ip::make_address(endpoint_address);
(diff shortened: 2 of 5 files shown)


libraries/appbase/include/appbase/application.hpp
```diff
@@ -156,6 +156,18 @@ namespace appbase {
             return *ptr;
          }
 
+         template<typename Plugin>
+         bool is_plugin_initialized()const {
```
spoonincode, line 160: Could something like `app().get_plugin<producer_plugin>().get_state()` work instead of adding this?

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -366,6 +378,8 @@ void state_history_plugin::set_program_options(options_description& cli, options
            "enable debug mode for trace history");
    options("context-free-data-compression", bpo::value<string>()->default_value("zlib"), 
            "compression mode for context free data in transaction traces. Supported options are \"zlib\" and \"none\"");
+   options("state-history-allow-legacy-history-plugin", bpo::bool_switch()->default_value(false),
+           "allow legacy history plugin to run at the same time. Caution: it will cause random crashes");
```
heifner, line 382: Why have this? I don't think this is needed. No one should run both.

heifner:
With the addition of `trace_api_plugin` & `chain_plugin` `--enable-account-queries` I think we should remove `history_plugin` in `develop`. @b1bart

plugins/state_history_plugin/state_history_plugin.cpp
larryk85: @heifner There are cases where you would want to run both for some amount of time to transition on a single node.  People have used it in this way and that's why we found out about the bug, so if they are okay with throwing caution to the wind then they can.  @linhuang-blockone I would state that it can cause random crashes, not that it will every time cases crashes.

heifner: We have known issues history_plugin, it flat out does the wrong thing in a number of known cases. In my opinion we should kill it immediately even in 2.0.x.

matthewdarwin: I agree.  I don't know why we still have this around and are enabling people to do the wrong thing.  No valid use case for running history + state_history in the same node at the same time.

libraries/appbase/include/appbase/application.hpp
linhuang-blockone: Thanks Matt. Just tried and get_state worked.

plugins/state_history_plugin/state_history_plugin.cpp
linhuang-blockone: I will follow any decision once agreed upon.

FYI: This fix has already uncovered a number of tests which use both state_history plugin and directly or indirectly history plugin. I looked at one failed test: Ubuntu 18.04 ShiP test: it used history_api plugin (which requires history plugin) and state_history plugin.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -296,6 +298,16 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
    std::map<session*, std::shared_ptr<session>> sessions;
 
    void listen() {
+      if ( app().is_plugin_initialized<history_plugin>() ) {
+         wlog( "Legacy history plugin is running; this will cause random crashes." );
+         if ( allow_legacy_history_plugin ) {
+            wlog( "With state-history-allow-legacy-history-plugin set, proceed at your own risk" );
+         } else {
+            wlog( "With state-history-allow-legacy-history-plugin not set, shutting down." );
```
tbfleming, line 306: I'm afraid this message will encourage people to set the flag. Also, since it's shutting down, use `elog`.

Maybe "Stopping to prevent random crashes; disable the legacy history plugin to continue". People who are desperate to do the wrong thing will find the flag without you mentioning it.

tbfleming:
If we end up having to keep the legacy plugin, I recommend:
* Completely purging all mention of it in the documentation (doc team task)
* If we don't purge it from the documentation, do not mention this flag in the documentation (doc team anti-task)

Is there a way to hide the option from the command-line --help?

tbfleming:
> With the addition of `trace_api_plugin` & `chain_plugin` `--enable-account-queries` I think we should remove `history_plugin` in `develop`. @b1bart I will add I also think we should remove it in `release/2.0.x`. We added these new features in `release/2.0.x` so seems we should be able to remove `history_plugin` in `release/2.0.x` as well.

I agree with removing it in a major release. Removing it in a patch release seems a bit aggressive. I'm on the fence because of the known bugs.

PR #9217, opened 2020-06-17
Improve parsing of RabbitMQ-related command line arguments in rodeos - develop

revl:
## Change Description

This commit tightens up parsing of RabbitMQ-related command line arguments (it simplifies the code, improves readability, and adds a comment and a unit test).

## Change Type

- [x] Stability bug fix

On Linux, specifying `--stream-rabbits-exchange` crashes rodeos with SEGFAULT. On macOS, it was observed that the log contained "RabbitMQ Connected Successfully! Exchange п��". Apparently, there's a read access of deleted memory in the onSuccess lambda. It can no longer be reproduced after this commit. If this commit doesn't fix the Linux bus error, I'll investigate it separately.

## Consensus Changes

- [ ] Consensus Changes

## API Changes

- [ ] API Changes

## Documentation Additions

- [x] Documentation Additions

A comment to `parse_rabbitmq_address` in `rabbitmq.hpp` expands a little bit on the format of the argument of the rodeos options `--stream-rabbits` and `--stream-rabbits-exchange`.

### DIFF
--- a/programs/rodeos/streams/rabbitmq.hpp
+++ b/programs/rodeos/streams/rabbitmq.hpp
@@ -81,7 +81,8 @@ class rabbitmq : public stream_handler {
       }
 
       auto& exchange = channel_->declareExchange(exchangeName_, type);
-      exchange.onSuccess([this]() { ilog("RabbitMQ Connected Successfully!\n Exchange ${e}", ("e", exchangeName_)); });
+      exchange.onSuccess(
+            [en = exchangeName_]() { ilog("RabbitMQ Connected Successfully!\n Exchange ${e}", ("e", en)); });
       exchange.onError([](const char* error_message) {
          throw std::runtime_error("RabbitMQ Exchange error: " + std::string(error_message));
       });
--- a/programs/rodeos/streams/rabbitmq.hpp
+++ b/programs/rodeos/streams/rabbitmq.hpp
@@ -1,9 +1,9 @@
 #pragma once
 
-#include "stream.hpp"
 #include "amqpcpp.h"
 #include "amqpcpp/libboostasio.h"
 #include "amqpcpp/linux_tcp.h"
+#include "stream.hpp"
 #include <fc/log/logger.hpp>
 #include <memory>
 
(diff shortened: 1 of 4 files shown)


programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -20,24 +20,24 @@ class rabbitmq : public stream_handler {
...
+      init(io_service, address);
 
       exchangeName_ = DEFAULT_EXCHANGE;
 
       declare_queue();
    }
 
-   rabbitmq(boost::asio::io_service& io_service, std::vector<eosio::name> routes, std::string address, std::string exchange_name, std::string exchange_type)
+   rabbitmq(boost::asio::io_service& io_service, std::vector<eosio::name>&& routes, const AMQP::Address& address,
+            std::string&& exchange_name, const std::string& exchange_type)
```
heifner, line 36: Why change `std::vector` & `std::string` to rvalue `&&` types? By value and move is more appropriate for most "sink" values.
See "Item 41: Consider pass by value for copyable parameters that are cheap to move and always copied" of Effective Modern C++.

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -72,17 +72,18 @@ class rabbitmq : public stream_handler {
...
+      } else {
+         throw std::runtime_error("Unsupported RabbitMQ exchange type: " + exchange_type);
       }
 
       auto& exchange = channel_->declareExchange(exchangeName_, type);
-      exchange.onSuccess([this]() {
-         ilog("RabbitMQ Connected Successfully!\n Exchange ${e}",
-              ("e", exchangeName_));
-      });
+      exchange.onSuccess([this]() { ilog("RabbitMQ Connected Successfully!\n Exchange ${e}", ("e", exchangeName_)); });
```
heifner, line 86: This still has the problem that `this` might be destroyed before this is called. Copy the `exchangeName_` into the lambda.

revl: That's what I did first, because I thought that was the cause of SEGFAULT, but then I saw that the lambda is contained in the channel_ data member, so I put `this` back. Because this->exchangeName_ is not in the immediate scope of the lambda, I'd have to use capture with an initializer, which is c++14. Should I do that?

revl:
It's been confirmed that this commit also fixes the Linux SEGFAULT.

systemv:
> It's been confirmed that this commit also fixes the Linux SEGFAULT.

Confirmed to be working fine on Ubuntu 20.04, which previously had segfault issue.

programs/rodeos/streams/rabbitmq.hpp
revl: I made the requested change.

PR #9220, opened 2020-06-18
Migrate fc::optional to std::optional

TimothyBanks:
TBD


libraries/fc/include/fc/io/raw.hpp
```diff
@@ -267,13 +266,13 @@ namespace fc {
 
     // optional
     template<typename Stream, typename T>
-    void pack( Stream& s, const fc::optional<T>& v ) {
-      fc::raw::pack( s, bool(!!v) );
-      if( !!v ) fc::raw::pack( s, *v );
+    void pack( Stream& s, const std::optional<T>& v ) {
+      fc::raw::pack( s, v );
```
heifner, line 270: This looks infinite recursive.

TimothyBanks: Good catch.

TimothyBanks:
Moving back to original PR.  My changes have caused a build break somehow.

PR #9222, opened 2020-06-18
Fix log of pending block producer - develop

heifner:
## Change Description

- With `transaction_tracing` debug level logging enabled `nodeos` can EOS_ASSERT from trying to log the pending block producer when no pending block is available.
- `develop` version of #9223 

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -445,7 +445,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                if (_pending_block_mode == pending_block_mode::producing) {
                   fc_dlog(_trx_trace_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} is REJECTING tx: ${txid} : ${why} ",
                         ("block_num", chain.head_block_num() + 1)
-                        ("prod", chain.pending_block_producer())
+                        ("prod", get_pending_block_producer())
                         ("txid", trx->id())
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1560,7 +1569,8 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
                if( pbm == pending_block_mode::producing ) {
                   fc_dlog( _trx_trace_log,
                            "[TRX_TRACE] Block ${block_num} for producer ${prod} is EXPIRING PERSISTED tx: ${txid}",
-                           ("block_num", chain.head_block_num() + 1)("prod", chain.pending_block_producer())("txid", txid));
+                           ("block_num", chain.head_block_num() + 1)("txid", txid)
+                           ("prod", chain.is_building_block() ? chain.pending_block_producer() : name()) );
```
brianjohnson5972, line 1573: Is there a reason you don't use the new method here as well?

heifner: impl this not available and I didn't want to have to calculate it and pass it in when most of the time logging is off and it is not needed.

PR #9235, opened 2020-06-22
Migrate fc::static_variant to std::variant

TimothyBanks:
## Change Description
This PR is to replace the use of `fc::static_variant` with `std::variant`.  Based on performance metrics, std::variant is the better alternative.

Jira:  https://blockone.atlassian.net/browse/EPE-17
Jira issue has links to benchmarks describing performance improvement of `std::variant` over `fc::static_variant`.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
Can we get a better description on this PR please. The standard template should be used. Specifically the check box sections are needed. 
Also, when we stop using JIRA again we will have no information on this change.

TimothyBanks:
This PR should be ready for review now.

I am currently unsure which boxes need to be check in the description.

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -669,7 +669,7 @@ namespace impl {
        * and can be degraded to the normal ::from_variant(...) processing
        */
       template<typename M, typename Resolver, not_require_abi_t<M> = 1>
-      static void extract( const variant& v, M& o, Resolver, abi_traverse_context& ctx )
+      static void extract( const fc::variant& v, M& o, Resolver, abi_traverse_context& ctx )
```
allenhan2, line 672: Is fc::variant kept here due to the api consistency? using std::variant will cause much changing on the user side?

libraries/fc/include/fc/static_variant.hpp
```diff
@@ -14,587 +14,86 @@
...
-      v = std::move(vars);
-   }
-   template<typename... T> void from_variant( const fc::variant& v, fc::static_variant<T...>& s )
-   {
-      auto ar = v.get_array();
-      if( ar.size() < 2 ) return;
-      s.set_which( ar[0].as_uint64() );
-      s.visit( to_static_variant(ar[1]) );
-   }
+template<typename... T> struct get_typename { static const char* name()   { return BOOST_CORE_TYPEID(std::variant<T...>).name();   } };
```
allenhan2, line 97: are they equivalent, name() for std::variant and fc::static_variant?

libraries/fc/include/fc/static_variant.hpp
```diff
@@ -14,587 +14,86 @@
...
+{
+  variants vars(2);
+  vars[0] = s.index();
+  visit( from_static_variant(vars[1]), s );
+  v = std::move(vars);
+}
+template<typename... T> void from_variant( const fc::variant& v, std::variant<T...>& s )
+{
+  auto ar = v.get_array();
+  if( ar.size() < 2 ) return;
```
allenhan2, line 92: Does s need to be void (initialized) if size() < 2, or caller needs to initialize?

libraries/fc/include/fc/static_variant.hpp
```diff
@@ -14,587 +14,86 @@
...
-   {
-      sv.init(v);
-   }
-};
-
-template<typename StaticVariant>
-struct move_construct
+  if constexpr(I >= std::variant_size_v<variant>)
+  {
+    throw std::runtime_error{"Variant index " + std::to_string(I + index) + " out of bounds"};
```
allenhan2, line 30: should be std::exception or FC::exception? and should catch all std exception here and rethrow with fc::exception?

libraries/fc/test/static_variant/test_static_variant.cpp
```diff
@@ -1,234 +0,0 @@
-#define BOOST_TEST_MODULE static_variant
```
allenhan2, line 1: see some new functions being added in static_variant.hpp, this test needs to test those new apis.

libraries/chain/transaction.cpp
```diff
@@ -600,11 +599,11 @@ const vector<bytes>* packed_transaction::get_context_free_data()const {
 const bytes* maybe_get_context_free_data(const packed_transaction::prunable_data_type::none&, std::size_t) { return nullptr; }
 const bytes* maybe_get_context_free_data(const packed_transaction::prunable_data_type::partial& p, std::size_t i) {
    if( p.context_free_segments.size() <= i ) return nullptr;
-   return p.context_free_segments[i].visit(
+     return std::visit(
```
heifner, line 602: Indentation looks wrong. It is diff throughout actually, but this one looks like it could cause confusion.

libraries/fc/include/fc/static_variant.hpp
```diff
@@ -14,587 +14,86 @@
...
-
-template< typename Visitor, typename ...Ts>
-struct const_result_type_info;
-
-template< typename Visitor, typename ...Ts>
-struct result_type_info;
-
-template< typename Visitor, typename ...Ts>
-struct rvalue_ref_result_type_info;
+// This will go away.
```
heifner, line 21: When?

libraries/fc/include/fc/rpc/json_connection.hpp
```diff
@@ -58,92 +58,92 @@ namespace fc { namespace rpc  {
          void notice( const fc::string& method, const variant_object& named_args );
```
heifner, line 58: I don't think `json_connection.hpp` is used, can you just remove it?

libraries/chain/include/eosio/chain/abi_serializer.hpp
TimothyBanks: `fc::variant` wasn't a consideration for this effort.  Bart wanted only `fc:::static_varaint` taken out.

libraries/fc/include/fc/static_variant.hpp
TimothyBanks: Which exception type is preferred?

libraries/fc/include/fc/static_variant.hpp
TimothyBanks: This is actually pre-existing functionality that was needed after removing `fc::static_variant`.  I'm not quite sure of if that is the intent or not.

libraries/fc/include/fc/static_variant.hpp
TimothyBanks: No they aren't.  Is that problematic?

libraries/fc/include/fc/static_variant.hpp
TimothyBanks: The comment itself actually just needs to go away.

libraries/chain/transaction.cpp
TimothyBanks: Do we have clang format file that I could run on these changes?

heifner: No, but we have various ones around. Maybe create a JIRA issue for someone to commit one to the repo.

libraries/fc/include/fc/static_variant.hpp
heifner: It doesn't look like we use `fc::static_variant` in chainbase objects, so I believe this is fine. You can double check by searching for "chainbase::object" and verifying none of the types are `fc::static_variant` or contain a `fc::static_variant`

libraries/fc/include/fc/static_variant.hpp
heifner: fc

libraries/fc/include/fc/static_variant.hpp
heifner: The goal is to remove `fc::static_variant`, right? If this needs to be kept then I think it should be move to `raw.hpp`.

libraries/chain/include/eosio/chain/abi_serializer.hpp
allenhan2: cool, thanks for explanation

libraries/fc/include/fc/static_variant.hpp
allenhan2: yeah, exiting static_variant throws FC exception, and the current error handling logic is based on handling FC exception, so need to keep using FC.

libraries/fc/include/fc/static_variant.hpp
allenhan2: Seems like converting to std::variant, if it is empty, can we set it as std::monostate.

libraries/fc/include/fc/static_variant.hpp
TimothyBanks: Yes that was the goal but there is some custom functionality that was needed (`get_index`, `from_index`) that `std::variant` doesn't provide. So I don't know if this file is going away unless I can move all that functionality into raw.hpp or maybe there is a better place that I don't know about?

libraries/fc/include/fc/static_variant.hpp
TimothyBanks: We can only set it to the monostate if that type is part of the variant's type set and this is done by default constructing a variant instance.  So I will default construct a variant instance and assign it to `s`.

TimothyBanks:
PR is updated.

TimothyBanks:
I'm open to suggestions as to what to call it.  Or I can just call `get`.

libraries/fc/include/fc/static_variant.hpp
```diff
@@ -14,587 +14,112 @@
...
-
-template<int Pos, typename... Ts>
-struct type_at;
+template<typename Result>
+struct visitor {};
 
-template<typename... Ts>
-struct type_info;
+template<typename T, typename... Types>
+constexpr T& get(std::variant<Types...>& v)
```
heifner, line 25: Why are you not just using `std::get` ? Add comment if there is a good reason.

TimothyBanks: @allenhan2 mentioned trying to maintain the same exception semantics.  So instead of a `std::bad_variant_access` exception being raised on an unchecked call to `std::get`, it would be better to raise the same `fc::assert_exception` exception so I created `fc::get` to mimic `std::get` but raise the fc exception.

libraries/fc/include/fc/static_variant.hpp
```diff
@@ -14,587 +14,128 @@
...
-
-template<typename Visitor, typename T, typename ... Ts>
-struct rvalue_ref_result_type_info<Visitor, T, Ts...> {
-   using result_type = typename rvalue_ref_result_type_info<Visitor,T>::result_type;
+template <typename variant, int32_t i = 0>
+void from_index(variant& v, int index) 
+{
+  if constexpr(i >= std::variant_size_v<variant>)
+  {
+    FC_THROW_EXCEPTION(fc::out_of_range_exception, "Provided index out of range for variant.");
```
heifner, line 67: Did this exception type change? Was it `fc::assert_exception` before?

libraries/chain/controller.cpp
```diff
@@ -1767,14 +1767,14 @@ struct controller_impl {
...
 
       // Create (unsigned) block:
       auto block_ptr = std::make_shared<signed_block>( pbhs.make_block_header(
-         bb._trx_mroot_or_receipt_digests.contains<checksum256_type>() ?
-               bb._trx_mroot_or_receipt_digests.get<checksum256_type>() :
-               merkle( std::move( bb._trx_mroot_or_receipt_digests.get<digests_t>() ) ),
-         merkle( std::move( pending->_block_stage.get<building_block>()._action_receipt_digests ) ),
+         std::holds_alternative<checksum256_type>(bb._trx_mroot_or_receipt_digests) ?
+               std::get<checksum256_type>(bb._trx_mroot_or_receipt_digests) :
+               merkle( std::move( std::get<digests_t>(bb._trx_mroot_or_receipt_digests) ) ),
```
allenhan2, line 1776: the second get should be fc

libraries/chain/controller.cpp
```diff
@@ -1819,10 +1819,10 @@ struct controller_impl {
       });
 
       try {
-         EOS_ASSERT( pending->_block_stage.contains<completed_block>(), block_validate_exception,
+         EOS_ASSERT( std::holds_alternative<completed_block>(pending->_block_stage), block_validate_exception,
                      "cannot call commit_block until pending block is completed" );
 
-         auto bsp = pending->_block_stage.get<completed_block>()._block_state;
+         auto bsp = std::get<completed_block>(pending->_block_stage)._block_state;
```
allenhan2, line 1825: This works fine, but better to change to fc since the pattern changes.

libraries/chain/controller.cpp
```diff
@@ -1926,7 +1926,7 @@ struct controller_impl {
          start_block( b->timestamp, b->confirmed, new_protocol_feature_activations, s, producer_block_id);
 
          // validated in create_block_state_future()
-         pending->_block_stage.get<building_block>()._trx_mroot_or_receipt_digests = b->transaction_mroot;
+         std::get<building_block>(pending->_block_stage)._trx_mroot_or_receipt_digests = b->transaction_mroot;
```
allenhan2, line 1929: same here

libraries/chain/controller.cpp
```diff
@@ -1963,10 +1963,10 @@ struct controller_impl {
          bool explicit_net = self.skip_trx_checks();
 
          size_t packed_idx = 0;
-         const auto& trx_receipts = pending->_block_stage.get<building_block>()._pending_trx_receipts;
+         const auto& trx_receipts = std::get<building_block>(pending->_block_stage)._pending_trx_receipts;
```
allenhan2, line 1966: same here

libraries/chain/controller.cpp
```diff
@@ -3093,17 +3093,17 @@ const producer_authority_schedule& controller::pending_producers()const {
...
-      const auto& new_prods_cache = my->pending->_block_stage.get<assembled_block>()._new_producer_authority_cache;
+   if( std::holds_alternative<assembled_block>(my->pending->_block_stage) ) {
+      const auto& new_prods_cache = std::get<assembled_block>(my->pending->_block_stage)._new_producer_authority_cache;
       if( new_prods_cache ) {
          return *new_prods_cache;
       }
    }
 
-   const auto& bb = my->pending->_block_stage.get<building_block>();
+   const auto& bb = std::get<building_block>(my->pending->_block_stage);
```
allenhan2, line 3106: same here

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -7,15 +7,15 @@
...
-using shared_public_key_data = fc::static_variant<fc::ecc::public_key_shim, fc::crypto::r1::public_key_shim, shared_string>;
+using shared_public_key_data = std::variant<fc::ecc::public_key_shim, fc::crypto::r1::public_key_shim, shared_string>;
 
 struct shared_public_key {
    shared_public_key( shared_public_key_data&& p ) :
       pubkey(std::move(p)) {}
 
    operator public_key_type() const {
       fc::crypto::public_key::storage_type public_key_storage;
-      pubkey.visit(overloaded {
```
allenhan2, line 18: visit has similar issue, static_variant::visit throws  FC_THROW_EXCEPTION( fc::assert_exception, "Internal error: static_variant tag is invalid.");  std:visit throws std::bad_variant_access, need similar wrap as fc::get

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -44,10 +44,10 @@ namespace eosio { namespace chain {
          fc::raw::pack( enc, status );
          fc::raw::pack( enc, cpu_usage_us );
          fc::raw::pack( enc, net_usage_words );
-         if( trx.contains<transaction_id_type>() )
-            fc::raw::pack( enc, trx.get<transaction_id_type>() );
+         if( std::holds_alternative<transaction_id_type>(trx) )
+            fc::raw::pack( enc, std::get<transaction_id_type>(trx) );
          else
-            fc::raw::pack( enc, trx.get<packed_transaction_v0>().packed_digest() );
+            fc::raw::pack( enc, std::get<packed_transaction_v0>(trx).packed_digest() );
```
allenhan2, line 50: fc::get

libraries/chain/include/eosio/chain/block.hpp
```diff
@@ -122,10 +124,10 @@ namespace eosio { namespace chain {
          fc::raw::pack( enc, status );
          fc::raw::pack( enc, cpu_usage_us );
          fc::raw::pack( enc, net_usage_words );
-         if( trx.contains<transaction_id_type>() )
-            fc::raw::pack( enc, trx.get<transaction_id_type>() );
+         if( std::holds_alternative<transaction_id_type>(trx) )
+            fc::raw::pack( enc, std::get<transaction_id_type>(trx) );
          else
-            fc::raw::pack( enc, trx.get<packed_transaction>().packed_digest() );
+            fc::raw::pack( enc, std::get<packed_transaction>(trx).packed_digest() );
```
allenhan2, line 130: fc::get

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -158,17 +158,17 @@ namespace eosio { namespace chain {
...
+   using block_signing_authority = std::variant<block_signing_authority_v0>;
 
    struct producer_authority {
       name                    producer_name;
       block_signing_authority authority;
 
       template<typename Op>
       static void for_each_key( const block_signing_authority& authority, Op&& op ) {
-         authority.visit([&op](const auto &a){
+         std::visit([&op](const auto &a){
```
allenhan2, line 169: fc::visit maybe

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -177,31 +177,31 @@ namespace eosio { namespace chain {
       }
 
       static std::pair<bool, size_t> keys_satisfy_and_relevant( const std::set<public_key_type>& keys, const block_signing_authority& authority ) {
-         return authority.visit([&keys](const auto &a){
+         return std::visit([&keys](const auto &a){
```
allenhan2, line 180: same here

TimothyBanks:
Can you please show me where this new exception type is going to cause problems in the system?  I'm adding overhead to the use of `std::variant` just to maintain throwing the same exception which seems a bit overkill.  In the case of using `std::get`, I thought it was a good compromise to invoke `std::get_if` and if that fails, then throw a new exception.  In the case of `std::visit`, I don't have a conditional equivalent and I don't like adding the overhead of another try/catch block just to repackage the exception that is bubbling out.

allenhan2:
> Can you please show me where this new exception type is going to cause problems in the system? I'm adding overhead to the use of `std::variant` just to maintain throwing the same exception which seems a bit overkill. In the case of using `std::get`, I thought it was a good compromise to invoke `std::get_if` and if that fails, then throw a new exception. In the case of `std::visit`, I don't have a conditional equivalent and I don't like adding the overhead of another try/catch block just to repackage the exception that is bubbling out.
> 
> I would much prefer to update the parts of the system that are expecting this exact fc exception message to be thrown and move them over to the std equivalent.



> Can you please show me where this new exception type is going to cause problems in the system? I'm adding overhead to the use of `std::variant` just to maintain throwing the same exception which seems a bit overkill. In the case of using `std::get`, I thought it was a good compromise to invoke `std::get_if` and if that fails, then throw a new exception. In the case of `std::visit`, I don't have a conditional equivalent and I don't like adding the overhead of another try/catch block just to repackage the exception that is bubbling out.
> 
> I would much prefer to update the parts of the system that are expecting this exact fc exception message to be thrown and move them over to the std equivalent.


Appreciate your concern on the performance impact, however, logic correctness triumphs speed before we have good test coverage. The change should not jeopardize the error handling, which seems to me noway to verify.   
For std::get, if it is not guarded with std::holds_alternative, please use fc::get, which will throw fc exception instead of std::exception.

heifner:
I think if you eval everywhere std::variant is used and make sure exception handling does introduce a change in behavior, then it would be really nice to complete remove dependency on anything in `fc`.

TimothyBanks:
A quick survey of EOS (and submodules) shows only a couple of places that we explicitly catch the `fc::assert_exception`.  Everywhere else is within tests.  Are there any other places that could check?  Or is this too ignorant of an evaluation?

TimothyBanks:
`static_variant` has been removed with the exception of a file named `static_variant`.  It currently only holds functionality that is needed by EOS that isn't available with `std::variant`.

As for tracking down all code paths, I think everyone needs to come to an agreement on what is acceptable and what the path forward should.  @larryk85 Could you please chime in on this was well.

TimothyBanks:
Could we move the exception part into another JIRA issue and push this PR in?

TimothyBanks:
Updated.

TimothyBanks:
@heifner @allenhan2 Please review again.

libraries/chain/controller.cpp
```diff
@@ -1367,18 +1392,12 @@ struct controller_impl {
...
-         trace->except_ptr = std::current_exception();
-         trace->elapsed = fc::time_point::now() - trx_context.start;
-
-         if (auto dm_logger = get_deep_mind_logger()) {
-            fc_dlog(*dm_logger, "DTRX_OP FAILED ${action_id}",
-               ("action_id", trx_context.get_action_id())
-            );
-         }
+        handle_exception(e);
+      } catch ( const std::exception& e) {
```
heifner, line 1396: Here and elsewhere. This now catches `boost::interprocess::bad_alloc` which we want to handle differently.

TimothyBanks: Good catch.  I was too focused on the std::exception/fc::exception changes.

TimothyBanks: So basically in these cases where I've added `std::exception`, handle the boost bad_alloc and std bad_alloc as well to just bubble up the exception?

libraries/chain/apply_context.cpp
```diff
@@ -131,12 +141,15 @@ void apply_context::exec_one()
...
-      trace.error_code = controller::convert_exception_to_error_code( e );
-      trace.except = e;
-      finalize_trace( trace, start );
+   } catch ( const std::bad_alloc& ) {
+      throw;
+   } catch ( const boost::interprocess::bad_alloc& ) {
       throw;
+   } catch( const fc::exception& e ) {
+      handle_exception(e);
+   } catch ( const std::exception& e ) {
```
heifner, line 150: Why was catch of std::exception added here? Do we really want all std::exceptions to be ignored? Either this is not needed OR this is a consensus change since nodes without this code will fail transaction differently.

heifner:
I think all the exception changes should be pulled out into a separate PR. They are very dangerous and deserve their own PR.

libraries/chain/apply_context.cpp
```diff
@@ -58,6 +58,16 @@ void apply_context::exec_one()
...
    const auto& cfg = control.get_global_properties().configuration;
    const account_metadata_object* receiver_account = nullptr;
+
+   auto handle_exception = [&](const auto& e)
+   {
+      action_trace& trace = trx_context.get_action_trace( action_ordinal );
+      trace.error_code = controller::convert_exception_to_error_code( e );
+      trace.except = e;
+      finalize_trace( trace, start );
+      throw;
```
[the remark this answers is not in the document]
TimothyBanks: The original code was re-throwing.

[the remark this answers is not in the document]
heifner: Ah, so it was. Sorry. The diff confused me. That is good and explains why no tests failed.

libraries/chain/apply_context.cpp
TimothyBanks: Would it be better to catch just the specific std errors being thrown by variant and optional?  I can change the `std::exception` handler to `std::bad_variant_access` and whatever specific std exceptions that are thrown.

heifner: I think what you have is likely fine. I would only argue that someone does a thorough review of possible exceptions in code paths where this was added. I suspect there are not any other std exceptions we really need to worry about. So just need someone to review and give approval.

TimothyBanks:
Replay was successful.
https://buildkite.com/EOSIO/auto-eosio-full-replay-test/builds/245#_
https://buildkite.com/EOSIO/auto-eosio-replay-test/builds/246
Merging develop now.

Please take one last look.

TimothyBanks:
@heifner @allenhan2 I think this is ready to go.  Please take another look.

PR #9251, opened 2020-06-25
Add action return value to trace_api_plugin

softprofe:
## Change Description

The new action return value and `trace_api_plugin` were added at around the same time-frame. This change includes the action return value to the `trace_api_plugin` output.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/include/eosio/trace_api/extract_util.hpp
+++ b/plugins/trace_api_plugin/include/eosio/trace_api/extract_util.hpp
@@ -5,13 +5,18 @@
 
 namespace eosio { namespace trace_api {
 
-/// Used by to_transaction_trace_v0 for creation of action_trace_v0
-inline action_trace_v0 to_action_trace_v0( const chain::action_trace& at ) {
-   action_trace_v0 r;
+
+
+/// Used by to_transaction_trace_v0 for creation of action_trace_v1
+inline action_trace_v1 to_action_trace_v1( const chain::action_trace& at ) {
+   action_trace_v1 r;
    r.receiver = at.receiver;
    r.account = at.act.account;
    r.action = at.act.name;
    r.data = at.act.data;
+      // action_trace_v1 has return value, It's the only difference with action_trace_v0 
+   r.return_value = at.return_value;
+   
    if( at.receipt ) {
       r.global_sequence = at.receipt->global_sequence;
    }
@@ -23,8 +28,8 @@ inline action_trace_v0 to_action_trace_v0( const chain::action_trace& at ) {
 }
 
 /// @return transaction_trace_v0 with populated action_trace_v0
-inline transaction_trace_v0 to_transaction_trace_v0( const chain::transaction_trace_ptr& t ) {
-   transaction_trace_v0 r;
+inline transaction_trace_v1 to_transaction_trace_v0( const chain::transaction_trace_ptr& t ) {
+   transaction_trace_v1 r;
    if( !t->failed_dtrx_trace ) {
       r.id = t->id;
    } else {
@@ -33,7 +38,7 @@ inline transaction_trace_v0 to_transaction_trace_v0( const chain::transaction_tr
    r.actions.reserve( t->action_traces.size());
    for( const auto& at : t->action_traces ) {
       if( !at.context_free ) { // not including CFA at this time
-         r.actions.emplace_back( to_action_trace_v0( at ));
+         r.actions.emplace_back( to_action_trace_v1( at ));
       }
    }
    return r;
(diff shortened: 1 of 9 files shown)


heifner:
Can we get a better description on this PR please. The standard template should be used. Specifically the check box sections are needed.

plugins/trace_api_plugin/include/eosio/trace_api/extract_util.hpp
```diff
@@ -5,13 +5,18 @@
 
 namespace eosio { namespace trace_api {
 
-/// Used by to_transaction_trace_v0 for creation of action_trace_v0
-inline action_trace_v0 to_action_trace_v0( const chain::action_trace& at ) {
-   action_trace_v0 r;
+
+
+/// Used by to_transaction_trace_v0 for creation of action_trace_v1
```
heifner, line 10: Update comment

plugins/trace_api_plugin/include/eosio/trace_api/extract_util.hpp
```diff
@@ -23,8 +28,8 @@ inline action_trace_v0 to_action_trace_v0( const chain::action_trace& at ) {
 }
 
 /// @return transaction_trace_v0 with populated action_trace_v0
```
heifner, line 30: update comment

plugins/trace_api_plugin/include/eosio/trace_api/trace.hpp
```diff
@@ -21,11 +21,21 @@ namespace eosio { namespace trace_api {
...
+      chain::bytes                        return_value = {};   // add return_value 
+   };
+
   struct transaction_trace_v0 {
       using status_type = chain::transaction_receipt_header::status_enum;
 
       chain::transaction_id_type    id = {};
-      std::vector<action_trace_v0>  actions = {};
+      std::vector<action_trace_v1>  actions = {};
    };
```
heifner, line 39: Did you test reading an old `trace_api_plugin` file created before this change? I don't think it will work as the `transaction_trace_v0` format is changed here. It would be good to add a test where we read old supported versions of `trace_api_plugin`.

plugins/trace_api_plugin/request_handler.cpp
```diff
@@ -48,7 +48,8 @@ namespace {
                ("account", a.account.to_string())
                ("action", a.action.to_string())
                ("authorization", process_authorizations(a.authorization, yield))
-               ("data", fc::to_hex(a.data.data(), a.data.size()));
+               ("data", fc::to_hex(a.data.data(), a.data.size()))
+               ("return_value", fc::to_hex(a.return_value.data(), a.return_value.size()));    // add return_value if input is action_trace_v1
```
heifner, line 52: Similar to "data" -> "params", you should attempt to deserialize the return value and provide it deserialized.
Maybe "return_value_data" and "return_value" for the json names.

plugins/trace_api_plugin/request_handler.cpp
```diff
@@ -77,7 +78,7 @@ namespace {
       return result;
    }
 
-    fc::variants process_transactions(const std::vector<transaction_trace_v1>& transactions, const data_handler_function& data_handler, const yield_function& yield ) {
+    fc::variants process_transactions(const std::vector<transaction_trace_v1>& transactions, const data_handler_function & data_handler, const yield_function& yield ) {
```
heifner, line 81: Extra space added

plugins/trace_api_plugin/include/eosio/trace_api/trace.hpp
softprofe: As this plugin just commit for few months, there shouldn't be many old trace. so, This is a quick way to support new structure member return_value.  If we want have function of reading old trace and new trace, We have to define trace_block_v3. and then if any member added in future, we have to define trace_block_v4, and v5, v6 ..., and the process function would be more and more, do you think that is a better way to support return_value?      

Usually a version field is used to read different version, so as to deal different version in one function, but there is no version filed in action_trace and transaction_trace and also block_track, if you want support different version I can add a version field and then I think too much code will be changed.

heifner: The plugin has already been released, so I would think we would want to support old file formats. That is a question for @b1bart

softprofe: That's OK, I am going to define a new block structure name block_tarce_v2, this way we touch the old code. I will give a new pull later.

softprofe: type wrong, I mean this way we don't touch the old code. just add new things.

emorybarlow:
This is a duplicate of https://github.com/EOSIO/eos/pull/9261

PR #9252, opened 2020-06-25
Transaction logging rel 2.0.x

huangminghuang:
## Change Description
This PR splits the transaction_tracing logger into transaction_success_tracing and transaction_failure_tracing loggers; in addition, it adds logging statement for failed signature condition.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions
 `transaction_tracing` logger is removed

The following two loggers are added
 -  `transaction_success_tracing` - detailed log that emits successful verdicts from relay nodes on the P2P network.
 -  `transaction_failure_tracing` - detailed log that emits failed verdicts from relay nodes on the P2P network.

### DIFF
--- a/docs/01_nodeos/06_logging/index.md
+++ b/docs/01_nodeos/06_logging/index.md
@@ -80,7 +80,8 @@ The logging library built into EOSIO currently supports the following loggers:
 - `net_plugin_impl` - detailed logging for the net plugin.
 - `http_plugin` - detailed logging for the http plugin.
 - `producer_plugin` - detailed logging for the producer plugin.
-- `transaction_tracing` - detailed log that emits verdicts from relay nodes on the P2P network.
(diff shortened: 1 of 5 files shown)


heifner:
Do we want to replace existing logging config in `release/2.0.x`. My opinion is that we have changed much greater things in `release/2.0.x`, but this needs to be flagged and well documented.

heifner:
Need to update `programs/nodeos/logging.json`. Also in `develop`.

brianjohnson5972:
It was decided with @b1bart that we should keep the existing transaction_trace logging identical to the user and just added transaction_failure_trace to allow user to just get failure logging.

PR #9281, opened 2020-07-06
Refactor to use std::unique_ptr instead of naked pointers 

nksanthosh:
## Change Description

Since delete just marks the memory as "being free for reuse" and lhost does not immediately go out of scope and is used in several places it is safer to use std::unique_ptr instead of naked pointers and the corresponding delete
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/eosio-launcher/main.cpp
+++ b/programs/eosio-launcher/main.cpp
@@ -799,7 +799,7 @@ launcher_def::define_network () {
   }
   else {
     int ph_count = 0;
-    host_def *lhost = nullptr;
+    std::unique_ptr<host_def> lhost{nullptr};
     size_t host_ndx = 0;
     size_t num_prod_addr = servers.producer.size();
     size_t num_nonprod_addr = servers.nonprod.size();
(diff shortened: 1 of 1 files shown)


programs/eosio-launcher/main.cpp
```diff
@@ -809,6 +809,7 @@ launcher_def::define_network () {
         if (lhost) {
           bindings.emplace_back(move(*lhost));
           delete lhost;
+          lhost = nullptr;
```
heifner, line 812: I think it would be better to change this to use a `std::optional`

programs/eosio-launcher/main.cpp
```diff
@@ -849,6 +850,7 @@ launcher_def::define_network () {
     } // for i
     bindings.emplace_back( move(*lhost) );
     delete lhost;
+    lhost = nullptr;
```
brianjohnson5972, line 853: Since this is a local variable, I don't think this is needed, but if we make this change, I think for maintainability we should instead make lhost a unique_ptr and change line 810 and 851 to be "bindings.push_back(*lhost);" so we don't have to keep track of memory. This use case will not loop through this enough times for the moving of memory to make a significant impact.

programs/eosio-launcher/main.cpp
nksanthosh: Decided to make it a std::unique_ptr.

programs/eosio-launcher/main.cpp
nksanthosh: Brian - please take a look at the revised implementation. Thanks.

PR #9282, opened 2020-07-06
use nullptr following delete

nksanthosh:
## Change Description

Since delete just marks the memory as "being free for reuse" and lhost does not immediately go out of scope and is used in several places it is safer to set it to nullptr following the delete

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


larryk85:
Removing the release/2.1.x and renaming to stale-release/2.1.x for posterity.

PR #9284, opened 2020-07-06
Refactor to use std::unique_ptr instead of naked pointers

nksanthosh:
## Change Description

Since delete just marks the memory as "being free for reuse" and lhost does not immediately go out of scope and is used in several places it is safer to use std::unique_ptr instead of naked pointers and the corresponding delete

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/eosio-launcher/main.cpp
+++ b/programs/eosio-launcher/main.cpp
@@ -816,6 +816,7 @@ launcher_def::define_network () {
         if (lhost) {
           bindings.emplace_back(move(*lhost));
           delete lhost;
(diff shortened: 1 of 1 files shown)


nksanthosh:
1.8.x will only take critical bug fixes - so closing this PR

PR #9286, opened 2020-07-06
fix the issue caused by \r in cleos warnings and infos

vzqzhang:
## Change Description
The character '\r' causes issues for the string concatenation. The characters before '\r' are overwritten. This change removes all '\r' in the message before the it is appended.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/fc/src/log/console_appender.cpp
+++ b/libraries/fc/src/log/console_appender.cpp
@@ -1,4 +1,4 @@
-%s/^M$//#include <fc/log/console_appender.hpp>
+#include <fc/log/console_appender.hpp>
 #include <fc/log/log_message.hpp>
 #include <fc/string.hpp>
 #include <fc/variant.hpp>
@@ -121,7 +121,7 @@ namespace fc {
                line += "<7>";
                break;
          }
-      }^M
+      }
 
       line += fixed_size(  5, context.get_log_level().to_string() ); line += ' ';
       // use now() instead of context.get_timestamp() because log_message construction can include user provided long running calls
--- a/libraries/fc/src/log/console_appender.cpp
+++ b/libraries/fc/src/log/console_appender.cpp
@@ -1,4 +1,4 @@
-#include <fc/log/console_appender.hpp>
+%s/^M$//#include <fc/log/console_appender.hpp>
 #include <fc/log/log_message.hpp>
 #include <fc/string.hpp>
 #include <fc/variant.hpp>
@@ -105,6 +105,7 @@ namespace fc {
 
       std::string line;
       line.reserve( 256 );
+      line += "[ ";
       if(my->use_syslog_header) {
          switch(m.get_context().get_log_level()) {
             case log_level::error:
@@ -120,7 +121,8 @@ namespace fc {
                line += "<7>";
                break;
          }
-      }
+      }^M
+
       line += fixed_size(  5, context.get_log_level().to_string() ); line += ' ';
       // use now() instead of context.get_timestamp() because log_message construction can include user provided long running calls
       line += string( time_point::now() ); line += ' ';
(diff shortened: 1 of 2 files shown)


vzqzhang:
https://blockone.atlassian.net/browse/EPE-254

programs/cleos/main.cpp
```diff
@@ -584,7 +584,6 @@ void print_result( const fc::variant& result ) { try {
             for( const auto& a : actions ) {
                print_action_tree( a );
             }
-            wlog( "\rwarning: transaction executed locally, but may not be confirmed by the network yet" );
```
heifner, line 587: I think you misunderstood. The idea was not to remove this message, but to find the memory overrun that was causing its output to be garbled.

heifner:
Please use the standard PR template.

heifner:
Please provide example before/after log output for this PR including an example from nodeos and cleos.

heifner:
I think this needs to be tested/worked after https://github.com/EOSIO/eos/pull/9172 has been ported to `develop` branch. https://blockone.atlassian.net/browse/EPE-143

vzqzhang:
Here are the log output before and after the change:

before the change:
image

after the change:
image

libraries/fc/src/log/console_appender.cpp
```diff
@@ -105,6 +105,7 @@ namespace fc {
 
       std::string line;
       line.reserve( 256 );
+      line += "[ ";
```
heifner, line 108: I'm not sure that we want this, it changes the existing output format.

vzqzhang: The reason I added it is because there is an unmatched "]" at the end of the line. If this output is verified in some tests, we should remove it. However, I did not see any failure from buildkite. I can remove it if you think it is risky for some automation runs.

jeffreyssmith2nd: It is probably best to stay consistent with what was previously there, in the case that logs are being parsed and used.

vzqzhang: Reverted that line.

spoonincode:
>after the change:
 86954538-ff94d880-c10a-11ea-9ae5-151b4cf1136c

That looks a little weird IMO. Have we considered just changing
```c++
wlog( "\rwarning: transaction executed locally, but may not be confirmed by the network yet" );
```
To something like
```c++
wlog( "            \rwarning: transaction executed locally, but may not be confirmed by the network yet" );
```

heifner:
> ```c++
> wlog( "\rwarning: transaction executed locally, but may not be confirmed by the network yet                          " );
> ```

Probably best to just avoid using `wlog` if we are trying to abuse it not to report the "header" info. We could always just use `std::cout/std::cerr` like the rest of `cleos` adding color coding if we think that is important.

vzqzhang:
> Probably best to just avoid using wlog if we are trying to abuse it not to report the "header" info. We could always just use std::cout/std::cerr like the rest of cleos adding color coding if we think that is important.

I wonder if wlog dumps messages to files, not the console, in some cases.

vzqzhang:
Screen Shot 2020-07-15 at 11 27 08 AM

Does this look good?

heifner:
> Screen Shot 2020-07-15 at 11 27 08 AM
> 
> Does this look good?

Looks fine to me. But that is really a question for @b1bart or @aclark-b1

jeffreyssmith2nd:
> I think this needs to be tested/worked after #9172 has been ported to `develop` branch. [blockone.atlassian.net/browse/EPE-143](https://blockone.atlassian.net/browse/EPE-143)

This work has been done, so @vzqzhang you're good to reevaluate how this PR fits into the new code.

vzqzhang:
> > I think this needs to be tested/worked after #9172 has been ported to `develop` branch. [blockone.atlassian.net/browse/EPE-143](https://blockone.atlassian.net/browse/EPE-143)
> 
> This work has been done, so @vzqzhang you're good to reevaluate how this PR fits into the new code.

After the merge, the extra characters are gone. I wonder if the file console_appender.cpp is used by other files. The issue is fixed for command line of cleos. However, if console_appender.cpp is used by other files and '\r' is in the string, the issue could happen in other files.

Otherwise, we can discard this PR.

vzqzhang:
Close it since the changes in the merge already fixed the issue.

PR #9298, opened 2020-07-09
Add payment support for kv storage

huangminghuang:
## Change Description

Allow application developers to choose who pays for resources (feature parity with multi-index table functionality).

EPE-218

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/kv_context.cpp
```diff
@@ -6,22 +6,22 @@
...
    namespace {
-      void kv_resource_manager_update_ram(apply_context& context, int64_t delta, const kv_resource_trace& trace) {
+      void kv_resource_manager_update_ram(apply_context& context, int64_t delta, const kv_resource_trace& trace, account_name payer) {
          std::string event_id;
          if (context.control.get_deep_mind_logger() != nullptr) {
             event_id = STORAGE_EVENT_ID("${id}", ("id", fc::to_hex(trace.key.data(), trace.key.size())));
          }
 
-         context.add_ram_usage(account_name(context.get_receiver()), delta, storage_usage_trace(context.get_action_id(), event_id.c_str(), "kv", trace.op_to_string()));
+         context.add_ram_usage(payer, delta, storage_usage_trace(context.get_action_id(), event_id.c_str(), "kv", trace.op_to_string()));
```
swatanabe-b1, line 15: Needs to use `update_db_usage` instead of `add_ram_usage`.  `add_ram_usage` doesn't validate that the payer authorized the action.

libraries/chain/kv_context.cpp
```diff
@@ -6,22 +6,22 @@
...
 
-      void kv_resource_manager_update_disk(apply_context& context, int64_t delta, const kv_resource_trace& trace) {
+      void kv_resource_manager_update_disk(apply_context& context, int64_t delta, const kv_resource_trace& trace, account_name payer) {
          std::string event_id;
          if (context.control.get_deep_mind_logger() != nullptr) {
             event_id = STORAGE_EVENT_ID("${id}", ("id", fc::to_hex(trace.key.data(), trace.key.size())));
          }
 
-         context.add_disk_usage(account_name(context.get_receiver()), delta, storage_usage_trace(context.get_action_id(), event_id.c_str(), "kv", trace.op_to_string()));
+         context.add_disk_usage(payer, delta, storage_usage_trace(context.get_action_id(), event_id.c_str(), "kv", trace.op_to_string()));
```
swatanabe-b1, line 24: Same as with `update_ram`.  Also, need to forbid increasing disk usage in a notify.

libraries/chain/apply_context.cpp
```diff
@@ -1229,7 +1229,14 @@ void apply_context::add_ram_usage( account_name account, int64_t ram_delta, cons
    }
 }
 
-void apply_context::add_disk_usage( account_name account, int64_t disk_delta, const storage_usage_trace& trace ) {
+void apply_context::update_disk_usage( account_name account, int64_t disk_delta, const storage_usage_trace& trace ) {
+
+   if (disk_delta > 0 && !privileged) {
```
swatanabe-b1, line 1234: It's also acceptable if `account == receiver`

libraries/chain/apply_context.cpp
```diff
@@ -1229,7 +1229,14 @@ void apply_context::add_ram_usage( account_name account, int64_t ram_delta, cons
...
+      EOS_ASSERT(receiver == act->account, subjective_block_production_exception,
```
swatanabe-b1, line 1235: This should not be a subjective failure.  The similar check in `update_db_usage` is a subjective mitigation for before `RAM_RESTRICTIONS` is activated.  Since DISK in new, we don't need subjective mitigations and can always apply the restriction.  The actual check that you should copy is in `exec_one`.

swatanabe-b1: The behaviour of the check in `exec_one` is slightly different from the subjective check in `update_db_usage`.  Unless we've decided that that was a mistake it would be better to change `exec_one`.

libraries/chain/apply_context.cpp
```diff
@@ -50,6 +50,32 @@ apply_context::apply_context(controller& con, transaction_context& trx_ctx, uint
    context_free = trace.context_free;
 }
 
+template <typename Excpetion>
```
jeffreyssmith2nd, line 53: `Exception`

libraries/chain/apply_context.cpp
```diff
@@ -50,6 +50,32 @@ apply_context::apply_context(controller& con, transaction_context& trx_ctx, uint
...
+   const size_t checktime_interval    = 10;
+   const bool   not_in_notify_context = (receiver == act->account);
+   size_t counter = 0;
+   for (const auto& entry : deltas) {
+      if (counter == checktime_interval) {
+         trx_context.checktime();
+         counter = 0;
+      }
+      if (entry.delta > 0 && entry.account != receiver) {
+         EOS_ASSERT(not_in_notify_context, Excpetion,
```
jeffreyssmith2nd, line 64: type

libraries/chain/apply_context.cpp
```diff
@@ -50,6 +50,32 @@ apply_context::apply_context(controller& con, transaction_context& trx_ctx, uint
...
+                     "unprivileged contract cannot increase ${resource} usage of another account within a notify context: "
+                     "${account}",
+                     ("resource", resource)
+                     ("account", entry.account));
+         EOS_ASSERT(has_authorization(entry.account), Excpetion,
```
jeffreyssmith2nd, line 69: typo

libraries/chain/apply_context.cpp
```diff
@@ -95,27 +121,12 @@ void apply_context::exec_one()
...
-                                 "unprivileged contract cannot increase RAM usage of another account that has not authorized the action: ${account}",
-                                 ("account", itr->account)
-                     );
-                  }
+            if (!privileged) {
+               if (control.is_builtin_activated(builtin_protocol_feature_t::ram_restrictions)) {
+                  check_unprivileged_resource_usage<unauthorized_ram_usage_increase>("RAM", _account_ram_deltas);
                }
+
+               check_unprivileged_resource_usage<unauthorized_disk_usage_increase>("DISK", _account_disk_deltas);
```
jeffreyssmith2nd, line 129: I believe everything "DISK" related is not supposed to be a part of this release, so I think this param can be removed.

libraries/chain/apply_context.cpp
```diff
@@ -1229,10 +1240,11 @@ void apply_context::add_ram_usage( account_name account, int64_t ram_delta, cons
    }
 }
 
-void apply_context::add_disk_usage( account_name account, int64_t disk_delta, const storage_usage_trace& trace ) {
-   trx_context.add_disk_usage( account, disk_delta, trace );
+void apply_context::add_disk_usage( account_name payer, int64_t disk_delta, const storage_usage_trace& trace ) {
```
jeffreyssmith2nd, line 1243: Same thing here, I'm not sure if this should be included.

@b1bart, thoughts on this?

libraries/chain/apply_context.cpp
huangminghuang: > I believe everything "DISK" related is not supposed to be a part of this release, so I think this param can be removed.

The disk charging part and its corresponding unit tests existed before this PR.  This PR only refactor the code so that the permission checking mechanism for DISK is the same with that for RAM. Unless we don't want to charge for DISK at all in the future. I don't see why we need to remove the logic.

PR #9307, opened 2020-07-13
rodeos filters: print_time_us intrinsic

tbfleming:
## Change Description

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/rodeos/include/b1/rodeos/callbacks/filter.hpp
```diff
@@ -16,10 +16,28 @@ struct filter_callbacks {
       derived().get_filter_callback_state().push_data(data.data(), data.size());
    }
 
+   int32_t clock_gettime(int32_t id, uint64_t* data) {
```
swatanabe-b1, line 19: This should have a different name, so that a libc wrapper can set errno.  (The eosio-tester intrinsic has the same problem).

libraries/rodeos/include/b1/rodeos/callbacks/filter.hpp
```diff
@@ -16,10 +16,28 @@ struct filter_callbacks {
...
+      std::chrono::nanoseconds result;
+      if (id == 0) { // CLOCK_REALTIME
+         result = std::chrono::system_clock::now().time_since_epoch();
```
swatanabe-b1, line 22: I thought we were avoiding non-determinism in rodeos.

swatanabe-b1: Also, libc `time` uses `current_time`.  The divergence between which functions return block time and which functions return the time at which the filter is running is likely to be confusing.

libraries/rodeos/include/b1/rodeos/callbacks/filter.hpp
tbfleming: That'd force me to make changes to CDT

tbfleming: Which would break eosio-tester

libraries/rodeos/include/b1/rodeos/callbacks/filter.hpp
tbfleming: This PR is to support running performance tests on portions of filter code.

tbfleming: We goofed when we named the current block time `current_time` instead of `current_block_time`. We goofed again by hooking libc `time` to it. It'd be a shame to lose `std::chrono` for perf testing now because of our blunder back then.

libraries/rodeos/include/b1/rodeos/callbacks/filter.hpp
swatanabe-b1: We will need to fix this eventually, but I can live with it for now.

libraries/rodeos/include/b1/rodeos/callbacks/filter.hpp
swatanabe-b1: I don't really think that those were mistakes.  They make perfect sense from the POV of a contract, where block time is the only time.

tbfleming:
> auditability through (in part) determinism is one of the primary value propositions of EOSIO

Yes. At some point we need to decide (this question was up in the air last I heard, even at the top org levels) whether we should force determinism at all points in the system, or allow opt-in (e.g. by using this intrinsic) non-determinism where it doesn't harm block consensus.

> I would expect such a release to only enable sources of non-determinism like this when in an explicit testing mode which does its best to disable production use.

A certain customer requested detailed timing information from within filters go into production use. I'm open to trying to talk them out of it, but need help making that case.

> constructing and packing that buffer

If the compiler isn't optimizing `fc::raw::pack` well for this case, then we have a potential major perf gain throughout nodeos by fixing that function.

> Was their consideration of a profiling/tracing addition to the eos-vm layer rather than hooks for the application layer?

No, but it'd be worth considering. My initial guess: it'd be more expensive than the intrinsic call.

tbfleming:
I'm not apposed to gating it behind a flag, but this will trigger a nasty existing usability issue: eos-vm doesn't indicate which intrinsics are missing when it fails to load a wasm.

tbfleming:
Scratch the usability issue. The flag could cause the intrinsic to throw, not remove the intrinsic.

tbfleming:
New proposal: add a new intrinsic which tags a log message with a high-resolution time.

* Keeps the existing state WRT nondeterminism in filters: CFD presence / non-presence, and NaN
* Keeps this from ballooning into a major addition to eos-vm (profiling/tracing)
* Punts on the standard library block-time issue for now
* Punts on the stderr issue with the existing eosio-tester time for now
* Punts on quantum behavior for now (measurement affects results). This will get worse with this newer approach.

PR #9314, opened 2020-07-15
Prevent an older version of g++ to build eosio

tony-b1:
## Change Description
Updating the CMakeLists.txt to prevent older g++ versions to build eosio.

## Change Type
**Select ONE**
- [ ] Documentation
- [ ] Stability bug fix
- [X] Other
- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
fwiw I tried building 2.0.0 with gcc 7.5 and it failed (not that it should be a surprise since the troublesome file has not been changed in 2.0.x). So LGTM -- I don't think anyone out there would likely be using a 2.0.x with gcc 7

PR #9315, opened 2020-07-15
rodeos: retry on bind failure

tbfleming:
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

Sometimes when rodeos starts up, the OS fails the socket bind saying the port is already in use. The OS then allows it after some unknown amount of time. We've seen this behavior on OSX after killing (gracefully) the process then restarting it. We've also seen this behavior when starting containers on k8s.

rodeos now automatically retries the bind, waiting ~1s between each try. A new option, `--wql-retries` limits the number of retries (0 = no retrying, default = no limit). rodeos shuts down if it hits the retry limit.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/rodeos/wasm_ql_plugin.cpp
+++ b/programs/rodeos/wasm_ql_plugin.cpp
@@ -44,12 +44,12 @@ struct wasm_ql_plugin_impl : std::enable_shared_from_this<wasm_ql_plugin_impl> {
                FC_LOG_AND_RETHROW()
             } catch (...) {
                elog("shutting down");
-               app().shutdown();
+               app().quit();
             }
          });
       } else {
          elog("hit --wql-retries limit; shutting down");
-         app().shutdown();
+         app().quit();
       }
    }
 
--- a/programs/rodeos/wasm_ql_plugin.cpp
+++ b/programs/rodeos/wasm_ql_plugin.cpp
@@ -17,14 +17,45 @@ namespace b1 {
 
 struct wasm_ql_plugin_impl : std::enable_shared_from_this<wasm_ql_plugin_impl> {
    bool                                         stopping     = false;
+   uint32_t                                     max_retries  = 0xffff'ffff;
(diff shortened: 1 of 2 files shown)


spoonincode:
Did you try `SO_REUSEADDR`?

```c++
boost::asio::socket_base::reuse_address option(true);
acceptor.set_option(option);
```

tbfleming:
Ick. I confused that option with `SO_REUSEPORT`; I didn't want to open a vulnerability with that.

tbfleming:
Those options are poorly named.

programs/rodeos/wasm_ql_plugin.cpp
```diff
@@ -17,14 +17,45 @@ namespace b1 {
...
+         timer.async_wait([this](auto&) {
+            ilog("retry...");
+            try {
+               try {
+                  start_http();
+               }
+               FC_LOG_AND_RETHROW()
+            } catch (...) {
+               elog("shutting down");
+               app().shutdown();
```
swatanabe-b1, line 47: Directly calling `app().shutdown()` is dangerous.  It's better to call `app().quit()` and let appbase break out of the event loop normally.

heifner: We really should make that method private.

tbfleming: Yep. And switch http_plugin to using quit.

heifner: I saw that. I'll create a PR to develop for it.

spoonincode:
Do you really need to keep the loop around still?

tbfleming:
Probably not. The round-trip-time for customer testing is so high that I prefer to leave it in just in case...

PR #9317, opened 2020-07-16
fix UB with rvalue reference

dimas1185:
## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
@@ -189,7 +189,7 @@ namespace eosio {
       if ((left == right) && (body[left] == ' ')) {
          return {};
       }
-      return {&body[left], right-left+1};
+      return std::string_view(body).substr(left, right-left+1);
    }
 
    inline bool is_empty_content(const std::string& body) {
--- a/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
@@ -159,6 +159,10 @@ namespace eosio {
       error_info error;
    };
 
+   /**
+    * @brief Used to trim whitespace from body. 
+    * Returned string_view valid only for lifetime of body
+    */
    inline std::string_view make_trimmed_string_view(const std::string& body) {
       if (body.empty()) {
          return {};
--- a/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
@@ -185,7 +185,7 @@ namespace eosio {
       if ((left == right) && (body[left] == ' ')) {
          return {};
       }
-      return {body.substr(left, right-left+1)};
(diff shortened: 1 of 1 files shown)


plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
```diff
@@ -185,7 +185,7 @@ namespace eosio {
       if ((left == right) && (body[left] == ' ')) {
          return {};
       }
-      return {body.substr(left, right-left+1)};
+      return {&body[left], right-left+1};
```
heifner, line 188: This is still rather fragile in my opinion. I think you should just remove this method and replace its one use with direct implementation.

dimas1185: @heifner not sure I follow. do you mean to remove `make_trimmed_string_view` and paste its implementation in its only usage? 
The functionality looks deserving separate function. trimming strings is something people use though we don't have it in stl.
Alternative we can use more standard stuff like `std::string::find_first_not_of`/`std::string::find_last_not_of` but I'm not sure if custom implementation was added for a performance reason.

@allenhan2 any ideas?

allenhan2: It is better to use  "return std::string_view(body).substr(left, right-left+1);"

heifner: > It is better to use "return std::string_view(body).substr(left, right-left+1);"

This is not better as `body` is potentially only valid for the function call.

heifner: @dimas1185 What I mean is reimplement `is_empty_content` without `make_trimmed_string_view` and remove `make_trimmed_string_view`.

allenhan2: This is the nature of string_view, should not assume that it will be valid when underline string is not valid, otherwise, should use boost::algorithm::trim.  In this case, body string is not changed.
  
the function have 2 purpose, one is NO dynamic allocation, the other is NO change for the original string, body.

heifner: @allenhan2 understood and as such I think it is too fragile of a function hence my recommendation to just remove it.

allenhan2: The code and test case readability always bothers me, prefer short function, simple logic, and meaningful function name. it is easy for review and for other people to take over. Also, easy to write unit test cases, since part of conditions have been covered already. The drawback is end up like java, which is over engineered. 
For this function, I will vote for a separated function, since it could be reused for other scenarios when a trimed_string_view is needed without allocation/copy/change.
hope to move to a better location (util) instead of being put in http_plugin.hpp.

PR #9336, opened 2020-07-23
rodeos reliable amqp connection - develop

heifner:
## Change Description

Pulling over AMQP work now used in rodeos to develop.

===

As EOSIO gains more support for AMQP it gains the need in some use cases to reliably publish messages to AMQP exchanges. reliable_amqp_publisher is a small class for publishing messages to an exchange that:

* Reconnects to the AMQP server on connection failure
* Reconnects to the AMQP server on channel failure (such as when publishing to a non-existing exchange)
* Publishes messages in an AMQP transaction for positive confirmation the broker received them
* Queues unconfirmed messages while not connected
* Saves unconfirmed messages on disk on exit, and restores on relaunch

reliable_amqp_queue expects that the exchange it is configured to publish to will exist. It will not attempt to create it.

I am expecting and fully accept grief due to the lack of integration tests. I would like to explore with the team how we can best add integration tests for AMQP.

===

===

- Add `reliable_amqp_publisher` to rodeos streamer to provide reliable publishing of messages.
- Added ability to make the reliable publishing by block optional so that messages are published immediately as the filter is processed.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

New options:
Command line option: `--stream-delete-unsent`
  "Delete unsent AMQP stream data retained from previous connections"

Config option: `stream-rabbits-immediately=true`
  "Stream to RabbitMQ immediately instead of batching per block. Disables reliable message delivery."


libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+      in_flight = msgs_this_transaction;
+
+      channel->commitTransaction().onSuccess([this](){
+         message_deque.erase(message_deque.begin(), message_deque.begin()+in_flight);
+         logged_exceeded_max_depth = false;
+      })
+      .onFinalize([this]() {
+         in_flight = 0;
+         //unfortuately we don't know if an error is due to something recoverable or if an error is due
```
kimjh2005: #1. If commitTransaction() fails, we may need to log it.  We can get useful information such as how often it happens, the message_id, the number of messages in the envelop, the size of envelop, etc.

#2. Can the size of AMQP::Envelop be too big for rabbitmq if the the number or message with the same message_id is too many and too big?

libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
```diff
@@ -0,0 +1,236 @@
...
+      //if a message is in flight, let it try to run to completion
+      if(in_flight) {
+         auto until = std::chrono::steady_clock::now() + std::chrono::seconds(5);
+         ctx.restart();
+         while(in_flight && ctx.run_one_until(until));
+      }
+
+      if(message_deque.size()) {
+         try {
+            boost::filesystem::create_directories(data_file_path.parent_path());
```
kimjh2005: It needs to capture exception or handle error code for create_directories()

spoonincode: It's inside an FC_LOG_AND_DROP(), not really sure what else can be done here on failure since we're shutting down.

libraries/reliable_amqp_publisher/reliable_amqp_publisher.cpp
spoonincode: 2) The default message size in rabbitmq is 128MB; rabbitmq's max configurable is 512MB; amqp's max is 2GB. These limits were no where close to what I was expecting this to be used for originally. @heifner is rodeos usage a lot of data?

spoonincode: 3) I think any error is going to bubble up as a channel error or connection error, both of which have a 1s retry timer on them. You can see this by doing actions like publishing to a non-exsisting exchange, or publishing to an exchange that is bound to a queue that is full (and rejecting publishing due to that). Both of those result in a channel error and a 1s retry before attempting again.

heifner: 2. No way they will be anywhere near that size.

libraries/amqp/reliable_amqp_publisher.cpp
```diff
@@ -0,0 +1,272 @@
...
+         fc::raw::unpack(file, message_deque);
+         if( !message_deque.empty() )
+            batch_num = message_deque.back().num;
+      } FC_RETHROW_EXCEPTIONS(error, "Failed to load previously unconfirmed AMQP messages from ${f}", ("f", (fc::path)data_file_path));
+   }
+   else {
+      boost::filesystem::ofstream o(data_file_path);
+      FC_ASSERT(o.good(), "Failed to create unconfirmed AMQP message file at ${f}", ("f", (fc::path)data_file_path));
+   }
+   boost::filesystem::remove(data_file_path, ec);
```
kimjh2005, line 88: If preserving unsent messages is important, Instead of removing file here can  it be renamed as a backup file or deleted later?

heifner: I believe our current plan is to always persist to disk as we go along to support hard-failure cases like `kill -9`

kimjh2005: I just think of a case that the file has some unconfirmed messages because of the previous problem.  And nodeos restarts and reads the file and crashed or killed by -9  before writing to a data file.  Since we deleted the data file already, the next rerun will not have a valid data file.

spoonincode: (the reason it works this way) was IMO if nodeos crashes I feel like you're going to have bigger problems. You're going to have to start anew from a snapshot or reply/resync _anyways_. So you can pick a block to restart from that covers the area affected.

But as Kevin explains, there is a desire to harden usage of file further.

heifner:
Way out of date.

PR #9341, opened 2020-07-27
Allow HTTP-RPC with empty response

conr2d:
## Change Description

Current http request handlers always return the json string converted from fc::variant, it is not allowed to make HTTP-RPC with empty response. (At least, it will have "null" in the case of `fc::variant()`)

I am writing jsonrpc_plugin working upon http_plugin, and when request is given without "id" (it's called notification), JSONRPC server must not reply to it. This patch makes it possible to create API with empty response.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -168,6 +168,21 @@ namespace eosio {
          } catch(...) {}
          return 0;
       }
+
+      /**
+       * Helper method to calculate the "in flight" size of a std::optional<T>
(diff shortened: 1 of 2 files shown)


heifner:
Replaced by #9633 
Thanks @conr2d

PR #9347, opened 2020-07-27
Update abieos submodule to point to eosio branch

emorybarlow:
## Change Description

Update `abieos` submodule to point to `eosio` branch

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


emorybarlow:
> I don't know why I could not hover over lines where I could add a comment.
> Here are my two general comments:
> 
> 1. I am a little concern that "#ifndef ABIEOS_NO_INT128" scatters around files. Do we want to keep this as an option or just keep it on?
> 2. Do we have test cases for ABIEOS_NO_INT128 being both turned on and off?

I did not make those changes so I’ll defer comment to @tbfleming . This PR is just to update the submodule.

tbfleming:
It varies by architecture:
* x86-64 compilers support 128-bit
* wasm compilers support 128-bit
* iOS and Android compilers don't

The default supports 128-bit.

tbfleming:
The tests build in the same mode as the library. The only real way to test iOS and Android compatibility is to build abieos for those platforms and integrate into an iOS or Android test bench (probably an emulator).

Issue #9335, opened 2020-07-23
verify producer name is an eosio name


allenhan2:
when starting a BP and producer name is not presented after '-p' in the command line, the next word will be read as producer name, such as "-p --data-dir=xxx". and, nodeos starts with  default data-dir instead of given one

sanaraufx:
Upon further investigation, this issue should now be resolved. Thanks

PR #9348, opened 2020-07-27
verify the producer name before loading chain

allenhan2:
## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

Associated https://github.com/EOSIO/eos/issues/9335

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -710,10 +710,10 @@ if( options.count(op_name) ) { \
 
 void producer_plugin::plugin_initialize(const boost::program_options::variables_map& options)
 { try {
+   LOAD_VALUE_SET(options, "producer-name", my->_producers)
    my->chain_plug = app().find_plugin<chain_plugin>();
    EOS_ASSERT( my->chain_plug, plugin_config_exception, "chain_plugin not found" );
    my->_options = &options;
-   LOAD_VALUE_SET(options, "producer-name", my->_producers)
 
    chain::controller& chain = my->chain_plug->chain();
    unapplied_transaction_queue::process_mode unapplied_mode =


allenhan2:
Kevin points out that:
1. Should not add application dependency on eosio_chain
2. Move the startup work from chain_plugin initialization to start up

libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -53,7 +53,7 @@ namespace eosio::chain {
       return n;
    }
 
-   /// Immutable except for fc::from_variant.
+   /// Immutable except for fc::from_variant and istream.
```
tbfleming, line 56: The original comment is incorrect since name has an implicit assignment operator. Probably should remove this comment.

allenhan2: will do, thanks!

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -634,7 +634,7 @@ void producer_plugin::set_program_options(
           "Limits the maximum time (in milliseconds) that is allowed a pushed transaction's code to execute before being considered invalid")
          ("max-irreversible-block-age", bpo::value<int32_t>()->default_value( -1 ),
           "Limits the maximum age (in seconds) of the DPOS Irreversible Block for a chain this node will produce blocks on (use negative value to indicate unlimited)")
-         ("producer-name,p", boost::program_options::value<vector<string>>()->composing()->multitoken(),
+         ("producer-name,p", boost::program_options::value<vector<eosio::name>>()->composing()->multitoken(),
```
heifner, line 637: You need to update `LOAD_VALUE_SET` to use `eosio::name` instead of `std::string`

PR #9350, opened 2020-07-31
RocksDB - db intrinsics refactor

brianjohnson5972:
## Change Description
The current interface::db_**** was pulled into a db_context interface and into an implementation of db_context_chainbase.  apply_context currently is only creating and using the db_context_chainbase, db_context_rocksdb will be added in a subsequent commit to implement db_instrinsics in RockDB.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/db_context_chainbase.cpp
+++ b/libraries/chain/db_context_chainbase.cpp
@@ -42,7 +42,7 @@ namespace eosio { namespace chain {
       }
 
       int32_t db_lowerbound_i64(uint64_t code, uint64_t scope, uint64_t table, uint64_t id) override {
-         return context.db_find_i64(name(code), name(scope), name(table), id);
+         return context.db_lowerbound_i64(name(code), name(scope), name(table), id);
       }
 
       int32_t db_upperbound_i64(uint64_t code, uint64_t scope, uint64_t table, uint64_t id) override {
--- a/libraries/chain/db_context_chainbase.cpp
+++ b/libraries/chain/db_context_chainbase.cpp
@@ -1,0 +1,311 @@
+#include <eosio/chain/apply_context.hpp>
+#include <eosio/chain/exceptions.hpp>
+#include <eosio/chain/db_context.hpp>
+
+namespace eosio { namespace chain {
+
+   struct db_context_chainbase : db_context {
+      apply_context&               context;
+      const name                   receiver;
+
+      db_context_chainbase(apply_context& context, name receiver)
+         : context{ context }, receiver{ receiver } {}
+
+      ~db_context_chainbase() override {}
+
+      int32_t db_store_i64(uint64_t scope, uint64_t table, account_name payer, uint64_t id, const char* buffer , size_t buffer_size) override {
+         return context.db_store_i64(name(scope), name(table), payer, id, buffer, buffer_size);
+      }
+
+      void db_update_i64(int32_t itr, account_name payer, const char* buffer , size_t buffer_size) override {
+         context.db_update_i64(itr, payer, buffer, buffer_size);
+      }
+
+      void db_remove_i64(int32_t itr) override {
+         context.db_remove_i64(itr);
+      }
+
+      int32_t db_get_i64(int32_t itr, char* buffer , size_t buffer_size) override {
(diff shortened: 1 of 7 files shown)


libraries/chain/db_context_chainbase.cpp
```diff
@@ -0,0 +1,311 @@
...
+namespace eosio { namespace chain {
+
+   struct db_context_chainbase : db_context {
+      apply_context&               context;
+      const name                   receiver;
+
+      db_context_chainbase(apply_context& context, name receiver)
+         : context{ context }, receiver{ receiver } {}
+
+      ~db_context_chainbase() override {}
```
TimothyBanks, line 14: Nitpick: This should be implicitly generated by the compiler.

libraries/chain/db_context_chainbase.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+      void db_idx64_update(int32_t iterator, account_name payer, const uint64_t& secondary) override {
+         context.idx64.update( iterator, payer, secondary );
+      }
+
+      void db_idx64_remove(int32_t iterator) override {
+         return context.idx64.remove(iterator);
+      }
+
+      int32_t
```
TimothyBanks, line 72: Why is the return type declared on a different line?

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -553,6 +554,8 @@ class apply_context {
       int  db_upperbound_i64( name code, name scope, name table, uint64_t id );
       int  db_end_i64( name code, name scope, name table );
 
+      db_context& db_get_context();
```
TimothyBanks, line 557: General question, but is apply_context ever used in a const fashion?

libraries/chain/include/eosio/chain/db_context.hpp
```diff
@@ -0,0 +1,194 @@
...
+
+namespace chainbase {
+   class database;
+}
+
+namespace eosio { namespace chain {
+
+      class apply_context;
+
+      inline constexpr name dbram_id  = N(eosio.dbram);
```
TimothyBanks, line 15: Is this being defined in multiple translation units?  Otherwise, I don't think `inline` is necessary as the compiler will make that decision on its own regardless.  Same comment for line below.

libraries/chain/include/eosio/chain/db_context.hpp
```diff
@@ -0,0 +1,194 @@
...
+         virtual int32_t db_idx128_end(uint64_t code, uint64_t scope, uint64_t table) = 0;
+
+         virtual int32_t db_idx128_next(int32_t iterator, uint64_t& primary) = 0;
+
+         virtual int32_t db_idx128_previous(int32_t iterator, uint64_t& primary) = 0;
+
+         /**
+          * interface for 256-bit interger secondary
+          */
+         inline static constexpr uint32_t idx256_array_size = 2;
```
TimothyBanks, line 102: I think constexpr in this case implies inline so the inline keyword isn't necessary.

libraries/chain/include/eosio/chain/apply_context.hpp
brianjohnson5972: Technically, no.  It is used to call apply_context::db_*** methods and apply_context::idx***'s methods, which are all defined as non-const methods.  I'm putting a # warning to circle back when we have gotten past the hard stuff.

libraries/chain/include/eosio/chain/db_context.hpp
brianjohnson5972: removed inline and removed un-needed disk variable

libraries/chain/include/eosio/chain/db_context.hpp
brianjohnson5972: unneeded copy from another file, removed

TimothyBanks:
Looks good to me.

PR #9356, opened 2020-08-04
Increse max_transaction_cpu_usage to 90k

allenhan2:
## Change Description

When testing with genesis mode on my mac desktop, get transaction fails,
The first error looks like (it is ok due to wasm initialization, then retry should fix): 
Error 3080006: Transaction took too long

Then following error is (consistent):
Error 3080004: Transaction exceeded the current CPU usage limit imposed on the transaction

See the transaction took about 50k-70k us,

need to increase "max_transaction_cpu_usage", the upper limit is "max_block_cpu_usage" in the same config.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tutorials/bios-boot-tutorial/genesis.json
+++ b/tutorials/bios-boot-tutorial/genesis.json
@@ -11,7 +11,7 @@
     "context_free_discount_net_usage_den": 100,
     "max_block_cpu_usage": 100000,
     "target_block_cpu_usage_pct": 500,
-    "max_transaction_cpu_usage": 50000,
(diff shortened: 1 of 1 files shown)


allenhan2:
> This PR should be into develop

Will do, will investigate the root cause of test failure here also.

allenhan2:
any PR against develop

PR #9361, opened 2020-08-05
Add script support for CentOS 8

softprofe:
## Change Description

Start 9361 description ......
In script eos/scripts/eosio_build.sh, line 122, set -i  pop error on centos7 and conetos8, so, add OS judge and not use set -i on CentOS, tested can be built without "set -i";
In script eos/scripts/eosio_build_centos.sh, it regard centos version as 7 without judge, and install package which only can be used on centos 7. I add judge and install package which can be used on centos8.  centos 7 will still install the original packages.
End 9361 description...

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


scripts/eosio_build_centos.sh
```diff
@@ -23,6 +24,14 @@ if [[ -d /opt/rh/devtoolset-8 ]]; then
 fi
 # Ensure packages exist
 ensure-yum-packages "${REPO_ROOT}/scripts/eosio_build_centos7_deps"
+fi
+
+if [[ "$(echo ${VERSION} | sed 's/ .*//g')" == 8 ]]; then
+        echo "Install Development Tools ..."
+        dnf install -y  https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
+        dnf group install -y  "Development Tools"
```
spoonincode, line 32: as is, these will require the script to run as root

spoonincode:
will not work once #9198 is merged because it is lacking installation of llvm

softprofe:
> will not work on develop once #9198 is merged because it is lacking installation of llvm
> (won't work as is on 2.0)

Based on my experience of build eos on Centos,  after dnf group install -y  "Development Tools", llvm will  still  be installed by ensure-llvm()

spoonincode:
> llvm will still be installed by ensure-llvm()

Even for an unpinned build?

spoonincode:
My centos8 container was populated with a "minimal" install and interestingly failed to build because the lack of openssl dev being installed. Maybe we should add that as a installed package

softprofe:
> My centos8 container was populated with a "minimal" install and interestingly failed to build because the lack of openssl dev being installed. Maybe we should add that as a installed package

Sure we can install openssl-dev   for centos 8

softprofe:
> > llvm will still be installed by ensure-llvm()
> 
> Even for an unpinned build?

should not, because unpinned build don't need it I think.

spoonincode:
>should not, because unpinned build don't need it I think.

If you merge develop in to this branch I think you'll find you need it

softprofe:
> > should not, because unpinned build don't need it I think.
> 
> If you merge develop in to this branch I think you'll find you need it
Why  ensure-llvm() won't work?

PR #9365, opened 2020-08-06
split amqp_witness_plugin in to two (witness & amqp_witness)

spoonincode:
## Change Description

This splits up amqp_witness_plugin to a base witness_plugin and then an amqp_witness_plugin that ferries that witness signature over amqp. Splitting this up gives more flexibility to allow other options of distributing witness signatures (like placing them on chain); potentiality multiple options enabled at once.

appbase shutdown semantics make dealing with multiple threads a little tricky here. I've modeled the interface a bit how boost signals2 tracking works, though I couldn't use that verbatim because I couldn't find a way to control the "locking" granular enough.

No config items have changed; previous configs should still work as before.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/witness_plugin/witness_plugin.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+   app().get_plugin<chain_plugin>().chain().irreversible_block.connect([&](const chain::block_state_ptr& bsp) {
+      if(bsp->block->timestamp.to_time_point() < fc::time_point::now() - fc::seconds(my->staleness_limit))
+         return;
+
+      std::list<std::shared_ptr<wrapped_shared_ptr_base>> locks;
+      for(auto it = my->callbacks.begin(); it != my->callbacks.end(); ++it) {
+         auto lock = it->weakptr->lock();
+         if(!lock->valid())
+            it = my->callbacks.erase(it);
```
spoonincode, line 62: eh there threading problem here I'll take care of it..

plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
```diff
@@ -44,32 +36,29 @@ void amqp_witness_plugin::set_program_options(options_description& cli, options_
...
-         FC_LOG_AND_DROP();
+      my->rqueue->post_on_io_context([&my, &bsp, &sig, &made_to_other_thread_promise]() {
+         my->rqueue->publish_message(std::make_pair(bsp->header.action_mroot, sig));
+         made_to_other_thread_promise.set_value();
       });
+
+      //don't return from this callback until confirmation the amqp's queue has processed the posted message. If nodeos is shutting down
+      // this plugin's impl (and thus rqueue) could be destroyed immediately after the callback before the post_on_io_context()
+      // above has been dispatched
+      fut.wait();
```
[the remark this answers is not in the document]
spoonincode: I think you're right in this case it's unnecessary due to that

plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
```diff
@@ -44,32 +36,23 @@ void amqp_witness_plugin::set_program_options(options_description& cli, options_
...
-         return;
+   my->rqueue = std::make_unique<reliable_amqp_publisher>(my->amqp_server, my->exchange, my->routing_key, witness_data_file_path, "eosio.node.witness_v0");
 
-      my->rqueue->post_on_io_context([this, bsp]() {
-         try {
-            my->rqueue->publish_message(std::make_pair(bsp->header.action_mroot, my->signature_provider(bsp->header.action_mroot)));
-         }
-         FC_LOG_AND_DROP();
+   app().get_plugin<witness_plugin>().add_on_witness_sig([my=my](const chain::block_state_ptr& bsp, const chain::signature_type& sig) {
+      my->rqueue->post_on_io_context([my, bsp, sig]() {
```
swatanabe-b1, line 50: Whoops.  I missed the self-join that happens when `my` is the last reference to the `amqp_witness_plugin_impl`.  If you capture a raw pointer to `*rqueue`, I *think* it's safe, due to the destructor of `reliable_amqp_publisher`.

plugins/amqp_witness_plugin/amqp_witness_plugin.cpp
```diff
@@ -46,14 +46,14 @@ void amqp_witness_plugin::plugin_startup() {
 
    my->rqueue = std::make_unique<reliable_amqp_publisher>(my->amqp_server, my->exchange, my->routing_key, witness_data_file_path, "eosio.node.witness_v0");
 
-   app().get_plugin<witness_plugin>().add_on_witness_sig([my=my](const chain::block_state_ptr& bsp, const chain::signature_type& sig) {
+   app().get_plugin<witness_plugin>().add_on_witness_sig([my=my.get()](const chain::block_state_ptr& bsp, const chain::signature_type& sig) {
       my->rqueue->post_on_io_context([my, bsp, sig]() {
          my->rqueue->publish_message(std::make_pair(bsp->header.action_mroot, sig));
```
swatanabe-b1, line 51: This can be executed after the destructor of rqueue starts.  "If an object of a standard library type is accessed, and ... the access does not happen before the end of the object’s lifetime, the behavior is undefined unless otherwise specified." [res.on.objects]

PR #9366, opened 2020-08-06
Initial implementation of session

TimothyBanks:
## Change Description
This PR contains a new set of types to hopefully replace the current undo stack implementation.  This version contains an iterator on the session type that will collapse the current state of the session list and iterator over that.  An alternative implementation will be presented in another PR.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+}
+
+template <typename persistent_data_store, typename cache_data_store>
+auto session<persistent_data_store, cache_data_store>::iterator::move_previous_() -> void
+{
+    if (m_database_iterator_state.current == m_database_iterator_state.begin
+        && m_cache_iterator_state.current == std::begin(m_cache_iterator_state.updated_keys))
+    {
+        // We need to wrap around to the end.
+        m_database_iterator_state.current = m_database_iterator_state.end--;
```
TimothyBanks, line 967: This should be pre-decrement.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+        m_cache_iterator_state.current = std::end(m_cache_iterator_state.updated_key)--;
```
TimothyBanks, line 968: This should be pre-decrement.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+
+// Factor for creating the initial session iterator.
+//
+// \tparam predicate A functor used for getting the starting iterating in the cache list and the persistent data store.
+// \tparam comparator A functor for determining if the "current" iterator points to the cache list or the persistent data store.
+template <typename persistent_data_store, typename cache_data_store>
+template <typename predicate, typename comparator>
+auto session<persistent_data_store, cache_data_store>::make_iterator_(const predicate& p, const comparator& c) -> typename session<persistent_data_store, cache_data_store>::iterator
+{
+    if (m_impl)
```
TimothyBanks, line 653: Should be null check.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+    template <typename persistent_data_store_, typename cache_data_store_>
+    friend auto make_session(session<persistent_data_store_, cache_data_store_>& the_session) -> session<persistent_data_store_, cache_data_store_>;
+    
+    class iterator;
+    using const_iterator = const iterator;
+
+    // Defines a key ordered, cyclical iterator for traversing a session (which includes parents and children of each session).
+    // Basically we are iterating over a list of sessions and each session has its own cache of key_values, along with
+    // iterating over a persistent data store all the while maintaining key order with the iterator.
+    class iterator  final: public std::iterator<std::bidirectional_iterator_tag, key_value>
```
TimothyBanks, line 73: Correct white space.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+}
+
+template <typename persistent_data_store, typename cache_data_store>
+session<persistent_data_store, cache_data_store>::session_impl::session_impl(session_impl& parent_)
+: parent{parent_.shared_from_this()},
+  backing_data_store{parent_.backing_data_store}
+{
+    if (auto child_ptr = parent->child.lock())
+    {
+        child_ptr->backing_data_store = nullptr;
```
TimothyBanks, line 269: Should comment how the current logic means that the current child will be abandoned (undone).  The other option is to throw an exception if the parent session already has a child.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+        auto bds_kvs = m_impl->backing_data_store->read(not_found);
+        if (!bds_kvs.empty())
+        {
+            m_impl->cache.write(bds_kvs.first);
+        }
+        not_found = std::move(bds_kvs.second);
+        kvs.insert(std::end(kvs), std::begin(bds_kvs.first), std::end(bds_kvs.first));
+    }
+    
+    return std::make_pair(std::move(kvs), std::move(not_found));
```
TimothyBanks, line 558: `std::make_pair` is no longer necessary.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+            auto& kv = parent->cache.read(key);
+            if (kv == key_value::invalid)
+            {
+                continue;
+            }
+            
+            results.emplace_back(make_kv(kv.key().data(), kv.key().length(), kv.value().data(), kv.value().length(), ds.allocator()));
+        }
+    }
+    ds.write(results);
```
TimothyBanks, line 631: Currently this only pulls in keys that are in the caches.  If I can't find the key in the cache, should I hit the database to pull it in from there?

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1061 @@
...
+        {
+            not_found.emplace(key);
+        }
+    }
+    
+    // For all the keys that were found in one of the session caches, read them from the persistent data store
+    // and write them into our cache.
+    if (m_impl->backing_data_store)
+    {
+        auto bds_kvs = m_impl->backing_data_store->read(not_found);
```
TimothyBanks, line 549: Use structured bindings here.

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -0,0 +1,230 @@
...
+namespace std
+{
+template <>
+class less<b1::session::bytes> final
+{
+public:
+    constexpr auto operator()(const b1::session::bytes& lhs, const b1::session::bytes& rhs) const -> bool
+    {
+        // Shorter keys are "less than" longer keys
+        if (lhs.length() < rhs.length())
```
kimjh2005, line 129: Other existing blob compare functions such as compare_blob() in libraries/chain/include/eosio/chain/types.hpp does lexicographic order.  ex)  "aa" comes before "z"

libraries/chain_kv/include/b1/session/cache.hpp
```diff
@@ -0,0 +1,454 @@
...
+    // Namely, we need to be able to iterator over the entries in lexigraphical ordering on the keys.
+    using cache_type = std::map<bytes, key_value>;
+
+public:
+    using allocator_type = allocator;
+    
+    class iterator;
+    using const_iterator = const iterator;
+    
+    class iterator  final: public std::iterator<std::bidirectional_iterator_tag, key_value>
```
kimjh2005: std::iterator is deprecated in C++17

heifner: https://www.fluentcpp.com/2018/05/08/std-iterator-deprecated/

libraries/chain_kv/include/b1/session/cache.hpp
```diff
@@ -0,0 +1,454 @@
...
+{
+    auto it = m_cache.find(key);
+    
+    if (it == std::end(m_cache))
+    {
+        cache_miss();
+        return key_value::invalid;
+    }
+    
+    if (cache_hit(it))
```
kimjh2005, line 148: When does cache_hit() return false?

kimjh2005: I saw it returns false when it is erased inside cache_hit().  it may need a comment.

libraries/chain_kv/include/b1/session/bytes.hpp
TimothyBanks: I'll just try to pull that functionality in and invoke within this method call.

libraries/chain_kv/bytes.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+#include <boost/pool/pool.hpp>
+
+#include <b1/session/bytes.hpp>
+
+using namespace b1::session;
+
+const bytes bytes::invalid{};
+
+bytes::bytes(const bytes& b)
```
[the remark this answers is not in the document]
TimothyBanks: I'm open to suggestions.  What could another name be for this type?

libraries/chain_kv/bytes.cpp
```diff
@@ -0,0 +1,119 @@
+#include <string_view>
+
+#include <boost/pool/pool.hpp>
+
+#include <b1/session/bytes.hpp>
+
+using namespace b1::session;
+
+const bytes bytes::invalid{};
```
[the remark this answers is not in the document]
TimothyBanks: invalid is just a representation of a bytes instance where all data members are null which is how one of these types is default constructed.  Since the default constructor is hidden, I needed a way for a user to compare a bytes instance provided to them against what is being defined as an invalid instance.

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+template <typename other>
+bool boost_memory_allocator::equals(const other& right) const {
+  // Just pointer comparision for now.
+  return this == &right;
+}
+
+inline void boost_memory_allocator::initialize_() {
+  m_free_function 
+    = [weak = std::weak_ptr<boost_memory_allocator>(this->shared_from_this())](void* data, size_t length_bytes) {
```
heifner, line 46: `weak_from_this()` is a thing

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -0,0 +1,284 @@
...
+  size_t operator()(const eosio::session::bytes& b) const {
+    return std::hash<std::string_view>{}({reinterpret_cast<const char*>(b.data()), b.length()});
+  }
+};
+
+template <>
+class equal_to<eosio::session::bytes> final
+{
+public:
+    auto operator()(const eosio::session::bytes& lhs, const eosio::session::bytes& rhs) const -> bool
```
heifner, line 278: Am I missing something. Why is this: `auto ... -> bool` vs `bool ...`

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
```diff
@@ -0,0 +1,79 @@
...
+// methods as demonostrated in this type.
+class boost_memory_allocator : public std::enable_shared_from_this<boost_memory_allocator> {
+public:
+    void* malloc(size_t length_bytes) const;
+    void free(void* data, size_t length_bytes) const;
+    free_function_type& free_function();
+
+    static std::shared_ptr<boost_memory_allocator> make();
+
+    template <typename other>
```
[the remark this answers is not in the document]
TimothyBanks: If you're referring to the template parameters (I have seen a lot of type names such as classes, structs that are all lower case), it doesn't seem like that is necessarily what is standardized within the eos source directory.  I've seen camel case, pascal case, all lower case.  So this was just personal preference.  I could easily be convinced to change it if there is actually a casing that is desired.

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
```diff
@@ -0,0 +1,79 @@
...
+    return const_cast<boost::pool<>&>(m_pool).ordered_malloc(length_bytes);
+}
+
+// \brief Frees a chunks of memory.
+//
+// \param data A pointer to the memory to free.
+// \param length_bytes The size of the memory, in bytes.
+inline void boost_memory_allocator::free(void* data, size_t length_bytes) const {
+    const_cast<boost::pool<>&>(m_pool).ordered_free(data);
+}
```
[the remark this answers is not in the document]
TimothyBanks: The idea in this case was the `free` method is part of an adapter interface and it just happens that the adaptee doesn't need that parameter.  I could those "hide" the parameter name just in case there are warnings about unused variables.

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <boost/pool/pool.hpp>
+
+#include <b1/session/bytes_fwd_decl.hpp>
+
+namespace eosio::session {
+
+// \brief Defines a memory allocator backed by a boost::pool instance.
+//
+// \remarks This also demonstrates the "memory allocator" concept.  If we want to introduce another
```
swatanabe-b1, line 13: Is there a reason for introducing a custom allocator concept instead of using std::allocator?

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -0,0 +1,284 @@
...
+        return;
+    }
+    
+    if (--(*m_use_count_address) != 0) {
+        return;
+    }
+    
+    // The memory pool must remain in scope for the lifetime of this instance.
+    auto* free_function = reinterpret_cast<free_function_type*>(*m_memory_allocator_address);
+    (*free_function)(m_data, 3 * sizeof(size_t) + *m_length);
```
swatanabe-b1, line 174: Why do you need this indirection?  Couldn't m_memory_allocator_address be a `free_function_type*`?

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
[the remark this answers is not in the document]
swatanabe-b1: Template parameters are typically CamelCase.

libraries/chain_kv/bytes.cpp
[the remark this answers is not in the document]
swatanabe-b1: A bigger risk is eosio::chain::bytes:
```c++
// #include <eosio/chain/types.hpp>
using bytes               = vector<char>;
```

Maybe `shared_bytes` to describe its ownership semantics.

libraries/chain_kv/include/b1/session/key_value.hpp
```diff
@@ -0,0 +1,148 @@
...
+// \param key A pointer to the key.
+// \param key_length The size of the memory pointed by the key.
+// \param value A pointer to the value.
+// \param value_length The size of the memory pointed by the value.
+// \param a The memory allocator used to managed the memory used by the key_value instance.
+// \returns A key_value instance.
+// \remarks This factory guarentees that the memory needed for the key and value will be contiguous in memory.
+template <typename allocator>
+key_value make_kv(const void* key, size_t key_length, const void* value, size_t value_length, allocator& a) {
+    // TODO:  Not quite sure how to manage this memory chunk.
```
swatanabe-b1, line 97: To make the key and the value contiguous, you can allocate a single buffer and set its initial reference count to 2, letting it be shared by `m_key` and `m_value`.

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -0,0 +1,284 @@
...
+    bytes() = default;
+    
+private:
+    // The encoded address of the memory allocator that instantiated the memory this instance points to.
+    // \warning The memory allocator MUST stay in scope during the lifetime of this instance.
+    size_t* m_memory_allocator_address{nullptr};
+    // \brief A pointer to a counter to keep track of the number of bytes instances that point to the memory address.
+    // \remark When the counter reaches 0, the last instance will free the memory by invoking the free method on the memory allocator.
+    // \warning Assuming that this type will be accessed synchronously.
+    size_t* m_use_count_address{nullptr};
```
swatanabe-b1, line 67: Do you need two separate pointers for `m_use_count_address` and `m_memory_allocator_address`?  `shared_ptr` is typically implemented with a single pointer to a type that holds both the reference count and the deleter.

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
```diff
@@ -0,0 +1,79 @@
...
+    bool equals(const other& right) const;
+
+protected:
+    boost_memory_allocator() = default;
+    void initialize_();
+
+private:
+    // Currently just instantiating a boost::memory instance to start.
+    // We can change this to try other boost specific memory pools
+    boost::pool<> m_pool{sizeof(char)};
```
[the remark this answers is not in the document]
TimothyBanks: @swatanabe-b1 Do you have any ideas on which allocator would be a good one to start with?  I didn't spend a lot of time researching this as I just wanted to start with something to get a baseline for performance testing.

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -0,0 +1,284 @@
...
+}
+
+inline const void * const bytes::data() const {
+    return m_data;
+}
+
+inline size_t bytes::length() const {
+    if (!m_memory_allocator_address) {
+        // The size was encoded as the address of this data member.
+        return reinterpret_cast<size_t>(m_length);
```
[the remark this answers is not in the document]
TimothyBanks: Yeah it could.  I'll make the change.

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -0,0 +1,284 @@
...
+namespace eosio::session {
+
+// \brief An immutable type to represent a pointer and a length.
+//
+class bytes final {
+public:
+    friend class key_value;
+    
+    template <typename T, typename allocator>
+    friend bytes make_bytes(const T* data, size_t length, allocator& a);
```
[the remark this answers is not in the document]
TimothyBanks: I had a few reasons that I went this route:
1.  At some point I wanted to minimize the scope of friends and I preferred friending a function than a class.
2.  I was debating implementing some sort of pool of this types.  I also felt this design would allow for performance enhancement opportunities.
3.  I liked the idea of keeping the creation logic separate from the type so we could allow injecting new overloaded free functions.

At this point though, we could make them constructors as well.

libraries/chain_kv/include/b1/session/key_value.hpp
TimothyBanks: Perfect!

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
TimothyBanks: I have none.  Will make it conform to that concept.

libraries/chain_kv/include/b1/session/bytes.hpp
TimothyBanks: Just out of curiosity, do you know what advantages that provides?  I could understand getting rid of one indirection in the case of the last reference going out of scope but is there anything beyond that?

TimothyBanks:
Updated.

libraries/chain_kv/include/b1/session/key_value.hpp
```diff
@@ -0,0 +1,157 @@
...
+    auto populate_bytes = [&a](const void* data, size_t length, char* chunk) mutable
+    {
+        auto result = shared_bytes{};
+        
+        if (!data || length == 0)
+        {
+          return result;
+        }
+        
+        result.m_chunk_start = reinterpret_cast<uint64_t*>(chunk);
```
b1bart, line 111: This chunk format needs to be documented and could very well be a reified structure at this point.  I don't see any reason why its packed size done this way would be different than that of a struct.  Meanwhile, if it is on a given platform we would be able to respect the alignment using standard keywords.

libraries/chain_kv/include/b1/session/shared_bytes.hpp
```diff
@@ -0,0 +1,292 @@
...
+    // To allow for sharing a chunk of memory across multiple shared_byte instances, we need to keep track
+    // of the chunk start in each instance to pass back to the allocator.
+    uint64_t* m_chunk_start{nullptr};
+    // The encoded address of the memory allocator that instantiated the memory this instance points to.
+    // \warning The memory allocator MUST stay in scope during the lifetime of this instance.
+    uint64_t* m_memory_allocator_address{nullptr};
+    // \brief A pointer to a counter to keep track of the number of shared_bytes instances that point to the memory address.
+    // \remark When the counter reaches 0, the last instance will free the memory by invoking the free method on the memory allocator.
+    // \warning Assuming that this type will be accessed synchronously.
+    size_t* m_use_count_address{nullptr};
```
b1bart, line 70: This comment implies that this refcount makes the system not thread safe.  is there a reason to use a raw scalar vs an atomic type counter?

b1bart (changes requested):

libraries/chain_kv/include/b1/session/shared_bytes.hpp
TimothyBanks: I was under the impression that the system operated synchronously.  So given that, it didn't seem necessary to have the overhead of an atomic.  Is the system advertised as thread safe?  I can make it atomic if so.

libraries/chain_kv/include/b1/session/key_value.hpp
TimothyBanks: The chunk returned from the allocate call is already aligned, at least from what I've seen.

libraries/chain_kv/include/b1/session/shared_bytes.hpp
brianjohnson5972: @b1bart is your point that this could eventually be used asynchronously? If we want this code to be safe to access asynchronously, we first need to decide the use cases (i.e. what is going to be used asynchronously) and then make that usage safe.  I see this comment as a bread crumb for if this ever happens, but really shouldn't matter, since we will need to look at the whole design to make sure it is safe, but making this an atomic seems to me to be a lot of cost compared to the amount of work that will still be needed if we do make it asynchronous.

libraries/chain_kv/include/b1/session/boost_memory_allocator.hpp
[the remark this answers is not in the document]
swatanabe-b1: Actually, why roll your own reference counted pointer.  Would there be any problem with using `std::allocate_shared<char[]>(std::allocator<char>(), size)` (C++20 only) or `boost::allocate_shared<char[]>(std::allocator<char>(), size)`?

libraries/chain_kv/include/b1/session/bytes.hpp
swatanabe-b1: It reduces the size of the `shared_ptr` object and generally makes thing simpler.

libraries/chain_kv/include/b1/session/bytes.hpp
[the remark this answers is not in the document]
swatanabe-b1: Yes, but why?  I don't any reason why the size needs to be stored in the allocated block.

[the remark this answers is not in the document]
TimothyBanks: Maybe I don't understand what you're getting at.  The idea was that the allocated chunk would hold all data needed by the shared_bytes type so all of its data was contiguous in memory and moving/copying just involved passing pointers around .   Maybe I could define a structure for some parts of the allocated chunk and just overlay it, but there are other things that are shared between different shared_bytes instances, such as the chunk start and the reference count (and possibly the free function if I combine those two things into a structure) and those would need to be pointers.

libraries/chain_kv/include/b1/session/bytes.hpp
TimothyBanks: Ok.  I look at putting some of those data members into an internal struct.

libraries/chain_kv/include/b1/session/bytes.hpp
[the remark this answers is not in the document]
TimothyBanks: I think I got it.  I'll push a change up shortly.

libraries/chain_kv/include/b1/session/key_value.hpp
TimothyBanks: Injected a structured.

libraries/chain_kv/include/b1/session/shared_bytes.hpp
b1bart: I don't have a clear view of the whole system, and while it is true the db system operates synchronously, plenty of things that inspect data during operation operate asynchronously.  My concern is centered around the basic notion that a a future engineer may not have the nuance view to not place one of these spans in an action_trace etc.  

something like:
```
struct shared_span {
   size_t                     m_length = 0;
   std::shared_ptr<uint8_t[]> m_data;
}
```

seems conceptually simpler, is thread safe, and still supports safely aliasing within a span via `std::shared_ptr`'s aliasing constructor

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1244 @@
...
+//          transaction.backing_data_store()->memory_allocator()); transaction.write(std::move(kv));
+//      }
+//      {
+//          auto transaction = eosio::session::make_session(block);
+//          auto kv = eosio::session::make_kv("Hello", 5, "World2", 5,
+//          transaction.backing_data_store()->memory_allocator()); transaction.write(std::move(kv));
+//      }
+//  }
+// \endcode
+template <typename Persistent_data_store, typename Cache_data_store>
```
larryk85, line 40: This shouldn't necessarily be a `persistent` data store, it is just it's parent type, that parent type might be something that is more persistent, but this type shouldn't make any assumptions about that.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1244 @@
...
+   };
+
+   using persistent_data_store_type = Persistent_data_store;
+   using cache_data_store_type      = Cache_data_store;
+   using iterator_cache_type        = std::map<shared_bytes, iterator_state>;
+
+   static const session invalid;
+
+   template <typename T, typename Allocator>
+   friend shared_bytes make_shared_bytes(const T* data, size_t length, Allocator& a);
```
larryk85, line 59: I would remove these `friend` functions.  They shouldn't be needed with constructors.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -0,0 +1,1244 @@
...
+      void move_next_();
+      void move_previous_();
+
+    private:
+      typename iterator_cache_type::iterator m_active_iterator;
+      session*                               m_active_session{ nullptr };
+   };
+
+   struct iterator_traits final {
+      using difference_type     = long;
```
larryk85, line 125: Why is the difference_type a `long` this is likely to be 32-bits was that intended?  I don't fully understand what the difference of a `session_iterator` would mean.

larryk85 (changes requested):
I am not finished reviewing yet.  Please wait until I am done reviewing this before addressing the concerns, this PR is quite large and I don't want to lose my place with what I have reviewed so far.

libraries/chain_kv/include/b1/session/cache.hpp
```diff
@@ -0,0 +1,438 @@
...
+// following singature void(void). \param key The key to search the cache for. \param cache_hit The functor to invoke
+// when the key is found. \param cache_miss The functor to invoke when the key is not found. \return A reference to the
+// key_value instance if found, key_value::invalid otherwise.
+template <typename Allocator>
+template <typename on_cache_hit, typename on_cache_miss>
+const key_value& cache<Allocator>::find_(const shared_bytes& key, const on_cache_hit& cache_hit,
+                                         const on_cache_miss& cache_miss) const {
+   auto it = m_cache.find(key);
+
+   if (it == std::end(m_cache)) {
```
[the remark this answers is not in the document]
TimothyBanks: Yes.  The argument about it being more characters to type is highly dependent on the naming of the data member, I could have easily called `m_cache` something else that made it shorter to type.  I prefer the usage of `std::begin`, `std::end` because it is generic and  I could change m_cache to be a type that doesn't have that interface on it and not be bogged down in changing the `.end()` calls to `std::end`.

libraries/chain_kv/include/b1/session/cache.hpp
```diff
@@ -0,0 +1,438 @@
+#pragma once
```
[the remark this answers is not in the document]
TimothyBanks: Originally I implemented this type because of two reasons:
1.  Circular iterators.  `std::map` doesn't support circular iterators.  Which leads into the second reason...
2.  A common interface of all types in this design that store data.  I can cache, session, rocks_sessions interchangeably and in a generic fashion.  There was originally discussion about using `std::map`, 'std::unordered_map' or whatever container was discovered to be the best after performance tests. I wanted to insulate Session from the mess of changing between those types.  Ultimately when we discover the best backing container after performance testing, we can merge the container directly into session.

While I ultimately ended up with implementing the circular iterator logic in Session (and specializations of Session), the second reason still holds.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+
+    private:
+      std::shared_ptr<rocksdb::DB>       m_db;
+      std::shared_ptr<rocksdb::Iterator> m_iterator;
+      rocksdb::ColumnFamilyHandle*       m_column_family;
+      rocksdb::ReadOptions               m_read_options;
+      std::shared_ptr<Allocator>         m_allocator;
+   };
+
+   struct iterator_traits final {
```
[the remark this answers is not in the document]
TimothyBanks: There are many ways that this could be implemented.  I choose to keep it similar to what `std::iterator` used to provide.

libraries/chain_kv/include/b1/session/rocks_data_store_fwd_decl.hpp
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
```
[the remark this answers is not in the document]
TimothyBanks: Removed.

libraries/chain_kv/include/b1/session/cache.hpp
```diff
@@ -0,0 +1,438 @@
...
+// Searches the cache for the given key.
+//
+// \tparam on_cache_hit A functor to invoke when the key is found.  It has the following signature
+// void(cache_type::iterator& it) \tparam on_cache_miss A functor to invoke when the key is not found.  It has the
+// following singature void(void). \param key The key to search the cache for. \param cache_hit The functor to invoke
+// when the key is found. \param cache_miss The functor to invoke when the key is not found. \return A reference to the
+// key_value instance if found, key_value::invalid otherwise.
+template <typename Allocator>
+template <typename on_cache_hit, typename on_cache_miss>
+const key_value& cache<Allocator>::find_(const shared_bytes& key, const on_cache_hit& cache_hit,
```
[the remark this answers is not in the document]
TimothyBanks: Please elaborate on why you are calling this an anti-pattern.  This was intended to be the helper method and ultimately it is only used by this type in a way that defines the intended usage.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+// \tparam Iterable Any type that can be used within a range based for loop and returns key_value instances in its
+// iterator. \param key_values An Iterable instance that returns key_value instances in its iterator.
+template <typename Allocator>
+template <typename Iterable>
+void rocks_data_store<Allocator>::write(const Iterable& key_values) {
+   if (!m_db) {
+      return;
+   }
+
+   auto batch = rocksdb::WriteBatch{ 1024 * 1024 };
```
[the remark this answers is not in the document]
TimothyBanks: It was mainly chosen because I thought blocks were at most 1MB.  So I reserved space for 1MB.  Performance testing determine if a parameter is required or not.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+
+template <typename Allocator>
+void rocks_data_store<Allocator>::clear() {
+   // TODO:
+}
+
+template <typename Allocator>
+template <typename Iterable, typename Other_allocator>
+const std::pair<std::vector<key_value>, std::unordered_set<shared_bytes>>
+rocks_data_store<Allocator>::read_(const Iterable& keys, const std::shared_ptr<Other_allocator>& a) const {
```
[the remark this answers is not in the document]
TimothyBanks: I would argue that the type names chosen here are quality.  Just because they are long doesn't lessen the quality.  If the coding style guidelines details how non public members of a type should be named then I can change them to follow that.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+
+   auto kvs = std::vector<key_value>{};
+   kvs.reserve(key_slices.size());
+
+   for (size_t i = 0; i < values.size(); ++i) {
+      if (status[i].code() != rocksdb::Status::Code::kOk) {
+         continue;
+      }
+
+      not_found.erase(make_shared_bytes(key_slices[i].data(), key_slices[i].size()));
```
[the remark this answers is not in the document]
TimothyBanks: I created this function to explicitly do that.  This is the point at which all the work has already been done to know which keys were found and which ones weren't.

libraries/chain_kv/include/b1/session/session.hpp
TimothyBanks: Copy and paste errors.  Updated.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+      return;
+   }
+
+   auto key_slice = rocksdb::Slice{ reinterpret_cast<const char*>(key.data()), key.length() };
+   auto status    = m_db->Delete(m_write_options, column_family_(), key_slice);
+}
+
+template <typename Allocator>
+void rocks_data_store<Allocator>::clear() {
+   // TODO:
```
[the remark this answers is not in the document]
TimothyBanks: This specialization doesn't have an implementation for it, but it is on the session interface and does get called generically so this type needs an definition to compile.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+
+   auto key_slice = m_iterator->key();
+   auto value     = m_iterator->value();
+   return make_kv(key_slice.data(), key_slice.size(), value.data(), value.size(), m_allocator);
+}
+
+template <typename Allocator>
+template <typename Iterator_traits>
+bool rocks_data_store<Allocator>::rocks_iterator<Iterator_traits>::operator==(const rocks_iterator& other) const {
+   if (!m_iterator->Valid() && !other.m_iterator->Valid()) {
```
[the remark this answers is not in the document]
TimothyBanks: keyslices don't overload the equality operator so I will still need to use the compare method.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+   auto key_slices = std::vector<rocksdb::Slice>{};
+
+   for (const auto& key : keys) {
+      key_slices.emplace_back(reinterpret_cast<const char*>(key.data()), key.length());
+      not_found.emplace(key);
+   }
+
+   auto values = std::vector<std::string>{};
+   values.reserve(key_slices.size());
+   auto status = m_db->MultiGet(m_read_options, { key_slices.size(), column_family_() }, key_slices, &values);
```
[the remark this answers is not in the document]
TimothyBanks: There are also restrictions on that overload of the function.  We should wait until the parameters being passed into RocksDB have been fully discovered before going down that path.

libraries/chain_kv/include/b1/session/rocks_data_store.hpp
```diff
@@ -0,0 +1,684 @@
...
+
+template <typename Allocator>
+template <typename Iterable, typename Other_allocator>
+const std::pair<std::vector<key_value>, std::unordered_set<shared_bytes>>
+rocks_data_store<Allocator>::read_(const Iterable& keys, const std::shared_ptr<Other_allocator>& a) const {
+   if (!m_db) {
+      return {};
+   }
+
+   auto not_found  = std::unordered_set<shared_bytes>{};
```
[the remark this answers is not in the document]
TimothyBanks: The parameter doesn't necessarily give me that information.  I could reserve a size, but it would be a guess at the upper bound.

larryk85 (approved):
We will address the comments at a later date.

libraries/chain/CMakeLists.txt
```diff
@@ -139,7 +139,7 @@ target_include_directories( eosio_chain
                                    "${CMAKE_CURRENT_SOURCE_DIR}/../wasm-jit/Include"
                                    "${CMAKE_CURRENT_SOURCE_DIR}/libraries/eos-vm/include"
                                    "${CMAKE_CURRENT_SOURCE_DIR}/../rocksdb/include"
-                                   "${CMAKE_CURRENT_SOURCE_DIR}/../chain_kv/include/b1"
```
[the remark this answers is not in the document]
brianjohnson5972: Please don't do this.  Ignore any non-necessary changes.  I will never finish if things start moving.

libraries/chain/include/eosio/chain/backing_store/db_key_value_format.hpp
```diff
@@ -1,92 +1,110 @@
...
+      sec_double      = 4,
       sec_long_double = 5
    };
 
    namespace detail {
 
-      template<typename Key>
-      struct determine_sec_type {
-      };
+      template <typename Key>
```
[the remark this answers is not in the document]
brianjohnson5972: please ignore this and undo all changes to my code, thanks!

libraries/chain/include/eosio/chain/backing_store/db_key_value_format.hpp
```diff
@@ -1,92 +1,110 @@
...
          constexpr static auto scope_size = sizeof(scope);
          constexpr static auto table_size = sizeof(table);
-         constexpr static auto type_size = sizeof(key_type);
-         static_assert( type_size == 1, "" );
-         key_storage.reserve(scope_size + table_size + type_size + key_size);
-         b1::chain_kv::append_key(key_storage, scope.to_uint64_t());
-         b1::chain_kv::append_key(key_storage, table.to_uint64_t());
-         key_storage.push_back(static_cast<char>(kt));
+         constexpr static auto type_size  = sizeof(key_type);
+         static_assert(type_size == 1, "");
```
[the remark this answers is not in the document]
brianjohnson5972: please ignore

brianjohnson5972 (changes requested):
Please undo any changes that have nothing to do with this fix, it will only make my and Timothy's life much harder.

libraries/rodeos/include/b1/rodeos/callbacks/kv.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
-#include <b1/chain_kv/chain_kv.hpp>
 #include <b1/rodeos/callbacks/vm_types.hpp>
+#include <b1/chain_kv/chain_kv.hpp>
```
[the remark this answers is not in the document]
TimothyBanks: I'll undo these.  I originally encountered an inconsistency where some files used `b1/chain_kv` and others used `chain_kv`.  I think it was settled on making them all use `b1/chain_kv`.

TimothyBanks:
Updated.  This is ready for merging into feature branch.  Remaining comments will be taken care of in 2.1.

libraries/chain/combined_database.cpp
```diff
@@ -183,33 +196,30 @@ namespace eosio { namespace chain {
...
 
-   std::unique_ptr<kv_context> combined_database::create_kv_context(name receiver,
-                                                                    kv_resource_manager resource_manager,
+   std::unique_ptr<kv_context> combined_database::create_kv_context(name receiver, kv_resource_manager resource_manager,
                                                                     const kv_database_config& limits) {
       switch (backing_store) {
          case backing_store_type::ROCKSDB:
-            return create_kv_rocksdb_context<b1::chain_kv::view, b1::chain_kv::write_session, kv_resource_manager>(
-                  *kv_database, *kv_undo_stack, receiver, resource_manager, limits);
+            return create_kv_rocksdb_context<session_type, kv_resource_manager>(kv_undo_stack->top(), receiver,
```
TimothyBanks, line 211: @larryk85 @linhuang-blockone @revl 
I'm running into an exception where there is no top of the undo stack at this point.  I'm thinking that I might need to push a new session but I'm not quite sure where that should be done.  I was thinking one of two places:

1.  I can push a session right before this line.  
2.  I can push a new session when the combined database instance is creating.

With both of these places, would I always be pushing a session or would I only push a session when the undo stack is empty?

PR #9369, opened 2020-08-07
set medium priority for process signed block - develop

allenhan2:
## Change Description

This is change related to https://github.com/EOSIO/eosio-security/issues/340,

Still see the unlinkable block for each restart but it is expected when switching from in-sync to head-catchup and that is the current block that the node just produced

Co-Authored-By: Kevin Heifner heifnerk@objectcomputing.com

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
Introduced by: https://github.com/EOSIO/eos/pull/8142

allenhan2:
Will do more testing on develop with config bp<--API<--API.

allenhan2:
The test goes well, and see resync being reestablished much faster than 2.0.x

heifner:
I think the lesson learned here is you can't switch from a lower priority `app().push()` value to a higher value as the higher priority items are likely to then be executed before the lower priority items. In this particular case it results in blocks being received out of order leading to unlinkable blocks.

heifner:
`release/2.0.x` version #9370

heifner:
@allenhan2 can you look at the test failure: https://buildkite.com/EOSIO/eosio/builds/23856#60e2f506-0ae1-4257-a18e-df157add03a5 I don't think this is a test we ever see fail.

allenhan2:
This test runs well on my local machine and run well after rebuild

dimas1185:
merging on behalf of Yue

heifner:
> merging on behalf of Yue

Thanks for the indication that you merged on behalf on Yue. What of the failed test I asked about above?

allenhan2:
> > merging on behalf of Yue
> 
> Thanks for the indication that you merged on behalf on Yue. What of the failed test I asked about above?

commented above,
This test runs well on my local machine and run well after rebuild

heifner:
Failed tests should be investigated. Retrying to see if it passes causes us to have flakey tests. I don't remember seeing that particular test fail before. Can you take a look at the logs to see why it failed.

allenhan2:
If I remember correctly, it complains late produce of the block was late

allenhan2:
The test on this PR https://github.com/EOSIO/eos/pull/9403 (which has this change) ran well also

PR #9379, opened 2020-08-10
witness_trx_plugin: place witness signatures on chain

spoonincode:
## Change Description

amqp_witness_plugin sends witness signatures over AMQP, this new witness_trx_plugin places witness signatures on chain. For some oracles this can simplify their design considerably: they don't need to deal with AMQP at all, and witness signatures can no longer arrive earlier than the blocks they are witnessing.

My first cut at this plugin was extreme simplicity with a "fire and forget" attitude. However, if witness_trx_plugin simply submits transactions when witnessing the block, those transactions may have already been expired from the BP's standpoint if the node doing the witnessing is behind (due to syncing, etc).

This second attempt will make better efforts to ensure all witness signatures will make it in a transaction eventually. Witness signatures are buffered (and persisted to disk if need be) until the node is caught up. Additionally, the plugin expects these transactions to be included in an irreversible block and if they aren't by the expiry time, it reattempts.

The implementation for these hardening bits is a little crude (i.e. simple) and that can lead to some witness signatures being submitted more than once or being submitted out of order. This is considered acceptable at the moment as oracles on AMQP would have had to deal with those same (lack of) constraints.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/witness_trx_plugin/witness_trx_plugin.cpp
```diff
@@ -0,0 +1,243 @@
...
+               batch.splice(batch.begin(), sig_action_data_waiting_on_catch_up, sig_action_data_waiting_on_catch_up.begin(), it);
+               submit_witness_trx(std::move(batch));
+            }
+         }
+
+         auto it = outstanding_witness_transaction_index.get<by_trx_expiration>().begin();
+         while(it != outstanding_witness_transaction_index.get<by_trx_expiration>().end() && it->trx_expires < chain::time_point::now()) {
+            wlog("Witness transaction not confirmed by producers; resubmitting");
+            outstanding_witness_transaction_index.get<by_trx_expiration>().modify(it, [this](witness_trx_plugin_impl::outstanding_witness_transaction& row) {
+               submit_witness_trx(std::move(row.action_data));
```
swatanabe-b1, line 78: Does it make sense to resubmit transactions when we aren't `caught_up`?  Conversely, what is the motivation for delaying new transactions when we aren't `caught_up`?

swatanabe-b1: Also if outgoing transactions get backlogged, would we want to batch them?

plugins/witness_trx_plugin/witness_trx_plugin.cpp
```diff
@@ -0,0 +1,243 @@
...
+void witness_trx_plugin::plugin_startup() {
+   app().get_plugin<witness_plugin>().add_on_witness_sig([my=my](const chain::block_state_ptr& bsp, const chain::signature_type& sig) {
+      //if nodeos is shutting down, it is possible witness_trx_plugin_impl will be destroyed immediately after the callback returns before
+      // the post() below is actually dispatched. Prevent returning until certain post() has dispatched.
+      std::promise<void> made_to_other_thread_promise;
+      auto fut = made_to_other_thread_promise.get_future();
+
+      my->ctx.post([my, &bsp, &sig, &made_to_other_thread_promise]() {
+         std::vector<char> witness_action_data = fc::raw::pack(std::make_pair(bsp->header.action_mroot, sig));
+         made_to_other_thread_promise.set_value();
```
swatanabe-b1, line 193: If nodeos is shutting down, then once we notify the calling thread, the rest of nodeos could run to completion including returning from main, before the rest of this function executes.  We'll be left with a dangling reference to `chainplug` in that case AND the worker thread will join itself.

spoonincode: chainplug is easy to refactor out; it's stuff like the `app()` access that would be harder. Starting to like the old `weak_ptr` approach more...

swatanabe-b1: Wouldn't the `weak_ptr` approach only affect the outer callback, not this inner one?

spoonincode: I think `weak_ptr` may make it easier to deal with something like `app()`; or more generally, ensuring all dependent plugin_impls are destroyed before witness_plugin_impl stop() returns. But yeah the self join() problem remains.

PR #9380, opened 2020-08-10
change witness_plugin to by default sign blocks of any age

spoonincode:
## Change Description

witness_plugin's previous default was to only sign blocks timestamped within 10 minutes of its wall clock. I think this might be a nasty trap. If BP(s) out run a witness_plugin node by over 10 minutes, when that witness_plugin node catches up there will be some blocks that it won't sign. In the case that a quorum of witnesses fall behind the BP(s) by 10 minutes, that quorum will never sign to the threshold needed to take action based on that quorum.

Change the default to always sign any block regardless of the block's age.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/witness_plugin/witness_plugin.cpp
+++ b/plugins/witness_plugin/witness_plugin.cpp
@@ -6,7 +6,7 @@ static appbase::abstract_plugin& _witness_plugin = app().register_plugin<witness
 
 struct witness_plugin_impl {
    signature_provider_plugin::signature_provider_type signature_provider;
-   unsigned staleness_limit = 600;
+   std::optional<unsigned> staleness_limit;
 
    struct callback_entry {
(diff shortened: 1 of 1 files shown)


swatanabe-b1:
If it's a trap, then do we need this option at all?  I think the case of not signing on replay would be better handled with an absolute limit (either time or block number) or simply replaying without the plugin and then restarting nodeos.

spoonincode:
For the most part I'm anti-config-bloat so I'm game with removing it entirely.

spoonincode:
replaced with #9407 which removes the option entirely

Issue #9278, opened 2020-07-03
New accounts fail lookups via get_accounts_by_authorizers


aaroncox:
It's recently come to our attention that the new `v1/chain/get_accounts_by_authorizers` API endpoint will stop updating and adding new key/auth -> account associations to its dataset. All previous lookups will remain functional, but any account created after an unknown period in time (or event) will not be returned unless nodeos is restarted.

We have seen his both on our EOS API infrastructure as well as on Jungle 3. Here are some example calls against our API infrastructure that illustrate the account not returning in `v1/chain/get_accounts_by_authorizers`, but returning results through `v1/history/get_key_accounts`. 

```bash
curl https://jungle3.greymass.com/v1/chain/get_accounts_by_authorizers -d '{"keys": ["EOS7LQUHUHPNLY4T8gbZjtWhxqTNw8ukzqTvrk12uchnHM1gAaR71"]}'

curl https://jungle3.greymass.com/v1/history/get_key_accounts -d '{"public_key":"EOS7LQUHUHPNLY4T8gbZjtWhxqTNw8ukzqTvrk12uchnHM1gAaR71"}'
```

We will leave our Jungle 3 API running in this state if it helps further troubleshoot this issue - and feel free to create new Jungle 3 accounts and observe how they are returned by the v1/history plugin and not the new v1/chain plugin.

aaroncox:
To add additional context, it looks like it's any new accounts after the new API endpoint initializes. 

I performed the following steps over the weekend while experimenting more:

- Restarted nodeos w/ `--enable-account-queries=true`
- Immediately created a new account.
- Performed key lookup using `get_accounts_by_authorizers`, the account is not found. 
- Performed key lookup using `get_key_accounts`, the account name was returned.
- Awaited irreversibility, still not found via `get_accounts_by_authorizers`.
- Restarted nodeos again w/ `--enable-account-queries=true`
- Account can now be found.

apolloaotc:
{
  "account_names": []
}

nsjames:
bump, is there an ETA on this?

nickjjzhao:
Hello nsjames,

Sorry for the late response. I am actively working on this issue, and trying to fix it by next Wednesday. Please feel free to let me know if you have any other concerns. Thanks! – Jingjun

matthewdarwin:
Is suppose this issue is resolved now?

https://github.com/EOSIO/eos/pull/9382 (2.0.x)
[...]

PR #9382, opened 2020-08-10
Handle case newaccount in Account Query DB

nickjjzhao:
## Change Description

To fix issue #9278, adding code to handle case newaccount so new created accounts can be returned by calling API get_accounts_by_authorizers. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## API Changes
- [x] API Changes


jeffreyssmith2nd:
> This PR should be against develop not master

And if its something we plan to backport, an additional PR should be opened against the release/2.0.x branch

heifner:
Can we have some tests added to cover this case please.

b1bart:
> This PR should be against develop not master

The bug is against 2.0.x so, we should push to develop and backport or commit to 2.0.x and upport eitherway

b1bart:
> Can we have some tests added to cover this case please.

It looks like the original author omitted tests in general (hey, that was me).  I agree we should add them

b1bart:
I went ahead and adjusted this PR's base to `release/2.0.x`

nickjjzhao:
> Can we have some tests added to cover this case please.

Okay, I will add some tests.

nickjjzhao:
Please let me know if I need to update #9278 with current status. There is an ETA request in #9278 from a user.

nickjjzhao:
> This PR should be against develop not master
Looks like feature "Account Query DB" has not been added to the develop branch yet as per EPE-106.

plugins/chain_plugin/test/test_account_query_db.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+    produce_blocks(10);
+
+    //link aq_db to the `accepted_block` signal on the controller
+	auto c2 = control->accepted_block.connect([&](const block_state_ptr& blk) {
+        aq_db.commit_block( blk);
+	});
+
+	account_name tester_account = N(tester);
+	const auto tracePtr =  create_account(tester_account);
```
jeffreyssmith2nd, line 54: Sorry to be that guy, can we make this `trace_ptr`?

plugins/chain_plugin/test/test_account_query_db.cpp
```diff
@@ -0,0 +1,100 @@
...
+    auto c = control->accepted_block.connect([&](const block_state_ptr& blk) {
+        aq_db.commit_block( blk);
+    });
+
+	const auto& tester_account = N(tester);
+	const string role = "first";
+	produce_block();
+	create_account(tester_account);
+
+	auto tracePtr = push_action(config::system_account_name, updateauth::get_name(), tester_account, fc::mutable_variant_object()
```
jeffreyssmith2nd, line 82: Same here

plugins/chain_plugin/test/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+add_executable( test_account_query_db test_account_query_db.cpp )
+
+target_link_libraries( test_account_query_db chain_plugin eosio_testing)
+
+target_include_directories( test_account_query_db PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_SOURCE_DIR}/../../../libraries/chain/include" "${CMAKE_CURRENT_SOURCE_DIR}/../../../libraries/chainbase/include")
```
b1bart, line 5: It should not be necessary to explicitly add these include directories.  linking `chain_plugin` which depends on `eosio_chain` which depends on `chainbase` should have brought all those paths in.  Does it not?

plugins/chain_plugin/test/test_account_query_db.cpp
```diff
@@ -0,0 +1,100 @@
...
+BOOST_AUTO_TEST_SUITE(account_query_db_tests)
+
+BOOST_FIXTURE_TEST_CASE(newaccount_test, TESTER) { try {
+
+	// instantiate an account_query_db
+	auto aq_db = account_query_db(*control);
+
+    produce_blocks(10);
+
+    //link aq_db to the `accepted_block` signal on the controller
```
b1bart, line 48: Even though it will not have an effect, this should be moved before `produce_blocks(10)` to better simulate what the real integration would do.

b1bart: by this I mean the connection to the accepted block signal

plugins/chain_plugin/test/CMakeLists.txt
nickjjzhao: > It should not be necessary to explicitly add these include directories. linking `chain_plugin` which depends on `eosio_chain` which depends on `chainbase` should have brought all those paths in. Does it not?

I got some build errors earlier without adding those include directories, but now the build errors are disappeared. Will remove the include directories from the code.

plugins/chain_plugin/test/test_account_query_db.cpp
nickjjzhao: fixed.

plugins/chain_plugin/test/test_account_query_db.cpp
nickjjzhao: fixed.

plugins/chain_plugin/test/CMakeLists.txt
nickjjzhao: The unnecessary include directories have been removed.

plugins/chain_plugin/test/test_account_query_db.cpp
nickjjzhao: I wonder if there is any resource to read to better understand how the real integration works.

nickjjzhao: Rearranged the code.

PR #9383, opened 2020-08-11
Read transaction consensus fix

heifner:
## Change Description

Always pack the transaction in read_transaction as exact pack format is part of consensus and an alternative packed format could be stored in get_packed_transaction()

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [x] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/webassembly/cf_transaction.cpp
+++ b/libraries/chain/webassembly/cf_transaction.cpp
@@ -6,24 +6,35 @@ namespace eosio { namespace chain { namespace webassembly {
       if( data.size() == 0 ) return transaction_size();
 
       const packed_transaction& packed_trx = context.trx_context.packed_trx;
-      const bytes& trx =
-            packed_trx.get_compression() == packed_transaction::compression_type::none ?
-               packed_trx.get_packed_transaction() :
-               fc::raw::pack( static_cast<const transaction&>( packed_trx.get_transaction() ) );
-
+      bytes trx = fc::raw::pack( static_cast<const transaction&>( packed_trx.get_transaction() ) );
       size_t copy_size = std::min( static_cast<size_t>(data.size()), trx.size() );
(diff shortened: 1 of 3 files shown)


swatanabe-b1:
Unit test?

swatanabe-b1:
LGTM

libraries/chain/webassembly/cf_transaction.cpp
```diff
@@ -6,24 +6,35 @@ namespace eosio { namespace chain { namespace webassembly {
       if( data.size() == 0 ) return transaction_size();
 
       const packed_transaction& packed_trx = context.trx_context.packed_trx;
-      const bytes& trx =
-            packed_trx.get_compression() == packed_transaction::compression_type::none ?
-               packed_trx.get_packed_transaction() :
-               fc::raw::pack( static_cast<const transaction&>( packed_trx.get_transaction() ) );
-
+      bytes trx = fc::raw::pack( static_cast<const transaction&>( packed_trx.get_transaction() ) );
```
linhuang-blockone, line 11: Just curious, what problem does this fix? Thanks.

heifner: The packed transaction packed bytes stored in `packed_transaction::packed_trx` does not have to be the exact format that is returned by `fc::raw::pack` as long as it can be unpacked into a `transaction`. We store the original array of bytes in the `packed_transaction` however if a contract uses read_transaction it expects the format provided by `fc::raw::pack` as it might hash it to get the transaction id. So the optimization in the replaced code can't be used as the format might be different.

linhuang-blockone: Thanks Kevin for the detailed information!

PR #9387, opened 2020-08-11
PR-to-feature branch: change to log based snapshot test

b1bart:
This PR proposes to migrate the `test_compatible_versions` unit test from snapshots to a `blocks.log` based system.  This removes any dependency on the post-conditions of the methods used to construct blockchains in our tester framework. 

Instead of building a "new" chain in a way that was intended to be deterministic but rarely was in practice, the new process will load replay a small `blocks.log` file to reconstruct the HEAD state of the chain with the latest code.   It will then systematically check that this state is also recoverable from `v2`, `v3` and `v4` snapshots of that same chain.

this PR also restores the full integrity checksum based check.

the blocks.log was created using `v1.8.14` so as to have minimal requirements on the reading process.  If the minimum compatible blocks.log version exceeds the version produced by `v1.8.14` then this blocks.log may need to be migrated to a newer version.

The snapshots were created with the following versions based on this blocks.log:
- v2 : `v1.8.14` tag
- v3 : `v2.0.7` tag
- v4 :  `26e4100` commit (aka HEAD before the new snapshot was added)

the unit test now takes an additional program option:
- `--generate-snapshot-log` which will attempt to deterministically recreate the `blocks.log` file prior to running the test.  This will not overwrite the repo version automatically.  There is _no guarantee_ that the resulting chain matches the existing log but this may be useful in the future when a snapshot bump cleans out old versions and we want to start fresh

the pre-existing `--save-snapshot` still works and will be useful as a forward compatible way of generating the future snapshots for this test.  When a new version is created, the author would need to add it to the list of tested versions, run the unit-test with `--save-snapshot` then copy the file(s) from `<build>/unitttests/snapshosts/snap_v*.{bin,json}.gz` to the working tree.


b1bart:
attached are two patches (to `v1.8.14` and `v2.0.7` respectively) that created the faux unit tests which generated the blocks.log and snapshots.
[patches.zip](https://github.com/EOSIO/eos/files/5058757/patches.zip)

PR #9389, opened 2020-08-11
Remove fc::uint128_t typedef

heifner:
## Change Description

Removed `fc::uint128_t` typedef to avoid confusion with `eosio::chain::config::uint128_t` typedef.
- `fc::uint128_t` was `fc::uint128` class
- `eosio::chain::config::uint128_t` is `__uint128_t`

Removed `using namespace boost::multiprecision` to avoid pulling in `boost::multiprecision::uint128_t` which we also use.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/eosio_contract.cpp
+++ b/libraries/chain/eosio_contract.cpp
@@ -25,7 +25,7 @@ namespace eosio { namespace chain {
 
 
 uint128_t transaction_id_to_sender_id( const transaction_id_type& tid ) {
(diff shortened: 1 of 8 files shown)


heifner:
> Why did coverage test fail?

Because they are all failing. See EPE-305.

PR #9390, opened 2020-08-12
test contracts fix

dimas1185:
## Change Description

Resolving build issues with test-contracts
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/test-contracts/kv_test/kv_test.cpp
+++ b/unittests/test-contracts/kv_test/kv_test.cpp
@@ -4,32 +4,6 @@
 using namespace eosio;
 using namespace eosio::internal_use_do_not_use;
 
-enum it_stat : int32_t {
-   iterator_ok     = 0,
-   iterator_erased = -1,
-   iterator_end    = -2,
-};
-
-#define IMPORT extern "C" __attribute__((eosio_wasm_import))
-
-// clang-format off
-// IMPORT int64_t  kv_erase(uint64_t db, uint64_t contract, const char* key, uint32_t key_size);
-// IMPORT int64_t  kv_set(uint64_t db, uint64_t contract, const char* key, uint32_t key_size, const char* value, uint32_t value_size, name payer);
-// IMPORT bool     kv_get(uint64_t db, uint64_t contract, const char* key, uint32_t key_size, uint32_t& value_size);
-// IMPORT uint32_t kv_get_data(uint64_t db, uint32_t offset, char* data, uint32_t data_size);
-// IMPORT uint32_t kv_it_create(uint64_t db, uint64_t contract, const char* prefix, uint32_t size);
-// IMPORT void     kv_it_destroy(uint32_t itr);
-// IMPORT int32_t  kv_it_status(uint32_t itr);
-// IMPORT int      kv_it_compare(uint32_t itr_a, uint32_t itr_b);
-// IMPORT int      kv_it_key_compare(uint32_t itr, const char* key, uint32_t size);
-// IMPORT int32_t  kv_it_move_to_end(uint32_t itr);
-// IMPORT it_stat  kv_it_next(uint32_t itr, uint32_t* found_key_size, uint32_t* found_value_size);
-// IMPORT it_stat  kv_it_prev(uint32_t itr, uint32_t* found_key_size, uint32_t* found_value_size);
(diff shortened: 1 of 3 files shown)


unittests/test-contracts/test_api/test_datastream.cpp
```diff
@@ -74,7 +74,24 @@ void test_datastream::test_basic()
         int a[2];
         bool operator==( const StaticArray &o ) const { return a[0] == o.a[0] && a[1] == o.a[1]; }
     };
-    testtype<StaticArray>::run( {{10,20}}, "StaticArray" );
+    /**
+     * Removed test:
+     * testtype<StaticArray>::run( {{10,20}}, "StaticArray" );
```
jeffreyssmith2nd, line 79: Can we instead construct the StaticArray differently and keep the test or is brace elision the entire point of this test?

Based on that Github issue, the author has no intention on fixing it so we should either adjust the test or remove it outright.

unittests/test-contracts/kv_test/kv_test.cpp
```diff
@@ -12,21 +13,21 @@ enum it_stat : int32_t {
...
+// IMPORT int32_t  kv_it_status(uint32_t itr);
+// IMPORT int      kv_it_compare(uint32_t itr_a, uint32_t itr_b);
+// IMPORT int      kv_it_key_compare(uint32_t itr, const char* key, uint32_t size);
+// IMPORT int32_t  kv_it_move_to_end(uint32_t itr);
+// IMPORT it_stat  kv_it_next(uint32_t itr, uint32_t* found_key_size, uint32_t* found_value_size);
+// IMPORT it_stat  kv_it_prev(uint32_t itr, uint32_t* found_key_size, uint32_t* found_value_size);
+// IMPORT it_stat  kv_it_lower_bound(uint32_t itr, const char* key, uint32_t size, uint32_t* found_key_size, uint32_t* found_value_size);
+// IMPORT it_stat  kv_it_key(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size);
+// IMPORT it_stat  kv_it_value(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size);
 // clang-format on
```
jeffreyssmith2nd, line 31: This should just be deleted

unittests/test-contracts/test_api/test_datastream.cpp
dimas1185: I spent a while trying to fix that. was experimenting with different sets of constructors but got no luck.
so to not spend many time on this I just commented this out.

jeffreyssmith2nd: Yeah looking into this further, this has nothing to do with how it is initialized. It looks like its a bug in PFR when it counts the number of fields as part of the reflection. It's saying this is a size 2 when it should be a size 1.

I will spend some time investigating this, as it appears to be a possible regression in CDT.

dimas1185: Yes, it is PFR bug that is based on braces elision. how would you determine number of elements in class with braces elision feature being active? Library author is still looking for suggestions.
I tried to experiment with adding some constructors like this to restrict braces elision but no luck:
```
struct my{
   int val[2];
   my(const int (&v)[2]) { }
};
int main ()
{
   boost::pfr::for_each_field<my>({{1,2}}, [](const auto& v){ cout << "kuku" << endl;});
}
```
in case above it doesn't find any constructor at all.
 I'll ask for suggestions in blockchain channel.

unittests/test-contracts/kv_test/kv_test.cpp
```diff
@@ -2,33 +2,14 @@
 
 
 using namespace eosio;
+using namespace eosio::internal_use_do_not_use;
```
johndebord, line 5: Not sure if we should continue the practice of hiding the usage of the namespace `internal_use_do_not_use` with `using namespace eosio::internal_use_do_not_use;`.

dimas1185: @johndebord those are unit tests that are specific to intrinsics from `internal_use_do_not_use`. code will be verbose if we will keep it. I think unit tests for intrinsics are ok for this.

Issue #9255, opened 2020-06-26
state_history_plugin maybe can not distinguish old and new delta


gzliudan:
https://github.com/EOSIO/eos/blob/master/plugins/state_history_plugin/state_history_plugin.cpp#L532
```cpp
            for (auto& old : undo.old_values) {
               auto& row = index.get(old.first);
               if (include_delta(old.second, row))
                  delta.rows.obj.emplace_back(true, pack_row(row));
            }
            for (auto& old : undo.removed_values)
               delta.rows.obj.emplace_back(false, pack_row(old.second));
            for (auto id : undo.new_ids) {
               auto& row = index.get(id);
               delta.rows.obj.emplace_back(true, pack_row(row));
            }
```
The above codes use true for old records and new record, use false for removed records. But when there is no removed records,  we can't distinguish old records between new records, because all flags of records are true!

swang-b1:
I think there is block number in the row delta update, which will indicate new and old.

gzliudan:
I will investigate block number in the row delta update laster. But I think it's more clear that use a uint8_t to replace bool here, which value stards for 1=old; 0=removed; 2=new.

gzliudan:
I decoded table delta data from state history plugin, no block number in row, so use bool can't distinguish between old values and new values.

> I think there is block number in the row delta update, which will indicate new and old.

sanaraufx:
Should be fixed in 2.1

PR #9391, opened 2020-08-12
distinguish old and new table_delta in state history plugin

gzliudan:
replace bool with uint8_t to distinguish old and new table_delta in state history plugin: 0=removed; 1=old; 2=new


b1bart:
In addition to considering backwards compatibility changes such as these should be made against the `develop` branch or an appropriate `release/` branch.  `master` only contains stable tagged releases.  

This seems related to #9255 

We appreciate your submission but, as an acceptable submission will require a fair amount of re-work I am closing this PR out in favor of a future PR that can address the issues.

PR #9392, opened 2020-08-12
EPE-306 fix

huangminghuang:
## Change Description
Avoid using sizeof() for billable_size to prevent consensus problem.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/kv_chainbase_objects.hpp
+++ b/libraries/chain/include/eosio/chain/kv_chainbase_objects.hpp
@@ -34,9 +34,11 @@ namespace eosio { namespace chain {
 namespace config {
    template<>
    struct billable_size<kv_object> {
-      static constexpr uint64_t overhead = overhead_per_row_per_index_ram_bytes * 2;
-      static constexpr uint64_t serialized_kv_object_size = sizeof(kv_object) - 2*sizeof(shared_blob) + (8 + 4) * 2; ///< 8 for vector data 4 for vector size
-      static constexpr uint64_t value = serialized_kv_object_size + overhead; 
+      // NOTICE: Do not change any of the constants defined here; otherwise it would cause backward concensus compatibility problem.
+      static constexpr uint64_t overhead                     = overhead_per_row_per_index_ram_bytes * 2;
+      static constexpr uint64_t serialized_shared_blob_size  = 8 + 4; // 8 for vector data 4 for vector size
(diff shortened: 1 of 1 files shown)


larryk85:
I think this is a bit shallow of a solution.  For data types like `kv_object` where we know that they are directly impactful for consensus affecting code, we should be more diligent with these types and do a few things.  

1) Make some comments around that data type that this is a 'critical' consensus affecting data type.
2) Be diligent with these types to either be "packed" or hand laid out if we require the `sizeof` to be consistent no matter what.
3) In the struct in question, we should have `static_assert`'s that verify that the `sizeof` and `alignof` and any assumptions needed are held.

This will help us in the future.  As when we compile with either different compilers, for different OSes or architectures, it would be better to pop the asserts at compile time and not require us to rely on just not using things like `sizeof`.

b1bart:
> I think this is a bit shallow of a solution. For data types like `kv_object` where we know that they are directly impactful for consensus affecting code, we should be more diligent with these types and do a few things.
> 
> 1. Make some comments around that data type that this is a 'critical' consensus affecting data type.
> 2. Be diligent with these types to either be "packed" or hand laid out if we require the `sizeof` to be consistent no matter what.
> 3. In the struct in question, we should have `static_assert`'s that verify that the `sizeof` and `alignof` and any assumptions needed are held.
> 
> This will help us in the future. As when we compile with either different compilers, for different OSes or architectures, it would be better to pop the asserts at compile time and not require us to rely on just not using things like `sizeof`.

I agree with the general plan here but I would prefer an additional task to do this across _all_ of our billed structures as opposed to individual PRs per structure.  This PR should bring this structure into a stable state, then we can put in a new ticket/task to put in better protections.

Fair compromise?

larryk85:
> FWIW, there is no requirement that these numbers match any reality, the closer the approximate reality the more accurately we can bill things but as this is consensus we should leave some slack for alternative and future implementations to shift the details.

I agree here, the only thing I am concerned with is helping flag possible problem areas as they arise (which keeping the bounds tighter will give).  Any fake or synthetic consensus numbers should have their own checks to ensure whatever assumptions they make are held.

And yes I agree that another ticket and explicit task would be the preferable way to go.

PR #9394, opened 2020-08-12
Remove DISK mode and db specifier in host functions phase 1; add back…

linhuang-blockone:
…ing_store option to nodeos

## Change Description

This PR completes following tasks:
* Introduce ```backing-store``` option to nodoes, which at system level indicates where the state is stored (```Chainbase``` or ```Rocksdb```).
* Prepare eventual removal of ```DISK``` mode and ```db``` specifier in KV API host functions.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -91,9 +91,11 @@ void apply_context::exec_one()
          kv_iterators.resize(1);
          kv_destroyed_iterators.clear();
          if (!context_free) {
-#warning TODO: Remove kv_ram. Rename kv_disk
-            kv_ram = create_kv_chainbase_context(db, kvram_id, receiver, create_kv_resource_manager_ram(*this), control.get_global_properties().kv_configuration.kvram);
-            kv_disk = create_kv_chainbase_context(db, kvdisk_id, receiver, create_kv_resource_manager_disk(*this), control.get_global_properties().kv_configuration.kvdisk);
+            if (control.get_config().backing_store == backing_store_type::ROCKSDB) {
+               kv_backing_store = create_kv_rocksdb_context(control.kv_database(), control.kv_undo_stack(), receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+            } else {
+               kv_backing_store = create_kv_chainbase_context(db, receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+            }
 
             _db_context = create_db_chainbase_context(*this, receiver);
          }
@@ -1129,24 +1131,23 @@ int apply_context::db_end_i64( name code, name scope, name table ) {
    return keyval_cache.cache_table( *tab );
 }
 
-int64_t apply_context::kv_erase(uint64_t db, uint64_t contract, const char* key, uint32_t key_size) {
-   return kv_get_db(db).kv_erase(contract, key, key_size);
+int64_t apply_context::kv_erase(uint64_t contract, const char* key, uint32_t key_size) {
+   return kv_get_backing_store().kv_erase(contract, key, key_size);
 }
 
-int64_t apply_context::kv_set(uint64_t db, uint64_t contract, const char* key, uint32_t key_size, const char* value, uint32_t value_size, account_name payer) {
-   return kv_get_db(db).kv_set(contract, key, key_size, value, value_size, payer);
+int64_t apply_context::kv_set(uint64_t contract, const char* key, uint32_t key_size, const char* value, uint32_t value_size, account_name payer) {
+   return kv_get_backing_store().kv_set(contract, key, key_size, value, value_size, payer);
 }
 
-bool apply_context::kv_get(uint64_t db, uint64_t contract, const char* key, uint32_t key_size, uint32_t& value_size) {
-   return kv_get_db(db).kv_get(contract, key, key_size, value_size);
+bool apply_context::kv_get(uint64_t contract, const char* key, uint32_t key_size, uint32_t& value_size) {
+   return kv_get_backing_store().kv_get(contract, key, key_size, value_size);
 }
 
-uint32_t apply_context::kv_get_data(uint64_t db, uint32_t offset, char* data, uint32_t data_size) {
-   return kv_get_db(db).kv_get_data(offset, data, data_size);
+uint32_t apply_context::kv_get_data(uint32_t offset, char* data, uint32_t data_size) {
+   return kv_get_backing_store().kv_get_data(offset, data, data_size);
 }
 
-uint32_t apply_context::kv_it_create(uint64_t db, uint64_t contract, const char* prefix, uint32_t size) {
-   auto& kdb = kv_get_db(db);
+uint32_t apply_context::kv_it_create(uint64_t contract, const char* prefix, uint32_t size) {
    uint32_t itr;
    if (!kv_destroyed_iterators.empty()) {
       itr = kv_destroyed_iterators.back();
@@ -1157,7 +1158,7 @@ uint32_t apply_context::kv_it_create(uint64_t db, uint64_t contract, const char*
       itr = kv_iterators.size();
       kv_iterators.emplace_back();
    }
-   kv_iterators[itr] = kdb.kv_it_create(contract, prefix, size);
+   kv_iterators[itr] = kv_get_backing_store().kv_it_create(contract, prefix, size);
    return itr;
 }
 
@@ -1211,14 +1212,6 @@ int32_t apply_context::kv_it_key(uint32_t itr, uint32_t offset, char* dest, uint
 int32_t apply_context::kv_it_value(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size) {
    kv_check_iterator(itr);
    return static_cast<int32_t>(kv_iterators[itr]->kv_it_value(offset, dest, size, actual_size));
-}
-
-kv_context& apply_context::kv_get_db(uint64_t db) {
-   if (db == kvram_id.to_uint64_t())
-      return *kv_ram;
-   else if (db == kvdisk_id.to_uint64_t())
-      return *kv_disk;
-   EOS_ASSERT(false, kv_bad_db_id, "Bad key-value database ID");
 }
 
 void apply_context::kv_check_iterator(uint32_t itr) {
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -91,10 +91,15 @@ void apply_context::exec_one()
          kv_iterators.resize(1);
          kv_destroyed_iterators.clear();
          if (!context_free) {
-            if (control.get_config().backing_store == backing_store_type::ROCKSDB) {
-               kv_backing_store = create_kv_rocksdb_context(control.kv_database(), control.kv_undo_stack(), receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
-            } else {
-               kv_backing_store = create_kv_chainbase_context(db, receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+            switch (control.get_config().backing_store) {
+               case backing_store_type::ROCKSDB:
+                  kv_backing_store = create_kv_rocksdb_context(control.kv_database(), control.kv_undo_stack(), receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+                  break;
+               case backing_store_type::NATIVE:
+                  kv_backing_store = create_kv_chainbase_context(db, receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+                  break;
+               default:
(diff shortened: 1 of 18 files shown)


libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -585,6 +585,9 @@ class apply_context {
       int32_t  kv_it_lower_bound(uint32_t itr, const char* key, uint32_t size, uint32_t* found_key_size, uint32_t* found_value_size);
       int32_t  kv_it_key(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size);
       int32_t  kv_it_value(uint32_t itr, uint32_t offset, char* dest, uint32_t size, uint32_t& actual_size);
+      kv_context& kv_get_backing_store() const {
```
TimothyBanks, line 588: Is kv_context an immutable type?  Or should we return a const reference since this method is marked const?

libraries/chain/include/eosio/chain/apply_context.hpp
```diff
@@ -585,6 +585,9 @@ class apply_context {
...
+         return *kv_backing_store;
```
TimothyBanks, line 589: Is this guaranteed to be non null?

libraries/chain/include/eosio/chain/apply_context.hpp
linhuang-blockone: It is a mutable type. Could not return a const reference.

libraries/chain/include/eosio/chain/apply_context.hpp
linhuang-blockone: Thanks for spotting this! In theory kv_backing_store should have been created before kv_get_backing_store is called, but we cannot prevent it from being called prematurely. Will add an assert to check the pointer.

libraries/chain/include/eosio/chain/apply_context.hpp
TimothyBanks: Then should this method be invocable on a const instance?

linhuang-blockone: This method is invoked on apply_context, which is not constant. Its use looks like "int64_t apply_context::kv_set(...) {
 return kv_get_backing_store().kv_set(...);" 

If I returned a const reference, I got a compile error:

/home/lin/Work/kv/eos/libraries/chain/apply_context.cpp:1140:11: error:
'this' argument to member function 'kv_set' has type 'const
      eosio::chain::kv_context', but function is not marked const
   return kv_get_backing_store().kv_set"

TimothyBanks: What I was thinking is that if the method is marked const, then we might want to return a const refererence.  Otherwise, if the method doesn't need to be constant, then can we remove the const from the method?

linhuang-blockone: Oh I see your point. I'll remove the const. Thanks.

unittests/kv_tests.cpp
```diff
@@ -40,7 +40,6 @@ FC_REFLECT(itparam, (db)(count)(erase))
 class kv_tester : public tester {
  public:
```
jeffreyssmith2nd, line 41: Should all the references to DISK in these tests be removed?

linhuang-blockone: As the test contracts are still in the old format, those references to DISK cannot be removed for now.
We are doing the removal in phases. They will be removed in next phase when focusing on tests,
or removed in the final phase when CDT removes DISK.

unittests/kv_tests.cpp
```diff
@@ -832,13 +779,14 @@ BOOST_FIXTURE_TEST_CASE(kv_resource_limit, kv_tester) try { //
...
 FC_LOG_AND_RETHROW()
+*/
 
 BOOST_FIXTURE_TEST_CASE(kv_key_value_limit, kv_tester) try { //
    test_kv_key_value_limit_common();
 }
 FC_LOG_AND_RETHROW()
 
-constexpr name databases[] = { N(eosio.kvram), N(eosio.kvdisk) };
+constexpr name databases[] = { N(eosio.kvdisk) };
```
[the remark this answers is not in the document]
brianjohnson5972: If we are just using 1 database, then can we remove "databases" and just use N(eosio.kvram) everywhere it was?

unittests/kv_tests.cpp
```diff
@@ -701,90 +688,59 @@ class kv_tester : public tester {
...
    void test_kv_key_value_limit_common() { //
-      test_key_value_limit(N(eosio.kvram));
       test_key_value_limit(N(eosio.kvdisk));
    }
 
    abi_serializer abi_ser;
    abi_serializer sys_abi_ser;
 };
 
+class kv_rocksdb_tester : public kv_tester {
```
brianjohnson5972, line 734: get rid of this class and add get_config to kv_tester constructor and pass in backing_store, then you can use BOOST_DATA_TEST_CASE_F to pass in CHAINBASE and ROCKSDB as the dataset (instead of databases, which isn't needed anymore).

brianjohnson5972: also, we won't need all these helper methods anymore, so move the BOOST_.. macros instead of the functions that surround the implemenation.

unittests/snapshot_tests.cpp
```diff
@@ -663,11 +663,15 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_kv_snapshot, SNAPSHOT_SUITE, snapshot_suites)
    //for (bool rocks_save : { false, true }) {
    //   for (bool rocks_load : { false, true }) {
 
+   // snapshot handling is being rewritten.
```
brianjohnson5972, line 666: put a #warning here so that we don't forget to turn it back on

unittests/snapshot_tests.cpp
```diff
@@ -697,13 +701,17 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_kv_snapshot, SNAPSHOT_SUITE, snapshot_suites)
             auto snapshot = SNAPSHOT_SUITE::finalize(writer);
 
             auto cfg = chain.get_config();
-            cfg.use_rocksdb_for_disk = rocks_load;
+            if (rocks_load) {
```
brianjohnson5972, line 704: It looks like rocks_save isn't needed anymore, right?  It also looks like rocks_load will only be true, shouldn't the for loop be:
for (bool rocks_load : { true, false }) {

unittests/snapshot_tests.cpp
```diff
@@ -697,13 +701,17 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_kv_snapshot, SNAPSHOT_SUITE, snapshot_suites)
...
+            } else {
+               cfg.backing_store = eosio::chain::backing_store_type::CHAINBASE;
+            }
             // create a new child at this snapshot
             sub_testers.emplace_back(cfg, SNAPSHOT_SUITE::get_reader(snapshot), generation);
 
             // increment the test contract
             signed_transaction trx;
-            trx.actions.push_back({{{N(snapshot), N(active)}}, N(snapshot), N(eosio.kvram), {}});
+            //trx.actions.push_back({{{N(snapshot), N(active)}}, N(snapshot), N(eosio.kvram), {}});
```
brianjohnson5972, line 714: remove the next line (with kvdisk) and un-comment this line

libraries/chain/apply_context.cpp
```diff
@@ -91,9 +91,11 @@ void apply_context::exec_one()
...
          kv_destroyed_iterators.clear();
          if (!context_free) {
-#warning TODO: Remove kv_ram. Rename kv_disk
-            kv_ram = create_kv_chainbase_context(db, kvram_id, receiver, create_kv_resource_manager_ram(*this), control.get_global_properties().kv_configuration.kvram);
-            kv_disk = create_kv_chainbase_context(db, kvdisk_id, receiver, create_kv_resource_manager_disk(*this), control.get_global_properties().kv_configuration.kvdisk);
+            if (control.get_config().backing_store == backing_store_type::ROCKSDB) {
+               kv_backing_store = create_kv_rocksdb_context(control.kv_database(), control.kv_undo_stack(), receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+            } else {
+               kv_backing_store = create_kv_chainbase_context(db, receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
+            }
```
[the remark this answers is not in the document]
linhuang-blockone: Thanks for pointing out explicit check. Will do that.

libraries/chain/controller.cpp
```diff
@@ -2661,6 +2648,11 @@ uint32_t controller::get_max_nonprivileged_inline_action_size()const
    return my->conf.max_nonprivileged_inline_action_size;
 }
 
+const controller::config& controller::get_config()const
```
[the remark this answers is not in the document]
linhuang-blockone: To be consistent with the style in the surrounded existing code, no space was added.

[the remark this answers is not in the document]
linhuang-blockone: This function is used by apply_context::exec_one too.

unittests/kv_tests.cpp
```diff
@@ -646,27 +643,17 @@ class kv_tester : public tester {
...
-      BOOST_TEST(iterlimit({{N(eosio.kvram), 4, false}, {N(eosio.kvdisk), 6, false},
-                            {N(eosio.kvram), 2, true}, {N(eosio.kvdisk), 2, true},
-                            {N(eosio.kvram), 3, false}, {N(eosio.kvdisk), 3, false}}) == "");
-      BOOST_TEST(iterlimit({{N(eosio.kvram), 4, false}, {N(eosio.kvdisk), 6, false},
-                            {N(eosio.kvram), 2, true}, {N(eosio.kvdisk), 2, true},
-                            {N(eosio.kvram), 3, false}, {N(eosio.kvdisk), 4, false}}) == "Too many iterators");
-      BOOST_TEST(iterlimit({{N(eosio.kvram), 4, false}, {N(eosio.kvdisk), 6, false},
-                            {N(eosio.kvram), 2, true}, {N(eosio.kvdisk), 2, true},
-                            {N(eosio.kvram), 4, false}, {N(eosio.kvdisk), 3, false}}) == "Too many iterators");
+      BOOST_TEST(iterlimit({{N(eosio.kvdisk), 6, false},
```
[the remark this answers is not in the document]
linhuang-blockone: All references to kvdisk will be removed at the final phase of removing DISK mode. It depends on removing 
database ID in the test contract. Jeff will do that.

unittests/kv_tests.cpp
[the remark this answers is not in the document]
linhuang-blockone: They will be removed.

unittests/kv_tests.cpp
```diff
@@ -872,106 +820,84 @@ FC_LOG_AND_RETHROW()
...
-BOOST_FIXTURE_TEST_CASE(kv_iterase_rocksdb, kv_tester) try { //
-   use_rocksdb();
+BOOST_FIXTURE_TEST_CASE(kv_iterase_rocksdb, kv_rocksdb_tester) try { //
    test_kv_iterase_common();
 }
 FC_LOG_AND_RETHROW()
 
-BOOST_FIXTURE_TEST_CASE(kv_ram_usage_rocksdb, kv_tester) try { //
-   use_rocksdb();
+  /*
```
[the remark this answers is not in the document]
linhuang-blockone: All the unit tests will be cleaned up at the final phase.

unittests/kv_tests.cpp
linhuang-blockone: Thanks for letting me know about those techniques. I will utilize them in future tests.

unittests/snapshot_tests.cpp
linhuang-blockone: Will do.

unittests/snapshot_tests.cpp
linhuang-blockone: Will do.

unittests/snapshot_tests.cpp
linhuang-blockone: This will be removed in final phase.

unittests/kv_tests.cpp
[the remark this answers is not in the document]
brianjohnson5972: So why did we remove eosio.kvram and not eosio.kvdisk?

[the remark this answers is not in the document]
linhuang-blockone: We could remove eosio.kvdisk and keep eosio.kvram for now
But since in original design, only kvdisk would trigger RocksDB.
That's why I kept kvdisk for easier testing. It will be removed once contract and CDT are done.

PR #9400, opened 2020-08-13
[docs] add return value from actions cleos output explanation and samples

iamveritas:
[docs] add return value from actions cleos output explanation and samples in cleos reference documentation

### DIFF
--- a/docs/02_cleos/03_command-reference/push/push-action.md
+++ b/docs/02_cleos/03_command-reference/push/push-action.md
@@ -34,7 +34,25 @@ Push a transaction with a single action
 
 For actions that return a value `cleos` will have an extra line which will be prefixed with `=>`.
 
-Below you can see three examples of pushing actions `rstring` and `ruint` which return the value passed as input parameter to them:
+Let's consider for exemplification the three actions defined below. The first two are `rstring` and `ruint`, which return the value passed as input parameter to them, and the third one is `rtmp` returning a default `tmp` object.
+
+```c++
+struct tmp {
+    uint32_t id = 1;
+    std::vector<uint8_t> list{ 1, 2 };
+};
+
+[[eosio::action]]
+std::string rstring(const std::string & str) { return str;   }
+
+[[eosio::action]]
+uint16_t    ruint(uint16_t i)                { return i;     }
+
+[[eosio::action]]
+tmp         rtmp()                           { return tmp{}; }
+```
+
+Below you can see examples of `cleos` commands pushing each action and their respective outputs.
 
 ```shell
 > cleos push action eosio rstring '{"str":null}'  -p eosio@active
@@ -56,18 +74,7 @@ executed transaction: 12c51fd27fad9bb0fa959037a72093ad0b168f5846e916a0e9a295c841
 =>                                return value: 42
 ```
 
-It can also work with C++ structures or classes as well. For a C++ struct like this:
-
-```cpp
-struct tmp {
-    uint32_t id = 1;
-    std::vector<uint8_t> list{ 1, 2 };
-
-    EOSLIB_SERIALIZE(tmp, (id)(list))
-};
-```
-
-and the action `rtmp` returning a default tmp object, output will look like this:
+The `rtmp` action output, returning a default `tmp` object, output will look like this:
 
 ```shell
 > cleos push action eosio rtmp '{}'  -p eosio@active
@@ -77,7 +84,7 @@ executed transaction: 42386687d94695d77d67cf901cf8d1a4f83ac9f89382f0b01f3fb51267
 =>                                return value: {"id":1,"list":[1,2]}
 ```
 
-When it will not be possible to decode the return value, output will look like this:
+When it will not be possible to decode the return value, for example when ABI doesn’t contain information about an action return value, output will look like this:
 
 ```shell
 =>                          return value (hex): 1400000001000770656e64696e6700000000
--- a/docs/02_cleos/03_command-reference/push/push-action.md
+++ b/docs/02_cleos/03_command-reference/push/push-action.md
@@ -32,7 +32,7 @@ Push a transaction with a single action
 
 ## Output
 
-For actions that return a value `cleos` will have an extra line which will be prefixed with `=>`.
+For actions that return a value `cleos` has a line which is prefixed with `=>`.
 
 Let's consider for exemplification the three actions defined below. The first two are `rstring` and `ruint`, which return the value passed as input parameter to them, and the third one is `rtmp` returning a default `tmp` object.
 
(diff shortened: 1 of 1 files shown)


docs/02_cleos/03_command-reference/push/push-action.md
```diff
@@ -31,3 +29,65 @@ Push a transaction with a single action
...
+```
+
+It can also work with C++ structures or classes as well. For a C++ struct like this:
+
+```cpp
+struct tmp {
+    uint32_t id = 1;
+    std::vector<uint8_t> list{ 1, 2 };
+
+    EOSLIB_SERIALIZE(tmp, (id)(list))
```
Max-B1, line 66: I would empathize the fact that EOSLIB_SERIALIZE() is a must.

docs/02_cleos/03_command-reference/push/push-action.md
```diff
@@ -31,3 +29,65 @@ Push a transaction with a single action
...
+
+```shell
+> cleos push action eosio rtmp '{}'  -p eosio@active
+
+executed transaction: 42386687d94695d77d67cf901cf8d1a4f83ac9f89382f0b01f3fb51267cc21cd  96 bytes  110 us
+#         eosio <= eosio::rtmp                  ""
+=>                                return value: {"id":1,"list":[1,2]}
+```
+
+When it will not be possible to decode the return value, output will look like this:
```
Max-B1, line 80: _> When it will not be possible to decode the return value_
@iamveritas You can add an example: ABI doesn't contain information about an action return value.

docs/02_cleos/03_command-reference/push/push-action.md
iamveritas: I've asked Larry and he told me it is not necessary, how sure are you?

Max-B1: inviting @larryk85 to the thread

docs/02_cleos/03_command-reference/push/push-action.md
```diff
@@ -31,3 +29,65 @@ Push a transaction with a single action
 `--max-net-usage` _UINT_ - set an upper limit on the net usage budget, in bytes, for the transaction (defaults to 0 which means no limit)
 
 `--delay-sec` _UINT_ - set the delay_sec seconds, defaults to 0s
+
+## Output
+
+For actions that return a value `cleos` will have an extra line which will be prefixed with `=>`.
+
+Below you can see three examples of pushing actions `rstring` and `ruint` which return the value passed as input parameter to them:
```
Max-B1, line 37: here is the code of the actions:
```
[[eosio::action]]
std::string rstring(const std::string & str) { return str;   }

[[eosio::action]]
uint16_t    ruint(uint16_t i)                { return i;     }

[[eosio::action]]
tmp         rtmp()                           { return tmp{}; }
```
It might be useful to include it as well

docs/02_cleos/03_command-reference/push/push-action.md
larryk85: EOSLIB_SERIALIZE is not a must.  It helps in terms of compile time for datastream operations and is a must for "some" user defined types that PFR twirls into infinite on.  We will punt back to Boost::PFR if you don't add that macro and we will replace that with something even better before 2.2, so it's not a requirement for most data types.  We mainly used it for the in library stuff to save on compile time for the end user, but for their contracts the small amount of data types they have is negligible.

iamveritas: @larryk85 
how can the users know when they fall into that case of "some" user defined type?
you are mentioning an infinite loop?... will the execution of the action using this specific user defined type brake?
I still think we should mention this case, define it, so if anyone hits it they will know how to get around it: explicitly use the EOSLIB_SERIALIZE

larryk85: I would rather not mention it.  It is a macro that will be going away very soon.  The macro doesn't help the cases that I was thinking about anyway.  PFR will fail on inherited types or types that aren't trivially constructible, which means they would have to define the datastream operators by hand anyway.

iamveritas: ok so I will not mention the macro, I will even take it out from the example code in this doc.

but, I realize now that saying "any user defined type can be returned from an action" is not actually accurate. I'd like to make this statement more accurate, something like this: you can return almost any user defined type, however there are some exceptions, those which are not trivially constructible, for which you'll have to define custom the datastream operators to be able to serialize them correctly.

When will the user know they are in trouble with the user defined type they want to return which will need custom datastream operators? The serialization will fail? Will be any error? Will the return value just be empty, not present?

docs/02_cleos/03_command-reference/push/push-action.md
iamveritas: I will change, thank you.

docs/02_cleos/03_command-reference/push/push-action.md
iamveritas: thanks, done.

docs/02_cleos/03_command-reference/push/push-action.md
larryk85: But, that restriction has nothing to do with returning a value.  That is a general restriction of datastream.  It will fail at compile time.  We have or should have all of this documented with the datastream system.

iamveritas: I took out the macro from this doc page.
the rest of the discussion pertains to the return values from action explainer page, so I will continue it there.

halsaphi (commented):
For actions that return a value `cleos` will have an extra line which will be prefixed with `=>`.
-to-
For actions that return a value `cleos` now has a line which is prefixed with `=>`.

halsaphi (commented):
Let's consider for exemplification the three actions defined below
-to-
In the following three actions:
rstring  - returns the std::string input parameter.
ruint - returns the uint16_t input parameter. 
rtmp - returns a default tmp object.

Below you can see examples of cleos commands pushing each action and their respective outputs.
-to-
The cleos commands and the cleos outputs are:
...

halsaphi:
The rtmp action output, returning a default tmp object, output will look like this:
-to-
The rtmp action returns a default tmp object as json:

halsaphi:
When it will not be possible to decode the return value, for example when ABI doesn’t contain information about an action return value, output will look like this:
- to -
The rtmp action output when no ABI is available to decode the return value will look like this:

halsaphi:
If the length of the return value string is more than 100 chars the string will be truncated to the first 100 chars and three dots ... will be added at the end. It's similar to what cleos does to action parameters output.
-to -
If the length of the return value string is more than 100 chars the string will be truncated to the first 100 chars and three dots ... will be added at the end.

halsaphi (approved):

PR #9401, opened 2020-08-13
set max_irreversible_block_age to -1

allenhan2:
## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

plugin_http_api_test may hang due to pause and resume via producer api, set max_irreversible_block_age to -1 to bypass irreversible block is too old check.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


allenhan2:
The PR change does not require code coverage analysis, so, the code coverage timeout fail is irrelevant to this PR.

PR #9402, opened 2020-08-14
Extensible Blockchain Parameters

dimas1185:
## Change Description

**Extensible Blockchain Parameters**

**Background**

The basic means of manipulating consensus parameters for a blockchain has been a pair of intrinsic functions:

- `get_blockchain_parameters_packed`
- `set_blockchain_parameters_packed`

These intrinsics are tied to a specific and inflexible definition of what blockchain parameters are and included no convenient means to "version" that set of parameters so that parameters could be added/removed/modified in future consensus upgrades.

**Task Description**

This PR is to implement a new pair of intrinsics that is intended to eventually supplant the existing intrinsics and provide greater flexibility for future consensus upgrades.

As this task will add intrinsics it **MUST** be a consensus upgrade itself which, when activated, will allow contracts to link to the new intrinsics.   There is a system present in the `chain` library under `whitelisted_intrinsics.cpp|hpp` to facilitate additions like these. 

This PR support **only** the blockchain parameters that are already accessible via `get|set_blockchain_parameters_packed`.  Future tasks will add their associated parameters in conjunction with those consensus upgrades.

`set_parameters_packed( span<const char> packed_parameters )`

This is a new **privileged** intrinsic that allows a system contract the ability to set parameters in a flexible manner.  

The input buffer is a packed data stream which represents an encoded sequence of parameter_id:paramter_value pairs with the following format:

`|varuint32:sequence_length | varuint32:parameter_id | <various>:parameter_value | ...`

The encoding of parameter_values should be specific to the parameter being set

**Notes**

- It is illegal to have duplicate parameter_ids encoded in the sequence and should result in aborting the transaction context.
- Presence of a parameter_id which is unknown OR which is known but tied to an unactivated consensus protocol is illegal and should result in aborting the transaction context.
  - future consensus upgrades will need the ability to "add" valid protocol_ids 
- There is to be no requirement for ordering of items in the sequence.

`uint32_t get_parameters_packed( span<const char> packed_parameter_ids, span<char> packed_parameters)`

This is a new **privileged** intrinsic that allows a system contract the ability to get the current values of parameters in a flexible manner.  

The input buffer is a packed data stream which represents an encoded sequence of **parameter_id** pairs with the following format:

`|varuint32:sequence_length | varuint32:parameter_id | ...`

The output buffer is a packed data stream which represents an encoded sequence of parameter_id:paramter_value pairs with the following format:

`|varuint32:sequence_length | varuint32:parameter_id | <various>:parameter_value | ...`

The encoding of `parameter_value`s should be specific to the parameter being set

note the output buffer format should be valid input for **set_parameters_packed**

For each known parameter_id in the input sequence there should be an associated entry in the output sequence with the current encoded parameter_value

**Notes**

- It is illegal to have duplicate `parameter_id` s encoded in the input sequence and should result in aborting the transaction context.
- Presence of a `parameter_id` which is unknown OR which is known but tied to an unactivated consensus protocol is legal but should not result in additional entries in the output sequence.
  - future consensus upgrades will need the ability to "add" valid `protocol_id`s 
  - This means that the output sequence length can be less than the input sequence length
- There is to be no requirement or guarantee for ordering of items in the input OR output sequence.
- It is illegal to pass in a non-zero sized output buffer that cannot completely contain the output and should result in aborting the transaction context
  - If callers want to determine the required output size at runtime, they should pass in a zero-sized output buffer and the return value of the call should represent the smallest valid size for a buffer.

**New protocol feature**
This PR adds new protocol feature called `BLOCKCHAIN_PARAMETERS`. Protocol feature enables above mentioned intrinsics for privileged contracts.

**Acceptance Criteria**

In addition to presence of the above intrinsics the following expectations should be met:
- unittests guaranteeing that existing system contracts are valid and continue to work before and after the activation of this protocol feature
- unittests that demonstrate that calling setcode with WASM that imports these intrinsics fails prior to the activation of this consensus upgrade and succeeds immediately after activation.
- uinttests that demonstrate that parameters written with the new intrinsics can be read with the old intrinsics and vice versa

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

## Consensus Changes
- [x] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


libraries/chain/include/eosio/chain/chain_config_helper.hpp
```diff
@@ -0,0 +1,186 @@
...
+
+#include <boost/hana/map.hpp>                         /* make_map */
+#include <boost/hana/pair.hpp>                        /* make_pair */
+#include <boost/hana/type.hpp>                        /* type_c */
+#include <boost/hana/find.hpp>                        /* find */
+#include <boost/preprocessor/seq/for_each.hpp>        /* BOOST_PP_SEQ_FOR_EACH */
+#include <boost/preprocessor/seq/for_each_i.hpp>      /* BOOST_PP_SEQ_FOR_EACH_I */
+#include <boost/preprocessor/seq/size.hpp>            /* BOOST_PP_SEQ_SIZE */
+#include <boost/preprocessor/seq/elem.hpp>            /* BOOST_PP_SEQ_ELEM */
+#include <boost/preprocessor/repetition/repeat.hpp>   /* BOOST_PP_REPEAT */
```
heifner, line 14: Do we really need to add all these macros? Can't we just pack/unpack manually. Seems like a maintenance mess.

heifner:
> Please change the title to something clearer and provide better details in the description. These are used to generate the Release Notes, it's helpful if they are descriptive.

I would like to add the details here are likely to long outlive whatever is in Jira when we dump Jira and move to something else.

libraries/chain/include/eosio/chain/chain_config_helper.hpp
dimas1185: @heifner tbh, we have a lot of metaprogramming & macro boiler plates in our code (e.g. fc reflect) and I was not so experienced with this stuff so I decided to train a bit to get more familiarity with this in future. I had plans to make some kind of FC_REFLECT_RANGE to make it generic but later I put this in config helper.
So since this code is not going to be reusable like FC_REFLECT we can put there just resolved macros.
I'm really 50/50 on this. Either works for me.

heifner: The contract also has to do something similar, right? I assume we are not going to have all of this for the contract code.

dimas1185: @heifner CDT&contracts part is not implemented yet but you can check the unit tests:
unittests/test-contracts/params_test/params_test.cpp

dimas1185:
> Please change the title to something clearer and provide better details in the description. These are used to generate the Release Notes, it's helpful if they are descriptive.

@heifner @jeffreyssmith2nd description was update.
when you say title, do you mean inline description title or commit message?

heifner:
> @heifner @jeffreyssmith2nd description was updated.
> when you say title, do you mean inline description title or commit message?

The PR title: "new v1 config, intrinsics, UT"

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -219,6 +219,17 @@ Allows privileged contracts to set the constraints on WebAssembly code.
...
+            "BLOCKCHAIN_PARAMETERS",
+            fc::variant("70787548dcea1a2c52c913a37f74ce99e6caae79110d7ca7b859936a0075b314").as<digest_type>(),
+            {}
+         } )
+            // SHA256 hash of the raw message below within the comment delimiters (do not modify message below).
+/*
+Builtin protocol feature: BLOCKCHAIN_PARAMETERS
+
+Allows privileged contracts to get and set subsets of blockchain parameters.
+*/
```
heifner, line 232: Add information about this protocol feature to the PR description.

libraries/chain/include/eosio/chain/chain_config_helper.hpp
```diff
@@ -0,0 +1,186 @@
...
+DEFINE_ENUM_DERIVED(void, CLASS, MEMBERS)
+
+/**
+ * Returns field in in the structure
+ * This is enum-kind implementation.
+ * See DEFINE_FIELD_ID macro for implementation
+ */
+template<auto T>
+constexpr size_t field_id(){
+   FC_THROW_EXCEPTION(config_parse_error, "only specializations of field_id can be used");
```
jeffreyssmith2nd, line 113: Is there a way these could be static_asserts instead, making it a compile time failure?

libraries/chain/include/eosio/chain/chain_config_helper.hpp
jeffreyssmith2nd: I think I'm in agreement with Kevin here. While a bit tedious, maintaining manual pack/unpack would likely be easier than trying to maintain moderately complex macros.

jeffreyssmith2nd: If you want to make it easy to add the manual code, you could probably run just the preprocessor step and then copy/paste that to help you avoid some typing.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -202,27 +202,62 @@ namespace eosio { namespace chain { namespace webassembly {
...
+      config_range config_range(cfg, {context.control});
+
+      fc::raw::unpack(ds, config_range);
+      
+      config_range.config.validate();
+      context.db.modify( context.control.get_global_properties(),
+         [&]( auto& gprops ) {
+              gprops.configuration = config_range.config;
+      });
+   }
```
heifner, line 260: See comments in slack.

libraries/chain/protocol_feature_manager.cpp
dimas1185: added

libraries/chain/include/eosio/chain/chain_config_helper.hpp
dimas1185: @heifner @jeffreyssmith2nd please check my recent commit

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -115,3 +192,258 @@ FC_REFLECT(eosio::chain::chain_config,
...
+ * that is used as part of packing data_range<chain_config_v0, ...>
+ * @param s datastream
+ * @param entry contains config reference and particular id
+ * @throws config_parse_error if id is unknown
+ */
+template <typename DataStream>
+inline DataStream &operator<<(DataStream &s, const eosio::chain::data_entry<eosio::chain::chain_config_v0, eosio::chain::config_entry_validator> &entry){
+   using namespace eosio::chain;
+
+   if (!entry.is_allowed())
```
jeffreyssmith2nd, line 212: Should we consider some sort of logging here to let users know they tried to read something that doesn't exist?

unittests/test-contracts/params_test/params_test.cpp
```diff
@@ -0,0 +1,238 @@
...
+      ASSERT_EQ(old_way_params.max_authority_depth, 10);
+
+      params_object(17_ui) (0_ui)(1024*1024*4_64)
+                           (1_ui)(35*100_32)
+                           (2_ui)(1024*1024*2_32)
+                           (3_ui)(1024_32)
+                           (4_ui)(700_32)
+                           (5_ui)(50_32)
+                           (6_ui)(120_32)
+                           (7_ui)(250'000_32)
```
jeffreyssmith2nd, line 155: Did you determine if these are encoded correctly and just shown weird by Github?

dimas1185: yes, it just weird Github highlight.
I checked, this is regular single quote:
Char | Dec | Hex | Oct
-- | -- | -- | --
' | 39 | 27 | 47

libraries/chain/include/eosio/chain/chain_config.hpp
dimas1185: this part will be tested after adding first parameter into v1 config. I'll add logs, good idea.

PR #9405, opened 2020-08-16
added unit tests

Max-B1:
## Change Description

Added 3 unit tests that Emory forgot.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/abi_serializer.hpp
+++ b/libraries/chain/include/eosio/chain/abi_serializer.hpp
@@ -507,11 +507,9 @@ namespace impl {
             if (abi.valid()) {
                auto type = abi->get_action_result_type(act.name);
                if (!type.empty()) {
-                  try {
-                     binary_to_variant_context _ctx(*abi, ctx, type);
-                     _ctx.short_path = true; // Just to be safe while avoiding the complexity of threading an override boolean all over the place
-                     mvo( "return_value_data", abi->_binary_to_variant( type, act_trace.return_value, _ctx ));
-                  } catch(...) {}
+                  binary_to_variant_context _ctx(*abi, ctx, type);
+                  _ctx.short_path = true; // Just to be safe while avoiding the complexity of threading an override boolean all over the place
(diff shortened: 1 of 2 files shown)


unittests/abi_tests.cpp
```diff
@@ -2835,4 +2835,172 @@ BOOST_AUTO_TEST_CASE(serialize_optional_struct_type)
...
+}
+
+BOOST_AUTO_TEST_CASE(abi_to_variant__add_action__no_return_value)
+{
+   action_trace at;
+   std::string expected_json;
+   std::tie(at, expected_json) = generate_action_trace();
+
+   auto abi = R"({
+      "version": "eosio::abi/1.0",
```
kimjh2005, line 2978: is the version 1.0?

Max-B1: I just copied it from other tests. Not sure what is the difference between 1.0 and 1.2

kimjh2005: Does abi/1.0 has action return type?

Max-B1: `abi_to_variant__add_action__good_return_value` works with return values just fine

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -507,11 +507,9 @@ namespace impl {
             if (abi.valid()) {
                auto type = abi->get_action_result_type(act.name);
                if (!type.empty()) {
-                  try {
-                     binary_to_variant_context _ctx(*abi, ctx, type);
-                     _ctx.short_path = true; // Just to be safe while avoiding the complexity of threading an override boolean all over the place
-                     mvo( "return_value_data", abi->_binary_to_variant( type, act_trace.return_value, _ctx ));
-                  } catch(...) {}
```
johndebord, line 514: What was the reasoning for this code to have a try/catch block with the catch being a no-op? Was _special-case_ reason for that?

PR #9406, opened 2020-08-17
Implementation of a priority queue based iterator on Session

TimothyBanks:
## Change Description
This PR provides an implementation of a recursive iterator for the session type.  Details to follow...

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain_kv/include/b1/session/bytes.hpp
+++ b/libraries/chain_kv/include/b1/session/bytes.hpp
@@ -255,10 +255,10 @@ public:
             return false;
         }
         if (lhs == eosio::session::bytes::invalid) {
-            return false;
+            return true;
         }
         if (rhs == eosio::session::bytes::invalid) {
-            return true;
+            return false;
         }
 
         return std::string_view{reinterpret_cast<const char*>(lhs.data()), lhs.length()} < std::string_view{reinterpret_cast<const char*>(rhs.data()), rhs.length()};
@@ -273,10 +273,10 @@ public:
             return false;
         }
         if (lhs == eosio::session::bytes::invalid) {
-            return false;
+            return true;
         }
         if (rhs == eosio::session::bytes::invalid) {
-            return true;
+            return false;
         }
 
         return std::string_view{reinterpret_cast<const char*>(lhs.data()), lhs.length()} > std::string_view{reinterpret_cast<const char*>(rhs.data()), rhs.length()};
--- a/libraries/chain_kv/include/b1/session/bytes.hpp
+++ b/libraries/chain_kv/include/b1/session/bytes.hpp
@@ -251,6 +251,16 @@ template <>
 struct less<eosio::session::bytes> final {
 public:
     bool operator()(const eosio::session::bytes& lhs, const eosio::session::bytes& rhs) const {
+        if (lhs == eosio::session::bytes::invalid && rhs == eosio::session::bytes::invalid) {
+            return false;
+        }
+        if (lhs == eosio::session::bytes::invalid) {
+            return false;
+        }
+        if (rhs == eosio::session::bytes::invalid) {
+            return true;
+        }
+
         return std::string_view{reinterpret_cast<const char*>(lhs.data()), lhs.length()} < std::string_view{reinterpret_cast<const char*>(rhs.data()), rhs.length()};
     };
 };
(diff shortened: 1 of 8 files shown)


libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -842,92 +929,300 @@ const std::shared_ptr<const cache_data_store>& session<persistent_data_store, ca
...
+      it.current = it.end;
+      --it.current;
+    };
 
-template <typename persistent_data_store, typename cache_data_store, typename iterator_traits>
-using session_iterator_alias = typename session<persistent_data_store, cache_data_store>::template session_iterator<iterator_traits>;
+    // Prime the queue.
+    auto top = m_previous.top();
+    m_previous.pop();
+    m_previous.push(top);
```
revl, line 1192: I'm not sure I understand what this was supposed to do. Seems like a no-op.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -842,92 +929,300 @@ const std::shared_ptr<const cache_data_store>& session<persistent_data_store, ca
...
+using session_iterator_alias = typename session<persistent_data_store, cache_data_store>::template session_iterator<iterator_traits>;
+
 template <typename persistent_data_store, typename cache_data_store>
 template <typename iterator_traits>
-const key_value& session<persistent_data_store, cache_data_store>::session_iterator<iterator_traits>::read_(const bytes& key) const {
-    auto* current = m_cache_iterator_state.child.get();
-    while (current) {
-        if (current->cache.contains(key)) {
-            return current->cache.read(key);
+session<persistent_data_store, cache_data_store>::session_iterator<iterator_traits>::session_iterator(const session_iterator& it)
```
revl: I think these copy and move constructors are a bit too computation-heavy. Would it be possible to use a pointer to some data structure, like shared_ptr or something? In that case, copy and move constructors would just copy or move that pointer.

revl (approved):

libraries/chain_kv/include/b1/session/session.hpp
TimothyBanks: The queues maintain a pointer into the iterator states container (which holds stack based instances).  So moving the queues and the iterator states would result in a new iterator states container and the data in the queues pointing into the old iterator states container which means they would be pointing to reclaimed memory.  It's a similar issue with copying, except the queues will be pointing back to the original container instead of the new one.  

The iterator state struct isn't that heavy to copy or move around so it should be fine not be stored on the heap somewhere.

libraries/chain_kv/include/b1/session/session.hpp
TimothyBanks: The problem is, for example, that we could be incrementing along and while this is updating the queue that is holding the value of the next iterator, the queue that is holding the values for the previous container is not updating because it nothing is being pushed/popped to it.  So the idea was to force an update on the queue before using just to make sure it was correctly sorted before asking for the top element.

libraries/chain_kv/include/b1/session/bytes.hpp
```diff
@@ -248,31 +248,51 @@ inline bool bytes::operator!=(const bytes& other) const {
...
         return std::string_view{reinterpret_cast<const char*>(lhs.data()), lhs.length()} > std::string_view{reinterpret_cast<const char*>(rhs.data()), rhs.length()};
     };
 };
 
 template <>
-class hash<eosio::session::bytes> final {
+struct hash<eosio::session::bytes> final {
 public:
   size_t operator()(const eosio::session::bytes& b) const {
     return std::hash<std::string_view>{}({reinterpret_cast<const char*>(b.data()), b.length()});
```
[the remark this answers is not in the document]
TimothyBanks: Hashing a `std::string_view` with nullptr data or 0 length string should be able to handle this.  Unless we want to define the result differently that what `std::string_view` returns.

TimothyBanks (commented):
@revl I merged the fixes for this review comments into https://github.com/EOSIO/eos/pull/9366.

PR #9408, opened 2020-08-17
Rocksdb composite key construction for db intrinsic primary and secondary keys

brianjohnson5972:
## Change Description
db intrinsics are "keyed" off of scope, table, and a primary key.  This is how all objects are stored.  It also allows for a secondary key for each type. In chainbase, the secondary keys are ordered by the order of that secondary key, and then by the order of the primary key that is stored with it.
db_key_value_format is a class designed to provide the formatting for creating composite keys to store in RocksDB for primary and secondary keys.
Unit tests verify key construction and deconstruction, ordering and error handling.
Also, threw in an empty place holder for db_context_rocksdb.cpp.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/db_key_value_format.cpp
+++ b/libraries/chain/db_key_value_format.cpp
@@ -1,0 +1,49 @@
+#include <eosio/chain/db_key_value_format.hpp>
+
+namespace eosio { namespace chain {
+   std::string db_key_value_format::to_string(const key_type& kt) {
+      switch (kt) {
+         case key_type::primary: return "primary key of type uint64_t";
+         case key_type::sec_i64: return "seconday key of type uint64_t";
+         case key_type::sec_i128: return "seconday key of type uint128_t";
+         case key_type::sec_i256: return "seconday key of type key256_t";
+         case key_type::sec_double: return "seconday key of type float64_t";
+         case key_type::sec_long_double: return "seconday key of type float128_t";
+         default:
+            const int kt_as_int = static_cast<char>(kt);
+            return std::string("<invalid key_type: ") + std::to_string(kt_as_int) + ">";
+      }
+   }
+
+   db_key_value_format::key_type db_key_value_format::extract_from_composite_key(bytes::const_iterator& key_loc, bytes::const_iterator key_end, name& scope, name& table) {
+      uint64_t temp_name = 0;
+      EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, temp_name), bad_composite_key_exception,
+                 "DB intrinsic key-value store composite key is malformed, it does not contain a scope");
+      scope = name{temp_name};
+      EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, temp_name), bad_composite_key_exception,
+                 "DB intrinsic key-value store composite key is malformed, it does not contain a table");
+      table = name{temp_name};
+      EOS_ASSERT(key_loc != key_end, bad_composite_key_exception,
+                 "DB intrinsic key-value store composite key is malformed, it does not contain an indication of the "
+                 "type of the db-key (primary uint64_t or secondary uint64_t/uint128_t/etc)");
+      return key_type(*key_loc++);
+   }
+
+   bytes db_key_value_format::create_primary_key(name scope, name table, uint64_t db_key) {
+      bytes composite_key;
+      prepare_composite_key<uint64_t, key_type::primary>(composite_key, scope, table, sizeof(uint64_t));
+      b1::chain_kv::append_key(composite_key, db_key);
+      return composite_key;
+   }
+
+   void db_key_value_format::get_primary_key(const bytes& composite_key, name& scope, name& table, uint64_t& db_key) {
+      bytes::const_iterator composite_loc = composite_key.cbegin();
+      auto kt = extract_from_composite_key(composite_loc, composite_key.cend(), scope, table);
+      EOS_ASSERT(kt == key_type::primary, bad_composite_key_exception,
+                 "DB intrinsic key-value store composite key is malformed, it is supposed to be a primary key, "
+                 "but it is a: " + to_string(kt));
+      EOS_ASSERT(b1::chain_kv::extract_key(composite_loc, composite_key.cend(), db_key), bad_composite_key_exception,
+                 "DB intrinsic key-value store composite key is malformed, it is supposed to have a primary key");
+   }
+
+}} // namespace eosio::chain
--- a/libraries/chain/db_key_value_format.cpp
+++ b/libraries/chain/db_key_value_format.cpp
@@ -4,11 +4,11 @@ namespace eosio { namespace chain {
    std::string db_key_value_format::to_string(const key_type& kt) {
       switch (kt) {
          case key_type::primary: return "primary key of type uint64_t";
-         case key_type::sec_i64: return "seconday key of type uint64_t";
-         case key_type::sec_i128: return "seconday key of type uint128_t";
-         case key_type::sec_i256: return "seconday key of type key256_t";
-         case key_type::sec_double: return "seconday key of type float64_t";
-         case key_type::sec_long_double: return "seconday key of type float128_t";
+         case key_type::sec_i64: return "secondary key of type uint64_t";
+         case key_type::sec_i128: return "secondary key of type uint128_t";
+         case key_type::sec_i256: return "secondary key of type key256_t";
+         case key_type::sec_double: return "secondary key of type float64_t";
+         case key_type::sec_long_double: return "secondary key of type float128_t";
          default:
             const int kt_as_int = static_cast<char>(kt);
             return std::string("<invalid key_type: ") + std::to_string(kt_as_int) + ">";
--- a/libraries/chain/db_key_value_format.cpp
+++ b/libraries/chain/db_key_value_format.cpp
@@ -15,7 +15,7 @@ namespace eosio { namespace chain {
       }
    }
 
-   db_key_value_format::key_type db_key_value_format::extract_from_composite_key(bytes::const_iterator& key_loc, bytes::const_iterator key_end, name& scope, name& table) {
+   db_key_value_format::key_type db_key_value_format::extract_from_composite_key(b1::chain_kv::bytes::const_iterator& key_loc, b1::chain_kv::bytes::const_iterator key_end, name& scope, name& table) {
       uint64_t temp_name = 0;
       EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, temp_name), bad_composite_key_exception,
                  "DB intrinsic key-value store composite key is malformed, it does not contain a scope");
@@ -29,15 +29,15 @@ namespace eosio { namespace chain {
       return key_type(*key_loc++);
    }
 
-   bytes db_key_value_format::create_primary_key(name scope, name table, uint64_t db_key) {
-      bytes composite_key;
+   b1::chain_kv::bytes db_key_value_format::create_primary_key(name scope, name table, uint64_t db_key) {
+      b1::chain_kv::bytes composite_key;
       prepare_composite_key<uint64_t, key_type::primary>(composite_key, scope, table, sizeof(uint64_t));
       b1::chain_kv::append_key(composite_key, db_key);
       return composite_key;
    }
 
-   void db_key_value_format::get_primary_key(const bytes& composite_key, name& scope, name& table, uint64_t& db_key) {
-      bytes::const_iterator composite_loc = composite_key.cbegin();
+   void db_key_value_format::get_primary_key(const b1::chain_kv::bytes& composite_key, name& scope, name& table, uint64_t& db_key) {
+      b1::chain_kv::bytes::const_iterator composite_loc = composite_key.cbegin();
       auto kt = extract_from_composite_key(composite_loc, composite_key.cend(), scope, table);
       EOS_ASSERT(kt == key_type::primary, bad_composite_key_exception,
                  "DB intrinsic key-value store composite key is malformed, it is supposed to be a primary key, "
--- a/libraries/chain/db_key_value_format.cpp
+++ b/libraries/chain/db_key_value_format.cpp
@@ -1,50 +1,52 @@
 #include <eosio/chain/db_key_value_format.hpp>
 
-namespace eosio { namespace chain {
-   std::string db_key_value_format::to_string(const key_type& kt) {
-      switch (kt) {
-         case key_type::primary: return "primary key of type uint64_t";
-         case key_type::sec_i64: return "secondary key of type uint64_t";
-         case key_type::sec_i128: return "secondary key of type uint128_t";
-         case key_type::sec_i256: return "secondary key of type key256_t";
-         case key_type::sec_double: return "secondary key of type float64_t";
-         case key_type::sec_long_double: return "secondary key of type float128_t";
-         default:
-            const int kt_as_int = static_cast<char>(kt);
-            return std::string("<invalid key_type: ") + std::to_string(kt_as_int) + ">";
+namespace eosio { namespace chain { namespace db_key_value_format {
+   namespace detail {
+      std::string to_string(const key_type& kt) {
+         switch (kt) {
+            case key_type::primary: return "primary key of type uint64_t";
+            case key_type::sec_i64: return "secondary key of type uint64_t";
+            case key_type::sec_i128: return "secondary key of type uint128_t";
+            case key_type::sec_i256: return "secondary key of type key256_t";
+            case key_type::sec_double: return "secondary key of type float64_t";
+            case key_type::sec_long_double: return "secondary key of type float128_t";
+            default:
+               const int kt_as_int = static_cast<char>(kt);
+               return std::string("<invalid key_type: ") + std::to_string(kt_as_int) + ">";
+         }
       }
-   }
 
-   db_key_value_format::intermittent_decomposed_values db_key_value_format::extract_from_composite_key(b1::chain_kv::bytes::const_iterator begin, b1::chain_kv::bytes::const_iterator key_end) {
-      auto key_loc = begin;
-      uint64_t scope_name = 0;
-      EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, scope_name), bad_composite_key_exception,
-                 "DB intrinsic key-value store composite key is malformed, it does not contain a scope");
-      uint64_t table_name = 0;
-      EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, table_name), bad_composite_key_exception,
-                 "DB intrinsic key-value store composite key is malformed, it does not contain a table");
-      EOS_ASSERT(key_loc != key_end, bad_composite_key_exception,
-                 "DB intrinsic key-value store composite key is malformed, it does not contain an indication of the "
-                 "type of the db-key (primary uint64_t or secondary uint64_t/uint128_t/etc)");
-      return { name{scope_name}, name{table_name}, key_loc, key_type(*key_loc++) };
-   }
+      intermittent_decomposed_values extract_from_composite_key(b1::chain_kv::bytes::const_iterator begin, b1::chain_kv::bytes::const_iterator key_end) {
+         auto key_loc = begin;
+         uint64_t scope_name = 0;
+         EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, scope_name), bad_composite_key_exception,
+                    "DB intrinsic key-value store composite key is malformed, it does not contain a scope");
+         uint64_t table_name = 0;
+         EOS_ASSERT(b1::chain_kv::extract_key(key_loc, key_end, table_name), bad_composite_key_exception,
+                    "DB intrinsic key-value store composite key is malformed, it does not contain a table");
+         EOS_ASSERT(key_loc != key_end, bad_composite_key_exception,
+                    "DB intrinsic key-value store composite key is malformed, it does not contain an indication of the "
+                    "type of the db-key (primary uint64_t or secondary uint64_t/uint128_t/etc)");
+         return { name{scope_name}, name{table_name}, key_loc, key_type(*key_loc++) };
+      }
+   } // namespace detail
 
-   b1::chain_kv::bytes db_key_value_format::create_primary_key(name scope, name table, uint64_t db_key) {
+   b1::chain_kv::bytes create_primary_key(name scope, name table, uint64_t db_key) {
       b1::chain_kv::bytes composite_key;
-      prepare_composite_key<uint64_t, key_type::primary>(composite_key, scope, table, sizeof(uint64_t));
+      detail::prepare_composite_key<uint64_t, key_type::primary>(composite_key, scope, table, sizeof(uint64_t));
       b1::chain_kv::append_key(composite_key, db_key);
       return composite_key;
    }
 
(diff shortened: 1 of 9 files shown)


libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
    return next_prefix;
 }
 
+namespace detail {
+   typedef __uint128_t uint128_t;
```
TimothyBanks, line 115: What's the stance on typedef vs alias at block.one?

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+         from_ptr += Size;
+      }
+   }
+
+   template<typename T>
+   struct value_storage;
+
+   template<typename T>
+   struct value_storage<T*> {
+      char* as_char_ptr(T* value) const { return reinterpret_cast<char*>(value); }
```
TimothyBanks, line 133: Could these methods just be static to avoid the need for instantiation an instance to invoke?

I think the `as_char_ptr` can be constexpr too?

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      constexpr std::size_t size() const { return sizeof(T); }
+   };
+
+   template<typename T>
+   struct value_storage {
+      char* as_char_ptr(T& value) const { return reinterpret_cast<char*>(&value); }
```
TimothyBanks, line 139: Similar question here, could these just be static methods?

I think the `as_char_ptr` can be constexpr too?

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      constexpr std::size_t size() const { return sizeof(T); }
+   };
+
+   template <typename T, std::size_t N>
+   auto append_key(bytes& dest, T value) {
+      constexpr auto type_size = value_storage<T>().size();
```
TimothyBanks, line 145: This could be static too.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      constexpr auto buf_size = type_size * N;
```
TimothyBanks, line 146: This could be static too.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      char buf[buf_size];
+      detail::copy_and_swap_to_buffer<type_size, N>(buf, value_storage<T>().as_char_ptr(value));
+      std::reverse(std::begin(buf), std::end(buf));
+      dest.insert(dest.end(), std::begin(buf), std::end(buf));
```
TimothyBanks, line 150: Would there be any benefit in doing this operation in place in `dest`?  Expand `dest` first and then copy_and_swap and reverse on that new chunk in `dest`?

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      value = {value ^ (mask | signbit)};
+      T float_result;
+      std::memcpy(&float_result, &value, sizeof(UInt));
+      return float_result;
+   }
+
+   template<typename Key, std::size_t N>
+   bool extract_key(bytes::const_iterator& key_loc, bytes::const_iterator key_end, Key& key) {
+      const auto distance = std::distance(key_loc, key_end);
+      constexpr auto type_size = value_storage<Key>().size();
```
TimothyBanks, line 184: This could be static.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      constexpr auto key_size = type_size * N;
```
TimothyBanks, line 185: This could be static.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+      if (distance < key_size)
+         return false;
+
+      key_end = key_loc + key_size;
+      bytes key_store(key_loc, key_end);
```
TimothyBanks, line 190: It seems like there might be a solid use case for a reverse iterator on bytes.  Might save use from needing to do the `std::reverse` in some places.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
 auto append_key(bytes& dest, T value) -> std::enable_if_t<std::is_unsigned_v<T>, void> {
-   char buf[sizeof(value)];
-   memcpy(buf, &value, sizeof(value));
-   std::reverse(std::begin(buf), std::end(buf));
-   dest.insert(dest.end(), std::begin(buf), std::end(buf));
+   detail::append_key<T, 1>(dest, value);
+}
+
+template <typename T, std::size_t N>
+auto append_key(bytes& dest, std::array<T, N> value) -> std::enable_if_t<std::is_unsigned_v<T>, void> {
```
TimothyBanks, line 205: Is this style acceptable in eosio?  I hope you say it is because I strongly prefer it.

libraries/chain/include/eosio/chain/db_key_value_format.hpp
```diff
@@ -0,0 +1,90 @@
...
+         b1::chain_kv::append_key(composite_key, db_key);
+         b1::chain_kv::append_key(composite_key, primary_key);
+         return composite_key;
+      }
+
+      template<typename Key>
+      static bool get_secondary_key(const b1::chain_kv::bytes& composite_key, name& scope, name& table, Key& db_key, uint64_t& primary_key) {
+         b1::chain_kv::bytes::const_iterator composite_loc = composite_key.cbegin();
+         auto kt = extract_from_composite_key(composite_loc, composite_key.cend(), scope, table);
+         const auto expected_kt = determine_sec_type<Key>::kt;
```
TimothyBanks, line 31: This could be static constexpr.

libraries/chain/include/eosio/chain/db_key_value_format.hpp
```diff
@@ -0,0 +1,90 @@
...
+      struct determine_sec_type<float64_t> {static constexpr key_type kt = key_type::sec_double; };
+
+      template<>
+      struct determine_sec_type<float128_t> { static constexpr key_type kt = key_type::sec_long_double; };
+
+      static std::string to_string(const key_type& kt);
+
+      template<typename Key, key_type kt = determine_sec_type<Key>::kt>
+      static void prepare_composite_key(b1::chain_kv::bytes& key_storage, name scope, name table, std::size_t key_size) {
+         const auto scope_size = sizeof(scope);
```
TimothyBanks, line 77: The following few variables could be static constexpr as well.

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+
+#include <boost/test/unit_test.hpp>
+#include <cstring>
+
+using key_type = eosio::chain::db_key_value_format::key_type;
+using name = eosio::chain::name;
+
+BOOST_AUTO_TEST_SUITE(db_to_kv_tests)
+
+const uint64_t overhead_size = sizeof(name) * 2 + 1; // 8 (scope) + 8 (table) + 1 (key type)
```
TimothyBanks, line 12: Should we prefer constexpr over const here?

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+};
+
+template<typename T, typename KeyFunc>
+void check_ordered_keys(const std::vector<T> ordered_keys, KeyFunc key_func, bool skip_trailing_prim_key = false) {
+   verify_scope_table_order(ordered_keys[0], ordered_keys[1], key_func, skip_trailing_prim_key);
+   name scope{0};
+   name table{0};
+   const uint64_t prim_key = 0;
+   std::vector<b1::chain_kv::bytes> composite_keys;
+   for (const auto key: ordered_keys) {
```
TimothyBanks, line 103: Might want to use reference on the key type.

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+   }
+}
+
+// using this function to allow check_ordered_keys to work for all key types
+b1::chain_kv::bytes prim_drop_extra_key(name scope, name table, uint64_t db_key, uint64_t unused_primary_key) {
+   return eosio::chain::db_key_value_format::create_primary_key(scope, table, db_key);
+};
+
+template<typename T, typename KeyFunc>
+void check_ordered_keys(const std::vector<T> ordered_keys, KeyFunc key_func, bool skip_trailing_prim_key = false) {
```
TimothyBanks, line 97: Why not pass the `'ordered_keys` parameters by reference too?

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+      auto comp_key = key_func(scope, table, key, prim_key);
+      composite_keys.push_back(comp_key);
+   }
+   for (unsigned int i = 0; i < composite_keys.size() - 1; ++i) {
+      verify_assending_composite_keys(composite_keys, i, "check_ordered_keys");
+   }
+}
+
+BOOST_AUTO_TEST_CASE(ui64_keys_conversions_test) {
+   const name scope = N(myscope);
```
TimothyBanks, line 113: Could the following three variables be constexpr?

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+   std::vector<uint64_t> keys2;
+   keys2.push_back(0x0);
+   keys2.push_back(0x1);
+   keys2.push_back(0x7FFF'FFFF'FFFF'FFFF);
+   keys2.push_back(0x8000'0000'0000'0000);
+   keys2.push_back(0xFFFF'FFFF'FFFF'FFFF);
+   check_ordered_keys(keys2, &eosio::chain::db_key_value_format::create_secondary_key<uint64_t>);
+}
+
+eosio::chain::uint128_t create_uint128(uint64_t upper, uint64_t lower) {
```
TimothyBanks, line 161: This method could be constexpr too.

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+   return (eosio::chain::uint128_t(upper) << 64) + lower;
+}
+
+BOOST_AUTO_TEST_CASE(ui128_key_conversions_test) {
+   const name scope = N(myscope);
```
TimothyBanks, line 166: Could the following variables be constexpr as well?

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+   keys.push_back(create_uint128(0x0, 0xFFFF'FFFF'FFFF'FFFF));
+   keys.push_back(create_uint128(0x1, 0x0));
+   keys.push_back(create_uint128(0x1, 0xFFFF'FFFF'FFFF'FFFF));
+   keys.push_back(create_uint128(0x7FFF'FFFF'FFFF'FFFF, 0xFFFF'FFFF'FFFF'FFFF));
+   keys.push_back(create_uint128(0x8000'0000'0000'0000, 0xFFFF'FFFF'FFFF'FFFF));
+   keys.push_back(create_uint128(0xFFFF'FFFF'FFFF'FFFF, 0xFFFF'FFFF'FFFF'FFFF));
+   check_ordered_keys(keys, &eosio::chain::db_key_value_format::create_secondary_key<eosio::chain::uint128_t>);
+}
+
+eosio::chain::key256_t create_key256(eosio::chain::uint128_t upper, eosio::chain::uint128_t lower) {
```
TimothyBanks, line 200: constexpr function?

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
swatanabe-b1: `reinterpret_cast` is excluded from constant expressions.  [expr.const]

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -110,12 +111,141 @@ inline bytes get_next_prefix(const bytes& prefix) {
...
+namespace detail {
+   typedef __uint128_t uint128_t;
+
+   template<std::size_t Size, std::size_t N>
+   void copy_and_swap_to_buffer(char* to, const char* from) {
+      char* to_ptr = to + Size * (N - 1);
+      const char* from_ptr = from;
+      for (unsigned int i = 0; i < N; ++i) {
+         memcpy(to_ptr, from_ptr, Size);
+         to_ptr -= Size;
```
swatanabe-b1, line 123: "If the expression `P` points to element `x[i]` of an array object x with n elements, ... the expression `P - J` points to the (possibly-hypothetical) element `x[i − j]` if `0 ≤ i − j ≤ n;` *otherwise, the behavior is undefined*" [expr.add] (emphasis added)

libraries/chain/include/eosio/chain/db_key_value_format.hpp
```diff
@@ -0,0 +1,90 @@
...
+         const auto table_size = sizeof(table);
+         const auto type_size = sizeof(key_type);
```
b1bart, line 79: below we always write this as a single byte.  `sizeof(key_type)` may not be `1`.  Perhaps some static asserting here to guarantee that we don't one day mess this up?

libraries/chain/include/eosio/chain/db_key_value_format.hpp
```diff
@@ -0,0 +1,90 @@
...
+         key_storage.reserve(scope_size + table_size + type_size + key_size);
+         b1::chain_kv::append_key(key_storage, scope.to_uint64_t());
+         b1::chain_kv::append_key(key_storage, table.to_uint64_t());
+         key_storage.push_back(static_cast<char>(kt));
+      }
+
+      static key_type extract_from_composite_key(b1::chain_kv::bytes::const_iterator& key_loc, b1::chain_kv::bytes::const_iterator key_end, name& scope, name& table);
```
b1bart, line 86: I know its a pedantic nitpick but the function signature here seems hard to understand.

it has a return value and 4 parameters that are
-input&output
-input
-output
-output

I do think it would be more reasonable to take in a either an iterator range (begin and end) or a `bytes` type and return either a structure with explicit fields for the output OR a tuple that can be destructured.

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+   BOOST_CHECK_EQUAL(key, decomposed_key);
+   BOOST_CHECK_EQUAL(primary_key, decomposed_primary_key);
+
+   verify_secondary_wont_convert(composite_key, key_type::sec_double);
+
+   std::vector<float64_t> keys;
+   keys.push_back(to_softfloat64(-100000.0));
+   keys.push_back(to_softfloat64(-99999.9));
+   keys.push_back(to_softfloat64(-1.9));
+   keys.push_back(to_softfloat64(0.0));
```
b1bart, line 266: should we verify positive and negative 0.0?

unittests/db_to_kv_tests.cpp
```diff
@@ -0,0 +1,309 @@
...
+   keys.push_back(to_softfloat64(0.1));
+   keys.push_back(to_softfloat64(1.9));
+   keys.push_back(to_softfloat64(99999.9));
+   keys.push_back(to_softfloat64(100000.0));
+   check_ordered_keys(keys, &eosio::chain::db_key_value_format::create_secondary_key<float64_t>);
```
b1bart, line 271: Do we need to check for cases with denormalized floating point values and other things like NaN's?

swatanabe-b1: IIRC, the db intrinsics forbid NaN keys.  Subnormal and infinities are both valid, however.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
brianjohnson5972: I had quickly copied it, but I would definitely prefer using, so changing it.  Don't know what other reasons there are, but it is way easier to search and find "using TYPE-NAME =" then having to make a regex to find "typedef ... TYPE-NAME;"

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
brianjohnson5972: I am presuming that the stl developers can write more performant code than me, since this isn't a specialized case for this, unless it is the case that std::begin/end and vector(dest) doesn't have an optimization.  Without knowing, I would default to the simpler approach and we can revisit when we have performance numbers. (A little hint, if know one else decides to fact check you, you can just say "know it doesn't optimize it" and I won't look it up :), I'll take your word for it)

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
brianjohnson5972: Not sure which part of the style you are referring to, but if follows the style of the append_key method that was already there (or at least as far as I notice)

unittests/db_to_kv_tests.cpp
brianjohnson5972: We can, so I changed it.

unittests/db_to_kv_tests.cpp
brianjohnson5972: I think we could, but I think here it is cleaner to read to just be a const, since this is test code.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
brianjohnson5972: @swatanabe-b1 Are you saying that I am incorrectly bounding the use of this pointer?  Can you put the example in this specific logic, rather than using the more abstract example, so I can follow what you are referring to?  Is the problem in this method, or is it based on one (or more) of the instances of using the function?

libraries/chain/db_key_value_format.cpp
```diff
@@ -0,0 +1,49 @@
+#include <eosio/chain/db_key_value_format.hpp>
+
+namespace eosio { namespace chain {
+   std::string db_key_value_format::to_string(const key_type& kt) {
+      switch (kt) {
+         case key_type::primary: return "primary key of type uint64_t";
+         case key_type::sec_i64: return "seconday key of type uint64_t";
```
revl, line 7: "Seconda**r**y"

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
swatanabe-b1: In the last iteration of the loop, `to_ptr == to`.  The subtraction leaves it as `to - Size`, which is undefined behavior (assuming that `to` points to the beginning of an array), even though the out-of-bounds pointer is never dereferenced.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
revl: This form of function declaration is required here because the return type is referring to the template parameter T. I have no preference, but I doubt the majority would prefer using the arrow syntax in every function.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
swatanabe-b1: In an ideal world the compiler would not go through memory for the byte swapping, rending such optimizations pointless.  Unfortunately, neither gcc nor clang turns this use of std::reverse into bswap (or pshufb).

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
swatanabe-b1: The return type can always name template parameters.  The trailing return type allows referencing `dest` and `value` as well, but that isn't being used here.

libraries/chain/include/eosio/chain/db_key_value_format.hpp
brianjohnson5972: I was feeling the same way, but ignored my spidey sense, in order to keep on track.  I'll pass in const begin, end and return a tuple.

unittests/db_to_kv_tests.cpp
brianjohnson5972: yes, yes we should. Added separate test for 0.0 == -0.0.

unittests/db_to_kv_tests.cpp
brianjohnson5972: NaNs are precluded, added test for -inf and inf ordering compared to min and max.  What others do we need to check for? @b1bart

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -130,22 +131,27 @@ namespace detail {
...
-      constexpr static auto buf_size = type_size * N;
+
+      using t_type = typename value_storage<T>::type;
+      constexpr static auto buf_size = sizeof(t_type) * N;
       char buf[buf_size];
-      detail::copy_and_swap_to_buffer<type_size, N>(buf, value_storage<T>().as_char_ptr(value));
+      t_type* buf_as_type = reinterpret_cast<t_type*>(buf);
+      const t_type* first = value_storage<T>().as_ptr(value);
+      const t_type* last = first + N;
+      std::reverse_copy(first, last, buf_as_type);
```
swatanabe-b1, line 153: An array of `char` is not guaranteed to be correctly aligned to hold objects of type `t_type`.

brianjohnson5972: @swatanabe-b1 Are you saying that this is not guaranteed to work for uint64_t, uint128_t, key256_t, float64_t, and float128_t or are you saying that it isn't guaranteed to work for an arbitrary type_t?

brianjohnson5972: @swatanabe-b1 Huang-Ming helped me, I'll push the fix in a few minutes

PR #9411, opened 2020-08-18
fix the bug that the flight bytes are cacculated incorrect

vzqzhang:
## Change Description
There is an option for nodeos that specifies the max flight bytes. However, it did not work fine since the calculation of the bytes in flight was wrong. The bytes in flight was always 0 or 1 so it never exceeds the max flight bytes. 

It causes an issue that nodeos continues to accept new transactions when there is no new blocks until the process reaches the maximum file descriptors. It causes the nodeos malfunctioning and cannot be recovered. In some environments it causes the whole OS malfunctioning.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -397,7 +397,7 @@ namespace eosio {
             :object(std::move(object))
             ,impl(impl)
             {
-               count = detail::in_flight_sizeof(object);
+               count = detail::in_flight_sizeof(this->object);
                impl.bytes_in_flight += count;
             }
 
@@ -467,7 +467,7 @@ namespace eosio {
          detail::internal_url_handler make_app_thread_url_handler( int priority, url_handler next ) {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [this, priority, next_ptr=std::move(next_ptr)]( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) mutable {
-               auto tracked_b = make_in_flight(std::move(b), *this);
+               auto tracked_b = make_in_flight<string>(std::move(b), *this);
                if (!conn->verify_max_bytes_in_flight()) {
                   return;
                }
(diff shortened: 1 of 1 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -448,7 +449,7 @@ namespace eosio {
           */
          template<typename T>
          static auto make_in_flight(T&& object, http_plugin_impl& impl) {
-            return in_flight<T>(std::forward<T>(object), impl);
+            return std::make_shared<in_flight<T>>(in_flight<T>(std::forward<T>(object), impl));
```
heifner, line 452: Does this have to be a `shared_ptr`? Can't it just be moved into `wrapped_then`  then when `wrapped_then` is destructed it will as well.

vzqzhang: > Does this have to be a `shared_ptr`? Can't it just be moved into `wrapped_then` then when `wrapped_then` is destructed it will as well.

No. Since in_flight disables its copy constructor, tracked_b gets deleted if it's not wrapped by a shared pointer.

heifner: Move it, not copy it.

vzqzhang: > Move it, not copy it.

It failed to move.

b1bart: there are a few reasons why we moved this to a `std::shared_ptr` the first is that it makes it copyable which is necessary for use with `std::function`.  The second is that we have multiple asynchronous paths so there was no single lifecycle for `tracked_b`.  We are leaning on the `std::shared_ptr` 's reference counting to give us the lifetime guarantees we want.

PR #9412, opened 2020-08-18
port the fix of flight bytes bug into release/2.0.x

vzqzhang:
## Change Description
There is an option for nodeos that specifies the max flight bytes. However, it did not work fine since the calculation of the bytes in flight was wrong. The bytes in flight was always 0 or 1 so it never exceeds the max flight bytes. 

It causes an issue that nodeos continues to accept new transactions when there is no new blocks until the process reaches the maximum file descriptors. It causes the nodeos malfunctioning and cannot be recovered. In some environments it causes the whole OS malfunctioning.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -337,6 +337,7 @@ namespace eosio {
                con->send_http_response();
                return false;
             }
+
             return true;
          }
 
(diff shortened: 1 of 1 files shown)


vzqzhang:
The original PR is: https://github.com/EOSIO/eos/pull/9411

PR #9416, opened 2020-08-19
fix template instantiation for host function

huangminghuang:
## Change Description

Directly address the build time issue caused by the extra template instantiations for host function wrappers in multiple transaction units.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/webassembly/runtimes/host_function_registrator.cpp
```diff
@@ -0,0 +1,342 @@
...
+   inline static auto NAME##_registrator = NAME##_registrator_impl();
+
+#define REGISTER_LEGACY_CF_ONLY_HOST_FUNCTION(NAME, ...)                                                               \
+   static host_function_registrator<&interface::NAME, legacy_static_check_wl_args, context_free_check, ##__VA_ARGS__>  \
+       NAME##_registrator_impl() {                                                                                     \
+      return {BOOST_HANA_STRING("env"), BOOST_HANA_STRING(#NAME)};                                                     \
+   }                                                                                                                   \
+   inline static auto NAME##_registrator = NAME##_registrator_impl();
+
+REGISTER_LEGACY_CF_ONLY_HOST_FUNCTION(get_context_free_data)
```
larryk85, line 66: Please add the comment grouping that is the same as in `interface.hpp` to help navigate this better.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
```diff
@@ -74,7 +74,7 @@ static void segv_handler(int sig, siginfo_t* info, void* ctx)  {
 
 static intrinsic grow_memory_intrinsic EOSVMOC_INTRINSIC_INIT_PRIORITY("eosvmoc_internal.grow_memory", IR::FunctionType::get(IR::ResultType::i32,{IR::ValueType::i32,IR::ValueType::i32}),
   (void*)&eos_vm_oc_grow_memory,
-  find_intrinsic_index("eosvmoc_internal.grow_memory")
+  std::integral_constant<std::size_t, find_intrinsic_index("eosvmoc_internal.grow_memory")>::value
```
larryk85, line 77: Please change these to `std::integral_constant_v`

b1bart: I don't think there is a `std::integral_constant_v<>`. it is either `::value` as it was instantiate it and it will cast to a `std::size_t`

huangminghuang:
> Can you make this `std::tuple_size_v` instead.

std::tuple_size_v<> only works for tuple, not for std::array for now.  Library TS has it but not merged into the standard yet.

libraries/chain/webassembly/runtimes/eos-vm-oc/executor.cpp
larryk85: Gotcha.

PR #9419, opened 2020-08-20
AMQP ordered transaction processing

heifner:
## Change Description

- `amqp_trx_plugin` now submits one transaction at a time so that order of transaction execution matches order of transactions in the AMQP queue.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

New `amqp_trx_plugin` option:

```
  --amqp-trx-queue-size arg (=1000)     The maximum number of transactions to 
                                        pull from the AMQP queue.
```


plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
```diff
@@ -0,0 +1,212 @@
...
+                  auto trx_meta = i.fut.get();
+                  dlog("posting trx: ${id}", ("id", trx_meta->id()));
+                  app().post(priority::low, [self, trx{std::move(trx_meta)}, next{std::move(i.next)}](){
+                     auto retry_later = [self](const chain::transaction_metadata_ptr& trx, producer_plugin::next_function<chain::transaction_trace_ptr>& next) {
+                        std::promise<chain::transaction_metadata_ptr> p;
+                        q_item i;
+                        i.fut = p.get_future();
+                        p.set_value(trx);
+                        i.next = next;
+                        self->queue_.push_front( std::move( i ) );
```
swatanabe-b1, line 168: Is this busy wait okay?

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
```diff
@@ -0,0 +1,212 @@
...
+                     };
+                     self->prod_plugin_->execute_incoming_transaction(trx, next, retry_later);
```
swatanabe-b1, line 170: Do we allow this plugin to run in speculative mode?  Aborting a block will add trx to the `unapplied_transaction_queue`.

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
heifner: If configured for 500ms `max_block_cpu_usage`, then yes. However, I need to put some thought to what to do if configured differently.

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
heifner: All bets are off (in terms of order) if run on a non-bp node. I'll add a warning to that affect. Maybe I should just add back in non-order processing and use that if not configured as a producer. Also should only process when this block is "our" block.

libraries/amqp/include/eosio/amqp/amqp_handler.hpp
```diff
@@ -98,7 +98,10 @@ class amqp {
          std::promise<void> stop_promise;
          auto future = stop_promise.get_future();
          boost::asio::post( *handler_->amqp_strand(), [&]() {
-            stop_promise.set_value();
+            auto& cb = channel_->close();
```
swatanabe-b1, line 101: If initialization timed out or otherwise failed, it's possible for `channel_` to be null.

libraries/chain/controller.cpp
```diff
@@ -1606,9 +1616,8 @@ struct controller_impl {
 
       emit( self.block_start, head->block_num + 1 );
 
-      auto guard_pending = fc::make_scoped_exit([this, head_block_num=head->block_num](){
-         protocol_features.popped_blocks_to( head_block_num );
-         pending.reset();
+      auto guard_pending = fc::make_scoped_exit([this](){
+         abort_block();
```
swatanabe-b1, line 1620: Should not call potentially throwing code in a destructor.

heifner: abort_block is not any more likely to throw than what was there before, but a try-catch couldn't hurt

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
heifner: Still need to add a test to verify that abort_block does not process trx out of order.

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
heifner: It now respects cpu effort percent and does not have a busy wait.

libraries/amqp/include/eosio/amqp/amqp_handler.hpp
heifner: Add check for null

swatanabe-b1 (approved):

PR #9421, opened 2020-08-20
Rocksdb backing store files move

brianjohnson5972:
## Change Description
Created a directory under libraries/chain called backing_store and moved all db intrinsic related backing store classes.
There are no logic changes for this, only moving files and changed some class names.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/backing_store/db_chainbase_iter_cache.hpp
+++ b/libraries/chain/include/eosio/chain/backing_store/db_chainbase_iter_cache.hpp
@@ -5,7 +5,7 @@
 #include <algorithm>
 #include <set>
 
-namespace eosio { namespace chain { namespace backend_store {
+namespace eosio { namespace chain { namespace backing_store {
 
 template<typename T>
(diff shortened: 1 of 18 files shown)


libraries/chain/include/eosio/chain/backing_store/db_chainbase_iter_cache.hpp
```diff
@@ -5,7 +5,7 @@
 #include <algorithm>
```
linhuang-blockone, line 5: *.hpp files are in the same backing_store as *.cpp files, not in .../chain/include/eosio/chain/backing_store?

TimothyBanks: I think this is a private header file.

PR #9424, opened 2020-08-21
new chain_config param: action return value limit

dimas1185:
## Change Description

Consensus change.

Currently action return values are limited only by CPU and max wasm memory which is 20MB.
We need to provide node operators a way to prevent users from filling their SHiP logs with noise.

This change includes following changes:
1. adding `max_action_return_value_size` to `chain_config`. default value is 256 bytes.
this config value is used in `set_action_return_value` host function:
2. adding `wql-max-action-return-value` command line option to rodeos with default value of 20MB.
it is used in `set_action_return_value` from `rodeos/callbacks/action.hpp`

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [x] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/chain_config.cpp
```diff
@@ -41,7 +41,8 @@ namespace eosio { namespace chain {
 
 void chain_config_v1::validate() const {
    chain_config_v0::validate();
-   //add validation on new v1 parameters here
+   EOS_ASSERT( max_action_return_value_size <= MAX_SIZE_OF_BYTE_ARRAYS, action_validate_exception,
+               "max action return value size should be less than MAX_SIZE_OF_BYTE_ARRAYS" );
```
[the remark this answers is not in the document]
swatanabe-b1: I'm pretty sure that `MAX_SIZE_OF_BYTE_ARRAYS` already affects consensus.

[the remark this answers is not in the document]
dimas1185: any suggestion on maximum value? I tied it to `MAX_SIZE_OF_BYTE_ARRAYS` because of I thought it is a maximum value action return value can return.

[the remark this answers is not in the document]
heifner: @swatanabe-b1 is likely correct that consensus already depends on MAX_SIZE_OF_BYTE_ARRAYS. I think a static_assert here of `MAX_SIZE_OF_BYTE_ARRAYS == 20*1024*1024` would be good.

swatanabe-b1:
For the record, max wasm memory is 33 MiB, not 20.

libraries/chain/chain_config.cpp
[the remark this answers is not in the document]
dimas1185: @heifner if `MAX_SIZE_OF_BYTE_ARRAYS` will be changed doesn't it mean that consensus changed and `chain_config` needs to make changed accordingly?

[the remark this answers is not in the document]
heifner: Yes, exactly. Hence, the static_assert to make sure it doesn't happen on accident.

libraries/state_history/abi.cpp
```diff
@@ -418,18 +418,48 @@ extern const char* const state_history_plugin_abi = R"({
...
+                { "type": "uint16", "name": "max_authority_depth" },
+                { "type": "uint32", "name": "max_action_return_value_size" }
+            ]
+        },
         {
             "name": "global_property_v0", "fields": [
                 { "type": "uint32?", "name": "proposed_schedule_block_num" },
                 { "type": "producer_schedule", "name": "proposed_schedule" },
-                { "type": "chain_config", "name": "configuration" }
+                { "type": "chain_config_v0", "name": "configuration" }
```
swatanabe-b1, line 447: Don't change the format of existing types.

libraries/state_history/abi.cpp
```diff
@@ -418,18 +418,48 @@ extern const char* const state_history_plugin_abi = R"({
...
             ]
         },
         {
             "name": "global_property_v1", "fields": [
                 { "type": "uint32?", "name": "proposed_schedule_block_num" },
                 { "type": "producer_authority_schedule", "name": "proposed_schedule" },
-                { "type": "chain_config", "name": "configuration" },
+                { "type": "chain_config_v0", "name": "configuration" },
```
swatanabe-b1, line 454: Is it necessary to add a new version of `global_property`?  The only thing that changed was `chain_config` which is already a variant.

libraries/state_history/abi.cpp
swatanabe-b1: The format needs to match what is created by `libraries/state_history/include/eosio/state_history/serialization.hpp`, which also needs to be updated.

libraries/state_history/abi.cpp
dimas1185: in eos it is structure and its size is changed as per this PR. not sure how state history stuff works though

swatanabe-b1: If `state_history` represents a type as a variant, then changes to the type's layout don't need to be propagated up to containing types.

libraries/state_history/include/eosio/state_history/serialization.hpp
```diff
@@ -313,6 +313,7 @@ ST& operator<<(ST& ds, const history_serial_wrapper<eosio::chain::chain_config>&
    fc::raw::pack(ds, as_type<uint32_t>(obj.obj.max_inline_action_size));
    fc::raw::pack(ds, as_type<uint16_t>(obj.obj.max_inline_action_depth));
    fc::raw::pack(ds, as_type<uint16_t>(obj.obj.max_authority_depth));
+   fc::raw::pack(ds, as_type<uint16_t>(obj.obj.max_action_return_value_size));
```
swatanabe-b1, line 316: Increment the version at the top of the function.

libraries/state_history/abi.cpp
dimas1185: @swatanabe-b1 I made changes accordingly. also I believe I need to make similar changes to:
[libraries/abieos/include/eosio/ship_protocol.hpp](https://github.com/EOSIO/abieos/blob/develop/include/eosio/ship_protocol.hpp)

dimas1185:
discarded due to develop merge issues (for history clarity purpose)
new PR: https://github.com/EOSIO/eos/pull/9451

PR #9425, opened 2020-08-21
Refactor parts of controller into combined_database

revl:
## Change Description

This PR extracts parts of controller and controller_impl into a new class called `combined_database`. Also, it replaces `maybe_session` in controller.cpp with combined_session, which removes one indirection and improves readability.

## Change Type

**Select ONE**

- [ ] Documentation
- [ ] Stability bug fix
- [x] Other
- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/apply_context.cpp
```diff
@@ -91,16 +91,7 @@ void apply_context::exec_one()
          kv_iterators.resize(1);
          kv_destroyed_iterators.clear();
          if (!context_free) {
-            switch (control.get_config().backing_store) {
-               case backing_store_type::ROCKSDB:
-                  kv_backing_store = create_kv_rocksdb_context(control.kv_database(), control.kv_undo_stack(), receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
-                  break;
-               case backing_store_type::NATIVE:
-                  kv_backing_store = create_kv_chainbase_context(db, receiver, create_kv_resource_manager(*this), control.get_global_properties().kv_configuration);
```
linhuang-blockone, line 99: I would hesitate to move this piece of code away, as "db" is created by apply_context. I don't see combined_session has access to "db". This would break KV on Chainbase.

libraries/chain/controller.cpp
```diff
@@ -848,308 +734,10 @@ struct controller_impl {
       resource_limits.add_indices();
    }
 
-   void add_contract_tables_to_snapshot( const snapshot_writer_ptr& snapshot ) const {
```
linhuang-blockone, line 851: Where is those snapshot code moved to? Just saw Jeff's PR https://github.com/EOSIO/eos/pull/9426 
 (removing DISK mode). He simplifies snapshot code. You can take a look there, to consolidate with his code.

libraries/chain/apply_context.cpp
revl: Lin and I discussed this issue over Slack.

libraries/chain/controller.cpp
revl: Thanks for bringing Jeff's PR to my attention. I think those changes were a mistake.

libraries/chain/include/eosio/chain/combined_database.hpp
```diff
@@ -14,77 +36,92 @@
...
+
+      combined_session(chainbase::database& cb_database);
+
+      combined_session(chainbase::database& cb_database, b1::chain_kv::undo_stack& kv_undo_stack);
+
+      combined_session(combined_session&& src) noexcept;
+
+      ~combined_session() { undo(); }
+
+      combined_session& operator=(combined_session&& src) noexcept;
```
brianjohnson5972, line 60: If you didn't add something that needs this, I'd say we should remove it.  I am not sure where we would have a need for an already existing combined_session to have the contents of another combined_session moved into it.  If we don't have it, then we will see if someone does something that needs it, and if it makes sense, we can add it then.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,499 @@
...
+                        rocksdb::Slice key   = { move_to_rocks->kv_key.data(), move_to_rocks->kv_key.size() };
+                        rocksdb::Slice value = { move_to_rocks->kv_value.data(), move_to_rocks->kv_value.size() };
+                        batch.Put(b1::chain_kv::to_slice(b1::chain_kv::create_full_key(
+                                        prefix, move_to_rocks->contract.to_uint64_t(), key)),
+                                  value);
+                        db.remove(*move_to_rocks);
+                     }
+                  }
+               });
+               // TODO: Split the batch to avoid storing it all in memory at once.
```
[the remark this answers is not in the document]
revl: Done

libraries/chain/controller.cpp
```diff
@@ -487,7 +367,7 @@ struct controller_impl {
          throw;
       }
 
-      //commit( fork_head->dpos_irreversible_blocknum ); // redundant
+      //kv_db.commit( fork_head->dpos_irreversible_blocknum ); // redundant
```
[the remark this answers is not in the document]
revl: Done

libraries/chain/controller.cpp
```diff
@@ -3553,7 +3131,7 @@ chain_id_type controller::extract_chain_id(snapshot_reader& snapshot) {
    });
 
    // check if this is a legacy version of the snapshot, which has a genesis state instead of chain id
-   fc::optional<genesis_state> genesis = controller_impl::extract_legacy_genesis_state(snapshot, header.version);
+   fc::optional<genesis_state> genesis = extract_legacy_genesis_state(snapshot, header.version);
```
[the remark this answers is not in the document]
revl: Yes, I moved `extract_legacy_genesis_state()` to `combined_database.hpp/cpp`.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,499 @@
...
+   combined_database::combined_database(backing_store_type backing_store_, chainbase::database& chain_db,
+                                        const std::string& rocksdb_path, bool rocksdb_create_if_missing,
+                                        uint32_t rocksdb_threads, int rocksdb_max_open_files)
+       : backing_store(backing_store_), db(chain_db),
+         kv_database(rocksdb_path.c_str(), rocksdb_create_if_missing, rocksdb_threads, rocksdb_max_open_files),
+         kv_undo_stack(kv_database, rocksdb_undo_prefix) {}
+
+   void combined_database::set_backing_store(backing_store_type backing_store) {
+      if (backing_store == backing_store_type::ROCKSDB) {
+         if (db.get<kv_db_config_object>().using_rocksdb_for_disk)
```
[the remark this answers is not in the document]
revl: Introducing a variable of backing_store_type would require declaring a pair of functions to convert it to and from variant. Not sure if that's necessary. I don't think we'll have another type of backing store anytime soon. In the meantime, I renamed `using_rocksdb_for_disk` to `using_rocksdb_for_backing_store`.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,499 @@
...
+         cb_session        = std::move(src.cb_session);
+         kv_undo_stack     = src.kv_undo_stack;
+         src.type          = session_type::no_op;
+         src.kv_undo_stack = nullptr;
+      }
+      return *this;
+   }
+
+   void combined_session::push() {
+      switch (type) {
```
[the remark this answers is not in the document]
revl: When session_type is `chainbase_only`, push/squash/undo should not look like the `chainbase_and_chain_kv` path because that path terminates the application in case of any exceptions. That would change the current behavior of nodeos, which is to let exceptions through. Additionally, session_type is not boolean. It has three possible values. The `no_op` value allowed `combined_session` to absorb both `maybe_session` and `optional<combined_session>` elsewhere in the code.

libraries/chain/include/eosio/chain/combined_database.hpp
revl: I didn't add this operator, and it was used by the `transaction_context` constructor. But because it's a constructor, it was easy to switch from assignment to construction, so I deleted the assignment operator. Sure, we can always add it back when needed.

libraries/chain/combined_database.cpp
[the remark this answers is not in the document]
brianjohnson5972: It should be called backing_store and be an enum.  The whole intent was that they would support other "level" databases in the future and we don't want to have to have to migrate from this flag to a new flag in the future.

libraries/chain/combined_database.cpp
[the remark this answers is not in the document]
brianjohnson5972: Why are we changing the behavior of cb_session for chainbase_and_chain_kv?  I want to make sure @b1bart and @larryk85 agree.  The same issue was brought up in Lin's PR, but I am not sure if it ever was resolved.

And for no_op is it any different than not processing because of 2 nullptrs, aside from having to check the nullptrs?

libraries/chain/combined_database.cpp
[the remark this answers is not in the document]
revl: I changed it to the backing_store enum.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+      switch (backing_store) {
+         case backing_store_type::ROCKSDB:
+            return create_kv_rocksdb_context(kv_database, kv_undo_stack, receiver, resource_manager, limits);
+         case backing_store_type::NATIVE: return create_kv_chainbase_context(db, receiver, resource_manager, limits);
+         default: EOS_ASSERT(false, action_validate_exception, "Unknown backing store.");
+      }
+   }
+
+   void combined_database::add_to_snapshot(
+         const eosio::chain::snapshot_writer_ptr& snapshot, const eosio::chain::block_state& head,
```
larryk85, line 170: this is a bit difficult to read.  I'm fine with punching these out to new lines and aligning the variable, but I would do it for all of them then.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+         const eosio::chain::resource_limits::resource_limits_manager& resource_limits) const {
+
+      snapshot->write_section<chain_snapshot_header>(
+            [this](auto& section) { section.add_row(chain_snapshot_header(), db); });
+
+      snapshot->write_section<block_state>(
+            [this, &head](auto& section) { section.template add_row<block_header_state>(head, db); });
+
+      eosio::chain::controller_index_set::walk_indices([this, &snapshot](auto utils) {
+         using value_t = typename decltype(utils)::index_t::value_type;
```
larryk85, line 181: `utils` in this context will always copy.  Is that what is intended?  If not change to `auto&` and `std::decay_t<decltype(utils)>`.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+         if (std::is_same<value_t, database_header_object>::value) {
+            return;
+         }
+
+         // skip the kv_db_config as it only determines where the kv-database is stored
+         if (std::is_same_v<value_t, kv_db_config_object>) {
+            return;
+         }
+
+#warning TODO: Revisit this and adpat to new design
```
larryk85, line 198: misspelled `adapt`

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+                    ("snapshot_head_block", snapshot_head_block)("block_log_first_num",
+                                                                 blog_start)("block_log_last_num", blog_end));
+
+         fork_db.reset(head_header_state);
+         head                = fork_db.head();
+         snapshot_head_block = head->block_num;
+      }
+
+      controller_index_set::walk_indices([this, &snapshot, &header](auto utils) {
+         using value_t = typename decltype(utils)::index_t::value_type;
```
larryk85, line 288: same as above.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+         });
+      });
+
+      add_contract_tables_to_snapshot(snapshot);
+
+      authorization.add_to_snapshot(snapshot);
+      resource_limits.add_to_snapshot(snapshot);
+   }
+
+   void combined_database::read_from_snapshot(const snapshot_reader_ptr& snapshot, uint32_t blog_start,
```
larryk85, line 244: Same readability concerns as `add_to_snapshot`.

libraries/chain/controller.cpp
```diff
@@ -233,8 +148,7 @@ struct controller_impl {
    std::function<void()>               shutdown;
    chainbase::database                 db;
    chainbase::database                 reversible_blocks; ///< a special database to persist blocks that have successfully been applied but are still reversible
-   b1::chain_kv::database              kv_database;
-   b1::chain_kv::undo_stack            kv_undo_stack;
+   combined_database                   kv_db;
```
larryk85, line 151: I'm confused with this too.  It seems as if you don't need `db` anymore and would just use `combined_database db;`. I'm sure why 1) combined_database is only being used for the rocksdb side and 2) why we still have both, that is the point of the combined_database. Or am I missing something?

libraries/chain/include/eosio/chain/combined_database.hpp
```diff
@@ -14,77 +36,88 @@
...
-               cb_session = std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true));
-               kv_undo_stack.push(false);
-            }
-            FC_LOG_AND_RETHROW()
-         }
-         CATCH_AND_EXIT_DB_FAILURE()
-      }
+   class combined_database {
+    public:
+      combined_database(backing_store_type backing_store, chainbase::database& chain_db,
```
larryk85, line 75: Some of the same readability concerns as the others.

libraries/chain/include/eosio/chain/backing_store.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <fc/reflect/variant.hpp>
+#include <stdexcept>
+
+namespace eosio { namespace chain {
+
+   enum class backing_store_type {
+      NATIVE, // A name for regular users. Uses Chainbase.
+      ROCKSDB
```
larryk85, line 10: We shouldn't be calling this `ROCKSDB` either, it's not consistent with the naming used everywhere else in your code. `CHAIN_KV` would be a better fit.

libraries/chain/include/eosio/chain/combined_database.hpp
```diff
@@ -14,77 +36,88 @@
...
+                              eosio::chain::fork_database& fork_db, eosio::chain::block_state_ptr& head,
+                              uint32_t& snapshot_head_block, const eosio::chain::chain_id_type& chain_id);
+
+    private:
+      void add_contract_tables_to_snapshot(const snapshot_writer_ptr& snapshot) const;
+      void read_contract_tables_from_snapshot(const snapshot_reader_ptr& snapshot);
+
+      backing_store_type       backing_store;
+      chainbase::database&     db;
+      b1::chain_kv::database   kv_database;
```
larryk85, line 115: If you utilized an `std::variant` for the db and kv_db then you could do away with the explicit backing_store_type and variable.

libraries/chain/controller.cpp
brianjohnson5972: We have both because the native data (like account_object, etc) is still in chainbase, so both need to be managed when we have a chain_kv.

libraries/chain/include/eosio/chain/backing_store.hpp
brianjohnson5972: EOS is all about mixed metaphors, but I thought the idea was that chain_kv could be designed to work with several databases.  Right now it may be tied all together, but eventually, chain_kv could be loaded for any number of databases, but this backing store will determine which type of database you are storing in.

libraries/chain/combined_database.cpp
revl: I'm using a .clang-format file, which I found in a sibling directory. The resulting formatting gives me the creeps. I think it's time to discuss the formatting rules that we use.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+         }
+         FC_LOG_AND_RETHROW()
+      }
+      CATCH_AND_EXIT_DB_FAILURE()
+   }
+
+   std::unique_ptr<kv_context> combined_database::create_kv_context(name receiver, kv_resource_manager resource_manager,
+                                                                    const kv_database_config& limits) {
+      switch (backing_store) {
+         case backing_store_type::ROCKSDB:
```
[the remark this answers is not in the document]
revl: This is horrible, I completely agree. Our clang-format rules need work.

libraries/chain/include/eosio/chain/backing_store.hpp
```diff
@@ -0,0 +1,34 @@
+#pragma once
+
+#include <fc/reflect/variant.hpp>
+#include <stdexcept>
+
+namespace eosio { namespace chain {
+
+   enum class backing_store_type {
+      NATIVE, // A name for regular users. Uses Chainbase.
```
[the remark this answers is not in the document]
revl: Renamed to `CHAINBASE`.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+   combined_session::combined_session(chainbase::database& cb_database, b1::chain_kv::undo_stack& kv_undo_stack)
+       : kv_undo_stack{ &kv_undo_stack } {
+      try {
+         try {
+            cb_session = std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true));
+            kv_undo_stack.push(false);
+         }
+         FC_LOG_AND_RETHROW()
+      }
+      CATCH_AND_EXIT_DB_FAILURE()
```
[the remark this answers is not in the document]
revl: I replaced `exit` with `std::abort` per our conversation in Slack and created a ticket to address the inconsistency in error handling.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+         } else {
+            try {
+               try {
+                  cb_session->push();
+                  cb_session    = nullptr;
+                  kv_undo_stack = nullptr;
+               }
+               FC_LOG_AND_RETHROW()
+            }
+            CATCH_AND_EXIT_DB_FAILURE()
```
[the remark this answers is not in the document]
revl: As per our Slack thread, this will remain as is for now.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+         }
+      }
+   }
+
+   // chainlib reserves prefixes 0x10 - 0x2F.
+   static const std::vector<char> rocksdb_undo_prefix{ 0x10 };
```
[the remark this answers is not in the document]
revl: There are two more prefixes below this one. Changing just one introduces an inconsistency. Changing all three makes some of the references less readable. I don't think this presents a significant performance degradation, especially given that we'll replace chain_kv and this code will have to be reworked anyway.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+       : backing_store(backing_store_), db(chain_db),
+         kv_database(rocksdb_path.c_str(), rocksdb_create_if_missing, rocksdb_threads, rocksdb_max_open_files),
+         kv_undo_stack(kv_database, rocksdb_undo_prefix) {}
+
+   void combined_database::set_backing_store(backing_store_type backing_store) {
+      if (backing_store == backing_store_type::ROCKSDB) {
+         if (db.get<kv_db_config_object>().backing_store == backing_store_type::ROCKSDB)
+            return;
+         auto& idx = db.get_index<kv_index, by_kv_key>();
+         auto  it  = idx.lower_bound(boost::make_tuple(kvdisk_id, name{}, std::string_view{}));
```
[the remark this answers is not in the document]
revl: Yes, there's another PR waiting on this one that will take care of `kvdisk`.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,470 @@
...
+
+   void combined_database::undo() {
+      try {
+         try {
+            db.undo();
+            kv_undo_stack.undo(false);
+         }
+         FC_LOG_AND_RETHROW()
+      }
+      CATCH_AND_EXIT_DB_FAILURE()
```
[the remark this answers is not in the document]
revl: Per our Slack thread, it will be handled later.

libraries/chain/combined_database.cpp
revl: Fixed

libraries/chain/combined_database.cpp
revl: `clang-format` hasn't been configured properly. I don't know what to do about it.

libraries/chain/include/eosio/chain/combined_database.hpp
revl: We need to come together and unify and improve our `.clang-format` files.

libraries/chain/include/eosio/chain/backing_store.hpp
revl: Done

libraries/chain/include/eosio/chain/combined_database.hpp
revl: As Brian said above, chainbase is used even when KV entries are stored in chain_kv.

libraries/chain/combined_database.cpp
revl: Because all four occurrences of similar constructs pass `utils` by value, I assume that was the author's intent.

libraries/chain/combined_database.cpp
revl: Yes, there are four such constructs in this file. I assume they're intentional.

libraries/chain/combined_database.cpp
larryk85: Manually edit it. I don't think many people on our team are relying on clang-format.

libraries/chain/include/eosio/chain/combined_database.hpp
larryk85: As a code base and group, we've treated any of the clang-format stuff as purely optional and only to be applied to the code changes that are being made.  Also, if those rules result in bad formatting, then those that do use clang-format will just fix it up as they go along.  There are two issues with clang-format, 1) our codebase is not consistent and we don't particularly care about that. 2) it can format things terribly and there are certain rules that you are just not going to be able to do (and I really don't want to see disable clang formats all over the place).  

For formatting the code, there are only a few rules you need.  Stay consistent with what is around it and make it as readable as possible, at the end of the day that is all we care about.

brianjohnson5972 (approved):

libraries/chain/combined_database.cpp
revl: I changed the formatting manually.

libraries/chain/include/eosio/chain/combined_database.hpp
revl: I changed the formatting manually.

PR #9427, opened 2020-08-21
Fix static initialization problem

huangminghuang:
## Change Description
PR 9416 introduced a "static initialization order fiasco" bug to the system. It caused the unit tests to fail on non-unity build. This PR fixes the problem.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -6,6 +6,7 @@
 #include <eosio/chain/wasm_eosio_constraints.hpp>
 //eos-vm includes
 #include <eosio/vm/backend.hpp>
+#include "host_function_registrator.cpp"
```
larryk85, line 9: I'm fine with including at this layer.  But, I am strongly against including a cpp file.  So, just move this back to the includes and include like normal and rename to .hpp.

PR #9429, opened 2020-08-24
Abi kv nodeos

kimjh2005:
## Change Description

The ABI Serializer is defined at libraries/chain/abi_serializer.cpp and various supporting headers. 
https://blockone.atlassian.net/browse/EPE-209

-kv_tables defined in abi_def.
-custom to_variant() and from_variant() defined as the existing codes treat them as an array.
-added variant_to_binary and binary_to_add to abi_serializer.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -243,13 +243,12 @@ namespace eosio { namespace chain {
       if (size_t len = str.size(); len > 13)
          return false;
 
-      for (const char c : str)
-      {
-         if (c >= 'a' && c <= 'z')
+      for( const char c : str ) {
+         if( c >= 'a' && c <= 'z' )
             continue;
-         else if (c >= '1' && c <= '5')
+         else if( c >= '1' && c <= '5' )
             continue;
-         else if (c == '.')
+         else if ( c == '.' )
             continue;
          else
             return false;
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -264,8 +264,7 @@ namespace eosio { namespace chain {
       if( typedefs.find(type) != typedefs.end() ) return _is_type(typedefs.find(type)->second, ctx);
       if( structs.find(type) != structs.end() ) return true;
       if( variants.find(type) != variants.end() ) return true;
-      if (is_string_valid_name(rtype))
-      {
+      if( is_string_valid_name(rtype) ) {
          if( kv_tables.find(name(type)) != kv_tables.end() ) return true;
       }
       return false;
@@ -392,7 +391,6 @@ namespace eosio { namespace chain {
       auto rtype = resolve_type(type);
       auto ftype = fundamental_type(rtype);
       auto btype = built_in_types.find(ftype );
-      
       if( btype != built_in_types.end() ) {
          try {
             return btype->second.first(stream, is_array(rtype), is_optional(rtype), ctx.get_yield_function());
@@ -430,29 +428,26 @@ namespace eosio { namespace chain {
          } EOS_RETHROW_EXCEPTIONS( unpack_exception, "Unable to unpack presence flag of optional '${p}'", ("p", ctx.get_path_string()) )
          return flag ? _binary_to_variant(ftype, stream, ctx) : fc::variant();
       } else {
-         if(kv_tables.size() > 0 && is_string_valid_name(rtype)) {
-            auto kv_itr = kv_tables.find(name(rtype));
-            if (kv_itr != kv_tables.end() )
-            {
-               auto &kv_table = kv_itr->second;
-                  return _binary_to_variant(kv_table.type, stream, ctx);
-               }
-            }
-
-           auto v_itr = variants.find(rtype);
-           if (v_itr != variants.end())
-           {
-              ctx.hint_variant_type_if_in_array(v_itr);
-              fc::unsigned_int select;
-              try
-              {
-                 fc::raw::unpack(stream, select);
-              } EOS_RETHROW_EXCEPTIONS( unpack_exception, "Unable to unpack tag of variant '${p}'", ("p", ctx.get_path_string()) )
+         auto v_itr = variants.find(rtype);
+         if( v_itr != variants.end() ) {
+            ctx.hint_variant_type_if_in_array(v_itr);
+            fc::unsigned_int select;
+            try {
+               fc::raw::unpack(stream, select);
+            } EOS_RETHROW_EXCEPTIONS( unpack_exception, "Unable to unpack tag of variant '${p}'", ("p", ctx.get_path_string()) )
             EOS_ASSERT( (size_t)select < v_itr->second.types.size(), unpack_exception,
                         "Unpacked invalid tag (${select}) for variant '${p}'", ("select", select.value)("p",ctx.get_path_string()) );
             auto h1 = ctx.push_to_path( impl::variant_path_item{ .variant_itr = v_itr, .variant_ordinal = static_cast<uint32_t>(select) } );
             return vector<fc::variant>{v_itr->second.types[select], _binary_to_variant(v_itr->second.types[select], stream, ctx)};
-           }
+         }
+
+         if(kv_tables.size() > 0 && is_string_valid_name(rtype)) {
+            auto kv_itr = kv_tables.find(name(rtype));
+            if( kv_itr != kv_tables.end() ) {
+               auto &kv_table = kv_itr->second;
+                  return _binary_to_variant(kv_table.type, stream, ctx);
+            }
+         }
       }
 
       fc::mutable_variant_object mvo;
@@ -588,9 +583,7 @@ namespace eosio { namespace chain {
                auto& kv_table = kv_itr->second;
                _variant_to_binary( kv_table.type, var, ds, ctx );
            }
-         }
-         else
-         {
+         } else {
             EOS_THROW(invalid_type_inside_abi, "Unknown type ${type}", ("type", ctx.maybe_shorten(type)));
          }
       } else {
@@ -829,7 +822,6 @@ namespace eosio { namespace chain {
             const auto& str = item.variant_itr->first;
             output_name( s, str, shorten_names );
          }
-
       };
 
       struct path_item_type_visitor {
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -220,8 +220,6 @@ namespace eosio { namespace chain {
       return kv_tables.find(name(type)) != kv_tables.end();
    }
 
-
-
    std::string_view abi_serializer::fundamental_type(const std::string_view& type)const {
       if( is_array(type) ) {
          return type.substr(0, type.size()-2);
@@ -246,6 +244,10 @@ namespace eosio { namespace chain {
       if( typedefs.find(type) != typedefs.end() ) return _is_type(typedefs.find(type)->second, ctx);
       if( structs.find(type) != structs.end() ) return true;
       if( variants.find(type) != variants.end() ) return true;
+      if (is_string_valid_name(rtype))
+      {
+         if( kv_tables.find(name(type)) != kv_tables.end() ) return true;
+      }
       return false;
    }
 
@@ -305,13 +307,11 @@ namespace eosio { namespace chain {
         EOS_ASSERT(_is_type(t.second, ctx), invalid_type_inside_abi, "${type}", ("type",impl::limit_size(t.second)) );
       } FC_CAPTURE_AND_RETHROW( (t)  ) }
 
-      /*
       for( const auto& kt : kv_tables ) { try {
         ctx.check_deadline();
-        //EOS_ASSERT(_is_type(kt.second.primary_index.type, ctx), invalid_type_inside_abi, "${type}", ("type",impl::limit_size(kt.second.primary_index.type)) );
+        //EOS_ASSERT(_is_type(kt.second.type, ctx), invalid_type_inside_abi, "${type}", ("type",impl::limit_size(kt.second.type))) );
+        //EOS_ASSERT( kt.second.primary_index.type.size() > 0, invalid_type_inside_abi, "missing primary index$ {p}", ("p",kt.first.to_string());
       } FC_CAPTURE_AND_RETHROW( (kt)  ) }
-
-	  */
 
       for( const auto& r : action_results ) { try {
         ctx.check_deadline();
@@ -371,6 +371,7 @@ namespace eosio { namespace chain {
       auto rtype = resolve_type(type);
       auto ftype = fundamental_type(rtype);
       auto btype = built_in_types.find(ftype );
+      
       if( btype != built_in_types.end() ) {
          try {
             return btype->second.first(stream, is_array(rtype), is_optional(rtype), ctx.get_yield_function());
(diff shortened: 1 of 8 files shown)


libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -524,9 +524,11 @@ namespace eosio { namespace chain {
                                     3015016, "ABI has an unsupported version" )
       FC_DECLARE_DERIVED_EXCEPTION( duplicate_abi_action_results_def_exception,  abi_exception,
                                     3015017, "Duplicate action results definition in the ABI" )
+      FC_DECLARE_DERIVED_EXCEPTION(duplicate_abi_kv_table_def_exception, abi_exception,
+                                   3015018, "Duplicate kv_table definition in the ABI")
 
-   FC_DECLARE_DERIVED_EXCEPTION( contract_exception,           chain_exception,
-                                 3160000, "Contract exception" )
+      FC_DECLARE_DERIVED_EXCEPTION(contract_exception, chain_exception,
+                                   3160000, "Contract exception")
```
heifner, line 531: The indentation here should be restored to what it was.

libraries/chain/abi_serializer.cpp
```diff
@@ -228,13 +237,35 @@ namespace eosio { namespace chain {
          return type;
    }
 
+   // true if std::string can be converted to name
+   bool abi_serializer::is_string_valid_name(std::string_view str) const
+   {
+      if (size_t len = str.size(); len > 13)
```
TimothyBanks, line 243: Curious why we need to store the size in a variable and then do the comparison instead of just comparing the value returned from `str.size()`

libraries/chain/abi_serializer.cpp
```diff
@@ -228,13 +237,35 @@ namespace eosio { namespace chain {
          return type;
    }
 
+   // true if std::string can be converted to name
+   bool abi_serializer::is_string_valid_name(std::string_view str) const
+   {
```
TimothyBanks, line 242: I think this needs to be moved to the line above.

libraries/chain/abi_serializer.cpp
```diff
@@ -228,13 +237,35 @@ namespace eosio { namespace chain {
...
+            continue;
+         else if( c >= '1' && c <= '5' )
+            continue;
+         else if ( c == '.' )
+            continue;
+         else
+            return false;
+      }
+      return true;
+   }
```
heifner, line 257: This should be moved to `name.hpp` and reuse existing code.

libraries/chain/abi_serializer.cpp
```diff
@@ -401,6 +439,14 @@ namespace eosio { namespace chain {
             auto h1 = ctx.push_to_path( impl::variant_path_item{ .variant_itr = v_itr, .variant_ordinal = static_cast<uint32_t>(select) } );
             return vector<fc::variant>{v_itr->second.types[select], _binary_to_variant(v_itr->second.types[select], stream, ctx)};
          }
+
+         if(kv_tables.size() > 0 && is_string_valid_name(rtype)) {
```
TimothyBanks, line 443: Nitpick:  I think `!kv_tables.empty()` is the idiomatic approach instead of `kv_tables.size() > 0`.  It used to be that `size()` wasn't guaranteed to be a constant operation.  I don't know if there are other compelling reasons now other than possibly it being more concise.

libraries/chain/abi_serializer.cpp
```diff
@@ -170,6 +174,7 @@ namespace eosio { namespace chain {
       EOS_ASSERT( structs.size() == abi.structs.size(), duplicate_abi_struct_def_exception, "duplicate struct definition detected" );
       EOS_ASSERT( actions.size() == abi.actions.size(), duplicate_abi_action_def_exception, "duplicate action definition detected" );
       EOS_ASSERT( tables.size() == abi.tables.size(), duplicate_abi_table_def_exception, "duplicate table definition detected" );
+      EOS_ASSERT( kv_tables.size() == abi.kv_tables.value.size(), duplicate_abi_kv_table_def_exception, "duplicate table definition detected" );
```
heifner, line 177: "duplicate kv table definition detected"

libraries/chain/abi_serializer.cpp
```diff
@@ -228,13 +237,35 @@ namespace eosio { namespace chain {
...
+   }
+
    bool abi_serializer::_is_type(const std::string_view& rtype, impl::abi_traverse_context& ctx )const {
       auto h = ctx.enter_scope();
       auto type = fundamental_type(rtype);
       if( built_in_types.find(type) != built_in_types.end() ) return true;
       if( typedefs.find(type) != typedefs.end() ) return _is_type(typedefs.find(type)->second, ctx);
       if( structs.find(type) != structs.end() ) return true;
       if( variants.find(type) != variants.end() ) return true;
+      if( is_string_valid_name(rtype) ) {
```
heifner, line 266: This should be `type` not `rtype`, right?

libraries/chain/abi_serializer.cpp
```diff
@@ -294,6 +325,13 @@ namespace eosio { namespace chain {
         EOS_ASSERT(_is_type(t.second, ctx), invalid_type_inside_abi, "${type}", ("type",impl::limit_size(t.second)) );
       } FC_CAPTURE_AND_RETHROW( (t)  ) }
 
+      for( const auto& kt : kv_tables ) { try {
+        ctx.check_deadline();
+        EOS_ASSERT(_is_type(kt.second.type, ctx), invalid_type_inside_abi,
+                   "Invalid reference in struct ${type}", ("type", impl::limit_size(kt.second.type)));
+        EOS_ASSERT( kt.second.primary_index.type.size() > 0, invalid_type_inside_abi, "missing primary index$ {p}", ("p",kt.first.to_string()));
```
heifner, line 332: `kt.first` needs to be restricted by limit_size

libraries/chain/abi_serializer.cpp
```diff
@@ -529,6 +575,16 @@ namespace eosio { namespace chain {
          } else {
             EOS_THROW( pack_exception, "Unexpected input encountered while processing struct '${p}'", ("p",ctx.get_path_string()) );
          }
+      } else if( var.is_object() ) {
+         if( kv_tables.size() > 0 && is_string_valid_name(rtype) ) {
```
TimothyBanks, line 579: Same thing here about size.

libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -70,6 +71,45 @@ struct table_def {
...
+struct primary_index_def {
+   primary_index_def() = default;
+   primary_index_def(const index_name &name, const type_name &type)
+      : name(name), type(type)
+   {}
+
+   index_name name;                // the name of primary index
+   type_name  type;                // the kind of index
+
+   bool operator==(const primary_index_def &other) const {
```
TimothyBanks, line 83: Would it be useful to overload the `operator!=` as well?

libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -70,6 +71,45 @@ struct table_def {
...
+struct secondary_index_def
+{
+   secondary_index_def() = default;
+   secondary_index_def(const type_name &type)
+      : type(type)
+   {}
+
+   type_name  type;              // the kind of index
+
+   bool operator==(const secondary_index_def &other) const {
```
TimothyBanks, line 97: Would it be useful to overload the `operator!=` as well?

libraries/chain/abi_serializer.cpp
```diff
@@ -294,6 +325,13 @@ namespace eosio { namespace chain {
         EOS_ASSERT(_is_type(t.second, ctx), invalid_type_inside_abi, "${type}", ("type",impl::limit_size(t.second)) );
       } FC_CAPTURE_AND_RETHROW( (t)  ) }
 
+      for( const auto& kt : kv_tables ) { try {
+        ctx.check_deadline();
+        EOS_ASSERT(_is_type(kt.second.type, ctx), invalid_type_inside_abi,
+                   "Invalid reference in struct ${type}", ("type", impl::limit_size(kt.second.type)));
+        EOS_ASSERT( kt.second.primary_index.type.size() > 0, invalid_type_inside_abi, "missing primary index$ {p}", ("p",kt.first.to_string()));
+      } FC_CAPTURE_AND_RETHROW( (kt)  ) }
```
heifner, line 333: I think it better not to capture `kt` here.

libraries/chain/abi_serializer.cpp
```diff
@@ -401,6 +439,14 @@ namespace eosio { namespace chain {
             auto h1 = ctx.push_to_path( impl::variant_path_item{ .variant_itr = v_itr, .variant_ordinal = static_cast<uint32_t>(select) } );
             return vector<fc::variant>{v_itr->second.types[select], _binary_to_variant(v_itr->second.types[select], stream, ctx)};
          }
+
+         if(kv_tables.size() > 0 && is_string_valid_name(rtype)) {
+            auto kv_itr = kv_tables.find(name(rtype));
+            if( kv_itr != kv_tables.end() ) {
+               auto &kv_table = kv_itr->second;
+                  return _binary_to_variant(kv_table.type, stream, ctx);
```
heifner, line 447: fix indentation

libraries/chain/abi_serializer.cpp
```diff
@@ -228,13 +237,35 @@ namespace eosio { namespace chain {
          return type;
    }
 
+   // true if std::string can be converted to name
+   bool abi_serializer::is_string_valid_name(std::string_view str) const
```
dimas1185, line 241: what about moving this code to libraries/chain/include/eosio/chain/name.hpp ?

libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -165,16 +214,80 @@ void from_variant(const fc::variant& v, eosio::chain::may_not_exist<T>& e) {
...
+    }
+}
 } // namespace fc
 
 FC_REFLECT( eosio::chain::type_def                         , (new_type_name)(type) )
 FC_REFLECT( eosio::chain::field_def                        , (name)(type) )
 FC_REFLECT( eosio::chain::struct_def                       , (name)(base)(fields) )
 FC_REFLECT( eosio::chain::action_def                       , (name)(type)(ricardian_contract) )
 FC_REFLECT( eosio::chain::table_def                        , (name)(index_type)(key_names)(key_types)(type) )
+FC_REFLECT( eosio::chain::primary_index_def                      , (name)(type) )
```
dimas1185, line 285: fix indentation

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -176,6 +180,7 @@ namespace impl {
       map<type_name, variant_def>::const_iterator variant_itr;
    };
 
+
```
dimas1185, line 183: remove unnecessary empty lines

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -219,6 +224,7 @@ namespace impl {
       void hint_struct_type_if_in_array( const map<type_name, struct_def>::const_iterator& itr );
       void hint_variant_type_if_in_array( const map<type_name, variant_def>::const_iterator& itr );
 
+
```
dimas1185, line 227: remove unnecessary empty lines

unittests/abi_tests.cpp
```diff
@@ -26,7 +26,7 @@ using namespace chain;
 
 BOOST_AUTO_TEST_SUITE(abi_tests)
 
-fc::microseconds max_serialization_time = fc::seconds(1); // some test machines are very slow
+fc::microseconds max_serialization_time = fc::seconds(10000); // some test machines are very slow
```
dimas1185, line 29: 10'000 seconds is almost 3 hours. do you really need here so long timeout?

libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -524,9 +524,11 @@ namespace eosio { namespace chain {
...
       FC_DECLARE_DERIVED_EXCEPTION( duplicate_abi_action_results_def_exception,  abi_exception,
                                     3015017, "Duplicate action results definition in the ABI" )
+      FC_DECLARE_DERIVED_EXCEPTION( duplicate_abi_kv_table_def_exception, abi_exception,
+                                    3015018, "Duplicate kv_table definition in the ABI" )
 
-   FC_DECLARE_DERIVED_EXCEPTION( contract_exception,           chain_exception,
-                                 3160000, "Contract exception" )
+      FC_DECLARE_DERIVED_EXCEPTION( contract_exception, chain_exception,
+                                    3160000, "Contract exception" )
       FC_DECLARE_DERIVED_EXCEPTION( invalid_table_payer,             contract_exception,
```
heifner, line 532: Please revert this change.

unittests/abi_tests.cpp
heifner: This change should be reverted.

libraries/chain/include/eosio/chain/abi_def.hpp
```diff
@@ -165,16 +214,80 @@ void from_variant(const fc::variant& v, eosio::chain::may_not_exist<T>& e) {
...
+
+template<typename ST, typename T>
+ST& operator << (ST& s, const eosio::chain::kv_tables_as_object<T>& v) {
+   raw::pack(s, v.value);
+   return s;
+}
+
+template<typename ST, typename T>
+ST& operator >> (ST& s, eosio::chain::kv_tables_as_object<T>& v) {
+   if (s.remaining())
```
dimas1185, line 226: why are you making this object optional? can you point to the code where it is used?
few lines above this code there is similar method with `may_not_exist` template that is designed for such situations. so if you have optional object, what about wrapping it into `stream >> may_no_exist<object>` ?

libraries/chain/abi_serializer.cpp
```diff
@@ -228,13 +237,35 @@ namespace eosio { namespace chain {
...
+   {
+      if (size_t len = str.size(); len > 13)
+         return false;
+
+      for( const char c : str ) {
+         if( c >= 'a' && c <= 'z' )
```
dimas1185, line 247: check `name.hpp`
this is not full constraint. 13th character can't have same range. it needs special check.

libraries/chain/include/eosio/chain/abi_def.hpp
kimjh2005: It is not optional.  In libraries/fc/include/fc/variant.hpp,  from_variant treats all var as an array. In kv_table it is varaint_object.

    454    template<typename K, typename T>
    455    void from_variant( const variant& var,  std::map<K, T>& vo )
    456    {
    457       const variants& vars = var.get_array();

dimas1185: So if it is not optional, any reason to call ‘s.remaining()’?

kimjh2005: No, I will remove the remaining() check

libraries/chain/abi_serializer.cpp
jeffreyssmith2nd: Also, shouldn't we already have a helper function somewhere that does this? I hope we're not reimplementing this logic all over the code.

libraries/chain/abi_serializer.cpp
```diff
@@ -294,6 +306,13 @@ namespace eosio { namespace chain {
         EOS_ASSERT(_is_type(t.second, ctx), invalid_type_inside_abi, "${type}", ("type",impl::limit_size(t.second)) );
       } FC_CAPTURE_AND_RETHROW( (t)  ) }
 
+      for( const auto& kt : kv_tables ) { try {
+        ctx.check_deadline();
+        EOS_ASSERT(_is_type(kt.second.type, ctx), invalid_type_inside_abi,
+                   "Invalid reference in struct ${type}", ("type", impl::limit_size(kt.second.type)));
+        EOS_ASSERT( kt.second.primary_index.type.size() > 0, invalid_type_inside_abi, "missing primary index$ {p}", ("p",impl::limit_size(kt.first.to_string())));
+      } FC_CAPTURE_AND_RETHROW( (impl::limit_size(kt.first.to_string()))) }
```
heifner, line 314: This does not produce valid json which can be problematic for things that report exceptions in json format.
If `kt.first` is of type `name` then the `limit_size` is not needed. 
I would just not capture anything here.

kimjh2005: removed try FC_CAPTURE_AND_RETHROW

libraries/chain/include/eosio/chain/abi_def.hpp
kimjh2005: removing the remaining breaks existing unit tests

libraries/chain/include/eosio/chain/exceptions.hpp
kimjh2005: restored the indentation

libraries/chain/abi_serializer.cpp
kimjh2005: moved to name.hpp

libraries/chain/abi_serializer.cpp
kimjh2005: checked the 13th character to be one of ".12345abcdefghij" and added tests to name_tests.

libraries/chain/include/eosio/chain/abi_def.hpp
dimas1185: so you may want to adjust tests. e.g. use `may_no_exist` or catch exception if it was expected

kimjh2005: When old abi which doesn't have kv_tables is passed and tried to read it gets an exception. The if(s.remaining()) is needed to backward compatibility with abi 1.0 and 1.1.
As it is not optional field in abi 1.2,  may_no_exist is not a good one to apply.

dimas1185: I approve it now but suggest you to learn about SFINAE technique. that way you don't need this object at all.
with SFINAE you can define your `to_variant/from_variant` that way:
`template<typename T, typename = enable_if_t<is_same_v<T, your_object_type>, void>>
void from_variant(const fc::variant& v, eosio::chain::may_no_exist<T>& o) {}`

PR #9431, opened 2020-08-24
Restrict the maximum number of open HTTP RPC requests

vzqzhang:
## Change Description
Added a new option for `http-plugin` : `http-max-in-flight-requests`. It restricts the maximum requests the flight can hold if --http-max-in-flight-requests = positive number. If the limit exceeds, it sends a 429 error to the client. if --http-max-in-flight-requests = -1, there is no limit so the feature is turned off. It protects non-producing nodeos against the flood of requests when it cannot get new blocks.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

Added a new http_plugin option --http-max-in-flight-requests. "--http-max-in-flight-requests = positive number" sets the limit of requests that the flight can hold. If the limit is exceeded, a 429 error is sent to the client. if --http-max-in-flight-requests = -1, the feature is turned off. The feature is turned off in default. The doc is: "Maximum number of requests http_plugin should use for processing http requests. 429 error response when exceeded."

A change is made for the http_plugin option http-max-bytes-in-flight-mb. When the limit is exceeded, the error coode is 429, instead of 503. The new doc is: "Maximum size in megabytes http_plugin should use for processing http requests. 429 error response when exceeded."

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -405,7 +405,7 @@ namespace eosio {
             }
 
             void update_connection(const std::string & body, int code) override {
-               _conn->set_body(std::move(body));
+               _conn->set_body(body);
                _conn->set_status( websocketpp::http::status_code::value( code ) );
                _conn->send_http_response();
             }
@@ -568,7 +568,7 @@ namespace eosio {
                   try {
                      std::string json = fc::json::to_string( *(*tracked_response), fc::time_point::now() + max_response_time );
                      auto tracked_json = make_in_flight(std::move(json), *this);
-                     abstract_conn_ptr->update_connection(*(*tracked_json) , code);
+                     abstract_conn_ptr->update_connection(std::move(*(*tracked_json)), code);
                   } catch( ... ) {
                      abstract_conn_ptr->handle_exception();
                   }
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -136,8 +136,7 @@ namespace eosio {
          virtual bool verify_max_requests_in_flight() = 0;
          virtual void handle_exception() = 0;
 
-         virtual const void* operator* () const = 0;
-         virtual void* operator* () = 0;
+         virtual void update_connection(const std::string & body, int code) = 0;
       };
 
       using abstract_conn_ptr = std::shared_ptr<abstract_conn>;
@@ -329,7 +328,7 @@ namespace eosio {
          }
 
          template<typename T>
-         void report_429_error( const T& con, string what) {
+         void report_429_error( const T& con, const std::string & what) {
             error_results::error_info ei;
             ei.code = websocketpp::http::status_code::too_many_requests;
             ei.name = "Busy";
@@ -405,20 +404,10 @@ namespace eosio {
                http_plugin_impl::handle_exception<T>(_conn);
             }
 
-            /**
-             * const accessor
-             * @return const reference to the contained _conn
-             */
-            const void* operator* () const override {
-               return (const void *) &_conn;
-            }
-
-            /**
-             * mutable accessor (can be moved frmo)
-             * @return mutable reference to the contained _conn
-             */
-            void* operator* () override {
-               return (void *) &_conn;
+            void update_connection(const std::string & body, int code) override {
+               _conn->set_body(std::move(body));
+               _conn->set_status( websocketpp::http::status_code::value( code ) );
+               _conn->send_http_response();
             }
 
             detail::connection_ptr<T> _conn;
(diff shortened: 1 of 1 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -512,23 +567,24 @@ namespace eosio {
...
                   return;
                }
 
                // post  back to an HTTP thread to to allow the response handler to be called from any thread
-               boost::asio::post( thread_pool->get_executor(), [this, con, code, tracked_response=std::move(tracked_response)]() {
+               boost::asio::post( thread_pool->get_executor(), [this, abstract_conn_ptr, code, tracked_response=std::move(tracked_response)]() {
                   try {
                      std::string json = fc::json::to_string( *(*tracked_response), fc::time_point::now() + max_response_time );
                      auto tracked_json = make_in_flight(std::move(json), *this);
+                     auto con = *(static_cast<detail::connection_ptr<T>*>(*(*abstract_conn_ptr)));
```
b1bart, line 582: instead of `static_cast` here and the associated `operator*` 's above, we should move the next three lines (set_body, set_status, send) into a `virtual` method on the abstract connection that takes in  `body` and `code` and does this for us.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -465,7 +520,7 @@ namespace eosio {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [this, priority, next_ptr=std::move(next_ptr)]( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) mutable {
                auto tracked_b = make_in_flight<string>(std::move(b), *this);
-               if (!conn->verify_max_bytes_in_flight()) {
+               if (!conn->verify_max_bytes_in_flight()  || !conn->verify_max_requests_in_flight()) {
```
b1bart, line 523: I don't think its necessary to call verify_max_requests_in_flight here, we establish this request's presence in the in-flight set on line 622 & 623

plugins/http_plugin/http_plugin.cpp
```diff
@@ -133,7 +133,10 @@ namespace eosio {
       struct abstract_conn {
          virtual ~abstract_conn() {}
          virtual bool verify_max_bytes_in_flight() = 0;
+         virtual bool verify_max_requests_in_flight() = 0;
          virtual void handle_exception() = 0;
+
+         virtual void update_connection(const std::string & body, int code) = 0;
```
b1bart, line 139: I think a more appropriate name for this method would be `send_response` 

As the responses can be quite large and the back-end supports it we do want to `std::move` the body.  @heifner is correct that the `std::move` does nothing for a `const std::string&` however, I think the fix went the wrong direction.

Instead, this method should take a `std::string` (no reference) and then `std::move` from it. 

The caller is already `std::move`'ing into this argument so, that should move it all the way from the caller to the backend without creating a copy.

vzqzhang: > I think a more appropriate name for this method would be `send_response`
> 
> As the responses can be quite large and the back-end supports it we do want to `std::move` the body. @heifner is correct that the `std::move` does nothing for a `const std::string&` however, I think the fix went the wrong direction.
> 
> Instead, this method should take a `std::string` (no reference) and then `std::move` from it.
> 
> The caller is already `std::move`'ing into this argument so, that should move it all the way from the caller to the backend without creating a copy.

Changed the function.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -689,6 +732,8 @@ namespace eosio {
              "The maximum body size in bytes allowed for incoming RPC requests")
             ("http-max-bytes-in-flight-mb", bpo::value<uint32_t>()->default_value(500),
              "Maximum size in megabytes http_plugin should use for processing http requests. 503 error response when exceeded." )
+            ("http-max-in-flight-requests", bpo::value<int32_t>()->default_value(-1),
+             "Maximum number of requests http_plugin should use for processing http requests. 503 error response when exceeded." )
```
heifner, line 736: This doc, is wrong right? We are sending error 429 not 503.

vzqzhang: > This doc, is wrong right? We are sending error 429 not 503.

It's the same as http-max-bytes-in-flight-mb. So we should change the dic for http-max-bytes-in-flight-mb as well?

heifner: yes

heifner: Make sure to update the PR description with the documentation changes.

vzqzhang: > Make sure to update the PR description with the documentation changes.

Updated the code. PR description is the "Change Description" above, right?

heifner: Specifically the "Documentation Additions" above. You need to list your new option with its doc and note that the description changed on the other one.

PR #9438, opened 2020-08-25
Added iterator store for db intrinsic secondary keys in RocksDB.

brianjohnson5972:
## Change Description
Added an iterator store for secondary key lookup of db instrinsics in a RocksDB backing store.  This has similar requirements to db_chainbase_iter_cache, except that for chainbase the table and secondary key info was stored in an external object in a multi-index container, while this storage is all internal to the cache. Also, changed db_chainbase_iter_cache to db_chainbase_iter_store to be more accurate.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+template<typename T>
+struct secondary_key {
+   int       table_ei;
+   T         secondary;
+   uint64_t  primary;
+   uint64_t  payer;
+};
+
+struct unique_table_hash {
```
TimothyBanks, line 22: I need my memory refreshed on this, but couldn't we add a partial specialization of `std::hash`?  I thought the standard allowed for partial specialization of class templates as long as it depended on a user defined type.

libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
```diff
@@ -0,0 +1,156 @@
...
+         return _iterator_to_object.size() - 1;
+      }
+
+   private:
+      void validate_object_iterator(int iterator, const char* explanation_for_no_end_iterators) const {
+         EOS_ASSERT( iterator != _invalid_iterator, invalid_table_iterator, "invalid iterator" );
+         EOS_ASSERT( iterator >= 0, table_operation_not_permitted, explanation_for_no_end_iterators );
+         EOS_ASSERT( (size_t)iterator < _iterator_to_object.size(), invalid_table_iterator, "iterator out of range" );
+      }
+      unordered_map<unique_table, int, unique_table_hash > _table_cache;
```
swatanabe-b1, line 144: `unordered_map` is vulnerable to DOS attacks caused by hash collisions.

libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
```diff
@@ -0,0 +1,156 @@
...
+struct secondary_key {
+   int       table_ei;
+   T         secondary;
+   uint64_t  primary;
+   uint64_t  payer;
+};
+
+struct unique_table_hash {
+   std::size_t operator()(const unique_table& t) const {
+      return std::hash<name>()(t.contract) ^ std::hash<name>()(t.scope) ^ std::hash<name>()(t.table);
```
swatanabe-b1, line 24: `contract == scope` is fairly common and will cause them to cancel.

libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
brianjohnson5972: dropping hash and switching to map, so this is a moot point

libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
brianjohnson5972: changing to map

libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
```diff
@@ -0,0 +1,156 @@
...
+   return lhs.table_ei == rhs.table_ei &&
+          lhs.secondary == rhs.secondary &&
+          lhs.primary == rhs.primary;
+   // ignoring payer for uniqueness
+}
+
+template<typename SecondaryKey>
+class db_key_value_iter_cache {
+   public:
+      const int _invalid_iterator = -1;
```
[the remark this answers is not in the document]
brianjohnson5972: In the interest of how the code looks, I went with a constexpr method, then we get compile time, but not having to use it in reference to the long name plus template type.  I'm glad to change it all the way if that is still preferred.

libraries/chain/include/eosio/chain/backing_store/db_key_value_iter_cache.hpp
brianjohnson5972: Please verify you are ok with secondary_key operator<

swatanabe-b1: operator< can be implemented using std::tie, which is easier to understand.

PR #9445, opened 2020-08-25
move is_string_valid_name to cpp file

huangminghuang:
## Change Description

eosio::chain::is_string_valid_name() is static function in the header file which caused it be generated in every translation units include that header file. This PR fixes the problem. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/name.hpp
+++ b/libraries/chain/include/eosio/chain/name.hpp
@@ -30,31 +30,7 @@ namespace eosio::chain {
    }
 
    // true if std::string can be converted to name
-   static bool is_string_valid_name(std::string_view str)
-   {
-      size_t slen = str.size();
-      if( slen > 13)
-         return false;
-
-      size_t len = (slen <= 12) ? slen : 12;
-      for( size_t i = 0; i < len; ++i ) {
-         char c = str[i];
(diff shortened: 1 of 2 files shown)


libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -30,31 +30,7 @@ namespace eosio::chain {
...
-         char c = str[12];
-         if ((c >= 'a' && c <= 'j') || (c >= '1' && c <= '5') || (c == '.'))
-            return true;
-         else
-            return false;
-      }
-
-      return true;
-   }
+   bool is_string_valid_name(std::string_view str);
```
heifner, line 33: What about making it a constexpr instead?

huangminghuang: make it constexpr only make if it is called with the input parameter can be known at compile time. I don't see any of callers of the function meet that criteria except those in unit tests.

heifner: Yeah, use case is likely not there.

PR #9449, opened 2020-08-26
Support for storing kv and db intrinsics in Chainbase or RocksDB.

brianjohnson5972:
## Change Description
Support for storing kv and db intrinsics in Chainbase or RocksDB.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -228,6 +262,9 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
 
    std::string default_wasm_runtime_str= eosio::chain::wasm_interface::vm_type_string(eosio::chain::config::default_wasm_runtime);
 
+   // It is optimal that number of threads equals to the number of cores
+   uint16_t default_rocksdb_threads = std::thread::hardware_concurrency();
```
heifner, line 266: This is not a good default. There are many other threads in play besides just rocksdb.

libraries/chain/backing_store/db_context.cpp
```diff
@@ -0,0 +1,134 @@
...
+std::string db_context::table_event(name code, name scope, name table, name qualifier) {
+   return STORAGE_EVENT_ID("${code}:${scope}:${table}:${qualifier}",
+                           ("code", code)
+                           ("scope", scope)
+                           ("table", table)
+                           ("qualifier", qualifier)
+   );
+}
+
+void db_context::write_insert_table(fc::logger& deep_mind_logger, uint32_t action_id, name code, name scope, name table, account_name payer) {
```
heifner, line 24: `write_` seems confusing to me. Mayb `log_` instead? `write_` makes me think this writes to the database.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,609 @@
...
+            }
+        }
+        FC_LOG_AND_RETHROW()
+      }
+      CATCH_AND_EXIT_DB_FAILURE()
+   }
+
+   combined_session::combined_session(combined_session&& src) noexcept
+       : cb_session(std::move(src.cb_session)), kv_undo_stack(std::move(src.kv_undo_stack)) {
+      kv_undo_stack = nullptr;
```
heifner, line 24: Should this be `src.kv_undo_stack` ? Seems some more tests are in order.

plugins/chain_plugin/chain_plugin.cpp
linhuang-blockone: We were discussing the same concern when put the hardware_concurrency here. We thought as this was configurable, it could be changed if it turns out to be bad. But we can just set the default to 1.

libraries/chain_kv/include/b1/chain_kv/chain_kv.hpp
```diff
@@ -605,7 +775,7 @@ struct write_session {
 }; // write_session
 
 // A view of the database with a restricted range (prefix). Implements part of
-// https://github.com/EOSIO/spec-repo/blob/master/esr_key_value_database_intrinsics.md,
+// https://github.com/EOSIO/spec-repo/blob/master/esr_key_value_database_intrinsics.md,class
```
heifner, line 778: oops

libraries/chain/include/eosio/chain/combined_database.hpp
```diff
@@ -0,0 +1,145 @@
...
+      void undo();
+
+    private:
+      std::unique_ptr<chainbase::database::session> cb_session    = {};
+      eosio::session::undo_stack<rocks_db_type>*     kv_undo_stack = nullptr;
+   };
+
+   class combined_database {
+    public:
+      combined_database(chainbase::database& chain_db);
```
heifner, line 86: Should be `explicit`

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,609 @@
...
+              kv_undo_stack->push();
+            }
+        }
+        FC_LOG_AND_RETHROW()
+      }
+      CATCH_AND_EXIT_DB_FAILURE()
+   }
+
+   combined_session::combined_session(combined_session&& src) noexcept
+       : cb_session(std::move(src.cb_session)), kv_undo_stack(std::move(src.kv_undo_stack)) {
```
heifner, line 23: Does not make any sense to move a pointer.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,609 @@
...
+
+      if (std::clamp(version, v2::minimum_version, v2::maximum_version) == version) {
+         genesis.emplace();
+         snapshot.read_section<eosio::chain::genesis_state>(
+               [&genesis = *genesis](auto& section) { section.read_row(genesis); });
+      }
+      return genesis;
+   }
+
+   std::vector<char> make_rocksdb_undo_prefix() { return rocksdb_undo_prefix; }
```
heifner, line 605: Not used, can be removed.

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,609 @@
...
+               }
+               FC_LOG_AND_RETHROW()
+            }
+            CATCH_AND_EXIT_DB_FAILURE()
+         }
+      }
+   }
+
+   // chainlib reserves prefixes 0x10 - 0x2F.
+   static const std::vector<char> rocksdb_undo_prefix{ 0x10 };
```
heifner, line 87: Not used.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,749 @@
...
+      const auto& table_store = primary_iter_store.get_table(key_store);
+      auto exact =
+            get_exact_iterator(table_store.contract, table_store.scope, table_store.table, key_store.primary);
+      EOS_ASSERT( exact.valid, db_rocksdb_invalid_operation_exception,
+                  "invariant failure in db_next_i64, iter store found to update but it does not exist in the database");
+      auto& session_iter = exact.itr;
+      auto& type_prefix = exact.type_prefix;
+      ++session_iter;
+      return find_and_store_primary_key(session_iter, key_store.table_ei, type_prefix,
+                                        key_store.table_ei, __func__, primary);
```
heifner, line 318: __func__ is not used by `find_and_store_primary_key`

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,749 @@
...
+      return primary_iter_store.add(primary_key_iter(table_ei, found_key, found_payer));
+   }
+
+   // returns the exact iterator and the bounding key (type)
+   db_context_rocksdb::exact_iterator db_context_rocksdb::get_exact_iterator(
+         name code, name scope, name table, uint64_t primary) {
+      const auto slice_primary_key = get_primary_slice_in_primaries(code, scope, table, primary);
+      auto session_iter = current_session.lower_bound(slice_primary_key.full_key);
+      const bool valid = primary_lookup.match(slice_primary_key.full_key, session_iter);
+      return { .valid = valid, .itr = std::move(session_iter), .type_prefix = std::move(slice_primary_key.prefix_key) };
```
heifner, line 681: `shared_bytes` is not movable

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,749 @@
...
+      primary_iter_store.swap(iterator, noop_secondary, payer);
+   }
+
+   // gets a prefix that allows for only primary key iterators
+   prefix_bundle db_context_rocksdb::get_primary_slice_in_primaries(name code, name scope, name table, uint64_t id) {
+      bytes primary_key = db_key_value_format::create_primary_key(scope, table, id);
+      return { primary_key, end_of_prefix::at_type, code };
+   }
+
+   pv_bundle db_context_rocksdb::get_primary_key_value(name code, name scope, name table, uint64_t id) {
```
heifner, line 646: This method could be made `const`

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,749 @@
...
+      prefix_bundle primary_key = get_primary_slice_in_primaries(code, scope, table, id);
+      return { primary_key, current_session.read(primary_key.full_key) };
+   }
+
+   void db_context_rocksdb::set_value(const shared_bytes& primary_key, const payer_payload& pp) {
+      current_session.write(primary_key, pp.as_payload());
+   }
+
+   void db_context_rocksdb::update_db_usage( const account_name& payer, int64_t delta, const storage_usage_trace& trace ) {
```
heifner, line 655: This should be removed. It hides the implementation at db_context with the same impl.

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: Added it to the assert

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: Are you saying it can never be moved, or that it cannot be moved because it is const in this case?

plugins/chain_plugin/chain_plugin.cpp
brianjohnson5972: Made 1

libraries/chain/backing_store/db_context.cpp
brianjohnson5972: Changed

libraries/chain/combined_database.cpp
brianjohnson5972: Changed to sec.kv_info_stack. This is only used for pending_state, which only ever calls push() which doesn't (effectively) do anything with kv_undo_stack.

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: Removed

libraries/chain/backing_store/db_context_rocksdb.cpp
heifner: Because it is const.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -0,0 +1,451 @@
...
+   };
+
+   template <typename Session>
+   struct kv_iterator_rocksdb : kv_iterator {
+      using session_type = std::remove_pointer_t<Session>;
+
+      uint32_t&                       num_iterators;
+      uint64_t                        kv_contract{0};
+      const char*                     kv_user_prefix{nullptr};
+      uint32_t                        kv_user_prefix_size{0};
```
heifner, line 91: `kv_user_prefix` & `kv_user_prefix_size` are not used.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -0,0 +1,451 @@
...
+      uint32_t kv_get_data(uint32_t offset, char* data, uint32_t data_size) override {
+         const char* temp        = nullptr;
+         uint32_t      temp_size = 0;
+         if (current_value) {
+            temp      = backing_store::actual_value_start(current_value->data());
+            temp_size = backing_store::actual_value_size(current_value->size());
+         }
+         if (offset < temp_size)
+            memcpy(data, temp + offset, std::min(data_size, temp_size - offset));
+         return temp_size;
```
heifner, line 420: The docs says "@return number of bytes written in data." However, this does not return the number of bytes written to `data` in case where `offset >= temp_size` or when `data_size` is less than `temp_size - offset`. This goes for the kv_context_chainbase impl as well.

The test cases for this seem to have been written for what was implemented not for what it should do.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -0,0 +1,451 @@
...
+            }
+            FC_LOG_AND_RETHROW()
+         }
+         CATCH_AND_EXIT_DB_FAILURE()
+
+         account_name payer = backing_store::get_payer(old_value->data());
+         return erase_table_usage(resource_manager, payer, key, key_size, actual_old_value_size);
+      }
+
+      int64_t kv_set(uint64_t contract, const char* key, uint32_t key_size, const char* value, uint32_t value_size,
```
heifner, line 353: The kv_context_chainbase logs to deep mind.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -0,0 +1,451 @@
...
+      uint32_t                                    num_iterators = 0;
+      eosio::session::shared_bytes                current_key;
+      std::optional<eosio::session::shared_bytes> current_value;
+
+      kv_context_rocksdb(session_type& the_session, name receiver, Resource_manager /*kv_resource_manager*/ resource_manager,
+                         const kv_database_config& limits)
+          : session{ &the_session }, receiver{ receiver },
+            resource_manager{ resource_manager }, limits{ limits } {}
+
+      int64_t kv_erase(uint64_t contract, const char* key, uint32_t key_size) override {
```
heifner, line 329: kv_context_chainbase logs to deep mind.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,745 @@
...
+   int32_t db_context_rocksdb::db_idx128_previous(int32_t iterator, uint64_t& primary) {
+      return sec_lookup_i128.previous_secondary(iterator, primary);
+   }
+
+   /**
+    * interface for 256-bit interger secondary
+    */
+   int32_t db_context_rocksdb::db_idx256_store(uint64_t scope, uint64_t table, account_name payer, uint64_t id,
+                           const uint128_t* data) {
+      const eosio::chain::key256_t& secondary = *reinterpret_cast<const eosio::chain::key256_t*>(data);
```
larryk85, line 485: strict aliasing violation.

Change to:
```c++
eosio::chain::key256_t secondary;
std::mempcy(secondary.data(), data, secondary.size());
```

The compiler is smart enough to vaporize the memcpy calls and not cause any actual copying, but it is conformant with the strict-aliasing rules.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,745 @@
...
+      return sec_lookup_i256.store(name{scope}, name{table}, payer, id, secondary);
+   }
+
+   void db_context_rocksdb::db_idx256_update(int32_t iterator, account_name payer, const uint128_t* data) {
+      const eosio::chain::key256_t& secondary = *reinterpret_cast<const eosio::chain::key256_t*>(data);
```
larryk85, line 490: strict-aliasing, see above.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,745 @@
...
+      sec_lookup_i256.update(iterator, payer, secondary);
+   }
+
+   void db_context_rocksdb::db_idx256_remove(int32_t iterator) {
+      sec_lookup_i256.remove(iterator);
+   }
+
+   int32_t db_context_rocksdb::db_idx256_find_secondary(uint64_t code, uint64_t scope, uint64_t table, const uint128_t* data,
+                                    uint64_t& primary) {
+      const eosio::chain::key256_t& secondary = *reinterpret_cast<const eosio::chain::key256_t*>(data);
```
larryk85, line 500: strict-aliasing, see above.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,745 @@
...
+      return sec_lookup_i256.find_secondary(name{code}, name{scope}, name{table}, secondary, primary);
+   }
+
+   int32_t db_context_rocksdb::db_idx256_find_primary(uint64_t code, uint64_t scope, uint64_t table, uint128_t* data,
+                                  uint64_t primary) {
+      eosio::chain::key256_t& secondary = *reinterpret_cast<eosio::chain::key256_t*>(data);
```
larryk85, line 506: strict-aliasing, see above.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,745 @@
...
+      return sec_lookup_i256.find_primary(name{code}, name{scope}, name{table}, secondary, primary);
+   }
+
+   int32_t db_context_rocksdb::db_idx256_lowerbound(uint64_t code, uint64_t scope, uint64_t table, uint128_t* data,
+                                uint64_t& primary) {
+      eosio::chain::key256_t& secondary = *reinterpret_cast<eosio::chain::key256_t*>(data);
```
larryk85, line 512: strict-aliasing, see above.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -0,0 +1,745 @@
...
+      return sec_lookup_i256.lowerbound_secondary(name{code}, name{scope}, name{table}, secondary, primary);
+   }
+
+   int32_t db_context_rocksdb::db_idx256_upperbound(uint64_t code, uint64_t scope, uint64_t table, uint128_t* data,
+                                uint64_t& primary) {
+      eosio::chain::key256_t& secondary = *reinterpret_cast<eosio::chain::key256_t*>(data);
```
larryk85, line 518: strict-aliasing, see above.

larryk85 (changes requested):

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
linhuang-blockone: The document https://github.com/EOSIO/spec-repo/blob/master/esr_key_value_database_intrinsics.md says
"Fetches data from temporary buffer starting at offset. Copies up to data_size bytes into data. Returns amount of data in temporary buffer.". Which document are you referring to?

heifner: `libraries/chain/include/eosio/chain/webassembly/interface.hpp` which is what the documentation is generated from I believe.

heifner: https://github.com/EOSIO/spec-repo/blob/master/esr_key_value_database_intrinsics.md says "It still aborts the transaction if [data, data + data_size) is out of the WASM's linear memory range." which is not true of the implementation.

larryk85: This is handled because of the fact that we are only using eos-vm backends now.  There is no need to explicitly force any of that as it will segfault all of the backends.

heifner: Is the consensus here that the comment in `interface.hpp` needs to be updated? Under the current impl how do you know if it failed?

linhuang-blockone: Should we  change the comment in `libraries/chain/include/eosio/chain/webassembly/interface.hpp` to make it consistent with the implementation and https://github.com/EOSIO/spec-repo/blob/master/esr_key_value_database_intrinsics.md?

larryk85 (commented):
These comments are for @gssagoo.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
brianjohnson5972: This is a doc problem.  @nksanthosh  who is the doc person, I cannot think of his name right now?

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
brianjohnson5972: Added

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
brianjohnson5972: added

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: fixed

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: fixed

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: fixed

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: fixed

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: fixed

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: fixed

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,849 @@
...
+       : backing_store(backing_store_type::CHAINBASE), db(chain_db) {}
+
+   combined_database::combined_database(chainbase::database& chain_db,
+                                        const controller::config& cfg)
+       : backing_store(backing_store_type::ROCKSDB), db(chain_db), kv_database{ [&]() {
+            rocksdb::Options options;
+
+            options.create_if_missing = !cfg.read_only; // Creates a database if it is missing
+            options.level_compaction_dynamic_level_bytes = true;
+            options.bytes_per_sync = 1048576; // used to control the write rate of flushes and compactions.
```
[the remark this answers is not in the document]
brianjohnson5972: EPE-459 will handle this

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,849 @@
...
+
+            // By default, RocksDB uses only one background thread
+            // for flush and compaction.
+            // Good value for `total_threads` is the number of cores
+            options.IncreaseParallelism(cfg.rocksdb_threads);
+
+            options.OptimizeLevelStyleCompaction(512ull << 20); // optimizes level style compaction
```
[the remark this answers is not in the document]
brianjohnson5972: EPE-459 will handle this

libraries/chain/combined_database.cpp
```diff
@@ -0,0 +1,849 @@
...
+            // faster and more accurate, especially for high bits
+            // per key or millions of keys in a single (full) filter.
+            rocksdb::BlockBasedTableOptions table_options;
+            table_options.format_version               = 5;
+            table_options.index_block_restart_interval = 16;
+
+            // Sets the bloom filter - Given an arbitrary key, 
+            // this bit array may be used to determine if the key 
+            // may exist or definitely does not exist in the key set.
+	          table_options.filter_policy.reset(rocksdb::NewBloomFilterPolicy(15, false));
```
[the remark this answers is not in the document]
brianjohnson5972: EPE-460 will handle this

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
brianjohnson5972: EPE-461 now is in charge of this

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
brianjohnson5972: removed

libraries/chain/combined_database.cpp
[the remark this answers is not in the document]
gssagoo: @larryk85 - That's a good question.  I was debating the same thing a while back and was wondering how much of the rocksDB configuration we want to expose.  I guess it depends on who the users are.  I am ok either way as you would know about the audience better.  However there are some configurations that depend on each other that I feel we should not make configurable and those have already been coded that way.   We can have a call to discuss this.  Thanks.

libraries/chain/combined_database.cpp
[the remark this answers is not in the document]
gssagoo: @larryk85 I guess my answer to options.bytes_per_sync applies here as well.  We can go either way, depends on the audience and how much control we want to give.   thanks

libraries/chain/combined_database.cpp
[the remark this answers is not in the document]
gssagoo: @larryk85 I test the performance of rocksDB locally on my machine using https://github.com/EOSIO/eosio-chain-bench and didn't notice any degradation in the performance after adding this.  There is a JIRA ticket open (EPE 443) related to this to explore this further.   Thanks

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
heifner: Under the current impl how do you know if `kv_get_data` failed? 

Say you pass in too large of an offset. `kv_get_data` will return the size of the temp buffer but not copy anything into `data`.

linhuang-blockone: Should we throw an exception in this case? I'd also like to assert on `data` not being NULL.

PR #9451, opened 2020-08-26
new chain_config param: action return value limit

dimas1185:
## Change Description

Consensus change.

Currently action return values are limited only by CPU and max wasm memory which is 20MB.
We need to provide node operators a way to prevent users from filling their SHiP logs with noise.

This change includes following changes:
1. adding `max_action_return_value_size` to `chain_config`. default value is 256 bytes.
this config value is used in `set_action_return_value` host function:
2. adding `wql-max-action-return-value` command line option to rodeos with default value of 20MB.
it is used in `set_action_return_value` from `rodeos/callbacks/action.hpp`

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [x] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/fc/include/fc/utility.hpp
```diff
@@ -18,6 +18,8 @@
 #define MAX_NUM_ARRAY_ELEMENTS (1024*1024)
 #define MAX_SIZE_OF_BYTE_ARRAYS (20*1024*1024)
 
+static_assert(MAX_SIZE_OF_BYTE_ARRAYS == 20*1024*1024, "Changing MAX_SIZE_OF_BYTE_ARRAYS breaks consensus. Make sure this is expected");
```
jeffreyssmith2nd, line 21: This will need to be ported to the fc repo as well

heifner: I don't think this belongs in `fc` when `fc` is a separate repo. Someone might be using fc for something completely different. I think we should move this static_assert into libraries/chain/include/eosio/chain/config.hpp

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -281,14 +274,17 @@ template <typename DataStream>
 inline DataStream &operator<<(DataStream &s, const eosio::chain::data_entry<eosio::chain::chain_config_v1, eosio::chain::config_entry_validator> &entry){
    using namespace eosio::chain;
 
-   if (!entry.is_allowed())
-      return s;
+   //initial requirements were to skip packing field if it is not activated.
+   //this approach allows to spam this function with big buffer so changing this behavior
```
swatanabe-b1, line 278: Actually, skipping entries was wrong
- The contract has no way to know that the value was skipped and is likely to behave incorrectly.
- When the protocol feature is not activated, the old version of nodeos that doesn't know about the entry MUST behave the same as the new version of nodeos that does.  Skipping known but unactivated entries violates this.

unittests/test-contracts/action_results/action_results.cpp
```diff
@@ -11,5 +13,10 @@ class [[eosio::contract]] action_results : public contract {
          return 10;
       }
 
+      [[eosio::action]]
+      vector<char> retoverlim() {
+         return vector<char>(512, '0');
```
swatanabe-b1, line 18: I'd like to see tests for:
- exactly at the limit
- exactly one byte over the limit
- The maximum allowed limit

This may require calling `set_action_return_value` directly.

libraries/chain/chain_config.cpp
```diff
@@ -39,17 +39,53 @@ namespace eosio { namespace chain {
                   "max authority depth should be at least 1" );
 }
 
+template<typename Stream>
+Stream& chain_config_v0::log(Stream& out) const{
```
swatanabe-b1, line 43: Is this used anywhere?  If it's defined here it won't be instantiated if it's used in another translation unit.

libraries/chain/include/eosio/chain/chain_config.hpp
dimas1185: @swatanabe-b1  do you propose to remove the requirement for skipping/throwing exception in `get_parameters_packed` ? so inactive config should be presented regardless activation status

swatanabe-b1: I'm not suggesting any functional changes.  The comment indicates that this change is an optimization.  I'm saying that it's also necessary for correctness.  I think throwing an exception is the correct behavior for inactive config entries.

libraries/chain/include/eosio/chain/chain_config.hpp
```diff
@@ -15,6 +15,7 @@ namespace eosio { namespace chain {
  */
 struct chain_config_v0 {
 
+   //order must match parameters as ids are used in serialization
```
tbfleming, line 18: Add an additional comment: do not add more to this list; doing so will break chain_config_v1

tbfleming:
ship changes and rodeos changes look correct

libraries/chain/include/eosio/chain/chain_config.hpp
tbfleming: Never mind. It's already implied by the v0/v1 inheritance.

PR #9453, opened 2020-08-27
Reverting some libs

larryk85:
## Change Description
Previously we had un-submodularized some libraries.  This became apparent that this was a misstep and needs to be reverted back to the way it was.

We currently have quite a few external clients of these libraries so when we patch we need to create the same commits against the repo, and this is a non-trivial task now.  It also leads to a lot of issues when checking out `develop` from other branches such as `master`, and currently any of the `release/*` branches.  The way that these libraries were carried over lost the git history before the point of un-submodularization.  Because of these issues, I am making this PR to help get us back on track with this.

This will cause issues if you have tasks that are currently changing files in `fc`, `chainbase` or `appbase`.  To resolve this a couple of scripts are going to be provided below.  These scripts will copy your repo filter based on which branch your changes are at and which library your changes are in ('fc', 'chainbase' or 'appbase').  It will then do a `git log` so you can find the commit that is the one directly before your commits start.  It will then ask for that commit and construct a set of patches to be applied.  The second script should be called from within the repo of question ('fc', 'chainbase', or 'appbase') and checked out to the branch you want.  This will apply those patches previously made and clean up the copy that was made.  You should now have all of your commits in the repo and no longer stuck in the library in eos.

If you are navigating between branches or commits and you run into issues checking out into the previous 'un-submodularized' zone then you will need to remove the directories `libraries/fc libraries/chainbase libraries/appbase`.

These scripts are a quick effort to give engineers some degree of automation of migrating back, but these are not designed to be robust and will not do much checking.  Meaning that it will have to used in the ways described above.

### git-fixup-a.sh
```sh
#! /bin/bash

EOSIO_DIR=$1
GIT_BRANCH=$2
WHICH_LIB=$3 # currently this with either be fc, appbase or chainbase

# GIT_BOUNDS -- this is the non-inclusive lower bound to create the patches. I.e. this will be the commit before your work started.

cp -R ${EOSIO_DIR} /tmp/eos_copy

pushd /tmp/eos_copy &> /dev/null

git checkout ${GIT_BRANCH}

git filter-branch --prune-empty --subdirectory-filter ./libraries/${WHICH_LIB}

git log

printf "\n\nWhich commit to use: "
read GIT_BOUNDS

git format-patch ${GIT_BOUNDS}..HEAD
```

### git-fixup-b.sh
```sh
#! /bin/bash

git am -3 /tmp/eos_copy/*.patch

rm -rf /tmp/eos_copy
```

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


larryk85:
This PR is failing the checks because of the submodule revert check.  This isn't an issue and is only caused because of the issues mentioned in the description.

larryk85:
> see /tmp/eos_copy has the latest commits as in develop branch

Thanks for your evaluation @allenhan2.

heifner:
Note I had to add a trailing `/`, `./libraries/fc/` for the `filter-branch` to work

larryk85:
> Note I had to add a trailing `/`, `./libraries/fc/` for the `filter-branch` to work

Thanks @heifner, yeah that will cause it to fail.  I am updating the description to have the trailing `/`.

PR #9459, opened 2020-09-01
Kv unit test

linhuang-blockone:
## Change Description

This PR adds unit test cases for KV API on RocksDb using templated mocks. Those tests do not rely on external dependencies and cover much more normal and error execution paths than otherwise.

The PR also moves KV API related files to corresponding backing_store directories for better organization.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -189,23 +190,31 @@ namespace eosio { namespace chain {
       }
    }; // kv_iterator_rocksdb
 
+   template<typename View, typename Iterator, typename Write_session, typename Resource_manager>
```
b1bart, line 193: Do we expect that `View` and `Iterator` can really be independent types?  It seems that in both our runtime and testing systems the `Iterator` tightly coupled to (and a inner class of) `View`. 

If that will always be the case, we should remove the degree of freedom for `Iterator` here and using something like:
```
using iterator_type = typename View::iterator_type;
```

That way the `View` will dictate the iterator type.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -14,21 +14,22 @@ namespace eosio { namespace chain {
...
-      memcpy(&payer, data, kv_payer_size);
+      memcpy(&payer, data, kv_payer_size); // Before this method is called, data was checked to be at least kv_payer_size long
       return payer;
    }
 
    static const char* actual_value_start(const char* data) {
       return data + kv_payer_size;
    }
 
+   template <typename View, typename Iterator>
```
b1bart, line 25: If we decide to collapse `Iterator` into `View` in some way, this will also collapse

libraries/chain/backing_store/tests/test_kv_rocksdb.cpp
```diff
@@ -0,0 +1,689 @@
...
+      mock_is_erased = []( ) -> bool {
+         return false;
+      };
+      std::unique_ptr<kv_iterator> it = my_kv_context->kv_it_create(contract, prefix.c_str(), it_key_size);
+      BOOST_CHECK(it->kv_it_move_to_end() == kv_it_stat::iterator_end);
+   }
+
+   BOOST_FIXTURE_TEST_CASE(test_kv_it_next_erased, kv_rocksdb_fixture)
+   {
+      check_kv_it_move(true, true, true, kv_it_stat::iterator_ok);
```
b1bart, line 631: The use of `bool`'s here makes it hard to really understand what these test cases are doing without going back and forth with the definition of `check_kv_it_*`

Maybe we can make a parameters structure so we can have some names aka:
```
enum it_dir {
   NEXT,
   PREV
}
struct check_kv_it_move_params {
   it_dir direction = NEXT;
   bool is_erased = false;
   bool is_end = false;
}

check_kv_it_move({}, kv_it_stat::iterator_ok);
check_kv_it_move({.is_end=true}, kv_it_stat::iterator_end);
check_kv_it_move({.direction=PREV, .is_erased=true, .is_end=true}, kv_it_stat::iterator_ok}
```

I'm not sold on the above approach but I think there is something that can be done for readability here.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
linhuang-blockone: Thank you Bart so much for taking your time to review the code. Yes, the View and Iterator are tightly coupled. I will collapse them.

libraries/chain/backing_store/tests/test_kv_rocksdb.cpp
linhuang-blockone: I was concern about those true/false sequences for readability too. Thanks for pointing this out. Even for test code we should strive for best code. Will modify the code.

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
linhuang-blockone: I have collapsed them.

revl (approved):

PR #9460, opened 2020-09-01
rpc kv access implement get_kv_table_rows

kimjh2005:
## Change Description

Implemented get_kv_table_rows equivalent to get_table_rows.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -12,6 +12,8 @@
 #include <eosio/chain/generated_transaction_object.hpp>
 #include <eosio/chain/global_property_object.hpp>
 #include <eosio/chain/snapshot.hpp>
+#include <eosio/chain/kv_context.hpp>
+#include <eosio/to_key.hpp>
 
 #include <eosio/chain/eosio_contract.hpp>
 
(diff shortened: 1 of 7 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,207 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+
+   lb_key.resize(lb_key_size);
+   status_lb = lb_itr->kv_it_key(0, lb_key.data(), lb_key_size, lb_key_actual_size);
+   ub_key.resize(ub_key_size);
+   status_ub = ub_itr->kv_it_key(0, ub_key.data(), ub_key_size, ub_key_actual_size);
+
+   // iterate the range
+   auto walk_table_row_range = [&]( auto &itr, auto &end_itr, bool reverse ) {
+      auto cur_time = fc::time_point::now();
+      auto end_time = cur_time + fc::microseconds(1000 * 1000);
```
heifner, line 2053: We have talked about future http responses to be limited by `http-max-response-time-ms` instead of hard-coded. Other hard-coded values are 10ms not 1s.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,207 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+      auto success = kv_context.kv_get(p.code.to_uint64_t(), key.data(), key.size(), value_size);
+      if( success ) {
+         vector<char> value;
+         value.resize(value_size);
+         auto return_size = kv_context.kv_get_data(0, value.data(), value_size);
+         EOS_ASSERT(return_size == value_size, chain::contract_table_query_exception,  "point query value size mismatch: ${s1} ${s2}", ("s1", return_size)("s2", value_size));
+
+         fc::variant row_var;
+         if( p.json ) {
+          row_var = abis.binary_to_variant( p.table.to_string(), value, abi_serializer::create_yield_function( abi_serializer_max_time ), shorten_abi_errors );
```
heifner, line 1996: indentation. Should try-catch and revert to fc::variant( value ) on exception.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,207 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+}
+
+read_only::get_table_rows_result read_only::get_kv_table_rows( const read_only::get_kv_table_rows_params& p)const {
+   name database_id = chain::kvram_id;
+   auto& gp = db.get_global_properties();
+   auto& kv_config = gp.kv_configuration;
+   const chain::kv_database_config& limits = kv_config.kvram;
+
+   auto&database = db.db();
+   auto kv_context = chain::create_kv_chainbase_context(const_cast<chainbase::database&>(database), database_id, chain::name{0}, {}, limits);
```
jeffreyssmith2nd, line 1941: I could be wrong, but I thought we had added a way to get a non-const reference to `chainbase::database` then we wouldn't have to cast away the const.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,207 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+
+      unsigned int count = 0;
+      for( count = 0; cur_time <= end_time && count < p.limit && cmp < 0; cur_time = fc::time_point::now() ) {
+         row_value.clear();
+         row_value.resize(val_size);
+         auto status = itr->kv_it_value(0, row_value.data(), val_size, actual_size);
+
+         fc::variant row_var;
+         if( p.json ) {
+            row_var = abis.binary_to_variant( p.table.to_string(), row_value, abi_serializer::create_yield_function( abi_serializer_max_time ), shorten_abi_errors );
```
heifner, line 2077: Same comment as above. Also should probably adjust yield time according to how much time is left instead of just going with `abi_serializer_max_time`

plugins/chain_plugin/chain_plugin.cpp
kimjh2005: get_kv_tables_rows() needs to be revisited once rocksdb changes are merged to develop to enable rocksdb  kv_context.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,213 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+   auto&database = db.db();
+   auto kv_context = chain::create_kv_chainbase_context(const_cast<chainbase::database&>(database), database_id, chain::name{0}, {}, limits);
+   const abi_def abi = eosio::chain_apis::get_abi( db, p.code );
+
+   return get_kv_table_rows_context( p, *kv_context, abi );
+}
+
+read_only::get_table_rows_result read_only::get_kv_table_rows_context( const read_only::get_kv_table_rows_params& p, kv_context  &kv_context, const abi_def &abi )const {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-aliasing"
```
revl, line 1949: I couldn't find where exactly aliasing is happening at first glance. Wouldn't it make sense to only surround the problematic lines with pragmas?

kimjh2005: removed unused pragma.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,209 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
 #pragma GCC diagnostic pop
 }
```
heifner, line 1913: Are the `chain_api_plugin` changes coming in under a diff PR?

kimjh2005: It is not in the original scope.  I can add one line but we need to have a separate task to test http plugin.

heifner: This is unusable without http support. If there is a JIRA issue to add that, then fine otherwise please create a JIRA issue to add http support.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,209 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+         value.resize(value_size);
+         auto return_size = kv_context.kv_get_data(0, value.data(), value_size);
+         EOS_ASSERT(return_size == value_size, chain::contract_table_query_exception,  "point query value size mismatch: ${s1} ${s2}", ("s1", return_size)("s2", value_size));
+
+         fc::variant row_var;
+         if( p.json ) {
+            try {
+               row_var = abis.binary_to_variant( p.table.to_string(), value, abi_serializer::create_yield_function( wait_time ), shorten_abi_errors );
+            } catch (fc::exception &e) {
+               elog( "Error occurred while converting binary to variant in get_kv_table_rows: '${t}') to: ${i}", ("t", p.table.to_string())("i", p.index_name.to_string()));
```
heifner, line 2000: Should not `elog` this as it will fill up logs of public api providers.

plugins/chain_plugin/chain_plugin.cpp
jeffreyssmith2nd: Can we add a warning here in the code, and an additional Jira ticket to make sure that work is captured?

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,209 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+            row_var = fc::variant( value );
+         }
+         result.rows.emplace_back( std::move(row_var) );
+      }
+      return result;
+   }
+
+   ///////////////////////////////////////////////////////////
+   // ranged query
+   ///////////////////////////////////////////////////////////
```
jeffreyssmith2nd, line 2013: What happens if I specify index_value, lower_bound, and upper_bound? The way I'm reading this, it will do the value lookup as well as the lb/ub range. That doesn't seem like the desired behaviour, so maybe the range query should be in an `else` clause?

kimjh2005: There is "return result" inside point query.  If point query is successful, it returns.

plugins/chain_plugin/chain_plugin.cpp
kimjh2005: removed

plugins/chain_plugin/chain_plugin.cpp
kimjh2005: added chain_api_plugin change.

plugins/chain_plugin/chain_plugin.cpp
kimjh2005: I added a TO DO comment and will make a JIRA ticket for it.

plugins/chain_plugin/chain_plugin.cpp
heifner: Make sure to update the PR description with docs on the new apis.

heifner: Also add a test for it to `plugin_http_api_test.py`. Also what about `cleos` support? Is that a diff PR?

jeffreyssmith2nd: Cleos support is a separate task by a different person

kimjh2005: added get_kv_table_rows to plugin_http_api_test.

kimjh2005: updated chain.swagger.yaml.

jeffreyssmith2nd (approved):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1909,6 +1911,210 @@ read_only::get_table_rows_result read_only::get_table_rows( const read_only::get
...
+      unsigned int count = 0;
+      for( count = 0; cur_time <= end_time && count < p.limit && cmp < 0; cur_time = fc::time_point::now() ) {
+         row_value.clear();
+         row_value.resize(val_size);
+         auto status = itr->kv_it_value(0, row_value.data(), val_size, actual_size);
+
+         fc::variant row_var;
+         if( p.json ) {
+            auto time_left = end_time - cur_time;
+            row_var = abis.binary_to_variant( p.table.to_string(), row_value, abi_serializer::create_yield_function( time_left ), shorten_abi_errors );
```
heifner, line 2082: try-catch here as well

PR #9480, opened 2020-09-09
Kv microbenchmarking

linhuang-blockone:
## Change Description

This is a draft PR seeking comments about microbenchmarking KV APIs on RocksDb and ChainKV.
generate_keys.cpp is used to generate keys which are used as an input to benchmark_kv.cpp for creating
key-values on RocksDb or ChainKV.
Currently get, get_data, set, it_next, it_key_value are benchmarked.
I will modify and add more operations based on comments.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+// Global test data
+constexpr account_name receiver = N(kvrdb);
+constexpr uint64_t contract = receiver.to_uint64_t();
+constexpr account_name payer = N(payer);
+std::string prefix = "prefix";
+constexpr uint32_t it_key_size = 10;
+constexpr uint64_t default_billable_size = 12;
+
+constexpr uint32_t default_value_size = 1024;
+std::string default_operation = "get";
```
TimothyBanks, line 23: `constexpr const char* default_operation`?

libraries/chain/backing_store/tests/generate_keys.cpp
```diff
@@ -0,0 +1,117 @@
...
+      << "   -k=KEY, the first key to be generated\n"
+      << "   -h, print this help\n"
+      << std::endl;
+}
+
+// As this program is only for internal use,
+// we don't use sophisticated library to parse command line
+void parse_cmd_line(int argc, char* argv[], cmd_args& args) {
+   if (argc != 7) {
+      std::cout << "Require 3 arguments" << std::endl;
```
TimothyBanks, line 31: We are checking for 7 arguments.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+   std::string key;
+   auto i = 0;
+   std::unordered_set<std::string> unordered_keys;
+   while (i < work_set_size && getline(key_file, key)) {
+      unordered_keys.insert(key);  // randomize
+      ++i;
+   }
+   key_file.close();
+
+   std::vector<std::string> keys;
```
TimothyBanks, line 137: We could reserve the space we need for this vector too.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      kv_context_ptr->kv_set(contract, key.c_str(), key.size(), value.c_str(), value.size(), payer);
+      ++args.num_keys;
+   }
+
+   key_file.close();
+}
+
+void benchmark_get(cmd_args& args, const std::unique_ptr<kv_context> kv_context_ptr, std::vector<std::string> work_set, wall_clock_t& start, wall_clock_t& end, rusage& usage_start, rusage& usage_end) {
+   start = std::chrono::steady_clock::now();
+   getrusage(RUSAGE_SELF, &usage_start);
```
TimothyBanks, line 164: Is this something we want included in the performance metric?  What are we measuring in these methods?  I would think that we would want to move the measurement to only clocking the time it takes to return from `kv_get` which means the we would be counting only within the inner most loop.

Similar question in regards to the other benchmarking functions.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+   std::string key;
+   while (getline(key_file, key)) {
+      kv_context_ptr->kv_set(contract, key.c_str(), key.size(), value.c_str(), value.size(), payer);
+      ++args.num_keys;
+   }
+
+   key_file.close();
+}
+
+void benchmark_get(cmd_args& args, const std::unique_ptr<kv_context> kv_context_ptr, std::vector<std::string> work_set, wall_clock_t& start, wall_clock_t& end, rusage& usage_start, rusage& usage_end) {
```
TimothyBanks, line 162: `kv_context_ptr` is passed by value.  Is that desired?

Could `work_set` be passed as a const reference?

Same for the other benchmarking functions.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      benchmark_set(args, std::move(kv_context_ptr), work_set, start, end, usage_start, usage_end);
+   } else if (args.operation == "it_next" ) {
+      benchmark_it_next(args, std::move(kv_context_ptr), start, end, usage_start, usage_end);
+   } else if (args.operation == "it_key_value" ) {
+      benchmark_it_key_value(args, std::move(kv_context_ptr), start, end, usage_start, usage_end);
+   };
+}
+
+void print_results(cmd_args& args, wall_clock_t& start, wall_clock_t& end, const rusage& usage_start, const rusage& usage_end) {
+   std::string title = "Benchmarked \"" + args.operation + "\" for ";
```
TimothyBanks, line 266: Maybe consider `stringstream` instead.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      << std::endl << std::endl;
+}
+
+void benchmark(cmd_args& args) {
+   uint32_t clock_duration_us;
+   wall_clock_t start, end;
+   struct rusage usage_start, usage_end;
+
+   if (args.is_rocksdb) {
+      boost::filesystem::remove_all("kvrdb-tmp");  // Use a clean RocksDB
```
TimothyBanks, line 311: rocksdb has a method for deleting the database files given the database name.  I don't know if there are other files that need to be removed.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      boost::filesystem::remove_all(chain::config::default_state_dir_name);
```
TimothyBanks, line 312: This happens in either condition, could we just move it before the conditional?

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      b1::chain_kv::database      kv_db{"kvrdb-tmp", true};
+      b1::chain_kv::undo_stack    kv_undo_stack{kv_db, vector<char>{make_rocksdb_undo_prefix()}};
+
+      std::unique_ptr<kv_context> kv_context_ptr = create_kv_rocksdb_context<b1::chain_kv::view, b1::chain_kv::write_session, mock_resource_manager>(kv_db, kv_undo_stack, receiver, resource_manager, limits); 
+      benchmark_operation(args, std::move(kv_context_ptr), start, end, usage_start, usage_end); // kv_context_ptr must be in the same scope as kv_db and usage_start, since they are references in create_kv_rocksdb_context
```
TimothyBanks, line 317: This also happens in either condition, could we move this after the conditional?

libraries/chain/backing_store/tests/benchmark_kv.cpp
heifner: stringstream performance is terrible, not that it matters here. I much rather see std::string and to_string instead of stringstream.

TimothyBanks: I would like to see the metrics on that.  Assuming that you don't know the ultimate size of your string to be able to reserve the space before hand, appending to a string would result in reallocations and copies.

heifner: Google it. stringstream you are paying for all the overhead of locale and the terrible iostream library.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+using namespace eosio;
+using namespace eosio::chain;
+
+namespace kv_benchmark {
+
+// Global test data
+constexpr account_name receiver = N(kvrdb);
+constexpr uint64_t contract = receiver.to_uint64_t();
+constexpr account_name payer = N(payer);
+std::string prefix = "prefix";
```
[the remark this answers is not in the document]
linhuang-blockone: Thanks for spotting this. It was left from an earlier version. Will remove it.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Will change.

libraries/chain/backing_store/tests/generate_keys.cpp
linhuang-blockone: What I meant we needed key file name, first key, and number of keys. Will reword.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Will do.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: getrusage returns more detailed information about performance. Bart used it in previous benchmarking activities and suggested use it here too.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Will clean up constness. Thanks.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: I prefer to std::string too for simplicity.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Will look around for how to start with a clean database.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: benchmark_operation uses kv_context_ptr which requires kv_db or chainbase_db to be in the scope. If we move benchmark_operation after the condition, we would need to move kv_db and chainbase_db before the condition and
initialize both even though only one is needed.

libraries/chain/backing_store/tests/generate_keys.cpp
```diff
@@ -0,0 +1,117 @@
...
+      }
+   }
+}
+
+// Generate next key alphabetically.
+// Lower case only without loss of generality.
+std::string generate_next_key(const std::string& current_key) {
+   std::string next_key = current_key;
+   uint32_t last_possible_index = current_key.size() - 1;
+   while (last_possible_index >= 0) {
```
swatanabe-b1, line 79: This is always true, because `last_possible_index` is unsigned.

libraries/chain/backing_store/tests/generate_keys.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+// Generates "num_keys" keys starting with "first_key" alphabetically
+// and saves them into "key_file".
+// One may use this program to generate multiple files of keys,
+// concatenate them, and use Linux shuf command to randomize the
+// resulting file.
+
+void print_usage(char* program_name) {
+   std::cout << program_name << " - Generates \"num_keys\" keys starting with \"first_key\" alphabetically and saves them into \"key_file\"\n\n"
+      << "   -f=FILENAME, file storing the resultingkeys\n"
```
swatanabe-b1, line 20: `parse_cmd_line` does not expect `=`

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      << "   -f=FILENAME, file storing keys one per line \n"
+      << "   -o=OPERATION, operation to be benchmarked: get, get_data, set, it_next, it_key_value, Default " << default_operation << std::endl
+      << "   -c, use Chainbase. Without this option: use RocksDb\n"
+      << "   -h, print this help\n"
+      << "   -v=NUMBER, value size. Default " << default_value_size << std::endl
+      << std::endl;
+}
+
+// As this program is only for internal use,
+// we don't use sophisticated library to parse command line
```
swatanabe-b1, line 88: Hmm.  Using a library is supposed to be easier.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+   if (!key_file.is_open()) {
+      std::cout << "Failed to open key file " << args.key_file << std::endl;
+      exit(2);
+   }
+
+   std::string key;
+   auto i = 0;
+   std::unordered_set<std::string> unordered_keys;
+   while (i < work_set_size && getline(key_file, key)) {
+      unordered_keys.insert(key);  // randomize
```
swatanabe-b1, line 132: This is a rather odd way of randomizing the order.  `std::shuffle` would be clearer and probably more efficient.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+         print_usage(argv[0]);
+         exit(1);
+      }
+   }
+}
+
+std::vector<std::string> get_work_set_keys(const cmd_args& args) {
+   std::ifstream key_file(args.key_file);
+   if (!key_file.is_open()) {
+      std::cout << "Failed to open key file " << args.key_file << std::endl;
```
swatanabe-b1, line 124: Should use `cerr`, not `cout` for error messages.

libraries/chain/backing_store/tests/benchmark_kv.cpp
swatanabe-b1: Since it's all just going to `cout` below, is a temporary string needed at all?

libraries/chain/backing_store/tests/benchmark_kv.cpp
larryk85: Did Bart recommend using both in this fashion?  It seems a bit counterintuitive.  You will be capturing the overhead of two `getrusage` calls within your timer.  In addition you appear to only output the time in `microseconds` which rusage already gives you.  It seems that you can get by with just the rusage and add the user time and kernel time together.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+struct cmd_args {
+   uint32_t num_keys = 0;
+   uint32_t value_size = default_value_size;
+   bool is_rocksdb = true; // false means chainbase
+   std::string operation = default_operation;
+   std::string key_file = "";
+};
+
+constexpr uint32_t work_set_size = 10000;
+constexpr uint32_t num_iterations = 100;
```
larryk85, line 33: Why is this so small?  As I mentioned elsewhere, this should be sweepable so that we can see when things fall over.

libraries/chain/backing_store/tests/benchmark_kv.cpp
```diff
@@ -0,0 +1,337 @@
...
+      for (auto& key: work_set) {
+         kv_context_ptr->kv_set(contract, key.c_str(), key.size(), value.c_str(), value.size(), payer);
+      }
+   }
+
+   getrusage(RUSAGE_SELF, &usage_end);
+   end = std::chrono::steady_clock::now();
+}
+
+void benchmark_it_next(cmd_args& args, const std::unique_ptr<kv_context> kv_context_ptr, wall_clock_t& start, wall_clock_t& end, rusage& usage_start, rusage& usage_end) {
```
larryk85, line 209: It seems like we would also want to know the overhead of creating an iterator.  So another function like this but just create random iterators from our key set.

larryk85 (changes requested):

libraries/chain/backing_store/tests/generate_keys.cpp
linhuang-blockone: Thanks for spotting this out! Yes, it will be always true. I have removed this file altogether. I am now using Linux utility pwgen to generate random keys, which is much more powerful and flexible.

libraries/chain/backing_store/tests/generate_keys.cpp
linhuang-blockone: This file is removed.

libraries/chain/backing_store/tests/benchmark_kv.cpp
swatanabe-b1: What is this `std::random_shuffle`?  I've never heard of that algorithm.

heifner: https://en.cppreference.com/w/cpp/algorithm/random_shuffle

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Thanks. I will change to use the command line parsing library.

libraries/chain/backing_store/tests/benchmark_kv.cpp
swatanabe-b1: Nope.  Doesn't exist.

heifner: lol

linhuang-blockone: Thanks! I had changed to use pwgen and shuf commands to generate keys and worksets. First I use pwgen to generate  key files for different key sizes, concatenate them, and use "sort -u" to get rid off duplicates, and run shuf three times on the resulting concatenated file. This file becomes the master random key file. I then run "shuf -n" to build workset files to make sure workset files are random. Those two files are supplied as arguments to the benchmarking program. Now I don't dynamically generate keys.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Thanks. Will change.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Each time I only want to benchmark a single operation. If we benchmark multiple operations in the same proces, will previous runs affect memory foot prints?

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: I will remove std::chrono::steady_clock::now. It is unnecessary. work_set_size has already been removed; it is now calculated from the set itself. num_iterations is now a parameter, with default value 100000.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Will add that. Thanks you Bucky and Steven for the excellent comments!

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Will only use getrusage.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: Thanks. Will use string_view.

libraries/chain/backing_store/tests/benchmark_kv.cpp
linhuang-blockone: I have removed the strings.

nksanthosh:
@linhuang-blockone Is this going to be merged soon? What help is needed from reviewers if any at this point? Thanks!

linhuang-blockone:
> @linhuang-blockone Is this going to be merged soon? What help is needed from reviewers if any at this point? Thanks!

Bucky is going to do a final review. Afterwards it will be merged.

larryk85 (approved):

linhuang-blockone:
An identical PR https://github.com/EOSIO/eos/pull/9521 which resolved merging conflicts has been merged. Close this PR accordingly.

PR #9494, opened 2020-09-14
add an integration of nodeos for crash when the nodes are killed

vzqzhang:
## Change Description

Added a new integration test for nodeos. This test sets up one producing nodes and one "bridge" node using test_control_api_plugin. We send transactions to the bridge node. After the accounts are initialized and allocated with symbols, we kill the producing node. Then we flood hundreds transactions to the bridge node. At the end, we ill the bridge. It tests that no crashes happen when the nodes are killed. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/nodeos_contrl_c_test.py
+++ b/tests/nodeos_contrl_c_test.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+
+from testUtils import Utils
+from datetime import datetime
+from datetime import timedelta
+import time
+from Cluster import Cluster
+import json
+from WalletMgr import WalletMgr
+from Node import Node
+from TestHelper import TestHelper
+
+import signal
+
+###############################################################
+# nodeos_contrl_c_lr_test
+#
+# This test sets up one producing nodes and one "bridge" node using test_control_api_plugin. We send
+# transactions to the bridge node. After the accounts are initialized and allocated with symbols, we
+# kill the producing node. Then we flood hundreds transactions to the bridge node. At the end, we
+# kill the bridge. It tests that no crashes happen when the nodes are killed.
+#
+###############################################################
+
+errorExit=Utils.errorExit
+
+args = TestHelper.parse_args({"--wallet-port", "-v"})
+
+cluster=Cluster(walletd=True)
+killAll=True
+totalProducerNodes=1
+totalNonProducerNodes=1
+totalNodes=totalProducerNodes+totalNonProducerNodes
+walletPort=args.wallet_port
+walletMgr=WalletMgr(True, port=walletPort)
+producerEndpoint = '127.0.0.1:8888'
+httpServerAddress = '127.0.0.1:8889'
+
+try:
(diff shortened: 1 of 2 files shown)


tests/nodeos_contrl_c_test.py
```diff
@@ -0,0 +1,111 @@
...
+    cluster.setWalletMgr(walletMgr)
+    cluster.killall(allInstances=killAll)
+    cluster.cleanup()
+
+    specificExtraNodeosArgs = {}
+    specificExtraNodeosArgs[0] = "--plugin eosio::producer_plugin --plugin eosio::chain_api_plugin --plugin eosio::http_plugin "
+    "--plugin eosio::txn_test_gen_plugin --plugin eosio::producer_api_plugin --plugin eosio::history_plugin "
+    "--plugin eosio::history_api_plugin --plugin eosio::state_history_plugin --plugin eosio::db_size_api_plugin "
+    "--p2p-listen-endpoint=0.0.0.0:9876 --p2p-server-address=0.0.0.0:9876 --http-server-address=127.0.0.1:8888 "
+    "--delete-all-blocks --disable-replay-opts --delete-state-history --access-control-allow-origin='*'"
```
heifner, line 50: You don't need history_plugin, history_api_plugin, state_history_plugin, db_size_api_plugin.

I'm not sure you even need anything here for extra args. @brianjohnson5972

vzqzhang: > You don't need history_plugin, history_api_plugin, state_history_plugin, db_size_api_plugin.
> 
> I'm not sure you even need anything here for extra args. @brianjohnson5972

Removed the extra options

nksanthosh:
Since this test is not consistent in results - should we close the PR?

heifner:
I asked if we could get automation to run this test a few hundred times to see if it fails a few times. If it does fail then it would be useful to keep. If it never fails, then I would like us to modify it such that it does cause develop to fail. We have hit shutdown bugs under load a number of times and it would be good to have a test for it.

nksanthosh:
@kj4ezj Will you be able to run this test a few hundred times and report back on results? (not urgent - can be scheduled per your weekly workload)

kj4ezj:
> @kj4ezj Will you be able to run this test a few hundred times and report back on results?

This required code changes to the CI system, so I opened [pull request 9551](https://github.com/EOSIO/eos/pull/9551) based on this branch. I successfully completed a short test run in [build 24759](https://buildkite.com/EOSIO/eosio/builds/24759) with 21 runs in rounds of 14 runs each.

Finally, I kicked off [build 24760](https://buildkite.com/EOSIO/eosio/builds/24760) to perform 2100 runs across 30 rounds of 70 runs each (ten runs per operating system).

heifner (approved):

PR #9503, opened 2020-09-16
[docs] Add slices, trace log, clog format explainers to Trace API plugin - 2.0

### LINKED ISSUE #9502: [docs] Need more explainers in Trace API plugin to use options
[docs/01_nodeos/03_plugins/trace_api_plugin/index.md](https://developers.eos.io/manuals/eos/latest/nodeos/plugins/trace_api_plugin/index)

- Need to define slices, trace log filename structure, clog format, etc.
- Also, need to add new compression option.

lparisc:
## Change Description

Resolves #9502

## Change Type
**Select ONE**
- [x] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


docs/01_nodeos/03_plugins/trace_api_plugin/index.md
```diff
@@ -101,23 +112,74 @@ nodeos --data-dir data_dir --config-dir config_dir --trace-dir traces_dir
...
 
-```console
+### Slices
+
+In the context of the `trace_api_plugin`, a *slice* is defined as a collection of all relevant trace data between a given starting block height (inclusive) and a given ending block height (exclusive). For instance, a slice from 0 to 10,000 is a collection of all blocks with block numbers greater than or equal to 0 and less than 10,000. The trace directory contains a collection of slices. Each slice consists of a *trace data* log file and a *trace index* metadata log file:
+
+  *  `trace_<S>-<E>.log`
+  *  `trace_index_<S>-<E>.log`
+
+where `<S>` and `<E>` are the starting and ending block numbers for the slice padded with leading 0's to a stride. For instance if the start block is 5, the last is 15, and the stride is 4, then the resulting `<S>` is `0000000005` and `<E>` is `0000000015`.
```
b1bart, line 126: I think the TDD was incorrect.  The given examples are of stride `10` which is also what we use in the code.  Originally they correctly showed as 4 digits but we didn't sync the "stride" value when we true'd them to the code

docs/01_nodeos/03_plugins/trace_api_plugin/index.md
```diff
@@ -101,23 +112,74 @@ nodeos --data-dir data_dir --config-dir config_dir --trace-dir traces_dir
...
+
+#### trace_index&#95;&lt;S&gt;-&lt;E&gt;.log
+
+The trace index log or metadata log is an append only log that stores a sequence of binary-serialized types. Currently two types are supported: `block_entry_v0` and `lib_entry_v0`. The log begins with a basic header that includes versioning information about the data stored in the log. `block_entry_v0` includes a block ID and block number with an offset to the location of that block within the data log. `lib_entry_v0` includes an entry for the latest known LIB. The reader will use the LIB information for reporting to users an irreversible status.
```
b1bart, line 134: By the time of this writing there are actually 3 types of entries, those two and `block_trace_v1` which contains the same data as `block_trace_v0` and **adds** the information from #9005

b1bart (changes requested):

docs/01_nodeos/03_plugins/trace_api_plugin/index.md
lparisc: These are the entries for the trace index log file.

docs/01_nodeos/03_plugins/trace_api_plugin/index.md
lparisc: This is fixed now.

docs/01_nodeos/03_plugins/trace_api_plugin/index.md
b1bart: yes, correct, my mistake

b1bart (approved):

PR #9508, opened 2020-09-18
[docs] Add WASM interface C++ reference documentation

victorj8:
## Change Description

Currently there is few documentation about host functions on EOS, this PR provides documentation for some host functions.

## Change Type
**Select ONE**
- [x] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -68,6 +68,10 @@
           },
           "disable_default_filters" :true,
           "filters" : []
+        },
+        {
+          "name": "doxybook",
+          "options": {}
         }
     ]
 }
--- a/docs.json
+++ b/docs.json
@@ -60,6 +60,14 @@
                 "disable_filters": true,
                 "disable_summary_gen": true
             }
+        },
+        {
+          "name": "doxygen_to_xml",
+          "options": {
+            "INPUT": "libraries/chain/include/eosio/chain/webassembly/interface.hpp"
+          },
+          "disable_default_filters" :true,
+          "filters" : []
         }
     ]
 }
--- a/libraries/chain/include/eosio/chain/webassembly/interface.hpp
+++ b/libraries/chain/include/eosio/chain/webassembly/interface.hpp
@@ -54,7 +54,7 @@ namespace webassembly {
           * Allows a privileged smart contract, e.g. the system contract, to pre-activate a consensus protocol upgrade feature.
           *
           * @ingroup privileged
-          * @param digest_type - 256-bit digest representing the feature to pre-activate.
+          * @param feature_digest - 256-bit digest representing the feature to pre-activate.
          */
          void preactivate_feature(legacy_ptr<const digest_type> feature_digest);
 
@@ -130,8 +130,7 @@ namespace webassembly {
            * @param[out] packed_parameters the ouput for the parameters.
            * @param max_version has no effect, but should be 0.
            *
-           * @retval the size of the packed parameters if packed_parameters is empty.
-           * @retval the amount of data written in packed_parameters.
+           * @return the size of the packed parameters if packed_parameters is empty, otherwise it returns the amount of data written in packed_parameters.
          */
          uint32_t get_wasm_parameters_packed( span<char> packed_parameters, uint32_t max_version ) const;
 
@@ -186,12 +185,12 @@ namespace webassembly {
           * Proposes a schedule change with extended features.
           *
           * Valid formats:
-          * 0 : serialized array of producer_keys. Using this format is exactly equivalent to set_proposed_producers
-          * 1 : serialized array of producer_authority's
+          * 0 : serialized array of producer_keys. Using this format is exactly equivalent to set_proposed_producers.
+          * 1 : serialized array of producer_authority's.
           *
           * @ingroup privileged
           *
-          * @param packed_producer_format - format of the producer data blob
+          * @param packed_producer_format - format of the producer data blob.
           * @param packed_producer_schedule - packed data of representing the producer schedule in the format indicated.
           *
           * @return -1 if proposing a new producer schedule was unsuccessful, otherwise returns the version of the new proposed schedule.
@@ -214,7 +213,7 @@ namespace webassembly {
           *
           * @ingroup privileged
           *
-          * @param packed_parameters - a span containing the packed blockchain config parameters.
+          * @param packed_blockchain_parameters - a span containing the packed blockchain config parameters.
          */
          void set_blockchain_parameters_packed(legacy_span<const char> packed_blockchain_parameters);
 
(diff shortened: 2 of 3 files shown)


libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -16,28 +16,292 @@ namespace webassembly {
...
+         /**
+          * Gets the maximum key size, maximum value size, and maximum iterators of a kv database and returns the size of the data.
+          * The kv parameters are encoded as 16 bytes, representing four 32-bit little-endian values.
+          * +-------+---------------+---------------+---------------+---------------+
+          * | byte  | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 |10 |11 |12 |13 |14 |15 |
+          * +-------+---------------+---------------+---------------+---------------+
+          * | field |   version     |   key limit   |  value limit  | max iterators |
+          * +-------+---------------+---------------+---------------+---------------+
+          * | type  |      0        |   32-bits LE  |  32-bits LE   |  32-bits LE   |
+          * +-------+---------------+---------------+---------------+---------------+
```
lparisc, line 265: This table is not rendering correctly. Check https://www.doxygen.nl/manual/tables.html

docs.json
```diff
@@ -60,6 +60,14 @@
...
                 "disable_summary_gen": true
             }
+        },
+        {
+          "name": "doxygen_to_xml",
+          "options": {
+            "INPUT": "libraries/chain/include/eosio/chain/webassembly/interface.hpp"
+          },
+          "disable_default_filters" :true,
+          "filters" : []
```
lparisc, line 70: Need to add the "doxybook" generator in "docs.json" (after the "doxygen_to_xml" generator) for the contents to render on the left navigation menu:
```
            ...
        },
        {
            "name": "doxybook",
            "options": {}
        }
```

lparisc (changes requested):

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -100,175 +364,1547 @@ namespace webassembly {
...
+          *
+          * @ingroup producer
+          * @param[out] producers - output buffer containing the names of the current active producer names.
+          *
+          * @return number of bytes required (if the buffer is empty), or the number of bytes written to the buffer.
+         */
+         int32_t get_active_producers(legacy_span<account_name> producers) const;
+
+         /**
+          * Tests a given public key with the recovered public key from digest and signature.
```
spoonincode, line 378: It may be worth noting somehow that this assertion does not require the EOSIO canonical check to pass. That is, a signature that would have been disallowed during transaction acceptance would potentially be allowed here. Many contract users probably don't care about that detail.. but it's an important subtle behavior of the interface

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -100,175 +364,1547 @@ namespace webassembly {
...
          void abort() const;
-         void eosio_assert(bool, null_terminated_ptr) const;
-         void eosio_assert_message(bool, legacy_span<const char>) const;
-         void eosio_assert_code(bool, uint64_t) const;
-         void eosio_exit(int32_t) const;
 
-         // action api
-         int32_t read_action_data(legacy_span<char>) const;
+         /**
+          * Aborts processing of this action if the test condition is true.
```
spoonincode, line 612: Is this correct?

spoonincode:
Should the documentation offer any guidance or recommendations on usage patterns of any of these? For example, should the documentation recommend to use the assert_sha256() variant over sha256() if contract's usage pattern allows it. Or, avoiding spamage of eosio_assert() with negative checks.

docs.json
victorj8: done

libraries/chain/include/eosio/chain/webassembly/interface.hpp
victorj8: there is an exception being thrown 

`EOS_THROW( eosio_assert_message_exception, "assertion failure with message: ${s}", ("s",message) );`

but correct me if this is not the case or it requires a different wording

spoonincode: I think the description is negated? Isn't the action aborted when the condition is _false_

victorj8: that's right, thanks for catching that.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
victorj8: I am somewhat out of my element here, but if you have a suggested text i can add it, let me know what text should i add.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
victorj8: Thanks for the observation, but after trying the table solution the html table still was not rendering properly

/**
 * Gets the maximum key size, maximum value size, and maximum iterators of a kv database and returns the size of the data.
 * The kv parameters are encoded as 16 bytes, representing four 32-bit little-endian values.
 * <table>
 * <tr><th>byte <th> 0 | 1 | 2 | 3 <th> 4 | 5 | 6 | 7 <th> 8 | 9 | 10 | 11 <th> 12 | 13 | 14 | 15
 * <tr><td>field <td>version <td>key limit <td>value limit <td>max iterators
 * <tr><td>type <td>0 <td>32-bits LE <td>32-bits LE <td>32-bits LE
 * </table>
 *
 * @ingroup privileged
 * @param[out] packed_kv_parameters - the buffer containing the packed kv parameters.
 * @param max_version - has no effect, but should be 0.
 *
 * @return Returns the size required in the buffer (if the buffer is too small, nothing is written).
 *
*/

I switched to use codeblocks instead.

lparisc:
> Should the documentation offer any guidance or recommendations on usage patterns of any of these? For example, should the documentation recommend to use the assert_sha256() variant over sha256() if contract's usage pattern allows it. Or, avoiding spamage of eosio_assert() with negative checks.

Yes, of course. If the usage suggestions are short and to the point they can be added in a "remarks" section within the applicable function. However, I believe a separate "overview" section that includes best practices is also desired for devs to be able to use the WASM interface correctly and effectively. To that end, I have created this issue #9518 to keep this effort active. In the meantime I'm approving this unless there are any hard corrections. @victorj8 please wait at least 24 hours before merging in case there are any additional corrections. Thanks

lparisc (approved):

PR #9510, opened 2020-09-18
Switch default backing_store from RocksDB to chainbase

revl:
## Change Description

Previously, the default backing store was set to ROCKSDB, which was a mistake. This PR switches the default backing store to CHAINBASE and fixes the fallout of this change.

## Change Type

- [ ] Documentation
- [x] Stability bug fix
- [ ] Other
- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -4,19 +4,20 @@
 
 namespace eosio { namespace chain {
 
-   combined_session::combined_session(chainbase::database& cb_database)
-       : cb_session{ std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true)) } {}
-
-   combined_session::combined_session(chainbase::database& cb_database, b1::chain_kv::undo_stack& kv_undo_stack)
-       : kv_undo_stack{ &kv_undo_stack } {
-      try {
+   combined_session::combined_session(chainbase::database& cb_database, b1::chain_kv::undo_stack* kv_undo_stack)
+       : kv_undo_stack{ kv_undo_stack } {
+      if (!kv_undo_stack) {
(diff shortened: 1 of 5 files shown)


libraries/chain/combined_database.cpp
```diff
@@ -4,19 +4,20 @@
...
-   combined_session::combined_session(chainbase::database& cb_database)
-       : cb_session{ std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true)) } {}
-
-   combined_session::combined_session(chainbase::database& cb_database, b1::chain_kv::undo_stack& kv_undo_stack)
-       : kv_undo_stack{ &kv_undo_stack } {
-      try {
+   combined_session::combined_session(chainbase::database& cb_database, b1::chain_kv::undo_stack* kv_undo_stack)
+       : kv_undo_stack{ kv_undo_stack } {
+      if (!kv_undo_stack) {
+         cb_session = std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true));
```
brianjohnson5972, line 10: I'm guessing I am missing something, but how come you don't just initialize cb_session (not in an if check), and then just do if(kv_undo_stack) { ... push(false) ... }

brianjohnson5972 (changes requested):

libraries/chain/combined_database.cpp
revl: That is so that if an exception is thrown from the cb_session constructor, the application is terminated by analogy with other methods. Should I make this a special case?

brianjohnson5972: I guess do this for now, it passed by @larryk85 before.

brianjohnson5972 (approved):

PR #9512, opened 2020-09-21
Keep http_plugin_impl alive while connection objects are alive

vzqzhang:
## Change Description

When the nodeos is killed, the thread_pool and associated io_context of the http_plugin_impl should keep alive for the lifetime of the connection object 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -224,7 +224,7 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
          websocket_local_server_type unix_server;
 #endif
 
-         bool                     validate_host;
+         bool                     validate_host = true;
          set<string>              valid_hosts;
 
          bool host_port_is_valid( const std::string& header_host_port, const string& endpoint_local_host_port ) {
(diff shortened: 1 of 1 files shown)


nksanthosh:
@heifner Can you please review this PR? Thanks!

heifner:
@nksanthosh this is all my code Qing just made the PR. Needs to be reviewed by someone else.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -938,12 +951,15 @@ namespace eosio {
          my->thread_pool->stop();
       }
 
+      // release http_plugin_impl_ptr shared_ptrs captured in url handlers
+      my->url_handlers.clear();
```
swatanabe-b1, line 955: This isn't quite enough.  The thread pool's `io_context` may contain unexecuted handlers that hold `shared_ptrs` to the `http_plugin_impl`.

swatanabe-b1 (changes requested):

plugins/http_plugin/http_plugin.cpp
heifner: I'm not following. The `io_context` destructor will destroy all the unexecuted handlers that hold `shared_ptrs` to `http_plugin_impl`.
https://www.boost.org/doc/libs/develop/doc/html/boost_asio/reference/io_context/_io_context.html

Do you just mean the comment is not clear? Or do you see an issue here?

swatanabe-b1: You're only stopping the thread pool, not resetting it.  This leaves a `shared_ptr` cycle.

heifner: Ah. Thanks!

swatanabe-b1 (approved):

PR #9515, opened 2020-09-22
Improve handling of the backing_store configuration parameter

revl:
## Change Description

This commit makes the code around setting the `backing_store` parameter in `combined_database` more readable and adds an additional check for when nodeos is relaunched with a different type of `backing_store`.

## Change Type

- [ ] Documentation
- [ ] Stability bug fix
- [x] Other
- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -108,9 +108,7 @@ namespace eosio { namespace chain {
                     "Chainbase indicates that RocksDB is in use; resync, replay, or restore from snapshot to switch back to chainbase");
          break;
       case backing_store_type::ROCKSDB:
-         {
-            if (db.get<kv_db_config_object>().backing_store == backing_store_type::ROCKSDB)
-               return;
+         if (db.get<kv_db_config_object>().backing_store != backing_store_type::ROCKSDB) {
             auto& idx = db.get_index<kv_index, by_kv_key>();
             auto  it  = idx.lower_bound(boost::make_tuple(name{}, std::string_view{}));
             EOS_ASSERT(it == idx.end(), database_move_kv_disk_exception,
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -101,18 +101,25 @@ namespace eosio { namespace chain {
          kv_database(std::make_unique<b1::chain_kv::database>(rocksdb_path.c_str(), rocksdb_create_if_missing, rocksdb_threads, rocksdb_max_open_files)),
          kv_undo_stack(std::make_unique<b1::chain_kv::undo_stack>(*kv_database, rocksdb_undo_prefix)) {}
 
-   void combined_database::set_backing_store(backing_store_type backing_store) {
-      if (backing_store == backing_store_type::ROCKSDB) {
-         if (db.get<kv_db_config_object>().backing_store == backing_store_type::ROCKSDB)
-            return;
-         auto& idx = db.get_index<kv_index, by_kv_key>();
-         auto  it  = idx.lower_bound(boost::make_tuple(name{}, std::string_view{}));
-         EOS_ASSERT(it == idx.end(), database_move_kv_disk_exception,
-                    "Chainbase already contains KV entries; use resync, replay, or snapshot to move these to "
-                    "rocksdb");
-         db.modify(db.get<kv_db_config_object>(), [](auto& cfg) { cfg.backing_store = backing_store_type::ROCKSDB; });
+   void combined_database::check_backing_store_setting() {
+      switch (backing_store) {
+      case backing_store_type::CHAINBASE:
+         EOS_ASSERT(db.get<kv_db_config_object>().backing_store != backing_store_type::ROCKSDB, database_move_kv_disk_exception,
+                    "Chainbase indicates that RocksDB is in use; resync, replay, or restore from snapshot to switch back to chainbase");
+         break;
+      case backing_store_type::ROCKSDB:
+         {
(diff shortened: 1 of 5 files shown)


libraries/chain/combined_database.cpp
```diff
@@ -101,18 +101,25 @@ namespace eosio { namespace chain {
...
-         auto  it  = idx.lower_bound(boost::make_tuple(name{}, std::string_view{}));
-         EOS_ASSERT(it == idx.end(), database_move_kv_disk_exception,
-                    "Chainbase already contains KV entries; use resync, replay, or snapshot to move these to "
-                    "rocksdb");
-         db.modify(db.get<kv_db_config_object>(), [](auto& cfg) { cfg.backing_store = backing_store_type::ROCKSDB; });
+   void combined_database::check_backing_store_setting() {
+      switch (backing_store) {
+      case backing_store_type::CHAINBASE:
+         EOS_ASSERT(db.get<kv_db_config_object>().backing_store != backing_store_type::ROCKSDB, database_move_kv_disk_exception,
+                    "Chainbase indicates that RocksDB is in use; resync, replay, or restore from snapshot to switch back to chainbase");
```
linhuang-blockone, line 108: Where/when is backing_store_type::CHAINBASE state saved in the chainbase? Should a statement like db.modify(db.get<kv_db_config_object>(), [](auto& cfg) { cfg.backing_store = backing_store_type::CHAINBASE; }); appear somewhere?

libraries/chain/combined_database.cpp
```diff
@@ -101,18 +101,25 @@ namespace eosio { namespace chain {
...
+         break;
+      case backing_store_type::ROCKSDB:
+         {
+            if (db.get<kv_db_config_object>().backing_store == backing_store_type::ROCKSDB)
+               return;
```
linhuang-blockone, line 113: Add a log saying "using rocksdb for backing store" before "return". Otherwise the users will get confused sometimes see backing store setting while other times not.

libraries/chain/combined_database.cpp
revl: Where would you suggest this statement should appear?

libraries/chain/combined_database.cpp
revl: Done.

libraries/chain/combined_database.cpp
linhuang-blockone: Whenever configuration is changed from RocksDB to Chainbase, I think that should be reflected in the kv_db_config_object in the database. Maybe this kv_db_config_object's name should be changed too as it contains
information not only for kv.

revl: It's not possible to switch from RocksDB to Chainbase, because, unlike the opposite case - Chainbase to RocksDB, there is no check for whether RocksDB was actually used. Once nodeos has been switched to RocksDB, that change is permanent, and a replay or restore from snapshot would be required to switch back to Chainbase.

linhuang-blockone: I see. Thanks. If backing_store is Chainbase, kv_db_config_object will be created to its default value backing_store_type::Chainbase.

linhuang-blockone (approved):

libraries/chain/combined_database.cpp
revl: Yes.

nksanthosh:
@revl Can you please see why the tests are failing?

libraries/chain/combined_database.cpp
```diff
@@ -101,18 +101,23 @@ namespace eosio { namespace chain {
...
-         auto& idx = db.get_index<kv_index, by_kv_key>();
-         auto  it  = idx.lower_bound(boost::make_tuple(name{}, std::string_view{}));
-         EOS_ASSERT(it == idx.end(), database_move_kv_disk_exception,
-                    "Chainbase already contains KV entries; use resync, replay, or snapshot to move these to "
-                    "rocksdb");
-         db.modify(db.get<kv_db_config_object>(), [](auto& cfg) { cfg.backing_store = backing_store_type::ROCKSDB; });
+   void combined_database::check_backing_store_setting() {
+      switch (backing_store) {
+      case backing_store_type::CHAINBASE:
+         EOS_ASSERT(db.get<kv_db_config_object>().backing_store != backing_store_type::ROCKSDB, database_move_kv_disk_exception,
```
brianjohnson5972, line 107: this should just verify == backing_store_type::CHAINBASE

pipeline.jsonc
```diff
@@ -51,7 +51,7 @@
         "test":
         [
             {
-                "commit": "5160fa853650669d245e91dccdd5b0275b62748d"
+                "commit": "0090b4bcdbf20db925d290f7a2122ed84c6caf87"
```
brianjohnson5972, line 54: Do you know why this needed to change?

brianjohnson5972 (changes requested):

pipeline.jsonc
revl: My previous change of this setting after switching the default backing store from ROCKSDB to CHAINBASE was overlapped by a recent merge of another PR, which apparently unrolled this setting back to an earlier commit.

libraries/chain/combined_database.cpp
revl: Done

brianjohnson5972 (approved):

PR #9519, opened 2020-09-24
Fix missing payer in kv snapshots and backing_store setting error in …

linhuang-blockone:
…kv_snapshot_test

## Change Description

Corrected backing_store error, added payer to kv_object_view, added code to convert payer representation between chainbase and rocksdb. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
+++ b/libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
@@ -4,28 +4,28 @@
 #include <eosio/chain/kv_chainbase_objects.hpp>
 #include <eosio/chain/backing_store/kv_context.hpp>
 
-namespace eosio { namespace chain {
+namespace eosio { namespace chain { namespace backing_store { namespace kv_rksdb {
    static constexpr auto kv_payer_size = sizeof(account_name);
 
-   static uint32_t kv_rksdb_actual_value_size(const uint32_t raw_value_size) {
+   inline static uint32_t actual_value_size(const uint32_t raw_value_size) {
       EOS_ASSERT(raw_value_size >= kv_payer_size, kv_rocksdb_bad_value_size_exception , "The size of value returned from RocksDB is less than payer's size");
       return (raw_value_size - kv_payer_size);
    }
 
-   static account_name kv_rksdb_get_payer(const char* data) {
+   inline static account_name get_payer(const char* data) {
       account_name payer;
       memcpy(&payer, data, kv_payer_size); // Before this method is called, data was checked to be at least kv_payer_size long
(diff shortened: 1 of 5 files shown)


libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
```diff
@@ -7,21 +7,35 @@
 namespace eosio { namespace chain {
    static constexpr auto kv_payer_size = sizeof(account_name);
 
-   static uint32_t actual_value_size(const uint32_t raw_value_size) {
+   static uint32_t kv_rksdb_actual_value_size(const uint32_t raw_value_size) {
```
larryk85, line 10: Why were these functions renamed? The `kv_rksdb_` prefixes are a bit much and indicative that we should be namespacing these.  Was there clashing of function names?  Also, none of these functions are `inline`, meaning that they will cause duplicate symbols if we include in multiple compilation units.

larryk85 (changes requested):

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
linhuang-blockone: I thought the original names were too generic to be used outside kv_context_rocksdb. Will change to use name space and inline. Thanks.

brianjohnson5972: Can you revert them?  I renamed them in my branch.

brianjohnson5972: I renamed them and moved them, since I needed them as well.

brianjohnson5972 (changes requested):
After you revert the name changes. I have a class to replace your new method, but I will do that when I merge, but would rather not have to fix all those merges if I can avoid it.

TimothyBanks (approved):

libraries/chain/include/eosio/chain/backing_store/kv_context_rocksdb.hpp
linhuang-blockone: I am going to revert the naming changes, not introducing namespace for now so we can move forward quickly today. Will revisit this next week once everything is in.

larryk85 (approved):

brianjohnson5972 (approved):

PR #9523, opened 2020-09-25
Merge issue exposed a bug in undo_stack::commit method.

TimothyBanks:
## Change Description
A merge issue in the snapshot_tests exposed a bug in the new undo_stack commit method.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain_kv/include/b1/session/undo_stack.hpp
+++ b/libraries/chain_kv/include/b1/session/undo_stack.hpp
@@ -99,20 +99,16 @@ void undo_stack<Session>::commit(int64_t revision) {
 
    auto  start_index       = revision - initial_revision;
    auto& session_to_commit = m_sessions[start_index++];
-   auto  current_index     = 0;
-   auto  parent            = m_sessions[0].parent();
 
-   std::visit([&](auto* parent) { session_to_commit.attach(*parent); }, parent);
+   session_to_commit.detach();
+   session_to_commit.attach(*m_head);
    session_to_commit.commit();
    session_to_commit.detach();
 
-   for (int64_t i = 0; i <= start_index; ++i) { m_sessions[i].detach(); }
-   for (int64_t i = start_index; i < static_cast<int64_t>(m_sessions.size()); ++i) {
-      m_sessions[current_index++] = std::move(m_sessions[i]);
-   }
-   m_sessions.erase(std::begin(m_sessions) + current_index, std::end(m_sessions));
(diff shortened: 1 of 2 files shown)


libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -99,20 +99,16 @@ void undo_stack<Session>::commit(int64_t revision) {
...
    session_to_commit.commit();
    session_to_commit.detach();
 
-   for (int64_t i = 0; i <= start_index; ++i) { m_sessions[i].detach(); }
-   for (int64_t i = start_index; i < static_cast<int64_t>(m_sessions.size()); ++i) {
-      m_sessions[current_index++] = std::move(m_sessions[i]);
-   }
-   m_sessions.erase(std::begin(m_sessions) + current_index, std::end(m_sessions));
+   for (int64_t i = 0; i < start_index; ++i) { m_sessions[i].detach(); }
+   m_sessions.erase(std::begin(m_sessions), std::begin(m_sessions) + start_index);
```
linhuang-blockone, line 109: So now you remove everything from initial revision to current revision?

unittests/snapshot_tests.cpp
```diff
@@ -731,10 +731,6 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_kv_snapshot, SNAPSHOT_SUITE, snapshot_suites)
             auto writer = SNAPSHOT_SUITE::get_writer();
             chain.control->write_snapshot(writer);
             auto snapshot = SNAPSHOT_SUITE::finalize(writer);
-
-            // Set backing_store for load snapshot
-            set_backing_store(chain, resulting_backing_store);
-
```
linhuang-blockone, line 737: Good.

libraries/chain_kv/include/b1/session/undo_stack.hpp
TimothyBanks: The original implementation was based off of a `std::vector` and what I did was move all the remaining sessions to the front and then erased the end of the vector.  I moved it to a `std::deque` because I needed pointers to not become invalidated and erasing from the front of the `std::deque` is a lot less expensive than a vector.

So the intent was to remove everything from the initial revision to the revision that was being committed in both cases.

linhuang-blockone (approved):

PR #9540, opened 2020-10-01
Configurable ack-mode for amqp_trx_plugin

heifner:
## Change Description

- Add new option to `amqp_trx_plugin` that controls when AMQP is ack'ed.
```
     --amqp-trx-ack-mode arg (=in_block)   AMQP ack when 'received' from AMQP,
                                        when 'executed', or when 'in_block' is
                                        produced that contains trx.
                                        Options: received, executed, in_block
```

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

See above.

### DIFF
--- a/plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
+++ b/plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
@@ -16,7 +16,35 @@ namespace {
 
 static appbase::abstract_plugin& amqp_trx_plugin_ = appbase::app().register_plugin<eosio::amqp_trx_plugin>();
 
-constexpr auto def_max_trx_in_progress_size = 100*1024*1024; // 100 MB
+enum class ack_mode {
+   received,
+   executed,
+   in_block
+};
+
+std::istream& operator>>(std::istream& in, ack_mode& m) {
+   std::string s;
+   in >> s;
+   if( s == "received" )
+      m = ack_mode::received;
+   else if( s == "executed" )
+      m = ack_mode::executed;
+   else if( s == "in_block" )
+      m = ack_mode::in_block;
+   else
+      in.setstate( std::ios_base::failbit );
+   return in;
+}
+
+std::ostream& operator<<(std::ostream& osm, ack_mode m) {
+   if( m == ack_mode::received )
+      osm << "received";
+   else if( m == ack_mode::executed )
+      osm << "executed";
+   else if( m == ack_mode::in_block )
+      osm << "in_block";
+   return osm;
+}
 
 } // anonymous
 
@@ -31,6 +59,8 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
    std::optional<amqp> amqp_trx;
 
    std::string amqp_trx_address;
+   ack_mode acked = ack_mode::executed;
+   std::map<uint32_t, chain::deque<eosio::amqp::delivery_tag_t>> tracked_delivery_tags;
    uint32_t trx_processing_queue_size = 1000;
    bool allow_speculative_execution = false;
    std::shared_ptr<fifo_trx_processing_queue<producer_plugin>> trx_queue_ptr;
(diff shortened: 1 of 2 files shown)


plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -77,7 +134,11 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
 
       trx_queue_ptr->push( trx,
                 [my=shared_from_this(), delivery_tag, trx](const fc::static_variant<fc::exception_ptr, chain::transaction_trace_ptr>& result) {
-            my->amqp_trx->ack( delivery_tag );
+            if( my->acked == ack_mode::executed ) {
+               my->amqp_trx->ack( delivery_tag );
+            } else if ( my->acked == ack_mode::in_block ) {
```
swatanabe-b1, line 139: If the transaction failed, does it make sense to ack it with the block, when it isn't in the block?

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -58,12 +88,39 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
...
       amqp_trx->reject( delivery_tag );
    }
 
+   void on_block_start( uint32_t bn ) {
+      trx_queue_ptr->on_block_start();
+   }
+
+   void on_block_abort( uint32_t bn ) {
+      if( acked == ack_mode::in_block ) {
+         tracked_delivery_tags.erase( bn );
```
swatanabe-b1, line 106: Won't this mean that we never ack the transaction?  Does it even make sense to use `in_block` when running in speculative mode?

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: Good point, should ack it on execute if it failed.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: You are right. I was thinking it would execute in the next block, but we have lost the tracking of the delivery_tag at that point. To avoid confusion, I like your idea of just not allowing `in_block` and `amqp-trx-speculative-execution=true`

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -58,12 +88,39 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
...
+      }
+      trx_queue_ptr->on_block_stop();
+   }
+
+   void on_accepted_block( const chain::block_state_ptr& bsp ) {
+      if( acked == ack_mode::in_block ) {
+         const auto& entry = tracked_delivery_tags.find( bsp->block_num );
+         if( entry != tracked_delivery_tags.end() ) {
+            for( const auto& q : entry->second ) {
+               amqp_trx->ack( q );
```
swatanabe-b1, line 116: Doesn't AMQP have a batch ack?

heifner: AMQP has an option to ack a delivery_tag and all delivery_tags before it. Since we are executing in order, can just store last delivery_tag per block and then ack just that last delivery_tag.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -58,12 +88,34 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
...
+
+   void on_block_abort( uint32_t bn ) {
+      trx_queue_ptr->on_block_stop();
+   }
+
+   void on_accepted_block( const chain::block_state_ptr& bsp ) {
+      if( acked == ack_mode::in_block ) {
+         const auto& entry = tracked_delivery_tags.find( bsp->block_num );
+         if( entry != tracked_delivery_tags.end() ) {
+            amqp_trx->ack( entry->second );
```
swatanabe-b1, line 112: `, true `

heifner: wow, of course. thanks

swatanabe-b1 (approved):

PR #9545, opened 2020-10-02
Session optimizations

TimothyBanks:
## Change Description

Various changes after benchmarking and profiling the Session API code.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -333,11 +333,11 @@ namespace eosio { namespace chain {
                      // In KV RocksDB, payer and actual data are packed together.
                      // Extract them.
                      auto           value = (*it).second;
-                     auto           payer = get_payer(value.data());
-                     auto           actual_value_data = actual_value_start(value.data());
+                     auto           payer = get_payer(value->data());
+                     auto           actual_value_data = actual_value_start(value->data());
                      kv_object_view row{ name(contract),
                                          { { key.data() + key_prefix_size, key.data() + key.size() } },
-                                         { { actual_value_data, actual_value_data + actual_value_size(value.size()) } },
+                                         { { actual_value_data, actual_value_data + actual_value_size(value->size()) } },
                                          payer
                      };
                      section.add_row(row, db);
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -488,12 +488,10 @@ namespace eosio { namespace chain {
                                       move_to_rocks->kv_key.data() + move_to_rocks->kv_key.size());
 
                         // Pack payer and actual key value
-                        bytes final_kv_value;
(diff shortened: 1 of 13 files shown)


libraries/chain/combined_database.cpp
```diff
@@ -112,7 +112,7 @@ namespace eosio { namespace chain {
             if (!status.ok())
                throw std::runtime_error(std::string{ "database::database: rocksdb::DB::Open: " } + status.ToString());
             auto rdb        = std::shared_ptr<rocksdb::DB>{ p };
-            return std::make_unique<rocks_db_type>(eosio::session::make_session(std::move(rdb)));
+            return std::make_unique<rocks_db_type>(eosio::session::make_session(std::move(rdb), 1024));
```
linhuang-blockone, line 170: Can you add a comment or use a constexpr about 1024?

I see below it is max_iterators. Then there is a system consensus limits (whose type kv_database_config defines  max_key_size, max_value_size, max_iterators). I think you should use max_iterators from limits.

libraries/chain/combined_database.cpp
```diff
@@ -260,8 +260,10 @@ namespace eosio { namespace chain {
                     kv_undo_stack->push();
                     undo = true;
                   }
-
-                  for (auto it = kv_undo_stack->top().lower_bound(prefix_key); it != kv_undo_stack->top().lower_bound(next_prefix); ++it) {
+                  
+                  auto begin_it = kv_undo_stack->top().lower_bound(prefix_key);
+                  auto end_it = kv_undo_stack->top().lower_bound(next_prefix);
```
linhuang-blockone, line 320: Good, you don't call lower_bound every time.

libraries/chain_kv/include/b1/session/shared_bytes.hpp
```diff
@@ -114,10 +107,11 @@ inline shared_bytes shared_bytes::operator++(int) {
       buffer.pop_back();
    }
 
-   return eosio::session::make_shared_bytes(buffer.data(), buffer.size());
+   return eosio::session::shared_bytes(buffer.data(), buffer.size());
 }
```
heifner, line 111: `operator++(int)` 
- Seems like this should be a method and not an operator++ as it does not modify this. I expect an operator++(int) to modify this and return the old value. If it was an operator++ then it should be pre increment since it returns a new value not the current value.
- It looks like `if (++buffer.back())` is to detect overflow and not to detect a value of -1. Either add check for -1 or verify not max char as overflow of signed is undefined.
- This method could be optimized. It makes two copies of bytes. One into temp buffer and another into the return value. It should only make one copy at most.

linhuang-blockone (approved):

TimothyBanks:
> Are left and right pointers?

No.  They are `std::optional` types.

PR #9547, opened 2020-10-05
Trace API plugin - Add support for action return values

victorj8:
## Change Description
The intent of this PR is to add support to the trace api plugin for a new field representing a value that may be returned by an action.

To accomplish this task 3 new structs were added:

action_trace_v1 : representing the new version of an action trace that now may contain a field representing the return value.

transaction_trace_v2: representing the new version for the transaction trace that supports action_trace_v1.

block_trace_v2: representing the new version for block trace that supports transaction_trace_v2.

The general idea is that, for example, a new version for action traces may now not require the creation of new structs for transaction trace and block trace, but only a modification to the actions type on the transaction trace struct (add a new possibility for the variant).

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


victorj8:
> It appears unnecessary to include `std::vector<action_trace_v0>` in this as it will never be filled in. Is the choice to include it intentional such that the serialized format has a `which` value that lines up with the version?

That was not the intention, i added just as a way to experiment how it would look like, but as you mention, currently it will not be filled in, so i just deleted it.

> Is it worth considering taking a hard stance that "create_" code paths only make the latest types and can therefore be simplified with that assumption.

I modified to create the latest block_trace version .

b1bart (approved):

PR #9548, opened 2020-10-05
Improve RockDB performance option configuration

linhuang-blockone:
## Change Description

This PR improves RocksDB performance option configuration in several ways:
1. For `rocksdb_threads`, if user does not configure it, use the number of CPU cores dynamically found from the system.
2. `target-file-size-base` and `max-bytes-for-level-base` rely on `write-buffer-size` only. Remove them from options to prevent possible errors from the user. Nodeos now calculates them automatically.
3. Change `rocksdb-write-buffer-size` to `rocksdb-write-buffer-size-mb` for simpler and consistent configuration.
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -126,16 +126,13 @@ namespace eosio { namespace chain {
             options.level0_file_num_compaction_trigger = 2;
 
             // Size of L0 = write_buffer_size * min_write_buffer_number_to_merge * level0_file_num_compaction_trigger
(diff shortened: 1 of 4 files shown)


gssagoo (approved):
Looks good Lin.

TimothyBanks (approved):

PR #9571, opened 2020-10-14
Log store implementation

huangminghuang:
## Change Description
This PR implements the nuraft log_store interface to persist log entries which allows the state to be recovered from restart.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/blockvaultd/log_store.cpp
+++ b/programs/blockvaultd/log_store.cpp
@@ -1,0 +1,279 @@
+#include "log_store.hpp"
+#include <cassert>
+#include <iostream>
+
+namespace blockvault {
+
+using namespace nuraft;
+namespace fs = std::filesystem;
+
+ptr<log_entry> empty_log_entry() {
+   static ptr<buffer>    buf = buffer::alloc(sz_ulong);
+   static ptr<log_entry> r   = cs_new<log_entry>(0, buf);
+   return r;
+}
+
+fs::path path_for_index(const fs::path& log_dir, ulong index) {
+   // the log is stored in log_dir/<log_index>.log
+   return (log_dir / std::to_string(index)).concat(".log");
+}
+
+fs::path store_log(const fs::path& log_dir, const unsigned char* data, int sz) {
+   namespace fs = std::filesystem;
+
+   auto file = log_dir / std::tmpnam(nullptr);
+   auto fp   = std::unique_ptr<FILE, decltype(&fclose)>{fopen(file.c_str(), "w"), &fclose};
+   if (!fp)
+      throw fs::filesystem_error("file open error", file, std::make_error_code(static_cast<std::errc>(errno)));
+   fwrite(data, 1, sz, fp.get());
+   return file;
+}
+
+struct log_entry_meta {
+   ulong term;
+   byte  type;
+
+   log_entry_meta(ulong tm = 0, log_val_type tp = log_val_type::app_log)
+       : term(tm)
+       , type(static_cast<byte>(tp)) {}
+};
+
+constexpr int log_entry_meta_size = sizeof(ulong) + sizeof(byte);
+
+fs::path store_log(const fs::path& log_dir, nuraft::ptr<nuraft::log_entry>& entry) {
+   namespace fs = std::filesystem;
+
+   auto file = log_dir / std::tmpnam(nullptr);
+   auto fp   = std::unique_ptr<FILE, decltype(&fclose)>{fopen(file.c_str(), "w"), &fclose};
+   if (!fp)
+      throw fs::filesystem_error("file open error", file, std::make_error_code(static_cast<std::errc>(errno)));
+
+   log_entry_meta meta{entry->get_term(), entry->get_val_type()};
+
+   fwrite(&meta, log_entry_meta_size, 1, fp.get());
+   auto& buf = entry->get_buf();
+   buf.pos(0); // reset the position
+   fwrite(buf.data(), 1, buf.size(), fp.get());
+   return file;
+}
+
+class log_entry_reader {
+ public:
+   log_entry_reader() = default;
+   log_entry_reader(fs::path log_dir, ulong index) { set(log_dir, index); }
+   ~log_entry_reader() {
+      if (!fname.empty()) {
+         std::error_code ec;
+         fs::remove(fname, ec);
+      }
+   }
+
+   log_entry_reader(const log_entry_reader&) = delete;
+   log_entry_reader& operator=(const log_entry_reader&) = delete;
+   log_entry_reader(log_entry_reader&& other) { fname.swap(other.fname); }
+
+   void set(fs::path log_dir, ulong index) {
+      auto original = path_for_index(log_dir, index);
+      if (fs::exists(original)) {
+         fname = log_dir / std::tmpnam(nullptr);
+         fs::create_hard_link(path_for_index(log_dir, index), fname);
+      }
+   }
+
+   ptr<log_entry> get() {
+      auto fp = std::unique_ptr<FILE, decltype(&fclose)>(fopen(fname.c_str(), "r"), &fclose);
+      if (fp) {
+         int            sz  = std::filesystem::file_size(fname) - log_entry_meta_size;
+         ptr<buffer>    buf = buffer::alloc(sz);
+         log_entry_meta meta;
+         fread(&meta, log_entry_meta_size, 1, fp.get());
+         fread(buf->get_raw(sz), sz, 1, fp.get());
+         buf->pos(0); // reset the position
+         return cs_new<log_entry>(meta.term, buf, static_cast<log_val_type>(meta.type));
+      }
+
+      return empty_log_entry();
+   }
+
+   size_t size() const {
+      std::error_code ec;
+      size_t          sz = fs::file_size(fname, ec);
+      return ec ? 0 : sz - log_entry_meta_size;
+   }
+
+ private:
+   fs::path fname;
+};
+
+log_store::log_store(std::filesystem::path p)
+    : log_dir_(p)
+    , start_idx_(1)
+    , next_slot_(1) {}
+
+ulong log_store::next_slot() const { return next_slot_; }
+
+ulong log_store::start_index() const { return start_idx_; }
+
+ptr<log_entry> log_store::last_entry() const {
+   log_entry_reader reader;
+   {
+      std::lock_guard<std::mutex> l(logs_lock_);
+      reader.set(log_dir_, next_slot_ - 1);
+   }
+   return reader.get();
+}
+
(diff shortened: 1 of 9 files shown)


programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+      index_reader.seek(offset);
+      index_reader.read(position);
+      return position;
+   }
+
+   ptr<buffer> pack(ulong index, int32 cnt) {
+      if (cnt == 0)
+         return ptr<buffer>{};
+      data_collection_t::const_iterator start_pack_itr = collection_.upper_bound(index);
+      if (start_pack_itr == collection_.cbegin()) {
```
kimjh2005, line 304: Shouldn't it be  (start_pack_itr == collection_.cend())

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+      for (; i < collection_.cend(); ++i, ++j) {
+         if (next_start_index != 0 && next_start_index != i->first) {
+            throw std::runtime_error("log_store: the persistent log store has missing entries");
+         }
+         next_start_index = i->first + j->second;
+      }
+
+      if (next_start_index > 0) {
+         // when the index and data file does not match, we trust index instead of the data file,
+         // because it's earier to make writing to the index file transactional.
```
kimjh2005, line 164: earier?

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+   void write_data_file_preamble(ulong start_index) {
+      char buf[128];
+      snprintf(buf, 128, "log-entry-%llu", start_index);
+
+      fs::path active_prefix = log_dir_ / buf;
+      catalog_.add(start_index, active_prefix);
+      data_out_  = file_writer(active_prefix.replace_extension("data"), "wb");
+      index_out_ = file_writer(active_prefix.replace_extension("index"), "wb");
+
+      data_out_.write(data_file_preamble{start_index});
```
b1bart, line 421: Footnote (1)

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+   }
+
+   fs::path current_log_prefix() { return collection_.rbegin()->second; }
+
+   int open(const fs::path& log_dir) {
+
+      const char*                               pattern = "log-entry-\\d+\\.index";
+      boost::container::flat_map<ulong, size_t> num_entries;
+
+      for_each_file_in_dir_matches(log_dir, pattern, [this, &num_entries](const fs::path& index_path) {
```
b1bart, line 135: Footnote (2)

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+   void append(ulong term, log_val_type type, const byte* payload, size_t payload_size) {
+      if (next_slot_ % log_stride_ == 0) {
+         write_data_file_preamble(next_slot_);
+      } else if (!data_out_.is_open() || !index_out_.is_open()) {
+         prepare_out_files_for_append();
+      }
+
+      data_out_.write(log_entry_header{term, type, static_cast<uint32_t>(payload_size)});
+      data_out_.sync_write(payload, payload_size);
+      uint64_t position = data_out_.cur_position();
```
b1bart, line 439: Footnote (3)

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+   void compact_back(ulong last_log_index) {
+      auto itr = collection_.lower_bound(last_log_index);
+      if (itr == collection_.end()) {
+         return;
+      }
+
+      std::for_each(collection_.rbegin(), data_collection_t::reverse_iterator(itr), [](const auto& entry) {
+         auto filebase = entry.second;
+         fs::remove(filebase.replace_extension("index"));
```
b1bart, line 244: Footnote (4)

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+      auto [seg_index, prefix] = *collection_.rbegin();
+      collection_.erase(itr, collection_.end());
+
+      if (!collection_.empty()) {
+         auto [seg_index, prefix] = *collection_.rbegin();
+         ulong num_entry_to_keep  = last_log_index - seg_index + 1;
+         auto  data_file_size     = end_of_entry_position(collection_.end() - 1, last_log_index);
+         fs::resize_file(fs::path(prefix).replace_extension("index"), num_entry_to_keep * sizeof(uint64_t));
+         fs::resize_file(fs::path(prefix).replace_extension("data"), data_file_size);
```
b1bart, line 256: Footnote (5)

programs/blockvaultd/log_store.cpp
```diff
@@ -0,0 +1,627 @@
...
+      data_collection_t::const_iterator itr = collection_.upper_bound(last_log_index);
+      if (itr == collection_.cbegin()) {
+         return;
+      }
+
+      if (itr->first != last_log_index + 1)
+         --itr;
+      std::for_each(collection_.cbegin(), itr, [](const auto& entry) {
+         auto filebase = entry.second;
+         fs::remove(filebase.replace_extension("index"));
```
b1bart, line 271: Footnote (6)

b1bart (changes requested):
This will be a substantial comment so I've taken the approach of putting the full thoughts into one place (here) and then footnotes as comments in the files which are relevant.

Broadly, I am concerned with the atomicity of updates to this log store and the effect that *may* have on node stability.  I will point out where my concerns stem from *but* I will also accept that my concerns may be unfounded if we can document and make clear that even without atomicity of updates the operation of a node and recovery from any reachable state is consistent. 

I will also be borrowing heavily from [this](https://danluu.com/file-consistency/) as a good resource on the various peculiarities of filesystems.  

I think it is important that we share some of the burden with operators such that we are employing best practices to signal to the OS, filesystem, hardware etc even if the operators have not deployed on a strictly optimal configuration.

## Concerns with the architecture

### Need for documentation
This PR implements a storage system with several logical operations however, there is no accessible document, in code or otherwise, that enumerates these operations.  I have tried to reconstruct a list below but, it may be incomplete as a result

### Create New Log/Index
Looking at (1) we have opened a data file and an index file and written to the data file.  There are no syncs and as such, program termination at this time *may* leave a stale data file and it *may* leave an empty index file but as the directory is not synced nor the writes we have little guarantees that either of those files exists or that their content is as expected. 

On recovery, we enumerate all the _index_ files (see (2)) but, as shown above we may have a data file with no index file.  The result of such a failure mode would be that upon the next write we would append an _additional_ preamble to the data file.  Furthermore, we cannot guarantee the content of the previously written preamble so, this can create a silently corrupted data log going forward.

### Append entry to existing Log/Index
At (3) we see that the current position is read after a write followed by a synchronous write.  This implies that there are no other operations on this file between the writes and this which in turn implies that any additional processes that are manipulating these files (for instance, another instance of BlockVault running in error or maintenance tools).  While we cannot guard against all possible interactions, that this will create a silent corruption in the log perhaps undiscovered for days, we should at least attempt to create some sort of lockfile in this log directory so that well-behaved processes can detect each other's presence and terminate rather than corrupt logs.

Also at (3), failure between the writes to the data and index seem handled in the restoration process IF all the data in the index is valid.  However, failure between the write and sync calls to the index file can leave that data in an undetermined state.  It may be worthwhile to augment the index with some basic checksum so that corrupted writes to the index file are detectable.  If we can guarantee that the index will not be written until the data is atomically and correctly stored then we should be able to avoid a checksum there.

Additionally, I believe there needs to be some amount of directory syncing in order to make sure that these operations are safe.  Please see the document linked at the top.

### Truncate Log/Index (remove the "newest" entries)
At (4) we see the removal of index/data file pairs.  While in code this appears in the "correct" order and removes the index before the data, without a sync on the directory containing these files, I am uncertain that the actual filesystem state will represent that strong ordering.

Likewise at (5), the partially truncated files are resized. This most certainly requires a sync at the directory level to be safe

Finally, there is a generalized atomicity violation here which we cannot easily fix.  Failure or termination of the process in serveral places will leave the *recovered* system in a state which is neither the state from before this logical operation nor the state after.  At best, this *may* represent a valid historical state of the log store but I am skeptical and it would assume that any other durably written artifacts (snapshots, cluster config, etc) are consistent with this rewound historical state. 

In that best case scenario, after recovery every thing would be consistent and then the expectation is that some work prior to the truncate will be repeated, building the log back up to the truncation point, at which point the operation will hopefully succeed?

Either we need to re-evaluate this operation to avoid the atomicity violation OR we need a lot more documentation that shows safety in all of the possible intermediate states.

### Compact Log (remove the oldest entries)
At (6) we see the same need for directory syncing to guarantee that indices are always removed before data files

Additionally, this has the same generalized atomicity concerns *however* the nature of this operation and its usage make this a non-issue.  For observers, this compaction step is an "optimization" of the data structure to save space.  Any intermediate state would merely be less optimal as long as the correctness concern is addressed.

## Concerns with aesthetics

There is an inconsistent usage of types like `uint32_t` which are very explicit in their size and sign and types like `ulong`.  Understandably, nuRaft is dictating some of these types however, I prefer to use `static_assert` to ensure the equivalence of the nuRaft type and the explicit type (eg `uint32_t`) for future proofing across platforms and then standardize on using only the explicit types in the implementation so that their representation when written to buffers/files is easier to intuit.

## The End

Considering the broad issues to address, it does not seem worthwhile to dig into more of the PR as there is likely to be collateral damage when addressing the above.

heifner:
https://danluu.com/file-consistency/ - "Sqlite is quite good in terms of reliability if you want a good default". Why don't we just use Sqlite?

kimjh2005:
I think that considering this log store is for raft library, the log store itself doesn't have to be perfect in every aspects. There can be a compromise between performance and stability. We have similar data/index scheme in blocks log which is stable.

heifner:
> I think that considering this log store is for raft library, the log store itself doesn't have to be perfect in every aspects. There can be a compromise between performance and stability. We have similar data/index scheme in blocks log which is stable.

Did you see that sqlite is actually more performant than raw file io.

kimjh2005:
In sqlite.org/fasterthanfs.html, sqlite is 35% faster than fs due to open() and close() call invoked only once.
"The performance difference arises (we believe) because when working from an SQLite database, the open() and close() system calls are invoked only once,"
But I am not against using sqlite but it is introduction of an another library. Can rocksdb be used instead of sqlite?

PR #9358, opened 2020-08-04
Telemetry - add cppkin support

heifner:
## Change Description

- Add support for https://github.com/EOSIO/cppKin via new `cppkin_plugin`
- EOSIO/cppkin support both Zipkin and New Relic
- Enabled by new CMake option: `ENABLE_CPPKIN`
  -- Needs `EOSIO/cppkin` installed to build

Adds cppkin logging for http/amqp transactions, block production, SHiP block receive/send, rodoes block receive/processed.

### **Trace entries added**:
- amqp_trx_plugin
  -- Trace: "Transaction", Span: "AMQP Received", Tag: "trx_id" [received an input transaction from AMQP "trx" queue]
  -- Trace: "Transaction", Span: "Processed", Tag: "trx_id" [executed transaction and placed in pending block]
      ---- Includes: Tags: "block_num", "block_time", "elapsed" and either "status" or "error"

- chain_plugin
  -- Trace: "Transaction", Span: "HTTP Received", Tag: "trx_id" [received an input transaction from http push_transaction]
  -- Trace: "Transaction", Span: "Processed", Tag: "trx_id" [executed transaction and placed in pending block]
      ---- Includes: Tags: "block_num", "block_time", "elapsed" and either "status" or "error"
  -- Trace: "Transaction", Span: "HTTP Received", Tag: "trx_id" [received an input transaction from http send_transaction]
  -- Trace: "Transaction", Span: "Processed", Tag: "trx_id" [executed transaction and placed in pending block]
      ---- Includes: Tags: "block_num", "block_time", "elapsed" and either "status" or "error"

- net_plugin
  -- Trace: "Block", Span: "Accepted", Tags: "block_id", "block_num", "block_time"
  -- Trace: "Block", Span: "PreAccepted", Tags: "block_id", "block_num", "block_time", "producer" [Logged when trusted producer allows for sending of block before processing]

- state_history_plugin
  -- Trace: "Block", Span: "SHiP-Accepted", Tags: "block_id", "block_num", "block_time"
  -- Trace: "Block", Span: "SHiP-Send", Tags: "block_id", "block_num", "block_time"

- rodeos
  -- Trace: "Block", Span: "rodeos-received", Tags: "block_id", "block_num", "block_time"
  -- Trace: "Block", Span: "Filter-Start", Tags: "block_num"
  -- Trace: "Block", Span: "Filter-Stop", Tags: "block_num"

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

[...]

PR #9579, opened 2020-10-20
Zipkin appender

heifner:
## Change Description

- Add new `zipkin_appender` for sending trace log messages to zipkin.
- Enabled by adding `zipkin` appender type and `zipkin` logger to `logging.json`:
```
   {
      "name": "zip",
      "type": "zipkin",
      "args": {
        "endpoint": "http://127.0.0.1:9411",
        "path": "/api/v2/spans",
        "service_name": "nodeos",
        "timeout_us": 200000
      },
      "enabled": true
    }
```
`zipkin` logger with `debug` level required for the new logger.
```
    {
      "name": "zipkin",
      "level": "debug",
      "enabled": true,
      "additivity": false,
      "appenders": [
        "zip"
      ]
   }
```
The `zipkin_appender` sends zipkin Spans to zipkin via http json on a separate thread. `timeout_us` can be configured for how long to allow each http post before hitting the time out. Nine consecutive failures will disable the zipkin_appender until a restart of application or SIGHUP.

- This PR removes the `EOSIO/cppkin` integration added by #9358. `EOSIO/cppkin` is a more full-featured zipkin solution but it pulls in more shared libraries than desired.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


libraries/fc/src/log/zipkin_appender.cpp
```diff
@@ -0,0 +1,164 @@
...
+      if( !endpoint ) {
+         endpoint = url( cfg.endpoint + cfg.path );
+         std::cout << "Connecting to zipkin: " << (std::string) *endpoint << std::endl;
+      }
+
+      http.post_sync( *endpoint, create_zipkin_variant( message, cfg.service_name ), deadline );
+
+      consecutive_errors = 0;
+   } catch( const fc::exception& e ) {
+      std::cerr << "Unable to connect to zipkin: " << cfg.endpoint + cfg.path << ", error: " << e.to_detail_string() << std::endl;
```
spoonincode, line 155: can these be something like a `wlog()` instead? I know it seems weird for an appender to be... appending more logs. But it seems the zipkin_appender is intended solely for zipkin trace spans, not the default logger, if I'm understanding it right?

heifner: Nothing would prevent someone from setting up the `default` logger from using the `zip` appender causing infinite logging.

spoonincode: does that make sense though? (should it even be allowed?)

heifner: It should not be allowed. I added logic to not allows it and used default logging in `zipkin_appender`.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -123,37 +123,36 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
       const auto& tid = trx->id();
       dlog( "received packed_transaction ${id}", ("id", tid) );
 
-      auto trx_trace = fc_create_trace("Transaction");
+      auto trx_trace = fc_create_trace_with_id("Transaction", tid);
```
swatanabe-b1, line 126: This is before transaction deduplication, so if we receive the same transaction more than once, they'll get the same trace id.

libraries/fc/src/log/zipkin_appender.cpp
```diff
@@ -0,0 +1,164 @@
...
+   }
+   i = d.find( "zipkin.duration" );
+   if( i != d.end() ) {
+      mvo( "duration", i->value() );
+      d.erase( "zipkin.duration" );
+   }
+
+   mvo( "localEndpoint", fc::variant_object( "serviceName", service_name ) );
+
+   mvo( "tags", d );
```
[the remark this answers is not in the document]
heifner: Thanks. That is what the spec says. In practice int was allowed, but certainly do not want to allow object. So I'll restrict it to string.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
swatanabe-b1: Although, maybe that's what we want?

heifner: Yes. I think that would be desired so that they are grouped together in the zipkin ui. Note the zipkin ui does not overwrite when they have the same id, it groups them together.

libraries/fc/include/fc/log/trace.hpp
```diff
@@ -1,63 +1,163 @@
...
+
+      template<typename T>
+      static uint64_t to_id( const T& id ) {
+         static_assert( std::is_same_v<decltype(id.data()), const uint64_t*>, "expected uint64_t" );
+         return id.data()[3];
+      }
+
+   private:
+      uint64_t id;
+      // zipkin traceId and parentId are same since only allowing trace with span children, no children of spans
```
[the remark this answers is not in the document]
heifner: My understanding is "root span" <==> "trace". So one span (the root span, or trace) has no parentId. All the spans then share the traceId. Since currently only supporting trace with spans (no child spans of spans) then traceId and parentId are always the same.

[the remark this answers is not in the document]
heifner: Updated the comment to be more clear.

libraries/fc/src/log/zipkin_appender.cpp
```diff
@@ -0,0 +1,164 @@
...
+      uint64_t v = i->value().as_uint64();
+      mvo( "traceId", fc::to_hex( reinterpret_cast<const char*>(&v), sizeof( v ) ) );
+      d.erase( "zipkin.traceId" );
+   }
+   i = d.find( "zipkin.id" );
+   if( i != d.end() && i->value().is_uint64() ) {
+      uint64_t v = i->value().as_uint64();
+      mvo( "id", fc::to_hex( reinterpret_cast<const char*>(&v), sizeof( v ) ) );
+      d.erase( "zipkin.id" );
+   }
```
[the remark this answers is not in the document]
heifner: My thought was to just send whatever was present. But enforcing required `id` and `traceId` would probably be helpful to indicate problems early.

libraries/fc/src/log/zipkin_appender.cpp
```diff
@@ -0,0 +1,164 @@
...
+   //   int64_t     duration  // microseconds of span
+
+   log_context context = message.get_context();
+   fc::mutable_variant_object d = message.get_data();
+
+   fc::mutable_variant_object mvo;
+   auto i = d.find( "zipkin.traceId" );
+   if( i != d.end() && i->value().is_uint64() ) {
+      uint64_t v = i->value().as_uint64();
+      mvo( "traceId", fc::to_hex( reinterpret_cast<const char*>(&v), sizeof( v ) ) );
```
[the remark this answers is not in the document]
heifner: Discussed and since transaction id and block id are often used, going to keep the endianess as provided.

libraries/fc/src/log/logger_config.cpp
```diff
@@ -25,6 +27,17 @@ namespace fc {
       return true;
    }
 
+   uint64_t log_config::get_next_unique_id() {
+      auto& lc = log_config::get();
+      std::lock_guard g( lc.log_mutex );
+      if( lc.next_id == 0 ) {
+         std::mt19937_64 engine( std::random_device{}() );
+         std::uniform_int_distribution<uint64_t> distribution;
+         lc.next_id = distribution( engine );
```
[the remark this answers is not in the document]
heifner: Fixed.

libraries/fc/src/log/logger_config.cpp
```diff
@@ -76,6 +96,9 @@ namespace fc {
          }
          auto ap = fact_itr->second->create( cfg.appenders[i].args );
          log_config::get().appender_map[cfg.appenders[i].name] = ap;
+         if( cfg.appenders[i].type == "zipkin" ) {
+            zipkin_appender = ap;
```
swatanabe-b1, line 100: What if there's more than one zipkin appender?

libraries/fc/include/fc/log/trace.hpp
```diff
@@ -1,63 +1,183 @@
...
+         constexpr explicit operator bool() const noexcept { return id != 0; }
+      private:
+         explicit token(uint64_t id)
+         : id(id) {}
+         uint64_t id;
+      };
 
+      token get_token() const { return token{id}; };
+
+      static uint64_t to_id( const fc::sha256& id ) {
```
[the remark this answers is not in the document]
swatanabe-b1: Also, what if it's 0?

libraries/fc/src/log/logger_config.cpp
```diff
@@ -49,6 +62,12 @@ namespace fc {
          iter.second->initialize( ios );
    }
 
+   void log_config::shutdown_appenders() {
+      std::lock_guard g( log_config::get().log_mutex );
+      for( auto& iter : log_config::get().appender_map )
+         iter.second->shutdown();
```
swatanabe-b1, line 68: Indirect calls while holding a mutex are dangerous.  In this case it can deadlock:
- this thread blocks joining the zipkin thread
- the zipkin thread blocks in dlog or wlog trying to acquire the `log_mutex`

swatanabe-b1 (changes requested):

libraries/fc/include/fc/log/trace.hpp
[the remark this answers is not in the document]
heifner: Collisions should be fine, zipkin will just group them together, but a monitor can tell them apart via the tags.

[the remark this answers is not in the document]
swatanabe-b1: I could only see 0 happening of someone intentionally "mines" transactions to slip something under the radar.

libraries/fc/src/log/logger_config.cpp
heifner: I agree this is a bad pattern. However, it follows initialize and log which are also called holding the lock. I think I can just reset the lock guard in `zipkin_appender::shutdown()`

heifner: Actually, I forgot why this was all even needed. It is needed because the loggers explicitly leak.
https://github.com/EOSIO/fc/blob/master/src/log/logger_config.cpp#L16
So I wanted a hook that would allow for purging of all queued up zipkin messages.
This is not very clean. :(

libraries/fc/src/log/zipkin_appender.cpp
heifner: I thought it would be safe to call `wlog` and `dlog` since from a diff thread. However, the `thread.join()` causes a potential deadlock. I think the safest thing is to just go back to `std::cout` & `std::cerr`

libraries/fc/src/log/logger_config.cpp
swatanabe-b1: The thread has to be joined, or you have undefined behavior (unless it only executes functions that are permitted in a signal handler, which I greatly doubt.).

heifner: BTW, this same deadlock is possible on `configure_logging` where appenders are destroyed via sighup.

programs/rodeos/cloner_plugin.cpp
```diff
@@ -283,15 +290,23 @@ void cloner_plugin::plugin_initialize(const variables_map& options) {
...
+   handle_sighup();
+   my->start();
+}
 
 void cloner_plugin::plugin_shutdown() {
    if (my->session)
       my->session->connection->close(false);
    my->timer.cancel();
+   fc::log_config::shutdown_appenders();
    ilog("cloner_plugin stopped");
```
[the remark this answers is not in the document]
heifner: Moved the interface to `logger` so individual loggers can shutdown appenders if desired.

libraries/fc/include/fc/log/trace.hpp
```diff
@@ -1,63 +1,183 @@
...
 #include <fc/log/logger.hpp>
-#include <cppkin.h>
+#include <fc/log/logger_config.hpp>
+#include <fc/crypto/sha256.hpp>
+#include <optional>
+#include <type_traits>
+
+/// Simple wrappers for zipkin tracing, see zipkin_appender
+namespace fc {
+   inline std::string zipkin_logger_name = "zipkin";
```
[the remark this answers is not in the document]
heifner: yep

libraries/fc/src/log/logger_config.cpp
heifner: Reworked this to look at type directly.

swatanabe-b1 (approved):

PR #9593, opened 2020-10-22
Speculative validation optimizations

heifner:
## Change Description

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
PR reviewer: Verify that run-away memory is not an issue with possibility of trxs making it into the `unapplied_transaction_queue` but never being cleared. I don't think this is an issue, but something to consider.

b1bart (approved):

PR #9595, opened 2020-10-23
fixed cleos get_kv_table_rows bugs

kimjh2005:
## Change Description

Fixed get_kv_table_rows bugs:
-Made upper_bound optional when reverse option is not set
-Made lower_bound optional when reverse option is set
-Fixed when upper_bound value is not found, the iterator points to end value
-Changed --index can return multiple rows
-Changed next_key format to hex string so that the next call can be called easily using "--encode-type bytes"
-Replaced -k option with -i and --index in order to be in sync with get_table_rows
 
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

cleos 
-k  option removed
-i, --input added.

> cleos get kv_table --help
Retrieve the contents of a database kv_table
Usage: bin/cleos get kv_table [OPTIONS] account table index_name

Positionals:
  account TEXT REQUIRED       The account who owns the table
  table TEXT REQUIRED         The name of the kv_table as specified by the contract abi
  index_name TEXT REQUIRED    The name of the kv_table index as specified by the contract abi

Options:
  -h,--help                   Print this help message and exit
  -l,--limit UINT             The maximum number of rows to return
  -i,--index TEXT             Index value
  -L,--lower TEXT             JSON representation of lower bound value of key, defaults to first
  -U,--upper TEXT             JSON representation of upper bound value of key, defaults to last
  --encode-type TEXT          The encoding type of index_value, lower bound, upper bound 'bytes' for hexdecimal encoded bytes 'string' for string value 'dec' for decimal encoding of (uint[64|32|16|8], int[64|32|16|8], float64) 'hex' for hexdecimal encoding of (uint[64|32|16|8], int[64|32|16|8], sha256, ripemd160
  -b,--binary                 Return the value as BINARY rather than using abi to interpret as JSON
  -r,--reverse                Iterate in reverse order
  --show-payer                Show RAM payer

## Documentation Additions
- [ ] Documentation Additions

When --reverse option is not set,  --upper is optional, If --upper is not set, the result includes the end of the matching rows.

When --reverse option is set , --lower is optional, if --lower is not set,  the result includes the begning of the matchin rows.

-k is replaced by -i/--input option to be in sync. with get_table_rows() call

When result has more = true and next_key is set,  the following call to get_kv_table_rows should set --encode-type bytes and 
set the next_key value to  index or lower/upper value.

When --index is used to search non unique secondary index,  result can return multiple rows.


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2172,80 +2172,59 @@ read_only::get_table_rows_result read_only::get_kv_table_rows_context( const rea
...
 
    ///////////////////////////////////////////////////////////
    // ranged query
    ///////////////////////////////////////////////////////////
    bool has_lower_bound = p.lower_bound.has_value() && !p.lower_bound.value().empty();
    bool has_upper_bound = p.upper_bound.has_value() && !p.upper_bound.value().empty();
-   EOS_ASSERT(has_lower_bound && has_upper_bound, chain::contract_table_query_exception,  "Unknown range: ${t} ${i}", ("t", p.table)("i", p.index_name));
+   unbounded = !has_lower_bound || !has_upper_bound;
+   // reverse mode has upper_bound value, non-reverse and point query has lower bound value
+   EOS_ASSERT((p.reverse && has_upper_bound) || (!p.reverse && has_lower_bound) || (has_lower_bound && point_query), chain::contract_table_query_exception, "Unknown/Invalid range: ${t} ${i}", ("t", p.table)("i", p.index_name));
```
linhuang-blockone, line 2192: Maybe you can change `(!p.reverse && has_lower_bound) || (has_lower_bound && point_query)` to `(!p.reverse || point_query) && has_lower_bound)` to match the comment and be consistent with `(p.reverse && has_upper_bound)`

tests/get_kv_table_nodeos_tests.cpp
```diff
@@ -33,10 +33,10 @@ using namespace eosio::chain;
 using namespace eosio::testing;
 using namespace fc;
```
linhuang-blockone, line 35: Why not this file in unittests directory? I thought test directory was for integration testing and they were all python files.

linhuang-blockone (approved):

PR #9603, opened 2020-10-27
add KV Table iterator bool operator

vzqzhang:
## Change Description

Added bool operator to KV table iterator so that developers can easily check if the iterator exists by using an `if (itr) {}` statement.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/rodeos/include/eosio/key_value.hpp
+++ b/libraries/rodeos/include/eosio/key_value.hpp
@@ -403,6 +403,10 @@ class kv_table {
          return compare(b) >= 0;
       }
 
+      explicit operator bool() {
+         return compare(this) != 0;
+      }
+
    private:
       uint32_t itr;
       status itr_stat;


libraries/rodeos/include/eosio/key_value.hpp
```diff
@@ -403,6 +403,10 @@ class kv_table {
          return compare(b) >= 0;
       }
 
+      explicit operator bool() {
```
swatanabe-b1, line 406: const

libraries/rodeos/include/eosio/key_value.hpp
```diff
@@ -403,6 +403,10 @@ class kv_table {
...
+         return compare(this) != 0;
```
swatanabe-b1, line 407: An iterator is always equal to itself.  Also, there is no matching overload of compare.

larryk85:
1) We don't need changes to rodeos at this time.
2) For internal eos needs, I don't think this would buys us much in the long run.

PR #9605, opened 2020-10-27
fix get_table_rows_by_seckey conversion

kimjh2005:
## Change Description

fixed function get_table_rows_by_seckey converts lower and upper value in eosio::name incorrectly.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


dimas1185:
can you explain why get_table_rows_by_seckey was working incorrectly if you used string representation of name instead of string representaton of underlying integer?

kimjh2005:
convert_to_type calls boost::lexical_cast<uint64_t>( str.c_str()) 
Before the fix,  lexical_cast<uint64_t>("1111" ) is called, 
After the fix, lexical_cast<uint64_t>("1324356665524535") is called.    //"1324356665524535" is eosio::name in uint64_t string.

The return value is:
Before the fix is 1111,
After the fix is eosio::name in uint64_t number.


template<>
uint64_t convert_to_type(const string& str, const string& desc) {
   try {
      return boost::lexical_cast<uint64_t>(str.c_str(), str.size());
   } catch( ... ) { }

dimas1185:
@kimjh2005 that is understood. I don't get why `lexical_cast<uint64_t>("1111" )` is an issue. it should return valid value, isn't it? so why raw uint64 representation is preferred?

kimjh2005:
the return value 1111 is not correct value for eosio::name.     "1111"_n should have uint64_t number which is not 1111.

dimas1185:
@kimjh2005 eosio::chain::name "1111"_n is valid name. However I looked in the code and I understood why it fails. I'll propose some alternative solution inline in comments.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -491,7 +491,7 @@ class read_only {
          if( p.lower_bound.size() ) {
             if( p.key_type == "name" ) {
                name s(p.lower_bound);
-               SecKeyType lv = convert_to_type<SecKeyType>( s.to_string(), "lower_bound name" ); // avoids compiler error
+               SecKeyType lv = convert_to_type<SecKeyType>( std::to_string(s.to_uint64_t()), "lower_bound name" ); // avoids compiler error
```
dimas1185, line 494: look what you are doing here:
1. convert string to uint64 name. 
2. convert resulted uint64 to string.
3. convert resulted string to uint64.

I propose template straightforward solution to avoid that:
1. change declaration of `convert_to_type`:
```
template<typename Type, bool IsName=false>
Type convert_to_type(const string& str, const string& desc)

```

2. declare corresponding template specialization:
```
template<>
uint64_t convert_to_type<uint64_t, true>(const string& str, const string& desc);
```

implementation is something like this:
```
template<>
uint64_t convert_to_type<uint64_t, true>(const string& str, const string& desc) {
   try {
      return chain::string_to_uint64_t(str);
   } FC_RETHROW_EXCEPTIONS(warn, "Could not convert ${desc} name string '${str}' to key type.", ("desc", desc)("str",str) );
}
```

change line 494 (and same for 505) to:
`SecKeyType lv = convert_to_type<SecKeyType, true>( p.lower_bound, "lower_bound" );`

larryk85: I see where are you are going here. But, you can't specialize template functions the way you are wanting.  The "specialization" is from the return type.

I think a sufficient approach is to just create an overload that takes an `eosio::name` as the first parameter.

So,

``` c++
uint64_t convert_to_type(const eosio::name& n, const std::string& desc) {
   // to_uint64_t is basically noexcept
   return n.to_uint64_t();
}
```

Then in your caller, 
``` c++
SecKeyType lv = convert_to_type(s /*or name{p.lower_bound} */, "lower_bound");
```

larryk85 (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
dimas1185: @larryk85 I tested it and it works. So at least clang permits this.
Your suggested alternative looks simpler than mine. Good idea.

larryk85: > @larryk85 I tested it and it works. So at least clang permits this.
> Your suggested alternative looks simpler than mine. Good idea.
> However it may need
> `constexpr if (is_same_v<uint64_t, SecKeyType>)` for compiler to compile.

1) Ah, I think I missed that you weren't trying to do partial specialization (didn't read it very well, my apologies).  I thought the pattern you were going with would've required partial specialization, I didn't pay attention to the defaulted value.

2) Yes, good call.  It will need an `if constexpr` wrapped around that call.  And an exception/error on the else branch, because it shouldn't be in that state.  If it did then something has gone awry.

dimas1185 (approved):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -490,9 +492,12 @@ class read_only {
 
          if( p.lower_bound.size() ) {
             if( p.key_type == "name" ) {
-               name s(p.lower_bound);
-               SecKeyType lv = convert_to_type<SecKeyType>( s.to_string(), "lower_bound name" ); // avoids compiler error
-               std::get<1>(lower_bound_lookup_tuple) = conv( lv );
+               if constexpr (std::is_same_v<uint64_t, SecKeyType>) {
```
brianjohnson5972, line 495: @larryk85 @b1bart I'm showing up a little late to this, but this sure doesn't seem like what we want.  If "name" means "expect a uint64_t" why don't we just call it "uint64_t".  I think what we really want is something that checks to see if the parameter has any letters, '.', or just 1-5 and we use string, otherwise we take it as a uint64_t.  For get_table_rows_ex, "name" is still treated as a string to be passed to a name{}, and I don't think it makes sense to handle it 2 different ways.

brianjohnson5972 (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
brianjohnson5972: Need to resolve this, since both of the mentioned interfaces effect the RPC code in working on

brianjohnson5972: I don't know if I just got confused by the comments that I read from earlier, or if my brain isn't working, because this is correct. I thought from the discussions that this was just testing everything as uint64.

larryk85 (approved):

PR #9632, opened 2020-11-03
Fixes for DB intrinsic replay logic

brianjohnson5972:
## Change Description
Fixed issue with finding a rocksdb key in a table to identify if the table needs to be removed.  Also fixed issue with identifying the last key entry in a table using the table end iterator.

## Change Type
**Select ONE**
- [] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/backing_store/db_context_rocksdb.cpp
+++ b/libraries/chain/backing_store/db_context_rocksdb.cpp
@@ -338,7 +338,7 @@ namespace eosio { namespace chain { namespace backing_store {
          if (past_end(session_iter)) {
             --session_iter;
             // either way, the iterator after our known table should have a primary key in this table as the previous iterator
-            if (!past_end(session_iter) || session_iter == current_session.begin()) {
+            if (past_end(session_iter)) {
                // NOTE: matching chainbase functionality, if iterator store found, but no key in db for table
                return primary_iter_store.invalid_iterator();
             }
--- a/libraries/chain/backing_store/db_key_value_any_lookup.cpp
+++ b/libraries/chain/backing_store/db_key_value_any_lookup.cpp
@@ -45,7 +45,7 @@ namespace eosio { namespace chain { namespace backing_store {
 
    void db_key_value_any_lookup::remove_table_if_empty(const shared_bytes& key) {
       // look for any other entries in the table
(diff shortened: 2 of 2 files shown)


libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -338,7 +338,7 @@ namespace eosio { namespace chain { namespace backing_store {
          if (past_end(session_iter)) {
             --session_iter;
             // either way, the iterator after our known table should have a primary key in this table as the previous iterator
-            if (!past_end(session_iter) || session_iter == current_session.begin()) {
+            if (past_end(session_iter)) {
```
larryk85, line 341: It seems like the original intent was off by one line (the `--session_iter` should've come after line 341).  Does this change semantically mean the same as that?

It seems like iterating past begin() would be "undefined" behavior.  The prefix may or may not match (just depends on what is there in memory), but maybe iterating past begin will give you `end`.  I couldn't fully determine that that is the case though.

larryk85 (changes requested):

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: The session::iterator has logic to set the iterator to current_session.end() (@TimothyBanks chime in here), so I was letting the second past_end call handle it (since it checks for current_session.end()).  But since you bring it up, we may not want to rely on that behavior. But if we don't want to, then I think we should assert for --current_session.begin(), or at the very least assure across the board that we are doing that and document that the begin() iterating to end() logic shouldn't be expected.

TimothyBanks: Iterators on session are circular.  So we can iterator past begin and wrap around to the end iterator.

larryk85: Gotcha, I didn't remember at which layer we were doing the wrapping (I was thinking it was being handled more towards the WASM API).

larryk85 (approved):

PR #9636, opened 2020-11-05
update the block_vault_interface with async semantic

huangminghuang:
## Change Description

* Move the blockvault.hpp to blockvault_client_plugin
* update block_vault_interface with async semantic
* update blockvault_client_plugin to fetch the block_vault_interface interface.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/blockvault_client_plugin/include/eosio/blockvault_client_plugin/blockvault.hpp
```diff
@@ -0,0 +1,98 @@
+#pragma once
+#include <eosio/chain/block.hpp>
+#include <eosio/chain/block_timestamp.hpp>
+#include <stdint.h>
+#include <string_view>
+
+namespace eosio {
+namespace blockvault {
+using watermark_t = std::pair<uint32_t, eosio::chain::block_timestamp_type>;
```
ndcgundlach, line 9: Isn't this defined elsewhere?

plugins/blockvault_client_plugin/include/eosio/blockvault_client_plugin/blockvault.hpp
```diff
@@ -0,0 +1,98 @@
...
+   /// interrupted due to inability to communicate with the chosen Block Vault
+   /// node, it must restart from the beginning with a new node.
+   ///
+   /// \param block_id The BlockID of the best known final block on the client
+   ///
+   /// \param callback The callback object to receive the snapshot and blocks
+   ///
+   virtual void sync(const eosio::chain::block_id_type& block_id, sync_callback& callback) = 0;
+};
+
```
ndcgundlach, line 95: Weren't we going to have some factory method (aka get())?

plugins/blockvault_client_plugin/include/eosio/blockvault_client_plugin/blockvault.hpp
huangminghuang: producer_plugin.cpp has it, but I cannot include it as my dependency.

ndcgundlach (dismissed):

plugins/blockvault_client_plugin/include/eosio/blockvault_client_plugin/blockvault.hpp
johndebord: It was defined here:
https://github.com/EOSIO/eos/blob/blockvault/libraries/blockvault/include/blockvault/blockvault.hpp#L8

So does this mean we're getting rid of `eos/libraries/blockvault`?

johndebord (dismissed):

b1bart:
One thing to consider, as this plugin is basically just a factory for an implementation of a pure virtual interface class, is changing this to something completely decoupled via an `appbase::method` with a `first_provider_policy` : 

https://github.com/EOSIO/appbase/blob/144b2e239d6fd93a8336543bf9eda7c52ea8c77e/include/appbase/method.hpp#L99-L116
 
This would allow you to provide a default implementation of the method which returns `nullptr` and then as a result of compilation of a program (like `nodeos`) the `block_vault_postgres_client_plugin` could provide a higher priority implementation that returns a pointer to that implementation.

In essence we get the ability to make providers first class plugins, potentially out-of-tree, rather than hiding that logic behind the factory plugin.  

This means in the future, a user could implement a custom plugin without any change to our core plugins that "takes over" the role of providing the blockvault interface factory.

b1bart:
you can see the "method" version in action here: https://github.com/EOSIO/eos/blob/develop/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp#L47-L48

both `transaction_async` and `block_sync` are provided by the `producer_plugin` at the default priority (0) but if that was provided at a non-zero priority some *other* plugin could take over the responsibility of handling those calls by providing a lower priority handler

huangminghuang:
> One thing to consider, as this plugin is basically just a factory for an implementation of a pure virtual interface class, is changing this to something completely decoupled via an `appbase::method` with a `first_provider_policy` :
> 
> https://github.com/EOSIO/appbase/blob/144b2e239d6fd93a8336543bf9eda7c52ea8c77e/include/appbase/method.hpp#L99-L116
> 
> This would allow you to provide a default implementation of the method which returns `nullptr` and then as a result of compilation of a program (like `nodeos`) the `block_vault_postgres_client_plugin` could provide a higher priority implementation that returns a pointer to that implementation.
> 
> In essence we get the ability to make providers first class plugins, potentially out-of-tree, rather than hiding that logic behind the factory plugin.
> 
> This means in the future, a user could implement a custom plugin without any change to our core plugins that "takes over" the role of providing the blockvault interface factory.

ndcgundlach (approved):

PR #9646, opened 2020-11-09
fix get_kv_table_rows secondary index search

kimjh2005:
## Change Description

fixed a bug that secondary index search failed to return values.
kv_table database entry whose key is secondary key contains the primarykey value. To retrieve the row value, you need to read the database entry containing the primarykey. The bug is it used the wrong value size for the  row value. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2293,17 +2296,19 @@ read_only::get_table_rows_result read_only::get_kv_table_rows_context( const rea
 
          row_value.clear();
          row_value.resize(value_size);
-         status = itr->kv_it_value(offset, row_value.data(), key_size, value_size);
-         EOS_ASSERT(status == chain::kv_it_stat::iterator_ok, chain::contract_table_query_exception,  "Invalid key in ${t} ${i}", ("t", p.table)("i", p.index_name));
+         status = itr->kv_it_value(offset, row_value.data(), value_size, actual_size);
```
kimjh2005, line 2299: This is the fix:  change key_size to value_size. It caused row_value trimmed and failed the subsequence search in line 2305.

linhuang-blockone: Good. All tests passed?

linhuang-blockone (approved):

PR #9651, opened 2020-11-10
[blockvault] New producer plugin logic for blockvault

johndebord:


plugins/blockvault_client_plugin/CMakeLists.txt
```diff
@@ -1,7 +1,13 @@
 file(GLOB HEADERS "include/eosio/blockvault_client_plugin/*.hpp")
+
 add_library( blockvault_client_plugin
              blockvault_client_plugin.cpp
              ${HEADERS} )
 
-target_link_libraries( blockvault_client_plugin appbase fc http_plugin chain_plugin eosio_testing )
-target_include_directories( blockvault_client_plugin PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" )
+target_link_libraries( blockvault_client_plugin appbase fc http_plugin chain_plugin signature_provider_plugin eosio_chain eosio_testing )
```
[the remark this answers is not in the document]
heifner: `http_plugin` & `chain_plugin` also not needed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -381,12 +383,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             throw;
          };
 
+         // Notes:
+         // - as soon as block is accepted via `push_block` a signal is emitted.
+         // - may be able to just overload `push_block` to _verify_ the block if blockvault is currently active; will have to just pass noop lambda:
+         //   - this is no longer planned.
```
heifner, line 389: Remove comments.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -381,12 +383,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
          try {
             chain.push_block( bsf, [this]( const branch_type& forked_branch ) {
                _unapplied_transactions.add_forked( forked_branch );
             }, [this]( const transaction_id_type& id ) {
                return _unapplied_transactions.get_trx( id );
             } );
+            if ( blockvault_plug ) {
+               uint32_t lib{chain.head_block_state()->dpos_irreversible_blocknum};
```
heifner, line 397: `head_block_state` is not necessarily the block you just pushed. Could return the `block_state_ptr` from `push_block`, or eval the future before calling `push_block`, or get it from the `accepted_block_header` signal.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -381,12 +383,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               eosio::chain::signed_block_ptr block{chain.head_block_state()->block};
```
heifner, line 398: No reason to make a copy of the `signed_block_ptr`, just pass it directly to `append_external_block`. Again `head_block_state` is not necessarily the block you just pushed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -381,12 +383,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               std::function<void(bool)> handler{[](bool){}};
```
heifner, line 399: I think this would be cleaner to just pass directly:  `blockvault_plug->append_external_block( .., .., [](bool){} );`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -718,7 +730,9 @@ if( options.count(op_name) ) { \
 void producer_plugin::plugin_initialize(const boost::program_options::variables_map& options)
 { try {
    my->chain_plug = app().find_plugin<chain_plugin>();
+   my->blockvault_plug = app().find_plugin<blockvault_client_plugin>();
    EOS_ASSERT( my->chain_plug, plugin_config_exception, "chain_plugin not found" );
+   EOS_ASSERT( my->blockvault_plug, plugin_config_exception, "blockvault_client_plugin not found" );
```
heifner, line 735: This should be removed. You don't want to require blockvault_plug to always be available.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2043,7 +2057,7 @@ void producer_plugin_impl::produce_block() {
    }
 
    //idump( (fc::time_point::now() - chain.pending_block_time()) );
-   chain.finalize_block( [&]( const digest_type& d ) {
+   chain.finalize_block( [&]( const digest_type& d ) { // get dpos irr block num
```
heifner, line 2060: Remove comment.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,7 +2069,47 @@ void producer_plugin_impl::produce_block() {
...
+   //   - or call `chain.head_block_state()` of which you might call `account_name controller::pending_block_producer() const`.
+   // - to get LIBid or LIBnumber:
+   //   - both can be found in controller.
+   //   - for this case do not get the chain last irreversible block number but get the LIB number from the head block state.
+   //   - get dpos irreversible block number
+   //   - block header state also has the producer.
+   //   - block header state also has the signed block as well.
+   // - it was talked about how another function could be added to controller for dealing with precursor cases prior to calling `commit_block`.
+   // - not sure why I wrote this function down but I'm keeping it around: `chain.fetch_block_by_number()`.
+   // - also keeping this note around `bsp = chain.finalize_block( [&]( const digest_type& d )` // get dpos irreversible block number
```
heifner, line 2091: Remove all these comments.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,7 +2069,47 @@ void producer_plugin_impl::produce_block() {
...
+
+   if ( blockvault_plug ) {
+      blockvault::watermark_t watermark{get_watermark(chain.head_block_state()->get_scheduled_producer(calculate_pending_block_time()).producer_name).value()};
```
heifner, line 2094: pending block is the block state you want, not `head_block_state`.
`finialize_block` returns the `block_state_ptr`, use that.
get_watermark can return an empty optional. That should not be possible in this case, but best to EOS_ASSERT that.
Use the `producer` of the block, not the scheduled producer.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,7 +2069,47 @@ void producer_plugin_impl::produce_block() {
...
+      uint32_t lib{chain.head_block_state()->dpos_irreversible_blocknum};
```
heifner, line 2095: Use the `dpos_irreversible_blocknum` of the blockstate being produced not the `head_block_state`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,7 +2069,47 @@ void producer_plugin_impl::produce_block() {
...
+      signed_block_ptr block{chain.head_block_state()->block};
```
heifner, line 2096: No reason to make a copy of the signed_block_ptr

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,7 +2069,47 @@ void producer_plugin_impl::produce_block() {
...
+         } else {
+            _unapplied_transactions.add_aborted( chain.abort_block() );
+         }
+      }};
+      blockvault_plug->propose_constructed_block(watermark, lib, block, handler);
+   } else {
+      chain.commit_block();
+   }
 
    block_state_ptr new_bs = chain.head_block_state();
```
heifner, line 2114: If `proposed_constructed_block` determines that the block is no good then you should not log a "Produced block" here.

heifner: `maybe_produce_block` should return `false` if unable to produce the block. It will return `true` even if the block is aborted now.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,7 +2069,47 @@ void producer_plugin_impl::produce_block() {
...
+           block_state_ptr new_bs = chain.head_block_state();
+           ilog("Produced block ${id}... #${n} @ ${t} signed by ${p} [trxs: ${count}, lib: ${lib}, confirmed: ${confs}]",
+                 ("p",new_bs->header.producer)("id",new_bs->id.str().substr(8,16))
+                 ("n",new_bs->block_num)("t",new_bs->header.timestamp)
+                 ("count",new_bs->block->transactions.size())("lib",chain.last_irreversible_block_num())("confs", new_bs->header.confirmed));
+         } else {
+            _unapplied_transactions.add_aborted( chain.abort_block() );
+         }
+      }};
+      blockvault_plug->propose_constructed_block(watermark, lib, block, handler);
```
heifner, line 2109: This should be called `async_propose_constructed_block`.

plugins/producer_plugin/producer_plugin.cpp
heifner: For that matter this method should not return until the handler has been called because it will `schedule_production_loop` which will `start_block`. That can't happen until the current block is commit'ed or aborted.
This implies you need to pass a future into the handler and only continue once the future has been set. That will actually simplify this logic as you can just have the handler set the promise and then leave all the existing logic as is. The future->get() can either throw on unaccepted block or just return on accepted block.

unittests/blockvault_client_plugin_tests.cpp
```diff
@@ -8,10 +8,29 @@
...
+
+   ilog( "produce block (implicitly invoking blockvault logic" );
+   node.produce_block();
+
+   ilog( "verify initial state of tester blockvault state" );
+   // BOOST_REQUIRE(eosio::blockvault_client_plugin::my->bvs.blockvault_initialized == true);
+   // BOOST_REQUIRE(++eosio::blockvault_client_plugin::my->bvs.producer_lib_id == block_id_type{});
+   // BOOST_REQUIRE(++eosio::blockvault_client_plugin::my->bvs.producer_watermark == std::pair<block_num, block_timestamp_type>{});
+   // BOOST_REQUIRE(eosio::blockvault_client_plugin::my->bvs.snapshot_lib_id == block_id_type{});
+   // BOOST_REQUIRE(eosio::blockvault_client_plugin::my->bvs.snapshot_watermark == std::pair<block_num, block_timestamp_type>{});
```
heifner, line 33: I would remove all this for now for this PR. You can add it in a new PR if you get something useful going here.

heifner (changes requested):

plugins/blockvault_client_plugin/CMakeLists.txt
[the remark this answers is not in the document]
johndebord: Removed `signature_provider_plugin`, `eosio_testing`, `http_plugin`, and `chain_plugin`.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Removed comments.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Changed the code to use the `block_state_ptr` returned from evaluating the future `bsf`.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Changed the code to just use the `block_state_ptr` returned from `push_block` in place of `lib` and `block`.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Change made.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Change made.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Removed comment.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Removed comments.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Changed code to use the `block_state_ptr` (pending block) returned from `finalize_block`; changed code to check for the empty optional case; changed code to use the producer of the block instead of the scheduled producer.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Changed code to grab the `dpos_irreversible_blocknum` using the `block_state_ptr` (pending block) returned from `finalize_block`.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Changed the code to just use the `block_state_ptr` returned from `finalize_block` in place of `lib` and `block`.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Changed to use `async_propose_constructed_block` and `async_append_external_block` in the aforementioned `on_incoming_block` code changes.

unittests/blockvault_client_plugin_tests.cpp
johndebord: Removed commented out code.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Implemented changes.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,15 +2069,23 @@ void producer_plugin_impl::produce_block() {
...
-   chain.commit_block();
+   if ( blockvault_plug && blockvault_plug->get() != nullptr ) {
+      std::promise<bool> p;
+      std::future<bool> f = p.get_future();
+      std::optional<producer_watermark> watermark{get_watermark(pending_blk_state->header.producer)};
+      EOS_ASSERT(watermark.has_value(), empty_watermark, "Attempting to use a watermark that does not exist");
+      blockvault_plug->get()->async_propose_constructed_block(watermark.value(), pending_blk_state->dpos_irreversible_blocknum, pending_blk_state->block, [&p](bool b) {
+         p.set_value( b );
+      });
+      EOS_ASSERT( f.get(), blockvault_failure, "Blockvault failure" );
```
heifner, line 2080: This isn't so much blockvault failure as it is invalid block. Seems like blockvault could report an error code indicating why the block was not accepted and that could be logged and included in the exception.
A blockvault failure would be if you f.wait_for() some duration. The timeout would be an indication that the blockvault failed. If blockvault was out of process you really should have a timeout. As is, probably not needed.

libraries/chain/controller.cpp
```diff
@@ -81,7 +81,7 @@ struct pending_state {
       if( std::holds_alternative<building_block>(_block_stage) )
          return std::get<building_block>(_block_stage)._pending_block_header_state;
 
-      return std::get<assembled_block>(_block_stage)._pending_block_header_state;     
+      return std::get<assembled_block>(_block_stage)._pending_block_header_state;
```
heifner, line 84: Try to avoid unneeded diffs. It makes it more difficult to go back and see why a line was changed.

heifner (changes requested):

libraries/chain/controller.cpp
johndebord: Made whitespace changes.

plugins/producer_plugin/producer_plugin.cpp
johndebord: Added a TODO for making a more granular approach to failures on the blockvault server-side.

ndcgundlach (dismissed):

PR #9661, opened 2020-11-12
Fix hard replay through exception when shared_mmeory.bin file has been copied

allenhan2:
## Change Description

Reported by Ian, on Ubuntu + boost 74, seems fc::copy does recursively, meaning when directory was copied, the underlying  shared_memory.bin is copied also. It can not be replicated on Macos (with boost 73).  
The fix is to check whether the file already exists before does copy. 

Huangming suggested to create dir instead of copying dir.  In function `recover_reversible_blocks`, it creates the `reversible_dir` when `new_db_dir` was not provided, however, did not do that when `new_db_dir` was provide. 

This fix has been verified working by Ian, and also works on macos. 

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
I'm not sure I completely follow, but it seems like maybe you're hitting
> **New, breaking change**: copy operation has been extended and reworked to implement behavior specified in C++20 [fs.op.copy]. This includes support for copy_options::recursive, copy_options::copy_symlinks, copy_options::skip_symlinks, copy_options::directories_only, copy_options::create_symlinks and copy_options::create_hard_links options. The operation performs additional checks based on the specified options. Applying copy to a directory with default copy_options will now also copy files residing in that directory (but not nested directories or files in those directories). 

?

Since fc's filesystem is intended to be an abstraction over boost::filesystem, I wonder if it makes any sense to modify fc::copy so that it has consistent behavior across all boost versions? Then again, fc filesystem will hopefully be removed soon so not sure it makes much sense to pour effort in to it.

allenhan2:
@spoonincode appreciate for the inputs, so, it is related to boost 74. Then, i agree with you that the fix should be enforced in fc::copy, if we will continue on using fc.

allenhan2:
how about using `std::filesystem::copy` instead

heifner:
> how about using `std::filesystem::copy` instead

Not available on all our platforms.

allenhan2:
any suggestion on this?

allenhan2:
according to ruling in EPE-487:
Make the changes required in FC to properly handle all versions of boost so far.  This would mean leaving it as is for anything under Boost 1.74 and creating a comparable solution for versions greater than 1.73.  This could be achieved by using Boost’s version macros that it supplies.

close this PR, and fix in FC instead, https://github.com/EOSIO/fc/pull/176

PR #9669, opened 2020-11-13
ship delta changes for issue 9255

venu-block1:
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tbfleming (changes requested):
This doesn't encode variant version changes correctly

tbfleming:
* When creating the deltas, it's storing the newer (version 1) table_delta format, but writing version 0 into the stream
* The ABI file decodes `version=0` mapping to `table_delta_v1` instead of `table_delta_v0`
* The ABI file is missing `row_v0` and `table_delta_v0`

venu-block1:
> * The ABI file is missing `row_v0` and `table_delta_v0`

May be I'm misunderstanding the not backward compatible part of this comment "... its a simple change but it is not strictly speaking backwards compatible" on the slack discussion thread. Now, I have both table_delta_v0 and table_delta_v1 in ABI file.

tbfleming:
Later on in that thread we decided to go with strict backwards compatibility.

venu-block1:
@tbfleming, I made changes in commit f447093 so that both table_delta_v0 and table_delta_v1 are in ABI file. 

Can you please review and confirm if I covered all the changes that are required to be done for this issue? Are there any other code changes I should be doing to advance this draft PR to PR? Thanks!

libraries/rodeos/rodeos.cpp
```diff
@@ -219,14 +219,14 @@ void rodeos_db_snapshot::write_deltas(uint32_t block_num, eosio::opaque<std::vec
    for (uint32_t i = 0; i < num; ++i) {
       ship_protocol::table_delta delta;
       deltas.unpack_next(delta);
-      auto&  delta_v0      = std::get<0>(delta);
+      auto&  delta_v1      = std::get<1>(delta);
```
tbfleming, line 222: rodeos needs to handle both v0 and v1

tbfleming:
This still needs fixing:

- When creating the deltas, it's storing the newer (version 1) table_delta format, but writing version 0 into the stream

libraries/rodeos/rodeos.cpp
venu-block1: Fixed it. Please review.

venu-block1:
> This still needs fixing:
> 
> * When creating the deltas, it's storing the newer (version 1) table_delta format, but writing version 0 into the stream

Please review commit f57011c, and confirm the change fixes this. Thanks!

tbfleming (approved):

tbfleming:
Approved. Switch it from draft mode.

libraries/state_history/include/eosio/state_history/rocksdb_receiver.hpp
```diff
@@ -162,7 +162,7 @@ class rocksdb_receiver_single_entry : public rocksdb_receiver {
       table_ = {};
    }
 
-   void set_delta_present_flag(bool value) {
+   void set_delta_present_flag(uint8_t value) {
```
victorj8, line 165: Now that present is not necessarily a bool, the "flag" in the name suffix doesn't seem quite correct, could you please change it to just set_delta_present ?

victorj8 (commented):
I noticed we currently don't have a unit test, for the case where we are updating a value (1), if you could add some that would be great, specially to test_deltas_contract and/or test_deltas_kv (which is where rocksdb and chainbase would be used for delta calculation)

libraries/state_history/include/eosio/state_history/rocksdb_receiver.hpp
venu-block1: Done

victorj8 (approved):

venu-block1:
> I noticed we currently don't have a unit test, for the case where we are updating a value (1), if you could add some that would be great, specially to test_deltas_contract and/or test_deltas_kv (which is where rocksdb and chainbase would be used for delta calculation)

Changes to exiting test cases to cover present values (0 and 2) are made in b0cf772. Created a ticket to cover present value 1.

PR #9670, opened 2020-11-15
disable building rodeos and eosio.tester

linhuang-blockone:
## Change Description

This PR disables building `rodeos` and `eosio-tester`. The main goal is to remove dependencies on `abieos` so that we can build `eosio` with more than just `clang` and `g++ 8` (currently it fails with other flavors).

After disabling rodeos and eosio-tester, it has been verified that `eosio` can be built with `g++ 9` and `g++ 10`.

The changes in `db_combined.hpp` is for fixing multiple definition link error found by `g++ 10`

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
```diff
@@ -207,7 +207,10 @@ class rocksdb_whole_db_table_collector {
          all_secondary_indices_;
 };
 
-auto process_all = []() { return true; };
+// This used to be "auto process_all = []() { return true; };"
+// g++10 correctly reports a link error of multiple definitions
+// as db_combined.hpp is included by mutiple source files.
```
heifner, line 212: I don't see that this comment adds anything. I would say remove it.

libraries/CMakeLists.txt
```diff
@@ -52,8 +52,6 @@ set_property(GLOBAL PROPERTY CTEST_CUSTOM_TESTS_IGNORE
   yubico_otp wrap_data wrap info import_rsa import_authkey generate_hmac generate_ec\
   attest pbkdf2 parsing ${_CTEST_CUSTOM_TESTS_IGNORE}")
 
-add_subdirectory( rodeos )
```
heifner, line 55: Can we make it a cmake option to enable rodeos and eosio-tester?

heifner (changes requested):

libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
larryk85: I agree, the comment is frivolous.

libraries/CMakeLists.txt
larryk85: Seems kind of heavy weight.  It currently is out of sync and needs to be updated to work correctly, so toggling it with an option is a lie to the end user.

larryk85 (approved):

libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
linhuang-blockone: Removed.

libraries/CMakeLists.txt
heifner: If that broken, then shouldn't we remove the directories out of the branch for now?

heifner (approved):

PR #9671, opened 2020-11-16
Huangminghuang/blockvault client fixes

huangminghuang:
## Change Description
1. Fix the problem where the connection uri is not properly passed to initialize pqxx::connection 
2. Add some more diagnostic logs
3. Change the order to upload big objects  after the table row insertion is successful.  This should avoid the unnecessary big object uploading when the transaction is doomed to fail. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
+++ b/plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
@@ -45,9 +45,7 @@ void blockvault_client_plugin::plugin_initialize(const variables_map& options) {
          }
       }
    }
-   catch (std::exception& ex) {
-      elog("blockvault_client_plugin initialization error: $msg, skipping it.", ("msg", ex.what()));
-   }
+   FC_RETHROW_EXCEPTIONS ( error, "blockvault_client plugin initialization error")
 #endif
 }
 
--- a/plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
+++ b/plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
@@ -9,6 +9,7 @@
 #endif
 
 namespace eosio {
+
 static appbase::abstract_plugin& _blockvault_client_plugin = app().register_plugin<blockvault_client_plugin>();
 
 using vault_impl = eosio::blockvault::block_vault_impl<eosio::blockvault::zlib_compressor>;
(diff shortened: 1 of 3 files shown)


plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
```diff
@@ -20,27 +21,33 @@ class blockvault_client_plugin_impl : public vault_impl {
...
+         if (boost::starts_with(uri, "postgresql://") || boost::starts_with(uri, "postgres://")) {
+            my.reset(new blockvault_client_plugin_impl(std::make_unique<blockvault::postgres_backend>(uri)));
+            ilog("blockvault_client plugin started");
+         } else if (uri.size()) {
+            elog("unknown block-vault-backend option, skipping it");
+         }
       }
    }
+   catch (std::exception& ex) {
+      elog("blockvault_client_plugin initialization error: $msg, skipping it.", ("msg", ex.what()));
```
heifner, line 49: Here and above. If not configured correctly I think you should throw an exception and not allow nodeos to start. This silent failure seems like an issue. If you asked for blockvault you should get it or nodeos should not start.

heifner (changes requested):

plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
```diff
@@ -20,27 +21,31 @@ class blockvault_client_plugin_impl : public vault_impl {
...
-      } else if (url.size()) {
-         elog("unknown block-vault-backend option, skipping it");
+   try {
+      if (options.count("block-vault-backend")) {
+         std::string uri = options["block-vault-backend"].as<std::string>();
+         if (boost::starts_with(uri, "postgresql://") || boost::starts_with(uri, "postgres://")) {
+            my.reset(new blockvault_client_plugin_impl(std::make_unique<blockvault::postgres_backend>(uri)));
+            ilog("blockvault_client plugin started");
+         } else if (uri.size()) {
+            elog("unknown block-vault-backend option, skipping it");
```
heifner, line 44: I think this should throw as well.

heifner (approved):

PR #9674, opened 2020-11-17
Move chainbase calls out of try-CATCH_AND_EXIT_DB_FAILURE block

linhuang-blockone:
## Change Description

In combined_database.cpp, when running with chainbase backing store, chainbase calls are outside the try-CATCH_AND_EXIT_DB_FAILUR,
while when  running with rocksdb backing store, they are inside the try-CATCH_AND_EXIT_DB_FAILURE.

This PR makes exception handling consistent by moving the chainbase calls out of the try-CATCH_AND_EXIT_DB_FAILURE
when running with rocksdb backing store.
## Change Type
**Select ONE**
- [ ] Documentation

- [X] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -9,9 +9,9 @@
 namespace eosio { namespace chain {
    combined_session::combined_session(chainbase::database& cb_database, eosio::session::undo_stack<rocks_db_type>* undo_stack)
        : kv_undo_stack{ undo_stack } {
+      cb_session = std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true));
       try {
         try {
-            cb_session = std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true));
             if (kv_undo_stack) {
               kv_undo_stack->push();
             }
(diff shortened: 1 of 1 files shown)


libraries/chain/combined_database.cpp
```diff
@@ -274,9 +274,9 @@ namespace eosio { namespace chain {
          db.set_revision(revision);
```
heifner, line 274: combined_session needs similar changes.

heifner (changes requested):

libraries/chain/combined_database.cpp
linhuang-blockone: Thanks for pointing this out. Done.

libraries/chain/combined_database.cpp
```diff
@@ -9,9 +9,9 @@
 namespace eosio { namespace chain {
    combined_session::combined_session(chainbase::database& cb_database, eosio::session::undo_stack<rocks_db_type>* undo_stack)
        : kv_undo_stack{ undo_stack } {
+      cb_session = std::make_unique<chainbase::database::session>(cb_database.start_undo_session(true));
```
heifner, line 12: Seems this would be better as a std::optional than a unique_ptr that way you avoid the dynamic memory alloc.

libraries/chain/combined_database.cpp
```diff
@@ -308,9 +296,9 @@ namespace eosio { namespace chain {
          db.commit(revision);
          break;
       case backing_store_type::ROCKSDB:
+         db.commit(revision);
          try {
             try {
-               db.commit(revision);
                kv_undo_stack->commit(revision);
```
heifner, line 302: This approach is fine, but I think removing the break and not duplicating the logic would be clearer. Here and all the other switch statements.

heifner (approved):

libraries/chain/combined_database.cpp
linhuang-blockone: I thought about refactoring that code too, but was afraid of making too big changes to impact existing outstanding branches. Will make the changes.

libraries/chain/combined_database.cpp
linhuang-blockone: Changing cb_session to "std::optional<chainbase::database::session> cb_session" involves multiple changes in combined_database.cpp and combined_database.cpp. Is it OK not to do it at this time?

heifner: Yep

heifner (approved):

libraries/chain/combined_database.cpp
linhuang-blockone: Thanks Kevin for all your help and insist on getting best possible code!

PR #9676, opened 2020-11-17
[blockvault] Fix accessing an empty watermark

johndebord:
The producer watermark is now constructed within the function call to `async_propose_constructed_block`, as opposed to calling `get_watermark`.

<hr>

**Pattern used in `on_block_header` to generate producer watermarks (for posterity):**
```c++
void on_block_header( const block_state_ptr& bsp ) {
   consider_new_watermark( bsp->header.producer, bsp->block_num, bsp->block->timestamp );
}
```
```c++
my->_accepted_block_header_connection.emplace(chain.accepted_block_header.connect( [this]( const auto& bsp ){ my->on_block_header( bsp ); } ));
```
```c++
void commit_block( bool add_to_fork_db ) {
   auto reset_pending_on_exit = fc::make_scoped_exit([this]{
      pending.reset();
   });

   try {
      // ...

      emit( self.accepted_block, bsp );

      // ...
   }

   // push the state for pending.
   pending->push();
}
```

<hr>

So `nodeos_run_test.py` currently has this output:
```bash
...
debug 2020-11-17T08:26:10.000 thread-0  producer_plugin.cpp:1472      start_block          ] Starting block #30 at 2020-11-17T08:26:10.000 producer defproducerb
debug 2020-11-17T08:26:10.001 thread-0  producer_plugin.cpp:1640      remove_expired_trxs  ] Processed 0 expired transactions of the 0 transactions in the unapplied queue, Persistent expired 0, Other expired 0
debug 2020-11-17T08:26:10.001 thread-0  producer_plugin.cpp:1943      schedule_maybe_produ ] Scheduling Block Production on Normal Block #30 for 2020-11-17T08:26:10.400
debug 2020-11-17T08:26:10.401 thread-0  producer_plugin.cpp:1957      operator()           ] Produce block timer for 30 running at 2020-11-17T08:26:10.401
debug 2020-11-17T08:26:10.401 thread-0  producer_plugin.cpp:2023      operator()           ] Signing took 165us
info  2020-11-17T08:26:10.412 thread-1  block_vault_impl.hpp:71       operator()           ] propose_constructed_block({30, 1317833541}, 28) returns false
warn  2020-11-17T08:26:10.412 thread-0  producer_plugin.cpp:2012      maybe_produce_block  ] 3170014 blockvault_failure: Blockvault server-side failure
Blockvault failure
    {}
    thread-0  producer_plugin.cpp:2082 produce_block

[...]

PR #9684, opened 2020-11-18
producer plugin changes

huangminghuang:
reporting block vault rejection instead of error

## Change Description

This PR just incorporates the changes from PR #9676 from John Debord which merges into the wrong branch. 

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -7,6 +7,7 @@
 #include <eosio/chain/thread_utils.hpp>
 #include <eosio/chain/unapplied_transaction_queue.hpp>
 #include <eosio/resource_monitor_plugin/resource_monitor_plugin.hpp>
+#include <eosio/blockvault_client_plugin/blockvault_client_plugin.hpp>
 
 #include <fc/io/json.hpp>
 #include <fc/log/logger_config.hpp>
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2055,15 +2076,24 @@ void producer_plugin_impl::produce_block() {
...
+   if ( blockvault_plug->get() != nullptr ) {
+      std::promise<bool> p;
+      std::future<bool> f = p.get_future();
+      blockvault_plug->get()->async_propose_constructed_block({pending_blk_state->block->block_num(), pending_blk_state->block->timestamp},
+                                                               pending_blk_state->dpos_irreversible_blocknum,
+                                                               pending_blk_state->block, [&p](bool b) {
+         p.set_value( b );
+      });
+      if (!f.get())
+         throw rejected_by_blockvault{};
```
heifner, line 2088: I think it is worth using `EOS_ASSERT` here instead of a raw throw with a struct. Declare a new exception type and use it.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
huangminghuang: The reason I chose not to use EOS_ASSERT and fc derived exception was because I didn't want to log the exception detail.  Logging the exception detail just makes it looks like error as you have mentioned. If we don't log the exception detail, I don't see why not just use a simple struct based exception. This exception should never be propagated outside this file, declare a new exception in `chain/exceptions.hpp` just make you recompile more than 300 files.

heifner: `block_validate_exception` would be appropriate. It is an exceptional condition that is an error. My point is that the error is with the block not with the blockvault as was reported before.

ndcgundlach (approved):

heifner:
Any idea what the unittest failure is about?

ndcgundlach:
I'm looking into the unit test failure.

PR #9692, opened 2020-11-18
RocksDB - Renaming / creation of some parameters and change of default value for create_if_missing

victorj8:
## Change Description

Changes to RocksDB parameter names:
- `rocksdb-threads` changed to `persistent-storage-num-threads`
- `rocksdb-files` changed to `persistent-storage-max-num-files`
- `rocksdb-write-buffer-size-mb`  changed to `persistent-storage-write-buffer-size-mb`

New RocksDB parameter:
- `persistent-storage-bytes-per-sync` (used to control the write rate of flushes and compactions)

Other
- create_if_missing is now set to true.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

We need to change the documentation regarding the new parameter names and the addition of the new parameter.

### DIFF
--- a/libraries/chain/combined_database.cpp
+++ b/libraries/chain/combined_database.cpp
@@ -161,30 +161,30 @@ namespace eosio { namespace chain {
        : backing_store(backing_store_type::ROCKSDB), db(chain_db), kv_database{ [&]() {
             rocksdb::Options options;
 
-            options.create_if_missing = !cfg.read_only; // Creates a database if it is missing
+            options.create_if_missing = true; // Creates a database if it is missing
             options.level_compaction_dynamic_level_bytes = true;
-            options.bytes_per_sync = 1048576; // used to control the write rate of flushes and compactions.
+            options.bytes_per_sync = cfg.persistent_storage_bytes_per_sync; // used to control the write rate of flushes and compactions.
 
             // By default, RocksDB uses only one background thread
             // for flush and compaction.
             // Good value for `total_threads` is the number of cores
-            options.IncreaseParallelism(cfg.rocksdb_threads);
+            options.IncreaseParallelism(cfg.persistent_storage_num_threads);
 
             options.OptimizeLevelStyleCompaction(512ull << 20); // optimizes level style compaction
 
             // Number of open files that can be used by the DB.  
             // Setting it to -1 means files opened are always kept open.
-            options.max_open_files = cfg.rocksdb_max_open_files;
+            options.max_open_files = cfg.persistent_storage_max_num_files;
 
             // Use this option to increase the number of threads
             // used to open the files.
-            options.max_file_opening_threads = cfg.rocksdb_threads; // Default should be the # of Cores
+            options.max_file_opening_threads = cfg.persistent_storage_num_threads; // Default should be the # of Cores
 
             // Write Buffer Size - Sets the size of a single
             // memtable. Once memtable exceeds this size, it is
             // marked immutable and a new one is created.
             // Default should be 128MB
-            options.write_buffer_size = cfg.rocksdb_write_buffer_size;
+            options.write_buffer_size = cfg.persistent_storage_write_buffer_size;
             options.max_write_buffer_number = 10; // maximum number of memtables, both active and immutable
             options.min_write_buffer_number_to_merge = 2; // minimum number of memtables to be merged before flushing to storage
 
@@ -193,7 +193,7 @@ namespace eosio { namespace chain {
 
             // Size of L0 = write_buffer_size * min_write_buffer_number_to_merge * level0_file_num_compaction_trigger
             // For optimal performance make this equal to L0
-            options.max_bytes_for_level_base = cfg.rocksdb_write_buffer_size * options.min_write_buffer_number_to_merge * options.level0_file_num_compaction_trigger; 
+            options.max_bytes_for_level_base = cfg.persistent_storage_write_buffer_size * options.min_write_buffer_number_to_merge * options.level0_file_num_compaction_trigger;
 
             // Files in level 1 will have target_file_size_base
             // bytes. It’s recommended setting target_file_size_base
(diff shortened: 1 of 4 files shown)


libraries/chain/combined_database.cpp
```diff
@@ -161,30 +161,30 @@ namespace eosio { namespace chain {
...
 
             // Number of open files that can be used by the DB.  
             // Setting it to -1 means files opened are always kept open.
-            options.max_open_files = cfg.rocksdb_max_open_files;
+            options.max_open_files = cfg.persistent_storage_max_num_files;
 
             // Use this option to increase the number of threads
             // used to open the files.
-            options.max_file_opening_threads = cfg.rocksdb_threads; // Default should be the # of Cores
+            options.max_file_opening_threads = cfg.persistent_storage_num_threads; // Default should be the # of Cores
```
gssagoo, line 181: @victorj8 - we should remove the comment on this line as it applied to the earlier setting and is confusing.

libraries/chain/combined_database.cpp
```diff
@@ -161,30 +161,30 @@ namespace eosio { namespace chain {
...
 
-            options.create_if_missing = !cfg.read_only; // Creates a database if it is missing
+            options.create_if_missing = true; // Creates a database if it is missing
             options.level_compaction_dynamic_level_bytes = true;
-            options.bytes_per_sync = 1048576; // used to control the write rate of flushes and compactions.
+            options.bytes_per_sync = cfg.persistent_storage_bytes_per_sync; // used to control the write rate of flushes and compactions.
 
             // By default, RocksDB uses only one background thread
             // for flush and compaction.
             // Good value for `total_threads` is the number of cores
```
gssagoo, line 170: @victorj8 - we should remove the 3 lines of comments and replace with: 
// Number of threads used for flush and compaction.

libraries/chain/combined_database.cpp
```diff
@@ -204,7 +204,7 @@ namespace eosio { namespace chain {
             // that will concurrently perform a compaction job by
             // breaking it into multiple,
             // smaller ones that are run simultaneously.
-            options.max_subcompactions = cfg.rocksdb_threads;	// Default should be the # of CPUs
+            options.max_subcompactions = cfg.persistent_storage_num_threads;	// Default should be the # of CPUs
```
gssagoo, line 207: @victorj8 - we should remove the comment on this line as it applied to the earlier setting and is confusing.

gssagoo (changes requested):
I have left comments in line to remove and update some of the comments accompanying the changes we have made.

libraries/chain/combined_database.cpp
victorj8: Done

libraries/chain/combined_database.cpp
victorj8: Done

libraries/chain/combined_database.cpp
victorj8: Done

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -306,12 +304,14 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
...
+         ("persistent-storage-num-threads", bpo::value<uint16_t>()->default_value(default_persistent_storage_num_threads),
           "Number of rocksdb threads for flush and compaction")
-         ("rocksdb-files", bpo::value<int>()->default_value(config::default_rocksdb_max_open_files),
+         ("persistent-storage-max-num-files", bpo::value<int>()->default_value(config::default_persistent_storage_max_num_files),
           "Max number of rocksdb files to keep open. -1 = unlimited.")
-         ("rocksdb-write-buffer-size-mb", bpo::value<uint64_t>()->default_value(config::default_rocksdb_write_buffer_size / (1024  * 1024)),
+         ("persistent-storage-write-buffer-size-mb", bpo::value<uint64_t>()->default_value(config::default_persistent_storage_write_buffer_size / (1024  * 1024)),
           "Size of a single rocksdb memtable (in MiB)")
+         ("persistent-storage-bytes-per-sync", bpo::value<uint64_t>()->default_value(config::default_persistent_storage_bytes_per_sync),
+          "Rocksdb write rate of flushes and compactions.")
```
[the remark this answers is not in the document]
victorj8: I was told that this value may be getting values such as 512K, that's why i didn't use the MiB here.

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -88,8 +88,10 @@ const static uint16_t   default_controller_thread_pool_size          = 2;
 const static uint32_t   default_max_variable_signature_length        = 16384u;
 const static uint32_t   default_max_nonprivileged_inline_action_size = 4 * 1024; // 4 KB
 const static uint32_t   default_max_action_return_value_size         = 256;
-const static int        default_rocksdb_max_open_files               = -1;
-const static uint64_t   default_rocksdb_write_buffer_size            = 128 * 1024 * 1024;
+const static uint16_t   default_persistent_storage_num_threads       = 1;
+const static int        default_persistent_storage_max_num_files     = -1;
+const static uint64_t   default_persistent_storage_write_buffer_size = 128 * 1024 * 1024;
+const static uint64_t   default_persistent_storage_bytes_per_sync    = 1048576;
```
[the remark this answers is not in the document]
victorj8: Done

linhuang-blockone (approved):

libraries/chain/combined_database.cpp
```diff
@@ -161,30 +161,28 @@ namespace eosio { namespace chain {
...
 
             // Use this option to increase the number of threads
             // used to open the files.
-            options.max_file_opening_threads = cfg.rocksdb_threads; // Default should be the # of Cores
+            options.max_file_opening_threads = cfg.persistent_storage_num_threads;
 
             // Write Buffer Size - Sets the size of a single
             // memtable. Once memtable exceeds this size, it is
             // marked immutable and a new one is created.
             // Default should be 128MB
```
gssagoo, line 184: @victorj8 Apologies - i missed this yesterday, but can you remove the comment '// Default should be 128MB'. thanks

gssagoo (commented):
All looks good besides that comment cleanup that i commented on which is not a show stopper and can be addressed afterwards.

PR #9694, opened 2020-11-18
net_plugin monitor heartbeat of peers

kimjh2005:
## Change Description

Implemented heartbeat monitoring of connections.
- Added config option --p2p-keepalive-interval-ms N  
- Updated time stamp of received messages
- ticker timer function checks each connections latest time stamp and current time stamp. 
As ticker() is executed on keepalive interval, heartbeat time out may not be the exact 2 * p2p-keepalive-interval-ms.
- For connection to peer-address, it will try to reconnect.  For accepted connection,  the connection is disconnected and destroyed.
- For connections from old network_version, the heartbeat timeout will not be enabled.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

>nodeos --help 
  ...                                     
                                        
  --p2p-keepalive-interval-ms arg (=32000)     heartbeat keepalive message interval in
                                        miliseconds
...

1. Every p2p-keepalive-interval-ms milliseconds a time_message is published to all the connections. 
2. If any messages not received over 2 * p2p-keepalive-interval-ms milliseoncds, a heartbeat timeout occurs and the  connection is closed.   If the connection is to peer, it will be reconnected when the peer restarts.  If it is an accepted connection, it will not be reconnected and the connection is closed and removed. 
3. Heartbeat timeout is disabled on connection from older version.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3576,7 +3576,7 @@ namespace eosio {
            "   _port  \tremote port number of peer\n\n"
            "   _lip   \tlocal IP address connected to peer\n\n"
            "   _lport \tlocal port number connected to peer\n\n")
-         ( "p2p-keepalive-interval", bpo::value<int>()->default_value(def_keepalive_interval), "peer heartbeat keepalive message interval in miliseconds")
+         ( "p2p-keepalive-interval-ms", bpo::value<int>()->default_value(def_keepalive_interval), "peer heartbeat keepalive message interval in milliseconds")
 
         ;
    }
@@ -3602,12 +3602,12 @@ namespace eosio {
          my->p2p_accept_transactions = options.at( "p2p-accept-transactions" ).as<bool>();
 
          my->use_socket_read_watermark = options.at( "use-socket-read-watermark" ).as<bool>();
-         my->keepalive_interval = std::chrono::milliseconds( options.at( "p2p-keepalive-interval" ).as<int>() );
+         my->keepalive_interval = std::chrono::milliseconds( options.at( "p2p-keepalive-interval-ms" ).as<int>() );
          EOS_ASSERT( my->keepalive_interval.count() > 0, chain::plugin_config_exception,
-                     "p2p-keepalive_interval must be greater than 0" );
+                     "p2p-keepalive_interval-ms must be greater than 0" );
 
-         if( options.count( "p2p-keepalive-interval" )) {
-            my->heartbeat_timeout = std::chrono::milliseconds( options.at( "p2p-keepalive-interval" ).as<int>() * 2 );
+         if( options.count( "p2p-keepalive-interval-ms" )) {
+            my->heartbeat_timeout = std::chrono::milliseconds( options.at( "p2p-keepalive-interval-ms" ).as<int>() * 2 );
          }
 
          if( options.count( "p2p-listen-endpoint" ) && options.at("p2p-listen-endpoint").as<string>().length()) {
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -233,8 +233,8 @@ namespace eosio {
       boost::asio::steady_timer::duration   connector_period{0};
       boost::asio::steady_timer::duration   txn_exp_period{0};
       boost::asio::steady_timer::duration   resp_expected_period{0};
-      boost::asio::steady_timer::duration   keepalive_interval{std::chrono::seconds{32}};
-      std::chrono::seconds                  heartbeat_timeout{std::chrono::seconds{32 * 2}};
+      std::chrono::milliseconds             keepalive_interval{std::chrono::milliseconds{32 * 1000}};
+      std::chrono::milliseconds             heartbeat_timeout{keepalive_interval * 2};
 
       int                                   max_cleanup_time_ms = 0;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1081,6 +1096,25 @@ namespace eosio {
       });
    }
 
+   void connection::check_heartbeat( const tstamp &current_time )
+   {
+      if( protocol_version >= heartbeat_interval ) {
+         if( latest_msg_time > 0 &&  current_time > latest_msg_time + heartbeat_timeout ) {
+            if( !peer_addr.empty() ) {
```
heifner, line 1103: use `peer_address()` instead of variable directly.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1081,6 +1096,25 @@ namespace eosio {
...
+               fc_wlog(logger, "heartbeat timed out for peer address ${adr}", ("adr", peer_addr));
+               no_retry = benign_other;
+            } else {
+               fc_wlog(logger, "heartbeat timed out from ${p} ${ag}", ("p", last_handshake_recv.p2p_address)("ag", last_handshake_recv.agent));
+               no_retry = fatal_other;
+            }
+
+            flush_queues();
```
heifner, line 1111: `close()` calls `flush_queues()` don't call here.
Also I think this would be a situation where you would want immediate reconnect. Pass `true` to close.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1081,6 +1096,25 @@ namespace eosio {
...
 
+   void connection::check_heartbeat( const tstamp &current_time )
+   {
+      if( protocol_version >= heartbeat_interval ) {
+         if( latest_msg_time > 0 &&  current_time > latest_msg_time + heartbeat_timeout ) {
+            if( !peer_addr.empty() ) {
+               fc_wlog(logger, "heartbeat timed out for peer address ${adr}", ("adr", peer_addr));
+               no_retry = benign_other;
+            } else {
+               fc_wlog(logger, "heartbeat timed out from ${p} ${ag}", ("p", last_handshake_recv.p2p_address)("ag", last_handshake_recv.agent));
```
heifner, line 1107: `last_handshake_recv` should not be accessed without holding the `conn_mtx`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1081,6 +1096,25 @@ namespace eosio {
...
+               no_retry = fatal_other;
```
heifner, line 1108: I don't see why you are setting `no_retry` to `fatal_other`. Why set it at all? For that matter why set it to `benign_other` above? That is set when sending a go_away message which we are not doing here.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1081,6 +1096,25 @@ namespace eosio {
...
+            }
+
+            flush_queues();
+            close();
+         }
+      }
+      send_time();
```
heifner, line 1115: Should not `send_time()` if you just closed the connection above.

heifner (changes requested):
You could manually test this by pausing a connected node and seeing if it disconnected and tried to re-connect.

spoonincode:
Were TCP keepalives considered? I realize this technically detects more failures (given it's a layer 7 check not a layer 4). But unless detecting stuck nodeos' is a requirement, TCP keepalives seem considerably less complex to implement and still detect dead network connections.

heifner:
> Were TCP keepalives considered? I realize this technically detects more failures (given it's a layer 7 check not a layer 4). But unless detecting stuck nodeos' is a requirement, TCP keepalives seem considerably less complex to implement and still detect dead network connections.

Yes. https://github.com/EOSIO/eos/tree/net-plugin-keepalive-2.0 but that didn't fix EOSUSA issue.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3532,6 +3576,8 @@ namespace eosio {
            "   _port  \tremote port number of peer\n\n"
            "   _lip   \tlocal IP address connected to peer\n\n"
            "   _lport \tlocal port number connected to peer\n\n")
+         ( "keepalive-interval", bpo::value<int>()->default_value(def_keepalive_interval), "heartbeat keepalive message interval in miliseconds")
```
heifner, line 3579: This should be called `p2p-keepalive-interval-ms`
Remember to update the PR description.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3556,6 +3602,11 @@ namespace eosio {
          my->p2p_accept_transactions = options.at( "p2p-accept-transactions" ).as<bool>();
 
          my->use_socket_read_watermark = options.at( "use-socket-read-watermark" ).as<bool>();
+         my->keepalive_interval = std::chrono::milliseconds( options.at( "keepalive-interval" ).as<int>() );
```
heifner, line 3605: Should EOS_ASSERT this is > 0

heifner (changes requested):

heifner (approved):

PR #9697, opened 2020-11-19
[blockvault] Fix CICD Test Failure

johndebord:
~~Installing the dependencies in `libpq` and `libpqxx` in the `.cicd/platforms/unpinned/macos-10.15-unpinned.sh` script is causing the test `snapshot_unit_test_eos-vm`. This change addresses that issue.~~
image


huangminghuang (dismissed):

heifner:
The problem was a bad merge with develop

PR #9698, opened 2020-11-19
[blockvault] Add Guide For Building And Testing The `blockvault_client_plugin`

johndebord:


plugins/blockvault_client_plugin/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+Build the `blockvault` branch:
+```bash
+~ $ git clone --recursive --branch blockvault git@github.com:EOSIO/eos.git blockvault
+~ $ cd blockvault
+~/blockvault $ mkdir build
+~/blockvault $ cd build
+
+# Caveat Alert! macOS will default to using its system version of `libpq`. To use the `libpq` installed by brew set the environment variable like so:
+~/blockvault $ PKG_CONFIG_PATH=/usr/local/opt/libpq/lib/pkgconfig
```
spoonincode, line 27: doesn't the cmake file already do this?

PR #9705, opened 2020-11-20
Merge blockvault into develop

huangminghuang:
## Change Description

This PR merges the blockvault branch into develop

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

plugins/blockvault_client_plugin/README.md provides some basic documentation for the overview and configurations.


heifner:
Since this will be the PR referenced in future release notes. Can the description be updated with an overview of blockvault and with command line options.

plugins/blockvault_client_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+
+
+if (APPLE AND EXISTS "/usr/local/opt/libpq/lib/pkgconfig")
+   # to find libpq installed by homebrew, it is required to set the environment variable PKG_CONFIG_PATH
+   if (DEFINED ENV{PKG_CONFIG_PATH})
+      set(ENV{PKG_CONFIG_PATH} "/usr/local/opt/libpq/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
+   else()
+      set(ENV{PKG_CONFIG_PATH} "/usr/local/opt/libpq/lib/pkgconfig")
+   endif()
+endif(APPLE AND NOT DEFINED ENV{PKG_CONFIG_PATH})
```
spoonincode, line 16: this mismatch of if()/endif() causes a warning. No need to do this, just use `endif()`

spoonincode (changes requested):
Need to include libpqxx license file.

Was making libpqxx a submodule considered/attempted? It would make that easier, and at first look libpqxx looks to be somewhat easy to integrate as a submodule.

spoonincode:
The .deb files are no longer correct when libpq is linked in dynamically (which it is for at least for the 18.04 build). Since libpq is optional, when it is dynamically linked will need to wire through to the deb/rpm builder scripts that it is a dep.

spoonincode:
There are many places in the docs, help text, etc that refer to "PostgresQL", but the correct spelling is "PostgreSQL"

plugins/blockvault_client_plugin/include/eosio/blockvault_client_plugin/blockvault_client_plugin.hpp
```diff
@@ -0,0 +1,84 @@
+// CORE TODO
```
spoonincode, line 1: Should this comment section remain?

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -224,10 +225,18 @@ class chain_plugin_impl {
...
+           chain->startup(shutdown, check_shutdown, *genesis);
+       }else {
+           chain->startup(shutdown, check_shutdown);
+       }
+   }
 };
 
-chain_plugin::chain_plugin()
-:my(new chain_plugin_impl()) {
+    chain_plugin::chain_plugin()
```
spoonincode, line 238: Not sure why indentation changed here, but eosio code indentation is 3 spaces

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
```diff
@@ -0,0 +1,85 @@
+#pragma once
```
spoonincode, line 1: file indentation is 4 spaces

programs/nodeos/main.cpp
```diff
@@ -119,7 +120,7 @@ int main(int argc, char** argv)
          .default_unix_socket_path = "",
          .default_http_port = 8888
       });
-      if(!app().initialize<chain_plugin, net_plugin, producer_plugin, resource_monitor_plugin>(argc, argv)) {
+      if(!app().initialize<blockvault_client_plugin, chain_plugin, net_plugin, producer_plugin, resource_monitor_plugin>(argc, argv)) {
```
heifner, line 123: `blockvault_client_plugin` should be removed. It is an optional plugin.

programs/nodeos/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ target_link_libraries( ${NODE_EXECUTABLE_NAME}
         PRIVATE -Wl,${whole_archive_flag} test_control_plugin        -Wl,${no_whole_archive_flag}
         PRIVATE -Wl,${whole_archive_flag} test_control_api_plugin    -Wl,${no_whole_archive_flag}
         PRIVATE -Wl,${build_id_flag}
-        PRIVATE chain_plugin http_plugin producer_plugin http_client_plugin
+        PRIVATE chain_plugin http_plugin producer_plugin http_client_plugin blockvault_client_plugin
```
heifner, line 61: `blockvault_client_plugin` will need wrapped in the `whole_archive_flag` once removed from main.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -674,6 +679,8 @@ if( options.count(op_name) ) { \
 
 void producer_plugin::plugin_initialize(const boost::program_options::variables_map& options)
 { try {
+   my->blockvault_plug = app().find_plugin<blockvault_client_plugin>();
+   EOS_ASSERT( my->blockvault_plug, plugin_config_exception, "blockvault_client_plugin not found" );
```
heifner, line 683: Once blockvault_plugin removed from main this will need to be removed and check for null at use.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1318,15 +1327,19 @@ void chain_plugin::plugin_startup()
    try {
       auto shutdown = [](){ return app().quit(); };
       auto check_shutdown = [](){ return app().is_quiting(); };
-      if (my->snapshot_path) {
+      auto bvc_plug = app().find_plugin<blockvault_client_plugin>();
+      EOS_ASSERT( bvc_plug, plugin_config_exception, "blockvault_client_plugin not found" );
```
spoonincode, line 1331: fwiw, I think you can `get_plugin("blockvault_client_plugin")` which throws if plugin not found. will save a line or two

plugins/chain_plugin/test/test_blockvault_sync_strategy.cpp
```diff
@@ -0,0 +1,153 @@
+#define BOOST_TEST_MODULE blockvault_sync_strategy
```
spoonincode, line 1: this file has 4 spaces too

programs/nodeos/main.cpp
huangminghuang: Actually, blockvault_client_plugin is mandatory. It uses wether blockvault_client_plugin::get() returns nullptr or not to check if the plugin is functional. The reason for the design is to make sure blockvault_client_plugin is initialized before chain plugin. There no control of plugin initialization order of optional plugin AFAIK.

heifner: Really, I thought you could specify dependency even when optional. For example chain_api_plugin is dependent on chain_plugin and it is optional. I believe that is honored on init/startup.

spoonincode, line 123: Not sure you actually need this since chain_plugin already REQUIRES() blockvault_client_plugin

heifner: See application.hpp initialize and startup

heifner: I don't think chain_plugin should require blockvault_client_plugin unless I'm missing something.

plugins/blockvault_client_plugin/tests/test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+
+CONTAINER_ID=`docker run  -p 5432:5432 -e POSTGRES_PASSWORD=password -d postgres`
```
spoonincode, line 4: This exposes 5432 to the world. Unless that's needed it's probably better practice to do `-p  127.0.0.1:5432:5432` instead

huangminghuang:
> Need to include libpqxx license file.
> 
> Was making libpqxx a submodule considered/attempted? It would make that easier, and at first look libpqxx looks to be somewhat easy to integrate as a submodule.

Bucky made the call not to use submodule.

programs/nodeos/CMakeLists.txt
spoonincode: I tried locally after removing it. Didn't need whole_archive. I think maybe it being REQUIRE() from chain_plugin already gets it picked up. Not that it would really hurt anything being whole_archive'd anways

spoonincode:
The build scripts don't install the required dependencies, libpq & libpqxx, on all platforms.

More troubling is that the pinned build script is broken: it will use the system libpqxx during a pinned build which is invalid. Pinned builds cannot use any system provided c++ libraries. In this case, this is easily fixed by making libpqxx a submodule.

spoonincode:
something needs to be done with exceptions at init, as they seem to just be getting swallowed somewhere. In this case my postgresql wasn't running

```
$ programs/nodeos/nodeos --block-vault-backend postgresql://postgres:password@localhost/ -e -peosio
info  2020-11-21T21:51:50.699 nodeos    blockvault_client_plug:35     plugin_initialize    ] initializing blockvault_client plugin
Failed to initialize
$
```

spoonincode:
If the connection to the postgres server fails, it doesn't seem to be retried. Have to restart nodeos to get it to reconnect.

plugins/blockvault_client_plugin/postgres_backend.cpp
```diff
@@ -0,0 +1,196 @@
...
+#include <limits.h>
+#include <unistd.h>
+
+namespace eosio {
+namespace blockvault {
+
+postgres_backend::postgres_backend(const std::string& options)
+    : conn(options) {
+   pqxx::work w(conn);
+   w.exec("CREATE TABLE IF NOT EXISTS BlockData (watermark_bn bigint, watermark_ts bigint, lib bigint, "
```
b1bart, line 14: I'm surprised not to see indices over some of the selected columns like `watermark_(bn|ts)` `lib` and `block_id`, was that unnecessary or reserved for future perf analysis?

b1bart (changes requested):

plugins/blockvault_client_plugin/postgres_backend.cpp
```diff
@@ -0,0 +1,196 @@
...
+          "block_id bytea, previous_block_id bytea, block oid, block_size bigint);"
+          "CREATE TABLE IF NOT EXISTS SnapshotData (watermark_bn bigint, watermark_ts bigint, snapshot oid);");
+   w.commit();
+
+   conn.prepare(
+       "insert_constructed_block",
+       "INSERT INTO BlockData (watermark_bn, watermark_ts, lib, block_id, previous_block_id, block, block_size) "
+       "SELECT $1, $2, $3, $4, $5, $6, $7  WHERE NOT "
+       "EXISTS (SELECT * FROM BlockData WHERE (watermark_bn >= $1) OR (watermark_ts > $2) OR (lib > $3) OR (block_id "
```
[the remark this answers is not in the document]
huangminghuang: > if my read is correct, I think this should be `(watermark_ts >= $2)`. It is considered "double production" to produce two blocks at the same height OR the same timestamp. This statement seems to require that there is no known block with a higher or equal blocknumber OR higher timestamp which leaves open a higher blocknumber with the same timstamp.

The timestamp only changes when producer changes, i.e. it is only incremented every 12 blocks in the ideal cases. If you have the (watermark_ts >= $2), that means the producer can only produce one block every production round.

plugins/blockvault_client_plugin/postgres_backend.cpp
huangminghuang: This is what I got from https://www.w3schools.com/sql/sql_create_index.asp
```
Note: Updating a table with indexes takes more time than updating a table without (because the indexes also need an update). So, only create indexes on columns that will be frequently searched against.
```

Base on the above statement, I have double if it's worthwhile to create indexes.

.cicd/platforms/pinned/centos-7.7-pinned.dockerfile
```diff
@@ -46,7 +46,18 @@ RUN curl -LO https://dl.bintray.com/boostorg/release/1.72.0/source/boost_1_72_0.
     ./bootstrap.sh --with-toolset=clang --prefix=/usr/local && \
     ./b2 toolset=clang cxxflags='-stdlib=libc++ -D__STRICT_ANSI__ -nostdinc++ -I/usr/local/include/c++/v1 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fpie' linkflags='-stdlib=libc++ -pie' link=static threading=multi --with-iostreams --with-date_time --with-filesystem --with-system --with-program_options --with-chrono --with-test -q -j$(nproc) install && \
     cd / && \
-    rm -rf boost_1_72_0.tar.bz2 /boost_1_72_0
+    rm -rf boost_1_72_0.tar.bz2 /boost_1_72_0whi
```
johndebord, line 49: Is this a typo?

huangminghuang: yes, wIll be fixed in PR #9711

plugins/blockvault_client_plugin/postgres_backend.cpp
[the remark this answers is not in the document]
b1bart: The timestamp should change with every block.  If it isn't that is a bug.  That time should be the blocks' timestamps.  

### recording watermarks
this code is fired from the `accepted_block_header` signal from controller:

https://github.com/EOSIO/eos/blob/efadbb1/plugins/producer_plugin/producer_plugin.cpp#L229-L231

That signal is fired from `push_block` for every block that is accepted to fork-db (has proper headers) and from `commit_block` for locally constructed blocks.  That should result in it being updated every block. 


### use
This code is what prevents double signing on a node that has complete data:

https://github.com/EOSIO/eos/blob/efadbb1/plugins/producer_plugin/producer_plugin.cpp#L1374-L1389

It is a little convoluted because it is checks `>` the block num of the *current head block* (which will be our previous block making our block num 1 higher) AND it checks that our candidate `block_timestamp` is `>=` the watermarks timestamp.

plugins/blockvault_client_plugin/postgres_backend.cpp
b1bart: If we are not seeing performance issues at this time, then we can leave it as-is.  It is true that updates/inserts are more costly with indices *however* our updates have dependent queries associated with them to enforce our guarantees.

We should profile this with a substantially sized database and see if the trade of certain indices improves update time by reducing those dependent read-times more than it increases the time to write to indices.  

There is also another option where we do some fancy materialized view of the current highest watermarks BUT we should not engage in these optimizations until we have baseline performance numbers and a plan to iterate.

plugins/blockvault_client_plugin/postgres_backend.cpp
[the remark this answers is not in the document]
huangminghuang: fixed in commit 900909f2acc93a6b7b7ccfbeb5d35b7eab628b82

plugins/blockvault_client_plugin/postgres_backend.cpp
b1bart: resolving this for now as it was mostly a thing to pay attention to in our next iteration.

b1bart (approved):

b1bart:
@huangminghuang I've completed a logic review and turned it to approve.  When you and @johndebord have the build/ci issues resolved it can go in as far as I am concerned.

plugins/chain_plugin/chain_plugin.cpp
ndcgundlach: blockvault_client_plugin is no longer required, so this is no longer relevant

tests/CMakeLists.txt
```diff
@@ -105,6 +106,7 @@ add_test(NAME keosd_auto_launch_test COMMAND tests/keosd_auto_launch_test.py WOR
 set_property(TEST keosd_auto_launch_test PROPERTY LABELS nonparallelizable_tests)
 add_test(NAME db_modes_test COMMAND tests/db_modes_test.sh WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 set_tests_properties(db_modes_test PROPERTIES COST 6000)
+set_property(TEST db_modes_test PROPERTY LABELS nonparallelizable_tests)
```
spoonincode, line 109: Why is this needed? That test is designed to be parallelizable.

huangminghuang: It listens on some default TCP ports which cannot be shared with others.

spoonincode: What ports? It's launched with `--http-server-address '' --p2p-listen-endpoint ''` which has historically disabled listening on all ports.

.cicd/platforms/pinned/macos-10.14-pinned.sh
```diff
@@ -3,6 +3,8 @@ set -eo pipefail
 VERSION=1
 brew update
 brew install git cmake python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl@1.1 jq || :
+curl -LO  https://raw.githubusercontent.com/Homebrew/homebrew-core/106b4b8a421dda33c79a4018c3c3816234076331/Formula/libpqxx.rb
```
spoonincode, line 6: We've done stuff like this before and were burned by it. Why are we doing these sort of hacks instead of just using it as a submodule like it appears to be designed for? @larryk85 why are you against this being a submodule?

larryk85: I'm not against submodules as a whole. I'm against this particular dependency being a submodule for a few reasons.

We need clearly defined rules as to what is or is not a submodule in our codebase.  It is clear with some of the other recent additions as submodules that without solid rules this can lead to less than ideal situations (we have to install components or the entirety of the dependencies with our own installation, possibility of building certain things inefficiently for certain targets, etc).  

This is also in effect to aid in consistency with our dependency management and overall project management.  There are no questions as to what to do when or why, if we modify it or own it then it is a submodule.  Other than that, it lives outside of that space and we have to treat it like any other external dependency.  Also if that means that we need to have custom brew/repo mirrors for our needed versions to fill in certain targets, then that is a sufficient answer as well.  Once again we don't modify these, as such the need to update them is going to be low.

We don't own libpqxx and we do not modify libpqxx, therefore it makes little sense at this point to have it in our codebase because it is probably the easier thing to do.  I fully understand your points and if in the future we need a modified variant of the lib, then it can live in our codebase and install 'along-side' our main code install.

spoonincode: We don't own rocksdb, we don't own yubihsm, we don't own rapidjson, and we don't own amqp-cpp either. Nor did we modify those. Not submoduling looks contrarian not consistent.

It's not just the "easier thing to do", being a submodule has many obvious engineering benefits to us and usability benefits to users:
* Can universally, consistently pin at a known good version; something this PR does multiple ways contrary to your "aid in consistency with our dependency management" goal
* Seamlessly works for pinned builds in all environments; something this PR doesn't handle in all cases AFAIK
* Easier to install the required license file; something this PR doesn't handle
* Doesn't require users to chase down a dependency that isn't provided by common distros (we should _not_ start leaning on the build scripts to do that -- those need to go away. If we need a build script we've made it too hard. Good OSS should not require some obfuscated build script)

Instead we're spending significant engineering time crafting fragile unique solutions to these issues scattered throughout the CI and build scripts in an inconsistent way.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1131,6 +1143,10 @@ void producer_plugin::create_snapshot(producer_plugin::next_function<producer_pl
                ("message", ec.message()));
 
          my->_pending_snapshot_index.emplace(head_id, next, pending_path.generic_string(), snapshot_path.generic_string());
+
+         if ( my->blockvault != nullptr ) {
+            my->blockvault->propose_snapshot( blockvault::watermark_t{head_block_num, head_block_time}, pending_path.generic_string().c_str() );
```
swatanabe-b1, line 1148: What if this block gets forked out?

heifner: Does look like this should be called from `pending_snapshot::finalize`

huangminghuang:
> Need to include libpqxx license file.

addressed in commit 020c2867d23b73224c3d5b841ed2c13062add007

tests/CMakeLists.txt
huangminghuang: revert the change in commit b068313ccfa8b3be9d9d70f861a7dd2eccc69b46

PR #9706, opened 2020-11-23
Default to returning WASM as returning WAST has been deprecated.

deckb:
## Change Description

As there is no more support for `get_code` to return WAST the default to return WASM should be true.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ X] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


b1bart (approved):
Seeing as though the OpenAPI docs for this endpoint was already suggesting this defaults to `true` it would be nice if it did 👍

PR #9712, opened 2020-11-24
Snapshot memory exhaustion

brianjohnson5972:
# Change Description
Code previously was batching up all key-value pairs from db intrinsic tables during initialization from a snapshot. Changed to track how much memory was being stored and writing out the batch when it passes a newly added configuration threshold.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

New configuration parameter:
"persistent-storage-mbytes-snapshot-batch"
          Rocksdb threshold, in MB, to determine how  much memory in Key-Value pairs to batch before writing data read from a snapshot. Depending on the resources available on your machine, increasing this number may result in faster loading of a snapshot.

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -312,6 +312,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "Size of a single rocksdb memtable (in MiB)")
          ("persistent-storage-bytes-per-sync", bpo::value<uint64_t>()->default_value(config::default_persistent_storage_bytes_per_sync),
           "Rocksdb write rate of flushes and compactions.")
+         ("persistent-storage-mbytes-snapshot-batch", bpo::value<uint32_t>()->default_value(config::default_persistent_storage_mbytes_batch),
+          "Rocksdb batch size threshold before writing read in snapshot data to database.")
 
          ("reversible-blocks-db-size-mb", bpo::value<uint64_t>()->default_value(config::default_reversible_cache_size / (1024  * 1024)), "Maximum size (in MiB) of the reversible blocks database")
          ("reversible-blocks-db-guard-size-mb", bpo::value<uint64_t>()->default_value(config::default_reversible_guard_size / (1024  * 1024)), "Safely shut down node when free space remaining in the reverseible blocks database drops below this size (in MiB).")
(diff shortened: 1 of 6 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -312,6 +312,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "Size of a single rocksdb memtable (in MiB)")
          ("persistent-storage-bytes-per-sync", bpo::value<uint64_t>()->default_value(config::default_persistent_storage_bytes_per_sync),
           "Rocksdb write rate of flushes and compactions.")
+         ("persistent-storage-mbytes-snapshot-batch", bpo::value<uint32_t>()->default_value(config::default_persistent_storage_mbytes_batch),
+          "Rocksdb batch size threshold before writing read in snapshot data to database.")
```
heifner, line 317: Do we really need a knob for something only used when reading a snapshot?

brianjohnson5972: The intent was that having this means that if someone is using rocksdb because they have limited resources, that they won't be forced to have as high of memory available the we default to, but that if someone has plenty of memory, that they can jack it way up and start faster.
@tbfleming @larryk85 @b1bart

larryk85: I don't see any harm in having the knob.  And it would allow for operators with more robust machines to set the value higher and read the snapshots in as fast as possible (bringing a reserve node up to parity or a main node back after failure).

spoonincode: How much faster is it when turning the knob up?

tbfleming: I prefer to have a knob; fill-rocksdb had perf gains with high batch sizes (somewhere between 100-150 blocks of full-history deltas on mainnet). I no longer have the numbers, but the difference was large.

heifner: I'll defer. I just believe we need to question every new knob as each and every one means additional testing and potential additional error reports when used incorrectly.

larryk85 (approved):

plugins/chain_plugin/chain_plugin.cpp
spoonincode: My measurements with a mainnet snapshot:
50MB: 63s
200MB: 63s
1024MB: 61s

PR #9716, opened 2020-11-30
eosio-bios and eosio-boot contracts support for KV inside eosio

linhuang-blockone:
## Change Description

This PR provides `eosio-bios` and `eosio-boot `contracts to support users to boot up a minimal chain with KV features in it.
- If `EOSIO_COMPILE_TEST_CONTRACTS` is set, `eosio-bios` and `eosio-boot` are dynamically built from source.

- If `EOSIO_COMPILE_TEST_CONTRACTS` is not set, pre-built `eosio-bios` and `eosio-boot` are used.

Those contracts are installed in `<install-dir>/etc/eosio/contracts` and packaged in `${EOS_PREFIX}/etc/eosio/contracts` in in brew and deb.

The PR also fixes compilation errors of `unittests/test-contracts/kv_addr_book/kv_addr_book.cpp`, `unittests/test-contracts/kv_bios/kv_bios.cp`,  `unittests/test-contracts/kv_test/kv_test.cpp` when `EOSIO_COMPILE_TEST_CONTRACTS` is set.

The PR has been tested with EOSIO_COMPILE_TEST_CONTRACTS on and off, and running `make install` and `<build-dir>/packages/generate_package.sh brew`

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


contracts/CMakeLists.txt
```diff
@@ -0,0 +1,63 @@
...
+   set(EOSIO_WASM_OLD_BEHAVIOR "Off")
+   # Remove visibility set by top CMakeList
+   set(CMAKE_VISIBILITY_INLINES_HIDDEN 0)
+   # Remove C++ compiler set by top CMakeList (contracts use eosio-cpp)
+   unset(CMAKE_CXX_STANDARD)
+   
+   find_package(eosio.cdt)
+endif()
+
+add_subdirectory(eosio.bios)
```
larryk85, line 62: These should CMake External projects.  If you look at the way eosio.contracts is setup and the way building the test contracts are. They don't include directly.

The reason for this is two fold.  The old behavior that you are relying on, is a kludge of CMake to hack in CDT's toolchain.  This is problematic for a couple of reasons, it can corrupt the cmake cache and fails for certain OSes and the behavior from doing the 'raw' find_package(eosio.cdt) will be gone after this release of CDT.

It has been left in CDT to build very legacy versions of eosio.contracts, but that no longer makes much sense.

So, a quick solution is to use `eosio-init` to produce the CMake scaffolding for these two projects and then you can just copy and paste them into this.

You can steal the CMake from `/unittests/CMakeLists.txt` for setting the toolchain file and the external project.

This will also remove the issue where it is building as a 'generic' target and forcing you have to manually append the `.wasm` on to the binary.

larryk85 (changes requested):

contracts/CMakeLists.txt
linhuang-blockone: Thank you Bucky very much for your comments! I have made the changes. Now the names are perfect.

larryk85 (approved):

PR #9721, opened 2020-12-01
compile time option to disable light validation options 📦

spoonincode:
## Change Description

In some high assurance scenarios it is desirable to enforce that nodeos fully validates all blocks. This is particularly true when the witness plugin is enabled. After all, a witness signature is intended to convey full validation by the signing node. The witness plugin does try to enforce that configuration options are set appropriately for full validation but it is possible to concoct sequences of restarts that can elude that check.

This PR adds a compile time option to disable light validation options which will decisively ensure full validation was used.

I realize ifdef’ing around the config items is a little hokey, but this seems to be a rather unintrusive way of accomplishing this. A couple unit tests need to be disabled as well; I’m not sure it’s useful modifying eosio_blocklog_prune_test to always do full validation so we can keep it in this mode?

`force-all-checks`? It does seem like maybe that ought to be always enabled when this new compile option is enabled. There are two reasons I’m not right now:
* witness plugin will not sign blocks during replay
* current expected high assurance deployments will not give untrusted write access to the block log

Certainly open to opinions on this one way or another

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
The point of `eosio_blocklog_prune_test` is to test that light validation works with pruned blocks. This is not possible if you do not have light validation. So this test should be disabled in this "mode".

heifner:
Should `disable-all-subjective-mitigations` be added to the list?

spoonincode:
`disable-all-subjective-mitigations` requires another different compile time option to be enabled, so I think maybe okay?

heifner (approved):

PR #9740, opened 2020-12-04
improve EOS VM OC cache lookup performance via hashed index

spoonincode:
## Change Description

The OC code cache was using an ordered index for the cache lookup. This results in surprisingly expensive lookups compared to a hashed index which suffices for this use case. Switch over to a hashed index.

Frees over 1% of main thread time on a busy chain in my tests. Actual results will vary depending on chain activity, hardware, etc

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
LGTM

b1bart (approved):

PR #9743, opened 2020-12-04
fixed get_kv_table_rows issues

kimjh2005:
## Change Description

Fixed issues:
-fixed a crash when --reverse and --index_value used at the same time.
-fixed the behavior or end range  to be inclusive  as in other get_table apis. 
 Before upper was exclusive [lower, upper), now inclusive [lower, upper]
-fixed a incorrect end of range problem when iterator doesn't have an exact match

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

Range is now inclusive on both ends like in get_table_rows.
--lower value1 --upper value2     :  Return result from value1 to value2 (now value2 is included)
ex) data set:1, 2, 3, 4, 5  
        --lower 1  --upper 3   will return  [1, 2, 3] instead of [1, 2]

-r -lower value1 --upper value2  : In reverse mode,  return result from value2 to value1 (value1 is included)
ex) data set:1, 2, 3, 4, 5  
     -r   --lower 1  --upper 3   will return  [3, 2, 1]  instead of [3, 2]

## Documentation Additions
- [ ] Documentation Additions


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2219,6 +2220,7 @@ read_only::get_table_rows_result read_only::get_kv_table_rows_context( const rea
       const string &index_value = *p.index_value;
       p.lower_bound = p.index_value;
       unbounded = true;
+      p.reverse = false;
```
kimjh2005, line 2223: This fixes the crash when --reverse is enabled with --index_value.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2290,8 +2292,8 @@ read_only::get_table_rows_result read_only::get_kv_table_rows_context( const rea
    if( has_upper_bound ) {
       auto exact_match = kv_context.kv_get(p.code.to_uint64_t(), ub_key.data(), ub_key.size(), value_size);
       status_ub = ub_itr->kv_it_lower_bound(ub_key.data(), ub_key.size(), &ub_key_size, &ub_value_size);
-      if( p.reverse && !point_query && !exact_match ) {
-            status_ub = ub_itr->kv_it_prev(&ub_key_size, &ub_value_size);
+      if( !point_query && !exact_match ) {
+         status_ub = ub_itr->kv_it_prev(&ub_key_size, &ub_value_size);
```
kimjh2005, line 2296: This fixes the incorrect end iterator resulting empty set.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2326,7 +2328,7 @@ read_only::get_table_rows_result read_only::get_kv_table_rows_context( const rea
       }
 
       unsigned int count = 0;
-      for( count = 0; cur_time <= end_time && count < p.limit && cmp < 0; cur_time = fc::time_point::now() ) {
+      for( count = 0; cur_time <= end_time && count < p.limit && cmp <= 0; cur_time = fc::time_point::now() ) {
```
kimjh2005, line 2331: cmp<=0  for inclusive range

plugins/chain_plugin/chain_plugin.cpp
linhuang-blockone: I like your review comment explaining how it fixes a particular problem.

linhuang-blockone (approved):

PR #9749, opened 2020-12-05
Added a new application method for sync from blockvault

ndcgundlach:
## Change Description

Added a new application method for blockvault to use for syncing.  It avoids a lot logic that is not necessary when syncing externally.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [ x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
@@ -70,7 +70,10 @@ struct blockvault_sync_strategy : public sync_callback {
    }
 
    void on_block(eosio::chain::signed_block_ptr block) override final {
-      ilog("Received block number ${bn}", ("bn", block->block_num()));
+      if (0 == (_num_blocks_received % 100)) {
+         ilog("Received block number ${bn}", ("bn", block->block_num()));
+      }
+      
       if (_check_shutdown()) {
          _shutdown();
       }
--- a/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
@@ -71,9 +71,9 @@ struct blockvault_sync_strategy : public sync_callback {
 
    void on_block(eosio::chain::signed_block_ptr block) override final {
       if (0 == (_num_blocks_received % 100)) {
-         ilog("Received block number ${bn}", ("bn", block->block_num()));
+         dlog("Received block number ${bn}", ("bn", block->block_num()));
       }
-      
+
       if (_check_shutdown()) {
          _shutdown();
       }
--- a/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
@@ -45,6 +45,10 @@ struct blockvault_sync_strategy : public sync_callback {
          ilog("Received no data from blockvault.");
          run_startup();
       }
+
+      ilog("Sync from blockvault completed. ${snap}. ${blks} blocks received. ${ulnk} blocks unlinkable",
+           ("snap", _received_snapshot ? "Got snapshot" : "No snapshot")
+           ("blks", _num_blocks_received)("ulnk", _num_unlinkable_blocks));
    }
 
    void on_snapshot(const char* snapshot_filename) override final {
@@ -66,6 +70,7 @@ struct blockvault_sync_strategy : public sync_callback {
    }
 
    void on_block(eosio::chain::signed_block_ptr block) override final {
+      ilog("Received block number ${bn}", ("bn", block->block_num()));
       if (_check_shutdown()) {
          _shutdown();
       }
@@ -76,7 +81,7 @@ struct blockvault_sync_strategy : public sync_callback {
 
       try {
          ++_num_blocks_received;
-         _blockchain_provider.incoming_block_sync_method(block, block->calculate_id());
+         _blockchain_provider.incoming_blockvault_sync_method(block);
       } catch (unlinkable_block_exception& e) {
          if (block->block_num() == 2) {
             elog("Received unlinkable block 2. Please double check if --genesis-json and --genesis-timestamp are "
--- a/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
@@ -88,7 +88,7 @@ struct blockvault_sync_strategy : public sync_callback {
 
          ++_num_blocks_received;
          auto rc = _blockchain_provider.incoming_blockvault_sync_method(block,
-            (!_received_snapshot || (block->block_num() != _snapshot_height +1)));
+            !(_received_snapshot && block->block_num() == _snapshot_height +1));
 
          EOS_ASSERT(rc, plugin_exception,
                     "Unable to sync block from blockvault, block num=${bnum}, block id=${bid}",
--- a/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
@@ -84,7 +84,7 @@ struct blockvault_sync_strategy : public sync_callback {
 
       try {
          ++_num_blocks_received;
-         EOS_ASSERT(_blockchain_provider.incoming_blockvault_sync_method(block), plugin_exception,
+         EOS_ASSERT(_blockchain_provider.incoming_blockvault_sync_method(block, (1 != _num_blocks_received)), plugin_exception,
                     "Unable to sync block from blockvault, block num=${bnum}, block id=${bid}",
                     ("bnum", block->block_num())("bid", block->calculate_id()));
       } catch (unlinkable_block_exception& e) {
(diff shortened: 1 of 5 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -292,9 +293,56 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         auto bsf = chain.create_block_state_future( id, block );
+
+         // abort the pending block
+         _unapplied_transactions.add_aborted( chain.abort_block() );
+
+         // push the new block
+         auto handle_error = [&](const auto& e)
+         {
+            elog((e.to_detail_string()));
+            app().get_channel<channels::rejected_block>().publish( priority::medium, block );
```
heifner, line 314: Do not call `rejected_block`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -292,9 +293,56 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         };
+
+         try {
+            block_state_ptr blk_state = chain.push_block( bsf, [this]( const branch_type& forked_branch ) {
+               _unapplied_transactions.add_forked( forked_branch );
+            }, [this]( const transaction_id_type& id ) {
+               return _unapplied_transactions.get_trx( id );
+            } );
+            if ( blockvault != nullptr ) {
+               blockvault->async_append_external_block(blk_state->dpos_irreversible_blocknum, blk_state->block, [](bool){});
```
heifner, line 325: Do not call.

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
```diff
@@ -76,7 +81,7 @@ struct blockvault_sync_strategy : public sync_callback {
 
       try {
          ++_num_blocks_received;
-         _blockchain_provider.incoming_block_sync_method(block, block->calculate_id());
+         _blockchain_provider.incoming_blockvault_sync_method(block);
```
heifner, line 84: Should EOS_ASSERT that this returned true.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
ndcgundlach: Doh!

plugins/producer_plugin/producer_plugin.cpp
ndcgundlach: removed

plugins/producer_plugin/producer_plugin.cpp
ndcgundlach: removed

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
ndcgundlach: added assertion

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -292,9 +293,52 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+            } );
+         } catch ( const guard_exception& e ) {
+            chain_plugin::handle_guard_exception(e);
+            return false;
+         } catch ( const std::bad_alloc& ) {
+            chain_plugin::handle_bad_alloc();
+         } catch ( boost::interprocess::bad_alloc& ) {
+            chain_plugin::handle_db_exhaustion();
+         } catch( const fc::exception& e ) {
+            handle_error(e);
```
b1bart, line 331: I think we know there will be cases where blockvault may deliver an unlinkable block.  Consider a timeline with a fork, one fork wins and a snapshot is created at a blockheight that existed on both forks:

```
         LIB 
          \/
Common -> A1 -> A2 -> A3 -> A4 -> HEAD
      \-> B1 -> B2 (orphan)
```

A snapshot taken at `A1` will prune `A1`, `B1` and all their ancestors from Blockvault but not `B2` as blockvault does not reconstruct the tree and therefore cannot realize that `B2` is orphaned.  This means a `sync` can contain `B2` (and perhaps additional blocks in the `B` chain) that are unlinkable after loading the snapshot. 

After the sync, we should still be in the right place even with the erroneous unlinkable blocks.

This is a long-winded way of saying we should either pre-check that the `prev` is a known block and early out if it is not OR we should log and drop the unlinkable block exception instead of throwing out.

b1bart (changes requested):

plugins/producer_plugin/producer_plugin.cpp
larryk85: This solution would solve the side effect of these erroneous unlinkable blocks, but it is really papering over the root cause (we are incorrectly or inconsistently applying pruning).

My worry is that it will try to sync now and get a lot of legitimate unlinkable blocks because of either a bug in our implementation or an inconsistent DB on the users side.  This would continue to drop everything on the floor and appear that it is fine at the end, but in reality in only applied a small number of the blocks that it needed to.

b1bart: That was how we defined pruning for blockvault so, while it is a root cause it is also expected behavior.  

To do better we either need to replicate forkdb logic in the backends/SQL which will complicate their nominal operation so that we have cleaner output during (hopefully) exceptional syncing cases where snapshots have sliced branches.  

As an alternative, we can reconstruct the trees on the client side but for the cases you are describing we would gain no additional guarantees and functionally we'd be duplicating fork-db.

larryk85: Yeah, I know that it is expected behavior given how we defined the rules for pruning (looking back at my post the work "incorrectly" shouldn't have been used).  I meant it will produce inconsistencies from being able to reapply everything that block vault owns.

And I'm not stating that we need to replicate the entirety of forkdb on the non-eos side.

But, some degree of being able to determine legitimate vs. illegitimate errors would allow for the system to not mask over or hide the original error.  Because, as you have already mentioned if the blocks are illegitimate or orphaned then we will ultimately be in a consistent state after syncing.  By creating the blanket that says any unlinkable block is acceptable, we lose that tight guarantee.  This doesn't necessarily need to be done as part of this fix (I understand the needs of this fix for today).  But, I think it would be good to circle back to this very shortly after.

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
```diff
@@ -66,6 +70,7 @@ struct blockvault_sync_strategy : public sync_callback {
    }
 
    void on_block(eosio::chain::signed_block_ptr block) override final {
+      ilog("Received block number ${bn}", ("bn", block->block_num()));
```
heifner, line 73: Do not want an info level log for every block received.

plugins/producer_plugin/producer_plugin.cpp
ndcgundlach: auto previous = chain.fetchblock_by_id(block->previous);

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
ndcgundlach: dlog?

ndcgundlach: throttled it to every 100 blocks

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
```diff
@@ -66,6 +70,10 @@ struct blockvault_sync_strategy : public sync_callback {
    }
 
    void on_block(eosio::chain::signed_block_ptr block) override final {
+      if (0 == (_num_blocks_received % 100)) {
+         ilog("Received block number ${bn}", ("bn", block->block_num()));
```
heifner, line 74: I would prefer still making it `dlog`

ndcgundlach: 👌

plugins/producer_plugin/producer_plugin.cpp
b1bart: Dlog and short circuit without failing for now.  We can revisit a better solution later as we don’t have a concrete proposal for anything that improves upon this yet.

plugins/chain_plugin/include/eosio/chain_plugin/blockvault_sync_strategy.hpp
```diff
@@ -84,7 +84,7 @@ struct blockvault_sync_strategy : public sync_callback {
 
       try {
          ++_num_blocks_received;
-         EOS_ASSERT(_blockchain_provider.incoming_blockvault_sync_method(block), plugin_exception,
+         EOS_ASSERT(_blockchain_provider.incoming_blockvault_sync_method(block, (1 != _num_blocks_received)), plugin_exception,
```
b1bart, line 87: I don’t think we can assume it’s only the first block that needs to skip the connectivity check.  We still have the edge case of a snapshot that sliced 2 or more possible chains.  We cannot guarantee that the first block from BV is on the best eventual chain.  

Can this be based on block height?  

If the incoming block is exactly snapshot height + 1 then we cannot perform the connectivity check.

b1bart (changes requested):

b1bart (approved):

PR #9751, opened 2020-12-05
blockvault_client_plugin logger

huangminghuang:
## Change Description

This PR create a new separate logger of blockvault_client_plugin and add checks for double production check for blockvault.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

The Loggers Section in https://developers.eos.io/manuals/eos/latest/nodeos/logging/index should add

* `blockvault_client_plugin`  - detailed logging for the blockvault client plugin.

### DIFF
--- a/plugins/blockvault_client_plugin/block_vault_impl.hpp
+++ b/plugins/blockvault_client_plugin/block_vault_impl.hpp
@@ -1,12 +1,12 @@
 #pragma once
 
-#include <fc/scoped_exit.hpp>
+#include "backend.hpp"
 #include <boost/asio.hpp>
+#include <boost/filesystem.hpp>
 #include <fc/io/datastream.hpp>
+#include <fc/scoped_exit.hpp>
 #include <memory>
 #include <thread>
-#include <boost/filesystem.hpp>
-#include "backend.hpp"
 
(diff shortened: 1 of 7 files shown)


plugins/blockvault_client_plugin/block_vault_impl.hpp
```diff
@@ -25,7 +25,7 @@ struct transform_callback : backend::sync_callback {
 
       if (uncompressed_file.size())
          target.on_snapshot(uncompressed_file.c_str());
-      else 
+      else
          elog("snapshot uncompress failed");
```
heifner, line 29: Didn't change this one.

heifner (changes requested):
Not sure I'm a fan of having a separate logger for this. But if you do go ahead then add this to the example in `programs/nodeos/logging.json`. You will also need to notify the doc team so they can add it to the documentation.

larryk85 (approved):

PR #9752, opened 2020-12-05
sync from block vault during block production

huangminghuang:
## Change Description

This PR allows nodeos to sync with block vault when it doesn't receive the required block from peer logical producer node within 1 second.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2053,6 +2077,36 @@ static auto maybe_make_debug_time_logger() -> std::optional<decltype(make_debug_
    }
 }
 
+void block_only_sync::schedule() {
+   if (!_pending) {
+      // wait one second to see if we can actually get the block from net plugin before we try to resync from block vault
+      _start_sync_timer.expires_from_now(boost::posix_time::seconds(1));
+      _pending = true;
+      _start_sync_timer.async_wait(
+          app().get_priority_queue().wrap(priority::high, [this](const boost::system::error_code& ec) {
```
b1bart, line 2086: This should capture a weak pointer of some sort (perhaps this needs to be a method of `_impl` so that it can capture the weak ptr there.  There is a race condition between this and shutdown where `this` could be a stale pointer when this job fires.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2053,6 +2077,36 @@ static auto maybe_make_debug_time_logger() -> std::optional<decltype(make_debug_
...
+             if (ec != boost::asio::error::operation_aborted) {
+                auto id = _impl->chain_plug->chain().last_irreversible_block_id();
+                fc_dlog(_log, "Attempt to resync from block vault");
+                _impl->blockvault->sync(&id, *this);
```
b1bart, line 2090: this can throw (on a snapshot or a few different block errors) which will bubble from here out.   We should ensure that we don't throw from a job handler so something here needs to catch errors and respond to them appropriately

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2053,6 +2077,36 @@ static auto maybe_make_debug_time_logger() -> std::optional<decltype(make_debug_
...
+   }
+}
+
+void block_only_sync::on_snapshot(const char*) {
+   EOS_THROW(producer_exception, "Attempting to resync from blockvault encountered a snapshot and the node must restart to continue!");
+}
+
+void block_only_sync::on_block(eosio::chain::signed_block_ptr block) {
+   try {
+      _impl->on_sync_block(block, true);
```
b1bart, line 2103: we pass true here but I do not believe we are sure that is correct.  If this node has been initialized with a snapshot and has _no_ other input and constructs a block rejected by block-vault then we cannot verify connectivity.  On that edge case we would want to pass `false`.  

That case would only trigger if when we called `sync` the head_block was not available from controller.  in that case we would not verify connectivity for any block at exactly head_block_num + 1

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -353,6 +372,11 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             return false;
          }
 
+         if (_block_vault_resync.is_pending() && _producers.count( block->producer ) > 0 ) {
+            // Cancel any pending resync from blockvault if we received any blocks from the same logical producer
+            _block_vault_resync.cancel();
```
b1bart, line 377: At this point, we have not validated the signature.   This should be moved underneath the `chain.push_block` call and its error handling when we do other book-keeping like conditionally setting `_production_enabled=true`

b1bart (changes requested):

plugins/producer_plugin/producer_plugin.cpp
b1bart: See the remarks here https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/basic_deadline_timer/cancel/overload2.html

This is what the destructor calls and if the timer has expired but the job it posts hasn’t been processed then it will not carry an aborted error code

huangminghuang: Change to follow the usage example of producer_plugin_impl::_timer which is cancel on plugin shutdown so that we don't have to use week pointer.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2053,6 +2078,43 @@ static auto maybe_make_debug_time_logger() -> std::optional<decltype(make_debug_
...
+      _pending = true;
+      _start_sync_timer.async_wait(
+          app().get_priority_queue().wrap(priority::high, [this](const boost::system::error_code& ec) {
+             if (!ec) {
+                auto id = _impl->chain_plug->chain().last_irreversible_block_id();
+                fc_dlog(_log, "Attempt to resync from block vault");
+                try {
+                  _impl->blockvault->sync(&id, *this);
+                } catch( fc::exception& er ) {
+                   wlog("Attempting to resync from blockvault encountered ${details}; the node must restart to "
```
larryk85, line 2094: Should be `fc_wlog`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2053,6 +2078,43 @@ static auto maybe_make_debug_time_logger() -> std::optional<decltype(make_debug_
...
+   }
+}
+
+void block_only_sync::on_snapshot(const char*) {
+   EOS_THROW(producer_exception, "a snapshot");
+}
+
+void block_only_sync::on_block(eosio::chain::signed_block_ptr block) {
+   try {
+      _impl->on_sync_block(block,  block->block_num() != _impl->chain_plug->chain().head_block_num() + 1);
```
larryk85, line 2111: __(Not necessarily a change request, but more of a strong comment)__
What is the reason for having the logic for `check_connectivity` live outside of the function?

This seems like an implementation detail of `on_sync_block` that is spilling over.  Plus you already have all of the pieces to evaluate this as variables on `on_sync_block` (`chain` and `blk_num`).

plugins/producer_plugin/producer_plugin.cpp
larryk85: You missed:
line 326 `dlog`.
line 341 `elog`

Not sure if this was intentional.

plugins/producer_plugin/producer_plugin.cpp
b1bart: I think that still has the problem.   in appbase::shutdown we will call two tight loops that call `shutdown` on plugins and then erase them from the list.  There is no event queue pump/draining in between though it may be happening on a different thread depending on how we entered shutdown. 

Regardless, its not safe to assume that the call to `cancel` isn't racing a posted expiration/invocation already in the io_context queue.   Observe that other uses of `_timer` in producer plugin are capturing and locking weak pointers.

plugins/producer_plugin/producer_plugin.cpp
b1bart: @huangminghuang this looks unaddressed still?

huangminghuang: That has been changed.

plugins/producer_plugin/producer_plugin.cpp
huangminghuang: I didn't write that function `on_sync_block` . However, if if bart agrees that the usage of the function in `blockvault_sync_strategy.hpp:91` where it checks `_received_snapshot` is unnecessary, I can change it.

plugins/producer_plugin/producer_plugin.cpp
huangminghuang: Fix in commit a406b8513355290fa9efca7873dcc36415d10d8d

plugins/producer_plugin/producer_plugin.cpp
b1bart: We can internalize the check to `on_sync_block` but it will do more than the boolean passed in from the outside caller that has more information.  It looks trivial because this logic is not correct. 

We should bring in @ndcgundlach  and discuss.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2083,15 +2083,16 @@ void block_only_sync::schedule() {
       // wait one second to see if we can actually get the block from net plugin before we try to resync from block vault
       _start_sync_timer.expires_from_now(boost::posix_time::seconds(1));
       _pending = true;
-      _start_sync_timer.async_wait(
-          app().get_priority_queue().wrap(priority::high, [this](const boost::system::error_code& ec) {
-             if (!ec) {
+      _start_sync_timer.async_wait(app().get_priority_queue().wrap(
+          priority::high, [this, weak_impl = _impl->weak_from_this()](const boost::system::error_code& ec) {
+             auto shared_impl = weak_impl.lock();
+             if (shared_impl.get() && !ec) {
```
b1bart, line 2089: While I understand that `weak_impl.lock()` should protect situations were `this` is stale it seems sloppy to use `this->_impl` rather than `shared_impl`.  I would even go so far as to route `this->_pending = false` through that `shared_ptr<>` as well.

b1bart (changes requested):

b1bart (approved):

PR #9755, opened 2020-12-07
[docs] Remove sudo command from install/uninstall script instructions

lparisc:
## Change Description

Resolves DEVREL-1017

Removing the `sudo` command from the install/uninstall scripts since they might create unintended side effects. It was also agreed that the instructions already advice that adequate permission is required to install on system folders.

## Change Type
**Select ONE**
- [x] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


johndebord (approved):
LGTM

PR #9760, opened 2020-12-07
Optimizations

heifner:
## Change Description

- Pack blocks for writing to the block log in a thread pool.
- Destroy block_state objects and their associated transaction_metadata objects in thread pool.
- Calculate merkle roots in a thread pool.

## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -539,13 +539,13 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                if( exception_is_exhausted( *trace->except, deadline_is_subjective )) {
                   _unapplied_transactions.add_incoming( trx, persist_until_expired, next );
                   if( _pending_block_mode == pending_block_mode::producing ) {
-                     fc_dlog(_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} COULD NOT FIT, tx: ${txid} RETRYING ",
+                     fc_dlog(_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} COULD NOT FIT ${c}, tx: ${txid} RETRYING ",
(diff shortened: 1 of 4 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -539,13 +539,13 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                if( exception_is_exhausted( *trace->except, deadline_is_subjective )) {
                   _unapplied_transactions.add_incoming( trx, persist_until_expired, next );
                   if( _pending_block_mode == pending_block_mode::producing ) {
-                     fc_dlog(_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} COULD NOT FIT, tx: ${txid} RETRYING ",
+                     fc_dlog(_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} COULD NOT FIT ${c}, tx: ${txid} RETRYING ",
```
b1bart, line 542: I think it is possible that those scraping logs would be using the pattern `COULD NOT FIT, tx:(.*)` to capture the transaction ID.   I think it is safer to append this code to the end and label it as an error code.

My reasoning being that it is far less likely that `RETRYING ?$` was used in an integrated scraper.  Moreso given the inconsistency between a trailing space in the two versions of this message.

b1bart (changes requested):

plugins/producer_plugin/producer_plugin.cpp
heifner: Moved.

b1bart (approved):

PR #9765, opened 2020-12-09
added next_key_bytes

kimjh2005:
## Change Description

Added next_key_bytes field to get_table_rows_result structure to indicate there are more rows. 
As the value is encoded as "bytes", the subsequent query must use --encode-type "bytes" to use it.

## Change Type
**Select *ONE*:**
- [x] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

When the “more” field is set to true, the “next_key” field will be set to a value that can be provided with --lower (or --upper if --reverse is also being used) using the same --encode-type that was provided with the original request (NOTE: This feature is not yet implemented). Also the “next_key_bytes” field will be set to a value that can be provided with --lower (or --upper if --reverse is also being used) but “ --encode-type  bytes” must be used.


brianjohnson5972 (approved):
Please update the API documentation portion, indicating that this is reference to the chain/get_kv_table_rows endpoint.

PR #9772, opened 2020-12-10
Saving and restoring undo stack at shutdown and startup

linhuang-blockone:
## Change Description

This PR saves RocksDB backing store undo stack when shutdown and restores undo stack upon startup.

This is a draft PR calling for review and quick turn around of resolving any design/coding issues.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

In \<the configured data-dir\>/state directory, a new file **undo_stack.dat** is introduced if ` --backing-store rocksdb` is used . It saves RocksDB undo stack's state before node shutdown and is used to restore __nodeos__'s undo stack at startup.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -487,13 +487,8 @@ struct controller_impl {
 
    void startup(std::function<void()> shutdown, std::function<bool()> check_shutdown, const snapshot_reader_ptr& snapshot) {
       EOS_ASSERT( snapshot, snapshot_exception, "No snapshot reader provided" );
-# warning TODO: Chain_kv needs to fix kv_undo_stack's revision after restar
-      // Currently kv_undo_stack returns a wrong revision after restart,
-      // failing tests/terminate-scenarios-test.py.
-      // As chain_kv is being rewritten, comment out this check for now
-      // and retest after chain_kv rewriting is finished.
-      //EOS_ASSERT( db.revision() == kv_undo_stack.revision(), database_revision_mismatch_exception,
-      //            "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_undo_stack.revision()) );
+      EOS_ASSERT( db.revision() == kv_db.revision(), database_revision_mismatch_exception,
+                  "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_db.revision()) );
       this->shutdown = shutdown;
       ilog( "Starting initialization from snapshot, this may take a significant amount of time" );
       try {
@@ -526,13 +521,8 @@ struct controller_impl {
 
    void startup(std::function<void()> shutdown, std::function<bool()> check_shutdown, const genesis_state& genesis) {
       EOS_ASSERT( db.revision() < 1, database_exception, "This version of controller::startup only works with a fresh state database." );
-# warning TODO: Chain_kv needs to fix kv_undo_stack's revision
-      // Currently kv_undo_stack returns a wrong revision after restart,
-      // failing tests/terminate-scenarios-test.py.
-      // As chain_kv is being rewritten, comment out this check for now
-      // and retest after chain_kv rewriting is finished.
-      //EOS_ASSERT( db.revision() == kv_undo_stack.revision(), database_revision_mismatch_exception,
-      //            "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_undo_stack.revision()) );
+      EOS_ASSERT( db.revision() == kv_db.revision(), database_revision_mismatch_exception,
+                  "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_db.revision()) );
       const auto& genesis_chain_id = genesis.compute_chain_id();
       EOS_ASSERT( genesis_chain_id == chain_id, chain_id_type_exception,
                   "genesis state provided to startup corresponds to a chain ID (${genesis_chain_id}) that does not match the chain ID that controller was constructed with (${controller_chain_id})",
@@ -566,9 +556,8 @@ struct controller_impl {
 
    void startup(std::function<void()> shutdown, std::function<bool()> check_shutdown) {
       EOS_ASSERT( db.revision() >= 1, database_exception, "This version of controller::startup does not work with a fresh state database." );
-# warning TODO: Chain_kv needs to fix kv_undo_stack's revision
-      //EOS_ASSERT( db.revision() == kv_undo_stack.revision(), database_revision_mismatch_exception,
-      //            "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_undo_stack.revision()) );
+      EOS_ASSERT( db.revision() == kv_db.revision(), database_revision_mismatch_exception,
+                  "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_db.revision()) );
       EOS_ASSERT( fork_db.head(), fork_database_exception, "No existing fork database despite existing chain state. Replay required." );
 
       this->shutdown = shutdown;
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -252,6 +252,11 @@ struct controller_impl {
                            { check_protocol_features( timestamp, cur_features, new_features ); }
       );
 
+      // Might be too late?
+      if ( cfg.backing_store == backing_store_type::ROCKSDB ) {
+         kv_db.set_datadir( cfg.state_dir );
+      }
+
       set_activation_handler<builtin_protocol_feature_t::preactivate_feature>();
       set_activation_handler<builtin_protocol_feature_t::replace_deferred>();
       set_activation_handler<builtin_protocol_feature_t::get_sender>();
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -252,11 +252,6 @@ struct controller_impl {
                            { check_protocol_features( timestamp, cur_features, new_features ); }
       );
 
-      // Might be too late?
-      if ( cfg.backing_store == backing_store_type::ROCKSDB ) {
-         kv_db.set_datadir( cfg.state_dir );
-      }
-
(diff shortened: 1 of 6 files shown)


libraries/chain/controller.cpp
```diff
@@ -252,6 +252,11 @@ struct controller_impl {
                            { check_protocol_features( timestamp, cur_features, new_features ); }
       );
 
+      // Might be too late?
+      if ( cfg.backing_store == backing_store_type::ROCKSDB ) {
```
brianjohnson5972, line 256: Is there a reason we don't just do this in the constructor for kv_db?  It is passed cfg.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -139,7 +139,7 @@ class session {
 
    /// \brief Constructs a child session from another instance of the same session type.
    explicit session(session& parent, std::nullptr_t);
-   session(const session&) = delete;
+   //session(const session&) = delete;
```
brianjohnson5972, line 142: Do you know why you needed to remove this? I don't think you should need a copy constructor.

linhuang-blockone: Thanks! Tim also spot this. I have changed it back in the next commit.

libraries/chain/controller.cpp
linhuang-blockone: I will change to use constructor of kv_db.

libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -259,7 +259,8 @@ class session {
    It& first_not_deleted_in_iterator_cache_(It& it, const It& end) const;
 
  private:
-   parent_variant_type m_parent{ nullptr };
+   //parent_variant_type m_parent{ nullptr };
```
TimothyBanks, line 262: We can remove this line.

libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -2,11 +2,21 @@
...
+#include <fc/io/datastream.hpp>
+#include <fc/io/raw.hpp>
+#include <fc/reflect/typename.hpp>
+#include <fstream>
 #include <b1/session/session.hpp>
 #include <b1/session/session_variant.hpp>
 #include <eosio/chain/exceptions.hpp>
 
 namespace eosio::session {
+   const uint32_t undo_stack_magic_number = 0x30510ABC;
```
TimothyBanks, line 16: Can we make all of these either `const` or `constexpr`?

libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -197,4 +215,111 @@ typename undo_stack<Session>::const_variant_type undo_stack<Session>::bottom() c
...
+                    ("min", undo_stack_min_supported_version)
+                    ("max", undo_stack_max_supported_version)
+         );
+
+         int64_t rev; fc::raw::unpack( ds, rev );
+         revision(rev);
+
+         size_t num_sessions; fc::raw::unpack( ds, num_sessions );
+         for( size_t i = 0; i < num_sessions; ++i ) {
+            session_type session;
```
TimothyBanks, line 262: It might provide better encapsulation if we did this instead
```
push();
auto& session = top();
```
and then remove lines 263-268 and line 278.

This change might require setting the revision after this loop and you would have to directly set it `m_revision = rev;`.  Otherwise `revision()` will throw because the undo_stack is no longer empty.

libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -66,18 +76,26 @@ class undo_stack {
...
-undo_stack<Session>::undo_stack(Session& head) : m_head{ &head } {}
+undo_stack<Session>::undo_stack(Session& head, const fc::path& datadir) : m_head{ &head }, m_datadir{ datadir } {
+   open();
+}
 
 template <typename Session>
 undo_stack<Session>::~undo_stack() {
+   close();
+
    for (auto& session : m_sessions) { session.undo(); }
```
TimothyBanks, line 99: Since `close` is draining the stack, this line is no longer needed.

TimothyBanks (changes requested):

TimothyBanks:
Thanks for the help Lin.

linhuang-blockone:
> Thanks for the help Lin.

Thank you Tim, Bucky, and Brian for all the help. My last commit made it serialization and deserialization work (I verified via logging). I will incorporate your other comments shortly.

libraries/chain_kv/include/b1/session/session.hpp
linhuang-blockone: Removed.

libraries/chain_kv/include/b1/session/shared_bytes.hpp
```diff
@@ -508,8 +508,14 @@ shared_bytes::shared_bytes_iterator<Iterator_traits>::operator[](size_t index) c
    return m_buffer[index];
 }
 
+struct shared_bytes_reflect{
```
larryk85, line 511: So, instead of making this struct and reflecting it.  What you should do is this then.

Instead of using the FC_REFLECT, because the internals of shared_bytes are not very usable through reflection.

Add two public methods to `shared_bytes`:
```c++
template <typename Stream>
inline Stream& operator<<(Stream& ds, const shared_bytes& b) {
   ds << b.size();
   ds.write(b.data(), b.size());
   return ds;
}
```
and
```c++
template <typename Stream>
inline Stream& operator>>(Stream& ds, shared_bytes& b) {
   std::size_t sz;
   ds >> sz;
   shared_bytes tmp = {sz};
   ds.read(tmp.data(), tmp.size());
   b = tmp;
   return ds;
}
```

libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -197,4 +215,123 @@ typename undo_stack<Session>::const_variant_type undo_stack<Session>::bottom() c
...
+         for( size_t i = 0; i < num_sessions; ++i ) {
+            session_type session;
+            if (m_sessions.empty())
+               session.attach(*m_head);
+            else
+               session.attach(m_sessions.back());
+
+            size_t num_updated_keys; fc::raw::unpack( ds, num_updated_keys );
+            for( size_t j = 0; j < num_updated_keys; ++j ) {
+               shared_bytes_reflect raw_key; fc::raw::unpack( ds, raw_key );
```
larryk85, line 270: If you implement the above methods for `shared_bytes`, this becomes...
```c++
shared_bytes key;
ds >> key;

shared_bytes value;
ds >> value;
```

libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -197,4 +215,123 @@ typename undo_stack<Session>::const_variant_type undo_stack<Session>::bottom() c
...
+   while ( !m_sessions.empty() ) {
+      auto& session = m_sessions.front();
+      auto updated_keys = session.updated_keys();
+      fc::raw::pack( out, updated_keys.size() ); // number of updated keys
+
+      for (const auto& key: updated_keys) {
+        auto value = session.read(key);
+
+        if ( value ) {
+           shared_bytes_reflect raw_key;
```
larryk85, line 317: with the datastream operators this would become...
```c++
out << key;
out << value;
```

larryk85 (changes requested):

libraries/chain_kv/include/b1/session/shared_bytes.hpp
TimothyBanks: shared_bytes already has an out stream overload and a method that can create a shared_bytes from a hex string (which is what is streamed out).  You would just need to provide the operator for streaming in.

libraries/chain_kv/include/b1/session/undo_stack.hpp
linhuang-blockone: Changed to use constexpr.

libraries/chain_kv/include/b1/session/undo_stack.hpp
linhuang-blockone: Changed. Had to use `auto& session = m_sessions.back();` as `top` returns a variant.

libraries/chain_kv/include/b1/session/undo_stack.hpp
linhuang-blockone: Removed.

libraries/chain_kv/include/b1/session/undo_stack.hpp
```diff
@@ -197,4 +213,120 @@ typename undo_stack<Session>::const_variant_type undo_stack<Session>::bottom() c
...
+                    "Unsupported version of Undo stack data file '${filename}'. "
+                    "Undo stack data version is ${version} while code supports version(s) [${min},${max}]",
+                    ("filename", undo_stack_dat.generic_string())
+                    ("version", version)
+                    ("min", undo_stack_min_supported_version)
+                    ("max", undo_stack_max_supported_version)
+         );
+
+         int64_t rev; fc::raw::unpack( ds, rev );
+         revision(rev);
```
TimothyBanks, line 256: You don't need to call this anymore.

linhuang-blockone: Removed.

libraries/chain_kv/include/b1/session/undo_stack.hpp
linhuang-blockone: Thanks for the suggestion! This simplifies the code. Changed.

libraries/chain_kv/include/b1/session/undo_stack.hpp
linhuang-blockone: Changed.

TimothyBanks (approved):

larryk85 (approved):

PR #9775, opened 2020-12-11
add script to install linux binary package

huangminghuang:
## Change Description

Add script to install linux package. The script should be configured by cmake and then published as a release asset. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

Add description that  installing linux binary package from github release asset should download the eosio-linux-binary-install.sh and run it directly. To install package which has been downloaded, one can use the `-f` options like 
```
 eosio-linux-binary-install.sh -f eosio-2.1.0-1.x86_64.rpm
```


dimas1185:
it doesn't seems to fix ```Amazon 2``` install. after install I've got:
```nodeos: symbol lookup error: nodeos: undefined symbol: lo_lseek64```

spoonincode:
I think this disregards the spirit of the binary packages. Asking users (or, in the case of this script, doing it behind their back) to install third party package repos is a rather invasive action with potential non trivial ramifications. One of the biggest headaches is that we don't know what kind of package conflicts the repo will add for users (for example, what happens if they already have installed, and are using, the system provided postgresql packages and libraries). It's also harder for a user to undo this action as opposed to just simply uninstalling the eosio package. **It's just about impossible for us to _ever_ automatically undo the side effects of this script safely in the future.**

I realize we're already doing this in the build scripts due to libpq there, which I also find deeply troubling for much of the same reasons above. Installing new trusted package signing keys on to a user's OS is a very invasive action that should not be taken lightly. Requiring such actions just to use the binary builds is simply too much.

nksanthosh:
Will tackle this after the upcoming release

kj4ezj (changes requested):
I agree with our build expert @spoonincode that adding a binary installation script is moving EOSIO in the wrong direction. These linux variants each offer ways to declare package dependencies that the PPA then manages on our behalf, and the fact that we are trying to work outside of that system points to a more fundamental design failure on our part.

Worse more, maintenance of the existing BASH scripts costs us a ton of time and money. I think investing in more is the wrong decision.

I recommend this pull request be closed, then we work with Product and our build experts to find the correct way to do this.

PR #9782, opened 2020-12-14
Update supported operating systems in README.md

johndebord:


dimas1185 (approved):

nksanthosh:
MacOS 10.15 Catalina support is seen in CICD but not in this list - we can add a separate JIRA issue for that and check with Product

PR #9803, opened 2020-12-17
Fix stdout console logging: Merge back #9582 to 2.0.x

softprofe:
## Change Description

Merge back pull https://github.com/EOSIO/eos/pull/9582 to 2.0.x. 

- `std_out` option for console logging has never worked.
- Fix support of `std_out` `stream` option for logging.
- Add `flush` option to example `logging.json` since someone might want to set it to false for stdout logging.

## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

- The `flush` option is not documented and should be added to the nodeos logging section.
https://developers.eos.io/manuals/eos/latest/nodeos/logging/index

### DIFF
--- a/programs/nodeos/logging.json
+++ b/programs/nodeos/logging.json
@@ -15,7 +15,8 @@
             "level": "error",
             "color": "red"
           }
-        ]
+        ],
+        "flush": true
       },
       "enabled": true
     },{
@@ -33,7 +34,8 @@
             "level": "error",
             "color": "red"
           }
-        ]
+        ],
+        "flush": true
       },
       "enabled": true
     },{


programs/nodeos/logging.json
```diff
@@ -15,7 +15,8 @@
             "level": "error",
             "color": "red"
           }
-        ]
+        ],
+        "flush": true
```
nickjjzhao, line 19: Is there any difference by setting flush to be true or false here?

softprofe: Not very sure what will happen if set flush to false. However, as this PR is merge back, it should be synchronized with original PR.

nickjjzhao: I wasn't sure about flush's value because the change description "Add flush option to example logging.json since someone might want to set it to false for stdout logging." I am going to approve this PR as  it's a back-merge.

nickjjzhao (approved):
LGTM

PR #9805, opened 2020-12-17
chain_plugin db intrinsic table RPC calls incorrectly handling --lower and --upper in certain scenarios

vzqzhang:
## Change Description
This PR fixes issue in https://blockone.atlassian.net/browse/EPE-515 where table RPC calls incorrectly handling --lower and --upper in certain scenarios

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
+++ b/libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
@@ -457,17 +457,15 @@ namespace detail {
          EOS_ASSERT(begin_key < end_key, database_exception, "Invalid iterator_pair request: begin_key was greater than or equal to end_key.");
          if (is_reverse_) {
             current_ = session.lower_bound(end_key);
-            EOS_ASSERT(current_ != session.end(), database_exception, "iterator_pair: failed to find lower bound of end_key");
             end_ = session.lower_bound(begin_key);
             // since this is reverse iterating, then need to iterate backward if this is a greater-than iterator,
             // to get a greater-than-or-equal reverse iterator
-            if ((*current_).first > end_key) {
+            if (current_ == session.end() || (*current_).first > end_key) {
                --current_;
+               EOS_ASSERT(current_ != session.end(), database_exception, "iterator_pair: failed to find lower bound of end_key");
             }
             // since this is reverse iterating, then need to iterate backward to get a less-than iterator
-            if (end_ != session.end()) {
-               --end_;
-            }
+            --end_;
          }
          else {
             current_ = session.lower_bound(begin_key);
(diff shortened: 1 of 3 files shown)


libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
```diff
@@ -457,17 +457,15 @@ namespace detail {
          EOS_ASSERT(begin_key < end_key, database_exception, "Invalid iterator_pair request: begin_key was greater than or equal to end_key.");
          if (is_reverse_) {
             current_ = session.lower_bound(end_key);
-            EOS_ASSERT(current_ != session.end(), database_exception, "iterator_pair: failed to find lower bound of end_key");
             end_ = session.lower_bound(begin_key);
             // since this is reverse iterating, then need to iterate backward if this is a greater-than iterator,
             // to get a greater-than-or-equal reverse iterator
-            if ((*current_).first > end_key) {
+            if (current_ == session.end() || (*current_).first > end_key) {
```
linhuang-blockone, line 463: Will `current` ever be NULL?

linhuang-blockone (commented):
As this fixes a bug, can you add a unit test case so we don't regress in the future?

libraries/chain/include/eosio/chain/backing_store/db_combined.hpp
brianjohnson5972: it is an iterator type, not a pointer, so it will always be valid to do *current_

vzqzhang:
> As this fixes a bug, can you add a unit test case so we don't regress in the future?

Added unit tests

vzqzhang:
> `result.rows.size() ` was checked to be 5. No need to check it again.

Removed

linhuang-blockone (approved):

tests/get_table_tests.cpp
```diff
@@ -116,6 +116,17 @@ BOOST_AUTO_TEST_CASE_TEMPLATE( get_scope_test, TESTER_T, backing_store_ts) { try
       BOOST_REQUIRE_EQUAL(name("initc"_n), result.rows[1].scope);
    }
 
+   param.lower_bound = "initb";
+   param.upper_bound = "initd";
```
brianjohnson5972, line 120: I missed that there was a scope with "initd", so you need to set upper_bound = "inite" (or past the last scope, so you could do param.upper_bound = "z"; which should guarantee that is past the last scope, since I doubt it has any scope values that start with "z".

brianjohnson5972 (changes requested):

tests/get_table_tests.cpp
vzqzhang: Updated

brianjohnson5972 (approved):

PR #9814, opened 2020-12-18
Re-enable full multi_index_tests

TimothyBanks:
## Change Description

Pre RC1 debugging showed that the expected key, that could be seen to have been added previously, was not returned by iterating from begin() to end(), so the test (of ROCKSDB, using backing_store_ts_only_cb instead of backing_store_ts) was turned off.  This PR re-enables the full test as it currently passes against the code in develop.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/api_tests.cpp
+++ b/unittests/api_tests.cpp
@@ -2178,12 +2178,10 @@ BOOST_FIXTURE_TEST_CASE(db_notify_tests, TESTER) {
    BOOST_TEST_REQUIRE(push_action( action({}, "notifier"_n, name(), {}), "notifier"_n.to_uint64_t() ) == "");
 }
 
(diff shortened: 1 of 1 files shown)


brianjohnson5972 (approved):
The test failures had nothing to do with api_tests, so I restarted them.

Issue #9649, opened 2020-11-09
file_space_handler extremely verbose message.


deckb:
When the threshold is hit there starts to be message every 2 seconds alerting that there will be an issue at some point. It would be nice to have a configuration option that limits the output of this message to a certain cadence or if the message would output every **n** bytes/KiB/MiB

https://github.com/EOSIO/eos/blob/924423185fd596c0ef4fca25b0c7a397439a4de5/plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp#L113

bogniq:
@deckb Thank you for reporting this!
There is an option `resource-monitor-interval-seconds` to set up the interval from 1 sec to 300 sec (default value is 2 sec). Do you think this option is good enough to solve this issue? 
https://github.com/EOSIO/eos/blob/924423185fd596c0ef4fca25b0c7a397439a4de5/plugins/resource_monitor_plugin/resource_monitor_plugin.cpp#L61
(also, I'm working to reproduce this problem and try different options).

deckb:
I'm not sure I want to change how often the check is performed just how often the warning message is posted. Something like `resource-monitor-warning-interval` where 0 means output every check else output every <n> checks. So if:
```
resource-monitor-interval-seconds = 2
reousrce-monitor-warning-interval = 30
```

The message would be every minute. Is that too confusing?

bogniq:
Yes, thank you for reminding that checking and warning need separate interval settings. I think `reousrce-monitor-warning-interval` can also be in seconds, with some limitations (e.g. it should be no less than `resource-monitor-interval-seconds`). How do you think?

Also, your alternative suggestion is to set the interval by every n bytes/KiB/MiB. Does it mean the warning is output every time the available space is `reduced` by n bytes/KiB/MiB? This option vs. warning time interval, which one do you think the user would prefer (or both needed)? Thanks.

deckb:
>Yes, thank you for reminding that checking and warning need separate interval settings. I think reousrce-monitor-warning-interval can also be in seconds, with some limitations (e.g. it should be no less than resource-monitor-interval-seconds). How do you think?

That would work as well. The way I described was a bit clunky and even as I wrote it I didn't love my answer.

>Also, your alternative suggestion is to set the interval by every n bytes/KiB/MiB. Does it mean the warning is output every time the available space is reduced by n bytes/KiB/MiB? 

Yes that is exactly right. 

[...]

PR #9826, opened 2020-12-23
Add warning interval option for resource monitor plugin

bogniq:
## Change Description

Option `resource-monitor-warning-interval` is added to the resource monitor plugin to fix issue #9649. This option sets `the number of monitor intervals` after which the warning message is output if the resource threshold is hit (currently only used for filesystem space). 

Before this option is added, the warning is output after every monitoring (e.g. by default every 2 seconds). Now this option can be set from 1 to 450 and has a default value of 30 (e.g. if the monitor interval is 2 seconds, and this warning interval is (by default) 30, then the warning will be output every 60 seconds if the threshold is hit).

**Note**: the naming of this option is discussed in the issue and needs to be documented to avoid confusion.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
+++ b/plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
@@ -36,33 +36,55 @@ namespace eosio::resource_monitor {
          shutdown_on_exceeded = new_shutdown_on_exceeded;
       }
 
-      bool is_threshold_exceeded() const {
+      void set_warning_interval(uint32_t new_warning_interval) {
+         warning_interval_in_secs = new_warning_interval;
+      }
+
+      bool is_threshold_exceeded() {
          // Go over each monitored file system
          for (auto& fs: filesystems) {
             boost::system::error_code ec;
             auto info = space_provider.get_space(fs.path_name, ec);
+            const auto path_str = fs.path_name.string();
             if ( ec ) {
                // As the system is running and this plugin is not a critical
                // part of the system, we should not exit.
                // Just report the failure and continue;
                wlog( "Unable to get space info for ${path_name}: [code: ${ec}] ${message}. Ignore this failure.",
-                  ("path_name", fs.path_name.string())
+                  ("path_name", path_str)
                   ("ec", ec.value())
                   ("message", ec.message()));
 
                continue;
             }
 
-            if ( info.available < fs.warning_available ) {
-               wlog("Space usage on ${path}'s file system approaching threshold. available: ${available}, warning_available: ${warning_available}", ("path", fs.path_name.string()) ("available", info.available) ("warning_available", fs.warning_available));
+            if ( info.available < fs.shutdown_available ) {
+               const std::string format = "Space usage on ${path}'s file system exceeded threshold ${threshold}%, available: ${available}, Capacity: ${capacity}, shutdown_available: ${shutdown_available}";
+               const auto args = fc::mutable_variant_object()
+                  ("path", path_str) 
+                  ("threshold", shutdown_threshold) 
+                  ("available", info.available) 
+                  ("capacity", info.capacity ) 
+                  ("shutdown_available", fs.shutdown_available);
+
+               latest_warnings[path_str] = fc::format_string(format, args);
+               // If shutdown_on_exceeded is true, the warning messages will be logged before the app quits.
                if ( shutdown_on_exceeded ) {
-                  wlog("nodeos will shutdown when space usage exceeds threshold ${threshold}%", ("threshold", shutdown_threshold));
+                  log_warnings();
                }
 
-               if ( info.available < fs.shutdown_available ) {
-                  wlog("Space usage on ${path}'s file system exceeded threshold ${threshold}%, available: ${available}, Capacity: ${capacity}, shutdown_available: ${shutdown_available}", ("path", fs.path_name.string()) ("threshold", shutdown_threshold) ("available", info.available) ("capacity", info.capacity) ("shutdown_available", fs.shutdown_available));
-
-                  return true;
+               return true;
+            } else if ( info.available < fs.warning_available ) {
+               const std::string format = "Space usage on ${path}'s file system approaching threshold. available: ${available}, warning_available: ${warning_available}";
+               const auto args = fc::mutable_variant_object()
+                  ("path", path_str) 
+                  ("available", info.available) 
+                  ("warning_available", fs.warning_available);
+               
+               latest_warnings[fs.path_name.string()] = fc::format_string(format, args);
+               if ( shutdown_on_exceeded ) {
+                  latest_warnings[path_str] += "\n" + 
+                     fc::format_string("nodeos will shutdown when space usage exceeds threshold ${threshold}%", fc::mutable_variant_object("threshold", shutdown_threshold));
                }
             }
          }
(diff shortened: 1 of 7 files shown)


plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -53,16 +57,15 @@ namespace eosio::resource_monitor {
...
+               if (output_threshold_warning) {
+                  wlog("Space usage warning: ${path}'s file system exceeded threshold ${threshold}%, available: ${available}, Capacity: ${capacity}, shutdown_available: ${shutdown_available}", ("path", fs.path_name.string()) ("threshold", shutdown_threshold) ("available", info.available) ("capacity", info.capacity) ("shutdown_available", fs.shutdown_available));
                }
-
-               if ( info.available < fs.shutdown_available ) {
-                  wlog("Space usage on ${path}'s file system exceeded threshold ${threshold}%, available: ${available}, Capacity: ${capacity}, shutdown_available: ${shutdown_available}", ("path", fs.path_name.string()) ("threshold", shutdown_threshold) ("available", info.available) ("capacity", info.capacity) ("shutdown_available", fs.shutdown_available));
-
-                  return true;
+               return true;
+            } else if ( info.available < fs.warning_available && output_threshold_warning ) {
```
dimas1185, line 65: why you are not returning ```true``` if ```available``` is less than ```warning_available``` ?

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -53,16 +57,15 @@ namespace eosio::resource_monitor {
...
+               wlog("Space usage warning: ${path}'s file system approaching threshold. available: ${available}, warning_available: ${warning_available}", ("path", fs.path_name.string()) ("available", info.available) ("warning_available", fs.warning_available));
+               if ( shutdown_on_exceeded) {
+                  wlog("nodeos will shutdown when space usage exceeds threshold ${threshold}%", ("threshold", shutdown_threshold));
```
dimas1185, line 68: you have same piece of code on line 61. you could think of different structure to avoid copy-paste.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -53,16 +57,15 @@ namespace eosio::resource_monitor {
...
                }
-
-               if ( info.available < fs.shutdown_available ) {
-                  wlog("Space usage on ${path}'s file system exceeded threshold ${threshold}%, available: ${available}, Capacity: ${capacity}, shutdown_available: ${shutdown_available}", ("path", fs.path_name.string()) ("threshold", shutdown_threshold) ("available", info.available) ("capacity", info.capacity) ("shutdown_available", fs.shutdown_available));
-
-                  return true;
+               return true;
+            } else if ( info.available < fs.warning_available && output_threshold_warning ) {
+               wlog("Space usage warning: ${path}'s file system approaching threshold. available: ${available}, warning_available: ${warning_available}", ("path", fs.path_name.string()) ("available", info.available) ("warning_available", fs.warning_available));
```
dimas1185, line 66: you have same piece of code on line 62. you could think of different structure to avoid copy-paste.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -53,16 +57,15 @@ namespace eosio::resource_monitor {
...
+               if ( shutdown_on_exceeded) {
```
dimas1185, line 67: earlier this code was executed before ```if ( info.available < fs.shutdown_available )``` . now it is not. is it expected?

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -139,7 +147,11 @@ class resource_monitor_plugin_impl {
    static constexpr uint32_t space_threshold_high = 99; // in percentage
    static constexpr uint32_t space_threshold_warning_diff = 5; // Warning issued when space used reached (threshold - space_threshold_warning_diff). space_threshold_warning_diff must be smaller than space_threshold_low
 
-   boost::asio::io_context     ctx;
+   static constexpr uint32_t def_monitor_warning_interval = 30; // After this number of monitor intervals, warning is output if the threshold is hit
+   static constexpr uint32_t warning_interval_low = 1;
+   static constexpr uint32_t warning_interval_high = 450; // e.g. if the monitor interval is 2 sec, the warning interval is at most 15 minutes
```
dimas1185, line 152: consider using min and max in names

tests/resource_monitor_plugin_test.py
```diff
@@ -112,6 +114,62 @@ def testCommon(title, extraNodeosArgs, expectedMsgs):
         if not isMsgInStderrFile(msg):
             errorExit ("Log should have contained \"%s\"" % (expectedMsgs))
 
+def extractTimestamp(msg):
+    matches = re.compile("\s+([0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}.[0-9]{3})\s").search(msg)
+    return datetime.strptime(matches.group(0).strip(), '%Y-%m-%dT%H:%M:%S.%f')
+
+intervalTolerance = 0.1 # 10%
```
dimas1185, line 121: is 0.1 just random number? if yes could you run this test in a loop and dump to log ```curInterval - timedelta(seconds=expectedInterval)```
to see how volatile it could be?
This is to avoid another flakey test where we need restart tests 2-3 times to get it passed.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
bogniq: The current process is: only when `available` is less than `shutdown_available`, this method will return `true`; otherwise, if `available` is less than `warning_available`, it will not return but only output warnings. Name like `is_shutdown_threshold_exceeded` may be more accurate, do you think it's necessary to modify the name?

dimas1185: I missed that. Apologize. I think you can leave old code name.
Also in original code it assumed that shutdown limit is always lower than warning so it is checked only if warning limit reached. Are you sure this logic is not broken? if statements reorganization looks a bit confusing, can you preserve original structure and just add another if statement(s) inside or add more to existing if statement(s)? Wouldn't it be easier?

bogniq: Yes, shutdown limit is always lower than warning. My consideration is: in the original code, when `if (info.available < fs.shutdown_available)` is within `if (info.available < fs.warning_available)`, and if the shutdown threshold is hit, warning message (`...file system approaching threshold...`) will be (unnecessarily) output and immediately followed by the shutdown warning (`...file system exceeded threshold`). I think here the first warning message is unnecessary. But maybe I missed something? Thanks.

dimas1185: if there an assert to check that ```shutdown_available``` is always less than ```warning_available``` that makes sense. this part looks good than.

tests/resource_monitor_plugin_test.py
bogniq: 0.1 (i.e. 10%) is based on observed warning interval deviations on buildkite and my local dockers. On buildkite, the max deviation is observed on macOS 10.15, where for the first test case (10-seconds interval, from 2 seconds monitor interval and warning interval 5) the deviation is about 0.6 sec, i.e. 6%. The second test case (30 seconds) has smaller deviation percentage. 
To be safer, I changed this tolerance to be 0.15 (15%) in the new commit.

plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
bogniq: Thanks. I changed all `_low` and `high` in the limits names to `_min` and `_max`.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
bogniq: Please see my above comments for the `if` blocks change.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
bogniq: Yes, it's hard-coded that `shutdown_available` is less than `warning_available` (which are calculated from `shutdown_threshold` and `warning_threshold`, and `shutdown_threshold - warning_threshold = 5 (space_threshold_warning_diff)`).

dimas1185 (approved):
code looks good to me but please double check with @iamveritas if we need any documentation changes related to this as we are adding new command line option.

bogniq:
Thank you! Let me contact @iamveritas for the documents.

iamveritas:
> Thank you! Let me contact @iamveritas for the documents.

I have created a ticket to track this issue from documentation point of view.
thank you.

PR #9828, opened 2020-12-23
Fix packed transaction version conversion -- Release 2.1.x

linhuang-blockone:
## Change Description

https://blockone.atlassian.net/browse/EPE-579

In Jungle testnet, LIBs were not advancing in a mixed of Rel. 2.0.x and Rel. 2.1.x nodes.
This was due to 2.0.x validation failure of blocks produced by 2.1.x, and in turn caused by
incorrect conversion between packed_transaction_v0 and packed_transaction_v1.
The solution is to make sure a conversion does not stripped any data.

New test cases are added.
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/transaction.cpp
+++ b/libraries/chain/transaction.cpp
@@ -484,6 +484,7 @@ packed_transaction::packed_transaction(const packed_transaction_v0& other, bool 
    unpacked_trx(other.unpacked_trx),
    trx_id(other.id())
 {
+   EOS_ASSERT( legacy, transaction_exception, "Full type of prunable_data_type is not supported" );
    estimated_size = calculate_estimated_size();
 }
 
@@ -498,17 +499,18 @@ packed_transaction::packed_transaction(packed_transaction_v0&& other, bool legac
    unpacked_trx(std::move(other.unpacked_trx)),
    trx_id(other.id())
 {
+   EOS_ASSERT( legacy, transaction_exception, "Full type of prunable_data_type is not supported" );
    estimated_size = calculate_estimated_size();
 }
 
 packed_transaction_v0_ptr packed_transaction::to_packed_transaction_v0() const {
    const auto* sigs = get_signatures();
+   EOS_ASSERT( std::holds_alternative<packed_transaction::prunable_data_type::full_legacy>(prunable_data.prunable_data), transaction_exception, "Failed to get full_legacy variant in to_packed_transaction_v0" );
    auto& legacy = std::get<prunable_data_type::full_legacy>(prunable_data.prunable_data);
-
    return std::make_shared<const packed_transaction_v0>( packed_trx,
                             sigs != nullptr ? *sigs : vector<signature_type>(),
                             legacy.packed_context_free_data,
-                            compression.value );
+                            compression);
 }
 
 uint32_t packed_transaction::get_unprunable_size()const {
--- a/libraries/chain/transaction.cpp
+++ b/libraries/chain/transaction.cpp
@@ -268,6 +268,17 @@ packed_transaction_v0::packed_transaction_v0(const bytes& packed_txn, const vect
    }
 }
 
+packed_transaction_v0::packed_transaction_v0(const bytes& packed_txn, const vector<signature_type>& sigs, vector<bytes> cfd, compression_type _compression)
+:signatures(sigs)
+,compression(_compression)
+,packed_trx(packed_txn)
+{
+   local_unpack_transaction( std::move(cfd) );
+   if( !packed_context_free_data.empty() ) {
+      local_unpack_context_free_data();
+   }
+}
+
 packed_transaction_v0::packed_transaction_v0( bytes&& packed_txn, vector<signature_type>&& sigs, bytes&& packed_cfd, compression_type _compression )
 :signatures(std::move(sigs))
 ,compression(_compression)
@@ -504,10 +515,11 @@ packed_transaction::packed_transaction(packed_transaction_v0&& other, bool legac
 packed_transaction_v0_ptr packed_transaction::to_packed_transaction_v0() const {
    const auto* sigs = get_signatures();
    const auto* context_free_data = get_context_free_data();
-   signed_transaction strx( transaction( get_transaction() ),
+
+   return std::make_shared<const packed_transaction_v0>( bytes(packed_trx),
                             sigs != nullptr ? *sigs : vector<signature_type>(),
-                            context_free_data != nullptr ? *context_free_data : vector<bytes>() );
-   return std::make_shared<const packed_transaction_v0>( std::move( strx ), get_compression() );
+                            context_free_data != nullptr ? *context_free_data : vector<bytes>(),
+                            compression.value );
 }
 
 uint32_t packed_transaction::get_unprunable_size()const {
--- a/libraries/chain/transaction.cpp
+++ b/libraries/chain/transaction.cpp
@@ -268,17 +268,6 @@ packed_transaction_v0::packed_transaction_v0(const bytes& packed_txn, const vect
    }
 }
 
-packed_transaction_v0::packed_transaction_v0(const bytes& packed_txn, const vector<signature_type>& sigs, vector<bytes> cfd, compression_type _compression)
-:signatures(sigs)
-,compression(_compression)
-,packed_trx(packed_txn)
-{
-   local_unpack_transaction( std::move(cfd) );
-   if( !unpacked_trx.context_free_data.empty() ) {
-      local_pack_context_free_data();
-   }
-}
-
 packed_transaction_v0::packed_transaction_v0( bytes&& packed_txn, vector<signature_type>&& sigs, bytes&& packed_cfd, compression_type _compression )
 :signatures(std::move(sigs))
 ,compression(_compression)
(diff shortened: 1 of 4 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -428,6 +428,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          } catch ( boost::interprocess::bad_alloc& ) {
             chain_plugin::handle_db_exhaustion();
          } catch( const fc::exception& e ) {
+            elog((e.to_detail_string()));
```
heifner, line 431: This should be `fc_dlog` and I would add a bit of description for context.

libraries/chain/transaction.cpp
```diff
@@ -504,10 +515,11 @@ packed_transaction::packed_transaction(packed_transaction_v0&& other, bool legac
...
    const auto* sigs = get_signatures();
    const auto* context_free_data = get_context_free_data();
-   signed_transaction strx( transaction( get_transaction() ),
+
+   return std::make_shared<const packed_transaction_v0>( bytes(packed_trx),
                             sigs != nullptr ? *sigs : vector<signature_type>(),
-                            context_free_data != nullptr ? *context_free_data : vector<bytes>() );
-   return std::make_shared<const packed_transaction_v0>( std::move( strx ), get_compression() );
+                            context_free_data != nullptr ? *context_free_data : vector<bytes>(),
+                            compression.value );
```
heifner, line 522: - You can copy over the `packed_context_free_data` directly. No need to pass in unpacked and then pack. So you can use the existing constructor.
- `bytes(packed_trx)` is not needed, can be just `packed_trx`

We should consider removing the partial support for `full` as we don't actually allow a `full` to be used. Note that `get_prunable_size_impl` for `full` asserts. I think we should modify the constructor of `packed_transaction` that takes a `packed_transaction_v0` to assert of `legacy == false`.

unittests/misc_tests.cpp
```diff
@@ -826,6 +826,12 @@ BOOST_AUTO_TEST_CASE(transaction_test) { try {
    BOOST_CHECK( packed != fc::raw::pack( static_cast<const transaction&>(pkt8.get_transaction()) ));
    BOOST_CHECK( packed == pkt8.get_packed_transaction() ); // extra maintained
 
+   // Check extra maintained in transitions of v0 -> v1 -> v0
+   packed_transaction pkt8_v0_to_v1(pkt8, true); // Build v1 with extra
+   auto pkt8_v0_to_v1_to_v0 = pkt8_v0_to_v1.to_packed_transaction_v0();
+   BOOST_CHECK_EQUAL(pkt.get_transaction().id(), pkt8_v0_to_v1_to_v0->get_transaction().id());
+   BOOST_CHECK( packed != fc::raw::pack( static_cast<const transaction&>(pkt8_v0_to_v1_to_v0->get_transaction()) ));
+   BOOST_CHECK( packed == pkt8_v0_to_v1_to_v0->get_packed_transaction() ); // extra maintained
```
heifner, line 834: - Add tests for context_free_data being maintained.
- Add tests for get_prunable_size and get_unprunable_size being equal.
- Add tests for empty context free data going through conversion. This tests with context_free_data but not with an empty context free data. We saw indication that it could be handled differently. Would be good to repeat all the above tests with empty context free data.
- Add test where you start with a packed_transaction with extra data and make the round trip.

heifner (changes requested):

unittests/misc_tests.cpp
linhuang-blockone: Done.

libraries/chain/transaction.cpp
linhuang-blockone: Thanks for your suggestion. This simplifies the code. I will do removing the partial support for full later. It did not work in my first try. I'd like to finish other comments first.

plugins/producer_plugin/producer_plugin.cpp
linhuang-blockone: Done.

libraries/chain/transaction.cpp
```diff
@@ -503,11 +503,12 @@ packed_transaction::packed_transaction(packed_transaction_v0&& other, bool legac
 
 packed_transaction_v0_ptr packed_transaction::to_packed_transaction_v0() const {
    const auto* sigs = get_signatures();
-   const auto* context_free_data = get_context_free_data();
-   signed_transaction strx( transaction( get_transaction() ),
+   auto& legacy = std::get<prunable_data_type::full_legacy>(prunable_data.prunable_data);
```
heifner, line 506: Since it is possible for this to be `none` an `EOS_ASSERT` indicating unable to convert from prunable_data_type would be a better error message than the std exception that `std::get` will throw here.

libraries/chain/transaction.cpp
```diff
@@ -503,11 +503,12 @@ packed_transaction::packed_transaction(packed_transaction_v0&& other, bool legac
...
+
+   return std::make_shared<const packed_transaction_v0>( packed_trx,
                             sigs != nullptr ? *sigs : vector<signature_type>(),
-                            context_free_data != nullptr ? *context_free_data : vector<bytes>() );
-   return std::make_shared<const packed_transaction_v0>( std::move( strx ), get_compression() );
+                            legacy.packed_context_free_data,
+                            compression.value );
```
heifner, line 511: The `.value` is not needed on `compression.value`

unittests/misc_tests.cpp
```diff
@@ -826,6 +834,35 @@ BOOST_AUTO_TEST_CASE(transaction_test) { try {
...
+
+   // Round trip from packed_transaction to packed_transaction
+   packed_transaction pkt_v0_to_v1(*pkt8_v0_to_v1_to_v0, true);
+   BOOST_CHECK_EQUAL(pkt.get_transaction().id(), pkt_v0_to_v1.get_transaction().id());
+   BOOST_CHECK( packed != fc::raw::pack( static_cast<const transaction&>(pkt_v0_to_v1.get_transaction()) ));
+   BOOST_CHECK( packed == pkt_v0_to_v1.get_packed_transaction() );
+   BOOST_CHECK_EQUAL(pkt_v0_to_v1.get_prunable_size(), pkt_v0_to_v1.get_prunable_size());
+   BOOST_CHECK_EQUAL(pkt_v0_to_v1.get_unprunable_size(), pkt_v0_to_v1.get_unprunable_size());
+   BOOST_REQUIRE_EQUAL(pkt_v0_to_v1.get_context_free_data()->size(), pkt_v0_to_v1.get_context_free_data()->size());
+   BOOST_CHECK(std::equal(pkt_v0_to_v1.get_context_free_data()->begin(), pkt_v0_to_v1.get_context_free_data()->end(), pkt_v0_to_v1.get_context_free_data()->begin()));
```
heifner, line 865: It should be possible to have extra data at the end of `packed_context_free_data` just like you can have extra data at the end of packed_trx. Please add some tests for that as well.

heifner (changes requested):

libraries/chain/transaction.cpp
linhuang-blockone: Do you think we should pass empty bytes to packed_transaction_v0 constructor if we fail to convert prunable_data_type instead?

heifner: No. Their is no v0 representation of a pruned transaction.

linhuang-blockone: Thanks. Fixed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -428,6 +428,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          } catch ( boost::interprocess::bad_alloc& ) {
             chain_plugin::handle_db_exhaustion();
          } catch( const fc::exception& e ) {
+            // The log will help us debug
+            fc_dlog(_log, "${exp}", ("exp", e.to_detail_string()));
```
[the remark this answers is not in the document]
linhuang-blockone: Fixed.

libraries/chain/transaction.cpp
linhuang-blockone: Fixed.

unittests/misc_tests.cpp
linhuang-blockone: Extra data was added to packed_context_free_data in tests. Local variables were renamed to help understand the code.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -428,6 +428,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          } catch ( boost::interprocess::bad_alloc& ) {
             chain_plugin::handle_db_exhaustion();
          } catch( const fc::exception& e ) {
+            fc_dlog(_log, "exception during push_block: ${ex}", ("ex", e.to_detail_string()));
```
b1bart, line 431: In normal operation this was being reported as an error to the default log  as a result of `handle_error(e)`. 

If we need to also log this to `_log` we should do so in `handle_error` as well so that it will properly report all non-fatal exceptions that are thrown.   

It seems like this may just be an artifact of debugging that could be removed?

b1bart (changes requested):

plugins/producer_plugin/producer_plugin.cpp
linhuang-blockone: Thanks for pointing this out. I think we should just remove it as the original 2.1 RC1.

b1bart (approved):

heifner (approved):

linhuang-blockone:
Thanks @heifner and @b1bart!

PR #9831, opened 2020-12-23
Fix or inhibit various compiler warnings while building on clang.

TimothyBanks:
## Change Description

There were a few compiler warnings occurring while building eos on clang:
- There were a few warning pragmas within the code.  I removed those and created JIRA issues to address them.
- A few 3rdParty libraries not maintained by block.one produce various compiler warnings when built on clang.  I decided to inhibit all compiler warnings in those projects.
- The build produced quite a few `ld: warning: direct access in weak symbol`.  I eliminated the majority of them by changing the default symbol visibility to hidden.  There are still a couple that are being generated when linking against libpqxx because it seems that library isn't using hidden by default.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


unittests/snapshot_tests.cpp
```diff
@@ -377,6 +375,7 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(test_chain_id_in_snapshot, SNAPSHOT_SUITE, snapsho
    verify_integrity_hash<SNAPSHOT_SUITE>(*chain.control, *snap_chain.control);
 }
 
+/*
```
TimothyBanks, line 378: It appear that maybe I was out of sync with develop and that this test was resurrected.  I'll need to verify that.

linhuang-blockone: Good catch. That test was reenabled a couple of weeks ago.

linhuang-blockone (approved):

linhuang-blockone (commented):
Would it be helpful to list the JIRAs here?

CMakeLists.txt
```diff
@@ -49,7 +49,8 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" OR "${CMAKE_CXX_COMPILER_ID}
 endif()
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "AppleClang")
-  set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
+  set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+  set(CMAKE_VISIBILITY_INLINES_HIDDEN 1)
```
linhuang-blockone, line 53: What warnings did visibility rules cause?

TimothyBanks: ON and 1 are the same.  I was trying to make it consistent with other places in our source code.

TimothyBanks:
I was told not to list JIRAs in the PR.  But there is only one JIRA issue for resolving the compilation warnings on MacOS.

Issue #9794, opened 2020-12-16
provided keys in error message is wrong


n13:
I received an error message about missing permissions [1]

Subsequent invocations of the same action caused ever-changing "provided keys" - the provided key, according to nodeos, was always a different one.

As it turns out nodeos merely takes a guess at the "provided keys" but can't accurately determine them. This should be reflected in the error message. It should either omit the message about provided keys or say "either you provided key [key] or something else went wrong with the signing of this transaction" 

Steps to reproduce:
1 - Make any transaction on a local chain using the wrong chain ID

Expected result:
Accurate error message example

"Either you provided keys [keys] or something else went wrong with the signing of this transaction"

Actual result: 
Misleading error message

Reference (this is what is happening):
https://github.com/EOSIO/eosjs/issues/518#issuecomment-472176189

[1]
```
{"code":500,"message":"Internal Service Error","error":{"code":3090003,"name":"unsatisfied_authorization","what":"Provided keys, permissions, and delays do not satisfy declared authorizations","details":[{"message":"transaction declares authority '{\"actor\":\"owner\",\"permission\":\"active\"}', but does not have signatures for it under a provided delay of 0 ms, provided permissions [], provided keys [\"EOS6b4ZQkcMuHmLdTmWERUL1PikWvX9LMzgcfF9of1nZAjt9PoJow\"], and a delay max limit of 3888000000 ms","file":"authorization_manager.cpp","line_number":532,"method":"check_authorization"}]}}
```

aclark-b1:
Which version of Nodeos are you using?  This message has been updated in our 2.1 rc1 release.

rusty-block-one:
Thanks to @tbfleming and @heifner for their assistance and recommendations for the solution to this issue.  @heifner for pointing out that removing the information from the message was the incorrect approach and @tbfleming for recommending a preferred solution.
The solution was to change the message from stating these are provided keys to stating these are recovered keys.  It clarifies what is being reported.

PR #9832, opened 2020-12-23
Issue 9794 - Remove incorrect logging of "provided keys" in error 309003

rusty-block-one:
## Change Description
In error code 309003, nodeos incorrectly reports provided keys. The code has no way to determine the correct provided keys. The reporting of provided keys was removed from the error message.

See #9794

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/authorization_manager.cpp
+++ b/libraries/chain/authorization_manager.cpp
@@ -565,12 +565,11 @@ namespace eosio { namespace chain {
          EOS_ASSERT( checker.satisfied( p.first, p.second ), unsatisfied_authorization,
                      "transaction declares authority '${auth}', "
                      "but does not have signatures for it under a provided delay of ${provided_delay} ms, "
-                     "provided permissions ${provided_permissions}, provided keys ${provided_keys}, "
+                     "provided permissions ${provided_permissions}, "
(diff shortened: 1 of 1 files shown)


jgiszczak:
The purpose of including provided keys in the error message is to show the user which keys were actually considered by the software while processing that transaction to assist the user in determining where the problem lies when authorization fails.  If the problem here is that the keys are never printed to the log, the requisite converter for a vector of keys should be created rather than removing this information from the log message.

jgiszczak (changes requested):
The purpose of including provided keys in the error message is to show the user which keys were actually considered by the software while processing that transaction to assist the user in determining where the problem lies when authorization fails. If the problem here is that the keys are never printed to the log, the requisite converter for a vector of keys should be created rather than removing this information from the log message.

jgiszczak (approved):

PR #9855, opened 2020-12-31
Update merkle.cpp

goodengineer:
Evaluation proposal for execution computing power saving:

Passing vector by reference to save memory copying effort added in function digest_type merkle(...), ids.size() 4 calls, reduced to 3 calls by introducing service variable size_t sz.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/merkle.cpp
+++ b/libraries/chain/merkle.cpp
@@ -32,20 +32,20 @@ bool is_canonical_right(const digest_type& val) {
 }
 
 
-digest_type merkle(vector<digest_type> ids) {
-   if( 0 == ids.size() ) { return digest_type(); }
-
-   while( ids.size() > 1 ) {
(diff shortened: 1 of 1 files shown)


jgiszczak:
This is broken without #9856.  We do not accept patches which break the build.

We also do not accept patches to `master`.  Target the `develop` branch.

PR #9872, opened 2021-01-04
Return failure traces when running with amqp_trx_plugin

heifner:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

- When running with amqp_trx_plugin, `transaciton_trace` can be returned for partial executed transactions which includes the exception. It is up to the user to understand these traces as failure cases.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -458,23 +458,28 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
          return [this, &trx, &chain, &next](const fc::static_variant<fc::exception_ptr, transaction_trace_ptr>& response) {
             next(response);
-            if (response.contains<fc::exception_ptr>() || response.get<transaction_trace_ptr>()->except) {
-               _transaction_ack_channel.publish(priority::low, std::pair<fc::exception_ptr, transaction_metadata_ptr>(response.get<fc::exception_ptr>(), trx));
+
+            fc::exception_ptr except_ptr;
+            if (response.contains<fc::exception_ptr>()) {
+               except_ptr = response.get<fc::exception_ptr>();
+            } else if (response.get<transaction_trace_ptr>()->except) {
+               except_ptr = response.get<transaction_trace_ptr>()->except->dynamic_copy_exception();
+            }
+            _transaction_ack_channel.publish(priority::low, std::pair<fc::exception_ptr, transaction_metadata_ptr>(except_ptr, trx));
+
+            if (except_ptr) {
                if (_pending_block_mode == pending_block_mode::producing) {
                   fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} is REJECTING tx: ${txid} : ${why} ",
                         ("block_num", chain.head_block_num() + 1)
                         ("prod", get_pending_block_producer())
                         ("txid", trx->id())
-                        ("why", response.contains<fc::exception_ptr>() ?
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -546,8 +547,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   if( !exhausted )
                      exhausted = block_is_exhausted();
                } else {
-                  auto e_ptr = trace->except->dynamic_copy_exception();
-                  send_response( e_ptr );
+                  if( return_failure_trace ) {
+                     send_response( trace );
```
swatanabe-b1, line 551: This will make the log messages from `send_response` wrong.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -491,7 +491,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       bool process_incoming_transaction_async(const transaction_metadata_ptr& trx,
                                               bool persist_until_expired,
                                               next_function<transaction_trace_ptr> next,
-                                              RetryLaterFunc retry_later)
+                                              RetryLaterFunc retry_later,
+                                              bool return_failure_trace = false)
```
swatanabe-b1, line 495: This seems pretty fragile to me.  I think you'd be better of making sure that all downstream consumers can handle a transaction trace with an exception appropriately, and never extract the exception here.  Either that or somehow use types that enforce consistency between return_failure_trace and next.

swatanabe-b1 (changes requested):

plugins/producer_plugin/producer_plugin.cpp
heifner: Good point. Updated.

plugins/producer_plugin/producer_plugin.cpp
heifner: For now this will only be true for `amqp_trx_plugin`. However, a new `send_transaction2` in the future should also work this way (at least optionally). The community has requested this type of behavior.

swatanabe-b1 (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -458,18 +458,20 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
          return [this, &trx, &chain, &next](const fc::static_variant<fc::exception_ptr, transaction_trace_ptr>& response) {
             next(response);
-            if (response.contains<fc::exception_ptr>()) {
+            if (response.contains<fc::exception_ptr>() || response.get<transaction_trace_ptr>()->except) {
                _transaction_ack_channel.publish(priority::low, std::pair<fc::exception_ptr, transaction_metadata_ptr>(response.get<fc::exception_ptr>(), trx));
```
swatanabe-b1, line 462: Not necessarily an exception_ptr.

heifner: Thanks. Fixed

swatanabe-b1 (approved):
Looks like that fixed the test failures.

PR #9873, opened 2021-01-04
chain api plugin / get_activated_protocol_features modifications

victorj8:
Made the following modifications related to the get_activated_protocol_features API call:

- Corrected documentation that indicated a wrong format of the POST json payload needed for this call.
- When the search_by_block num was used, the range of values to act was always empty due to a bug on the lower_bound_value used for both upper_bound and lower_bound.
- Added some extra checks on the integration test plugin_http_api_test to take into account the data returned by the api call.  

## Change Type
**Select *ONE*:**
- [x] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1784,7 +1784,7 @@ read_only::get_activated_protocol_features( const read_only::get_activated_proto
    auto lower = ( params.search_by_block_num ? pfm.lower_bound( lower_bound_value )
                                              : pfm.at_activation_ordinal( lower_bound_value ) );
 
-   auto upper = ( params.search_by_block_num ? pfm.upper_bound( lower_bound_value )
+   auto upper = ( params.search_by_block_num ? pfm.upper_bound( upper_bound_value )
                                              : get_next_if_not_end( pfm.at_activation_ordinal( upper_bound_value ) ) );
 
    if( params.reverse ) {
--- a/tests/plugin_http_api_test.py
+++ b/tests/plugin_http_api_test.py
@@ -6,8 +6,10 @@ import time
 import unittest
 
 from testUtils import Utils
+from testUtils import Account
 from TestHelper import TestHelper
 from Node import Node
+from Cluster import Cluster
 from WalletMgr import WalletMgr
 
 class PluginHttpTest(unittest.TestCase):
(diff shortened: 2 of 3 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1784,7 +1784,7 @@ read_only::get_activated_protocol_features( const read_only::get_activated_proto
    auto lower = ( params.search_by_block_num ? pfm.lower_bound( lower_bound_value )
                                              : pfm.at_activation_ordinal( lower_bound_value ) );
 
-   auto upper = ( params.search_by_block_num ? pfm.upper_bound( lower_bound_value )
+   auto upper = ( params.search_by_block_num ? pfm.upper_bound( upper_bound_value )
```
linhuang-blockone, line 1787: Good catch.

tests/plugin_http_api_test.py
```diff
@@ -60,6 +62,28 @@ def startEnv(self) :
...
 
+    def activateAllBuiltinProtocolFeatures(self):
+        self.nodeos.activatePreactivateFeature()
+
+        contract = "eosio.bios"
+        contractDir = "unittests/contracts/old_versions/v1.7.0-develop-preactivate_feature/%s" % (contract)
+        wasmFile = "%s.wasm" % (contract)
+        abiFile = "%s.abi" % (contract)
+
+        producerKeys = Cluster.parseClusterKeys(0)
```
linhuang-blockone, line 73: Why 0 for the argument `totalNodes`. It could be the reason of the test failure.

linhuang-blockone (changes requested):

mutagubyaivan:
How do I gain, I get many mails everyday

tests/plugin_http_api_test.py
victorj8: I think the argument is the number of additional producers other than the main one, this test is only using one producer, so that was the reason of 0. Anyways, for some reason this test was passing locally (probably using a leftover directory config from a previous execution), but then i started seeing the problem locally too, so probably the method doesn't work well if the nodeos instance is not set through the Cluster class.

In the end, I realized that the only reason i used that method from the Cluster class was to get the public and private keys for the eosio account, but as these get a default value if not defined explicitly, i just opted to use those instead.

linhuang-blockone (approved):

PR #9876, opened 2021-01-05
Changing python test framework to use binaries in bin directory

brianjohnson5972:
## Change Description
The python test framework expects binaries to be in the directory their source files are in (like programs/cleos/) and it is better that they just expect them in the bin/ directory.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/nodeos_run_test.py
+++ b/tests/nodeos_run_test.py
@@ -25,6 +25,15 @@ errorExit=Utils.errorExit
 cmdError=Utils.cmdError
 from core_symbol import CORE_SYMBOL
 
+for root, dirs, files in os.walk("."):
+    for directory in dirs:
+        if directory == "bin":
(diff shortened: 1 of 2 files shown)


mutagubyaivan:
How do I gain because I get mails everyday

brianjohnson5972:
There is more to it than just that change, this isn't worth pursuing more right now, closing.

PR #9877, opened 2021-01-05
DeepMind logger holds `const char*` fields that it uses for some of its naming and indicators

TimothyBanks:
## Change Description

DeepMind logger holds `const char*` fields that it uses for some of its naming and indicators.  This is an issue as the logger is a separate thread and these pointers can go stale.  It is working currently as a side effect that C style string literals will be in the code section of the program and therefore globally available with a lifetime of the entire program.  If you were to pass it a pointer from an std::string or std::string_view this can lead to dangling pointers and corruption.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/trace.hpp
+++ b/libraries/chain/include/eosio/chain/trace.hpp
@@ -87,7 +87,7 @@ namespace eosio { namespace chain {
    struct storage_usage_trace {
    public:
       storage_usage_trace(uint32_t action_id, std::string event_id, std::string_view family, std::string_view operation)
-      :storage_usage_trace(action_id, std::move(event_id), std::move(family), std::move(operation), ".")
+      :storage_usage_trace(action_id, std::move(event_id), family, operation, ".")
       {}
 
       storage_usage_trace(uint32_t action_id, std::string&& event_id, std::string_view family, std::string_view operation, std::string_view legacy_tag)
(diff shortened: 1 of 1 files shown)


heifner:
`dmlog_appender` does not run a separate thread. It might be called from multiple different threads in the future, but currently only called from the main thread.

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -86,19 +86,19 @@ namespace eosio { namespace chain {
 
    struct storage_usage_trace {
    public:
-      storage_usage_trace(uint32_t action_id, std::string event_id, const char* family, const char* operation)
-      :storage_usage_trace(action_id, std::move(event_id), family, operation, ".")
+      storage_usage_trace(uint32_t action_id, std::string event_id, std::string_view family, std::string_view operation)
+      :storage_usage_trace(action_id, std::move(event_id), std::move(family), std::move(operation), ".")
```
heifner, line 90: `std::move` of `family` and `operation` does nothing here and is confusing.

larryk85 (approved):

Issue #8984, opened 2020-04-21
dependency on third-party distributions


cc32d9:
The EOSIO repository depends on submodules from third-party authors on Github. Those libraries may eventually disappear or break. Some of these libraries, like https://github.com/cryptonomex/secp256k1-zkp.git  haven't been updated since long time (2015). `secp256k1-zkp` has also no master branch which causes errors in recursive submodules upgrades.

I believe B1 needs to maintain clones of those repositories that the software depends upon.

spoonincode:
@cc32d9 you called out the secp256k1 fork explicitly, and I think that concern was reasonable. Now that it's resolved are there others you'd like to see handled differently?

cc32d9:
I don't know, but basically any removal of a dependency repository will disrupt the process, so it makes sense to clone all third-party repos

spoonincode:
Unless there is a specific concern I would vote to close this. We have to draw the line _somewhere_ -- we can't mirror the entire Ubuntu source code or something. Most of the current deps are quite high profile and unlikely to disappear IMO. Since this is tracked in Jira, calling @aclark-b1 for input

PR #9885, opened 2021-01-06
replace old libsecp256k1 fork with upstream; boosts k1 recovery performance

spoonincode:
## Change Description

Replace usage of an old libsecp256k1 fork with latest upstream. This allows us to make use of (recently expired patented) improvements boosting key recovery performance. My measurements of the improvement:

CPU | Previous | New
--- | --- | ---
Intel 9600k | 50µs | 37µs
AMD 3950X | 60µs | 43µs

(exact performance numbers will vary based on hardware and many other factors; these numbers we gathered with only a single core loaded)

Okay, full disclosure: the old secp256k1 fork we were using could have the patented code enabled (and the performance improvement) with a compile time option. However, I'm taking this opportunity to upgrade us to the upstream mainline code instead of an old fork. Some users have voiced concern over using an aging non-upstream fork (#8984) and while I don't think there is any reasonable chance that Cryptonomex's repos will disappear, secp256k1's main upstream repo almost certainly won't.

Needs EOSIO/fc#180, where much of the changes actually are

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


dimas1185 (approved):

spoonincode:
Hey guys, we really need to approve and merge EOSIO/fc#180 first. Otherwise, now develop's submodule is no longer pointing to fc's eosio branch

PR #9890, opened 2021-01-07
plugin_http_api_test fix for intermittent assert error

brianjohnson5972:
## Change Description
Test is intermittently failing on one platform. Since the features were being turned on and immediately being verified to be on, added waiting for blocks to be produced before validating.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/plugin_http_api_test.py
+++ b/tests/plugin_http_api_test.py
@@ -106,8 +106,9 @@ class PluginHttpTest(unittest.TestCase):
         allFeatureDigests = [d['feature_digest'] for d in allProtocolFeatures]
         ACT_FEATURE_DEFAULT_LIMIT = 10
 
+        prot_feat_cmd = cmd_base + "get_activated_protocol_features"
+        default_cmd = prot_feat_cmd
         def get_activated_protocol_features():
-            prot_feat_cmd = cmd_base + "get_activated_protocol_features"
             ret_json = Utils.runCmdReturnJson(prot_feat_cmd)
             activated_protocol_features = ret_json["activated_protocol_features"]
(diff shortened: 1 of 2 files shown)


tests/plugin_http_api_test.py
```diff
@@ -106,12 +106,29 @@ def test_ChainApi(self) :
...
+            activated_protocol_features = ret_json["activated_protocol_features"]
+            self.assertEqual(type(activated_protocol_features), list)
+            return activated_protocol_features
+
+        initial_prot_feat = get_activated_protocol_features()
+
+        # wait till all the protocol features are activated in the blocks
+        pre_block = self.nodeos.getBlockNum()
+        later_block = pre_block + 2
+        self.nodeos.waitForBlock(later_block, reportInterval=1)
```
victorj8, line 122: Why not modify preactivateAllBuiltinProtocolFeature on Node.py or maybe add some comment to it? i am not sure if this is is going to be used again in the future, but i think i was expecting that after returning from that method (on line 84) the nodeos node is expected to be on a state that has those activated and don't have to remember to do this wait separately

brianjohnson5972:
Since the problem is intermittent and it would be good to actually investigate why this is failing (the features are turned on directly, so not a block delay issue) I'm closing it.

PR #9898, opened 2021-01-11
fix build problem of non-root user using script

softprofe:
## Change Description

Start 9898 description ......

On a clean cloud instance, add a non-root user and then use this user try build eos using script, error will happen because libpqxx is not installed because no root permission. 

The epe560 says the same problem, it is on 2.1 branch.

End 9898 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/helpers/eosio.sh
+++ b/scripts/helpers/eosio.sh
@@ -318,24 +318,24 @@ function ensure-libpq-and-libpqxx() {
         if [[ $NAME == "Amazon Linux" ]]; then
             #install libpq
             if [ ! -d /usr/include/libpq ]; then
-                eval $LIBPQ_SUDO amazon-linux-extras enable postgresql11 && \
-                    eval $LIBPQ_SUDO yum install -y libpq-devel
+                $LIBPQ_SUDO amazon-linux-extras enable postgresql11 && \
+                    $LIBPQ_SUDO yum install -y libpq-devel
             fi
             EXTRA_CMAKE_FLAGS="-DPostgreSQL_TYPE_INCLUDE_DIR=/usr/include/libpq"
         elif [[ $NAME == "CentOS Linux" ]]; then
             #install libpq
             if [ ! -d /usr/pgsql-13 ]; then
                 CENTOS_VERSION=$(rpm -E %{rhel})
-                eval $LIBPQ_SUDO yum install -y https://download.postgresql.org/pub/repos/yum/reporpms/EL-$CENTOS_VERSION-x86_64/pgdg-redhat-repo-latest.noarch.rpm
-                [ $CENTOS_VERSION -lt 8 ] ||  eval $LIBPQ_SUDO dnf -qy module disable postgresql
-                eval $LIBPQ_SUDO yum install -y postgresql13-devel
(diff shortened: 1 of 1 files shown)


scripts/helpers/eosio.sh
```diff
@@ -315,26 +315,26 @@ function ensure-libpq-and-libpqxx() {
...
             fi
             export PostgreSQL_ROOT=/usr/pgsql-13   
             export PKG_CONFIG_PATH=/usr/pgsql-13/lib/pkgconfig
         elif [[ $NAME == "Ubuntu" ]]; then
             # install libpq
             if [ ! -d /usr/include/postgresql ]; then 
-                echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > /etc/apt/sources.list.d/pgdg.list && \
-                    curl -sL https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add - && \
-                    apt-get update && apt-get -y install libpq-dev  
+                sudo echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > sudo /etc/apt/sources.list.d/pgdg.list && \
```
spoonincode, line 335: Does this work properly? The redirection w/ sudo usually requires more care.

softprofe: Yes, Per my test it can pass the build, however, the line 'echo "deb http://apt... ' more like did nothing after run.

spoonincode: Won't
```
sudo echo "deb http://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" > sudo /etc/apt/sources.list.d/pgdg.list
```
create a file named `sudo`?

softprofe: no, create nothing. just like no echo called.

heifner: `sudo echo "deb http://apt.postgresql.org/pub/repos/apt -pgdg main" > sudo hello`
Creates file `sudo` with contents:
```
bash-3.2$ cat sudo
deb http://apt.postgresql.org/pub/repos/apt -pgdg main hello
```

heifner: Funny it actual includes "hello", did not expect that. But pretty clear this is not what you want.

softprofe: Maybe, it didn't check carefully in currently directly, I checked /etc/apt/sources.list.d nothing changed, even no pgdg.list there.

heifner:
Please use standard PR template for your PR description providing a nice description that will be helpful when viewed at time of release.

softprofe:
> Why not just add the required package into `scripts/eosio_build_ubuntu_deps` so it uses the 'standard' way packages are installed? Why invent a new way to install a package here?

This question is out or this PR, maybe you should ask Huangming(the author of ensure-libpq-and-libpqxx), I guess the things required more than 'standard way' can do.    In this PR we would like to fix the problem of non-root user building by script.

huangminghuang (approved):

PR #9899, opened 2021-01-12
use oob cmake if possible so as to save script building time

softprofe:
## Change Description

Start 9899 description ......

By  using oob cmake instead of download source and build cmake , will save some script building time.

End 9899 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


scripts/eosio_build.sh
```diff
@@ -138,6 +138,30 @@ execute cd $REPO_ROOT
...
+   if [[ ${NAME} == "Ubuntu" ]]; then
+      if [[  -z $(command -v cmake 2>/dev/null)    ]] ; then
+         if [[ ${VERSION_ID} == "18.04"  ||   ${VERSION_ID} == "20.04" ]]; then
+            install-package cmake
+         fi
+      fi
+   elif [[ ${NAME} == "CentOS Linux"  ||  ${NAME} == "Amazon Linux" ]] ; then
+      install-package cmake
+      install-package cmake3
+      sudo alternatives --install /usr/local/bin/cmake cmake /usr/bin/cmake 10 \
```
spoonincode, line 152: Any time we do sudo for something other then simply installing a package from the main package repo makes me really really nervous. It's something we shouldn't do IMO, but.. that's just my opinion. (It's hard to undo broad system actions beyond just installing a package, and the stakes are very high for anything we're running as root)

The script seems to already make use of a `${CMAKE}` variable every time calling cmake. Can we set that variable to the proper cmake executable here instead?

softprofe: Sure, we can set ${CMAKE} to proper path.

softprofe: Per my test,   https://buildkite.com/EOSIO/eosio-build-scripts/builds/1182;     remove sudo also works,;
Currently this PR serve Ubuntu 18/20, Amazon Linux2, CentOS 8.
Now we don't have "unconformable"  sudo.

Issue #9871, opened 2021-01-04
Enhancement: Path-aware error for block log version error


sergmetelin:
It would be great if errors such as this: 
<img width="1193" alt="Screenshot 2021-01-04 at 1 20 02 PM" src="https://user-images.githubusercontent.com/2141500/103566195-8f83c700-4e8f-11eb-88af-c180c0fdb1c9.png">
would also show a location of the block log.
This instance of nodeos is installed via brew, so finding where the block log is located is a bit of a research whereas it would be much easier if it's printed out.

sanaraufx:
Fixed in 2.1 rc3.

PR #9905, opened 2021-01-13
Add log path for unsupported block log version exception

allenhan2:
## Change Description

For issue: https://github.com/EOSIO/eos/issues/9871

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

Test:

warn  2021-01-19T17:37:30.340 thread-0  chain_plugin.cpp:1321         plugin_initialize    ] 3190001 block_log_unsupported_version: unsupported version of block log
Unsupported version of block log. Block log version is 808464432 while code supports version(s) [1,4], log file: /Users/yue.han/workspace/eosio/develop/bin/dogfooding/data/blocks/blocks.log
    {"version":808464432,"min":1,"max":4,"log":"/Users/yue.han/workspace/eosio/develop/bin/dogfooding/data/blocks/blocks.log"}
    thread-0  block_log.cpp:57 read_from

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [x] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -282,7 +282,6 @@ namespace eosio { namespace chain {
 
      block_log_data() = default;
      block_log_data(const fc::path& path, mapmode mode = mapmode::readonly) { 
-        preamble.log_path = path;
         open(path, mode); 
      }
 
@@ -293,6 +292,7 @@ namespace eosio { namespace chain {
            file.close();
         file.open(path.string(), mode);
         fc::datastream<const char*> ds(this->data(), this->size());
+        preamble.log_path = path;
         preamble.read_from(ds);
         first_block_pos = ds.tellp();
         return ds;
@@ -302,7 +302,6 @@ namespace eosio { namespace chain {
       uint32_t      first_block_num() const { return preamble.first_block_num; }
       uint64_t      first_block_position() const { return first_block_pos; }
       chain_id_type chain_id() const { return preamble.chain_id(); }
-      fc::path      log_path() const { return preamble.log_path; }
 
       std::optional<genesis_state> get_genesis_state() const {
          return std::visit(overloaded{[](const chain_id_type&) { return std::optional<genesis_state>{}; },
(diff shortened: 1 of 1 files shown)


libraries/chain/block_log.cpp
```diff
@@ -279,7 +281,10 @@ namespace eosio { namespace chain {
    public:
 
      block_log_data() = default;
-     block_log_data(const fc::path& path, mapmode mode = mapmode::readonly) { open(path, mode); }
+     block_log_data(const fc::path& path, mapmode mode = mapmode::readonly) { 
+        preamble.log_path = path;
+        open(path, mode); 
+     }
 
      const block_log_preamble& get_preamble() const { return preamble; }
```
heifner, line 290: `open` is public but does not set the `log_path`
Should either be made private or set `log_path`.

libraries/chain/block_log.cpp
```diff
@@ -297,6 +302,7 @@ namespace eosio { namespace chain {
       uint32_t      first_block_num() const { return preamble.first_block_num; }
       uint64_t      first_block_position() const { return first_block_pos; }
       chain_id_type chain_id() const { return preamble.chain_id(); }
+      fc::path      log_path() const { return preamble.log_path; }
```
heifner, line 305: This is not used.

heifner (changes requested):

libraries/chain/block_log.cpp
allenhan2: It is in constructor, preamble is struct

heifner: But if you call it `open` again with a diff log_path they will be out of sync.

allenhan2: I see the point, thanks, it is better to update in open also.

heifner (approved):

PR #9915, opened 2021-01-15
extra transaction data integration test

dimas1185:
## Change Description

New python test to check that LIB is not stuck in case of extra data added manually to packed transaction

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,199 @@
...
+        #adding random trailing data
+        packedTrx["packed_trx"] = packed_trx_param + "00000000"
+
+        exitMsg = "failed to send packed transaction: %s" % (packedTrx)
+        node.processCurlCmd("chain", "send_transaction", json.dumps(packedTrx), silentErrors=False, exitOnError=True, exitMsg=exitMsg)
+
+    info = node.getInfo(exitOnError=True)
+    oldLIB = int(info["last_irreversible_block_num"])
+
+    blocksGap = (12 * (2 * math.floor(2 * prodCount / 3)) + 12) * 1.3 #adding 30% more to be on the safe side
```
brianjohnson5972, line 180: I'm not following exactly what this remaining section is testing for, but I think we could greatly reduce the time this takes.  I think we should look for the transactions in blocks (using getTransaction(..finalized)), for both node(0) and node(prod-1), and verify that node(0) and node(prod-1) both have advanced LIB past oldLIB and that head has progressed at least 12*pnode blocks, if you do node.getBlock(lastTransactionBlock+12*pnode) it will do the waiting if you need it to, just make sure you check that it returns a block (getBlock unfortunately just returns None on timeout, so need to verify)

brianjohnson5972 (changes requested):

pipeline.jsonc
```diff
@@ -30,7 +30,8 @@
         },
         "configuration":
         [
-            "170=v1.7.0"
+            "170=v1.7.0",
+            "209=v2.0.9"
```
kj4ezj, line 34: What prevents the protocol features multiversion test (`nodeos_multiple_version_protocol_feature_mv_test`) from testing against this new version (`v2.0.9`)?

kj4ezj (changes requested):
Just one change on my side.

tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,199 @@
...
+    Print("SERVER: %s" % (server))
+    Print("PORT: %d" % (port))
+
+    if localTest and not dontLaunch:
+        cluster.killall(allInstances=killAll)
+        cluster.cleanup()
+        Print("Stand up cluster")
+        specificExtraNodeosArgs = {}
+        associatedNodeLabels = {}
+        if len(specificExtraNodeosArgs):
```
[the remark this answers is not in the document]
dimas1185: it supposed to be filled out with `--chainbase=rocksdb` for some nodes. thanks, you found the bug.

tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,199 @@
...
+        packed_trx_param = packedTrx["packed_trx"]
+        if packed_trx_param is None:
+            cmdError("packed_trx is None. Json: %s" % (packedTrx))
+            errorExit("Can't find packed_trx in packed json")
+        
+        #adding random trailing data
+        packedTrx["packed_trx"] = packed_trx_param + "00000000"
+
+        exitMsg = "failed to send packed transaction: %s" % (packedTrx)
+        node.processCurlCmd("chain", "send_transaction", json.dumps(packedTrx), silentErrors=False, exitOnError=True, exitMsg=exitMsg)
```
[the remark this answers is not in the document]
dimas1185: added

tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,199 @@
...
+# General test that tests a wide range of general use actions around nodeos and keosd
+#
+###############################################################
+
+Print=Utils.Print
+errorExit=Utils.errorExit
+cmdError=Utils.cmdError
+from core_symbol import CORE_SYMBOL
+
+args = TestHelper.parse_args({"--host","--port","--prod-count","--defproducera_prvt_key","--defproducerb_prvt_key"
```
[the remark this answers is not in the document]
dimas1185: done

pipeline.jsonc
dimas1185: it should work. however is this something we want? @brianjohnson5972 ?

tests/nodeos_extra_packed_data_test.py
dimas1185: I fixed everything you suggested but I don't follow this part. so we need not just block + 12. We need LIB + 12.
`cleos get block` returns any block, including irreversible. I checked the code. So how we can wait till LIB move more efficiently?

brianjohnson5972: that was supposed to be:
`node.getBlock(lastTransactionBlock+12 * pnode)`

dimas1185: but it doesn't return LIB, isn't it?

brianjohnson5972: basically, waiting long enough to verify that all producers built off of those blocks.  Of course it doesn't prove it, so, you could also retrieve each of those blocks and put the block["producer"] in a dictionary and verify that len(prodDict) == pnode, That would be a thorough test.  You would only need to do it till while ( len(prodDict) < pnode), and then fail if you got past
`lastTransactionBlock + 12 * pnode`

pipeline.jsonc
brianjohnson5972: I think it should work.  Turn it on and then we will investigate if it fails.

tests/nodeos_extra_packed_data_test.py
dimas1185: @brianjohnson5972 ok. original test was related to LIB being stuck so I made waiting till LIB move. If your mentioned test is enough and we don't need to wait for LIB (if stuck block in node reproduces it) than I'm ok to make those changes. please confirm.

brianjohnson5972: "original test" meaning the test you copied it from?  If yes, then I think this is sufficient, since we will be verifying that each producer was able to process it.

pipeline.jsonc
dimas1185: so after looking deeper into it..
version 1.7.0 was last without preactivate protocol feature. as per the test it is supposed to go out of sync after activation. So this test supposed to fail with 2.0.9 and has no sense with it.
having that we don't have much 1.7.0 we may want to deactivate this test. Instead we will either maintain extra data multiversion test or create another to test our current master branch with latest release.
based on discussion with Santhosh and Brian

tests/CMakeLists.txt
```diff
@@ -157,6 +158,10 @@ if(ENABLE_MULTIVERSION_PROTOCOL_TEST)
   add_test(NAME nodeos_multiple_version_protocol_feature_mv_test COMMAND tests/nodeos_multiple_version_protocol_feature_test.py
            -v --clean-run --dump-error-detail --alternate-version-labels-file ${ALTERNATE_VERSION_LABELS_FILE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
   set_property(TEST nodeos_multiple_version_protocol_feature_mv_test PROPERTY LABELS mixed_version_tests)
+
+  add_test(NAME nodeos_extra_packed_data_test COMMAND tests/nodeos_extra_packed_data_test.py
```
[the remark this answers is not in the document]
dimas1185: done

tests/nodeos_extra_packed_data_test.py
```diff
@@ -78,16 +77,18 @@
...
+        if alternateVersionLabelsFile is not None:
+            #that means we are in multiversion mode
+            for i in range( int(pnodes / 2) ):
+                associatedNodeLabels[str(i)] = "209"
         
-        if cluster.launch(totalNodes=prodCount, 
-                          prodCount=1, 
-                          onlyBios=onlyBios, 
+        if cluster.launch(totalNodes=pnodes, 
+                          pnodes=1,
```
brianjohnson5972, line 91: So do you only want 1 producer and all the rest just nodes that are processing?  I missed earlier since you were using prod_count I thought that was the number of producers you want.  If you only want that one processor, you shouldn't use "-p" above and instead you should use "-n" and set totalNodes=args.n above, and then use totalNodes throughout instead of pnodes.

brianjohnson5972 (changes requested):

brianjohnson5972 (approved):

pipeline.jsonc
kj4ezj: > having that we don't have much 1.7.0 we may want to deactivate this test.

I don't think we want to delete it. I asked about this on the video call where @arhag presented on the multiversion test, and it sounded like we want to keep it for now because it can still catch breaking changes with the protocol features, but he can speak more on that if he wants to.

We have `v1.7.0` builds and binaries safely stored in the cloud and we can keep them more or less forever so there is no technical reason to push for deleting this test because it is old, or anything like that.

> create another to test our current master branch with latest release.

I thought that is what this pull request is for. I must not understand, can you expand on these two suggestions a bit?

kj4ezj: @dimas1185 answered my question via IM. There is a second definition in the protocol features test [here](https://github.com/EOSIO/eos/blob/74b0ac7b3f3885e82f86c8d0681b71033e21bbe4/tests/nodeos_multiple_version_protocol_feature_test.py#L87) and in the new test [here](https://github.com/EOSIO/eos/blob/74b0ac7b3f3885e82f86c8d0681b71033e21bbe4/tests/nodeos_extra_packed_data_test.py#L88) which defines which versions the tests are running against. I did poke around the code looking for this and skimmed the Python, but I had missed these definitions and believed one or both tests were behaving in an unexpected way, hence my reservations.

kj4ezj (approved):
I still think we should separate the configuration of the two tests and only define what versions we want to use in one place. So instead of having a list of binaries we download [here](https://github.com/EOSIO/eos/blob/74b0ac7b3f3885e82f86c8d0681b71033e21bbe4/pipeline.jsonc#L31), then a list of binaries we test against [here](https://github.com/EOSIO/eos/blob/74b0ac7b3f3885e82f86c8d0681b71033e21bbe4/tests/nodeos_multiple_version_protocol_feature_test.py#L87) and [here](https://github.com/EOSIO/eos/blob/74b0ac7b3f3885e82f86c8d0681b71033e21bbe4/tests/nodeos_extra_packed_data_test.py#L88), we would have one definition for each test. Perhaps this would be in the pipeline.jsonc like this...
```json
{
  "eosio": {
    "nodeos_multiple_version_protocol_feature_mv_test": [
      "v1.7.0"
    ],
    "nodeos_extra_packed_data_multiversion_test": [
      "v2.0.9"
    ],
  }
}
```
...or this...
```json
{
  "eosio": {
    "test": {
      "nodeos_multiple_version_protocol_feature_mv_test": [
        "v1.7.0"
      ],
      "nodeos_extra_packed_data_multiversion_test": [
        "v2.0.9"
      ],
    }
  }
}
```
...or, perhaps, we would not have anything in the pipeline.jsonc and instead declare these in CMakeLists.txt:
```cmake
add_test(NAME nodeos_extra_packed_data_multiversion_test COMMAND tests/nodeos_extra_packed_data_test.py --version '["v2.0.9"]' --alternate-version-labels-file ${ALTERNATE_VERSION_LABELS_FILE} -v --clean-run -p 8 --dump-error-detail WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
--and--
```cmake
add_test(NAME nodeos_multiple_version_protocol_feature_mv_test COMMAND tests/nodeos_multiple_version_protocol_feature_test.py --version '["v1.7.0"]' -v --clean-run --dump-error-detail --alternate-version-labels-file ${ALTERNATE_VERSION_LABELS_FILE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
In any case, I think having places the version needs to be defined is error-prone. I think it should be defined in just one place. Furthermore, I think that definition should not live in the test code. The test code should just test against the versions it is given. But I am flexible on that.

PR #9923, opened 2021-01-18
get_producers unittest 

huangminghuang:
## Change Description
EPE-582
This PR 
  - fixes a problem to add table row with f64 secondary key.
  - add a unit test for get_producers

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/backing_store/db_key_value_format.cpp
+++ b/libraries/chain/backing_store/db_key_value_format.cpp
@@ -290,4 +290,31 @@ namespace eosio { namespace chain { namespace backing_store { namespace db_key_v
       bytes composite_key = kt ? create_prefix_type_key(scope, table, *kt) : create_prefix_key(scope, table);
       return create_full_key(composite_key, code);
    }
+
+   eosio::session::shared_bytes create_full_primary_key(const eosio::session::shared_bytes& prefix, uint64_t primary_key) {
+      const auto prefix_size = detail::prefix_size<eosio::session::shared_bytes>();
+      b1::chain_kv::bytes composite_key;
+      constexpr static auto type_size = sizeof(key_type);
+      static_assert( type_size == 1, "" ); // this changing will break check below of prefix.back()
+      if (prefix.size() == prefix_size + type_size) {
+         const key_type kt = static_cast<key_type>(prefix[prefix.size() - 1]);
+         EOS_ASSERT(kt == key_type::primary, bad_composite_key_exception,
+                    "DB intrinsic create_full_primary_key should only be called with a prefix for a primary key, this has a prefix for type: ${type}",
+		    ("type", static_cast<uint64_t>(kt)));
+         composite_key.reserve(detail::key_size(key_type::primary));
+      }
+      else {
+         EOS_ASSERT(prefix.size() == prefix_size, bad_composite_key_exception,
+                    "DB intrinsic create_full_primary_key should only be called with a prefix of code, scope, and table and optionally a primary type indicator");
+         composite_key.reserve(type_size + detail::key_size(key_type::primary));
+	 composite_key.push_back(static_cast<char>(key_type::primary));
+      }
+
+      b1::chain_kv::append_key(composite_key, primary_key);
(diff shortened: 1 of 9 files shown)


brianjohnson5972:
Adding tests and further fixes

libraries/chain/backing_store/db_key_value_format.cpp
```diff
@@ -290,4 +290,31 @@ namespace eosio { namespace chain { namespace backing_store { namespace db_key_v
       bytes composite_key = kt ? create_prefix_type_key(scope, table, *kt) : create_prefix_key(scope, table);
       return create_full_key(composite_key, code);
    }
+
+   eosio::session::shared_bytes create_full_primary_key(const eosio::session::shared_bytes& prefix, uint64_t primary_key) {
+      const auto prefix_size = detail::prefix_size<eosio::session::shared_bytes>();
+      b1::chain_kv::bytes composite_key;
```
huangminghuang, line 296: Not sure why you need dynamic memory allocation here. I would use something like the following to eliminate the dynamic memory and branches. 
```
char composite_key[sizeof(prmary_key) + type_size] = { static_cast<char>(key_type::primary) } ;
int prerix_contains_key_type =  (prefix.size() == prefix_size + type_size);
EOS_ASSERT(prefix.size() == prefix_size || ( prerix_contains_key_type &&   static_cast<key_type>(prefix[prefix.size() - 1]) == key_type::primary_key), ....);
memcpy(composite_key + type_size, &primary_key, sizeof(prmary_key)); 
return eosio::session::make_shared_bytes<std::string_view, 2>(
        {  std::string_view{prefix.data(),  prefix.size()},
            std::string_view{composite_key + prerix_contains_key_type, 
                                         sizeof(composite_key) - prerix_contains_key_type} });

File: tests/get_producers_tests.cpp [hunk 2]
```diff
@@ -23,28 +23,39 @@
...
+   chain.produce_block();
+   chain.set_producers( {"dan"_n,"sam"_n,"pam"_n} );
+   chain.produce_blocks(30);
+
+   results = plugin.get_producers(params);
+   BOOST_REQUIRE_EQUAL(results.rows.size(), 3);
+   auto owners = std::vector<std::string>{"dan", "sam", "pam"};
+   auto it     = owners.begin();
+   for (const auto& elem : results.rows) {
+      BOOST_CHECK_EQUAL(elem["owner"].as_string(), *it++);
```
linhuang-blockone: Will the order of owner in owners always be dan, sam, and pam?

brianjohnson5972: it is controlled by the setprods call so it will always be that order

linhuang-blockone (approved):

libraries/chain/backing_store/db_key_value_format.cpp
brianjohnson5972: The reason for needing the dynamic memory is that initially all this code was written around chain_kv code.  Since the new rocksdb session code was added afterward, the decision was to take that existing code and massage it to the new format (with the intrinsic type char and contract name at the beginning).  This code was designed to prevent constructing the whole key from scratch, since they would all have the same prefix.  All this code definitely needs to be refactored and cleaned up, and there are a lot of copied code, but at least it has the same pattern of code.  I think it would be better to remove this method and just construct the key from scratch rather than have a highly optimized method that is different than other functions performing similar actions, and this is only being used currently for test code.

huangminghuang: To me, code without branches is easier to read. I am OK if you insisted on the existing code.

PR #9927, opened 2021-01-19
trace history log messages should print nicely in syslog

vzqzhang:
## Change Description

trace history log messages should print nicely in syslog

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/state_history/log.cpp
+++ b/libraries/state_history/log.cpp
@@ -126,7 +126,7 @@ void state_history_log::recover_blocks(uint64_t size) {
          break;
       pos = pos + state_history_log_header_serial_size + header.payload_size + sizeof(suffix);
       if (!(++num_found % 10000)) {
-         dlog("${num_found} blocks found, log pos = ${pos}", ("num_found", (unsigned)num_found)("pos", (unsigned long long)pos));
+         dlog("${num_found} blocks found, log pos = ${pos}", ("num_found", num_found)("pos", pos));
       }
    }
    read_log.flush();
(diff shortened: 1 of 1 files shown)


libraries/state_history/log.cpp
```diff
@@ -126,8 +126,7 @@ void state_history_log::recover_blocks(uint64_t size) {
          break;
       pos = pos + state_history_log_header_serial_size + header.payload_size + sizeof(suffix);
       if (!(++num_found % 10000)) {
-         printf("%10u blocks found, log pos=%12llu\r", (unsigned)num_found, (unsigned long long)pos);
-         fflush(stdout);
+         dlog("${num_found} blocks found, log pos = ${pos}", ("num_found", (unsigned)num_found)("pos", (unsigned long long)pos));
```
heifner, line 129: The `unsigned` and `unsigned long long` casts are not needed.

heifner (changes requested):

libraries/state_history/log.cpp
huangminghuang, line 129: The type casts for `num_found` and `pos` are unnecessary.

heifner (approved):

PR #9939, opened 2021-01-19
Placeholder

johndebord:


dimas1185:
I can see that you changed `get_account` but added tests for `get_account_limits` and `get_account_cpu_limit_ex`.
Can you add a test for `get_account` ? i.e. to greylist specific account and to set `greylist-limit` and then check it's return value? Any issues with that?

dimas1185 (changes requested):
see my general comment

PR #9950, opened 2021-01-20
Update transaction_context.cpp

goodengineer:
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -134,8 +134,7 @@ namespace eosio { namespace chain {
       rl.update_account_usage( bill_to_accounts, block_timestamp_type(control.pending_block_time()).slot );
 
       // Calculate the highest network usage and CPU time that all of the billed accounts can afford to be billed
-      int64_t account_net_limit = 0;
-      int64_t account_cpu_limit = 0;
+      int64_t account_net_limit = 0,account_cpu_limit = 0;
       bool greylisted_net = false, greylisted_cpu = false;
       std::tie( account_net_limit, account_cpu_limit, greylisted_net, greylisted_cpu) = max_bandwidth_billed_accounts_can_pay();
       net_limit_due_to_greylist |= greylisted_net;
@@ -173,7 +172,7 @@ namespace eosio { namespace chain {
          validate_account_cpu_usage( billed_cpu_time_us, validate_account_cpu_limit, true );
       }
 
-      eager_net_limit = (eager_net_limit/8)*8; // Round down to nearest multiple of word size (8 bytes) so check_net_usage can be efficient
+      //eager_net_limit = (eager_net_limit/8)*8; // Round down to nearest multiple of word size (8 bytes) so check_net_usage can be efficient
 
       if( initial_net_usage > 0 )
          add_net_usage( initial_net_usage );  // Fail early if current net usage is already greater than the calculated limit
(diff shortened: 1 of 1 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -134,8 +134,7 @@ namespace eosio { namespace chain {
       rl.update_account_usage( bill_to_accounts, block_timestamp_type(control.pending_block_time()).slot );
 
       // Calculate the highest network usage and CPU time that all of the billed accounts can afford to be billed
-      int64_t account_net_limit = 0;
-      int64_t account_cpu_limit = 0;
+      int64_t account_net_limit = 0,account_cpu_limit = 0;
```
heifner, line 137: why? This just makes it less readable.

libraries/chain/transaction_context.cpp
```diff
@@ -173,7 +172,7 @@ namespace eosio { namespace chain {
          validate_account_cpu_usage( billed_cpu_time_us, validate_account_cpu_limit, true );
       }
 
-      eager_net_limit = (eager_net_limit/8)*8; // Round down to nearest multiple of word size (8 bytes) so check_net_usage can be efficient
+      //eager_net_limit = (eager_net_limit/8)*8; // Round down to nearest multiple of word size (8 bytes) so check_net_usage can be efficient
```
heifner, line 175: This should not be removed.

libraries/chain/transaction_context.cpp
```diff
@@ -511,10 +509,8 @@ namespace eosio { namespace chain {
       // Calculate the new highest network usage and CPU time that all of the billed accounts can afford to be billed
       auto& rl = control.get_mutable_resource_limits_manager();
       const static int64_t large_number_no_overflow = std::numeric_limits<int64_t>::max()/2;
-      int64_t account_net_limit = large_number_no_overflow;
-      int64_t account_cpu_limit = large_number_no_overflow;
-      bool greylisted_net = false;
-      bool greylisted_cpu = false;
+      int64_t account_net_limit = large_number_no_overflow,account_cpu_limit = large_number_no_overflow;
+      bool greylisted_net = false,greylisted_cpu = false;
```
heifner, line 513: This does nothing except make it less readable.

heifner (changes requested):

libraries/chain/transaction_context.cpp
goodengineer: Correct, because it got rounded after /8 if eager_net_limit is capable to be separated with reminder it does, otherwise it does nothing. I beg your pardon, from first sight it can be seen like /8 *8 does nothing if eager_net_limit is divisible by 8 without reminder.

libraries/chain/transaction_context.cpp
goodengineer: "Why?" - m just trying to contribute only what is correct, doesn't deteriorate and if spotted to enhance. It is also a matter of style and saving code storage space as I have explained already before. It's up to you to accept the amendment(restyling or similar in this specific case to save code storage rows) or leave it as per your view. Let us agree on this.

libraries/chain/transaction_context.cpp
goodengineer: For which reasons? m just trying to contribute only what is correct, doesn't deteriorate and if spotted to enhance. It is also a matter of style and saving code storage space as I have explained already before. It's up to you to accept the amendment(restyling or similar in this specific case to save code storage rows) or leave it as per your view. Let us agree on this.

heifner: What do you mean by "saving code storage space"?

PR #9976, opened 2021-01-29
EPE-389 Fix net_plugin stall during head catchup

rusty-block-one:
This fix handles situations where a node is in head_catchup state and
recieves the sync_last_requested_num and then does a sync_update_expected
for the next block. And the next block has already been received by the
node. In this instance, the block will not receive the next block and
will be effectively hung.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -148,7 +148,7 @@ namespace eosio {
 
    private:
       constexpr static auto stage_str( stages s );
-      void set_state( stages s );
+      bool set_state( stages s );
       bool is_sync_required( uint32_t fork_head_block_num );
       void request_next_chunk( std::unique_lock<std::mutex> g_sync, const connection_ptr& conn = connection_ptr() );
       void start_sync( const connection_ptr& c, uint32_t target );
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1965,6 +1965,9 @@ namespace eosio {
 
          if( set_state_to_head_catchup ) {
             set_state( head_catchup );
+            if( blk_num == sync_last_requested_num ) {
+                request_next_chunk( std::move( g_sync ));
```
heifner, line 1969: I have not looked at the logic of this, but you do need to lock `g_sync` before the call. `request_next_chunk` assumes `g_sync` is locked. It was unlocked on line 1945 above.

heifner: `g_sync` needs to be locked before accessing `sync_last_requested_num` as that var is protected by `sync_mtx`.

rusty-block-one: killing this pull request and creating a new one

rusty-block-one:
Reissuing the pull request

PR #9991, opened 2021-02-03
Provide better ASM compiler hints and use resulting variable

jgiszczak:
## Change Description
Improve LLVM bug 39427 workaround to behave better in the presence of multiple eligible compilers.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
What is this resolving? The problem is that llvm 7.0.x compiled with gcc cannot be called from a clang compiled file.

So I'm not sure why would ever set
```
set(CMAKE_ASM${ASM_DIALECT}_COMPILER_INIT "clang++")
```
because that would seem to completely negate the workaround?

jgiszczak:
On Ubuntu 20 with both gcc and multiple versions of clang installed this allows the build to succeed where it otherwise fails because it's unable to find a compiler that can handle the llvmWARshim.llvm.  Setting `CMAKE_ASM${ASM_DIALECT}_COMPILER_INIT` allows it to find a compiler.  Perhaps the changes should be conditional on the llvm version found?

spoonincode:
This work around is already only active strictly when llvm 7.0.x is used with clang
https://github.com/EOSIO/eos/blob/58f90b9666dc37cf35eac5dafb6db1c4138024e7/libraries/chain/CMakeLists.txt#L33-L38

Did you not have g++ installed? Maybe that's what is needed here: double check that g++ is installed.

jgiszczak:
I have gcc 9 and gcc 10 installed, along with llvm 7, 8, 9, 10, and 11.  (I'm not entirely sure why there are so many different compilers installed.)  But indeed I do not have any version of g++ installed in that VM.  I was correcting the logic path where CMake selects LLVM 10 as the compiler.  It was trying to use g++ to build llvmWARshim.llvmwar, which isn't installed.  It successfully builds it with clang++, which it has selected as the CMAKE_ASM-LLVMWAR_COMPILER.

jgiszczak:
Compiler identification is apparently not working correctly.  `CMAKE_CXX_COMPILER` is `/usr/bin/clang++` which self-identifies like so:
```
$ /usr/bin/clang++ --version
clang version 10.0.0-4ubuntu1 
Target: x86_64-pc-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```

spoonincode:
I think maybe you might be confusing the clang version you're compiling with and the llvm library version that is being linked with. These are two separate versions and do not need to be equal. Since the default llvm on 20.04 is llvm10 my guess is you've done something (maybe even just passing an argument to cmake) to pick the llvm7 library.

It's possible that newer clang versions do not have the ABI incompatibility any longer and that's why using clang++ here seems to work. But that negates the entire purpose of the workaround: the always-safe thing to do is build this source file with g++ when using llvm 7.0.x library.

jgiszczak:
I haven't passed an argument or otherwise actively done anything to pick any particular version of libllvm.  All the corresponding libllvm versions are installed so I'd have thought it would automatically select libllvm10.

spoonincode:
I tried myself on 20.04 installing `llvm-7-dev llvm-8-dev llvm-9-dev llvm-dev` and was indeed getting llvm10 when doing cmake. But on closer inspection of documentation I'm not exactly sure how/why cmake picked 10 over 9/8/7.

But it still seems like adding a check that g++ is installed and/or works in here
https://github.com/EOSIO/eos/blob/58f90b9666dc37cf35eac5dafb6db1c4138024e7/libraries/chain/CMakeLists.txt#L33-L36
is the right fix to me.

PR #9997, opened 2021-02-03
Addded caching value in iter store to avoid extra retrievals from RocksDB

brianjohnson5972:
## Change Description
Added caching value in the iter store and avoiding accessing RocksDB and removed unneeded RocksDB iterators from the typical use case path.  This is only an optimization.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/backing_store/db_context_rocksdb.cpp
+++ b/libraries/chain/backing_store/db_context_rocksdb.cpp
@@ -178,7 +178,7 @@ namespace eosio { namespace chain { namespace backing_store {
 
       EOS_ASSERT( !old_key_value.value, db_rocksdb_invalid_operation_exception, "db_store_i64 called with pre-existing key");
 
-      primary_lookup.add_table_if_needed(old_key_value.full_key, payer);
+      primary_lookup.add_table_if_needed(scope_name, table_name, payer);
 
       const payer_payload pp{payer, value, value_size};
       set_value(old_key_value.full_key, pp);
@@ -200,7 +200,9 @@ namespace eosio { namespace chain { namespace backing_store {
 
       const unique_table t { receiver, scope_name, table_name };
       const auto table_ei = primary_iter_store.cache_table(t);
-      return primary_iter_store.add(primary_key_iter(table_ei, id, payer));
+      const auto iterator = primary_iter_store.add(primary_key_iter(table_ei, id, payer));
+      primary_iter_store.set_value(iterator, pp.value, pp.value_size);
+      return iterator;
    }
 
    void db_context_rocksdb::db_update_i64(int32_t itr, account_name payer, const char* value , size_t value_size) {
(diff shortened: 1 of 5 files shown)


libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -252,6 +254,7 @@ namespace eosio { namespace chain { namespace backing_store {
          table_store.table, old_payer, payer, name(key_store.primary),
          old_key_value.value->data(), old_key_value.value->size(), value, value_size);
       }
+      primary_iter_store.set_value(itr, value, value_size);
```
linhuang-blockone, line 257: Should this be primary_iter_store.set_value(iterator, pp.value, pp.value_size); ?

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -288,19 +291,12 @@ namespace eosio { namespace chain { namespace backing_store {
 
    int32_t db_context_rocksdb::db_get_i64(int32_t itr, char* value , size_t value_size) {
       const auto& key_store = primary_iter_store.get(itr);
```
linhuang-blockone, line 293: Is it possible that itr is not cached in the iter_store?

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: no, itr is the "handle" for the identified code/scope/table/primary_key, and update cannot change any of those parameters, to do that you would need to call db_rem_i64 code/scope/table/primary_key then call db_store_i64 code/scope/table/primary_key2

brianjohnson5972: It could be, value and pp.value and value_size and pp.value_size are just copies of each other. The payer_payload makes a copy and then when you call pp.to_payload() it returns a packed buffer with both the payer and value data in it. But since we have value and value_size passed in, we can just use it.

libraries/chain/backing_store/db_context_rocksdb.cpp
brianjohnson5972: Only if other code was written incorrectly. the itr is a "handle" to the code/scope/table/primary_key that would have been stored and returned by a call to db_find_i64 (or db_lower/upper_bound, db_next/previous_i64, etc)

brianjohnson5972:
This failed for replaying blocks, and was not able to identify the bug and after analysis, it was determined that this would provide little improvement since in every scenario that the cache would have data, the value would be read directly from the most recent session's cache, so closing.

PR #10006, opened 2021-02-04
use p2p address for duplicate connection resolution

huangminghuang:
## Change Description

This PR fixes the problem where two nodeos establish p2p connections to each other, send go away messages to each other at the same time and then both connections are dropped without any retry. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -422,8 +422,9 @@ namespace eosio {
    constexpr uint16_t proto_block_id_notify = 2;     // reserved. feature was removed. next net_version should be 3
    constexpr uint16_t proto_pruned_types = 3;        // supports new signed_block & packed_transaction types
    constexpr uint16_t heartbeat_interval = 4;        // supports configurable heartbeat interval
+   constexpr uint16_t dup_goaway_resolution = 5;     // support peer address based duplicate connection resolution
 
-   constexpr uint16_t net_version = heartbeat_interval;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2841,18 +2842,24 @@ namespace eosio {
...
+                     g_check_conn.unlock();
+                     g_conn.lock();
+                     auto c_time = last_handshake_sent.time;
+                     g_conn.unlock();
+                     if (msg.time + c_time <= check_time)
+                        continue;
+                  }
+                  else if (my_impl->p2p_address < msg.p2p_address) {
+                     continue; 
+                  }
```
brianjohnson5972, line 2862: I'd like a comment like "only the p2p node with the higher of the 2 addresses will decide which connection is a duplicate, so there is no chance for both connections to be closed"

brianjohnson5972 (approved):
Just add that one comment and it is good.

brianjohnson5972 (approved):

PR #10012, opened 2021-02-05
EPE-165: Improve logic for unlinkable blocks while sync'ing

rusty-block-one:
The current logic for handling unlinkable blocks while sync'ing closes
the node prematurely when the node is flooded with unlinkable blocks in
a short time window. This fix groups unlinkable blocks into time windows
in order to manage the flooding condition.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -551,7 +551,7 @@ namespace eosio {
       /// @param[in] window_size          The time, in microseconds, of the rejected block window
       /// @param[in] max_rejected_windows The max consecutive number of rejected block windows
       /// @note   Copy ctor is not allowed
-      block_status_monitor(fc::microseconds window_size = fc::microseconds(2*1000*1000),
+      explicit block_status_monitor(fc::microseconds window_size = fc::microseconds(2*1000),
             uint32_t max_rejected_windows = 13) :
          window_size_(window_size) {}
       block_status_monitor( const block_status_monitor& ) = delete;
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1970,10 +1970,10 @@ namespace eosio {
 
    // called from connection strand
    void sync_manager::rejected_block( const connection_ptr& c, uint32_t blk_num ) {
-      std::unique_lock<std::mutex> g( sync_mtx );
       c->block_status_monitor_.rejected();
       if( c->block_status_monitor_.max_events_violated()) {
          fc_wlog( logger, "block ${bn} not accepted from ${p}, closing connection", ("bn", blk_num)("p", c->peer_name()) );
+         std::unique_lock<std::mutex> g( sync_mtx );
          sync_last_requested_num = 0;
          sync_source.reset();
          g.unlock();
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -130,33 +130,6 @@ namespace eosio {
       }
    };
 
-   /// monitors the status of blocks as to whether a block is accepted (sync'd) or
-   /// rejected. It groups consecutive rejected blocks in a (configurable) time
-   /// window (rbw) and maintains a metric of the number of consecutive rejected block
-   /// time windows (rbws).
-   class block_status_monitor {
-   private:
-      fc::microseconds window_size_{2*1000*1000};  ///< rbw time interval (2ms)
-      fc::time_point   window_start_;              ///< The start of the recent rbw (0 implies not started)
-      uint32_t         events_{0};                 ///< The number of consecutive rbws
-      
-   public:
-      block_status_monitor(fc::microseconds window_size = fc::microseconds(2*1000*1000)) :
-         window_size_(window_size) {}
-      block_status_monitor( const block_status_monitor& ) = delete;
-      block_status_monitor( block_status_monitor&& ) = delete;
-      ~block_status_monitor() = default;
-      /// called when a block is accepted (sync_recv_block)
-      void accepted();
-      /// called when a block is rejected
-      void rejected();
-      /// returns number of consecutive rbws
-      auto events() const { return events_; }
-      /// assignment not allowed
-      block_status_monitor& operator=( const block_status_monitor& ) = delete;
-      block_status_monitor& operator=( block_status_monitor&& ) = delete;
-   };
-
    class sync_manager {
    private:
       enum stages {
@@ -172,7 +145,6 @@ namespace eosio {
       uint32_t       sync_req_span{0};
       connection_ptr sync_source;
       std::atomic<stages> sync_state{in_sync};
-      block_status_monitor block_status_monitor_;
 
    private:
       constexpr static auto stage_str( stages s );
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1900,7 +1961,8 @@ namespace eosio {
    // called from connection strand
    void sync_manager::rejected_block( const connection_ptr& c, uint32_t blk_num ) {
       std::unique_lock<std::mutex> g( sync_mtx );
-      if( ++c->consecutive_rejected_blocks > def_max_consecutive_rejected_blocks ) {
+      block_status_monitor_.rejected();
+      if( block_status_monitor_.events() > def_max_consecutive_rejected_blocks ) {
```
heifner, line 1965: This needs to be tracked per connection.

heifner: The def_max_consecutive_rejected_blocks should be part of block_status_monitor

plugins/net_plugin/net_plugin.cpp
```diff
@@ -130,6 +130,33 @@ namespace eosio {
       }
    };
 
+   /// monitors the status of blocks as to whether a block is accepted (sync'd) or
+   /// rejected. It groups consecutive rejected blocks in a (configurable) time
+   /// window (rbw) and maintains a metric of the number of consecutive rejected block
+   /// time windows (rbws).
+   class block_status_monitor {
+   private:
+      fc::microseconds window_size_{2*1000*1000};  ///< rbw time interval (2ms)
```
heifner, line 139: This is 2 seconds not 2ms

plugins/net_plugin/net_plugin.cpp
```diff
@@ -130,6 +130,33 @@ namespace eosio {
...
+      fc::time_point   window_start_;              ///< The start of the recent rbw (0 implies not started)
+      uint32_t         events_{0};                 ///< The number of consecutive rbws
+      
+   public:
+      block_status_monitor(fc::microseconds window_size = fc::microseconds(2*1000*1000)) :
```
heifner, line 144: explicit

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1511,6 +1539,39 @@ namespace eosio {
...
+
+      // if window time has not elapsed
+      const auto elapsed = now - window_start_;
+      if( elapsed < window_size_ ) {
+         return;
+      }
+
+      // if window time exactly expired
+      ++events_;
+      if( elapsed == window_size_ ) {
```
heifner, line 1566: This is extremely unlikely. Not sure I'm following the logic here.
accepted() should restart this monitor always.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
rusty-block-one: My thinking is that for consecutive windows, the end of one window is not necessarily the start of the next window.  To illustrate:
t0, the start of the first time window
te0, the time the first window closes
t1, the start of the next window
s, the window size
t, the time of the rejected block that closes the first window

te0 = t0 + s
t != te0
t1 = t instead of te0

The window_start is set to "0" in accepted().  However, the rejected() logic uses a window_start=0 as a signal to set the start of the next window.  My thinking is that this ensures each time window has at least 1 rejected block. My concern is that if the interval between rejected blocks is > than a multiple of the window size (s), a rejected block could result in the count being increased by more than one. This may be overthinking the problem.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1935,9 +2010,9 @@ namespace eosio {
          c->close( false, true );
          return;
       }
-      c->consecutive_rejected_blocks = 0;
       sync_update_expected( c, blk_id, blk_num, blk_applied );
       std::unique_lock<std::mutex> g_sync( sync_mtx );
+      c->block_status_monitor_.accepted();
```
heifner, line 2015: This can be outside the mutex. Move up.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1900,7 +1974,8 @@ namespace eosio {
    // called from connection strand
    void sync_manager::rejected_block( const connection_ptr& c, uint32_t blk_num ) {
       std::unique_lock<std::mutex> g( sync_mtx );
```
heifner, line 1976: This mutex can be moved down to right above `sync_last_requested_num`

plugins/net_plugin/net_plugin.cpp
heifner: I think this overcomplicates it. Seems to me it should just be if you are in same window then don't increment; otherwise increment.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -534,6 +533,48 @@ namespace eosio {
...
+      uint32_t         events_{0};                 ///< The number of consecutive rbws
+      const uint32_t   max_consecutive_rejected_windows_{13};
+      
+   public:
+      /// ctor
+      ///
+      /// @param[in] window_size          The time, in microseconds, of the rejected block window
+      /// @param[in] max_rejected_windows The max consecutive number of rejected block windows
+      /// @note   Copy ctor is not allowed
+      block_status_monitor(fc::microseconds window_size = fc::microseconds(2*1000*1000),
```
heifner, line 553: explicit

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1511,6 +1552,39 @@ namespace eosio {
...
+         events_ = 0;
+      }
+   }
+
+   void block_status_monitor::rejected() {
+      const auto now = fc::time_point::now();
+      // if starting a new window
+      if( window_start_ == fc::time_point() ) {
+         window_start_ = now;
+         return;
```
heifner, line 1568: Seems to me if you are starting a new window you should increment events_.
I think it would be clearer to use `if( events_ == 0 ) {` here.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Changed logic to make it clearer

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1511,6 +1549,39 @@ namespace eosio {
...
 
+   //-----------------------------------------------------------
+   void block_status_monitor::reset() {
+      in_accepted_state_ = true;
+      events_ = 0;
+   }
+
+   void block_status_monitor::accepted() {
+      if(in_accepted_state_) {
+         return;
```
heifner, line 1560: Not that performance is a real issue here, but this is likely to be faster to just set the values. Although the compiler has probably removed this if anyway. I think it is clearer to just set the values. But not a huge deal either way.

heifner (approved):

heifner:
Approved, but please don't merge until LRT has been run on this change. All net_plugin changes should have LRTs run before merge.

plugins/net_plugin/net_plugin.cpp
rusty-block-one: I appreciate your comment and thinking.  I debated internally the difference between just resetting the values and testing the current state. I'll change it to just setting the values. 
My internal debate centered around wondering what the most likely state on each call.  If the most likely state is accepted then the if test would reduce cpu cycles with a hiccup on when the state changed.  If the state is continually going back and forth, then removing the if test would most likely reduce cpu cycles.

PR #10049, opened 2021-02-12
add compile time option to enforce a chain id 📦

spoonincode:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
hmmm... actually, let me think this over a little more

matthewdarwin:
@spoonincode having a way to get the chain_id from a genesis.json file would be very convenient when starting a new chain. Doesn't have to be a separate tool.. could be baked into nodeos itself  `nodeos --print-chain-id-from-genesis --genesis-json=filename`

spoonincode:
The issue I have with baking these sort of things in to nodeos is that they aren't scriptable like I'd expect a unixy tool to be. Look at say `--print-build-info` or `--print-genesis-json` you can't just
```
BUILD_INFO=$(nodeos --print-build-info)
```
because 1) it goes out stderr, and 2) you get all sorts of other log trash when you run this command.

spoonincode:
Opening back up since I thought about it more.

heifner (approved):

PR #10050, opened 2021-02-15
Add more key-value pairs and operations to test contracts for KV snapshot tests

linhuang-blockone:
## Change Description

Add more key-values and operations to test contracts for KV snapshot test.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [X] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/snapshot_tests.cpp
+++ b/unittests/snapshot_tests.cpp
@@ -590,12 +590,24 @@ static const char kv_snapshot_wast[] = R"=====(
   (func $kv_get (import "env" "kv_get") (param i64 i32 i32 i32) (result i32))
   (func $kv_get_data (import "env" "kv_get_data") (param i32 i32 i32) (result i32))
   (func $kv_set (import "env" "kv_set") (param i64 i32 i32 i32 i32 i64) (result i64))
+  (func $kv_erase (import "env" "kv_erase") (param i64 i32 i32) (result i64))
   (memory 1)
   (func (export "apply") (param i64 i64 i64)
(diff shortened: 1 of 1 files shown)


brianjohnson5972 (changes requested):
"Verify the kv_object content is correct" I would take this as meaning adding verification of the resulting values in the kv_objects.  Since no verifications changed or were added, I'm not sure what we are getting with this change.

linhuang-blockone:
> "Verify the kv_object content is correct" I would take this as meaning adding verification of the resulting values in the kv_objects. Since no verifications changed or were added, I'm not sure what we are getting with this change.

I think verify_integrity_hash by the subtesters verifies all the states (including KV tables) before and after a snapshot is taken. With more keys in the test contract, write_snapshot would cover more execution paths. I am not sure how to verify kv_object directly at this testing level.

brianjohnson5972:
I meant since we have wasm code that could be broken and we would never know that anything is set, since the action processing code in the controller does not require that anything actually process.
I was meaning for you to use the kv_context to retrieve the key and value.  I know we don't do that for the db intrinsic one, but it uses a contract that is verified in another test, so we know that the database is updated, versus this test where right now, if the wasm was messed up so that it didn't do anything, this test would just keep saying that everything is good.

PR #10051, opened 2021-02-15
Add new loggers for dumping transaction trace

nickjjzhao:
## Change Description
Current loggers `transaction_success_tracing` and `transaction_failure_tracing` in file logging.json are created for displaying brief transaction trace information. New loggers that could be used to dump entire transaction trace or entire transaction for all trxs including failed are requested. 

Everywhere we currently log REJECTING or ACCEPTING we should log to new loggers with either the trace or the transaction. We should  log to new loggers with the transaction for the case of  an invalid signature.

New added loggers `transaction_trace_success` and `transaction_trace_failure` have default logging level `info` in file logging.json. To enable them, their logging levels need to be changed to `debug`.

Two examples of log output from new trace loggers:

1. creating an account named as "bob" succeeded  

> debug 2021-02-23T20:12:22.681 thread-0  producer_plugin.cpp:553       operator()           ] [TRX_TRACE] Block 44 for producer eosio is ACCEPTING tx: {"id":"f205f092723a1e62d27846df6168905fa6e504323353138404dd9ef4bfeee365","block_num":44,"block_time":"2021-02-23T20:12:23.000","producer_block_id":null,"receipt":{"status":"executed","cpu_usage_us":657,"net_usage_words":25},"elapsed":657,"net_usage":200,"scheduled":false,"action_traces":[{"action_ordinal":1,"creator_action_ordinal":0,"closest_unnotified_ancestor_action_ordinal":0,"receipt":{"receiver":"eosio","act_digest":"1c83ef3abe3ca4905d0aac83fb5f94f5ce55e92f35f0eca5df18a5ec547d950e","global_sequence":44,"recv_sequence":44,"auth_sequence":[["eosio",44]],"code_sequence":0,"abi_sequence":0},"receiver":"eosio","act":{"account":"eosio","name":"newaccount","authorization":[{"actor":"eosio","permission":"active"}],"data":{"creator":"eosio","name":"bob","owner":{"threshold":1,"keys":[{"key":"EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV","weight":1}],"accounts":[],"waits":[]},"active":{"threshold":1,"keys":[{"key":"EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV","weight":1}],"accounts":[],"waits":[]}},"hex_data":"0000000000ea30550000000000000e3d01000000010002c0ded2bc1f1305fb0faac5e6c03ee3a1924234985427b6167ca569d13df435cf0100000001000000010002c0ded2bc1f1305fb0faac5e6c03ee3a1924234985427b6167ca569d13df435cf01000000"},"context_free":false,"elapsed":220,"console":"","trx_id":"f205f092723a1e62d27846df6168905fa6e504323353138404dd9ef4bfeee365","block_num":44,"block_time":"2021-02-23T20:12:23.000","producer_block_id":null,"account_ram_deltas":[{"account":"bob","delta":2724}],"account_disk_deltas":[],"except":null,"error_code":null,"return_value_hex_data":""}],"account_ram_delta":null,"except":null,"error_code":null}

2. creating an account named as "bob" failed

> debug 2021-02-23T20:13:28.728 thread-0  producer_plugin.cpp:528       operator()           ] [TRX_TRACE] Block 176 for producer eosio is REJECTING tx: {"code":3050001,"name":"account_name_exists_exception","message":"Account name already exists","stack":[{"context":{"level":"error","file":"eosio_contract.cpp","line":93,"method":"apply_eosio_newaccount","hostname":"","thread_name":"thread-0","timestamp":"2021-02-23T20:13:28.727"},"format":"Cannot create account named ${name}, as that name is already taken","data":{"name":"bob"}},{"context":{"level":"warn","file":"eosio_contract.cpp","line":127,"method":"apply_eosio_newaccount","hostname":"","thread_name":"thread-0","timestamp":"2021-02-23T20:13:28.727"},"format":"","data":{"create":{"creator":"eosio","name":"bob","owner":{"threshold":1,"keys":[{"key":"EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV","weight":1}],"accounts":[],"waits":[]},"active":{"threshold":1,"keys":[{"key":"EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV","weight":1}],"accounts":[],"waits":[]}}}},{"context":{"level":"warn","file":"apply_context.cpp","line":143,"method":"exec_one","hostname":"","thread_name":"thread-0","timestamp":"2021-02-23T20:13:28.727"},"format":"pending console output: ${console}","data":{"console":""}}]}

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Documentation Additions
- [x] Documentation Additions

The following trace loggers are added:
- `transaction_trace_success` - fully detailed log that emits successful verdicts from relay nodes on the P2P network; enabling this logger can greatly slow down nodeos.
- `transaction_trace_failure` - fully detailed log that emits failed verdicts from relay nodes on the P2P network; enabling this logger can greatly slow down nodeos.


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -516,6 +519,11 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                           ("txid", trx->id())
                           ("why",std::get<fc::exception_ptr>(response)->what()));
                }
+
+               //Dump transaction trace
+               fc_dlog(_trx_trace_dump_log, "[TRX_TRACE][REJECTING tx] ${entire_trace}",
+                        ("entire_trace", fc::json::to_pretty_string(response)));
```
heifner, line 525: If we are going to dump them, might as well dump them completely expanded via the abi_serializer. Also try-catch so this does not throw.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -516,6 +519,11 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                           ("txid", trx->id())
```
heifner, line 519: I know you called this `transaction_trace_dump` but for the case of an invalid signature, probably should log the transaction.

plugins/producer_plugin/producer_plugin.cpp
heifner: Will need to add the variant type to `single_type_requires_abi_v` in abi_serializer for abi_serializer to expand correctly.

plugins/producer_plugin/producer_plugin.cpp
heifner: Actually everywhere we currently log REJECTING or ACCEPTING we should log to this logger with either the trace or the transaction.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -503,18 +514,39 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
                } else {
                   fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why} ",
                           ("txid", trx->id())
                           ("why",std::get<fc::exception_ptr>(response)->what()));
+
+                  //Dump transaction trace if logger `_trx_trace_dump_log` is enabled
+                  fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] [Speculative execution is REJECTING tx: ${txid} : ${why}] \n${entire_trace}",
+                          ("txid", trx->id())
+                          ("why",std::get<fc::exception_ptr>(response)->what())
+                          ("entire_trace", fc::json::to_pretty_string(entire_trace)));
```
heifner, line 548: Use abi_serializer

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1656,23 +1703,46 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
    size_t num_expired_other = 0;
    size_t orig_count = _unapplied_transactions.size();
    bool exhausted = !_unapplied_transactions.clear_expired( pending_block_time, deadline,
-                  [&num_expired_persistent, &num_expired_other, pbm = _pending_block_mode,
+                  [this, &num_expired_persistent, &num_expired_other, pbm = _pending_block_mode,
                    &chain, has_producers = !_producers.empty()]( const transaction_id_type& txid, trx_enum_type trx_type ) {
+
+            auto trx = this->_unapplied_transactions.get_trx(txid);
+            fc::variant entire_trx = fc::mutable_variant_object()("transaction", trx->packed_trx()->get_transaction());
```
heifner, line 1710: Only want to do this if debug logger enabled.
I don't see a need to put this under "transaction".
Use abi_serializer.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2276,6 +2346,13 @@ void producer_plugin_impl::produce_block() {
 void producer_plugin::log_failed_transaction(const transaction_id_type& trx_id, const char* reason) const {
    fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why}",
            ("txid", trx_id)("why", reason));
+
+   // Dump failed transaction if logger `_trx_trace_dump_log` is enabled
+   auto trx = my->_unapplied_transactions.get_trx(trx_id);
```
heifner, line 2351: Wrap in if debug level enabled.

heifner (changes requested):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3418,7 +3418,7 @@ read_only::abi_bin_to_json_result read_only::abi_bin_to_json( const read_only::a
 }
 
 read_only::get_required_keys_result read_only::get_required_keys( const get_required_keys_params& params )const {
-   transaction pretty_input;
+    transaction pretty_input;
```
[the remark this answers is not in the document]
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -503,18 +514,39 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
             next(response);
+
+            auto& cp = app().get_plugin<chain_plugin>();
+            auto ro_api = cp.get_read_only_api();
+            auto entire_trace = ro_api.get_entire_trx_trace(response);
+
             if (std::holds_alternative<fc::exception_ptr>(response)) {
                _transaction_ack_channel.publish(priority::low, std::pair<fc::exception_ptr, transaction_metadata_ptr>(std::get<fc::exception_ptr>(response), trx));
-               if (_pending_block_mode == pending_block_mode::producing) {
+                if (_pending_block_mode == pending_block_mode::producing) {
```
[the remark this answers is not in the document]
nickjjzhao: Fixed

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3430,6 +3430,20 @@ read_only::get_required_keys_result read_only::get_required_keys( const get_requ
...
+fc::variant read_only::get_entire_trx_trace(const std::variant<fc::exception_ptr, transaction_trace_ptr>& tracePtr )const {
+
+    fc::variant pretty_output;
+    try {
+        abi_serializer::to_variant(tracePtr, pretty_output,
+                                   make_resolver(this, abi_serializer::create_yield_function(abi_serializer_max_time)),
+                                   abi_serializer::create_yield_function(abi_serializer_max_time));
+    }catch(const fc::exception& e){
+        wlog("problem encountered while expanding transaction trace via the abi serializer:\n${details}",
+                     ("details",e.to_detail_string()));
```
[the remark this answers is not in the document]
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -503,18 +514,39 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
          auto send_response = [this, &trx, &chain, &next](const std::variant<fc::exception_ptr, transaction_trace_ptr>& response) {
             next(response);
+
+            auto& cp = app().get_plugin<chain_plugin>();
+            auto ro_api = cp.get_read_only_api();
+            auto entire_trace = ro_api.get_entire_trx_trace(response);
```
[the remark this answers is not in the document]
nickjjzhao: Fixed. Moved this interface to chain_plugin.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -473,14 +476,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
          auto trx_id = trx->id();
          boost::asio::post(_thread_pool->get_executor(), [self = this, future{std::move(future)}, persist_until_expired,
-                                                          next{std::move(next)}, trx_id]() mutable {
+                                                          next{std::move(next)}, trx{std::move(trx)}, trx_id]() mutable {
```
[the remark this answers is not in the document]
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -473,14 +476,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx_id]() mutable {
-                  auto exception_handler = [&next, trx_id](fc::exception_ptr ex) {
+               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx{std::move(trx)}, trx_id]() mutable {
+                  auto exception_handler = [&next, &trx, trx_id](fc::exception_ptr ex) {
                     fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why} ",
                             ("txid", trx_id)("why",ex->what()));
                      next(ex);
+
+                    //Dump failed transaction if logger `_trx_trace_dump_log` is enabled
+                    fc::variant entire_trx = fc::mutable_variant_object()("transaction", *trx);
```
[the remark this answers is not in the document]
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -503,18 +514,39 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
                         ("txid", trx->id())
                         ("why",std::get<fc::exception_ptr>(response)->what()));
+
+                  //Dump transaction trace if logger `_trx_trace_dump_log` is enabled
+                  fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] [Block ${block_num} for producer ${prod} is REJECTING tx: ${txid} : ${why}] \n${entire_trace} ",
+                          ("block_num", chain.head_block_num() + 1)
+                          ("prod", get_pending_block_producer())
+                          ("txid", trx->id())
+                          ("why",std::get<fc::exception_ptr>(response)->what())
+                          ("entire_trace", fc::json::to_pretty_string(entire_trace)));
```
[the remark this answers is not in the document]
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3484,6 +3483,19 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+
+fc::variant chain_plugin::get_entire_trx_trace(const std::variant<std::shared_ptr<transaction>, transaction_trace_ptr>& obj )const {
+
+    fc::variant pretty_output;
+    auto ro_api = get_read_only_api();
+    try {
+        abi_serializer::to_variant(obj, pretty_output,
+                                   chain_apis::make_resolver(&ro_api, abi_serializer::create_yield_function(get_abi_serializer_max_time())),
+                                   abi_serializer::create_yield_function(get_abi_serializer_max_time()));
+    }EOS_RETHROW_EXCEPTIONS(chain::packed_transaction_type_exception, "Invalid packed transaction trace")
```
heifner, line 3495: I think you might have misunderstood my previous PR comments. This method should never throw. Instead of rethrow it should just assign directly to `pretty_output`.

heifner: Also I don't think abi_serializer will actually work as is like this by passing in a `std::variant`. You could pull out of the particular type and pass that in OR have the method take two args instead of a variant OR have two diff methods OR make it a template.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -473,14 +476,24 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
 
          auto trx_id = trx->id();
          boost::asio::post(_thread_pool->get_executor(), [self = this, future{std::move(future)}, persist_until_expired,
-                                                          next{std::move(next)}, trx_id]() mutable {
+                                                          next{std::move(next)}, &trx, trx_id]() mutable {
```
heifner, line 479: `trx` is on the stack you do not want to pass a reference to it here as it will not live as long as the `post` call.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -473,14 +476,24 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
             if( future.valid() ) {
                future.wait();
-               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx_id]() mutable {
-                  auto exception_handler = [&next, trx_id](fc::exception_ptr ex) {
+               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, &trx, trx_id]() mutable {
+                  auto exception_handler = [self, &next, &trx, trx_id](fc::exception_ptr ex) {
```
heifner, line 483: Same issue here with `&trx`

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed

plugins/chain_plugin/chain_plugin.cpp
nickjjzhao: Fixed

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3484,6 +3483,31 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+
+fc::variant chain_plugin::get_entire_trx_trace(const std::variant<fc::exception_ptr, transaction_trace_ptr, std::shared_ptr<transaction>>& obj)const {
+
+    fc::variant pretty_output;
+    auto ro_api = get_read_only_api();
+    try {
+        if (std::holds_alternative<fc::exception_ptr>(obj)) {
+            abi_serializer::to_variant(std::get<fc::exception_ptr>(obj), pretty_output,
+                                       chain_apis::make_resolver(&ro_api, abi_serializer::create_yield_function(get_abi_serializer_max_time())),
+                                       abi_serializer::create_yield_function(get_abi_serializer_max_time()));
```
heifner, line 3495: No need for fc::exception to be converted by abi_serializer. Just assignment is all that is needed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -477,10 +480,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
                future.wait();
                app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx_id]() mutable {
-                  auto exception_handler = [&next, trx_id](fc::exception_ptr ex) {
+                  auto exception_handler = [self, &next, trx_id](fc::exception_ptr ex) {
                     fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why} ",
                             ("txid", trx_id)("why",ex->what()));
                      next(ex);
+
+                    if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                        if (self->_unapplied_transactions.get_trx(trx_id)) {
```
heifner, line 489: Many/most of the time the transaction will not be in the unapplied_transactions queue. I think you are better off capturing a copy of `trx`. You could avoid capturing `trx_id` and just get that from `trx`.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3484,6 +3483,31 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
 }
 
 } // namespace chain_apis
+
+fc::variant chain_plugin::get_entire_trx_trace(const std::variant<fc::exception_ptr, transaction_trace_ptr, std::shared_ptr<transaction>>& obj)const {
```
heifner, line 3487: This method is misnamed now that it does 3 diff things.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -477,10 +480,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                            //Dump transaction
+                            std::shared_ptr<transaction> transaction_ptr = std::make_shared<transaction>(self->_unapplied_transactions.get_trx(trx_id)->packed_trx()->get_transaction());
+                            auto entire_trx = self->chain_plug->get_entire_trx_trace(transaction_ptr);
+                            fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] [Speculative execution is REJECTING tx: ${txid} : ${why}]: \n${entire_trx} ",
+                                    ("txid", trx_id)
+                                    ("why",ex->what())
+                                    ("entire_trx", fc::json::to_pretty_string(entire_trx)));
```
heifner, line 496: I think it is worth asking if file support is needed here. Having all this logging mixed in with the rest of the logging seems like it would make it rather unusable. I can't imagine wanting to_pretty_string unless this is going to a file.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1656,23 +1711,52 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
...
             if( trx_type == trx_enum_type::persisted ) {
                if( pbm == pending_block_mode::producing ) {
                   fc_dlog(_trx_failed_trace_log,
                            "[TRX_TRACE] Block ${block_num} for producer ${prod} is EXPIRING PERSISTED tx: ${txid}",
                            ("block_num", chain.head_block_num() + 1)("txid", txid)
                            ("prod", chain.is_building_block() ? chain.pending_block_producer() : name()) );
+
+                  if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                      // Dump transaction
+                      std::shared_ptr<transaction> transaction_ptr = std::make_shared<transaction>(this->_unapplied_transactions.get_trx(txid)->packed_trx()->get_transaction());
```
heifner, line 1726: Should not make a copy of the transaction just to log it. Worse case you can add another variant for `transaction`.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
heifner: Also the `\n` in the logging also messes with "normal" logging which is one statement per line.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Thanks! Is it possible that function `get_trx(trx_id)` could return  a `nullptr`, i.e., trx_id is valid but there is no corresponding trx in the queue?

heifner: yes

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: I tried passing `transaction` instead of `shared_ptr<transaction>` to the interface `get_entire_trx_trace()`, which had a variant for `transaction`, but there was compiling error `error: no viable conversion from ‘const eosio::chain::transaction’ to ‘const std::variant<fc::exception_ptr, transaction_trace_ptr, transaction>’` 

Will try agin.

heifner: That is likely because `transaction` constructor is explicit which is a good thing because you would not want to make a copy just so you can call this. At this point, I would give up on using a std::variant.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Ok. I agree.  I will delete all the call `to_pretty_string()` and `\n`. I though it would be more helpful to look at formatted transaction traces in the log while debugging.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: I will try making three diff methods to replace current `get_entire_trx_trace()`. I thought about making it a template and implementing it in `chain_plugin.hpp` but this interface needs to call `make_resolver` that is defined in `chain_plugin.cpp'.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/chain_plugin/chain_plugin.cpp
nickjjzhao: Fixed.

plugins/chain_plugin/chain_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3484,6 +3483,34 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
 }
 
 } // namespace chain_apis
+
+fc::variant chain_plugin::get_entire_trx_trace(const transaction_trace_ptr& trx_trace )const {
+
+    fc::variant pretty_output;
+    auto ro_api = get_read_only_api();
```
heifner, line 3490: I think you should modify `make_resolver` to take a `controller&` or `chainbase&` so you don't have to make a copy of `read_only` on every call.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -471,16 +474,21 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
 
-         auto trx_id = trx->id();
          boost::asio::post(_thread_pool->get_executor(), [self = this, future{std::move(future)}, persist_until_expired,
-                                                          next{std::move(next)}, trx_id]() mutable {
+                                                          next{std::move(next)}, trx]() mutable {
             if( future.valid() ) {
                future.wait();
-               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx_id]() mutable {
-                  auto exception_handler = [&next, trx_id](fc::exception_ptr ex) {
+               app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx]() mutable {
```
heifner, line 481: This `trx` can be moved into to lambda.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -471,16 +474,21 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                  auto exception_handler = [self, &next, trx](fc::exception_ptr ex) {
```
heifner, line 482: This `trx` can be moved into the lambda.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -471,16 +474,21 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
                     fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why} ",
-                            ("txid", trx_id)("why",ex->what()));
+                            ("txid", trx->id())("why",ex->what()));
                      next(ex);
+
+                    if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                        //Dump transaction
```
heifner, line 488: Remove comment

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -471,16 +474,21 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                        auto entire_trx = self->chain_plug->get_entire_trx(trx->get_transaction());
+                        fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] ${entire_trx}", ("entire_trx", entire_trx));
```
heifner, line 490: You should add "REJECTING" to the log output.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -516,6 +525,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                           ("txid", trx->id())
                           ("why",std::get<fc::exception_ptr>(response)->what()));
                }
+
+               if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                   //Dump transaction trace
```
heifner, line 530: Remove comment

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -516,6 +525,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                   auto entire_trace = std::get<fc::exception_ptr>(response);
+                   fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] ${entire_trace} ", ("entire_trace", entire_trace));
```
heifner, line 532: Should add "Speculative execution is REJECTING tx:"

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -527,6 +542,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                   fc_dlog(_trx_successful_trace_log, "[TRX_TRACE] Speculative execution is ACCEPTING tx: ${txid}",
                           ("txid", trx->id()));
                }
+
+               if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                   //Dump transaction trace
```
heifner, line 547: Remove comment

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -527,6 +542,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                   auto entire_trace = chain_plug->get_entire_trx_trace(std::get<transaction_trace_ptr>(response) );
+                   fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] ${entire_trace}", ("entire_trace", entire_trace));
```
heifner, line 549: Add "Speculative execution is ACCEPTING tx:"

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1656,8 +1678,9 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
    size_t num_expired_other = 0;
    size_t orig_count = _unapplied_transactions.size();
    bool exhausted = !_unapplied_transactions.clear_expired( pending_block_time, deadline,
-                  [&num_expired_persistent, &num_expired_other, pbm = _pending_block_mode,
+                  [this, &num_expired_persistent, &num_expired_other, pbm = _pending_block_mode,
```
heifner, line 1681: Instead of capturing `this`, how about just capturing `&chain_plug`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1667,12 +1690,25 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
                } else {
                   fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is EXPIRING PERSISTED tx: ${txid}", ("txid", txid));
                }
+
+               if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                   // Dump transaction
```
heifner, line 1695: Remove comment

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1667,12 +1690,25 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
...
+                   auto entire_trx = chain_plug->get_entire_trx(this->_unapplied_transactions.get_trx(txid)->packed_trx()->get_transaction());
```
heifner, line 1696: Instead of looking it up in `_unapplied_transactions`, change the callback on `clear_expired` to take a `const packed_transaction_ptr&` instead of a transaction id.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1667,12 +1690,25 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
...
+                   fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] ${entire_trx}", ("entire_trx", entire_trx));
```
heifner, line 1697: Add "EXPIRING PERSISTED tx:" to log output.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1667,12 +1690,25 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
...
             } else {
                if (has_producers) {
                   fc_dlog(_trx_failed_trace_log,
                         "[TRX_TRACE] Node with producers configured is dropping an EXPIRED transaction that was PREVIOUSLY ACCEPTED : ${txid}",
                         ("txid", txid));
+
+                  if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)) {
+                      // Dump transaction
+                      auto entire_trx = chain_plug->get_entire_trx(this->_unapplied_transactions.get_trx(txid)->packed_trx()->get_transaction());
+                      fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] ${entire_trx}", ("entire_trx", entire_trx));
```
heifner, line 1710: Same comments as above.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2276,6 +2312,18 @@ void producer_plugin_impl::produce_block() {
 void producer_plugin::log_failed_transaction(const transaction_id_type& trx_id, const char* reason) const {
    fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why}",
            ("txid", trx_id)("why", reason));
+
+   if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)){
+       transaction_metadata_ptr trx_meta_ptr = my->_unapplied_transactions.get_trx(trx_id);
```
heifner, line 2317: It will never be in the `unapplied_transactions`. Modify `log_failed_transaction` to take a `const packed_transaction_ptr&`

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: When is a transaction put into the `unapplied_transaction_queue`?

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: I guess `trx{std::move(trx)` is not the right way to do this as per your previous comments `Can't move as trx is a const&`. Could you be a bit more specific?

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Which version looks better or none of them?
1) `"[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why}; ${entire_trx} "`
2)`"[TRX_TRACE] Speculative execution is REJECTING tx: ${entire_trx} "`
3)`"[TRX_TRACE] [REJECTING] ${entire_trx}"`

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Ok. I am reverting to previous version of changes, i.e., `if` and `else` have different log messages. 

The log message in `if` block would be `"[TRX_TRACE] Block ${block_num} for producer ${prod} is REJECTING tx: ${entire_trace}` , and the log message in `else` block would be `[TRX_TRACE] Speculative execution is REJECTING tx: ${entire_trace}`. Does it look ok?

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Yes, that would be much better.

plugins/producer_plugin/producer_plugin.cpp
heifner: On line 478 above the `trx` is a `const&`. Here on line 481 it is not a `const` and can be moved.

plugins/producer_plugin/producer_plugin.cpp
heifner: See calls to `unapplied_transaction_queue` `add_forked`, `add_aborted`, `add_persisted`

nickjjzhao: I did search `unapplied_transactions` in the code to see where a trx could be in the queue, but was not sure. Thanks!

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Got error `error: 'chain_plug' in capture list does not name a variable` after replacing `this` with `&chain_plug`. Did I miss somthing?

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Fixed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2756,7 +2756,7 @@ namespace eosio {
          fc::raw::unpack( ds, trx_id );
          if( trx_id ) {
             if (trx_in_progress_sz > def_max_trx_in_progress_size) {
-               report_dropping_trx(*trx_id, trx_in_progress_sz);
+               report_dropping_trx(ptr, trx_in_progress_sz);
```
heifner, line 2759: `ptr` is null here. For this case the only thing you can report is the transaction id. 
Seems you need either two log_failed_transactions or have it take a transaction id and a packed_transaction_ptr which might be null.

plugins/producer_plugin/producer_plugin.cpp
heifner: `[chain_plug = chain_plug, ..]` should work.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2273,9 +2327,19 @@ void producer_plugin_impl::produce_block() {
...
-           ("txid", trx_id)("why", reason));
+           ("txid", packed_trx_ptr->id())("why", reason));
+
+   if (_trx_trace_dump_log.is_enabled(fc::log_level::debug)){
+       if (packed_trx_ptr) {
+           auto entire_trx = my->chain_plug->get_entire_trx(packed_trx_ptr->get_transaction());
+           fc_dlog(_trx_trace_dump_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${entire_trx}", ("entire_trx", entire_trx));
+       } else {
+           fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : ${why}",
+                   ("txid", packed_trx_ptr->id())("why", reason));
```
heifner, line 2340: `packed_trx_ptr` is null here.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
nickjjzhao: Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Sorry,  I missed this one. Fixed.

plugins/producer_plugin/producer_plugin.cpp
nickjjzhao: Yes. It works. Why didn't `&chain_plug` work here?

nickjjzhao: Fixed.

heifner: "Class members cannot be captured explicitly by a capture without initializer" https://en.cppreference.com/w/cpp/language/lambda

nickjjzhao: Thanks!

plugins/chain_plugin/chain_plugin.cpp
nickjjzhao: Fixed.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3484,6 +3479,32 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
 } // namespace chain_apis
+
+fc::variant chain_plugin::get_entire_trx_trace(const transaction_trace_ptr& trx_trace )const {
+
+    fc::variant pretty_output;
+    try {
+        abi_serializer::to_variant(trx_trace, pretty_output,
+                                   chain_apis::make_resolver(chain(), abi_serializer::create_yield_function(get_abi_serializer_max_time())),
+                                   abi_serializer::create_yield_function(get_abi_serializer_max_time()));
+     } catch (chain::abi_exception&) {
```
heifner, line 3490: In general this should catch everything. However, I think there are a few FC exceptions that also might be thrown. Since these are called from code we never want to throw, I think `...` would be better.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3484,6 +3479,32 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+    return pretty_output;
+}
+
+fc::variant chain_plugin::get_entire_trx(const transaction& trx) const {
+    fc::variant pretty_output;
+    try {
+        abi_serializer::to_variant(trx, pretty_output,
+                                   chain_apis::make_resolver(chain(), abi_serializer::create_yield_function(get_abi_serializer_max_time())),
+                                   abi_serializer::create_yield_function(get_abi_serializer_max_time()));
+    } catch (chain::abi_exception&) {
```
heifner, line 3502: In general this should catch everything. However, I think there are a few FC exceptions that also might be thrown. Since these are called from code we never want to throw, I think `...` would be better.

heifner (changes requested):

plugins/chain_plugin/chain_plugin.cpp
nickjjzhao: Changed.

plugins/chain_plugin/chain_plugin.cpp
nickjjzhao: Changed.

heifner (approved):

PR #10059, opened 2021-02-17
Cleos print error message from server when parse json failed

softprofe:
## Change Description

Start 10059 description ......

EPE 436

In some case, the server can't give Json string to caller, it give a html message to show the possible reason why the request failed. but the do_http_call function still try to parse the string as json string and throw an json parse error message(it is not unstandable), now just print the original string from server so easy to understand what happned.

However, do_http_call  is trying to parse string which is supposed to be handled by higher layer function, that is not usual.

End 10059 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -258,7 +258,8 @@ namespace eosio { namespace client { namespace http {
       response_result = fc::json::from_string(re);
    } catch(...) {
       // re reported below if print_response requested
-      std::cerr << "Error happens, below is the message received from server:" << std::endl << re << std::endl;
+      std::cerr << "Error happens, below is the message received from server:" << std::endl;
+      print_response = true;
    }
 
    if( print_response ) {
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -258,6 +258,7 @@ namespace eosio { namespace client { namespace http {
       response_result = fc::json::from_string(re);
    } catch(...) {
       // re reported below if print_response requested
+      std::cerr << "Error happens, below is the message received from server:" << std::endl << re << std::endl;
    }
 
    if( print_response ) {
--- a/programs/cleos/httpc.cpp
+++ b/programs/cleos/httpc.cpp
@@ -258,7 +258,6 @@ namespace eosio { namespace client { namespace http {
       response_result = fc::json::from_string(re);
    } catch(...) {
       // re reported below if print_response requested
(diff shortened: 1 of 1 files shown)


programs/cleos/httpc.cpp
```diff
@@ -258,6 +258,7 @@ namespace eosio { namespace client { namespace http {
       response_result = fc::json::from_string(re);
    } catch(...) {
       // re reported below if print_response requested
+      std::cerr << "Error happens, below is the message received from server:" << std::endl << re << std::endl;
```
heifner, line 261: Seems like `print_response = true;` would be a better way to handle this if you want to always `cerr` the `re`.

heifner (changes requested):

programs/cleos/httpc.cpp
softprofe: Ok

programs/cleos/httpc.cpp
```diff
@@ -258,6 +258,8 @@ namespace eosio { namespace client { namespace http {
       response_result = fc::json::from_string(re);
    } catch(...) {
       // re reported below if print_response requested
+      std::cerr << "Error happens, below is the message received from server:" << std::endl;
```
heifner, line 261: I don't think this cerr is needed. Also you should update the comment.

softprofe: That is for let user know error happened.   only print response don't means an error happened.

heifner: It is not necessarily an error, right? This only means it is not json.

softprofe: Ok. remove that line.

softprofe: The problem of this function do_http_call is layer chaos, a function in http layer try to handle higher layer thing, that caused you don't even know a exception throw is an error or not.

heifner (changes requested):
Please update PR description with standard template

heifner (changes requested):
I don't think the doc section in the PR description is for this PR

softprofe:
> I don't think the doc section in the PR description is for this PR

 Copied from another one, forget to delete the doc section. now it's gone.

heifner (approved):

PR #10087, opened 2021-02-24
add first_block_num for get_info command

softprofe:
## Change Description

Start 10087 description ......

EPE 737

add first_block_num to the result of get_info command, the first block num is fetched from block_log_imp object, and then fill this number to get info result.

End 10087 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -120,6 +120,7 @@ class read_only {
       string                               server_version;
       chain::chain_id_type                 chain_id;
       uint32_t                             head_block_num = 0;
+      uint32_t                             first_block_num = 0;
```
heifner, line 123: Did you test with an old `cleos`? I think you have to add this as an `std::optional` to the end.

libraries/chain/controller.cpp
```diff
@@ -2340,6 +2340,11 @@ const controller::config& controller::get_config()const
    return my->conf;
 }
 
+const uint32_t controller::get_first_block_num()const
```
heifner, line 2343: Why `const uint32_t` ?

heifner (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
softprofe: will test old cleos.

heifner: Actually I think it is old nodeos with new cleos that will break.

libraries/chain/controller.cpp
softprofe: That is because many functions in controller.cpp has const return value, just keep new function looks similar with other functions.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
softprofe: After test, with old cleos, it works good only no first_block_num as below:
$ ./cleos get info
{
  "server_version": "c5746b76",
  "chain_id": "8a34ec7df1b8cd06ff4a8abbaa7cc50300823350cadc59ab296cb00d104d2b8f",
  "head_block_num": 71774,
  "last_irreversible_block_num": 71773,
  "last_irreversible_block_id": "0001185d8e76bd53352ba97efaf74f94bf4801e4dbe4c460008697ff748fb58a",
  "head_block_id": "0001185e94a9337fe51d885f5b74b5e4ac50932329b367508792c36f974be62a",
  "head_block_time": "2021-02-26T01:49:12.000",
  "head_block_producer": "eosio",
  "virtual_block_cpu_limit": 200000000,
  "virtual_block_net_limit": 1048576000,
  "block_cpu_limit": 199900,
  "block_net_limit": 1048576,
  "server_version_string": "v2.1.0-rc1",
  "fork_db_head_block_num": 71774,
  "fork_db_head_block_id": "0001185e94a9337fe51d885f5b74b5e4ac50932329b367508792c36f974be62a",
  "server_full_version_string": "v2.1.0-rc1-c5746b76b610dbf336cc61db78196deeda373d5d-dirty",
  "last_irreversible_block_time": "2021-02-26T01:49:11.500"
}
while using new cleos, the first_block_num will appear as below:
$ ./cleos get info
{
  "server_version": "c5746b76",
  "chain_id": "8a34ec7df1b8cd06ff4a8abbaa7cc50300823350cadc59ab296cb00d104d2b8f",
  "head_block_num": 71883,
  "first_block_num": 1,
  "last_irreversible_block_num": 71882,
  "last_irreversible_block_id": "000118ca6ec45dd0064cc97f5d53bef4748f25642a03db735d455e674f93ffe6",
  "head_block_id": "000118cbe3c8577c571159f8ab333e94046fa66a6266f77b9816cebdb5625cf4",
  "head_block_time": "2021-02-26T01:50:06.500",
  "head_block_producer": "eosio",
  "virtual_block_cpu_limit": 200000000,
  "virtual_block_net_limit": 1048576000,
  "block_cpu_limit": 199900,
  "block_net_limit": 1048576,
  "server_version_string": "v2.1.0-rc1",
  "fork_db_head_block_num": 71883,
  "fork_db_head_block_id": "000118cbe3c8577c571159f8ab333e94046fa66a6266f77b9816cebdb5625cf4",
  "server_full_version_string": "v2.1.0-rc1-c5746b76b610dbf336cc61db78196deeda373d5d-dirty",
  "last_irreversible_block_time": "2021-02-26T01:50:06.000"
}

libraries/chain/controller.cpp
heifner: There are many methods on controller that return a `uint32_t`, not a one of them return a `const uint32_t`.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
heifner: What about new cleos with version 2.0.x of nodeos?

libraries/chain/controller.cpp
softprofe: Now it is uint32_t,  const is gone. uint32_t don't need const.  using const is to protect a reference return.  really don't need const.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
softprofe: what's the reason the user want update cleos but not update nodeos?

heifner: If you interact with a public nodeos API endpoint, you have no control on what version they are running. New cleos versions have to be backward compatible with old nodeos.

softprofe: New cleos with old nodeos doesn't break, but it print "first_block_num": 0,  it will make user confused, so let me try solve it.

softprofe: After tested,  compatible with old nodeos now.

heifner (approved):

PR #10093, opened 2021-02-25
cache key values in iterator store in db_context_rocksdb

linhuang-blockone:
## Change Description

https://blockone.atlassian.net/browse/EPE-786

Cache primary key values when first retrieving them in `db_find_i64` and iterator functions.
Use them in `db_get_i64`, `db_update_i64`, and `db_remove_i64`.

This PR is based on Brian's https://github.com/EOSIO/eos/pull/9997. Thanks Brian. It fixes a bug in the original PR, and makes `db_update_i64` and `db_move_i64` to use cached values too.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/backing_store/db_context_rocksdb.cpp
+++ b/libraries/chain/backing_store/db_context_rocksdb.cpp
@@ -200,16 +200,16 @@ namespace eosio { namespace chain { namespace backing_store {
 
       const unique_table t { receiver, scope_name, table_name };
       const auto table_ei = primary_iter_store.cache_table(t);
-      return primary_iter_store.add(primary_key_iter(table_ei, id, payer));
+      const auto store_itr = primary_iter_store.add(primary_key_iter(table_ei, id, payer));
+      primary_iter_store.set_value(store_itr, pp.value, pp.value_size);
+      return store_itr;
    }
 
    void db_context_rocksdb::db_update_i64(int32_t itr, account_name payer, const char* value , size_t value_size) {
       const auto& key_store = primary_iter_store.get(itr);
       const auto& table_store = primary_iter_store.get_table(key_store);
       EOS_ASSERT( table_store.contract == receiver, table_access_violation, "db access violation" );
-      const auto old_key_value = get_primary_key_value(receiver, table_store.scope, table_store.table, key_store.primary);
-
-      EOS_ASSERT( old_key_value.value, db_rocksdb_invalid_operation_exception,
+      EOS_ASSERT( key_store.value, db_rocksdb_invalid_operation_exception,
                   "invariant failure in db_update_i64, iter store found to update but nothing in database");
 
       // copy locally, since below the key_store memory will be changed
@@ -218,11 +218,11 @@ namespace eosio { namespace chain { namespace backing_store {
          payer = old_payer;
       }
 
-      const payer_payload old_pp{*old_key_value.value};
-      const auto old_value_actual_size = old_pp.value_size;
+      const auto old_value_actual_size = key_store.value->size();
 
       const payer_payload pp{payer, value, value_size};
-      set_value(old_key_value.full_key, pp);
+      prefix_bundle old_key = get_primary_slice_in_primaries(table_store.contract, table_store.scope,table_store.table, key_store.primary);
+      set_value(old_key.full_key, pp);
 
       std::string event_id;
       auto dm_logger = context.control.get_deep_mind_logger();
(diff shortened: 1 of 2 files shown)


libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -772,7 +775,14 @@ namespace eosio { namespace chain { namespace backing_store {
          primary_key = key;
       }
 
-      return primary_iter_store.add(primary_key_iter(table_ei, *primary_key, found_payer));
+      const auto store_itr = primary_iter_store.add(primary_key_iter(table_ei, *primary_key, found_payer));
+
+      // value pointed by the iterator might have changed. need to cache from the      source.
```
brianjohnson5972, line 780: We shouldn't have to go through constructing the key again and retrieving the value, since session_iter points to the key and value already.

Should change line769 to:
payer_payload pp(*(session_iter).second);
const account_name found_payer = pp.payer:

and change lines 780-783 to be:
primary_iter_store.set_value(store_itr, pp.value, pp.value_size);

brianjohnson5972 (changes requested):

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -708,8 +704,15 @@ namespace eosio { namespace chain { namespace backing_store {
                   ("func", calling_func));
 
       const account_name found_payer = payer_payload(*(*session_iter).second).payer;
+      const auto store_itr = primary_iter_store.add(primary_key_iter(table_ei, found_key, found_payer));
+
+      // value pointed by the iterator might have changed. need to cache from the source.
```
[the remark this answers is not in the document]
linhuang-blockone: We cannot reuse the value pointed by session_iter. It causes "can't find interest_token_name" assertion failure. This is one of the reasons the original solution failed.

The same response applies to the next comment.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -218,11 +218,11 @@ namespace eosio { namespace chain { namespace backing_store {
          payer = old_payer;
       }
 
-      const payer_payload old_pp{*old_key_value.value};
-      const auto old_value_actual_size = old_pp.value_size;
+      const auto old_value_actual_size = key_store.value->size();
 
       const payer_payload pp{payer, value, value_size};
-      set_value(old_key_value.full_key, pp);
+      prefix_bundle old_key = get_primary_slice_in_primaries(table_store.contract, table_store.scope,table_store.table, key_store.primary);
```
brianjohnson5972, line 224: get_primary_slice_in_primaries creates an extra key that we don't use, so we could create a new method that just creates the key_bundle instead of a prefix_bundle.

libraries/chain/backing_store/db_context_rocksdb.cpp
```diff
@@ -271,36 +271,31 @@ namespace eosio { namespace chain { namespace backing_store {
...
          db_context::log_row_remove(*dm_logger, context.get_action_id(), table_store.contract, table_store.scope,
-                                      table_store.table, old_payer, name(key_store.primary), pp.value,
-                                      pp.value_size);
+                                      table_store.table, old_payer, name(key_store.primary), key_store.value->data(),
+                                      key_store.value->size());
       }
 
-      current_session.erase(old_key_value.full_key);
-      primary_lookup.remove_table_if_empty(old_key_value.full_key);
+      prefix_bundle old_key = get_primary_slice_in_primaries(table_store.contract, table_store.scope,table_store.table, key_store.primary);
```
brianjohnson5972, line 282: other comment applies here as well

brianjohnson5972 (approved):

libraries/chain/backing_store/db_context_rocksdb.cpp
linhuang-blockone: Thanks for your suggestion. Will create a new method in a future PR in the performance activity.

libraries/chain/backing_store/db_context_rocksdb.cpp
linhuang-blockone: Will do it in a future PR.

PR #10109, opened 2021-03-04
Change blockvault client plugin initialization log message

huangminghuang:
## Change Description

This PR changes how blockvault client plugin log message during initialization. If the plugin is not enabled, no log message would be shown. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
+++ b/plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
@@ -32,7 +32,6 @@ void blockvault_client_plugin::set_program_options(options_description&, options
 }
 
 void blockvault_client_plugin::plugin_initialize(const variables_map& options) {
-   ilog("initializing blockvault_client plugin");
 #ifdef HAS_PQXX
    try {
       if (options.count("block-vault-backend")) {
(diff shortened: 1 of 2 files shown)


plugins/blockvault_client_plugin/blockvault_client_plugin.cpp
```diff
@@ -44,6 +43,9 @@ void blockvault_client_plugin::plugin_initialize(const variables_map& options) {
             elog("unknown block-vault-backend option, skipping it");
          }
       }
+      else {
+         ilog("no block-vault-backend option is specified, skipping it");
+      }
```
heifner, line 48: Ideally the plugin would not initialize at all if not specified. But chain_plugin requires blockvault_client_plugin currently. Would be great if that was not the case.

brianjohnson5972: If we are going to make that change, we need to do that in a Jira task just for that specifically.

brianjohnson5972 (approved):

brianjohnson5972 (changes requested):
since it could be a simple change, let's at least look

brianjohnson5972 (approved):

PR #10111, opened 2021-03-05
amqp shutdown hang

heifner:
## Change Description

- The `amqp_trace_plugin` was blocking on shutdown calling `amqp::stop()` since the channel was unusable and would not call `onFinalize` or `onError`. Add a check for if channel is usable. Also added an `onError` which should not be necessary but figure it can't hurt anything.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/amqp/include/eosio/amqp/amqp_handler.hpp
+++ b/libraries/amqp/include/eosio/amqp/amqp_handler.hpp
@@ -99,10 +99,18 @@ public:
          std::promise<void> stop_promise;
          auto future = stop_promise.get_future();
          boost::asio::post( *handler_->amqp_strand(), [&]() {
-            if( channel_ ) {
+            if( channel_ && channel_->usable() ) {
                auto& cb = channel_->close();
                cb.onFinalize( [&]() {
-                  stop_promise.set_value();
+                  try {
+                     stop_promise.set_value();
+                  } catch(...) {}
+               } );
+               cb.onError( [&](const char* message) {
+                  elog( "Error on stop: ${m}", ("m", message) );
+                  try {
+                     stop_promise.set_value();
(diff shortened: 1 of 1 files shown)


libraries/amqp/include/eosio/amqp/amqp_handler.hpp
```diff
@@ -99,10 +99,18 @@ class amqp {
...
          auto future = stop_promise.get_future();
          boost::asio::post( *handler_->amqp_strand(), [&]() {
-            if( channel_ ) {
+            if( channel_ && channel_->usable() ) {
                auto& cb = channel_->close();
                cb.onFinalize( [&]() {
-                  stop_promise.set_value();
+                  try {
+                     stop_promise.set_value();
+                  } catch(...) {}
```
swatanabe-b1, line 107: How would you get more that one call to `set_value`?  From [future.promise] it looks to me like that is the only way `set_value` will throw for `promise<void>`.

heifner: I thought it might be possible for channel to call both onError and onFinalize.

swatanabe-b1: I see.

swatanabe-b1 (approved):

PR #10122, opened 2021-03-10
Add security group participant message handling to net_plugin

rusty-block-one:
## Change Description
Added logic to net_plugin_impl and connection classes to support security group in privacy features. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2,6 +2,7 @@
 
 #include <eosio/net_plugin/net_plugin.hpp>
 #include <eosio/net_plugin/protocol.hpp>
+#include <eosio/net_plugin/security_group.hpp>
 #include <eosio/chain/controller.hpp>
 #include <eosio/chain/exceptions.hpp>
 #include <eosio/chain/block.hpp>
@@ -27,6 +28,7 @@
 
 #include <atomic>
 #include <shared_mutex>
+#include <functional>
 
 using namespace eosio::chain::plugin_interface;
 
@@ -336,6 +338,17 @@ namespace eosio {
       constexpr static uint16_t to_protocol_version(uint16_t v);
 
       connection_ptr find_connection(const string& host)const; // must call with held mutex
+      /** private security group management */
+      security_group_manager security_group;
+      /** \brief  Update security group information
+       *
+       * Retrieves the security group information from the indicated block and
+       * updates the cached security group information. If a participant is
+       * removed from the cached security group, the TLS connection is closed.
+       *
+       * \param   bs The block_state containing the security group update
+       */
+      void update_security_group(const block_state_ptr& bs);
    };
 
    const fc::string logger_name("net_plugin_impl");
@@ -575,7 +588,20 @@ namespace eosio {
 
    class connection : public std::enable_shared_from_this<connection> {
    public:
+      /** @brief  For non TLS connection setup
+       *
+       * @param endpoint    The connection endpoint
+       */
       explicit connection( string endpoint );
+      /** @brief  For TLS connection setup
+       *
+       * @param endpoint        The connection endpoint
+       * @param participant     The name of the participant
+       * @param participaing    Indicates if the participant is participating in block creation (default = true)
+       *                            true - block creation
+       *                            false - syncing only
+       */
+      explicit connection( string endpoint, chain::account_name participant, bool participaing = true );
       connection();
 
       ~connection() {}
@@ -789,6 +815,32 @@ namespace eosio {
             ( "_lport", local_endpoint_port );
          return mvo;
       }
+
+      // security group management
+      //
+   private:
+      std::optional<security_group_participant> security_group_info_;
+
+   public:
+      /** @brief  Set/change the participant name for the connection
+       *
+       *  @param name   The name of the participant
+       */
+      void participant_name(chain::account_name name) {
+         security_group_info_ = security_group_participant(name);
+      }
+      /** @brief  Returns the optional name of the participant */
+      std::optional<chain::account_name> participant_name();
+      bool is_participating() const {
+         return security_group_info_ && security_group_info_->is_participating();
+      }
+      bool is_syncing() const {
+         return security_group_info_ && security_group_info_->is_syncing();
+      }
+      /**   @brief    Flag the tls connection as participating in block production */
+      void now_participating();
+      /**   @brief   Flag the connection as only publishing sync information */
+      void now_syncing();
    };
 
    const string connection::unknown = "<unknown>";
@@ -875,6 +927,12 @@ namespace eosio {
         last_handshake_sent()
    {
       fc_ilog( logger, "creating connection to ${n}", ("n", endpoint) );
+   }
+
+   connection::connection( string endpoint, chain::account_name participant, bool participating )
+      : connection(endpoint)
+   {
+      security_group_info_ = security_group_participant(participant, participating);
    }
 
    connection::connection()
@@ -1286,6 +1344,22 @@ namespace eosio {
       });
 
       return true;
+   }
+
+   std::optional<chain::account_name> connection::participant_name() {
+      if(security_group_info_)
+         return security_group_info_.value().name();
+      return std::optional<chain::account_name>();
+   }
+
+   void connection::now_participating() {
+      if(security_group_info_)
+         security_group_info_.value().now_participating();
+   }
+
+   void connection::now_syncing() {
+      if(security_group_info_)
+         security_group_info_.value().now_syncing();
    }
 
    //------------------------------------------------------------------------
@@ -3428,10 +3502,40 @@ namespace eosio {
       }
    }
 
+   // called from any thread
+   void net_plugin_impl::update_security_group(const block_state_ptr& bs) {
+      // update connections
+      //
+      auto updater = [](chain::account_name name, auto handler) {
+         for_each_connection([name, handler](auto& cp) {
+            auto participant_name = cp->participant_name();
+            if(participant_name && participant_name.value() == name) {
+               handler(*cp);
+            }
+            return true;
+         });
+      };
+      // add a new participant to the security group
+      //
+      auto on_add = [updater](chain::account_name name) {
+         updater(name, [](connection& con) { con.now_participating(); });
+      };
+      // remove a participant from the security group
+      //
+      auto on_remove = [updater](chain::account_name name) {
+         updater(name, [](connection& con) { con.now_syncing(); });
+      };
+      // update cache
+      //
+      auto& update = bs->get_security_group_info();
+      security_group.update_cache(update.version, update.participants, on_add, on_remove);
+   }
+
    // called from application thread
    void net_plugin_impl::on_accepted_block(const block_state_ptr& bs) {
       update_chain_info();
       controller& cc = chain_plug->chain();
+      update_security_group(bs);
       dispatcher->strand.post( [this, bs]() {
          fc_dlog( logger, "signaled accepted_block, blk num = ${num}, id = ${id}", ("num", bs->block_num)("id", bs->id) );
 
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2227,17 +2227,13 @@ namespace eosio {
       for_each_block_connection( [this, &id, &bnum, &b, &buff_factory]( auto& cp ) {
          peer_dlog( cp, "socket_is_open ${s}, connecting ${c}, syncing ${ss}",
                     ("s", cp->socket_is_open())("c", cp->connecting.load())("ss", cp->syncing.load()) );
-         if( !cp->current() ) return true;
+         if( !cp->current() || !cp->is_participating() ) return true;
          send_buffer_type sb = buff_factory.get_send_buffer( b, cp->protocol_version.load() );
          if( !sb ) {
             peer_wlog( cp, "Sending go away for incomplete block #${n} ${id}...",
                        ("n", b->block_num())("id", b->calculate_id().str().substr(8,16)) );
             // unable to convert to v0 signed block and client doesn't support proto_pruned_types, so tell it to go away
             cp->enqueue( go_away_message( fatal_other ) );
-            return true;
-         }
-
-         if(!cp->is_participating()) {
             return true;
          }
 
@@ -3478,7 +3474,14 @@ namespace eosio {
    void net_plugin_impl::update_security_group(const block_state_ptr& bs) {
       // update cache
       //
+      // Check the version before taking the lock.  Since this is the only thread that
+      // touches the version information, no need to take the lock if the version is
+      // unchanged.
+      //
       auto& update = bs->get_security_group_info();
+      if(security_group.current_version() == update.version) {
+         return;
+      }
       std::lock_guard<std::shared_mutex> connection_guard(connections_mtx);
       if(!security_group.update_cache(update.version, update.participants)) {
          return;
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2237,6 +2237,10 @@ namespace eosio {
             return true;
          }
 
+         if(!cp->is_participating()) {
+            return true;
+         }
+
          cp->strand.post( [this, cp, id, bnum, sb{std::move(sb)}]() {
             std::unique_lock<std::mutex> g_conn( cp->conn_mtx );
             bool has_block = cp->last_handshake_recv.last_irreversible_block_num >= bnum;
@@ -2282,7 +2286,7 @@ namespace eosio {
 
       trx_buffer_factory buff_factory;
       for_each_connection( [this, &trx, &nts, &buff_factory]( auto& cp ) {
-         if( cp->is_blocks_only_connection() || !cp->current() ) {
+         if( cp->is_blocks_only_connection() || !cp->current() || !cp->is_participating() ) {
             return true;
          }
          nts.connection_id = cp->connection_id;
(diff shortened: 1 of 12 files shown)


rusty-block-one:
@brianjohnson5972 the PR is ready for review

heifner:
Can we use descriptive names for PRs please.

dimas1185:
FYI, EPE is private jira reference while PRs are public. Please substitute EPE with reasonable description.

rusty-block-one:
@heifner @dimas1185 Changed name to descriptive name

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3919,6 +4013,16 @@ namespace eosio {
       if( my->find_connection( host ) )
          return "already connected";
 
+      /** @todo when making a shared tls connection, need to use the ctor that takes
```
brianjohnson5972, line 4016: Since setting the participant name has to be handled after the certificate handshake, just remove these comments

tests/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ include_directories( "${CMAKE_SOURCE_DIR}/plugins/wallet_plugin/include" )
 file(GLOB UNIT_TESTS "*.cpp")
 
 add_executable( plugin_test ${UNIT_TESTS} )
-target_link_libraries( plugin_test eosio_testing eosio_chain chainbase chain_plugin wallet_plugin fc ${PLATFORM_SPECIFIC_LIBS} )
+target_link_libraries( plugin_test eosio_testing eosio_chain chainbase chain_plugin wallet_plugin fc net_plugin ${PLATFORM_SPECIFIC_LIBS} )
```
brianjohnson5972, line 8: Why did you need to add this?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +807,29 @@ namespace eosio {
             ( "_lport", local_endpoint_port );
          return mvo;
       }
+
+      // security group management
+      //
+   private:
+      std::optional<chain::account_name> participant_name_;
+      std::atomic<bool> participating_{false};
```
brianjohnson5972, line 815: default to true, the security_group changing logic and logic for TLS connections should handle setting it false when it needs to be

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +807,29 @@ namespace eosio {
...
+
+   public:
+      /** @brief  Set/change the participant name for the connection
+       *
+       *  @param name   The name of the participant
+       */
+      void participant_name(chain::account_name name) { participant_name_ = name; }
```
brianjohnson5972, line 822: The TLS handling for this should be in the connection, so leave off the setter

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Removed comment

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Removed setter

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Changed default value

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +807,29 @@ namespace eosio {
...
+      /** @brief  Returns the optional name of the participant */
+      auto participant_name() { return participant_name_; }
+      /** returns true if connection is in the security group */
+      bool is_participating() const { return participating_.load(std::memory_order_relaxed); }
+       /** returns false if the connection is not in the security group */
+      bool is_syncing() const { return !participating_.load(std::memory_order_relaxed); }
+      /**   @brief    Flag the tls connection as participating in block production */
+      void now_participating() { participating_.store(true, std::memory_order_relaxed); }
```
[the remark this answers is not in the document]
rusty-block-one: changed to use set_participating

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +807,29 @@ namespace eosio {
...
+   public:
+      /** @brief  Set/change the participant name for the connection
+       *
+       *  @param name   The name of the participant
+       */
+      void participant_name(chain::account_name name) { participant_name_ = name; }
+      /** @brief  Returns the optional name of the participant */
+      auto participant_name() { return participant_name_; }
+      /** returns true if connection is in the security group */
+      bool is_participating() const { return participating_.load(std::memory_order_relaxed); }
```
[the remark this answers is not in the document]
rusty-block-one: The default for atomic is memory_order_seq_cst. I assume that all we want to do is change the value of the flag and not hold up any threads simultaneously accessing this value. With the default, the compiler may generate could that could temporarily block a thread in order to make sure there are no visible side effects.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -336,6 +339,17 @@ namespace eosio {
       constexpr static uint16_t to_protocol_version(uint16_t v);
 
       connection_ptr find_connection(const string& host)const; // must call with held mutex
+      /** private security group management */
+      security_group_manager security_group;
+      /** \brief  Update security group information
+       *
+       * Retrieves the security group information from the indicated block and
+       * updates the cached security group information. If a participant is
+       * removed from the cached security group, the TLS connection is closed.
```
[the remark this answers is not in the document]
rusty-block-one: updated comment

plugins/net_plugin/net_plugin.cpp
```diff
@@ -336,6 +339,17 @@ namespace eosio {
       constexpr static uint16_t to_protocol_version(uint16_t v);
 
       connection_ptr find_connection(const string& host)const; // must call with held mutex
+      /** private security group management */
+      security_group_manager security_group;
```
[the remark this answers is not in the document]
rusty-block-one: moved

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1435,6 +1476,18 @@ namespace eosio {
 
    void connection::enqueue( const net_message& m ) {
       verify_strand_in_this_thread( strand, __func__, __LINE__ );
+      // for tls connections, when the connection is not in the security group
+      // certain message types will not be transmitted
+      if(participant_name_ && !participating_) {
```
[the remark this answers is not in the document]
rusty-block-one: changed now to just check _participating (by calling is_participating)

plugins/net_plugin/net_plugin.cpp
```diff
@@ -256,6 +258,7 @@ namespace eosio {
 
       mutable std::shared_mutex             connections_mtx;
       std::set< connection_ptr >            connections;     // todo: switch to a thread safe container to avoid big mutex over complete collection
+      std::unordered_multimap<chain::account_name, connection_ptr> particiant_connections;
```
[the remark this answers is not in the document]
rusty-block-one: removed

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +807,29 @@ namespace eosio {
...
+       *
+       *  @param name   The name of the participant
+       */
+      void participant_name(chain::account_name name) { participant_name_ = name; }
+      /** @brief  Returns the optional name of the participant */
+      auto participant_name() { return participant_name_; }
+      /** returns true if connection is in the security group */
+      bool is_participating() const { return participating_.load(std::memory_order_relaxed); }
+       /** returns false if the connection is not in the security group */
+      bool is_syncing() const { return !participating_.load(std::memory_order_relaxed); }
```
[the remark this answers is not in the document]
rusty-block-one: changed to use is_participating

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1435,6 +1476,18 @@ namespace eosio {
...
+      if(participant_name_ && !participating_) {
+         const bool ignore = std::holds_alternative<notice_message>(m) ||
+                             std::holds_alternative<signed_block_v0>(m) ||
+                             std::holds_alternative<packed_transaction_v0>(m) ||
+                             std::holds_alternative<signed_block>(m) ||
+                             std::holds_alternative<trx_message_v1>(m);
+         if(ignore) {
+            return;
+         }
+      }
```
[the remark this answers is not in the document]
rusty-block-one: Added

tests/CMakeLists.txt
rusty-block-one: When I initially developed the unit test, I placed it in the tests directory.  When we decided to create a test directory under net_plugin for unit tests, I did not backtrack and remove the change.  The file has been modified back to the original.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3481,51 @@ namespace eosio {
       }
    }
 
+   // called from any thread
+   void net_plugin_impl::update_security_group(const block_state_ptr& bs) {
+      // update cache
```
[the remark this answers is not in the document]
rusty-block-one: Now using connections_mtx

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3481,51 @@ namespace eosio {
...
+               added_connections.push_back(connection);
+            }
+         }
+         else {
+            connection->now_syncing();
+         }
+         return true;
+      };
+
+      for_each_connection(do_update);
```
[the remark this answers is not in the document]
rusty-block-one: changed to loop on connections instead of calling for_each_connection

plugins/net_plugin/net_plugin.cpp
```diff
@@ -254,8 +256,9 @@ namespace eosio {
       bool                                  use_socket_read_watermark = false;
       /** @} */
 
-      mutable std::shared_mutex             connections_mtx;
+      mutable std::shared_mutex             connections_mtx; // protects both connections and security_group
```
heifner, line 259: As implemented `connections_mtx` does not protect the `security_group`. It looks to me like `security_group` is only accessed and modified via the application thread.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3470,53 @@ namespace eosio {
...
+            connection->set_participating(false);
+         }
+      };
+
+      // send handshake when added to group
+      //    - connection mutex no longer needd
+      //
+      connection_guard.unlock();
+      for(auto& connection : added_connections) {
+         connection->send_handshake();
```
heifner, line 3511: This just does a post on a strand. I don't see the need to pull this out and send after releasing the mutex. Just call send_handshake() above so that `added_connections` is not needed.

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: It will need to be checked by the connection itself after the certificates are exchanged, to determine if the incoming participant is in the security group and can receive blocks or not.  I thought it would be simpler for it to be covered by the connections mutex rather than having a separate one and having to make sure that they are taken in the correct order.

heifner: I don't think it needs a mutex as it is only accessed via the application thread, right?

brianjohnson5972: The TLS connection portion isn't implemented yet. The connection will have an extra handshake where the participant will be identified and the connection will then have to search in the security group to see if the participant is in or out of the group, and that won't be on the application thread, it will be on one of the net_plugin thread pool threads.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3484,38 +3473,40 @@ namespace eosio {
...
-      // update connection
+      // update connections
+      //    - connection_mtx still needed
       //
-      auto do_update = [&](auto& connection) {
+      for(auto& connection : connections) {
          const auto& participant = connection->participant_name();
          if(!participant) {
-            return true;
+            return;
```
brianjohnson5972, line 3493: this should be continue instead of return

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3470,53 @@ namespace eosio {
       }
    }
 
+   // called from any thread
+   void net_plugin_impl::update_security_group(const block_state_ptr& bs) {
+      // update cache
+      //    - connection_mtx is needed
+      //
+      auto& update = bs->get_security_group_info();
+      std::shared_lock<std::shared_mutex> connection_guard(connections_mtx);
```
heifner, line 3479: This needs to be a `lock_guard` as it is modifying state.

plugins/net_plugin/net_plugin.cpp
heifner: In that case it needs to be a lock_guard not a shared_lock of the connections_mtx.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1446,6 +1485,9 @@ namespace eosio {
    }
 
    void connection::enqueue_block( const signed_block_ptr& b, bool to_sync_queue) {
+      if(!is_participating()) {
```
brianjohnson5972, line 1488: Need similar code (except with continue instead of return) also in dispatch_manager::bcast_block, after the if (!sb) block that sends the go away message, before block is sent to enque_buffer. Also in dispatch_manager::bcast_transaction, need to change:

if( cp->is_blocks_only_connection() || !cp->current() )

into:

if( cp->is_blocks_only_connection() || !cp->current() || !cp->is_participating() )

I believe this covers everything, but double check me that the "Need to check participating flag before calling enque_buffer for any transaction, block, or notice_messages for any connection." requirement has been met.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3470,53 @@ namespace eosio {
...
+
+      // update connections
+      //    - connection_mtx still needed
+      //
+      for(auto& connection : connections) {
+         const auto& participant = connection->participant_name();
+         if(!participant) {
+            return;
+         }
+         if(security_group.is_in_security_group(participant.value())) {
```
brianjohnson5972, line 3495: Need to change sec;urity_group_manager::is_in_security_group to always return true if cache_.empty().  This will allow this logic here to work if we ever remove all participants in the security group and it allows Dmytro's code to only have to call that method to determine what he should set participating_ to.

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
rusty-block-one: changed to continue

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Removed added_connections and now call send_handsahke in update loop

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Changed to lock_guard

plugins/net_plugin/net_plugin.cpp
rusty-block-one: Modified is_in_security_group to also return true when the cache_ is empty

plugins/net_plugin/net_plugin.cpp
heifner: This is called on every `on_accepted_block` which means we now have a mutex on the critical performance path. Can we re-design this so that we don't need to acquire a mutex on every block.

plugins/net_plugin/net_plugin.cpp
rusty-block-one: In dispatch_manager::bcast_block, the "if(!sb)" block is in the lambda passed to for_each_block_connection.  So "return true" actually results in a continue to the next connection.

I did a search on all references to enqueue_buffer and each has a preceding call to is_participating() in the code to bypass the call to enqueue_buffer.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2237,6 +2237,10 @@ namespace eosio {
             return true;
          }
 
+         if(!cp->is_participating()) {
```
brianjohnson5972, line 2240: Sorry, this block should go before line 2231 to avoid bothering to create the send_buffer

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3474,40 @@ namespace eosio {
       }
    }
 
+   // called from any thread
+   void net_plugin_impl::update_security_group(const block_state_ptr& bs) {
+      // update cache
+      //
+      auto& update = bs->get_security_group_info();
+      std::lock_guard<std::shared_mutex> connection_guard(connections_mtx);
```
brianjohnson5972, line 3482: still need to add something like:
// NOTE: since this is the only thread that ever updates security_group.version it is safe to check this before taking the lock
if (security_group.version == update.version) {
   return;
}

just before taking the lock to avoid taking the lock on every block when we only need to when the version changed.

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
rusty-block-one: moved the test prior to creating send_buffer

plugins/net_plugin/test/security_group_manager_tests.cpp
```diff
@@ -37,6 +37,16 @@ BOOST_AUTO_TEST_CASE(test_initial_population) {
       BOOST_REQUIRE(manager.is_in_security_group(participant));
    }
 }
+
+BOOST_AUTO_TEST_CASE(test_version) {
+   eosio::security_group_manager manager;
+   BOOST_REQUIRE(manager.current_version() == 0);
+
+      auto populate = create_list({ 1, 2, 3, 4, 5, 6});
```
brianjohnson5972, line 45: fix the indentation here

brianjohnson5972 (changes requested):
Indentation error is the last thing and then we can merge

plugins/net_plugin/test/security_group_manager_tests.cpp
rusty-block-one: indentation fixed

brianjohnson5972 (approved):

PR #10141, opened 2021-03-16
Separate amqp_trx_plugin from amqp_trace_plugin

heifner:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

```
  --amqp-trx-trace-reliable-mode arg (=queue)
                                        Reliable mode 'exit', exit application
                                        on any AMQP publish error.
                                        Reliable mode 'queue', queue
                                        transaction traces to send to AMQP on
                                        connection establishment.
                                        Reliable mode 'log', log an error and
                                        drop message when unable to directly
                                        publish to AMQP.

```

### DIFF
--- a/libraries/amqp/include/eosio/amqp/amqp_handler.hpp
+++ b/libraries/amqp/include/eosio/amqp/amqp_handler.hpp
@@ -22,7 +22,7 @@ public:
    // delivery_tag type of consume, use for ack/reject
    using delivery_tag_t = uint64_t;
    // called from amqp thread on consume of message
-   using on_consume_t = std::function<void(const AMQP::Message& message, const delivery_tag_t& delivery_tag, bool redelivered)>;
+   using on_consume_t = std::function<void(const AMQP::Message& message, delivery_tag_t delivery_tag, bool redelivered)>;
 
    /// @param address AMQP address
    /// @param name AMQP routing key
@@ -229,9 +229,8 @@ private:
          on_error( message );
          first_connect_.set();
       } );
-      consumer.onReceived( [&](const AMQP::Message& message, const delivery_tag_t& delivery_tag, bool redelivered) {
-         on_consume_( message, delivery_tag, redelivered );
-      } );
+      static_assert( std::is_same_v<on_consume_t, AMQP::MessageCallback>, "AMQP::MessageCallback interface changed" );
+      consumer.onReceived( on_consume_ );
    }
 
    // called from amqp thread
--- a/libraries/amqp/include/eosio/amqp/amqp_handler.hpp
+++ b/libraries/amqp/include/eosio/amqp/amqp_handler.hpp
@@ -22,7 +22,7 @@ public:
    // delivery_tag type of consume, use for ack/reject
    using delivery_tag_t = uint64_t;
    // called from amqp thread on consume of message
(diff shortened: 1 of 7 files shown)


libraries/amqp/include/eosio/amqp/amqp_handler.hpp
```diff
@@ -230,7 +230,7 @@ class amqp_handler {
          first_connect_.set();
       } );
       consumer.onReceived( [&](const AMQP::Message& message, const delivery_tag_t& delivery_tag, bool redelivered) {
-         on_consume_( delivery_tag, message.body(), message.bodySize() );
+         on_consume_( message, delivery_tag, redelivered );
```
swatanabe-b1, line 233: `on_consume_t` is almost the same type as `AMQP::MessageCallback` now.  Is there a reason for forwarding the arguments through an extra lambda?

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -187,15 +192,24 @@ void amqp_trx_plugin::set_program_options(options_description& cli, options_desc
    op("amqp-trx-ack-mode", bpo::value<ack_mode>()->default_value(ack_mode::in_block),
       "AMQP ack when 'received' from AMQP, when 'executed', or when 'in_block' is produced that contains trx.\n"
       "Options: received, executed, in_block");
+   op("amqp-trx-trace-address", bpo::value<std::string>(),
+      "AMQP address: Format: amqp://USER:PASSWORD@ADDRESS:PORT\n"
+      "Will publish to amqp-trx-trace-queue-name ('trace') queue. If not provided traces are not sent.");
+   op("amqp-trx-trace-queue-name", bpo::value<std::string>()->default_value("trace"),
+      "AMQP queue to publish transaction traces of amqp transactions.");
+   op("amqp-trx-trace-exchange", bpo::value<std::string>()->default_value(""),
+      "Existing AMQP exchange to send transaction trace messages.");
```
swatanabe-b1, line 201: I don't think that these should be options.  The trx plugin should _only_ send a trace via `reply-to`, it should _always_ send a trace when a `reply-to` is given, and it should always send the trace to the same AMQP server from which it receives transactions.  Broadcasting traces is the responsibility of the trace plugin.

swatanabe-b1 (changes requested):

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
heifner: This is a really good point. @arhag do you agree with this?

libraries/amqp/include/eosio/amqp/amqp_handler.hpp
heifner: `onReceived` takes a std::function. So no need for the extra lambda.

swatanabe-b1 (approved):

PR #10145, opened 2021-03-16
TLS connections and connection support for security group

dimas1185:
## Change Description

TLS implementation for net_plugin.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

3 options were added to nodeos:
`p2p-tls-security-group-ca-file` Certificate Authority's certificate file used for verifying peers TLS connection when security groups feature enabled"
`p2p-tls-own-certificate-file` Certificate file that will be used to authenticate running node if TLS is enabled"
`p2p-tls-private-key-file` Private key file that is used in conjunction with p2p-tls-own-certificate-file for server authorization in TLS connection. Together p2p-tls-private-key-file + p2p-tsl-own-certificate-file automatically enables TLS-only connection for peers.

## Documentation Additions
- [] Documentation Additions


plugins/net_plugin/net_plugin.cpp
```diff
@@ -975,9 +1000,22 @@ namespace eosio {
...
+            fc_dlog( logger, "connected to ${peer}", ("peer", peer_name()) );
+            socket_open = true;
+            start_read_message();
+         };
+         if (use_ssl) {
+            ssl_socket->async_handshake(ssl::stream_base::server,
+                                       [start_read](const auto& ec){
+                                          EOS_ASSERT(!ec, ssl_handshake_error, "ssl handshake error: ${e}", ("e", ec.message()));
+                                          start_read();
+                                       });
```
heifner, line 1014: You need to `bind_executor` to the `strand` the lambda. See for example `async_connect`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2449,10 +2497,10 @@ namespace eosio {
       connecting = true;
       pending_message_buffer.reset();
       buffer_queue.clear_out_queue();
-      boost::asio::async_connect( *socket, endpoints,
+      boost::asio::async_connect( get_socket(), endpoints,
          boost::asio::bind_executor( strand,
-               [resolver, c = shared_from_this(), socket=socket]( const boost::system::error_code& err, const tcp::endpoint& endpoint ) {
-            if( !err && socket->is_open() && socket == c->socket ) {
+               [resolver, c = shared_from_this(), psocket=&get_socket()]( const boost::system::error_code& err, const tcp::endpoint& endpoint ) {
```
heifner, line 2502: This needs to capture the shared_ptr of the socket so it maintains the lifetime of the socket.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -624,9 +634,11 @@ namespace eosio {
 
       std::atomic<connection_types>             connection_type{both};
 
+      std::shared_ptr<tcp::socket>              socket;     // only accessed through strand after construction
+      std::shared_ptr<ssl_stream>               ssl_socket;
    public:
+      bool                                      use_ssl;
```
heifner, line 640: make a const bool since it is not changed after construction.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1221,12 +1263,11 @@ namespace eosio {
       buffer_queue.fill_out_buffer( bufs );
 
       strand.post( [c{std::move(c)}, bufs{std::move(bufs)}]() {
-         boost::asio::async_write( *c->socket, bufs,
-            boost::asio::bind_executor( c->strand, [c, socket=c->socket]( boost::system::error_code ec, std::size_t w ) {
+         auto write_lambda = [c, psocket=&c->get_socket()]( boost::system::error_code ec, std::size_t w ) {
```
heifner, line 1266: This needs to capture the shared_ptr of the socket so it maintains the lifetime of the socket.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2464,16 +2512,17 @@ namespace eosio {
...
 
    void net_plugin_impl::start_listen_loop() {
-      connection_ptr new_connection = std::make_shared<connection>();
+      connection_ptr new_connection = std::make_shared<connection>(ssl_enabled);
       new_connection->connecting = true;
       new_connection->strand.post( [this, new_connection = std::move( new_connection )](){
-         acceptor->async_accept( *new_connection->socket,
-            boost::asio::bind_executor( new_connection->strand, [new_connection, socket=new_connection->socket, this]( boost::system::error_code ec ) {
+         acceptor->async_accept( new_connection->get_socket(),
+            boost::asio::bind_executor( new_connection->strand, [new_connection, this]( boost::system::error_code ec ) {
```
heifner, line 2519: This needs to capture the shared_ptr of the socket so it maintains the lifetime of the socket.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2565,13 +2613,9 @@ namespace eosio {
             close( false );
             return;
          }
-
-         boost::asio::async_read( *socket,
-            pending_message_buffer.get_buffer_sequence_for_boost_async_read(), completion_handler,
-            boost::asio::bind_executor( strand,
-              [conn = shared_from_this(), socket=socket]( boost::system::error_code ec, std::size_t bytes_transferred ) {
+         auto handle_read = [conn = shared_from_this(), psocket=&get_socket()]( boost::system::error_code ec, std::size_t bytes_transferred ) {
```
heifner, line 2616: This needs to capture the shared_ptr of the socket so it maintains the lifetime of the socket.

heifner (changes requested):
Didn't review the ssl certificate code.

plugins/net_plugin/net_plugin.cpp
dimas1185: @heifner we retain connection pointer and we use socket only to check if it's address was changed. so I think we good.

heifner: But the socket can be reset out from under you on a `close()`.

libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -681,4 +681,13 @@ namespace eosio { namespace chain {
    FC_DECLARE_DERIVED_EXCEPTION( state_history_exception,    chain_exception,
                                  3280000, "State history exception" )
 
+   FC_DECLARE_DERIVED_EXCEPTION( ssl_exception, chain_exception,
+                                 3290000, "SSL exception")
+
+      FC_DECLARE_DERIVED_EXCEPTION( ssl_incomplete_configuration, ssl_exception,
```
brianjohnson5972, line 687: make indentation on this line and following match the ones above

dimas1185: the one above is root class exception

dimas1185:
> The `flex_socket` does not solve the issue. Since `async_write` is using a raw socket THAT socket must live for the complete lifetime of the `async_write` the typical pattern followed is to capture a shared_ptr to the socket in the lambda so that the socket lives as long as the `async_write`.

`async_write` is using not raw but `flex_socket`.
`flex_socket                               socket;     // only accessed through strand after construction`
`auto write_lambda = [c, socket=c->socket](` <-- here we copy `flex_socket` instance by value hence increase `shared_ptr` counter

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3732,7 +3892,7 @@ namespace eosio {
            "   _lip   \tlocal IP address connected to peer\n\n"
            "   _lport \tlocal port number connected to peer\n\n")
          ( "p2p-keepalive-interval-ms", bpo::value<int>()->default_value(def_keepalive_interval), "peer heartbeat keepalive message interval in milliseconds")
-         ( "p2p-tls-ca-certificate-file", bpo::value<string>(), "Certificate Authority's certificate file used for verifying peers TLS connection when security groups feature enabled" )
+         ( "p2p-tls-security-group-ca-file", bpo::value<string>(), "Certificate Authority's certificate file used for verifying peers TLS connection when security groups feature enabled" )
```
brianjohnson5972, line 3895: `change these three parameters to use bps::value<bfs::path>()`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3843,6 +4003,22 @@ namespace eosio {
             my->chain_plug->enable_accept_transactions();
          }
 
+         //if we have certificate option that TLS must be enabled
+         if ( options.count("p2p-tls-own-certificate-file") ) {
+            auto certificate = options["p2p-tls-own-certificate-file"].as<string>();
```
brianjohnson5972, line 4008: use bfs::path instead of string for these 3 parameters

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3843,6 +4003,22 @@ namespace eosio {
...
+            auto pkey        = options["p2p-tls-private-key-file"].as<string>();
+            auto ca_cert     = options["p2p-tls-security-group-ca-file"].as<string>();
+
```
brianjohnson5972, line 4011: need to add this here:
            auto relative_to_absolute = [](bfs::path& file) {
               if( file.is_relative()) {
                  file = bfs::current_path() / file;
               }
            };
            relative_to_absolute(certificate);
            relative_to_absolute(pkey);
            relative_to_absolute(ca_cert);

to turn relative paths into absolute paths

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -969,15 +1070,31 @@ namespace eosio {
...
-         socket_open = true;
-         start_read_message();
+
+         auto start_read = [this](){
+            fc_dlog( logger, "connected to ${peer}", ("peer", peer_name()) );
+            socket_open = true;
+            start_read_message();
+         };
+         if (socket.use_ssl) {
+            socket.ssl_socket->async_handshake(ssl::stream_base::server,
```
heifner, line 1086: Nothing keeps this `ssl_socket` alive for life of async call.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1221,12 +1340,12 @@ namespace eosio {
       buffer_queue.fill_out_buffer( bufs );
 
       strand.post( [c{std::move(c)}, bufs{std::move(bufs)}]() {
-         boost::asio::async_write( *c->socket, bufs,
-            boost::asio::bind_executor( c->strand, [c, socket=c->socket]( boost::system::error_code ec, std::size_t w ) {
+         //capture by value is const so we need mutable keyword
+         auto write_lambda = [c, socket=c->socket]( boost::system::error_code ec, std::size_t w ) mutable {
```
heifner, line 1344: Good point. I missed that this was copying the flex_socket by value. It would be more efficient to only have one shared_ptr copy instead of two.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2464,16 +2592,17 @@ namespace eosio {
...
 
    void net_plugin_impl::start_listen_loop() {
-      connection_ptr new_connection = std::make_shared<connection>();
+      connection_ptr new_connection = std::make_shared<connection>(ssl_enabled);
       new_connection->connecting = true;
       new_connection->strand.post( [this, new_connection = std::move( new_connection )](){
-         acceptor->async_accept( *new_connection->socket,
-            boost::asio::bind_executor( new_connection->strand, [new_connection, socket=new_connection->socket, this]( boost::system::error_code ec ) {
+         acceptor->async_accept( new_connection->socket.get_socket(),
+            boost::asio::bind_executor( new_connection->strand, [new_connection, this]( boost::system::error_code ec ) {
```
heifner, line 2599: Does not capture a copy of the shared_ptr to socket.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
dimas1185: unfortunately we can't cast `tcp::socket` to `ssl::stream<tcp::socket>` that is why we have two pointers.
do you have any suggestions? perhaps we can use `variant<socket, ssl_socket>` instead of two pointers. anyway we instantiate only one of them.

plugins/net_plugin/net_plugin.cpp
dimas1185: good point. will be fixed.

plugins/net_plugin/net_plugin.cpp
heifner: You could use a shared_ptr<flex_socket> and then just unique_ptr or variant inside flex_socket.

dimas1185: since we care about increasing reference count in `flex_socket`'s object using `shared_ptr` for `flex_socket` looks redundant. I think using variant inside `flex_socket` will do the trick to avoid two shared_ptr copies (hence only one reference count increase which is costly atomic operation)

plugins/net_plugin/net_plugin.cpp
dimas1185: done

plugins/net_plugin/net_plugin.cpp
```diff
@@ -586,10 +598,53 @@ namespace eosio {
       block_status_monitor& operator=( block_status_monitor&& ) = delete;
    };
 
+   struct flex_socket {
+      using socket_ptr     = std::shared_ptr<tcp::socket>;
+      using ssl_socket_ptr = std::shared_ptr<ssl_stream>;
+      using socket_variant = std::variant<socket_ptr, ssl_socket_ptr>;
```
brianjohnson5972, line 604: Since we already have a bool flag to identify which is which, I think the variant just adds complexity and also adds an extra level of indirection and just have a tcp_socket_ptr and a ssl_socket_ptr (and make them protected so they are not exposed). And then inside this class use ssl_socket_ptr in place of ssl_socket. (You could also skip the use_ssl flag and use if ssl_socket_ptr bool operator(), but I'm fine either way).

Or even better, I think it would be good to make flex_socket a base class with pure virtual methods and then pull in all the logic in connection that uses this into the class itself by passing lambdas into the socket.  I think @huangminghuang could help you.

brianjohnson5972 (changes requested):
There may be more, but wanted to report these to you.  The refactoring logic tasks could wait for a subsequent PR, but logic changes need to happen with this.

tests/Cluster.py
```diff
@@ -245,9 +245,29 @@ def insertSpecificExtraNodeosArgs(node, insertStr):
...
             def getArguments(number):
-                nodeCert = os.path.join(privacyDir, "node{}.crt".format(number))
-                nodeKey = os.path.join(privacyDir, "node{}_key.pem".format(number))
-                return "--p2p-tls-own-certificate-file {} --p2p-tls-private-key-file {} --p2p-tls-ca-certificate-file {}".format(nodeCert, nodeKey, certAuth)
+                # this function converts number to eos name string
+                # eos name can have only numbers 1-5
+                # e.g. 0 -> 1, 6 -> 5.1, 12 -> 5.5.2
+                def normalizeNumber(number):
+                    if number <= 0:
+                        number = 1
```
brianjohnson5972, line 253: won't this method return "1" for number=1 and number=0? I think this should be
max = 4
if number >= 0 and number <= max:
    return str(number+1)
 ...
       cnt -= max
       ...

dimas1185: logic of this method is to normalize number so I think it shouldn't increase it. However your concern with 0 and 1 makes sense, it is may result in bug if caller will start from zero. What about forbidding zero at all? just putting an `assert(number > 0)`

brianjohnson5972: well this is called with the node number that starts with 0, so it has to take 0, or else we need to adjust the number we are passing in, but that makes no sense to me, just do it here.  We also need to replicate this logic in the generate certificates script, creating node1,node2,...node5,node1.1,...

brianjohnson5972: and the while would be:

while cnt > max:

brianjohnson5972: actually, you are correct, when getArgument calls getNumber, it should add 1 to the passed in number and it should treat 1 as "1", 2 as "2", etc. So yes, assert on num==0

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: We will clean this up in a subsequent change

dimas1185: I'm not sure how do you think it can be improved. variant decision looks to me more elegant than any virtual stuff. virtual stuff adds some little performance overhead and more complexity. Also `ssl_socker_ptr` and `tcp_socket_ptr` are not interchangeable as calls like boost::async_write can't accept any common interface.

dimas1185: there is an alternative to always use `ssl_socket_ptr` but even in that case we won't eliminate `if/else` and usage will be like that:
`         if (my_impl->ssl_enabled){
            boost::asio::async_write( *c->socket, bufs, boost::asio::bind_executor( c->strand, write_lambda ));
         }
         else {
            boost::asio::async_write( *c->socket.lowest_layer(), bufs, boost::asio::bind_executor( c->strand, write_lambda ));
         }`
this kind of hack as we are using ssl stream class without touching any ssl logic so I decided to make it more straightforward with `variant` and use proper classes in corresponding cases.

brianjohnson5972 (approved):

PR #10151, opened 2021-03-17
Revert "Run these tests against any builds."

williamblevins:
This reverts commit 66d575d106a7db4503fd74540967339c83a4abb7.

This is breaking local workflows. I wanted these tests to run locally, but too many build paths, etc.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/full-version-label.sh
+++ b/tests/full-version-label.sh
@@ -2,6 +2,15 @@
 set -eo pipefail
 # The purpose of this test is to ensure that the output of the "nodeos --full-version" command matches the version string defined by our CMake files
 echo '##### Nodeos Full Version Label Test #####'
+# If the environment variable BUILDKITE_TAG is empty or unset, this test will echo success
+if [[ "$BUILDKITE_TAG" == '' || "$BUILDKITE" != 'true' ]]; then
+    echo 'This test is only run in Buildkite against tagged builds.'
+    [[ "$BUILDKITE" != 'true' ]] && echo 'This is not Buildkite.'
+    [[ "$BUILDKITE_TAG" == '' ]] && echo 'This is not a tagged build.'
+    echo 'Exiting...'
+    exit 0
+fi
+echo 'Tagged build detected, running test.'
 # orient ourselves
 [[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/eos/')
 [[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/EOSIO/eosio/')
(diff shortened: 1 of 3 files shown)


tests/full-version-label.sh
```diff
@@ -2,6 +2,15 @@
...
+    echo 'This test is only run in Buildkite against tagged builds.'
+    [[ "$BUILDKITE" != 'true' ]] && echo 'This is not Buildkite.'
+    [[ "$BUILDKITE_TAG" == '' ]] && echo 'This is not a tagged build.'
+    echo 'Exiting...'
+    exit 0
+fi
+echo 'Tagged build detected, running test.'
 # orient ourselves
 [[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/eos/')
 [[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/EOSIO/eosio/')
```
kj4ezj, line 16: What if we change this...
```bash
[[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/eos/')
[[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/EOSIO/eosio/')
[[ "$EOSIO_ROOT" == '' ]] && EOSIO_ROOT=$(echo $(pwd)/ | grep -ioe '.*/build/' | sed 's,/build/,,')
```
...to a relative path...
```bash
EOSIO_ROOT="$(dirname ${BASH_SOURCE[0]})/../.."
```
...or an unresolved absolute path...
```bash
EOSIO_ROOT="$(pwd)/$(dirname ${BASH_SOURCE[0]})/../.."
```
...or even a resolved absolute path:
```bash
EOSIO_ROOT_RELATIVE="$(pwd)/$(dirname ${BASH_SOURCE[0]})/../.."
cd "$EOSIO_ROOT_RELATIVE"
EOSIO_ROOT="$(pwd)"
cd "$OLDPWD"
```
There are even fancier suggestions [here](https://stackoverflow.com/questions/284662/how-do-you-normalize-a-file-path-in-bash), and I played around with `relpath`, but found my test code didn't like it when the script was invoked from inside the build directory using an absolute path to the script so more testing may be needed.

kj4ezj (changes requested):
We know what folder `cmake` puts this test into, defined [here](https://github.com/EOSIO/eos/blob/474f77dc1665b098c01c96a12b6326e846a6ba91/tests/CMakeLists.txt#L49) as `CMAKE_CURRENT_BINARY_DIR`. This maps to `/build/tests`.

Instead of disabling these test locally, what if we find our way from there?

kj4ezj:
Closed in favor of [pull request 10160](https://github.com/EOSIO/eos/pull/10160).

PR #10154, opened 2021-03-17
execute action set for preexecution hook

victorj8:
## Change Description

Call the action set for the preexecution hook on transaction_context::init after we have set the bill_to_accounts member, with the objective that this gets overrided by the action callback code.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -128,6 +128,24 @@ namespace eosio { namespace chain {
             }
          }
       }
+
+      if( control.is_builtin_activated( builtin_protocol_feature_t::register_transaction_hook ) ) {
+         auto gprops = control.get_global_properties();
+         const uint32_t preexecution_hook_type = 0;
+
+         auto it = find_if(std::begin(gprops.transaction_hooks), std::end(gprops.transaction_hooks),
+                           [&](transaction_hook &t_hook) {
+                              return t_hook.type == preexecution_hook_type;
+                           }
+         );
+
+         if (it != std::end(gprops.transaction_hooks)) {
+            chain::action preexecution_hook_action{{}, it->contract, it->action, {}};
+            schedule_action(preexecution_hook_action, preexecution_hook_action.account, false, 0, 0);
+            execute_action( 1, 0 );
+         }
+      }
+
       validate_ram_usage.reserve( bill_to_accounts.size() );
 
       // Update usage values of accounts to reflect new time
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -141,10 +141,8 @@ namespace eosio { namespace chain {
 
          if (it != std::end(gprops.transaction_hooks)) {
             chain::action preexecution_hook_action{{}, it->contract, it->action, {}};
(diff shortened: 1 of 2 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -128,6 +128,24 @@ namespace eosio { namespace chain {
...
+         auto it = find_if(std::begin(gprops.transaction_hooks), std::end(gprops.transaction_hooks),
+                           [&](transaction_hook &t_hook) {
+                              return t_hook.type == preexecution_hook_type;
+                           }
+         );
+
+         if (it != std::end(gprops.transaction_hooks)) {
+            chain::action preexecution_hook_action{{}, it->contract, it->action, {}};
+            schedule_action(preexecution_hook_action, preexecution_hook_action.account, false, 0, 0);
+            execute_action( 1, 0 );
```
heifner, line 145: Not sure you want to do this. It will generate traces for the preexecution_hook_action. Also not sure the ordinals are correct. Need to determine what if anything should be logged to deep mind logger for this.

victorj8: I modified my code so that it doesn't add the action trace of the preexecution hook, but i am not 100% sure understanding why this is not required, if you can elaborate on why is this the case,  that would be helpful.

Also I removed the part of the logging to deep mind.

Any other feedback about how to better do this task is welcome.

libraries/chain/transaction_context.cpp
```diff
@@ -128,6 +128,26 @@ namespace eosio { namespace chain {
...
+                              return t_hook.type == preexecution_hook_type;
+                           }
+         );
+
+         if (it != std::end(gprops.transaction_hooks)) {
+            chain::action preexecution_hook_action{{}, it->contract, it->action, {}};
+            int action_ordinal = schedule_action(preexecution_hook_action, preexecution_hook_action.account, false, 0, 0);
+            apply_context acontext( control, *this, action_ordinal, 0 );
+            acontext.exec();
+            trace->action_traces.pop_back();
```
heifner, line 147: Going to all the expense of creating an action_trace only to throw it away does not seem very efficient.

libraries/chain/transaction_context.cpp
heifner: I don't know what the requirements are for an action trace for this preexecution_hook. Need to ask @b1bart if an action trace is desired. And if desired if it should be logged to deep mind. I would suspect that this extra action trace would be confusing to consumers of transaction traces.

victorj8: @heifner, after consulting @b1bart, the trace should be included for preexecution_hook action and send to deep mind.

libraries/chain/transaction_context.cpp
victorj8: I added it back

ndcgundlach (approved):

PR #10162, opened 2021-03-19
Rodeos per block ack

heifner:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
+++ b/libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
@@ -1,0 +1,55 @@
+#include <fc/time.hpp>
+
+#include <string>
+#include <memory>
+#include <vector>
+#include <functional>
+#include <deque>
+
+namespace eosio {
+
+/*
+ * A transactional_amqp_publisher manages a connection to a single AMQP server that publishes messages to an exchange
+ * in a reliable manner: connections to the AMQP server are retried, publishing occurs in a transaction,
+ * publish blocks until AMQP acks transactional group of messages.
+ *
+ * on_fatal_error is called if AMQP does not ack transactional group of messages in the specified time. Called from
+ * AMQP thread or calling thread.
+ *
+ * Be aware that the transactional_amqp_publisher will NOT create the exchange. It requires the exchange
+ * to already exist. Publishing to a non-existing exchange will effectively block as the transaction will not be
+ * ack'ed.
+ */
+
+class transactional_amqp_publisher {
+   public:
+      // Called from amqp thread when publish messages are not ack
+      using error_callback_t = std::function<void(const std::string& err)>;
+
+      /// Create a reliable queue to the given server publishing to the given exchange
+      /// \param server_url server url as amqp://...
+      /// \param exchange the exchange to publish to
+      /// \param time_out time to wait on transactional ack from AMQP before calling on_fatal_error
+      /// \param on_fatal_error called from AMQP does not ack transaction in time_out time
+      transactional_amqp_publisher(const std::string& server_url, const std::string& exchange,
+                                   const fc::microseconds& time_out, error_callback_t on_fatal_error);
+
+      /// Publish messages. May be called from any thread except internal thread (do not call from on_fatal_error)
+      /// Blocks until AMQP ack of success transactional commit or on_fatal_error if timeout waiting on ack
+      /// \param queue set of messages to send in one transaction <routing_key, message_data>
+      void publish_messages_raw(std::deque<std::pair<std::string, std::vector<char>>>&& queue);
+
+      /// Publish message directly to AMQP queue. May be called from any thread.
+      /// Bypasses all transactional mechanisms and just does a simple AMQP publish. Does not queue or retry or block.
+      /// \param routing_key if empty() uses class provided default routing_key
+      /// \param data message to send
+      /// \param on_error() call from AMQP thread if unable to directly publish (e.g. not currently connected)
+      void publish_message_direct(const std::string& routing_key, std::vector<char> data, error_callback_t on_error);
+
+      ~transactional_amqp_publisher();
+
+   private:
+      std::unique_ptr<struct transactional_amqp_publisher_impl> my;
+};
+
+}
--- a/libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
+++ b/libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
@@ -35,7 +35,8 @@ class transactional_amqp_publisher {
                                    const fc::microseconds& time_out, error_callback_t on_fatal_error);
 
       /// Publish messages. May be called from any thread except internal thread (do not call from on_fatal_error)
-      /// Blocks until AMQP ack of success transactional commit or on_fatal_error if timeout waiting on ack
+      /// All calls should be from the same thread or at the very least no two calls should be performed concurrently.
+      /// Blocks until AMQP ack of success transactional commit or on_fatal_error if timeout waiting on ack.
       /// \param queue set of messages to send in one transaction <routing_key, message_data>
       void publish_messages_raw(std::deque<std::pair<std::string, std::vector<char>>>&& queue);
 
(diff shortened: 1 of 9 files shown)


swatanabe-b1:
I don't understand the motivation for this change.
- commitTransaction only guarantees that rabbitmq has taken responsibility for the message.  It does not guarantee that the message has been delivered.
- Rodeos can still be ahead of what has been successfully published to rabbit, so it doesn't allow you to get rid of the local storage for messages.

heifner:
> 1. commitTransaction only guarantees that rabbitmq has taken responsibility for the message.  It does not guarantee that the message has been delivered.
> 2. Rodeos can still be ahead of what has been successfully published to rabbit, so it doesn't allow you to get rid of the local storage for messages.

1. The desired responsibility is that rabbitmq has taken responsibility of the message. Once it has ack'ed the message it is rabbitmq responsibility to deliver or recover the message on DR.
2. Rodeos commits the changes to rocksdb after the ack from rabbitmq.

swatanabe-b1:
> 2. Rodeos commits the changes to rocksdb after the ack from rabbitmq.

That's true, but the commit happens regardless of whether the ack was received or not.  The only thing that I see that checks for `app().quit()` is `write_deltas`, which happens before the filter is processed.  `rodeos_db_snapshot::end_block` may commit the changes before we return to appbase and check the quitting flag.

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -14,7 +14,7 @@
 namespace b1 {
 
 class rabbitmq : public stream_handler {
-   std::shared_ptr<eosio::reliable_amqp_publisher> amqp_publisher_;
+   std::unique_ptr<eosio::reliable_amqp_publisher> amqp_publisher_;
```
swatanabe-b1, line 17: I don't think it makes sense to use `reliable_amqp_publisher` here anymore.  You're adding workarounds to bypass most of the functionality that it provides.

heifner:
> > 1. Rodeos commits the changes to rocksdb after the ack from rabbitmq.
> 
> That's true, but the commit happens regardless of whether the ack was received or not. The only thing that I see that checks for `app().quit()` is `write_deltas`, which happens before the filter is processed. `rodeos_db_snapshot::end_block` may commit the changes before we return to appbase and check the quitting flag.

Oh, I see what you mean now, thanks.

programs/rodeos/streams/rabbitmq.hpp
heifner: Yeah. I'll just create a transactional_amqp_publisher roughly based on reliable_amqp_publisher

libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include <memory>
+#include <vector>
+#include <functional>
+#include <deque>
+
+namespace eosio {
+
+/*
+ * A transactional_amqp_publisher manages a connection to a single AMQP server that publishes messages to an exchange
+ * in a reliable manner: connections to the AMQP server are retried, publishing occurs in a transaction,
```
swatanabe-b1, line 13: How much sense does it make to retry the connection, when you completely stop rodeos after 15 seconds?  Or asked a different way, why is the timeout so short?  I think it's unlikely that a problem that requires opening a new connection would be resolved that quickly.

libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
```diff
@@ -0,0 +1,55 @@
...
+      /// Create a reliable queue to the given server publishing to the given exchange
+      /// \param server_url server url as amqp://...
+      /// \param exchange the exchange to publish to
+      /// \param time_out time to wait on transactional ack from AMQP before calling on_fatal_error
+      /// \param on_fatal_error called from AMQP does not ack transaction in time_out time
+      transactional_amqp_publisher(const std::string& server_url, const std::string& exchange,
+                                   const fc::microseconds& time_out, error_callback_t on_fatal_error);
+
+      /// Publish messages. May be called from any thread except internal thread (do not call from on_fatal_error)
+      /// Blocks until AMQP ack of success transactional commit or on_fatal_error if timeout waiting on ack
```
swatanabe-b1, line 38: This implementation may split the messages into multiple transactions.  The previous implementation using `reliable_amqp_publisher` guaranteed that they were all in the same transaction.  Is that an acceptable change?

heifner: Thanks for pointing that out. I meant to ask @spoonincode about that. Seems like we should just put them all into one transaction unless rabbitmq has something that makes putting into smaller needed.

libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
heifner: During testing I noticed that losing connection to rabbitmq happens fairly often (every couple of hours) and is quickly regained. Something more like 30-60 seconds probably would be better to give it some more time to become available.

libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
heifner: Changed to do all events of a block in one transaction.

libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
swatanabe-b1: Have you not pushed this up yet, or did you change the wrong timeout?

heifner: Looks like I accidentally reverted it when reverting another change to the interface.

libraries/amqp/include/eosio/amqp/transactional_amqp_publisher.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+      /// Create a reliable queue to the given server publishing to the given exchange
+      /// \param server_url server url as amqp://...
+      /// \param exchange the exchange to publish to
+      /// \param time_out time to wait on transactional ack from AMQP before calling on_fatal_error
+      /// \param on_fatal_error called from AMQP does not ack transaction in time_out time
+      transactional_amqp_publisher(const std::string& server_url, const std::string& exchange,
+                                   const fc::microseconds& time_out, error_callback_t on_fatal_error);
+
+      /// Publish messages. May be called from any thread except internal thread (do not call from on_fatal_error)
```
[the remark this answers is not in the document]
heifner: Excellent point. Really you should have one transaction_amqp_publisher per thread.

libraries/amqp/transactional_amqp_publisher.cpp
```diff
@@ -0,0 +1,249 @@
...
+
+   ctx.stop();
+   thread.join();
+
+   //if a message is in flight, let it try to run to completion
+   if(in_flight) {
+      auto until = std::chrono::steady_clock::now() + std::chrono::seconds(15);
+      ctx.restart();
+      while(in_flight && ctx.run_one_until(until));
+   }
```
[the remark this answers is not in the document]
swatanabe-b1: That scenario creates unavoidable races.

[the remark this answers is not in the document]
swatanabe-b1: There's no way to have such a situation that would not also allow the transaction_amqp_publisher to be called after the destructor finishes, which is definitely UB.

swatanabe-b1 (approved):

PR #10164, opened 2021-03-19
Allow rodeos to recover from kill-9

heifner:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/rodeos/rodeos.cpp
+++ b/libraries/rodeos/rodeos.cpp
@@ -297,7 +297,7 @@ rodeos_filter::rodeos_filter(eosio::name name, const std::string& wasm_filename,
             filter_state->eosvmoc_tierup.emplace(
                   cache_path, eosvmoc_config, code,
                   eosio::chain::digest_type::hash(reinterpret_cast<const char*>(code.data()), code.size()));
-         } catch( const eosio::chain::database_exception& ) {
+         } catch( const eosio::chain::database_exception& e ) {
             wlog( "eosvmoc cache exception ${e} removing cache ${c}", ("e", e.to_string())("c", cache_path.generic_string()) );
             // destroy cache and try again
             boost::filesystem::remove_all( cache_path );
--- a/libraries/rodeos/rodeos.cpp
+++ b/libraries/rodeos/rodeos.cpp
@@ -292,9 +292,19 @@ rodeos_filter::rodeos_filter(eosio::name name, const std::string& wasm_filename,
 #ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
    if (eosvmoc_enable) {
       try {
-         filter_state->eosvmoc_tierup.emplace(
-               eosvmoc_path, eosvmoc_config, code,
-               eosio::chain::digest_type::hash(reinterpret_cast<const char*>(code.data()), code.size()));
+         auto cache_path = eosvmoc_path / "rodeos_eosvmoc_cc";
+         try {
+            filter_state->eosvmoc_tierup.emplace(
+                  cache_path, eosvmoc_config, code,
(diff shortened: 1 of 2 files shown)


swatanabe-b1:
What happens if it's dirty because another rodeos process is currently using it?

heifner:
> What happens if it's dirty because another rodeos process is currently using it?

BadThings(TM). Any recommendations on how to avoid that?

spoonincode:
Really? My first guess would have been it's okay because removing a file doesn't invalidate the existing open handles to that file (and I don't believe OC ever re-opens its cache file to encounter it switched out from under it).

I do wonder if OC should simply always do this on its own instead of farming it out externally. But at this point since rodeos future is kinda uncertain, not sure worth over thinking it...

heifner:
> Really? My first guess would have been it's okay because removing a file doesn't invalidate the existing open handles to that file (and I don't believe OC ever re-opens its cache file to encounter it switched out from under it).

I didn't test the scenario. Let me test it and see what happens.

heifner:
> I didn't test the scenario. Let me test it and see what happens.

```
warn  2021-03-23T15:58:57.754 thread-0  wasm_ql_plugin.cpp:139        plugin_initialize    ] 13 N2b18chain_kv9exceptionE: database::database: rocksdb::DB::Open: IO error: While lock file: /Users/kevin.heifner/ext/develop-boxed-2.1/cmake-build-relwithdebinfo/var/lib/rodeos/rocksdb/LOCK: Resource temporarily unavailable
rethrow database::database: rocksdb::DB::Open: IO error: While lock file: /Users/kevin.heifner/ext/develop-boxed-2.1/cmake-build-relwithdebinfo/var/lib/rodeos/rocksdb/LOCK: Resource temporarily unavailable:
    {"what":"database::database: rocksdb::DB::Open: IO error: While lock file: /Users/kevin.heifner/ext/develop-boxed-2.1/cmake-build-relwithdebinfo/var/lib/rodeos/rocksdb/LOCK: Resource temporarily unavailable"}
    thread-0  wasm_ql_plugin.cpp:139 plugin_initialize
```

swatanabe-b1:
The file is opened before exit to reset the dirty flag.  I think the following is possible:
- start rodeos1
- start rodeos2 - creates a new file
- exit rodeos1 cleanly - resets dirty flag in the file created by rodeos2
- kill -9 rodeos2
- start rodeos3 - dirty flag _not_ set 💥

swatanabe-b1:
There's also a race on startup, but that's a problem that already exists and it's much less likely.  I usually run into this situation when I forget that I'm already running rodeos, which implies a significant time lapse.

heifner:
> The file is opened before exit to reset the dirty flag. I think the following is possible:

Not sure what more we can do here except: "Doctor it hurts when I do ...".
Open to suggestions.

heifner:
> > The file is opened before exit to reset the dirty flag. I think the following is possible:
> 
> Not sure what more we can do here except: "Doctor it hurts when I do ...".
> Open to suggestions.

I guess we could just ALWAYS remove any existing eosvmoc cache on startup. Thoughts?

spoonincode:
I still need to refactor the OC code cache so it can work properly on a `noexec` filesystem (mostly meaning adding the ability to do heap/locked for the OC code cache). Maybe at that point can eliminate any re-opening and eliminate that race.

spoonincode (approved):

swatanabe-b1 (approved):

Issue #10140, opened 2021-03-16
Restarting nodeos caused corruption in fork db


matthewdarwin:
Seen on Jungle with nodeos 2.1.0-rc3.

stop nodeos, no errors:

```
info  2021-03-15T22:22:51.299 nodeos    resource_monitor_plugi:122    plugin_shutdown      ] shutdown...
info  2021-03-15T22:22:51.300 nodeos    resource_monitor_plugi:129    plugin_shutdown      ] exit shutdown
info  2021-03-15T22:22:51.351 nodeos    main.cpp:179                  main                 ] nodeos successfully exiting
```
on startup:

```
...
info  2021-03-15T22:22:57.741 nodeos    chain_api_plugin.cpp:98       plugin_startup       ] starting chain_api_plugin
info  2021-03-15T22:22:57.743 nodeos    net_api_plugin.cpp:53         plugin_startup       ] starting net_api_plugin
info  2021-03-15T22:22:57.743 nodeos    producer_plugin.cpp:932       plugin_startup       ] producer plugin:  plugin_startup() begin
info  2021-03-15T22:22:57.743 nodeos    producer_plugin.cpp:957       plugin_startup       ] Launching block production for 1 producers at 2021-03-15T22:22:57.743.
debug 2021-03-15T22:22:57.743 nodeos    producer_plugin.cpp:2047      schedule_production_ ] Waiting till another block is received
info  2021-03-15T22:22:57.743 nodeos    producer_plugin.cpp:968       plugin_startup       ] producer plugin:  plugin_startup() end
info  2021-03-15T22:22:57.743 nodeos    producer_api_plugin.cp:87     plugin_startup       ] starting producer_api_plugin
info  2021-03-15T22:22:57.743 nodeos    resource_monitor_plugi:94     plugin_startup       ] Creating and starting monitor thread
info  2021-03-15T22:22:57.743 nodeos    file_space_handler.hpp:112    add_file_system      ] /var/lib/nodeos/blocks's file system monitored. shutdown_available: 49065250810, capacity: 490652508160, threshold: 90
debug 2021-03-15T22:22:57.744 nodeos    producer_plugin.cpp:382       on_incoming_block    ] received incoming block 67350332 0403af3cbe05ed9e0d2d3715b3c6a5774b05e355eb3a0c58cd0c1a4e0a9eb9a5
[...]

PR #10170, opened 2021-03-22
Rel.2.1.x: Exit nodeos if fork_db is corrupted and lib cannot advance

linhuang-blockone:
## Change Description

https://blockone.atlassian.net/browse/EPE-831

https://github.com/EOSIO/eos/issues/10140 reports that nodeos gets stuck after restart due to a corrupt forkdb. When this happens, log_irreversible keeps throwing the same exception whenever it executes. The solution is to exit nodeos. Log detailed information to help future debugging.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -330,11 +330,8 @@ struct controller_impl {
          // todo: move this check to startup so id does not have to be calculated
          EOS_ASSERT( root_id == log_head->calculate_id(), fork_database_exception, "fork database root does not match block log head" );
       } else {
-         if( fork_db.root()->block_num != lib_num ) {
-            ilog( "Empty block log expects the first appended block to build off a block that is not the fork database root, terminating. root block number: ${block_num}, lib: ${lib_num}", ("block_num", fork_db.root()->block_num) ("lib_num", lib_num) );
-            shutdown();
-            return;
-         }
+         EOS_ASSERT( fork_db.root()->block_num == lib_num, fork_database_exception,
+                     "empty block log expects the first appended block to build off a block that is not the fork database root" );
       }
 
       auto fork_head = (read_mode == db_read_mode::IRREVERSIBLE) ? fork_db.pending_head() : fork_db.head();
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -330,8 +330,11 @@ struct controller_impl {
          // todo: move this check to startup so id does not have to be calculated
          EOS_ASSERT( root_id == log_head->calculate_id(), fork_database_exception, "fork database root does not match block log head" );
       } else {
-         EOS_ASSERT( fork_db.root()->block_num == lib_num, fork_database_exception,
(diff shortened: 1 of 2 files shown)


libraries/chain/controller.cpp
```diff
@@ -330,8 +330,11 @@ struct controller_impl {
          // todo: move this check to startup so id does not have to be calculated
          EOS_ASSERT( root_id == log_head->calculate_id(), fork_database_exception, "fork database root does not match block log head" );
       } else {
-         EOS_ASSERT( fork_db.root()->block_num == lib_num, fork_database_exception,
-                     "empty block log expects the first appended block to build off a block that is not the fork database root" );
+         if( fork_db.root()->block_num != lib_num ) {
+            ilog( "Empty block log expects the first appended block to build off a block that is not the fork database root, terminating. root block number: ${block_num}, lib: ${lib_num}", ("block_num", fork_db.root()->block_num) ("lib_num", lib_num) );
+            shutdown();
```
heifner, line 335: This introduces unique error handling behavior for this one case. I think it would be better to handle the fork_database_exception differently in the producer_plugin.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -428,6 +428,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             chain_plugin::handle_bad_alloc();
          } catch ( boost::interprocess::bad_alloc& ) {
             chain_plugin::handle_db_exhaustion();
+         } catch ( const fork_database_exception& ) {
+            wlog("Cannot recover from fork_db exception. Shutting down.");
```
heifner, line 432: elog. Log the exception details. Add to the exception message the details you had in the log message.

libraries/chain/controller.cpp
linhuang-blockone: Was worried exit on any fork_database_exception would impact wider areas. But once fork_db becomes corrupt, nothing can be done. Code changed according to the comments. Thanks.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
linhuang-blockone: Agreed. Done.

heifner: Add to the EOS_ASSERT the details you had in the log message.

heifner (approved):

PR #10205, opened 2021-03-29
Fix for issue with security group test not able to progress through bootstrap

brianjohnson5972:
## Change Description
Needed to add a delay in the bootstrap sequence when TLS connections are configured because when the network is started up there is an unknown delay in the system of 30 seconds which is causing the producing nodes to not be ready to produce. Also added some small changes to the script as well.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -1509,6 +1509,8 @@ class Node(object):
 
     def getActivatedProtocolFeatures(self):
         latestBlockHeaderState = self.getLatestBlockHeaderState()
+        if "activated_protocol_features" not in latestBlockHeaderState or "protocol_features" not in latestBlockHeaderState["activated_protocol_features"]:
+            Utils.errorExit("getLatestBlockHeaderState did not return expected output, should contain [\"activated_protocol_features\"][\"protocol_features\"]: {}".format(latestBlockHeaderState))
         return latestBlockHeaderState["activated_protocol_features"]["protocol_features"]
 
     def modifyBuiltinPFSubjRestrictions(self, featureCodename, subjectiveRestriction={}):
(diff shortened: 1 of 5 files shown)


tests/Node.py
```diff
@@ -1424,16 +1424,50 @@ def isLibAdvancing():
             return self.getIrreversibleBlockNum() > currentLib
         return Utils.waitForTruth(isLibAdvancing, timeout)
 
+    def waitUntilBeginningOfProdTurn(self, producerName, timeout=30, sleepTime=0.4):
+        beginningOfProdTurnHead = 0
+        def isDesiredProdTurn():
+            beginningOfProdTurnHead = self.getHeadBlockNum()
```
jgiszczak, line 1430: This updates a local copy of `beginningOfProdTurnHead`, leaving the variable of the same name in the parent scope unchanged.

jgiszczak (changes requested):

tests/Node.py
brianjohnson5972: I changed it to return the value, and only set the variable where it was being used, it was never used after that, so I didn't reset the variable on subsequent calls

brianjohnson5972: I missed what you were meaning here, rather than in the original place it was being used.  Fixed it now.

jgiszczak (approved):

PR #10229, opened 2021-04-07
TLS connection timing issue fix

dimas1185:
## Change Description

tests fix for Amazon 2, CentOS and Ubuntu 16

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -647,7 +647,10 @@ namespace eosio {
 
       ~connection() {}
 
-      bool start_session(bool server);
+      //this function calls callback synchronously in case of non-ssl connection
+      //and asynchronously in case of ssl after asynchronous ssl handshake 
+      template<typename _Callback>
+      void start_session(bool server, _Callback callback);
 
       bool socket_is_open() const { return socket_open.load(); } // thread safe, atomic
       inline void set_socket_open(bool open = true) { socket_open = open; }
@@ -1086,7 +1089,8 @@ namespace eosio {
       return stat;
    }
 
-   bool connection::start_session(bool server) {
+   template<typename _Callback>
+   void connection::start_session(bool server, _Callback callback) {
       verify_strand_in_this_thread( strand, __func__, __LINE__ );
 
       update_endpoints();
@@ -1096,12 +1100,13 @@ namespace eosio {
       if( ec ) {
          fc_elog( logger, "connection failed (set_option) ${peer}: ${e1}", ("peer", peer_name())( "e1", ec.message() ) );
          close();
-         return false;
+         callback(false);
       } else {
 
          if (my_impl->ssl_enabled) {
             socket.ssl_socket()->async_handshake(server ? ssl::stream_base::server : ssl::stream_base::client,
-                                       boost::asio::bind_executor(strand, [c = shared_from_this(), socket=socket](const auto& ec){
+                                       boost::asio::bind_executor(strand, 
+                                       [c = shared_from_this(), socket=socket, callback](const auto& ec){
                                           //we use socket just to retain connection shared_ptr just in case it will be deleted
                                           //when we will need it inside start_read_message
                                           std::ignore = socket;
@@ -1118,14 +1123,15 @@ namespace eosio {
                                           fc_dlog( logger, "participant added: ${o}", ("o", c->participant_name()));
                                           
                                           c->start_read_message();
+                                          callback(true);
                                        }));
          }
          else {
             fc_dlog( logger, "connected to ${peer}", ("peer", peer_name()) );
             set_socket_open();
             start_read_message();
+            callback(true);
          }
-         return true;
       }
    }
 
@@ -2611,9 +2617,10 @@ namespace eosio {
              c = shared_from_this(), 
              socket=socket]( const boost::system::error_code& err, const tcp::endpoint& endpoint ) mutable {
             if( !err && socket.raw_socket().is_open() && &socket.raw_socket() == &c->socket.raw_socket() ) {
-               if( c->start_session(false) ) {
-                  c->send_handshake();
-               }
+               c->start_session(false, [c](bool success){
+                  if (success)
+                     c->send_handshake();
+               });
             } else {
                fc_elog( logger, "connection failed to ${peer}: ${error}", ("peer", c->peer_name())( "error", err.message()));
                c->close( false );
@@ -2652,10 +2659,12 @@ namespace eosio {
                   if( from_addr < max_nodes_per_host && (max_client_count == 0 || visitors < max_client_count)) {
                      fc_ilog( logger, "Accepted new connection: " + paddr_str );
                      new_connection->set_heartbeat_timeout( heartbeat_timeout );
-                     if( new_connection->start_session(true)) {
-                        std::lock_guard<std::shared_mutex> g_unique( connections_mtx );
-                        connections.insert( new_connection );
-                     }
+                     new_connection->start_session(true, [c = shared_from_this(), new_connection](bool success) {
+                        if (success) {
+                           std::lock_guard<std::shared_mutex> g_unique( c->connections_mtx );
+                           c->connections.insert( new_connection );
+                        }
+                     });
 
                   } else {
                      if( from_addr >= max_nodes_per_host ) {
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -653,7 +653,6 @@ namespace eosio {
       void start_session(bool server, _Callback callback);
 
       bool socket_is_open() const { return socket_open.load(); } // thread safe, atomic
-      inline void set_socket_open(bool open = true) { socket_open = open; }
       const string& peer_address() const { return peer_addr; } // thread safe, const
 
       void set_connection_type( const string& peer_addr );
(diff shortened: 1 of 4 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1097,32 +1100,38 @@ namespace eosio {
       if( ec ) {
          fc_elog( logger, "connection failed (set_option) ${peer}: ${e1}", ("peer", peer_name())( "e1", ec.message() ) );
          close();
-         return false;
+         callback(false);
```
brianjohnson5972, line 1103: should still return here, after callback is called

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3816,18 +3828,18 @@ namespace eosio {
       error_code ec;
 
       if (!ca.empty()){
-         fc_dlog(logger, "using verify file: ${p}", ("p", ca));
+         dlog("using verify file: ${p}", ("p", ca));
```
brianjohnson5972, line 3831: if you don't have a special reason for this fc_dlog to dlog changes and the ones below, please revert them

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1097,32 +1100,38 @@ namespace eosio {
...
+                                          fc_dlog( logger, "connected to ${peer}", ("peer", c->peer_name()) );
+                                          fc_dlog( logger, "participant added: ${o}", ("o", c->participant_name()));
+                                          
+                                          c->start_read_message();
+                                          callback(true);
                                        }));
          }
          else {
-            start_read();
+            fc_dlog( logger, "connected to ${peer}", ("peer", peer_name()) );
```
brianjohnson5972, line 1130: is there a reason that we couldn't still have this in a lambda, that both this path and the ssl path use? they still do exactly the same actions, except the ssl path has an extra log of "participant added...".

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1097,32 +1100,38 @@ namespace eosio {
...
+            set_socket_open();
+            start_read_message();
+            callback(true);
```
brianjohnson5972, line 1133: before calling lambda(see comment), need to call c->set_participating(true)

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -918,16 +922,14 @@ namespace eosio {
                
                std::lock_guard<std::shared_mutex> connection_guard(my_impl->connections_mtx);
                if(my_impl->security_group.is_in_security_group(participant)) {
-                  set_participating(true);
-                  peer_dlog(this, "participant added: ${o}", ("o", organization));
+                  participant_name_ = participant;
```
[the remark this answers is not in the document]
dimas1185: ok. btw so far it is not used except by ssl handshake

plugins/net_plugin/net_plugin.cpp
```diff
@@ -918,16 +922,14 @@ namespace eosio {
...
                   return true;
                }
             }
             
             peer_dlog(this, "received unauthorized participant: ${s}", ("s", organization));
          }
-         
-         set_participating(false);
+
```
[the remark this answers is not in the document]
dimas1185: the reason is we are doing this later in callback after ssl handshake execution

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1097,32 +1100,38 @@ namespace eosio {
...
                                           if (ec) {
+                                             c->set_participating(false);
                                              fc_elog( logger, "ssl handshake error: ${e}", ("e", ec.message()) );
                                              c->close();
                                              return;
                                           }
-                                          start_read();
+
+                                          c->set_socket_open();
+                                          c->set_participating(true);
```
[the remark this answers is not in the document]
dimas1185: I moved this code from `verify_certificate` for a reason. inside `verify_certificate` we do not guarantee that handshake was successful and we do not know if asio changed all internal structures already. So I moved it to later stage just because `socket_open` boolean and participating information can be used asynchronously.

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
brianjohnson5972: yes it is, in update_security_group

plugins/net_plugin/net_plugin.cpp
dimas1185: since method is asynchronous what will be meaning of return?

plugins/net_plugin/net_plugin.cpp
dimas1185: on some reason I don't see those lines in logs if `fc_dlog` is used. not sure why, need to investigate.

plugins/net_plugin/net_plugin.cpp
dimas1185: yes, we can, but we need to handle somehow `set_participating` logic that is unnecessary in case of non-ssl. I'll think of it.

plugins/net_plugin/net_plugin.cpp
heifner: That is because the `logger` is not initialized until plugin_startup and these are called in plugin_initialize

plugins/net_plugin/net_plugin.cpp
dimas1185: so we need `set_participating` called even in non-ssl scenario?

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
brianjohnson5972: is that in the error case, where we call close?

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: return means nothing later on gets processed.  Not a big deal, but since this is handling an error case early, it makes it clear that nothing later on in the method will be executed.

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
brianjohnson5972: based on your other comment, you don't need to add back in set_participating(true)

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: can you call c->set_participating(true) before calling the lambda, or is that problematic? This isn't absolutely necessary, but I like that it makes it clear that the same logic is happening for both (just later on for SSL)

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: so that sounds like you should use dlog then

tests/Node.py
```diff
@@ -1468,7 +1468,7 @@ def containsFeatures(self, features):
         return True
 
     def containsPreactivateFeature(self):
-        return containsFeatures(["PREACTIVATE_FEATURE"])
+        return self.containsFeatures(["PREACTIVATE_FEATURE"])
```
brianjohnson5972, line 1471: Thanks for fixing it

plugins/net_plugin/net_plugin.cpp
dimas1185: I think it adds additional complexity for caller, as it needs to catch errors in two places (via callback and via return value) so I removed return value. We always execute callback.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -918,12 +926,15 @@ namespace eosio {
...
             if (is_string_valid_name(organization)){
-               account_name participant{organization};
-               
+               participant_name_ = account_name{organization};
+
                std::lock_guard<std::shared_mutex> connection_guard(my_impl->connections_mtx);
-               if(my_impl->security_group.is_in_security_group(participant)) {
-                  participant_name_ = participant;
-                  return true;
+               if(my_impl->security_group.is_in_security_group(*participant_name_)) {
```
brianjohnson5972, line 932: using the verified temporary (has no use the rest of the life of the connection) flag is complex, instead move line 931 through 934 to the start_session logic, it will be much cleaner than caching the value here, and if a new security_group came in, between verify_certificate and the processing in start_session, we would be setting the wrong set_participating value.

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3932,7 +3934,7 @@ namespace eosio {
            "Number of worker threads in net_plugin thread pool" )
          ( "sync-fetch-span", bpo::value<uint32_t>()->default_value(def_sync_fetch_span), "number of blocks to retrieve in a chunk from any individual peer during synchronization")
          ( "use-socket-read-watermark", bpo::value<bool>()->default_value(false), "Enable experimental socket read watermark optimization")
-         ( "peer-log-format", bpo::value<string>()->default_value( "[\"${_name}\" ${_ip}:${_port}]" ),
+         ( "peer-log-format", bpo::value<string>()->default_value( "[\"${_name}\" ${_ip}:${_port}] " ),
```
brianjohnson5972, line 3937: undo this unneeded diff

plugins/net_plugin/net_plugin.cpp
```diff
@@ -881,6 +880,16 @@ namespace eosio {
          
          return true;
       }
+      inline void setup_participant() {
+         if(my_impl->ssl_enabled) {
+            account_name participant = participant_name_ ? *participant_name_ : account_name{};
+            bool participating = my_impl->security_group.is_in_security_group(participant);
```
brianjohnson5972, line 886: need to take the connection_mtx here for calling is_in_security_group

brianjohnson5972 (changes requested):

plugins/net_plugin/net_plugin.cpp
dimas1185: I did this because of default format lacks space and it harder to read.
so now it looks like this:
`["localhost:9877" 127.0.0.1:9877] preverified: [true] certificate subject: /C=US/ST=VA/L=Blockburg/O=node2/CN=test-domain2`
not like this:
`["localhost:9877" 127.0.0.1:9877]preverified: [true] certificate subject: /C=US/ST=VA/L=Blockburg/O=node2/CN=test-domain2`

plugins/net_plugin/net_plugin.cpp
dimas1185: I checked usage of this mutex and it is used to lock connections vector to iterate or add/remove new connections.
why do we need it to lock before call to `is_in_security_group` ?

brianjohnson5972 (approved):

PR #10237, opened 2021-04-11
fix db_modes_test pipeline fails, see epe889

softprofe:
## Change Description

Start 10237 description ......

EPE 889

In function 
run_expect_success() {
   run_nodeos "$@"
   local NODEOS_PID=$!
   sleep 10
   kill $NODEOS_PID
   wait $NODEOS_PID
}
wait throw exception, but kill not throw.  The biggest possible is the timing of wait. if the kill finished rapidly before call wait, so the wait $NODEOS_PID is an already killed PID so wait it will throw.    However the same fail didn't repro in my local PC. In my local Macbook, a case happened was that Nodes shutdown by it self. in this case, kill should throw instead of wait.  Because kill didn't throw but wait throws, I think it is very possible Nodes was killed before wait call.  in this case, don't let wait throw stop the test script.

End 10237 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/db_modes_test.sh
+++ b/tests/db_modes_test.sh
@@ -49,6 +49,11 @@ run_expect_success() {
    kill $NODEOS_PID
    rc=0
    wait $NODEOS_PID && rc=$? || rc=$?
+   if [[ rc -eq  127  || rc -eq 0 ]]; then
+      rc=0
+   else
+      exit rc
+   fi
 }
 
 run_and_kill() {
--- a/tests/db_modes_test.sh
+++ b/tests/db_modes_test.sh
@@ -51,9 +51,8 @@ run_expect_success() {
    wait $NODEOS_PID && rc=$? || rc=$?
    if [[ rc -eq  127  || rc -eq 0 ]]; then
       rc=0
-   else
-      exit rc
    fi
+   return $rc
 }
 
 run_and_kill() {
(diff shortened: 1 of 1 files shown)


pepe-invest-git:
pipiline

tests/db_modes_test.sh
```diff
@@ -47,7 +47,8 @@ run_expect_success() {
    local NODEOS_PID=$!
    sleep 10
    kill $NODEOS_PID
-   wait $NODEOS_PID
+   rc=0
+   wait $NODEOS_PID && rc=$? || rc=$?
```
jgiszczak, line 51: Once retrieved, test for macOS and explicitly return 0 if rc is 0 or 127.  If not macOS, just return rc.

jgiszczak (changes requested):

tests/db_modes_test.sh
softprofe: Hi Jonathan, actually ubuntu also return 127, if you try wait 98765, and the echo "$?"
jonathan.giszczak-ext:speech_balloon:  3:55 PM
Ah.  In that case, no need for the platform check.  Just check for 0 and 127 and return zero, otherwise return rc.

jgiszczak (approved):

PR #10245, opened 2021-04-12
Fix for Security Group on nodes just applying the block that was produced

brianjohnson5972:
## Change Description
Fixed problem where security group was not being set for any node that did not produce the block and only applied it.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/global_property_object.hpp
```diff
@@ -153,16 +153,23 @@ namespace eosio { namespace chain {
...
    }
 
    inline void set_gpo_extension(global_property_object&                        gpo,
                                  snapshot_global_property_object::extension_t&& extension) {
       std::visit(
           [&gpo](auto& ext) {
              gpo.proposed_security_group_block_num    = ext.proposed_security_group_block_num;
-             gpo.proposed_security_group_participants = std::move(ext.proposed_security_group_participants);
-             gpo.transaction_hooks                    = std::move(ext.transaction_hooks);
+             gpo.proposed_security_group_participants = {ext.proposed_security_group_participants.begin(),
```
dimas1185, line 166: I'm not familiar with boost containers but isn't same thing supposed to do copy constructor?

libraries/chain/include/eosio/chain/global_property_object.hpp
```diff
@@ -153,16 +153,23 @@ namespace eosio { namespace chain {
...
+                                                         ext.proposed_security_group_participants.end(),
+                                                         gpo.proposed_security_group_participants.key_comp(),
+                                                         gpo.proposed_security_group_participants.get_allocator()};
+             gpo.transaction_hooks                    = {ext.transaction_hooks.begin(),
+                                                         ext.transaction_hooks.end(),
+                                                         gpo.transaction_hooks.get_allocator()};
```
dimas1185, line 172: same here as for `proposed_security_group_participants ` comment

libraries/chain/include/eosio/chain/global_property_object.hpp
```diff
@@ -153,16 +153,23 @@ namespace eosio { namespace chain {
 
    inline snapshot_global_property_object::extension_t get_gpo_extension(const global_property_object& gpo) {
       return snapshot_global_property_object::extension_v0{
-          gpo.proposed_security_group_block_num, gpo.proposed_security_group_participants, gpo.transaction_hooks};
+          gpo.proposed_security_group_block_num,
+          {gpo.proposed_security_group_participants.begin(), gpo.proposed_security_group_participants.end()},
+          {gpo.transaction_hooks.begin(), gpo.transaction_hooks.end()}};
```
dimas1185, line 158: same here, why not to utilize copy-constructor here for `transaction_hooks` and `proposed_security_group_participants`?

libraries/chain/controller.cpp
```diff
@@ -1442,7 +1442,11 @@ struct controller_impl {
                }
 
                ++bb._pending_block_header_state.security_group.version;
-               bb._pending_block_header_state.security_group.participants = std::move(gpo.proposed_security_group_participants);
+               bb._pending_block_header_state.security_group.participants = {
+                  gpo.proposed_security_group_participants.begin(),
+                  gpo.proposed_security_group_participants.end(),
+                  bb._pending_block_header_state.security_group.participants.key_comp(),
+                  bb._pending_block_header_state.security_group.participants.get_allocator()};
```
dimas1185, line 1449: please correct me if I wrong, but we do not assign custom comparator or allocator anywhere, so constructor supposed to instantiate one from template parameter and we don't need to make here a copy from existing comparator and allocator

libraries/chain/controller.cpp
```diff
@@ -2274,7 +2282,10 @@ struct controller_impl {
 
       db.modify(gpo, [&proposed_participants, cur_block_num](auto& gp) {
          gp.proposed_security_group_block_num    = cur_block_num;
-         gp.proposed_security_group_participants = std::move(proposed_participants);
+         gp.proposed_security_group_participants = {proposed_participants.begin(),
+                                                    proposed_participants.end(),
+                                                    gp.proposed_security_group_participants.key_comp(),
+                                                    gp.proposed_security_group_participants.get_allocator()};
```
dimas1185, line 2288: same here regarding explicitly specifying allocator and comparator

dimas1185 (commented):
see my comments in files

tests/privacy_simple_network.py
```diff
@@ -0,0 +1,280 @@
...
+    Print("Validating system accounts after bootstrap")
+    cluster.validateAccounts(None)
+
+    def getNodes():
+        global cluster
+        nodes = [cluster.biosNode]
+        nodes.extend(cluster.getNodes())
+        return nodes
+
+#    cluster.biosNode.kill(signal.SIGTERM)
```
dimas1185, line 88: remove unnecessary comment

tests/privacy_simple_network.py
```diff
@@ -0,0 +1,280 @@
...
+        ignWallet = cluster.walletMgr.create("ignition")  # will actually just look up the wallet
+        cluster.walletMgr.importKey(newAcc, ignWallet)
+
+    # numAccounts = 4
+    # testAccounts = Cluster.createAccountKeys(numAccounts)
+    # accountPrefix = "testaccount"
+    # for i in range(numAccounts):
+    #     testAccount = testAccounts[i]
+    #     testAccount.name  = accountPrefix + str(i + 1)
+    #     createAccount(testAccount)
```
dimas1185, line 114: remove unnecessary comments

tests/privacy_simple_network.py
```diff
@@ -0,0 +1,280 @@
...
+
+    blockProducer = None
+
+    def verifyInSync(producerNum):
```
dimas1185, line 118: this function seems to be generic and I see you are using it in `privacy_startup_network.py` . What about moving it to `Cluster.py`

libraries/chain/include/eosio/chain/global_property_object.hpp
brianjohnson5972: They are 2 different types now, so you cannot use the copy constructor

libraries/chain/include/eosio/chain/global_property_object.hpp
brianjohnson5972: See other comment.

libraries/chain/include/eosio/chain/global_property_object.hpp
brianjohnson5972: see above

libraries/chain/include/eosio/chain/global_property_object.hpp
dimas1185: got it. if you don't need allocator and comparator to be custom instead of instantiating it from template parameter by default -  please remove.

libraries/chain/include/eosio/chain/global_property_object.hpp
brianjohnson5972: cleaning these up by using a set_*** method that just takes begin and end iterators.

libraries/chain/controller.cpp
brianjohnson5972: yes, this should just use begin and end iterator

libraries/chain/controller.cpp
brianjohnson5972: The global_property_object ctor macro sets the allocator, so the comparator has to come along for the ride since it cannot be defaulted if the allocator is not defaulted

tests/privacy_simple_network.py
```diff
@@ -0,0 +1,280 @@
...
+        error = None
+        for prod in producers:
+            if prod == producers[producerNum]:
+                continue
+
+            if prod.waitForBlock(headBlockNum, reportInterval = 1) is None:
+                error = "Producer node failed to get block number {}. Current node info: {}".format(headBlockNum, json.dumps(prod.getInfo(), indent=4, sort_keys=True))
+                break
+
+            if prod.getBlock(headBlockId) is None:
```
dimas1185, line 137: I see that intent here was to wait for block by id to ensure there are no forks. Maybe we can add this code to `Node.py` ? We could reuse method `waitForBlock` and if block parameter is string we consider this was block id otherwise block number. Or Alternatively we could add `waitForBlockId`

libraries/chain/include/eosio/chain/global_property_object.hpp
brianjohnson5972: Are you saying that there is a constructor that allows you to default the comparator, while setting the allocator and providing iterators? If there is point me to it. Allocator has to come from the original container in the global_property_object.
If you are wondering why transaction_hook doesn't worry about comparator, that is because vector doesn't have a comparator.

tests/privacy_simple_network.py
brianjohnson5972: I think that is a good idea, but would rather not tack that on here. The issue is that code relies on block numbers, so either we would have to pull in logic from nodeos to get the block num from the id, or else pass that in as well.

libraries/chain/controller.cpp
dimas1185: thanks, it was not straightforward, now I see it. it is `chainbase::allocator<Allocator>`

tests/privacy_simple_network.py
brianjohnson5972: Moved along with some other methods

dimas1185 (approved):
looks good.
the only minor thing to consider if you'll do any other code changes is to mark inline helper setters:
`set_proposed_security_group_participants` and `set_transaction_hooks`

huangminghuang (approved):

PR #10250, opened 2021-04-15
Rel 2.0.x: Subjective CPU billing cleos enhancement and adding subjective_cpu_bill to /v1/chain/get_account result

linhuang-blockone:
## Change Description

This PR adds `subjective_cpu_bill` to `/v1/chain/get_account` result and adds a new entity `subjective cpu bandwidth` to `cleos` `get_account` command. An example output looks like

```
permissions:
     owner     1:    1 EOS8Dq1KosJ9PMn1vKQK3TbiihgfUiDBUsz471xaCE6eYUssPB1KY
        active     1:    1 EOS8Dq1KosJ9PMn1vKQK3TbiihgfUiDBUsz471xaCE6eYUssPB1KY
memory:
     quota:       unlimited  used:      2.66 KiB

net bandwidth:
     used:               unlimited
     available:          unlimited
     limit:              unlimited

cpu bandwidth:
     used:               unlimited
     available:          unlimited
     limit:              unlimited

subjective cpu bandwidth:
     used:               186 us

```
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [x] Other

Manual testing.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

A new `subjective cpu bandwidth:` is added to `cleos get_account ` command to display the used subjective cpu of the account. Please see above example output.

### DIFF
--- a/plugins/chain_plugin/CMakeLists.txt
+++ b/plugins/chain_plugin/CMakeLists.txt
@@ -4,7 +4,7 @@ add_library( chain_plugin
              chain_plugin.cpp
              ${HEADERS} )
 
-target_link_libraries( chain_plugin eosio_chain appbase )
-target_include_directories( chain_plugin PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_SOURCE_DIR}/../chain_interface/include" "${CMAKE_CURRENT_SOURCE_DIR}/../../libraries/appbase/include")
+target_link_libraries( chain_plugin producer_plugin eosio_chain appbase )
+target_include_directories( chain_plugin PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_SOURCE_DIR}/../chain_interface/include" "${CMAKE_CURRENT_SOURCE_DIR}/../../libraries/appbase/include" "${CMAKE_CURRENT_SOURCE_DIR}/../producer_plugin/include")
 
 add_subdirectory( test )
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1,4 +1,5 @@
 #include <eosio/chain_plugin/chain_plugin.hpp>
+#include <eosio/producer_plugin/producer_plugin.hpp>
 #include <eosio/chain/fork_database.hpp>
 #include <eosio/chain/block_log.hpp>
 #include <eosio/chain/exceptions.hpp>
@@ -2435,6 +2436,11 @@ read_only::get_account_results read_only::get_account( const get_account_params&
    result.net_limit = rm.get_account_net_limit_ex( result.account_name, greylist_limit).first;
    result.cpu_limit = rm.get_account_cpu_limit_ex( result.account_name, greylist_limit).first;
    result.ram_usage = rm.get_account_ram_usage( result.account_name );
+
+   producer_plugin* producer_plug = app().find_plugin<producer_plugin>();
+   if ( producer_plug ) {
+      result.subjective_cpu_bill = producer_plug->get_subjective_bill( result.account_name, fc::time_point::now() );
+   }
 
    const auto& permissions = d.get_index<permission_index,by_owner>();
    auto perm = permissions.lower_bound( boost::make_tuple( params.account_name ) );
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -186,6 +186,7 @@ public:
 
 
    fc::optional<chain_apis::account_query_db>                        _account_query_db;
+   const producer_plugin* producer_plug;
 };
 
 chain_plugin::chain_plugin()
@@ -1145,6 +1146,9 @@ void chain_plugin::plugin_startup()
    EOS_ASSERT( my->chain_config->read_mode != db_read_mode::IRREVERSIBLE || !accept_transactions(), plugin_config_exception,
                "read-mode = irreversible. transactions should not be enabled by enable_accept_transactions" );
    try {
+      my->producer_plug = app().find_plugin<producer_plugin>();
+      EOS_ASSERT(my->producer_plug, plugin_exception, "Failed to find producer_plugin");
+
       auto shutdown = [](){ return app().is_quiting(); };
       if (my->snapshot_path) {
          auto infile = std::ifstream(my->snapshot_path->generic_string(), (std::ios::in | std::ios::binary));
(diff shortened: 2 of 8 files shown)


plugins/chain_plugin/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_library( chain_plugin
              chain_plugin.cpp
              ${HEADERS} )
 
-target_link_libraries( chain_plugin eosio_chain appbase )
-target_include_directories( chain_plugin PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_SOURCE_DIR}/../chain_interface/include" "${CMAKE_CURRENT_SOURCE_DIR}/../../libraries/appbase/include")
+target_link_libraries( chain_plugin producer_plugin eosio_chain appbase )
```
heifner, line 7: I had no idea that would work:
https://cmake.org/cmake/help/latest/command/target_link_libraries.html#cyclic-dependencies-of-static-libraries

However, I'm not sure if we should do this or use the method providers of plugin_interface. I'm personally not a huge fan of the plugin_interface but that is the current way this is done. Thoughts @larryk85

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2436,6 +2437,11 @@ read_only::get_account_results read_only::get_account( const get_account_params&
    result.cpu_limit = rm.get_account_cpu_limit_ex( result.account_name, greylist_limit).first;
    result.ram_usage = rm.get_account_ram_usage( result.account_name );
 
+   producer_plugin* producer_plug = app().find_plugin<producer_plugin>();
```
heifner, line 2440: The find_plugin is actually rather expensive. This should be setup in plugin init or startup and asserted to be there. Assuming you don't change this to use plugin_interface methods.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -160,6 +160,8 @@ class read_only {
       account_resource_limit     cpu_limit;
       int64_t                    ram_usage = 0;
 
+      uint32_t                   subjective_cpu_bill = 0;
```
heifner, line 163: This needs to be optional so that old cleos will work with new nodeos. Please test with new cleos <-> old nodeos and old cleos <-> new nodeos.

heifner (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
linhuang-blockone: Thanks. Will change and test different versions of cleos.

plugins/chain_plugin/chain_plugin.cpp
linhuang-blockone: Thanks. As this is in read_only class, I will initialize it in read_only constructor. I did this too rush last night.

plugins/chain_plugin/CMakeLists.txt
linhuang-blockone: I will look into how to use plugin_interface.

linhuang-blockone: Do you know an existing code example? I saw register and subscribe of method providers. But it is immediately clear to me how to use it.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
linhuang-blockone: I have tested with different versions of nodeos and cleos. cleos worked in all combinations.

programs/cleos/main.cpp
```diff
@@ -2304,13 +2304,20 @@ void get_account( const string& accountName, const string& coresym, bool json_fo
...
       std::cout << indent << std::left << std::setw(11) << "available:" << std::right << std::setw(18) << to_pretty_time( res.cpu_limit.available ) << "\n";
       std::cout << indent << std::left << std::setw(11) << "limit:"     << std::right << std::setw(18) << to_pretty_time( res.cpu_limit.max ) << "\n";
       std::cout << std::endl;
 
+      std::cout << "subjective cpu bandwidth:" << std::endl;
+      uint32_t subjective_cpu_bill = 0;
+      if( res.subjective_cpu_bill.valid() ) {
+         subjective_cpu_bill = *res.subjective_cpu_bill;
+      }
+      std::cout << indent << std::left << std::setw(11) << "used:"      << std::right << std::setw(18) << to_pretty_time( subjective_cpu_bill ) << "\n";
```
heifner, line 2318: I think it would be better to not print the section at all if not available. Most of the time it would be true to report 0 as it would not have subjective billing. However, we have released versions that have subjective billing but will not return a subjective bill amount.

linhuang-blockone: Will change.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
arhag: I would think this would make more sense as a field in its own structure similar to `net_limit` and `cpu_limit`. This is basically the `used` field in `account_resource_limit`. And eventually, not now, we will want the `max` and `available` too.

linhuang-blockone: Thanks. I will do that.

programs/cleos/main.cpp
linhuang-blockone: I will not log only when it is not valid.

linhuang-blockone: get_subjective_bill in subjective_billing.hpp always returns a value (returning 0 when subjective billing is disabled or not available). Looks like we need another API to tell if subject billing amount is present. Do that in 2.1.x?

heifner: There is a difference from a nodeos not having the code of this PR vs having subject billing disabled. If nodeos has this PR code then 0 is the correct response to print as that indicates either there is no subjective billing on the node or the account has no subjective billing applied to it. Either way there is no subjective billing for that account.

As for an indication if a node has subjective billing enabled or not at all that should be a differ api call. That api call if added would need to indicate if it is enable for API calls or P2P calls or both. In the future that api call might also provide information on any configurable values of subjective billing.

linhuang-blockone: Thanks for explanation.

heifner (approved):

PR #10252, opened 2021-04-15
Rodeos timeout fix

heifner:
## Change Description

```
if(! impl_->read.pending)
    BOOST_VERIFY(
            expiry_time) == 0);
if(! impl_->write.pending)
    BOOST_VERIFY(
            expiry_time) == 0);
```
and the comment from the code

```
*/
```

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

`wql-idle-timeout` option is now ignored.

### DIFF
--- a/programs/rodeos/wasm_ql_http.cpp
+++ b/programs/rodeos/wasm_ql_http.cpp
@@ -405,6 +405,10 @@ class http_session {
                 : self(self), msg(std::move(msg)) {}
 
             void operator()() {
+               // expires_after controls two timers (read & write) which are only reset if there are no pending calls
+               // set expires_after here before write and also below before read to make sure both timers are reset
+               self.derived_session().stream.expires_after(std::chrono::milliseconds(self.http_config->idle_timeout_ms));
+
                http::async_write(
                      self.derived_session().stream, msg,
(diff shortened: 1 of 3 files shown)


spoonincode (approved):
After going back and forth on this in DM, I can't help but wonder if just removing this really is a better option.
* this `expires_after()` behavior smells like a trap and scares me away from ever using that myself in the future
* you described the fix (I think this fix) as kludgy
* this may be the only place we have such a timeout -- http_plugin doesn't
* another config option that how do we expect users to know what to do with :woman_shrugging: 
* I think the only "hard" boost 1.70 feature being used in :package:

heifner:
>  I can't help but wonder if just removing this really is a better option.

You convinced me. Removed. If/when we add back support I think we should just use our own timers as you suggest.

spoonincode (approved):

PR #10253, opened 2021-04-16
Added Privacy Test Scenario #3 to privacy-simple-network.py

brianjohnson5972:
## Change Description
Added more complex scenarios and covered the Test Scenario #3. Also made improvements to Node and Cluster and pulled in common code.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Cluster.py
+++ b/tests/Cluster.py
@@ -538,6 +538,8 @@ class Cluster(object):
         if onlyBios:
             self.nodes=[biosNode]
 
+        self.totalNodes = totalNodes
+
         # ensure cluster node are inter-connected by ensuring everyone has block 1
         Utils.Print("Cluster viability smoke test. Validate every cluster node has block 1. ")
         if not self.waitOnClusterBlockNumSync(1):
@@ -1031,24 +1033,6 @@ class Cluster(object):
             producerKeys[m]=keys
 
         return producerKeys
-
-    @staticmethod
-    def parseProducers(nodeNum):
-        """Parse node config file for producers."""
-
-        configFile=Utils.getNodeConfigDir(nodeNum, "config.ini")
-        if Utils.Debug: Utils.Print("Parsing config file %s" % configFile)
-        configStr=None
-        with open(configFile, 'r') as f:
-            configStr=f.read()
-
-        pattern=r"^\s*producer-name\s*=\W*(\w+)\W*$"
-        producerMatches=re.findall(pattern, configStr, re.MULTILINE)
-        if producerMatches is None:
-            if Utils.Debug: Utils.Print("Failed to find producers.")
-            return None
-
-        return producerMatches
 
     @staticmethod
     def parseClusterKeys(totalNodes):
@@ -1877,3 +1861,12 @@ class Cluster(object):
         if error:
             self.reportInfo()
             Utils.errorExit(error)
+
+    def getParticipantNum(self, nodeToIdentify):
+        num = 0
+        for node in self.nodes:
+            if node == nodeToIdentify:
+                return num
+            num += 1
+        assert nodeToIdentify == self.biosNode
+        return self.totalNodes
--- a/tests/Cluster.py
+++ b/tests/Cluster.py
@@ -1870,3 +1870,14 @@ class Cluster(object):
             num += 1
         assert nodeToIdentify == self.biosNode
         return self.totalNodes
+
+    def getProducingNodeIndex(self, blockProducer):
+        featureProdNum = 0
+        while featureProdNum < pnodes:
+            if blockProducer in self.nodes[featureProdNum].getProducers():
+                return featureProdNum
+
+            featureProdNum += 1
+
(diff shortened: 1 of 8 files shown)


tests/privacy_simple_network.py
```diff
@@ -160,33 +163,120 @@ def verifyNonParticipants(transId):
...
 
+    # one by one remove each (original) nonParticipant from the security group
+    while len(participants) > pnodes:
+        publishTrans = remFromSg()
+        verifySecurityGroup(publishTrans)
+        cluster.reportInfo()
+
+
+    # if we have more than 1 api node, we will add and remove all those nodes in bulk, if not it is just a repeat of the above test
+    if totalNodes > pnodes + 1:
```
dimas1185, line 218: you have `apiNodes` and `nonParticipants` list with all API nodes. I think it will be more clear to use here `len(apiNodes)` or `len(nonParticipants)`. btw, do you need two arrays for API nodes?

tests/Node.py
```diff
@@ -1687,3 +1722,24 @@ def waitForIrreversibleBlockProducedBy(self, producer, startBlockNum=0, retry=10
             retry = retry - 1
             startBlockNum = latestBlockNum + 1
         return False
+
+    @staticmethod
+    def parseProducers(nodeNum):
```
dimas1185, line 1727: it looks like you didn't change that method but moved it. can you try to get it back? just for sake of keeping history of changes.

brianjohnson5972: I moved it here since it makes more logical sense to be associated with the Node.  Right now there are no dependencies from Node on Cluster, which I think is logically what we want, since a Cluster contains a Node, so I'd like to keep it here.

dimas1185 (approved):

tests/privacy_simple_network.py
```diff
@@ -122,20 +108,37 @@ def publishContract(account, file, waitForTransBlock=False):
...
-        action += ']]'
-
-        Utils.Print("adding {} to the security group".format(action))
-        trans = producers[0].pushMessage(cluster.eosioAccount.name, "add", action, "--permission eosio@active")
-        Utils.Print("add trans: {}".format(json.dumps(trans, indent=4, sort_keys=True)))
-        trans = producers[0].pushMessage(cluster.eosioAccount.name, "publish", "[0]", "--permission eosio@active")
+    # this is passed to limit the number of add/remove table entries are processed, but using it here to keep from getting duplicate transactions
+    publishProcessNum = 20
+    def security_group(addNodeNums=[], removeNodeNums=[]):
+        def createAction(nodeNums):
```
jgiszczak, line 114:
```
def createAction(nodeNums):
    return None if len(nodeNums) == 0 else '[[{}]]'.format(','.join(['"{}"'.format(Node.participantName(nodeNum)) for nodeNum in nodeNums]))
```

tests/privacy_simple_network.py
```diff
@@ -160,33 +163,120 @@ def verifyNonParticipants(transId):
...
+
+        done = False
+        # keep adding and removing nodes till we are done
+        while not done:
+            if blockNum:
+                participants[0].waitForNextBlock()
+
+            while not done and len(participants) > pnodes:
+                publishTrans = remFromSg()
+                Utils.Print("publishTrans: {}".format(json.dumps(publishTrans, indent=2)))
```
jgiszczak, line 249: Is there a reason indent doesn't match the indent used elsewhere?

jgiszczak (changes requested):

tests/privacy_simple_network.py
dimas1185: those one liners are hard to read.
we can have at least something like that:
```
def createAction(nodeNums):
     if len(nodeNums) == 0 
         return None
     return '[[{}]]'.format(','.join(['"{}"'.format(Node.participantName(n)) for n in nodeNums]))
 ```

File: tests/Cluster.py [hunk 5]
```diff
@@ -1877,3 +1861,23 @@ def verifyInSync(self, sourceNodeNum=0, specificNodes=None):
         if error:
             self.reportInfo()
             Utils.errorExit(error)
+
+    def getParticipantNum(self, nodeToIdentify):
+        num = 0
+        for node in self.nodes:
```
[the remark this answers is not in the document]
dimas1185: or alternatively we can do something like this:
```
range_nodes = range(len(self.nodes))
nodeDict = dict(zip(self.nodes, range_nodes)) | {self.biosNode : self.totalNodes}
return nodeDict[nodeToIndentify]
```

PR #10260, opened 2021-04-19
fix read only nodeos read mode is IRREVERSIBLE terminate block can go futher than LIB see epe915

softprofe:
## Change Description

Start 10260 description ......

EPE 915

When running nodeos in irreversible mode `read-mode = irreversible`, I would expect the terminate-at-block option to stop on an irreversible block not on a non-irreversible block.

The fix is add check in function
producer_plugin_impl::start_block_result producer_plugin_impl::start_block() the check read_only and read-mode = irreversible condition

End 10260 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1415,7 +1415,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
 
    const auto& hbs = chain.head_block_state();
 
-   if( chain.get_terminate_at_block() > 0 && chain.get_terminate_at_block() < hbs->block_num ) {
+   if( ( chain.get_terminate_at_block() > 0 && chain.get_terminate_at_block() < hbs->block_num ) || (chain.get_config().read_only && chain.get_config().read_mode == db_read_mode::IRREVERSIBLE && chain.last_irreversible_block_num() < hbs->block_num )) {
```
heifner, line 1418: How is it possible that `chain.last_irreversible_block_num() < hbs->block_num` is ever true?
I can't see how this fixes the issue.
Where you able to reproduce the reported issue?

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
softprofe: Can't repro using the way you told me, I wrote the steps under EPE915 and @heifner , but no response from you.

softprofe: Today reproduced, and change the code , and test it works. and then commit again.

heifner (changes requested):
Please add a test that verifies this fix. Setup a number of producers so that there is more than 1 diff between head and lib.

libraries/chain/controller.cpp
```diff
@@ -1879,7 +1879,7 @@ struct controller_impl {
          block_state_ptr bsp = block_state_future.get();
          const auto& b = bsp->block;
 
-         if( conf.terminate_at_block > 0 && conf.terminate_at_block < b->block_num() ) {
+         if( conf.terminate_at_block > 0 && ( (conf.read_mode != db_read_mode::IRREVERSIBLE && conf.terminate_at_block < b->block_num() ) ||  ( conf.read_mode == db_read_mode::IRREVERSIBLE && conf.terminate_at_block <= self.last_irreversible_block_num() ) )  ) {
```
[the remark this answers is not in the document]
softprofe: Add two line breaks.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1415,7 +1415,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
 
    const auto& hbs = chain.head_block_state();
 
-   if( chain.get_terminate_at_block() > 0 && chain.get_terminate_at_block() < hbs->block_num ) {
+   if( chain.get_terminate_at_block() > 0 && ( ( chain.get_config().read_mode != db_read_mode::IRREVERSIBLE && chain.get_terminate_at_block() < hbs->block_num ) || ( chain.get_config().read_mode == db_read_mode::IRREVERSIBLE && chain.get_terminate_at_block() <= chain.last_irreversible_block_num() ) ) ) {
```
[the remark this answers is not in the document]
softprofe: Add two line breaks.

libraries/chain/controller.cpp
```diff
@@ -1879,7 +1879,9 @@ struct controller_impl {
          block_state_ptr bsp = block_state_future.get();
          const auto& b = bsp->block;
 
-         if( conf.terminate_at_block > 0 && conf.terminate_at_block < b->block_num() ) {
+         if( conf.terminate_at_block > 0 &&
+            ( (conf.read_mode != db_read_mode::IRREVERSIBLE && conf.terminate_at_block < b->block_num() ) ||
+              ( conf.read_mode == db_read_mode::IRREVERSIBLE && conf.terminate_at_block <= self.last_irreversible_block_num() ) )  ) {
```
brianjohnson5972, line 1884: shouldn't this be '<', like the head/read only mode check, so that it terminates after applying the block, rather than before applying the block?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1415,7 +1415,9 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
 
    const auto& hbs = chain.head_block_state();
 
-   if( chain.get_terminate_at_block() > 0 && chain.get_terminate_at_block() < hbs->block_num ) {
+   if( chain.get_terminate_at_block() > 0 &&
+       ( ( chain.get_config().read_mode != db_read_mode::IRREVERSIBLE && chain.get_terminate_at_block() < hbs->block_num ) ||
+         ( chain.get_config().read_mode == db_read_mode::IRREVERSIBLE && chain.get_terminate_at_block() <= chain.last_irreversible_block_num() ) ) ) {
```
brianjohnson5972, line 1420: see other comment about '<' instead of '<="

libraries/chain/controller.cpp
softprofe: After test, it is currently works well.
let's assume terminate_at_block is 50.
In case of conf.read_mode != db_read_mode::IRREVERSIBLE,  when b->block_num() is 51, stop, then the chain already have 50.  
In case of conf.read_mode == db_read_mode::IRREVERSIBLE, when self.last_irreversible_block_num()  is 50, then stop, this time the chain already has LIB 50, and the head is a bigger number we don't care. 

Where's wrong you think?

softprofe:
> Please add a test that verifies this fix. Setup a number of producers so that there is more than 1 diff between head and lib.

As I'm going to add a test case, I found a test case have conflict with what we are doing. in unittest/forked_tests.cpp/
the read_mode test let producer's LIB equal to read_mode IRREVERSIBLE node's head. 
So, please make sure which idea is right.

378 BOOST_AUTO_TEST_CASE( read_modes ) try {
379    tester c;
380    c.produce_block();
381    c.produce_block();
382    auto r = c.create_accounts( {"dan"_n,"sam"_n,"pam"_n} );
383    c.produce_block();
384    auto res = c.set_producers( {"dan"_n,"sam"_n,"pam"_n} );
385    c.produce_blocks(200);
386    auto head_block_num = c.control->head_block_num();
387    auto last_irreversible_block_num = c.control->last_irreversible_block_num();
388
389    tester head(setup_policy::none, db_read_mode::HEAD);
390    push_blocks(c, head);
391    BOOST_CHECK_EQUAL(head_block_num, head.control->fork_db_head_block_num());
392    BOOST_CHECK_EQUAL(head_block_num, head.control->head_block_num());
393
394    tester read_only(setup_policy::none, db_read_mode::READ_ONLY);
395    push_blocks(c, read_only);
396    BOOST_CHECK_EQUAL(head_block_num, read_only.control->fork_db_head_block_num());
397    BOOST_CHECK_EQUAL(head_block_num, read_only.control->head_block_num());
398
399    tester irreversible(setup_policy::none, db_read_mode::IRREVERSIBLE);
400    push_blocks(c, irreversible);
401    BOOST_CHECK_EQUAL(head_block_num, irreversible.control->fork_db_pending_head_block_num());
402    BOOST_CHECK_EQUAL(last_irreversible_block_num, irreversible.control->fork_db_head_block_num());
403    BOOST_CHECK_EQUAL(last_irreversible_block_num, irreversible.control->head_block_num());
404
405 } FC_LOG_AND_RETHROW()

heifner:
> As I'm going to add a test case, I found a test case have conflict with what we are doing. in unittest/forked_tests.cpp/
> the read_mode test let producer's LIB equal to read_mode IRREVERSIBLE node's head.
> So, please make sure which idea is right.

> 399 tester irreversible(setup_policy::none, db_read_mode::IRREVERSIBLE);
> 400 push_blocks(c, irreversible);
> 401 BOOST_CHECK_EQUAL(head_block_num, irreversible.control->fork_db_pending_head_block_num());
> 402 BOOST_CHECK_EQUAL(last_irreversible_block_num, irreversible.control->fork_db_head_block_num());
> 403 BOOST_CHECK_EQUAL(last_irreversible_block_num, irreversible.control->head_block_num());

This test is correct. It indicates when running in IRREVERSIBLE mode that the head_block_num & fork_db_head_block_num are aways LIB. The existing code for terminate_at_block checks current block not head_block_num. I do think this illustrates a cleaner fix for this issue. The terminate_at_block should be compared to head_block_num and not the current block being processed.

You need to add a multi-producer test that shows terminate_at_block works for all modes: SPECULATIVE, HEAD, READ_ONLY, IRREVERSIBLE

softprofe:
> > As I'm going to add a test case, I found a test case have conflict with what we are doing. in unittest/forked_tests.cpp/
> > the read_mode test let producer's LIB equal to read_mode IRREVERSIBLE node's head.
> > So, please make sure which idea is right.
> 
> > 399 tester irreversible(setup_policy::none, db_read_mode::IRREVERSIBLE);
> > 400 push_blocks(c, irreversible);
> > 401 BOOST_CHECK_EQUAL(head_block_num, irreversible.control->fork_db_pending_head_block_num());
> > 402 BOOST_CHECK_EQUAL(last_irreversible_block_num, irreversible.control->fork_db_head_block_num());
> > 403 BOOST_CHECK_EQUAL(last_irreversible_block_num, irreversible.control->head_block_num());
> 
> This test is correct. It indicates when running in IRREVERSIBLE mode that the head_block_num & fork_db_head_block_num are aways LIB. The existing code for terminate_at_block checks current block not head_block_num. I do think this illustrates a cleaner fix for this issue. The terminate_at_block should be compared to head_block_num and not the current block being processed.
> 
> You need to add a multi-producer test that shows terminate_at_block works for all modes: SPECULATIVE, HEAD, READ_ONLY, IRREVERSIBLE

Python test script added, you can review now.

softprofe:
> Please add a test that verifies this fix. Setup a number of producers so that there is more than 1 diff between head and lib.
Done.

heifner:
> > Please add a test that verifies this fix. Setup a number of producers so that there is more than 1 diff between head and lib.
> > Done.

Sorry, I have not had time to review this. In general it seems to duplicate a lot of python code, can the code duplication be factored out?

softprofe:
> > > Please add a test that verifies this fix. Setup a number of producers so that there is more than 1 diff between head and lib.
> > > Done.
> 
> Sorry, I have not had time to review this. In general it seems to duplicate a lot of python code, can the code duplication be factored out?

Ok, removed some unused duplicated function, but some code were currently used can't remove. As these code are not from node.py  cluster.py or testUtils.py so, it's hard to reuse.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+   def stopProdNode():
+      if not producingNode.killed:
+         producingNode.kill(signal.SIGTERM)
+
+   def startProdNode():
+      if producingNode.killed:
+         relaunchNode(producingNode)
+
+   # Give some time for it to produce, so lib is advancing
+   waitForBlksProducedAndLibAdvanced()
```
dimas1185: we have similar method in `Node.waitForLibToAdvance`

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+   # Kill all nodes, so we can test all node in isolated environment
+   for clusterNode in cluster.nodes:
+      clusterNode.kill(signal.SIGTERM)
+   cluster.biosNode.kill(signal.SIGTERM)
```
dimas1185: there is a method for killing them all:
`Cluster.killall`

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+   producingNodeId = 0
+   producingNode = cluster.getNode(producingNodeId)
+
+   def stopProdNode():
+      if not producingNode.killed:
+         producingNode.kill(signal.SIGTERM)
+
+   def startProdNode():
+      if producingNode.killed:
```
dimas1185: if you expect it not always to be killed, add warning there or other message or assert, depending on your needs.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+         0:"--enable-stale-production",
+         4:"--read-mode irreversible",
+         6:"--read-mode irreversible",
+         9:"--plugin eosio::producer_api_plugin"})
+
+   producingNodeId = 0
+   producingNode = cluster.getNode(producingNodeId)
+
+   def stopProdNode():
+      if not producingNode.killed:
```
dimas1185: see my comment under `startProdNode`

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+         testResultMsgs.append(resultDesc)
+         # Kill node after use
+         if not nodeToTest.killed: nodeToTest.kill(signal.SIGTERM)
+      return testResult
+
+
+   # 1th test case: Replay in irreversible mode with reversible blks while connected to producing node
+   # Expectation: Node replays and launches successfully
+   #              and the head and lib should be no less than termAtBlockNum
+   def replayInIrrModeWithRevBlksAndConnectedToProdNodeTermAtBlock(nodeIdOfNodeToTest, nodeToTest):
```
dimas1185: I don't see `nodeIdOfNodeToTest` is used somewhere in the function but even if it does, there is a property for this:
`Node.nodeId`
So you don't need another parameter.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+         waitForBlksProducedAndLibAdvanced() # Wait
+         infoprod = producingNode.getInfo()
+         libprod = int(infoprod["last_irreversible_block_num"])
+         termAtBlockNum = libprod + termAtFutureBlockNum
+        
+         relaunchTestNode(nodeToTest, chainArg=" --read-mode irreversible  --terminate-at-block=" + str(termAtBlockNum))
+         Utils.Print("Test node begin recevie from producing node and is expected to stop at or little bigger than block number " + str(termAtBlockNum) )
+
+         while True :
+             info = nodeToTest.getInfo()
```
dimas1185: I think you can just use here `Node.waitForHeadToAdvance` with short timeout

dimas1185: Also, if you need to know exactly the very last head & lib after shutdown, you can parse blocks.log, there are methods for this:
`Cluster.getBlockLog`.
That way you'll eliminate any flakiness.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+         assert forkDbHead >= head, "Fork db head ({}) should be larger or equal to the head ({})".format(forkDbHead, head)
+      finally:
+         stopProdNode()
+         Utils.Print("End replayInIrrModeWithRevBlksAndConnectedToProdNodeTermAtBlock")
+
+   # 2th test case: Replay in speculative mode with reversible blks while connected to producing node
+   # Expectation: Node replays and launches successfully
+   #              and the head and lib should be no less than termAtBlockNum
+   def replayInSpecModeWithRevBlksAndConnectedToProdNodeTermAtBlock(nodeIdOfNodeToTest, nodeToTest):
+      try:
```
dimas1185: same comments apply as for `replayInIrrModeWithRevBlksAndConnectedToProdNodeTermAtBlock`

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+         assert head >= termAtBlockNum, "Head ({}) should be no less than termAtBlockNum ({})".format(head, termAtBlockNum)
+         assert forkDbHead >= head, "Fork db head ({}) should be larger or equal to the head ({})".format(forkDbHead, head)
+      finally:
+         stopProdNode()
+         Utils.Print("End replayInSpecModeWithRevBlksAndConnectedToProdNodeTermAtBlock")
+
+   # 3th test case: Replay in head mode with reversible blks while connected to producing node
+   # Expectation: Node replays and launches successfully
+   #              and the head and lib should be no less than termAtBlockNum
+   def replayInHeadModeWithRevBlksAndConnectedToProdNodeTermAtBlock(nodeIdOfNodeToTest, nodeToTest):
```
dimas1185: same comments apply as for:
`replayInIrrModeWithRevBlksAndConnectedToProdNodeTermAtBlock`

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+def waitForBlksProducedAndLibAdvanced():
+   requiredConfirmation = int(2 / 3 * numOfProducers) + 1
+   maxNumOfBlksReqToConfirmLib = (12 * requiredConfirmation - 1) * 2
+   # Give 6 seconds buffer time
+   bufferTime = 6
+   timeToWait = maxNumOfBlksReqToConfirmLib / 2 + bufferTime
+   time.sleep(timeToWait)
+
+# This function is special for "terminate-at-block" argument, if no this argument just use node.relaunch (see Node.py)
+def nodeRelaunch(node: Node, chainArg=None, newChain=False, skipGenesis=True, timeout=Utils.systemWaitTimeout, addSwapFlags=None, cachePopen=False, nodeosPath=None):
```
dimas1185: I'm not sure I understand why are you using custom method here. regular `relaunch` has an option to add new arguments or swap existing. If any issue with that, I suggest just to update framework with this functionality.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -0,0 +1,316 @@
...
+
+   # 1th test case: Replay in irreversible mode with reversible blks while connected to producing node
+   # Expectation: Node replays and launches successfully
+   #              and the head and lib should be no less than termAtBlockNum
+   def replayInIrrModeWithRevBlksAndConnectedToProdNodeTermAtBlock(nodeIdOfNodeToTest, nodeToTest):
+      try:
+         startProdNode()
+         # Kill node and replay in irreversible mode
+         nodeToTest.kill(signal.SIGTERM)
+         waitForBlksProducedAndLibAdvanced() # Wait
```
dimas1185: we have similar method in `Node.waitForLibToAdvance`

heifner (approved):
Approving the C++ code changes. The python changes still need to be made.

dimas1185 (changes requested):
please make changes to integration tests

tests/nodeos_read_terminate_at_block_test.py
hmmblockone: There does seem to be the occasional failure when using getInfo. When parsing the block log how do you distinguish between head and lib?  I can see the last block_num corresponds to the output of get info, and seems match the head.  Also, is there a way to get the fork_db_head_block_num from the block log?  Is there a good reference for this? Thanks

heifner: The blocklog only has lib in it. You can ask the controller for fork_db_head_block_num.

hmmblockone: Still learning, is the controller in this context a producer? I think there's a controller module as well? Thanks

heifner: Sorry for being terse. `cleos get info` returns this information:

```
{
  "server_version": "eb71ad75",
  "chain_id": "3dddd240fb31aea58e7b19efdbf4c33d865d9f0bdfb11111b32468fbde4e35dd",
  "head_block_num": 72336,
  "last_irreversible_block_num": 72335,
  "last_irreversible_block_id": "00011a8f84e867fe04bd5a44c7f0de40c37a0e3bb781167fea43cde407e2b0cc",
  "head_block_id": "00011a9068c7cf12dbc4464d6bb9b0dd3b4daf9320eaceeb9ba1c61665f661fe",
  "head_block_time": "2021-08-20T19:57:19.000",
  "head_block_producer": "eosio",
  "virtual_block_cpu_limit": 450000000,
  "virtual_block_net_limit": "10485760000",
  "block_cpu_limit": 450000,
  "block_net_limit": 10485760,
  "server_version_string": "v2.1.0",
  "fork_db_head_block_num": 72336,
  "fork_db_head_block_id": "00011a9068c7cf12dbc4464d6bb9b0dd3b4daf9320eaceeb9ba1c61665f661fe",
  "server_full_version_string": "v2.1.0-eb71ad7560e3293c2ed152cca459789f6abf9564",
  "last_irreversible_block_time": "2021-08-20T19:57:18.500"
}
```

The fork db data is in there, it is retrieved from nodeos state. If python `getInfo()` fails then it should just be retried.

dimas1185: I don't like race conditions here, executing `getInfo` until nodeos shutdown.. this test may become flaky.
what if we won't catch the very last block before nodeos shutdown?
`Cluster.getBlockLog` is using `eosio-blocklog` and later is using not just `blocks.log` but also `fork_db` that contains reversible blocks.
I'm not sure if it reads `fork_db_head_block_num` but if it is not maybe we can skip it in order to avoid flakiness.

hmmblockone: While the tests seem to pass fine, if I print the last element of block log vs. getInfo(), there's normally an additional block in the block log.

Assuming I drop fork_db_head_block number from the test to use getBlockLog(), I'm not clear how to retrieve the lib vs the head without using getInfo().

The log seems to be a list of json objects, and when comparing the log to the last successful getInfo() call, the log seems to include the head block (or maybe it just became the next lib and getInfo() missed it).

dimas1185: that looks unfortunate limitation of eosio-blocklog. In that case we could skip checking LIB at all. I don't see any benefit of checking it here.

PR #10267, opened 2021-04-19
Feature privacy

brianjohnson5972:
## Change Description
Added feature to prevent sending blocks and transactions to peers that are not part of the security group that has been defined. The security group is maintained in the block_header_state. add_security_group_participant and remove_security_group_participant intrinsics cause the current security group's state to be copied to the global_property_object and edited to create the proposed security group.  When the block for the last called intrinsic becomes irreversible, it promotes the proposed security group from the global_property_object to the proposed_block_header_state when the block is started and then to the block_header_state when the block is completed.
net_plugin caches the security_group state on accept block and then sets a flag on each connection to indicate if that connection's participant is participating in the network or not.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.gitignore
```diff
@@ -88,3 +88,5 @@ var/lib/node_*
 .DS_Store
 
 !*.swagger.*
+
+ZScaler*
```
brianjohnson5972, line 92: Do we know why we added this?

dimas1185: purely debug stuff. I was building local docker images and our Zscaler policy requires installing there certificates or all internet connections will be denied

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3657,45 @@ namespace eosio {
       }
    }
 
+   // called from any thread
```
heifner, line 3660: This is only called from the main application thread.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +858,108 @@ namespace eosio {
...
+         if (my_impl->ssl_enabled)
+            return participating_.load(std::memory_order_relaxed);
+         
+         return true;
+      }
+      inline void setup_participant() {
+         if(my_impl->ssl_enabled) {
+            account_name participant = participant_name_ ? *participant_name_ : account_name{};
+            
+            std::lock_guard<std::shared_mutex> connection_guard(my_impl->connections_mtx);
```
heifner, line 887: This can be a `shared_lock`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -254,8 +266,9 @@ namespace eosio {
       bool                                  use_socket_read_watermark = false;
       /** @} */
 
-      mutable std::shared_mutex             connections_mtx;
+      mutable std::shared_mutex             connections_mtx; // protects both connections and security_group
```
heifner, line 269: Why share this mutex with security_group? Seems like the code would be much simplier if security_group_manger had its own mutex to protect its internal data.

brianjohnson5972: that was my fault, I was thinking that it was going to be held while adding a connection on the TLS side, so I was trying to avoid the issues of double locking. Will change.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +858,108 @@ namespace eosio {
...
+      inline void set_participating(bool status) { 
+         if (my_impl->ssl_enabled)
+            participating_.store(status, std::memory_order_relaxed); 
+      }
+      /** returns true if connection is in the security group */
+      inline bool is_participating() const { 
+         if (my_impl->ssl_enabled)
+            return participating_.load(std::memory_order_relaxed);
+         
+         return true;
```
linhuang-blockone, line 881: Why returning true when ssl is not enabled?

brianjohnson5972: because everything i participating when ssl is not enabled.  That gets us the current behavior.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +858,108 @@ namespace eosio {
...
+            std::string organization{buf, (size_t)length};
+            if (is_string_valid_name(organization)){
+               participant_name_ = account_name{organization};
+               //participant name will be later used in start_session to determine if
+               //participant is participating or not
+            }
+            else {
+               peer_dlog(this, "received unauthorized participant: ${s}", ("s", organization));
+            }
+         }
```
linhuang-blockone, line 947: This section of code appears not be related to verify cert. Can it be a separate function like get_organization?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -789,6 +858,108 @@ namespace eosio {
...
+      /** returns true if connection is in the security group */
+      inline bool is_participating() const { 
+         if (my_impl->ssl_enabled)
+            return participating_.load(std::memory_order_relaxed);
+         
+         return true;
+      }
+      inline void setup_participant() {
+         if(my_impl->ssl_enabled) {
+            account_name participant = participant_name_ ? *participant_name_ : account_name{};
```
linhuang-blockone, line 885: Should just return when participant_name_ is null?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3428,10 +3657,45 @@ namespace eosio {
...
+
+      // update connections
+      //
+      fc_dlog( logger, "SecurityGroup changed to version: ${v}", ("v", security_group.current_version()) );
+      for(auto& connection : connections) {
+         const auto& participant = connection->participant_name();
+         if(participant && security_group.is_in_security_group(participant.value())) {
+            if(!connection->is_participating()) {
+               connection->set_participating(true);
+               connection->send_handshake();
```
linhuang-blockone, line 3685: Can send_handshake throw? If so, it is better to do set_participating(true) after send_handshake().

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3563,6 +3827,34 @@ namespace eosio {
...
    }
 
+   void net_plugin_impl::init_ssl_context(const std::string& cert, const std::string& pkey, const std::string& ca){
+      ssl_context.reset( new ssl::context(ssl::context::sslv23) );
+
+      //TLS-only connection, no SSL
+      ssl_context->set_options(ssl::context::default_workarounds | 
+                           ssl::context::no_sslv2 |
+                           ssl::context::no_sslv3 );
+      ssl_context->set_password_callback([](auto,auto){ return "test"; });
```
linhuang-blockone, line 3837: Is "test" just for test?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3563,6 +3827,34 @@ namespace eosio {
...
+
+      if (!ca.empty()){
+         dlog("using verify file: ${p}", ("p", ca));
+         ssl_context->load_verify_file(ca, ec);
+         EOS_ASSERT(!ec, ssl_configuration_error, "load_verify_file: ${e}", ("e", ec.message()));
+
+         //this ensures peer has trusted certificate. no certificate-less connections
+         ssl_context->set_verify_mode(ssl::context::verify_peer | ssl::context::verify_fail_if_no_peer_cert);
+      }
+      dlog("using private key file: ${p}", ("p", pkey));
```
linhuang-blockone, line 3849: Do we want to print out private key? Will this be taken out in production?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3762,6 +4056,30 @@ namespace eosio {
             my->chain_plug->enable_accept_transactions();
          }
 
+         //if we have certificate option that TLS must be enabled
+         if ( options.count("p2p-tls-own-certificate-file") ) {
+            auto certificate = options["p2p-tls-own-certificate-file"].as<bfs::path>();
+            auto pkey        = options["p2p-tls-private-key-file"].as<bfs::path>();
+            auto ca_cert     = options["p2p-tls-security-group-ca-file"].as<bfs::path>();
```
linhuang-blockone, line 4063: Should make sure key file and CA are configured if certificate is configured and print out the error, as user can make mistakes and forgets to configure all of them.

plugins/net_plugin/security_group_manager.cpp
```diff
@@ -0,0 +1,11 @@
+#include <eosio/net_plugin/security_group_manager.hpp>
+
+namespace eosio {
+   bool security_group_manager::update_cache(const uint32_t version, const participant_list_t& participant_list) {
+      if(version == version_)
+         return false;
+      version_ = version;
```
linhuang-blockone, line 7: Do we allow version going backward?

plugins/net_plugin/test/security_group_manager_tests.cpp
```diff
@@ -0,0 +1,116 @@
...
+      for(auto participant : removed) {
+         BOOST_REQUIRE(!manager.is_in_security_group(participant));
+      }
+
+      auto added = create_list({7, 8, 9});
+      for (auto participant : added) {
+         BOOST_REQUIRE(manager.is_in_security_group(participant));
+      }
+}
+BOOST_AUTO_TEST_SUITE_END()
```
linhuang-blockone, line 116: Can we have some unit tests for SSL, possibly in a separate file?

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: I pulled it into a method called extract_participant_name, but it still needs to be called in verify_cert. Is the real issue that we should call this method process_cert and then it is able to both verify the certificate and process any data it needs from it?

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: no, because you could setup a TLS connected network and never turn on the security_group and we want to keep this connection participating in that case. That will be handled in is_in_security_group(participant) since it checks to see if the security_group is empty, in which case ALL participants are in the security group.

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: It doesn't throw, and since it ends up processing that send on another thread, you want participating already set true, otherwise you may end up not completing the chain of events that sending a handshake was meant to cause.

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: @dimas1185 ^^^

plugins/net_plugin/include/eosio/net_plugin/security_group_manager.hpp
```diff
@@ -0,0 +1,30 @@
...
+      /** @brief Update the security group participants
+       *
+       * @param version             The version number for this update
+       * @param participant_list    The list of accounts for the security group
+       * @return True if an update was performed.
+       */
+      bool update_cache(const uint32_t version, const participant_list_t& participant_list);
+      /** @brief Determine if a participant is in the security group */
+      bool is_in_security_group(chain::account_name participant) const {
+         return cache_.empty() || (cache_.find(participant) != cache_.end());
```
linhuang-blockone, line 25: Why returns true when cache is empty?

plugins/net_plugin/net_plugin.cpp
linhuang-blockone: process_cert describes what the function does better.

plugins/net_plugin/test/security_group_manager_tests.cpp
brianjohnson5972: Not in unit tests, but we have privacy_tls_test.py that does integration tests around ensuring that TLS verification is occurring.

plugins/net_plugin/include/eosio/net_plugin/security_group_manager.hpp
brianjohnson5972: because an empty security group isn't saying that "no one" is in the security group, it is saying "everyone" is in the security group, so if the cache is empty, then is_in_security_group should return true.

plugins/net_plugin/security_group_manager.cpp
heifner: The assert verbiage could use a bit of work.

plugins/net_plugin/net_plugin.cpp
dimas1185: it is not a bug, it is feature.
we allow SSL without security groups and hence without CA.
However in that case on we may need that upon security group activation we have CA configured.
@brianjohnson5972

ndcgundlach (changes requested):

libraries/chain/include/eosio/chain/global_property_object.hpp
```diff
@@ -118,25 +129,77 @@ namespace eosio { namespace chain {
...
    };
 
+   inline snapshot_global_property_object::extension_t get_gpo_extension(const global_property_object& gpo) {
+      return snapshot_global_property_object::extension_v0{
+          gpo.proposed_security_group_block_num,
+          {gpo.proposed_security_group_participants.begin(), gpo.proposed_security_group_participants.end()}};
+   }
+
+   inline void set_gpo_extension(global_property_object&                        gpo,
+                                 snapshot_global_property_object::extension_t&& extension) {
```
heifner, line 166: Why a rvalue reference instead of just const& ? It is only copied from, so seems like const& would be better.

ndcgundlach (approved):

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: I fixed error check to require both certificate file and private key file be set to the same value and also require certificate file be set if security group cca file is set

plugins/net_plugin/net_plugin.cpp
brianjohnson5972: removed

plugins/net_plugin/security_group_manager.cpp
brianjohnson5972: I think I fixed it

libraries/chain/include/eosio/chain/global_property_object.hpp
brianjohnson5972: that was left over from the original change that didn't use shared_set in global_property_object and that was missed, cleaned up

.gitignore
brianjohnson5972: removed

linhuang-blockone (approved):

PR #10276, opened 2021-04-20
Ensure container.size() > 0 before initiating RocksDB writes/deletes

nksanthosh:
## Change Description

In the prior implementation RocksDB writes/deletes were always called even if the input size was zero. Added a check to only send those calls to rocksDB when size()>0. Improved performance of replays marginally. These zero sized calls to rocksDB numbered many thousands in 1M blocks replays of real world blockslogs.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain_kv/include/b1/session/session.hpp
+++ b/libraries/chain_kv/include/b1/session/session.hpp
@@ -502,9 +502,14 @@ void session<Parent>::commit() {
             updates.emplace(p.first, p.second.value);
          }
       }
-
+      if (deletes.size() > 0) {
       ds.erase(deletes);
+      }
+      
+      if (updates.size() > 0) {
       ds.write(updates);
+      }
+      
       clear();
    };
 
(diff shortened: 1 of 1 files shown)


libraries/chain_kv/include/b1/session/session.hpp
```diff
@@ -502,9 +502,14 @@ void session<Parent>::commit() {
...
       }
-
+      if (deletes.size() > 0) {
       ds.erase(deletes);
+      }
+      
+      if (updates.size() > 0) {
       ds.write(updates);
+      }
+
```
heifner, line 512: Fix indentation

heifner (changes requested):

linhuang-blockone (approved):

brianjohnson5972 (approved):

PR #10282, opened 2021-04-21
Created SecurityGroup.py script for providing helper methods

brianjohnson5972:
## Change Description
Pulled code out of privacy_simple_network.py for adding and removing participants to the security_group and verifying the network after changes.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+        intersect = list(set([x.getParticipant() for x in self.participants]) & set([x.getParticipant() for x in self.nonParticipants]))
+        if len(intersect) != 0:
+            Utils.exitError("Values: {}, are in the nonParticipants list: {}, and the participants list: {}".format(", ".join(intersect), SecurityGroup.createAction(self.nonParticipants), SecurityGroup.createAction(self.participants)))
+
+    def __addParticipants(self, nodes):
+        if Utils.Debug: Utils.Print("Moving the following: {}, from the nonParticipants list: {}, to the participants list: {}".format(SecurityGroup.createAction(nodes), SecurityGroup.createAction(self.nonParticipants), SecurityGroup.createAction(self.participants)))
+        for node in nodes:
+            if node not in self.nonParticipants:
+                for nonParticipant in self.nonParticipants:
+                    match = "" if node == nonParticipant else " not"
```
dimas1185, line 41: what is this loop for?
if `if node not in self.nonParticipants` was triggered how can you have a match here?

tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+                    Utils.Print("node: '{}' does{} match '{}'".format(node, match, nonParticipant))
+            assert node in self.nonParticipants, "Cannot remove {} from nonParticipants list: {}".format(node, SecurityGroup.createAction(self.nonParticipants))
+            self.nonParticipants.remove(node)
+
+        self.participants.extend(nodes)
+        if Utils.Debug: Utils.Print("nonParticipants list: {}, to the participants list: {}".format(SecurityGroup.createAction(self.nonParticipants), SecurityGroup.createAction(self.participants)))
+        self.__consistentStorage()
```
dimas1185, line 48: this method seem excessive check to me. `__addParticipants` method does remove from non participants and append to participants. So how mismatch can happen?

tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+    @staticmethod
+    def createAction(nodes):
+        return None if len(nodes) == 0 else \
+            "[[{}]]".format(','.join(['"{}"'.format(node.getParticipant()) for node in nodes]))
+
+    def editSecurityGroup(self, addNodes=[], removeNodes=[]):
+
+        addAction = SecurityGroup.createAction(addNodes)
+        # doing deep copy in case the passed in list IS participants or nonParticipants lists, which will be adjusted
+        self.__addParticipants(copy.deepcopy(addNodes))
```
dimas1185, line 68: I don't understand why do we use `deepcopy` here? So it will copy `Node` instance. Why do we need this?
passed list is not stored anywhere inside `__addParticipants`, its elements are added to class members.

tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+            Utils.Print("adding {} to the security group".format(addAction))
+            trans = self.defaultNode.pushMessage(self.contractAccount.name, "add", addAction, "--permission eosio@active")
+            Utils.Print("add trans: {}".format(json.dumps(trans, indent=4, sort_keys=True)))
+
+        if removeAction:
+            Utils.Print("removing {} from the security group".format(removeAction))
+            trans = self.defaultNode.pushMessage(self.contractAccount.name, "remove", removeAction, "--permission eosio@active")
+            Utils.Print("remove trans: {}".format(json.dumps(trans, indent=4, sort_keys=True)))
+
+        self.publishProcessNum += 1
```
dimas1185, line 82: do we really need this functionality? We deal with small amount of participants and can simply publish 0 every time

tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+        producerHead = self.defaultNode.getBlockNum()
+
+        for nonParticipant in self.nonParticipants:
+            nonParticipantPostLIB = nonParticipant.getBlockNum(blockType=BlockType.lib)
+            assert nonParticipantPostLIB < publishBlock, "Participants not in security group should not have advanced LIB to {}, but it has advanced to {}".format(publishBlock, nonParticipantPostLIB)
+            nonParticipantHead = nonParticipant.getBlockNum()
+            assert nonParticipantHead < producerHead, "Participants (that are not producers themselves) should not advance head to {}, but it has advanced to {}".format(producerHead, nonParticipantHead)
+
+    def verifySecurityGroup(self, publishTrans = None):
+        if publishTrans is None:
```
dimas1185, line 113: you have same check in `verifyNonParticipants` and `verifyParticipantsTransactionFinalized`.
I think we should keep only one. Maybe we can remove checks from called methods and leave this one?

tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+            publishTrans = self.publishTrans
+        publishTransId = Node.getTransId(publishTrans)
+        self.verifyParticipantsTransactionFinalized(publishTransId)
+        self.verifyNonParticipants(publishTransId)
+
+    def addToSecurityGroup(self):
```
dimas1185, line 119: this method looks confusing. its meaning looks like `moveHeadToSecurityGroup`. I suggest to rename it or add new member like this:
`moveToSecurityGroup(self, index=0)`

tests/SecurityGroup.py
```diff
@@ -0,0 +1,127 @@
...
+        trans = self.editSecurityGroup([self.nonParticipants[0]])
+        Utils.Print("Take a non-participant and make a participant. Now there are {} participants and {} non-participants".format(len(self.participants), len(self.nonParticipants)))
+        return trans
+
+    def remFromSecurityGroup(self):
```
dimas1185, line 124: same as `addToSecurityGroup`

tests/privacy_simple_network.py
```diff
@@ -98,127 +99,35 @@
...
-            participantNum = cluster.getParticipantNum(apiNode)
-            toAdd.append(participantNum)
-        participants.extend(nonParticipants)
-        nonParticipants = []
-
-        Utils.Print("Add all api nodes to security group")
-        publishTrans = security_group(addNodeNums=toAdd)
-        verifySecurityGroup(publishTrans)
+        Utils.Print("Add all api nodes to security group at the same time")
+        securityGroup.editSecurityGroup(addNodes=securityGroup.nonParticipants)
```
dimas1185, line 130: now I see why you used `copy` inside `editSecurityGroup`. First of all I think proper behavior is to use shallow copy `copy.copy` and not deep copy. Second, maybe we can add helper method like `moveAllToSecurityGroup` to avoid using internal structures as parameters?

tests/SecurityGroup.py
brianjohnson5972: Sorry, debug code, removed

tests/SecurityGroup.py
brianjohnson5972: leftover debug code, removed

tests/SecurityGroup.py
brianjohnson5972: so that calls like securityGroup.editSecurityGroup(securityGroup.nonParticipating) don't change addNodes when it calls remove self.nonParticipating, which otherwise would be the exact same container.

tests/SecurityGroup.py
brianjohnson5972: except then we would have identical transactions because there is nothing distinguishing one publish action from another. If we don't do this, we would need to add a nonce or else make sure we never send 2 publish actions in the same block

tests/privacy_simple_network.py
brianjohnson5972: changed all deepcopy to copy. Added a local function to only copy if it is one of our containers

tests/SecurityGroup.py
dimas1185: that is known limitation to transactions. I think we just should never send two publish actions in one block. Was it ever issue to you or it was just in case?

tests/SecurityGroup.py
```diff
@@ -0,0 +1,155 @@
...
+
+        featureDict = node.getSupportedProtocolFeatureDict()
+        Utils.Print("feature dict: {}".format(json.dumps(featureDict, indent=4, sort_keys=True)))
+
+        Utils.Print("{} Feature activated".format(feature))
+
+    # publish the eosio.secgrp contract
+    @staticmethod
+    def publishContract(node, account):
+            Utils.Print("Publish contract")
```
dimas1185, line 46: `"Publish eosio.secgrp contract"` ?

jgiszczak (approved):

PR #10285, opened 2021-04-22
Update security_group_tests.cpp

shahabi8:
## Change Description
Added test for security groups with 2 chains to verify both the producing chain and applying chain produce the same security group state.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/security_group_tests.cpp
+++ b/unittests/security_group_tests.cpp
@@ -420,6 +420,90 @@ BOOST_AUTO_TEST_CASE(test_participants_change_modified) {
    BOOST_CHECK(chain.control->in_active_security_group(participants_t{"bob"_n}));
 }
 
+BOOST_AUTO_TEST_CASE(test_participants_change_2_chains) { 
+   // chain2 trails chain a block behind since it reads the records of last finalized block from chain
+   eosio::testing::tester chain;
+   eosio::testing::tester chain2;
+   using namespace eosio::chain::literals;
+   
+   chain.create_accounts( {"alice"_n,"bob"_n,"charlie"_n} );
+
+   chain2.push_block(chain.produce_block());
+   
+   {
+      const auto& cur_security_group = chain.control->active_security_group();
+      BOOST_REQUIRE_EQUAL(cur_security_group.version, 0);
+      BOOST_REQUIRE_EQUAL(cur_security_group.participants.size(), 0);
+   }
+
+   chain.create_accounts({ "addmember"_n, "rmmember"_n });
+
+   chain2.push_block(chain.produce_block());
+
+   chain.set_code( "addmember"_n, add_security_group_participants_wast );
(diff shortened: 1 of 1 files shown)


unittests/security_group_tests.cpp
```diff
@@ -420,6 +420,90 @@ BOOST_AUTO_TEST_CASE(test_participants_change_modified) {
...
+   BOOST_TEST_REQUIRE(chain.push_action_no_produce(  eosio::chain::action({}, "addmember"_n, {}, participants_payload({"bob"_n})), "addmember"_n.to_uint64_t()));
+
+   BOOST_CHECK_EQUAL(chain.control->proposed_security_group_participants().size(), 2);
+   BOOST_CHECK_EQUAL(chain.control->active_security_group().participants.size(), 0);
+   BOOST_CHECK_EQUAL(chain2.control->proposed_security_group_participants().size(), 0);
+   BOOST_CHECK_EQUAL(chain2.control->active_security_group().participants.size(), 0);
+
+   chain2.push_block(chain.produce_block());
+
+   BOOST_CHECK_EQUAL(chain.control->proposed_security_group_participants().size() , 0);
```
brianjohnson5972, line 473: add a comment like:
// because produce_block calls start_block, and push_block does not, chain2 will trail behind chain

brianjohnson5972 (changes requested):

unittests/security_group_tests.cpp
shahabi8: Just added the comment as you suggested.

unittests/security_group_tests.cpp
```diff
@@ -421,7 +421,9 @@ BOOST_AUTO_TEST_CASE(test_participants_change_modified) {
 }
 
 BOOST_AUTO_TEST_CASE(test_participants_change_2_chains) { 
-   // chain2 trails chain a block behind since it reads the records of last finalized block from chain
+   /* Note:
+   * because produce_block calls start_block, and push_block does not, chain2 will trail behind chain
```
brianjohnson5972, line 425: Sorry, I realized that this is not clear, it need to be "because produce_block calls start_block after finalizing the produced block, and push_block does not, chain2 will trail behind chain". They both call start_block, but push_block is calling it at the start, while produce block is calling it after it finalizes the block it is producing.

brianjohnson5972 (approved):

PR #10288, opened 2021-04-23
Privacy Test Case #1

brianjohnson5972:
## Change Description
Created script to automate test case #1.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/SecurityGroup.py
+++ b/tests/SecurityGroup.py
@@ -15,7 +15,8 @@ class SecurityGroup(object):
         self.participants = []
         self.contractAccount = contractAccount
         assert len(nonParticipants) > 0
-        self.nonParticipants = copy.copy(nonParticipants)
+        # copy over all the running processes
+        self.nonParticipants = [x for x in nonParticipants if x.pid]
         if Utils.Debug: Utils.Print("Creating SecurityGroup with the following nonParticipants: []".format(SecurityGroup.createAction(self.nonParticipants)))
         self.defaultNode = defaultNode if defaultNode else nonParticipants[0]
         self.publishProcessNum = minAddRemEntriesToPublish
--- a/tests/SecurityGroup.py
+++ b/tests/SecurityGroup.py
@@ -151,7 +151,7 @@ class SecurityGroup(object):
             assert nonParticipantHead < producerHead, "Participants (that are not producers themselves) should not advance head to {}, but it has advanced to {}".format(producerHead, nonParticipantHead)
 
     def getLatestPublishTransId(self):
-        publishTransId = Node.getTransId(self.publishTrans)
+        return Node.getTransId(self.publishTrans)
 
     # verify that the participants' and nonParticipants' nodes are consistent based on the publish transaction
     def verifySecurityGroup(self, publishTrans = None):
--- a/tests/SecurityGroup.py
+++ b/tests/SecurityGroup.py
@@ -16,9 +16,16 @@ class SecurityGroup(object):
         self.contractAccount = contractAccount
         assert len(nonParticipants) > 0
         # copy over all the running processes
-        self.nonParticipants = [x for x in nonParticipants if x.pid]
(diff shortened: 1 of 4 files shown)


tests/SecurityGroup.py
```diff
@@ -15,7 +15,8 @@ def __init__(self, nonParticipants, contractAccount, defaultNode=None, minAddRem
         self.participants = []
         self.contractAccount = contractAccount
         assert len(nonParticipants) > 0
-        self.nonParticipants = copy.copy(nonParticipants)
+        # copy over all the running processes
+        self.nonParticipants = [x for x in nonParticipants if x.pid]
```
dimas1185, line 19: why do you need this? this will make more difficult to deal with `unstartedNodes`

tests/privacy_startup_network.py
```diff
@@ -18,8 +19,11 @@
 ###############################################################
 # privacy_startup_network
 #
-# General test for Privacy to verify TLS connections, and slowly adding participants to the security group and verifying
-# how blocks and transactions are sent/not sent.
+# Script implements Privacy Test Case #1. It pairs up producers with p2p connections with relay nodes
+# and the relay nodes connected to at least 2 or more API nodes.  The producers and relay nodes are
```
dimas1185, line 23: you use `non-producers`, `api nodes`, `relay nodes`. what is the difference?

tests/SecurityGroup.py
brianjohnson5972: This is needed so that the killed bios node (killed because it has a mesh connection to everything) to keep it from failing the checks for the nonParticipants.  I guess we could put checks in the code to check for node.pid before:
moving it to the participants list
validating not in sync
etc.

That should work for eventually starting an unstartedNodes.
Do you agree?

dimas1185: Yes, agree. I have test with `unstartedNodes` where I assign those at startup like this:
`SecurityGroup(self.getAllNodes() + self.unstartedNodes, self.eosioAccount, defaultNode=self.nodes[0])`
so that should work fine in my case

tests/privacy_startup_network.py
brianjohnson5972: Let me know if the note I added was helpful with this.

jgiszczak (changes requested):

dimas1185 (approved):

tests/SecurityGroup.py
```diff
@@ -130,11 +143,16 @@ def verifyNonParticipants(self, transId):
...
         # never have been forwarded to a nonParticipant
         for nonParticipant in self.nonParticipants:
+            if nonParticipant.pid is None:
+                continue
             nonParticipantPostLIB = nonParticipant.getBlockNum(blockType=BlockType.lib)
             assert nonParticipantPostLIB < publishBlock, "Participants not in security group should not have advanced LIB to {}, but it has advanced to {}".format(publishBlock, nonParticipantPostLIB)
             nonParticipantHead = nonParticipant.getBlockNum()
             assert nonParticipantHead < producerHead, "Participants (that are not producers themselves) should not advance head to {}, but it has advanced to {}".format(producerHead, nonParticipantHead)
 
+    def getLatestPublishTransId(self):
```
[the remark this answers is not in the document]
brianjohnson5972: whoops, fixed that. It was so that if the caller didn't want to call verifySecurityGroup, and just was calling one of the functions that calls, it can get the default value.

jgiszczak (approved):

PR #10296, opened 2021-04-24
scenario 3 test added to feature privacy tests

dimas1185:
## Change Description

scenario 3 implementation from test plan: https://blockone.atlassian.net/wiki/spaces/EOSIOProdEng/pages/1434419366/Private+Chain+Access+Test+Plan
some corrections been made:
steps m,n,s revised as they caused errors

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [x] Existing Tests

- [x] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -6,6 +6,7 @@ import os
 import re
 import json
 import signal
+import requests
 
 from datetime import datetime
 from datetime import timedelta
@@ -564,8 +565,8 @@ class Node(object):
         assert ret is not None or errorContext is None, Utils.errorExit("%s." % (errorContext))
         return ret
 
-    def waitForIrreversibleBlock(self, blockNum, timeout=WaitSpec.default(), blockType=BlockType.head):
-        return self.waitForBlock(blockNum, timeout=timeout, blockType=blockType)
+    def waitForIrreversibleBlock(self, blockNum, timeout=WaitSpec.default()):
+        return self.waitForBlock(blockNum, timeout=timeout, blockType=BlockType.lib)
 
     # Trasfer funds. Returns "transfer" json return object
     def transferFunds(self, source, destination, amountStr, memo="memo", force=False, waitForTransBlock=False, exitOnError=True, reportStatus=True, sign=False, dontSend=False, expiration=None, skipSign=False):
@@ -1121,7 +1122,7 @@ class Node(object):
     def kill(self, killSignal):
         if Utils.Debug: Utils.Print("Killing node: %s" % (self.cmd))
         assert (self.pid is not None)
-        Utils.Print("Killing node pid: {}", self.pid)
+        Utils.Print("Killing node pid: {}".format(self.pid))
         try:
             if self.popenProc is not None:
                Utils.Print("self.popenProc is not None")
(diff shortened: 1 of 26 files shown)


tests/Node.py
```diff
@@ -6,6 +6,7 @@
 import re
 import json
 import signal
+import requests
```
brianjohnson5972, line 9: It looks like this isn't available on most of the buildkite pipelines.

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,257 @@
...
+    # non-producing nodes do not receive the block that moved lib to or past the recorded block
+    Print("Node1 head = {} Prod1 head = {}".format(apiNode1_head, prod1_head))
+    assert apiNode1_head <= prod1_head
+
+    trans = securityGroup.editSecurityGroup(addNodes=[apiNode1])
+    # record block number (rbn1) transaction was added in
+    trans_block_num = trans["processed"]["block_num"]
+    Print("security group update [[\"node3\"]] published on block ".format(trans_block_num))
+
+    prod1.waitForIrreversibleBlock(trans_block_num)
```
brianjohnson5972, line 101: Is there a reason you don't use validateSecurityGroup method? If you are not going to use that, you should use waitForTransactionFinalization (it is named something close to that at least) instead of using the transaction block number, since that block number is a guess as to when the transaction is processed.  Since you have 2 producers, when prod1 is the producer, that will be the block number, but if it is prod2's turn to produce, then it may be in the next block or even later.

brianjohnson5972 (changes requested):

tests/Node.py
dimas1185: I'll fix PR after 2 PM EST. We need `pip install requests` for this. Do you know right place for this?

brianjohnson5972: I'm guessing in the .cicd pipeline files, but not sure where

dimas1185: @brianjohnson5972 I'm already resolved this for Ubuntu. working for other platforms

dimas1185: @brianjohnson5972 I'm having issues with `libpqxx` on `macos` . it seems like when rebuilding a base image it fails to install this library from particular commit. I tried to update the commit.
```
curl -LO  https://raw.githubusercontent.com/Homebrew/homebrew-core/106b4b8a421dda33c79a4018c3c3816234076331/Formula/libpqxx.rb
brew install -f ./libpqxx.rb
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  1442  100  1442    0     0   2089      0 --:--:-- --:--:-- --:--:--  2089
Warning: Calling `cellar` in a bottle block is deprecated! Use `brew style --fix` on the formula to update the style or use `sha256` with a `cellar:` argument instead.
Warning: Calling `sha256 "digest" => :tag` in a bottle block is deprecated! Use `brew style --fix` on the formula to update the style or use `sha256 tag: "digest"` instead.
Warning: Calling `sha256 "digest" => :tag` in a bottle block is deprecated! Use `brew style --fix` on the formula to update the style or use `sha256 tag: "digest"` instead.
Warning: Calling `sha256 "digest" => :tag` in a bottle block is deprecated! Use `brew style --fix` on the formula to update the style or use `sha256 tag: "digest"` instead.
Warning: Calling `sha256 "digest" => :tag` in a bottle block is deprecated! Use `brew style --fix` on the formula to update the style or use `sha256 tag: "digest"` instead.
Error: Failed to load cask: ./libpqxx.rb
Cask 'libpqxx' is unreadable: wrong constant name #<Class:0x00007fab95a8b540>
Warning: Treating ./libpqxx.rb as a formula.
==> Downloading https://ghcr.io/v2/homebrew/core/libpqxx/manifests/7.2.1
##O#- #
curl: (22) The requested URL returned error: 404 Not Found
Error: Failed to download resource "libpqxx_bottle_manifest"
Download failed: https://ghcr.io/v2/homebrew/core/libpqxx/manifests/7.2.1
```

PR #10297, opened 2021-04-26
EPE-412: use better error message when a port is taken

plroblox:
## Change Description

As requested by EPE-412:Use a better error message for nodeos startup, when a port for p2p-listen-endpoint is taken by another app 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

Use a port scanner to see if a port 9000+ like 9010 is taken by another app like zscaler in the test machine, then start nodeos with the following to see the new error message which clearly tells the cause of failure:
 nodeos -e -p eosio  --p2p-listen-endpoint 127.0.0.1:9010  --plugin eosio::producer_plugin 

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3820,7 +3820,7 @@ namespace eosio {
            my->acceptor->bind(listen_endpoint);
            my->acceptor->listen();
          } catch (const std::exception& e) {
-           elog( "net_plugin::plugin_startup failed to bind to port ${port}", ("port", listen_endpoint.port()) );
+           elog( "net_plugin::plugin_startup failed to bind to port ${port}. Please check if the port is in use by another application. Either choose a different port or stop that application", ("port", listen_endpoint.port()) );
            throw e;
          }
          fc_ilog( logger, "starting listener, max clients is ${mc}",("mc",my->max_client_count) );
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3820,9 +3820,8 @@ namespace eosio {
            my->acceptor->bind(listen_endpoint);
            my->acceptor->listen();
          } catch (const std::exception& e) {
-           fc_elog(logger, "net_plugin::plugin_startup failed to bind to port ${port}." 
-		 "Please check if the port is likely in use by another application." 
-		 "Either choose a different port or stop that application", ("port", listen_endpoint.port()) );
+           fc_elog(logger, "net_plugin::plugin_startup failed to bind to port ${port}. " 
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3820,7 +3820,7 @@ namespace eosio {
            my->acceptor->bind(listen_endpoint);
            my->acceptor->listen();
          } catch (const std::exception& e) {
-           elog( "net_plugin::plugin_startup failed to bind to port ${port}", ("port", listen_endpoint.port()) );
+           elog( "net_plugin::plugin_startup failed to bind to port ${port}. Please check if the port is in use by another application. Either choose a different port or stop that application", ("port", listen_endpoint.port()) );
```
heifner, line 3823: This should be `fc_elog` instead of `elog`.
Add line wrap so code is easier to read.
I would recommend something more like: ".. . The port is likely in use by another application."

heifner (changes requested):

nickjjzhao:
It's more appropriate to set Change Type as `other` instead of `stability bug fix`.

plugins/net_plugin/net_plugin.cpp
plroblox: will do, thanks Kevin!

plroblox: code change is in, thx

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3820,7 +3820,9 @@ namespace eosio {
            my->acceptor->bind(listen_endpoint);
            my->acceptor->listen();
          } catch (const std::exception& e) {
-           elog( "net_plugin::plugin_startup failed to bind to port ${port}", ("port", listen_endpoint.port()) );
+           fc_elog(logger, "net_plugin::plugin_startup failed to bind to port ${port}." 
+		 "Please check if the port is likely in use by another application." 
+		 "Either choose a different port or stop that application", ("port", listen_endpoint.port()) );
```
nickjjzhao, line 3825: I think you could use the exact error message recommended by Kevin, i.e., ".. . The port is likely in use by another application."

plroblox: done, see another commit

heifner (approved):

plroblox:
all buildkite tests have passed, merge my fix for EPE-412 into develop branch

PR #10301, opened 2021-04-27
huangminghuang/update-abios

huangminghuang:
## Change Description

Update abieos to fix cmake 3.8 issue

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode:
Instead of `INCLUDES DESTINATION include` shouldn't this be something like `INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}` like the others?

huangminghuang:
> Instead of `INCLUDES DESTINATION include` shouldn't this be something like `INCLUDES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}` like the others?
CMAKE_INSTALL_INCLUDEDIR is only defined after the inclusion of GNUInstallDirs, which I don't think it's necessary. 

From cmake documetation:
> If a relative path is given it is interpreted relative to the value of the CMAKE_INSTALL_PREFIX variable.

brianjohnson5972 (approved):

spoonincode:
every other install(TARGETS has it. Even other out of tree projects from us like eos-vm
https://github.com/EOSIO/eos-vm/blob/59879ce21ea61d796f7cbbbed3c03d158a72ab77/CMakeLists.txt#L129-L135
Is there some reason abieos should not be consistent?

PR #10303, opened 2021-04-27
Resource Payer - Producer Plugin changes

victorj8:
The logic on the producer is changed to check for the existence of a resource payer transaction extension. 

if it exists, then it will use the specified payer to change the authorizer of the transaction ( by default the authorizer will be the first authorizer of the transaction). 

This will ensure that the payer defined in the resource_payer extension is subjectively billed.

Since the signature was already validated during signature recovery, no check will be needed to verify that the payer signed.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/transaction.hpp
+++ b/libraries/chain/include/eosio/chain/transaction.hpp
@@ -127,8 +127,8 @@ namespace eosio { namespace chain {
          return account_name();
       }
 
-      account_name authorizer()const {
-         auto authorizer = first_authorizer();
+      account_name resource_payer()const {
+         auto resource_payer = first_authorizer();
 
          // Check for the existence of a extension of the type resource_payer, for the possibility
          // of changing the default account to authorize
@@ -137,12 +137,12 @@ namespace eosio { namespace chain {
          if(!extensions.empty()) {
             auto itr = extensions.lower_bound(resource_payer::extension_id());
             if(itr != extensions.end()) {
-               const auto& resource_payer_info = std::get<resource_payer>(itr->second);
-               authorizer = resource_payer_info.payer;
+               const auto& resource_payer_info = std::get<eosio::chain::resource_payer>(itr->second);
+               resource_payer = resource_payer_info.payer;
             }
          }
 
-         return authorizer;
+         return resource_payer;
       }
 
       flat_multimap<uint16_t, transaction_extension> validate_and_extract_extensions()const;
--- a/libraries/chain/include/eosio/chain/transaction.hpp
+++ b/libraries/chain/include/eosio/chain/transaction.hpp
@@ -127,6 +127,24 @@ namespace eosio { namespace chain {
          return account_name();
       }
 
+      account_name authorizer()const {
+         auto authorizer = first_authorizer();
+
+         // Check for the existence of a extension of the type resource_payer, for the possibility
+         // of changing the default account to authorize
+         auto extensions = validate_and_extract_extensions();
+
+         if(!extensions.empty()) {
+            auto itr = extensions.lower_bound(resource_payer::extension_id());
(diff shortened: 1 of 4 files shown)


libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -127,6 +127,24 @@ namespace eosio { namespace chain {
          return account_name();
       }
 
+      account_name authorizer()const {
```
ndcgundlach, line 130: I would change this to something like resource_payer()

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
```diff
@@ -119,50 +119,50 @@ class subjective_billing {
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
 
    /// @param in_pending_block pass true if pt's bill time is accounted for in the pending block
-   void subjective_bill( const transaction_id_type& id, const fc::time_point& expire, const account_name& first_auth,
+   void subjective_bill( const transaction_id_type& id, const fc::time_point& expire, const account_name& authorizer,
```
ndcgundlach, line 122: here again change authorizer to resource_payer and in other places in the code

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
```diff
@@ -119,50 +119,50 @@ class subjective_billing {
...
             if( in_pending_block ) {
-               _block_subjective_bill_cache[first_auth] += bill;
+               _block_subjective_bill_cache[authorizer] += bill;
             }
          }
       }
    }
 
-   void subjective_bill_failure( const account_name& first_auth, const fc::microseconds& elapsed, const fc::time_point& now )
+   void subjective_bill_failure( const account_name& authorizer, const fc::microseconds& elapsed, const fc::time_point& now )
```
ndcgundlach, line 141: change authorizer to resource_payer

libraries/chain/include/eosio/chain/transaction.hpp
victorj8: Done

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
victorj8: Done

plugins/producer_plugin/include/eosio/producer_plugin/subjective_billing.hpp
victorj8: Done

libraries/chain/include/eosio/chain/transaction.hpp
```diff
@@ -127,6 +127,24 @@ namespace eosio { namespace chain {
          return account_name();
       }
 
+      account_name resource_payer()const {
+         auto resource_payer = first_authorizer();
```
heifner, line 131: It would be more efficient to not take the time to retrieve the first_authorizer if resource_payer is in the trx extension.

victorj8: Thanks for the suggestion, i changed that part

ndcgundlach (approved):

PR #10309, opened 2021-04-28
Rodeos fixes for rocksdb high write amplification and large number of sst files created 

nksanthosh:
## Change Description

Rodeos fixes for rocksdb high write amplification and large number of sst files created 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/rodeos/rodeos.cpp
+++ b/libraries/rodeos/rodeos.cpp
@@ -132,8 +132,9 @@ void rodeos_db_snapshot::end_block(const get_blocks_result_base& result, bool fo
    irreversible_id = result.last_irreversible.block_id;
    if (!first || head < first)
       first = head;
-   if (write_now)
-      end_write(write_now);
+  
+      end_write(true);
+    
    if (near)
       db->flush(false, false);
 }
--- a/libraries/rodeos/rodeos.cpp
+++ b/libraries/rodeos/rodeos.cpp
@@ -120,7 +120,11 @@ void rodeos_db_snapshot::end_block(const get_blocks_result_base& result, bool fo
    if (!writing_block || result.this_block->block_num != *writing_block)
       throw std::runtime_error("call start_block first");
 
-   bool near       = result.this_block->block_num + 4 >= result.last_irreversible.block_num;
+   // if we are trying to get caught up to head and are already near (within 4 blocks) of LIB only then call db->flush to disk
+   bool near       = false;
+   if (result.this_block->block_num < result.last_irreversible.block_num)
+      near = result.this_block->block_num + 4 >= result.last_irreversible.block_num;
+    
    bool write_now  = !(result.this_block->block_num % 200) || near || force_write;
    head            = result.this_block->block_num;
    head_id         = result.this_block->block_id;
--- a/tests/rodeos_test.py
+++ b/tests/rodeos_test.py
@@ -171,7 +171,8 @@ try:
         loadSystemContract=False,
         specificExtraNodeosArgs={
             0: ("--plugin eosio::state_history_plugin --trace-history --chain-state-history --disable-replay-opts "
-                "--state-history-stride {0} " 
+          # stride unless it holds all the required data files -- it will cause test to fail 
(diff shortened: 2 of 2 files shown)


tests/rodeos_test.py
```diff
@@ -171,7 +171,8 @@ def get_info(self):
         loadSystemContract=False,
         specificExtraNodeosArgs={
             0: ("--plugin eosio::state_history_plugin --trace-history --chain-state-history --disable-replay-opts "
-                "--state-history-stride {0} " 
+          # stride unless it holds all the required data files -- it will cause test to fail 
+          #      "--state-history-stride {0} "
```
heifner, line 175: This should be removed instead of commented out. No need to have a `stateHistoryStride` that is not used.

libraries/rodeos/rodeos.cpp
```diff
@@ -120,16 +120,21 @@ void rodeos_db_snapshot::end_block(const get_blocks_result_base& result, bool fo
...
    head            = result.this_block->block_num;
    head_id         = result.this_block->block_id;
    irreversible    = result.last_irreversible.block_num;
    irreversible_id = result.last_irreversible.block_id;
    if (!first || head < first)
       first = head;
-   if (write_now)
-      end_write(write_now);
+  
+      end_write(true);
```
heifner, line 136: Fix indentation

libraries/rodeos/rodeos.cpp
```diff
@@ -120,16 +120,21 @@ void rodeos_db_snapshot::end_block(const get_blocks_result_base& result, bool fo
    if (!writing_block || result.this_block->block_num != *writing_block)
       throw std::runtime_error("call start_block first");
 
-   bool near       = result.this_block->block_num + 4 >= result.last_irreversible.block_num;
+   // if we are trying to get caught up to head and are already near (within 4 blocks) of LIB only then call db->flush to disk
+   bool near       = false;
+   if (result.this_block->block_num < result.last_irreversible.block_num)
+      near = result.this_block->block_num + 4 >= result.last_irreversible.block_num;
```
heifner, line 126: Once you are caught up with the chain, you never flush again? Seems like you will want a configurable flush threshold.

If I understand this `if`, you only flush for blocks: LIB-4, LIB-3, LIB-2, LIB-1. That seems like a really odd thing to do.

heifner (changes requested):

libraries/rodeos/rodeos.cpp
heifner: `force_write` and `write_now` are no longer used. They should be removed and the signature of the method should be updated.

heifner (changes requested):

libraries/rodeos/rodeos.cpp
nksanthosh: 1. Agree that having a configurable flush threshold will be useful. Will put that in. 

2. Agree with your comments on the odd looking `if` check. It was there from the prior implementation and I did not want to change that prior to demonstrating the fix. Once we get confirmation that the fix stands I will update this `if` condition as well.

libraries/rodeos/rodeos.cpp
nksanthosh: 1. Agree on indentation

2. Agree.  Will update the method signature once testing is completed.

tests/rodeos_test.py
nksanthosh: Agreed. Will take of it.

nksanthosh:
Superseded by later work for rodeos and SHiP changes

PR #10320, opened 2021-04-30
Added  Privacy Test Case #4

brianjohnson5972:
## Change Description
EPE-835
Added test case that starts up a 4+ producer network, adds all but one producer into the security group, verifies that the producers in the security group are building on each others blocks and that the excluded producer is on its own fork. After that, it adds the remaining producer into the security group and verifies the all eventually become in sync.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/SecurityGroup.py
+++ b/tests/SecurityGroup.py
@@ -151,9 +151,9 @@ class SecurityGroup(object):
             assert nonParticipantPostLIB < publishBlock, "Participants not in security group should not have advanced LIB to {}, but it has advanced to {}".format(publishBlock, nonParticipantPostLIB)
             nonParticipantHead = nonParticipant.getBlockNum()
             if nonParticipantHead > headAtTransFinalization:
-                producer = nonParticipants.getBlockProducerBy(nonParticipantHead)
+                producer = nonParticipant.getBlockProducerByNum(nonParticipantHead)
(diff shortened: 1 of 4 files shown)


tests/SecurityGroup.py
```diff
@@ -118,17 +119,18 @@ def copyIfNeeded(nodes):
...
         for part in self.participants:
             if part.pid is None:
                 continue
-            atLeastOne = True
             if part.waitForTransFinalization(transId) == None:
                 Utils.errorExit("Transaction: {}, never finalized".format(trans))
-        assert atLeastOne, "None of the participants are currently running, no reason to call verifyParticipantsTransactionFinalized"
+            headAtTransFinalized = part.getBlockNum()
+        assert headAtTransFinalized, "None of the participants are currently running, no reason to call verifyParticipantsTransactionFinalized"
+        return headAtTransFinalized
```
dimas1185, line 130: if we have an assert, do we really need a return here?

dimas1185 (approved):

PR #10328, opened 2021-04-30
scenario 3 test added to feature privacy tests

dimas1185:
## Change Description

scenario 3 from dogfooding + some refactoring based on framework changes
https://blockone.atlassian.net/wiki/spaces/EOSIOProdEng/pages/1434419366/Private+Chain+Access+Test+Plan

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [x] Existing Tests

- [x] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,265 @@
...
+    # after recorded block number (rbn1) becomes lib, verify production cycle continues without a fork
+    assert block1_id is not None and block1_id == block2_id
+
+    prod1_head    = prod1.getHeadBlockNum()
+    # on this point head should be already ahead of API node #1, but let's add idle wait just to ensure
+    prod1.waitForBlock(prod1_head + 20)
+
+    apiNode1 = cluster.getNode(2)
+    apiNode1_head = apiNode1.getHeadBlockNum()
+    Print("Node1 currently at {}".format(apiNode1_head))
```
brianjohnson5972, line 91: Since the participant for producer1 is "node1", to make the print statements clearer, I think we should use API Node 1 and API Node 2 instead of Node 1 and Node 2 in print statements (here and everywhere), or below you used apiNode1, either is fine, but make sure you use the same throughout

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,265 @@
...
+
+    # add node1 back to security group *
+    trans = securityGroup.editSecurityGroup(addNodes=[apiNode1])
+    # record block number (rbn5) transaction was added in
+    trans_block_num = trans["processed"]["block_num"]
+
+    # before last recorded block number (rbn5) is made irreversible, verify
+    # node1 and node2 do not receive blocks
+    #comparing here with trans_block_num - 12 to avoid possible flakiness when LIB advances inside while
+    while prod1.getIrreversibleBlockNum() <= trans_block_num - 12:
```
brianjohnson5972, line 212: make similar change to above to do get head on apiNode1 and 2 before getting lib from prod1

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,265 @@
...
+    # Restart shouldn't be needed here but due to bug in net_plugin we have to.
+    # this supposed to be fixed in nodeos and removed
+    apiNode2.kill(signal.SIGTERM)
+    apiNode2.relaunch(cachePopen=True)
+    #######################################################################
+    apiNode2.waitForBlock(trans_block_num, timeout=240, blockType=BlockType.lib, reportInterval=1)
+    Print("apiNode2 LIB block is now {}".format(trans_block_num))
+    apiNode2.waitForHeadToAdvance()
+
+    # remove node2 back to security group *
```
brianjohnson5972, line 234: change "back to" to "from"

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,265 @@
...
+    # record block number (rbn6) transaction was added in
+    trans_block_num = trans["processed"]["block_num"]
+
+    # after node1’s LIB advances to rbn5, verify
+    # node1 and node2 continue to sync
+    apiNode2_head = apiNode2.getHeadBlockNum()
+    Print("Node1 head = {} Node2 head = {}".format(apiNode2_head, apiNode1.getHeadBlockNum()))
+    apiNode2_Advancing = False
+    while apiNode1.getIrreversibleBlockNum() < trans_block_num:
+        temp = apiNode2.getHeadBlockNum()
```
brianjohnson5972, line 245: change to get data from apiNode2 before you get if from apiNode1, like above comments

brianjohnson5972 (changes requested):

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,265 @@
...
+
+    trans = securityGroup.editSecurityGroup(addNodes=[apiNode1])
+    # record block number (rbn1) transaction was added in
+    trans_block_num = trans["processed"]["block_num"]
+    Print("security group update [[\"node3\"]] published on block ".format(trans_block_num))
+    securityGroup.verifyParticipantsTransactionFinalized(trans["transaction_id"])
+
+    cluster.launchUnstarted(cachePopen=True)
+
+    prod1.waitForBlock(trans_block_num + 50)
```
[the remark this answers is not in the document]
dimas1185: I started unlaunched node above, so I needed some method for sleeping until it come up. Could you recommend anything better?

tests/privacy_scenario_3_test.py
dimas1185: `apiNode1_head` and `apiNode2_head` was not supposed to move till that moment. Maybe I misunderstood your comment?

brianjohnson5972: What I mean is (and similarly for other ones like this) is that if you talk to prod first, and get the number, then talk to the node(s) that are not supposed to change, it can lead to very intermittent failures, since the python script could retrieve the prod1 LIB, then not get processing time for a big gap, in which time prod1 could produce a block, move LIB and send to apiNode1.  I know it seems far fetched, but we have diagnosed similar things in the test framework.  But if you instead do:
while True:
    apiNode1_head = apiNode1.getHeadBlockNum()
    apiNode2_head = apiNode2.getHeadBlockNum()
    prod1Lib = prod1.getIrreversibleBlockNum()
    if prod1Lib >= trans_block_num:
        break
    ...
    
then you are guaranteed that you will stop before apiNode1_head and apiNode2_head should ever advance

tests/privacy_scenario_3_test.py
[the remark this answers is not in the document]
brianjohnson5972: I would remove this wait, then after you set apiNode2 do:
apiNode2.waitForBlock(trans_block_num, timeout=60)

60 would be way longer than you are waiting, based on what you have above, it would probably only wait about 13 seconds I think... after calling verifyParticipantsTransactionFinalized trans_block_num would be finalized, so we would be around trans_block_num+24(2 rounds of confirming producers), so 50-24=26 blocks -> 13 seconds.  I think 60 seconds should be plenty and it won't waste to much time if it really isn't advancing

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,290 @@
...
+    # node2 still receives no blocks or transactions
+    breakTimeout = 100
+    while breakTimeout > 0:
+        apiNode2Head = apiNode2.getHeadBlockNum()
+        apiNode1Lib  = apiNode1.getIrreversibleBlockNum()
+        if apiNode1Lib >= trans_block_num:
+            break
+        Print("API Node2 head block = {}".format(apiNode2Head))
+        assert apiNode2Head == 1
+        apiNode1.waitForNextBlock(sleepTime=0.5)
```
brianjohnson5972, line 148: assert apiNode1.waitforNextBlock(sleepTime=0.5) != None

this will just allow us to fail faster if apiNode1 head isn't moving

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,290 @@
...
+        curApiNode2_head = apiNode2.getHeadBlockNum()
+        Print("API Node2 head = {}".format(curApiNode2_head))
+        prod1LIB = prod1.getIrreversibleBlockNum()
+        if prod1LIB >= trans_block_num:
+            break
+        
+        assert curApiNode1_head == apiNode1_head
+        assert curApiNode2_head == apiNode2_head
+        
+        prod1.waitForNextBlock(sleepTime=0.5)
```
brianjohnson5972, line 240: assert ...wait.. != None

brianjohnson5972 (changes requested):

tests/privacy_scenario_3_test.py
```diff
@@ -0,0 +1,290 @@
...
+    trans = securityGroup.editSecurityGroup(addNodes=[prod1, prod2])
+    # Record block number (rbn1) transaction was added in
+    trans_block_num = prod1.getBlockNumByTransId(trans["transaction_id"], blocksAhead=3)
+    assert trans_block_num is not None
+    Print("security group update [[\"node1\", \"node2\"]] published on block {}".format(trans_block_num))
+
+    prod1.waitForTransFinalization(trans["transaction_id"])
+    prod2.waitForTransFinalization(trans["transaction_id"])
+
+    block1 = prod1.getBlock(trans_block_num)
```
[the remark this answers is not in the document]
dimas1185: we are doing this on line 70:
`trans_block_num = prod1.getBlockNumByTransId(trans["transaction_id"], blocksAhead=3)`

tests/privacy_scenario_3_test.py
dimas1185: that shouldn't result in flakiness as we have few seconds until API Node 1 LIB will be ahead of `trans_block_num` but I'll do just in case

tests/privacy_scenario_3_test.py
dimas1185: I found better solution, it does have `errorContext` string parameter and if it is not `None` it asserts

tests/privacy_scenario_3_test.py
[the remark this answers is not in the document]
brianjohnson5972: sorry, missed that

brianjohnson5972 (approved):

PR #10329, opened 2021-05-03
Initializing ssl_enabled.

brianjohnson5972:
## Change Description
Needed to initialize ssl_enabled to ensure that any nodeos not configured for TLS will function like it always has.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -217,7 +217,7 @@ namespace eosio {
    public:
       unique_ptr<tcp::acceptor>        acceptor;
       ssl_context_ptr                  ssl_context;
-      bool                             ssl_enabled;
+      bool                             ssl_enabled{false};
       std::atomic<uint32_t>            current_connection_id{0};
 
       unique_ptr< sync_manager >       sync_master;


heifner (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -217,7 +217,7 @@ namespace eosio {
    public:
       unique_ptr<tcp::acceptor>        acceptor;
       ssl_context_ptr                  ssl_context;
-      bool                             ssl_enabled;
+      bool                             ssl_enabled{false};
```
dimas1185, line 220: I suspect that an issue was how we initialize it:
```
   net_plugin::net_plugin()
      :my( new net_plugin_impl ) {
      my_impl = my.get();
   }
   ```
   `new net_plugin_impl{}` would fix it, isn't it?

brianjohnson5972: I'm not sure if that would fix it, but we do like to initialize primitive types explicitly and setting it explicitly to false fixes it.

Issue #10113, opened 2021-03-07
stop nodeos on ship error


cc32d9:
If state history plugin has problems writing the history, nodeos must stop immediately.

related issue: #9483 

Another condition is if the node starts from a snapshot that is newer than the state history's last block. The node moves along and does not write any history.

Currently it just prints a warning and goes along.

In case of an error, the operator needs to restart nodeos from a snapshot, and sometimes this condition is difficult to detect. Normally if someone installs a node with state history enabled, it's the primary job of the node to deliver history.

cc32d9:
A 2.0.13 WAX node encountered this error, and stopped as expected. But it can't restart, as state history is missing the block.

Failure where it stopped:
```
Sep 11 22:25:11 eosio nodeos[37573]: error 2021-09-11T22:25:11.972 nodeos    state_history_log.hpp:126     write_entry          ] chain::plugin_exception: 3110000 plugin_exception: Plugin exception
Sep 11 22:25:11 eosio nodeos[37573]: missed a block in trace_history.log
Sep 11 22:25:11 eosio nodeos[37573]:     {"name":"trace_history"}
Sep 11 22:25:11 eosio nodeos[37573]:     nodeos  state_history_log.hpp:93 write_entry
Sep 11 22:25:12 eosio nodeos[37573]: error 2021-09-11T22:25:12.011 nodeos    controller.cpp:1962           apply_block          ] e.to_detail_string(): 3140002 state_history_write_exception: State history write error
Sep 11 22:25:12 eosio nodeos[37573]: State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
Sep 11 22:25:12 eosio nodeos[37573]:     {}
Sep 11 22:25:12 eosio nodeos[37573]:     nodeos  state_history_log.hpp:134 write_entry
Sep 11 22:25:12 eosio nodeos[37573]:  
Sep 11 22:25:12 eosio nodeos[37573]: error 2021-09-11T22:25:12.079 nodeos    producer_plugin.cpp:361       on_incoming_block    ] 3140002 state_history_write_exception: State history write error
Sep 11 22:25:12 eosio nodeos[37573]: State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
Sep 11 22:25:12 eosio nodeos[37573]:     {}
Sep 11 22:25:12 eosio nodeos[37573]:     nodeos  state_history_log.hpp:134 write_entry
[...]

Issue #9483, opened 2020-09-10
ship stopped writing new blocks


cc32d9:
eosio-2.0.7, Ubuntu 18.04. The node has resynched Telos from genesis. All of a sudden, it stopped writing state history logs, without any messages related to the moment when it stopped (log level is set to error).

config.ini (`producer_api_plugin` was added several hours after this happened):
```
chain-state-db-size-mb = 65536
reversible-blocks-db-size-mb = 2048

wasm-runtime = eos-vm-jit
eos-vm-oc-enable = true

p2p-accept-transactions = false
api-accept-transactions = false

http-server-address = 0.0.0.0:8802
p2p-listen-endpoint = 0.0.0.0:9802
p2p-server-address = :9857

access-control-allow-origin = *

#verbose-http-errors = true
contracts-console = true

max-clients = 100

plugin = eosio::chain_plugin
plugin = eosio::chain_api_plugin
plugin = eosio::db_size_api_plugin

plugin = eosio::producer_api_plugin

plugin = eosio::state_history_plugin
trace-history = true
chain-state-history = true
trace-history-debug-mode = true
state-history-endpoint = 0.0.0.0:8082

sync-fetch-span = 10

p2p-peer-address = p2p.telosarabia.net:9876
p2p-peer-address = telos.cryptolions.io:9871
p2p-peer-address = telos.eosdublin.io:9876
```

```
root@eosio:/srv/telos# ls -al data/state-history/
total 291419499
drwxr-xr-x 2 root root            6 Sep  6 13:31 .
drwxr-xr-x 6 root root            6 Sep  6 13:31 ..
-rw-r--r-- 1 root root    831008728 Sep  9 08:12 chain_state_history.index
-rw-r--r-- 1 root root 117728914750 Sep  9 08:12 chain_state_history.log
-rw-r--r-- 1 root root    831008728 Sep  9 08:12 trace_history.index
-rw-r--r-- 1 root root 178821446038 Sep  9 08:12 trace_history.log
root@eosio:/srv/telos# ls -al data/blocks/
total 47089027
drwxr-xr-x 3 root root            5 Sep  6 13:31 .
drwxr-xr-x 6 root root            6 Sep  6 13:31 ..
-rw-r--r-- 1 root root    877516224 Sep 10 07:29 blocks.index
-rw-r--r-- 1 root root 128877141011 Sep 10 07:29 blocks.log
drwxr-xr-x 2 root root            3 Sep  6 13:31 reversible
```

[...]

PR #10332, opened 2021-05-03
state-history-exit-on-write-failure-2.0.x

huangminghuang:
## Change Description

This PR address IS #9483  and IS #10113 so that the nodeos would exit when the state history plugin is unable to write trace/delta because of some invariant violations. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

Add language to warn user that a nodoes node with state history plugin should not be configured as producer node because it may leave the fork db in an inconsistent state when it is unable to write traces/deltas to disk.

### DIFF
--- a/plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_log.hpp
+++ b/plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_log.hpp
@@ -124,9 +124,14 @@ class state_history_log {
       }
       catch(const chain::plugin_exception& e) {
          elog( "chain::plugin_exception: ${details}", ("details", e.to_detail_string()) );
-         elog("State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch "
-              "the process");
+         // Both app().quit() and exception throwing are required. Without app().quit(),
+         // the exception would be caught and drop before reaching main(). The exception is
+         // to ensure the block won't be commited. 
          appbase::app().quit();
+         EOS_THROW(
+             chain::state_history_write_exception,
+             "State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch "
+             "the process");
       }
    }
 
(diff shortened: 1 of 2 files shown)


plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_log.hpp
```diff
@@ -87,39 +87,45 @@ class state_history_log {
...
+         index.seek_end(0);
+         index.write((char*)&pos, sizeof(pos));
+         if (_begin_block == _end_block)
+            _begin_block = block_num;
+         _end_block    = block_num + 1;
+         last_block_id = header.block_id;
+      }
+      catch(const chain::plugin_exception& e) {
+         wlog( "chain::plugin_exception: ${details}", ("details", e.to_detail_string()) );
+         EOS_THROW(chain::controller_emit_signal_exception, "State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process")
```
heifner, line 127: The idea is to derive from `chain::controller_emit_signal_exception` not use it directly. See for example `checkpoint_exception`.

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_log.hpp
```diff
@@ -87,39 +87,47 @@ class state_history_log {
...
+         if (_begin_block == _end_block)
+            _begin_block = block_num;
+         _end_block    = block_num + 1;
+         last_block_id = header.block_id;
+      }
+      catch(const chain::plugin_exception& e) {
+         elog( "chain::plugin_exception: ${details}", ("details", e.to_detail_string()) );
+         elog("State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch "
+              "the process");
+         appbase::app().quit();
```
heifner, line 129: I think you should throw like you had before but also call `app().quit()`. You don't want the block produced because on restart you need to reprocess the block so it can be written to SHiP.

heifner (approved):

PR #10344, opened 2021-05-06
Update var prune_state after trxs in a signed_block get pruned

nickjjzhao:


heifner:
Don't forget to back-port to 2.1

tests/eosio_blocklog_prune_test.py
```diff
@@ -186,7 +186,7 @@
         Utils.Print("Post light validation waiting info for each node:\nproducer: {}\nfull: {},\nlight: {}".format(json.dumps(pvnPostInfo, indent=1), json.dumps(fvnPostInfo, indent=1), json.dumps(lvnPostInfo, indent=1)))
         Utils.Print("Post full validation waiting info for each node:\nproducer: {}\nfull: {},\nlight: {}".format(json.dumps(pvnPost2Info, indent=1), json.dumps(fvnPost2Info, indent=1), json.dumps(lvnPost2Info, indent=1)))
 
-    assert not headAdvanced, "the full validation node is still syncing"
+    assert (not headAdvanced) or (fvnPost2Info["head_block_num"] < cfTrxBlockNum), "the full validation node is still syncing"
```
heifner, line 189: Why were these two additions needed? `waitForHeadToAdvance()` should be enough, right?

nickjjzhao: It happened locally for full validation node that headAdvanced was `true` but the node hadn't synced up to the pruned block yet. Similar to the light validation node,  its headAdvanced was `false` but the node had already passed the pruned block. I think we should treat these situations as success not failure so I added two extra conditions. I might be wrong.

heifner (approved):

PR #10350, opened 2021-05-10
Keep all block logs in retained directory

shahabi8:
## Change Description
This PR resolves an issue on keeping all block logs in retained folder rather than moving them to archive. The default setting is changed from keeping 10 block logs in retained directory to keep all the logs there.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -276,12 +276,12 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
    cfg.add_options()
          ("blocks-dir", bpo::value<bfs::path>()->default_value("blocks"),
           "the location of the blocks directory (absolute path or relative to application data dir)")
-         ("blocks-log-stride", bpo::value<uint32_t>()->default_value(config::default_blocks_log_stride),
+         ("blocks-log-stride", bpo::value<uint32_t>(),
          "split the block log file when the head block number is the multiple of the stride\n"
          "When the stride is reached, the current block log and index will be renamed '<blocks-retained-dir>/blocks-<start num>-<end num>.log/index'\n"
          "and a new current block log and index will be created with the most recent block. All files following\n"
          "this format will be used to construct an extended block log.")
-         ("max-retained-block-files", bpo::value<uint32_t>()->default_value(config::default_max_retained_block_files),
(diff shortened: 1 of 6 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -281,7 +281,7 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
          "When the stride is reached, the current block log and index will be renamed '<blocks-retained-dir>/blocks-<start num>-<end num>.log/index'\n"
          "and a new current block log and index will be created with the most recent block. All files following\n"
          "this format will be used to construct an extended block log.")
-         ("max-retained-block-files", bpo::value<uint16_t>()->default_value(config::default_max_retained_block_files),
+         ("max-retained-block-files", bpo::value<uint32_t>()->default_value(config::default_max_retained_block_files),
```
spoonincode, line 284: should this be an optional configuration item (i.e. no default), instead of a configuration item that defaults to a user visible high number?

victorj8: @spoonincode do you mean to also add an `->implicit_value(config::default_max_retained_block_files)` ?

spoonincode: No I mean remove `->default_value(config::default_max_retained_block_files)`. As is, this shows up in the config file as

```
# the maximum number of blocks files to retain so that the blocks in those files can be queried.
# When the number is reached, the oldest block file would be moved to archive dir or deleted if the archive dir is empty.
# The retained block log files should not be manipulated by users. (eosio::chain_plugin)
# max-retained-block-files = 4294967295
```

shahabi8: @spoonincode, I addressed your comment above, please let me know if you have any other suggestions.

victorj8 (approved):

PR #10356, opened 2021-05-13
flakiness fix for forked chain test

dimas1185:
## Change Description

Fixes flakiness in `nodeos_forked_chain_lr_test`
Flakiness as caused by waitForBlock timeout that was default (60 seconds).
Sometimes it requires longer timeout because of producer schedule for 10 producers is 10 * (12 * 0.5) = 60.
So when wait starts simultaneously with wait for next producer schedule test fails because of error getting next block.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/nodeos_forked_chain_test.py
+++ b/tests/nodeos_forked_chain_test.py
@@ -376,6 +376,7 @@ try:
     # block number to start expecting node killed after
     preKillBlockNum=nonProdNode.getBlockNum()
     preKillBlockProducer=nonProdNode.getBlockProducerByNum(preKillBlockNum)
+    Print("preKillBlockProducer = {}".format(preKillBlockProducer))
     # kill at last block before defproducerl, since the block it is killed on will get propagated
     killAtProducer="defproducerk"
     nonProdNode.killNodeOnProducer(producer=killAtProducer, whereInSequence=(inRowCountPerProducer-1))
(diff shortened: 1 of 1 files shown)


tests/nodeos_forked_chain_test.py
```diff
@@ -417,19 +423,23 @@ def getBlock(self, blockNum):
         if not prodChanged:
             if preKillBlockProducer!=blockProducer0:
                 prodChanged=True
+                Print("prodChanged = True")
```
brianjohnson5972, line 426: Are we keeping these print statements for diagnosing more failures over time?

brianjohnson5972 (approved):

heifner:
Didn't fix the problem. https://buildkite.com/EOSIO/eosio-lrt/builds/4329

tests/nodeos_forked_chain_test.py
dimas1185: yes. it is hard to understand what is happening without those

PR #10358, opened 2021-05-13
Changed high transaction test to send token.kv transfers

brianjohnson5972:
## Change Description
Sets the token.kv contract and sends transfers to tokn.kv contract to better approximate the activity we are trying to diagnose.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/Cluster.py
```diff
@@ -1239,6 +1253,13 @@ def bootstrap(self, biosNode, totalNodes, prodCount, totalProducers, pfSetupPoli
             Utils.Print("ERROR: Failed to create account %s" % (eosioTokenAccount.name))
             return None
 
+        eosioKVTokenAccount=copy.deepcopy(eosioAccount)
+        eosioKVTokenAccount.name="token.kv"
+        trans=biosNode.createAccount(eosioKVTokenAccount, eosioAccount, 0)
+        if trans is None:
+            Utils.Print("ERROR: Failed to create account %s" % (eosioKVTokenAccount.name))
+            return None
+
```
huangminghuang, line 1262: Can we create the token.kv account only `if tokenType == TokenType.kv`?

tests/nodeos_high_transaction_test.py
```diff
@@ -285,7 +319,7 @@ def findTransInBlock(transId, transToBlock, node):
                     # delay and see if transfer is accepted now
                     Utils.Print("Transfer rejected, delay 1 second and see if it is then accepted")
                     time.sleep(1)
-                trans=node.transferFunds(fromAccount, toAccount, transferAmount, "transfer round %d" % (round), exitOnError=False, reportStatus=False, sign = True, dontSend = args.send_duplicates)
+                trans=transfer(node, contract, fromAccount, toAccount, transferAmount)
```
huangminghuang, line 322: If I understand it correctly, you are modifying the script only to use `token.kv` instead of having a configuration to choose between `eosio.token` and `token.kv`. Is that right?

brianjohnson5972: yes, but I went to the more generic use of "push action" with "transfer" than using the "transfer" command provided by cleos.

huangminghuang (approved):

Issue #8948, opened 2020-04-15
eosio-blocklog --trim-blocklog --first always fails


matthewdarwin:
I tried multiple chains and multiple values for "first" and every time it fails to prune the start off the blocks log.  Even got a segfault.

Seems like removing early blocks is really not supported.

Note: starting by copying blocks from a live running system.

```
rm -r /tmp/b/*; cp /var/lib/nodeos/blocks/blocks.index /var/lib/nodeos/blocks/blocks.log /tmp/b/; eosio-blocklog --trim-blocklog --first 2  --blocks-dir=/tmp/b
info  2020-04-15T13:09:51.451 eosio-blo block_log.cpp:963             trim_blocklog_front  ] In directory /tmp/b will trim all blocks before block 2 from blocks.log and blocks.index.
info  2020-04-15T13:09:51.451 eosio-blo block_log.cpp:1081            trim_data            ] block log version= 3
info  2020-04-15T13:09:51.451 eosio-blo block_log.cpp:1133            trim_data            ] first block= 1
info  2020-04-15T13:09:51.451 eosio-blo block_log.cpp:1134            trim_data            ] last block= 4424933
info  2020-04-15T13:09:51.508 eosio-blo block_log.cpp:888             write                ] block: 4194304      position in file: 806012308
info  2020-04-15T13:09:51.743 eosio-blo block_log.cpp:888             write                ] block: 3145728      position in file: 604684439
[1]    24492 segmentation fault  eosio-blocklog --trim-blocklog --first 2 --blocks-dir=/tmp/b
```

```
rm -r /tmp/b/*; cp /var/lib/nodeos/blocks/blocks.index /var/lib/nodeos/blocks/blocks.log /tmp/b/; eosio-blocklog --trim-blocklog --first 1000 --blocks-dir=/tmp/b
info  2020-04-15T13:08:44.752 eosio-blo block_log.cpp:963             trim_blocklog_front  ] In directory /tmp/b will trim all blocks before block 1000 from blocks.log and blocks.index.
info  2020-04-15T13:08:44.752 eosio-blo block_log.cpp:1081            trim_data            ] block log version= 3
info  2020-04-15T13:08:44.753 eosio-blo block_log.cpp:1133            trim_data            ] first block= 1
info  2020-04-15T13:08:44.753 eosio-blo block_log.cpp:1134            trim_data            ] last block= 4424801
[...]

PR #10370, opened 2021-05-17
Handle special case in trim_blocklog_front

nickjjzhao:
## Change Description

[EPE-895](https://blockone.atlassian.net/browse/EPE-895)  - GitHub Issue]( https://github.com/EOSIO/eos/issues/8948)

This line `uint64_t& pos_content = *(uint64_t*)(buf + buffer_index);` in method `trim_blocklog_front(`) can cause bad memory access exception when there are less than 8 bytes right after address  `buf + buffer_index` in the `buffer`, which is defined as `auto buffer =  make_unique<char[]>(detail::reverse_iterator::_buf_len);`

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/block_log.cpp
```diff
@@ -1036,18 +1036,40 @@ namespace eosio { namespace chain {
...
          EOS_ASSERT( num_read == 1, block_log_exception, "blocks.log read failed" );
 
+         // handle the case that the distance between `buf+buffer_index` and `the end of the buffer` is less than 8 bytes
+         uint64_t complete_pos_content;
+         bool re_write = false;
+         uint64_t re_write_offset;
+
          // walk this memory section to adjust block position to match the adjusted location
          // of the block start and store in the new index file
          while(original_pos >= start_of_blk_buffer_pos) {
```
huangminghuang, line 1059: The loop can be simplified to work within block boundaries only, for those doesn't fit into block boundaries, it should be re-read from the outer for-loop.

libraries/chain/block_log.cpp
```diff
@@ -1036,18 +1036,40 @@ namespace eosio { namespace chain {
...
             const auto buffer_index = original_pos - start_of_blk_buffer_pos;
-            uint64_t& pos_content = *(uint64_t*)(buf + buffer_index);
-            const auto start_of_this_block = pos_content;
-            pos_content = start_of_this_block - pos_delta;
-            index.write(pos_content);
-            original_pos = start_of_this_block - pos_size;
+            if (buffer_index + pos_size <= detail::reverse_iterator::_buf_len){
+                uint64_t& pos_content = *(uint64_t*)(buf + buffer_index);
```
huangminghuang: please use memcpy() instead

libraries/chain/block_log.cpp
nickjjzhao: Simplified the while loop. Will make another update if it's not the way you expected. Thanks!

libraries/chain/block_log.cpp
nickjjzhao: Since reference var `pos_content` is defined for both reading and writing, I replaced this line with two methods `read_buffer` and `write_buffer`.

libraries/chain/include/eosio/chain/block_log.hpp
```diff
@@ -81,6 +81,9 @@ namespace eosio { namespace chain {
 
          static bool trim_blocklog_front(const fc::path& block_dir, const fc::path& temp_dir, uint32_t truncate_at_block);
 
+         // used only for unit test to adjust the buffer length
+         static void set_buff_len(uint64_t len);
```
huangminghuang: This shouldn't be in the public interface. Just implement it in `block_log.cpp` as a regular function and forward declare it in`unittests/restart_chain_tests.cpp`.

unittests/restart_chain_tests.cpp
```diff
@@ -78,4 +78,64 @@ BOOST_AUTO_TEST_CASE(test_restart_with_different_chain_id)
...
+         break;
+      case buf_len_type::large:
+         block_log::set_buff_len(log_size);
+         break;
+      default:
+         return;
+   }
+
+   BOOST_CHECK( block_log::trim_blocklog_front(temp1.path, temp2.path, truncate_at_block) == true);
+}
```
huangminghuang, line 141: Could you follow the example in https://github.com/EOSIO/eos/blob/ce5031971d0878ca56456e1c72e6a1d8dd2248c0/unittests/restart_chain_tests.cpp#L651-L658 to verify the trimed block log is correct?

nickjjzhao: Verified.

libraries/chain/include/eosio/chain/block_log.hpp
nickjjzhao: Fixed.

libraries/chain/block_log.cpp
```diff
@@ -1185,3 +1211,8 @@ namespace eosio { namespace chain {
    }
 
    } } /// eosio::chain
+
+// used only for unit test to adjust the buffer length
+void set_buff_len(uint64_t len){
```
huangminghuang: The name of the function is too generic without any namespace. Consider to use `block_log_set_buff_len`.

nickjjzhao: Fixed.

huangminghuang (approved):

PR #10377, opened 2021-05-18
Refactor block tracking to avoid re-sending to sender - develop

heifner:
## Change Description

- Blocks were being echoed back to sender. Avoid sending to sender by updating block<->connection index before processing block.
- Remove have block tracking as it is no longer used.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -105,7 +105,6 @@ namespace eosio {
       block_id_type id;
       uint32_t      block_num = 0;
       uint32_t      connection_id = 0;
-      bool          have_block = false; // true if we have received the block, false if only received id notification
    };
 
    struct by_peer_block_id;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3508,11 +3504,13 @@ namespace eosio {
 
       go_away_reason reason = fatal_other;
       try {
+         my_impl->dispatcher->add_peer_block( blk_id, c->connection_id );
```
dimas1185, line 3507: what about moving this line after `try{}catch()` ? In that case we don't need to have `rm_peer_block` at all.

heifner: It has to be before `accept_block` because `accept_block` signals the sending of the block out to all the peers.

heifner: If the block did not validate or it asserted there is no reason to ever try the block again. If it was unlinkable that could be because it has not received the prior block yet for some reason

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2288,15 +2284,19 @@ namespace eosio {
       auto bptr = blk_state.get<by_id>().find( std::make_tuple( connection_id, std::ref( blkid )));
       bool added = (bptr == blk_state.end());
       if( added ) {
-         blk_state.insert( {blkid, block_header::num_from_id( blkid ), connection_id, true} );
-      } else if( !bptr->have_block ) {
-         blk_state.modify( bptr, []( auto& pb ) {
-            pb.have_block = true;
-         });
+         blk_state.insert( {blkid, block_header::num_from_id( blkid ), connection_id} );
```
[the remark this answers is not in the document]
heifner: It is cleared in `expire_blocks()`

dimas1185 (approved):

PR #10381, opened 2021-05-18
fixing flakiness in voting and forked chain tests

dimas1185:
## Change Description

Fixes:
1. nodeos_voting_lr_test. It was broken after bugfix in `Node.waitForIrreversibleBlock` where it was waiting for head instead of lib. After fix this increased execution time and lead to timeout fails.
2. nodeos_forked_chain_lr_test. It had flakiness due to race condition in head catchup and switching forks that lead to `go_away_message` being sent with status `benign_other` due to unknown block requested. Fix adding retry after closing connection.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [x] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/Node.py
+++ b/tests/Node.py
@@ -1221,7 +1221,10 @@ class Node(object):
         # The voted schedule should be promoted now, then need to wait for that to become irreversible
         votingTallyWindow=120  #could be up to 120 blocks before the votes were tallied
         promotedBlockNum=self.getHeadBlockNum()+votingTallyWindow
-        self.waitForIrreversibleBlock(promotedBlockNum, timeout=rounds/2)
+        # There was waitForIrreversibleBlock but due to bug it was waiting for head and not lib.
+        # leaving waitForIrreversibleBlock here slows down voting test by few minutes so since
+        # it was fine with head block for few years, switching to waitForBlock instead
+        self.waitForBlock(promotedBlockNum, timeout=rounds/2)
 
         ibnSchedActive=self.getIrreversibleBlockNum()
 
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3260,7 +3260,11 @@ namespace eosio {
       }
       if( msg.reason == wrong_version ) {
          if( !retry ) no_retry = fatal_other; // only retry once on wrong version
-      } else {
+      } 
+      else if ( msg.reason == benign_other ) {
+         if ( retry ) fc_dlog( logger, "received benign_other reason, retrying to connect");
+      }
+      else {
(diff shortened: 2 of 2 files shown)


tests/Node.py
```diff
@@ -1221,7 +1221,10 @@ def getNextCleanProductionCycle(self, trans):
         # The voted schedule should be promoted now, then need to wait for that to become irreversible
         votingTallyWindow=120  #could be up to 120 blocks before the votes were tallied
         promotedBlockNum=self.getHeadBlockNum()+votingTallyWindow
-        self.waitForIrreversibleBlock(promotedBlockNum, timeout=rounds/2)
+        # There was waitForIrreversibleBlock but due to bug it was waiting for head and not lib.
```
heifner, line 1224: Go ahead and fix `waitForIrreversibleBlock` which should be as simple as changing the default argument. It is fine to leave this as you have it, but should probably update this comment.

heifner (changes requested):

brianjohnson5972 (approved):

tests/Node.py
brianjohnson5972: I had fixed it, which is why this is taking longer. He could have extended the test, but since it had never failed before, we presumed it was fine to replace this call with what it had been doing before (unintentionally)

heifner: I'v fine with it using `waitForBlock`, but we should fix `waitForIrreversibleBlock` which is still:
```
    def waitForIrreversibleBlock(self, blockNum, timeout=WaitSpec.default(), blockType=BlockType.head):
        return self.waitForBlock(blockNum, timeout=timeout, blockType=blockType)
```

brianjohnson5972: ahh, what I fixed was waitForBlock to actually honor blockType, I missed that waitForIrreversibleBlock looked like this.  Which then makes me wonder why Dmytro's change had any effect.

dimas1185: @heifner perhaps you are looking at wrong branch.
```
    def waitForIrreversibleBlock(self, blockNum, timeout=WaitSpec.default()):
        return self.waitForBlock(blockNum, timeout=timeout, blockType=BlockType.lib)
```

heifner: Apparently so.

heifner (approved):

PR #10401, opened 2021-05-26
huangminghuang/ship-threaded-disk-write

huangminghuang:
## Change Description

Write state history traces/deltas logs into disk from another threads.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/state_history/log.cpp
+++ b/libraries/state_history/log.cpp
@@ -43,6 +43,26 @@ state_history_log::state_history_log(const char* const name, const state_history
    this->stride               = config.stride;
    open_log(config.log_dir / (std::string(name) + ".log"));
    open_index(config.log_dir / (std::string(name) + ".index"));
+
+   thr = std::thread([this] {
+      while (!ending.load()) {
+         std::unique_lock lock(mx);
+         cv.wait(lock, [this]{ 
+            return ending.load() || !entries.empty(); 
+         });
+         
+         if (entries.size()) {
+            write_entry(entries.begin()->second);
+            entries.erase(entries.begin());
+         }
+      }
+   });
+}
+
+state_history_log::~state_history_log() {
+   ending = true;
+   cv.notify_one();
+   thr.join();
 }
 
 void state_history_log::read_header(state_history_log_header& header, bool assert_version) {
@@ -69,6 +89,7 @@ void state_history_log::get_entry_header(state_history_log::block_num_type block
 }
 
 std::optional<chain::block_id_type> state_history_log::get_block_id(state_history_log::block_num_type block_num) {
+   std::lock_guard lock(mx);
    auto result = catalog.id_for_block(block_num);
    if (!result && block_num >= _begin_block && block_num < _end_block) {
       state_history_log_header header;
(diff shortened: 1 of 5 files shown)


libraries/state_history/log.cpp
```diff
@@ -288,21 +309,59 @@ void state_history_log::split_log() {
...
+std::shared_ptr<std::vector<char>> state_history_traces_log::get_log_entry(block_num_type block_num) {
+   
+   std::lock_guard<std::mutex> lk(mx);
+   
+   auto itr = entries.find(block_num);
+   if (itr != entries.end()) {
+      std::vector<state_history::transaction_trace> traces;
+      auto ds = fc::datastream<const char*>(itr->second.data->data(), itr->second.data->size());
+      state_history::trace_converter::unpack(ds, traces);
+      return std::make_shared<std::vector<char>>(fc::raw::pack(traces));
```
brianjohnson5972, line 351: Do you think the pack can take long enough that it would be worth releasing the lock here?

libraries/state_history/log.cpp
```diff
@@ -288,21 +309,59 @@ void state_history_log::split_log() {
...
+   }
```
brianjohnson5972, line 352: The lock is only for entries, so we can release it after this line, right?

unittests/state_history_tests.cpp
```diff
@@ -169,18 +169,18 @@ BOOST_AUTO_TEST_CASE(test_trace_log) {
    BOOST_REQUIRE(!std::holds_alternative<eosio::ship_protocol::prunable_data_type::none>(
        get_prunable_data_from_traces(traces, cfd_trace->id)));
 
-   std::vector<transaction_id_type> ids{cfd_trace->id};
-   log.prune_transactions(cfd_trace->block_num, ids);
-   BOOST_REQUIRE(ids.empty());
+//    std::vector<transaction_id_type> ids{cfd_trace->id};
```
brianjohnson5972, line 172: Any reason to save this commented out code?

brianjohnson5972 (approved):

libraries/state_history/log.cpp
huangminghuang: > Do you think the pack can take long enough that it would be worth releasing the lock here?

From the KV heavy contract, the traces have never been big enough to cause any significant impact, that's why I didn't try to make the lock more finer grained.

libraries/state_history/log.cpp
huangminghuang: > The lock is only for entries, so we can release it after this line, right?

yes

unittests/state_history_tests.cpp
huangminghuang: > Any reason to save this commented out code?

Pushed new commit for uncommenting these lines.

PR #10100, opened 2021-03-01
Standardize JSON formatting of action data (epe 151)

softprofe:
## Change Description

Start 100100 description ......

EPE 151

old code doing this:
  act.data ------can parse to json object ? if yes   "data"=json object  and  hex_data = "hex string"
                                                                   if no     "data" = "hex string"

new code doing this:
  act.data ------can parse to json object ? if yes   "data"=json object  and  hex_data = "hex string"
                                                                   if no     "hex_data" = "hex string"

and extract function also some changes correspondingly.

End 10100 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #10405, opened 2021-05-28
test for existence of "data" field when printing action result, introduced by PR #10100

praphael:
fix cleos bug  where "data" field does not exist in action results, introduced by PR #10100 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner (changes requested):
Can you add a test that would fail before and now work.

softprofe (changes requested):
The change looks safe to me, it only add a check for act["data"], however, add test case(s) will make this PR perfect. Please test for both the act['data'] exist/not exist before merge.

praphael:
Ok I added a test, in order to test run 
`ctest --output-on-failure -L cleos_test`
It should pass

To check against older branch, copy cleos executable into programs/cleos.  The test should now fail. You might want to make a backup of cleos executable here before you do this.

jgiszczak (approved):
Being a little less cavalier about our assumptions of what nodeos returns is an improvement we should embrace.

heifner (approved):

softprofe (approved):

Issue #3678, opened 2018-05-31
Migrate http_plugin to boost beast


wanderingbort:
Right now we are using websocketpp (and an old version to boot) for our HTTP-RPC server.  

One of the most unfortunate side effects of this is that when handling http requests it will advertise HTTP/1.1 but does not support keep-alive.  We have made a small change so that it properly uses the `Connection: close` header (as the default value for that header is `keep-alive` in HTTP/1.1) to notify the peer that it will be closing the connection.  However, it would be a huge win for performance of cleos and the new out-of-process key signing if connections could persist longer. 

This would amortize the cost of TCP and TLS connections/handshaking down to almost nothing.

brianjohnson5972:
@wanderingbort This looks like it is OBE (overcome by events), is it?  I will presume it is if I hear nothing.

aclark-b1:
In order to focus our efforts on issues that are currently creating difficulty for the community we are closing tickets that were created prior to the EOSIO 2.0 release.  If you believe this issue is still relevant please feel free to reopen it or create a new one.  Thank you for your continued support of EOSIO!

spoonincode:
This needs to be done, and I'm not sure if it's tracked anywhere else at the moment

jeffreyssmith2nd:
This is tracked in Jira. I'll defer to @aclark-b1 but I would prefer we only track in one place and close this.

spoonincode:
if it's in jira that's fine, I wasn't sure if it was or not

PR #10415, opened 2021-06-04
migrate http_plugin to boost::beast GH issue #3678 / EPE-9

praphael:
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,511 @@
...
+
+                auto self = shared_from_this();
+
+            
+                // We need to be executing within a strand to perform async operations
+                // on the I/O objects in this session.
+                asio::dispatch(stream_.get_executor(), [self]() {
+                    // Set the timeout.
+                    
+                    beast::get_lowest_layer(self->derived().stream_).expires_after(
```
spoonincode, line 450: `expires_after()` has been notoriously problematic for us in other areas, @heifner can probably elaborate more.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,511 @@
...
+
+                // Read another request
+                do_read();
+            }
+
+
+            // TODO 
+            bool allow_host(const http::request<http::string_body>& req) {
+                /*
+                bool is_secure = con->get_uri()->get_secure();
```
spoonincode, line 262: code could use a pass through to remove commented out code

plugins/http_plugin/http_plugin.cpp
```diff
@@ -766,15 +634,19 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
              "Additionaly acceptable values for the \"Host\" header of incoming HTTP requests, can be specified multiple times.  Includes http/s_server_address by default.")
             ("http-threads", bpo::value<uint16_t>()->default_value( my->thread_pool_size ),
              "Number of worker threads in http thread pool")
+            ("http-use-beast", bpo::value<bool>()->default_value(true),
```
spoonincode, line 637: I'm surprised to see this an option. Was leaving in the old impl a requirement?

plugins/http_plugin/http_plugin.cpp
```diff
@@ -880,7 +764,7 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
                try {
                   my->unix_server.clear_access_channels(websocketpp::log::alevel::all);
                   my->unix_server.init_asio( &my->thread_pool->get_executor() );
-                  my->unix_server.set_max_http_body_size(my->max_body_size);
+                  my->unix_server.set_max_http_body_size(my->plugin_state->max_body_size);
```
spoonincode, line 767: unix socket wasn't converted over to beast

spoonincode:
There are at least a couple usages of Boost 1.70+ in here such as `bind_front_handler()` and `expires_after()`. Currently our minimum requirement is 1.67. It would be worthwhile exploring how difficult it would be to continue supporting 1.67 as imo we should be prudent in bumping requirements.

plugins/http_plugin/http_plugin.cpp
praphael: @spoonincode I left that in for testing purposes ... although once its well tested it should be removed

plugins/http_plugin/http_plugin.cpp
praphael: @spoonincode Yes this was an oversight, needs to be done

plugins/http_plugin/beast_http_session.hpp
praphael: @heifner Could you elaborate on problem with `expires_after()` ?

plugins/http_plugin/beast_http_session.hpp
praphael: This should now be cleaned up

plugins/http_plugin/http_plugin.cpp
praphael: @spoonincode Implemented, tested with keosd (uses UNIX sockets by default)

plugins/http_plugin/beast_http_session.hpp
heifner: https://github.com/EOSIO/eos/pull/10252

praphael:
> There are at least a couple usages of Boost 1.70+ in here such as `bind_front_handler()` and `expires_after()`. Currently our minimum requirement is 1.67. It would be worthwhile exploring how difficult it would be to continue supporting 1.67 as imo we should be prudent in bumping requirements.

@spoonincode Where is the min. requirement for Boost 1.67?   The build script will install Boost 1.72 if you have an earlier version of Boost.  

Also, I'm getting other feedback that running older versions of Beast presents a security issue, so Boost 1.72 may need to be upgraded in any case.

plugins/http_plugin/common.hpp
```diff
@@ -0,0 +1,293 @@
...
+    auto make_http_response_handler( asio::io_context &ioc, http_plugin_state &plugin_state, detail::abstract_conn_ptr abstract_conn_ptr) {
+    return [&ioc, &plugin_state, abstract_conn_ptr]( int code, std::optional<fc::variant> response )
+         {
+            auto tracked_response = make_in_flight(std::move(response), plugin_state);
+            if (!abstract_conn_ptr->verify_max_bytes_in_flight()) {
+                return;
+            }
+
+            // post  back to an HTTP thread to to allow the response handler to be called from any thread
+            boost::asio::post( ioc, // my->thread_pool->get_executor()
```
heifner, line 219: Are you planning on adding back the use of a thread pool?

spoonincode:
The base CMakeLists.txt lists the requirements, it's defined as boost 1.67 there.

Yeah if there are documented security issues with some versions of boost it may be a good idea to bump the requirement appropriately. We're at 1.67 because websockets are pretty much broken in 1.66. But I didn't see any security issues explicitly called out in beast's release notes on a cursory look. Both security reviews -- including one done pre-boost inclusion -- look clean:
https://www.boost.org/doc/libs/1_76_0/libs/beast/doc/html/beast/quick_start/security_review_bishop_fox.html

At the moment this PR doesn't build with 1.67. Unless we find a security issue, personally I don't see any compelling reason to drop support because of one or two functions.

plugins/http_plugin/http_plugin.cpp
spoonincode: Does that mean before the PR is merged? What would be the criteria for removing it?

praphael: My plan was for sometime after it has been used enough to be sure there are no major issues related to the beast implementation vs. websocketpp.  The run-time switch makes it easier to isolate if a problem is caused by the beast implementation.    Once the version it is included in reaches stable (probably 2.2), it can be removed

praphael:
@spoonincode  Supporting multiple boost versions does not appear to be trivial.  Boost::beast's own examples in 1.67 will not compile in 1.72, and the examples are substantially different apart from std::bind vs. bind_front_handler()   Unless we want to sprinkle the code liberally with #ifdef or have multiple versions of the same files, this might be a tough problem.

spoonincode:
It wouldn't surprise me if some very small amount of ifdef'ing is required. But as long as we're not using the 1.70+ beast::tcp_stream/basic_stream (which seems like a bad idea given the headache's we've had before), I'm fairly optimistic that the differences will be minimal. What concerns are you seeing?

praphael:
> It wouldn't surprise me if some very small amount of ifdef'ing is required. But as long as we're not using the 1.70+ beast::tcp_stream/basic_stream (which seems like a bad idea given the headache's we've had before), I'm fairly optimistic that the differences will be minimal. What concerns are you seeing?

I'm in the middle of this work right now.   Its definitely possible, it just looks ... ugly.  Should finish up this afternoon

There might be a way to dig.deeper to produce a version which is compatible with both without resorting #if/#else/#endif macros.     I was mostly basing my work off the examples, in particular. https://www.boost.org/doc/libs/1_67_0/libs/beast/example/http/server/flex/http_server_flex.cpp. (although I do not detect SSLat runtime)   There does seem to be a significant difference in how 1.70+ handles strands and streams, at least exposed through the Beast API. 
 Here is the example for 1.70 https://www.boost.org/doc/libs/1_70_0/libs/beast/example/http/server/flex/http_server_flex.cpp . 
 Alternatively, we could only use the Beast parser/serializer, and just use ASIO for the network portion.

praphael:
> It wouldn't surprise me if some very small amount of ifdef'ing is required. But as long as we're not using the 1.70+ beast::tcp_stream/basic_stream (which seems like a bad idea given the headache's we've had before), I'm fairly optimistic that the differences will be minimal. What concerns are you seeing?

What are the headaches you've had with beast::tcp_streams ?

heifner:
> What are the headaches you've had with beast::tcp_streams ?

This for one: https://github.com/EOSIO/eos/pull/10252

praphael:
> > What are the headaches you've had with beast::tcp_streams ?
> 
> This for one: #10252

I have removed all the expires_after() calls, they aren't necessary to use the streams to my knowledge.   I think some type of timeout mechanism should be investigated, however. 

@spoonincode 
The current version uses streams for boost >= 1.70, and sockets for boost 1.67 - 1.69.   I've tested it compiles with boost 1.67, 1.72 and 1.75, and it is passing all checks save the long running tests which are only run during nightlys.  But if you think its worth investigating having all versions use sockets I can reevaluate.

spoonincode:
I agree that many ifdefs makes it ugly and unwieldy. Why can't we use just always use sockets no matter the boost version?

Another thing I've been thinking about is that I'm skeptical we need strands. Due to the half-duplex nature of HTTP we're always either reading or writing. There is never a possibility that, say, an async_read completes while another thread is in the middle of a composed async_write. This is in contrast to the p2p & ship protocols. We should think this through some as it could be a performance benefit to avoid strand locking.

praphael:
> I agree that many ifdefs makes it ugly and unwieldy. Why can't we use just always use sockets no matter the boost version?
> 
> Another thing I've been thinking about is that I'm skeptical we need strands. Due to the half-duplex nature of HTTP we're always either reading or writing. There is never a possibility that, say, an async_read completes while another thread is in the middle of a composed async_write. This is in contrast to the p2p & ship protocols. We should think this through some as it could be a performance benefit to avoid strand locking.

> I agree that many ifdefs makes it ugly and unwieldy. Why can't we use just always use sockets no matter the boost version?
> 
> Another thing I've been thinking about is that I'm skeptical we need strands. Due to the half-duplex nature of HTTP we're always either reading or writing. There is never a possibility that, say, an async_read completes while another thread is in the middle of a composed async_write. This is in contrast to the p2p & ship protocols. We should think this through some as it could be a performance benefit to avoid strand locking.

With respect to a single connection, you may be correct.  The primary benefit of the concurrency, I would suspect, is handling multiple connections.and in the processing, not on I/O on a single connection.   But that is something we can test.  For the UNIX sockets,  I disabled the async I/O because there were some compilation issues and it made sense to me that in the case of local connections you wouldn't have more than one connection at a time.  I can add something similar to TCP connections.

praphael:
> > I agree that many ifdefs makes it ugly and unwieldy. Why can't we use just always use sockets no matter the boost version?
> > Another thing I've been thinking about is that I'm skeptical we need strands. Due to the half-duplex nature of HTTP we're always either reading or writing. There is never a possibility that, say, an async_read completes while another thread is in the middle of a composed async_write. This is in contrast to the p2p & ship protocols. We should think this through some as it could be a performance benefit to avoid strand locking.
> 
> > I agree that many ifdefs makes it ugly and unwieldy. Why can't we use just always use sockets no matter the boost version?
> > Another thing I've been thinking about is that I'm skeptical we need strands. Due to the half-duplex nature of HTTP we're always either reading or writing. There is never a possibility that, say, an async_read completes while another thread is in the middle of a composed async_write. This is in contrast to the p2p & ship protocols. We should think this through some as it could be a performance benefit to avoid strand locking.
> 
> With respect to a single connection, you may be correct. The primary benefit of the concurrency, I would suspect, is handling multiple connections.and in the processing, not on I/O on a single connection. But that is something we can test. For the UNIX sockets, I disabled the async I/O because there were some compilation issues and it made sense to me that in the case of local connections you wouldn't have more than one connection at a time. I can add something similar to TCP connections.

Let me revise my statement.  With a single request per connection, the above is true - there is likely no benefit to async I/O on a single connection as the connection will make one request, then will close.   However, with keep-alive, there could potentially be a benefit to async I/O even on a single connection.   We have four basic stages for responding to an HTTP request. 1)read  2)parse. 3)process. 4)write.   Let's assume each of those stages takes one unit of time (yes, this is a a gross simplification, just bear with me) With a synchronous strategy, we have:

0. read request 1
1. parse request 1
2. process request 1
3. write response 1
4. read request 2
5. parse request 2
6. process request 2
7.  write response 2
etc...

However, with async you could have the following
0. read request 1
1. parse request 1.  read request 2
2. process request 1, parse request 2, read request 3
3. write response 1,  process request 2, parse request 3,  reqd request 4
5. write response 2, process request 3, parse request 4, read request 5
etc...

So you would be pipelining and could respond much faster.   However, you would have to dedicate more threads to get the benefit.   It could be possible that an async strategy like this would be slower, if you didn't have enough threads dedicated to handle the number fo requests, due to overhead of context switching/ CPU cache misses, etc.   Its also possible, that even if you did have enough threads, the CPU cache misses make it not worth it.   It is hard to say what the best strategy is.   Also, for most of our use cases with HTTP, the request/response bodies should not be very large.  The only exception I can think if is uploading/downloading contracts, but that is done infrequently.  Well that and perhaps get_block calls with lots of actions in the block    .

praphael:
With respect to the above comment about pipelining, I realized our code as it is structured, cannot take advantage of this without significant changes.   There is only one copy of the read buffer, parser and serializer.  With multiple concurrent requests within a single session object, simultaneously processing would overwrite the objects and cause unpredictable behavior.    This could be solved by queues or keeping all copies local and passing copies but its not worth it at the moment for the use cases of http_plugin. 

I feel the best course of action, for now, is to have the session/connection objects be totally synchronous, and am making changes to that effect.   The acceptor/listener will remain asynchronous - so you could have multiple concurrent connections that take advantage of multiple threads.    I will retain the "scaffolding" of completion handlers (the on_xxx functions) in case we want to revisit this at some point.

praphael:
Updates made - switched to sync operation within session which substantially simplified code.  I did the best I could to minimize the #if / #else segments, but there does appear to be to be substantial breaking changes introduced in boost::asio 1.70, particularly with regards to creation of strands.    I've tested it compiles under boost 1.67, 1.72 and 1.75 

One issue which remains is timeouts.  Right now, its calling blocking I/O operations for reads/writes, with no timeouts (which boost::asio does not expose for sync read/write calls) clients can easily hang the connection.    Can you clarify whether the opposition to expires_after() was in beast implemenation, or was it a problem with asio  timer classes such as asio::steady_timer?  Because it shouldn't be hard to add a steady_timer to take care of this problem and kill the connection after a certain period.  Alternatively, we could go with some other approach, like setting some type fo time limit on the entire session and monitoring this time within the listener classes and then killing the strand when  new connection attempt is made and the time limit is exceeded.  This way would not tie up a strand/thread for a timer.   @heifner @spoonincode

praphael:
Closing this PR for now - current implementation has fairly obvious flaw.  Currently testing if prior async version also has flaw.  Need to have discussion about the way forward..

PR #10418, opened 2021-06-09
Remove support of MacOS 10.14, see also EPE 979

softprofe:
## Change Description

Remove support of MacOS 10.14, see  also EPE 979

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [x] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


.cicd/generate-pipeline.sh
```diff
@@ -732,12 +697,12 @@ cat <<EOF
...
   - label: ":beer: Brew Updater"
     command:
-      - "buildkite-agent artifact download eosio.rb . --step ':darwin: macOS 10.14 - Package Builder'"
+      - "buildkite-agent artifact download eosio.rb . --step ':darwin: macOS 10.15 - Package Builder'"
       - "buildkite-agent artifact upload eosio.rb"
     agents:
       queue: "automation-basic-builder-fleet"
     timeout: "${TIMEOUT:-5}"
-    skip: ${SKIP_PACKAGE_BUILDER}${SKIP_MAC}${SKIP_MACOS_10_14}
+    skip: ${SKIP_PACKAGE_BUILDER}${SKIP_MAC}
```
kj4ezj, line 705: This line...
```yaml
skip: ${SKIP_PACKAGE_BUILDER}${SKIP_MAC}${SKIP_MACOS_10_14}
```
...should become...
```yaml
skip: ${SKIP_PACKAGE_BUILDER}${SKIP_MAC}${SKIP_MACOS_10_15}
```
...as this step now depends upon Catalina. If Catalina is skipped (by setting `SKIP_MACOS_10_15`), this step will fail.

Perhaps that is the wrong thing to do and this step should be pulling `brew` packages from macOS 11, but that is a larger discussion. For the purposes of this story, setting it to Catalina is equivalent to how it behaved before the upgrade.

scripts/generate_bottle.sh
```diff
@@ -1,16 +1,16 @@
 #!/usr/bin/env bash
 set -eo pipefail
 
-VERS=`sw_vers -productVersion | awk '/10\.14\..*/{print $0}'`
+VERS=`sw_vers -productVersion | awk '/10\.15\..*/{print $0}'`
 if [[ -z "$VERS" ]]; then
-   VERS=`sw_vers -productVersion | awk '/10\.15.*/{print $0}'`
+   VERS=`sw_vers -productVersion | awk '/11\.*.*/{print $0}'`
```
kj4ezj, line 6: The `\.*` in this regular expression does not make sense.
```bash
VERS=`sw_vers -productVersion | awk '/11\.*.*/{print $0}'`
```
Since the dot (`.`) is not required, it does nothing. We can either make the dot required...
```bash
VERS=`sw_vers -productVersion | awk '/11[.].*/{print $0}'`
```
-or-
```bash
VERS=`sw_vers -productVersion | awk '/11\..*/{print $0}'`
```
...or remove it altogether:
```bash
VERS=`sw_vers -productVersion | awk '/11.*/{print $0}'`
```

kj4ezj (changes requested):

scripts/generate_bottle.sh
softprofe: I like your second way (VERS=`sw_vers -productVersion | awk '/11\..*/{print $0}'`)
 , I think the first dot after 11 is better to stay there, and then after this dot could be anything (.*) .

.cicd/generate-pipeline.sh
softprofe: Now it is skip: ${SKIP_PACKAGE_BUILDER}${SKIP_MAC}${SKIP_MACOS_10_15}

kj4ezj (approved):
🗑️

PR #10429, opened 2021-06-15
Test packaged EOSIO Linux artifacts against standard dockers.

williamblevins:
## Change Description

Updates for CICD to test packaged EOSIO Linux artifacts against standard dockers: `*.deb` and `*.rpm` files.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [x] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/test-package.sh
+++ b/.cicd/test-package.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+SCRIPT_NAME='test-package.gen.sh'
+
+cat > $SCRIPT_NAME <<EOL
+#!/bin/bash
+set -eu
+
+if [[ \$(apt-get --version 2>/dev/null) ]]; then # debian family
+    UPDATE='apt-get update'
+    echo "$ \$UPDATE"
+    eval \$UPDATE
+    echo '--- :minidisc: Install EOSIO'
+    INSTALL="apt-get install -y /eos/*.deb"
+    echo "$ \$INSTALL"
+    eval \$INSTALL
+elif [[ \$(yum --version 2>/dev/null) ]]; then # RHEL family
+    UPDATE='yum check-update || :'
+    echo "$ \$UPDATE"
+    eval \$UPDATE
+    INSTALL="yum install -y /eos/*.rpm"
+    echo "$ \$INSTALL"
+    eval \$INSTALL
+else
+    echo 'ERROR: Package manager not detected!'
+    exit 3
+fi
+
+nodeos --full-version
+EOL
+
+chmod +x $SCRIPT_NAME
+
+docker pull $IMAGE
+docker run --rm -v "$(pwd):/eos" -w '/eos' -it $IMAGE ./$SCRIPT_NAME
--- a/.cicd/test-package.sh
+++ b/.cicd/test-package.sh
@@ -1,35 +1,7 @@
 #!/bin/bash
-
-SCRIPT_NAME='test-package.gen.sh'
-
-cat > $SCRIPT_NAME <<EOL
-#!/bin/bash
 set -eu
 
-if [[ \$(apt-get --version 2>/dev/null) ]]; then # debian family
-    UPDATE='apt-get update'
-    echo "$ \$UPDATE"
-    eval \$UPDATE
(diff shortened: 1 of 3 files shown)


sandie06 (approved):
LGTM

.cicd/test-package.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
+SCRIPT_NAME='test-package.gen.sh'
+
+cat > $SCRIPT_NAME <<EOL
```
kj4ezj, line 5: If we wrap the opening delimiter with single-quotes...
```bash
cat > $SCRIPT_NAME <<'EOL'
```
...then it [won't](https://stackoverflow.com/a/27921346) perform shell interpolation.

While your code is already working and it might suck to change it, I think it will prevent errors for anyone who has to edit it moving forward and forgets an escape sequence. Thoughts?

I, personally, prefer a [recursive solution](https://github.com/EOSIO/eos-vm/blob/615a19214c56473a0508226b5c8cc3aa75b7766f/.cicd/build.sh#L4) because then all the code has syntax highlighting and lives in the same file but I guess I never had time to make the same changes in eos as I did in eos-vm.

.cicd/test-package.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+
```
kj4ezj, line 2: Consider another log folding group here to match the style of the other job steps.
```bash
echo '--- :evergreen_tree: Configuring Environment'
```

.cicd/test-package.sh
```diff
@@ -0,0 +1,35 @@
...
+else
+    echo 'ERROR: Package manager not detected!'
+    exit 3
+fi
+
+nodeos --full-version
+EOL
+
+chmod +x $SCRIPT_NAME
+
```
kj4ezj, line 33: Consider another log folding group here to match the style of the other job steps.
```bash
echo '--- :docker: Selecting Container'
```

kj4ezj (dismissed):
Mostly just concerned with the first three comments.

.cicd/test-package.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
williamblevins: Yes, we should. I'm going to add `-u` also. I did it for the test script, but I guess not the wrapper...

.cicd/test-package.sh
```diff
@@ -0,0 +1,35 @@
...
+
+SCRIPT_NAME='test-package.gen.sh'
+
+cat > $SCRIPT_NAME <<EOL
+#!/bin/bash
+set -eu
+
```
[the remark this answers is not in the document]
williamblevins: ok. Should I put the the `nodeos --full-version` into a separate fold since its a basic test layer that nodeos runs after installed? We might put other stuff there eventually.

.cicd/test-package.sh
williamblevins: I'm fine with changing it for readability. I avoid escaping when I can.

.cicd/test-package.sh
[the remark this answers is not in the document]
kj4ezj: Up to you. If you do, I would make both the installation and invocation expanded log groups.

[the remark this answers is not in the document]
kj4ezj: Emojii are [here](https://github.com/buildkite/emojis).

.cicd/test-package.sh
williamblevins: Hmm, if we do it this way, vim stops highlighting the EOL block though... ;(

zach-cpu (approved):

Issue #10255, opened 2021-04-16
Zipkin: connection retry needs re-work


matthewdarwin:
According to https://developers.eos.io/manuals/eos/v2.1/nodeos/logging/third_party_logging/zipkin_tracer
<<
telemetry-timeout-us, optional parameter, specifies the timeout for sending zipkin spans with default value set to 200000. Zipkin tracer sends zipkin spans, in json format, to the Zipkin service via HTTP POST requests on a separate thread. telemetry-timeout-us specifies how long to allow each HTTP POST request before reaching the time out and thus fail. Nine consecutive failures will disable the Zipkin tracer until a restart of nodeos daemon occurs.
>>

on networks like EOS Mainnet, 9 attempts is very small and useless for intermittent failures like running a zipkin upgrade.

For production use, need to work more like:

- retry every 30 seconds
- have health reporting as to if connected or not
- process SIGHUP or similar to force re-connect

matthewdarwin:
Further if `telemetry-url` is a DNS name that points to multiple A or AAAA records, nodeos should try all the addreses returned before giving up.  It may already do this, I didn't test.  (probably doesn't use AAAA records)

PR #10432, opened 2021-06-16
Update Zipkin connection retry - develop

nickjjzhao:
## Change Description

[EPE-933](https://blockone.atlassian.net/browse/EPE-933) [GitHub:issue:[10255](https://github.com/EOSIO/eos/issues/10255)]

on networks like EOS Mainnet, 9 attempts is very small and useless for intermittent failures like running a zipkin upgrade.

For production use, need to work more like:

1. retry every 30 seconds
2. have health reporting as to if connected or not
3. process SIGHUP or similar to force re-connect
4. Further if `telemetry-url` is a DNS name that points to multiple A or AAAA records, nodeos should try all the addreses returned before giving up 

Notes:
- Existing code supports item 4 above. 
- Changes made in` fc submodule` https://github.com/EOSIO/fc/pull/200 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Documentation Additions
- [x] Documentation Additions

Method handle_sighup() defined in zipkin is to handle signal SIGHUP, and this method is not called directly from the original SIGHUP signal handler but from other handlers, e.g., handle_sighup() of net_plugin, one of the mandatory plugins, can be used to forward signal SIGHUP by calling zipkin's handle_sighup().

Add a new option:
`telemetry-retry-interval-us`, optional parameter, specifies the retry interval for connecting to zipkin with default value set to 30000000


programs/rodeos/cloner_plugin.cpp
```diff
@@ -258,7 +260,8 @@ void cloner_plugin::plugin_initialize(const variables_map& options) {
       if (options.count("telemetry-url")) {
          fc::zipkin_config::init( options["telemetry-url"].as<std::string>(),
                                   options["telemetry-service-name"].as<std::string>(),
-                                  options["telemetry-timeout-us"].as<uint32_t>() );
+                                  options["telemetry-timeout-us"].as<uint32_t>(),
+                                  options["telemetry-retry-interval-us"].as<uint32_t>() );
       }
```
heifner, line 265: Add `fc::zipkin_config::handle_sighup();` to `cloner_plugin::handle_sighup()`

heifner (changes requested):

programs/rodeos/cloner_plugin.cpp
nickjjzhao: Added.

PR #10434, opened 2021-06-17
Always attempt to resync when receiving a block less than LIB.

jgiszczak:
## Change Description
Option to allow nodeos to continue syncing with a peer that has sent blocks older than LIB due to latency.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -249,7 +249,6 @@ namespace eosio {
       uint32_t                              max_nodes_per_host = 1;
       bool                                  p2p_accept_transactions = true;
       bool                                  p2p_reject_incomplete_blocks = true;
-      bool                                  p2p_always_resync = false;
 
       /// Peer clock may be no more than 1 second skewed from our clock, including network latency.
       const std::chrono::system_clock::duration peer_authentication_interval{std::chrono::seconds{1}};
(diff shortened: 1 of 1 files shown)


nksanthosh:
Continue testing for today and Kevin can review as soon as he is back.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2918,9 +2919,18 @@ namespace eosio {
...
+
             if( blk_num < last_sent_lib ) {
                fc_ilog( logger, "received block ${n} less than sent lib ${lib}", ("n", blk_num)("lib", last_sent_lib) );
-               close();
+               if(my_impl->p2p_always_resync) {
+                  fc_ilog( logger, "attempting to resync with peer that is sending blocks less than sent lib ${lib}", ("lib", last_sent_lib) );
+                  enqueue( (sync_request_message) {0, 0} );
+                  send_handshake();
+                  cancel_wait();
+               }
```
heifner, line 2930: I think we should not add this new option and just make this the behavior.

nksanthosh: Please put in your review comments and you can approve when it is done.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3971,6 +3981,7 @@ namespace eosio {
            "   _port  \tremote port number of peer\n\n"
            "   _lip   \tlocal IP address connected to peer\n\n"
            "   _lport \tlocal port number connected to peer\n\n")
+         ( "p2p-always-resync", bpo::value<bool>()->default_value(false), "Always attempt to resync even when peer sends blocks earlier than this node's last irreversible block")
```
heifner, line 3984: I think we should not add this new option and just make this the behavior.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
jgiszczak: Option removed.

heifner: PR name/description needs to be updated.

jgiszczak: Updated.

heifner (approved):

PR #10441, opened 2021-06-18
Feature privacy with blockvault test

shahabi8:
This PR creates a Python test for feature privacy with blockvault. There is a nodeos cluster with three nodes, two of them has access to blockvault. Security group is introduced to restrict non-participant node access to produced blocks as well as change of password for postgres database such that the non-participant nodes cannot retrieve data from blockvault.

## Change Type
**Select *ONE*:**
- [x] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions
It should be mentioned that security group does not restrict non-participant nodes access to blockvault. Postgres login credentials need to get updated to restrict non-participant nodes access.

### DIFF
--- a/tests/feature_privacy_blockvault.py
+++ b/tests/feature_privacy_blockvault.py
@@ -1,0 +1,142 @@
+#!/usr/bin/env python3
+
+from testUtils import Utils
+from Cluster import Cluster
+from WalletMgr import WalletMgr
+from TestHelper import TestHelper
+
+
+import time
+import shutil
+import signal
+import subprocess
+import atexit
+
+def killAndRelaunch(cluster, nodeToKill, addSwapFlags={}):
+    nodeId = nodeToKill.nodeId
+    Print("Kill node %d and then remove the state and blocks" % nodeId)
+
+    nodeToKill.kill(signal.SIGTERM)
+    shutil.rmtree(Utils.getNodeDataDir(nodeId,"state"))
+    shutil.rmtree(Utils.getNodeDataDir(nodeId,"blocks"))
+
+    assert nodeToKill.relaunch(timeout=30, skipGenesis=False, cachePopen=True, addSwapFlags=addSwapFlags), "Fail to relaunch"
+
+def changePostgressPassword(password):
+    stmt = "ALTER USER postgres WITH PASSWORD '{}';".format(password)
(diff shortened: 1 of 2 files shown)


tests/feature_privacy_blockvault.py
```diff
@@ -0,0 +1,142 @@
...
+testSuccessful=False
+killEosInstances=not dontKill
+killWallet=not dontKill
+
+WalletdName=Utils.EosWalletName
+ClientName="cleos"
+
+try:
+    TestHelper.printSystemInfo("BEGIN")
+    cluster.setWalletMgr(walletMgr)
```
dimas1185, line 60: you can save 2 lines and initialize wallet manager inside Cluster constructor

tests/feature_privacy_blockvault.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Print("#################################################################################")
+    Print("# Scenario 2: Adding node 0 and node 1 to a security group, check access of node 2 and node 0 to blocks produced by node 1#")
+    Print("#################################################################################")
+
+    Print("Add nodes 0 and 1 to security group")
+    securityGroup.editSecurityGroup(addNodes=[node0, node1])
+    securityGroup.verifySecurityGroup()
+
+    Print("### change postgre password to restrict node 2 access to it ###")
```
dimas1185, line 123: my understanding of blockvault feature is that it is used to create a cluster with same producer key with backup instances. Hence all of it's instances supposed to have same password and same certificate. So I don't see much logic in the scenario you are testing. Correct me if I missed something.

dimas1185 (changes requested):
two things:
1. please add more comments, test description, similar to the one in PR and ideally comment every step. this is essential for other people to maintain your test or solve flakyness if it will appear.
2. please review your test scenario, see my comment on line 123

shahabi8:
Closed this PR. Blockvault is not supposed to use this way.

PR #10445, opened 2021-06-21
static link libpq and submodule libpqxx cicd

softprofe:
## Change Description

Start 10445 description ......

EPE 886

This PR for cicd first. and after this merged. another PR take care of build script. 

In this PR, on all platforms,  libpq is linked staticaly.  and libpqxx is submoduled under eos/libraries.

And also script build  are supported on all platforms. 

End 10445 description ......
## Change Type
**Select ONE**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


spoonincode:
It's probably better to maintain the git commit history in our fork. It's a lot easier to verify the fork matches upstream (up to our changes) when all commits in our fork have the exact same commit id as upsteam.

That said, if we're content with forking libpqxx, then I do question if the alternative approach I put forward where we modify libpqxx to support back to libpq 9.2 is the better choice. With a modified libpqxx supporting libpq 9.2, there is no need to build and static link postgresql on any of our supported platforms. That avoids quite a bit of complexity in the build system.

.cicd/platforms/unpinned/macos-10.15-unpinned.sh
```diff
@@ -3,10 +3,31 @@ set -eo pipefail
 VERSION=1
 export SDKROOT="$(xcrun --sdk macosx --show-sdk-path)"
 brew update
-brew install git cmake python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl jq boost libpq libpqxx postgres || :
+brew install git cmake python libtool libusb graphviz automake wget gmp pkgconfig doxygen openssl jq   || :
 pip3 install requests
+
+# install boost from source
```
spoonincode, line 9: why was boost build strategy changed here?

spoonincode:
There seems to be some assumptions somewhere in the build that the environment is that of the cicd, and it's not properly detecting where various pieces are located.

My expectation is that, for example on macos 11, I can simply do
```
brew install cmake libusb gmp pkgconfig openssl boost postgresql
git clone ...
cmake
make
```
and it will work. But it fails with finding `libpq-fe.h`

This is another example of what becomes a lot easier if we simply dynamic link, as it seems to build as expected on my PoC branch.

.cicd/platforms/unpinned/macos-10.15-unpinned.sh
softprofe: Brew install boost get boost 1.76, it looks some files changed so that 1.76 can't allow eos build successfully.  I tried brew install boost@1.72, unfortunately brew can't find 172.  so, just build from source to solve that.

softprofe:
> It's probably better to maintain the git commit history in our fork. It's a lot easier to verify the fork matches upstream (up to our changes) when all commits in our fork have the exact same commit id as upsteam.
> 
> That said, if we're content with forking libpqxx, then I do question if the alternative approach I put forward where we modify libpqxx to support back to libpq 9.2 is the better choice. With a modified libpqxx supporting libpq 9.2, there is no need to build and static link postgresql on any of our supported platforms. That avoids quite a bit of complexity in the build system.

Which way to go need follow PM(s) decision.  Personally I think using libpq 9.2 too nostalgic as the current version is 13, old things will leave and new things come, I prefer to keep step with new things.

.cicd/platforms/unpinned/macos-10.15-unpinned.sh
spoonincode: The spirit of the unpinned pipeline is just that -- to catch problems a user would have when not doing a pinned build. If 1.76 is broken it needs to be fixed (not in this PR of course).

spoonincode (changes requested):
These changes break manual builds of eosio

softprofe:
> These changes break manual builds of eosio

The file need change is ensure libpq and libpqxx,  as this are too many commit in this PR, I just want make cicd work firstly and then another PR will make build script works.

spoonincode:
The build scripts aren't passing. Also I don't see any changes that would resolve manual builds either but I haven't actually tried since last time. Changing boost is not appropriate to this PR, that needs to be backed out and put up as a separate PR.

Packages should be changed to not include pq/pqxx as a dependency. Double check we're including all appropriate licenses because I didn't see anything for libpq and we may need it.

softprofe:
> The build scripts aren't passing. Also I don't see any changes that would resolve manual builds either but I haven't actually tried since last time. Changing boost is not appropriate to this PR, that needs to be backed out and put up as a separate PR.
> 
> Packages should be changed to not include pq/pqxx as a dependency. Double check we're including all appropriate licenses because I didn't see anything for libpq and we may need it.

After this PR, pq/pqxx are not required at nodeos runtime, but for the blockvalut test, if run test and set server to local machine(by default), pq is requried to install and started as a server.

pqxx has  Apache License, as other open source projects which had been included in eos, don't have issue.

scripts/helpers/eosio.sh
```diff
@@ -300,72 +300,48 @@ function build-clang() {
...
-        elif [ ! -f /usr/local/lib/pkgconfig/libpqxx.pc ]; then
-            brew install libpqxx
-        fi
-    fi
+        #build libpq from source
+        curl -L https://github.com/postgres/postgres/archive/refs/tags/REL_13_3.tar.gz | tar zxvf -
+        cd postgres-REL_13_3
+        ./configure && make && sudo make install && cd .. && rm -rf postgres-REL_13_3
 
+        sudo rm -rf /usr/local/var/postgres
```
spoonincode, line 310: This looks really dangerous. Could it eliminate a user's database?

spoonincode:
>pqxx has Apache License, as other open source projects which had been included in eos, don't have issue.

Anything we static link we need to be mindful of to include the appropriate license. For libpq from quick glance I take it that it's covered by this license
https://github.com/postgres/postgres/blob/master/COPYRIGHT
that certainly makes it looks like we need to include that file. Take a look in root CMakeLists.txt to see how it's done for others

spoonincode:
> After this PR, pq/pqxx are not required at nodeos runtime

Right, so it needs to be removed from generate_rpm, generate_deb, etc

spoonincode:
The changes have been made in such a way that only libpq in the cicd/script locations work. This prevents a user from doing a "manual" build of eosio. For example, on macos I would expect something like
```
brew install git cmake libusb gmp pkgconfig openssl boost postgres
git clone ...
cmake
make
```
to work, however the build will fail on this branch.

Similarly on Ubuntu, the libpq supplied in 20.04 is only 12.7 meaning the functionality is skipped. This seems needlessly restrictive. On Ubuntu 21.04 where libpq is >13.0, the build suffers a similar failure to macos when doing something like
```
apt-get install build-essential git libgmp-dev libboost-all-dev libssl-dev cmake pkg-config llvm-11-dev libusb-1.0-0-dev libcurl4-openssl-dev libpq-dev postgresql-server-dev-all
git clone ...
cmake -DCMAKE_PREFIX_PATH=/usr/lib/llvm-11
make
```

scripts/helpers/eosio.sh
spoonincode: imo for scripted builds doesn't seem worth trying to do anything with the database. Running the postgres tests outside of CI seems to require either
* user able to run docker, or
* manually setting PGHOST
So it's doesn't seem reasonable for this to just work as there is no way the script can safely make it just work.

softprofe: The reason we remove old path of the test data is to give new test a clean space.  
Remove this line, the first time install eos will OK. but if user install eos and uninstall it and then reinstall eos, there would be some unexpected table in db.  Can we take the eos host as an dedicated machine?

softprofe: OK, removed. so no issue now.

softprofe:
> > After this PR, pq/pqxx are not required at nodeos runtime
> 
> Right, so it needs to be removed from generate_rpm, generate_deb, etc

Ok, that should be removed. done.

softprofe:
> > pqxx has Apache License, as other open source projects which had been included in eos, don't have issue.
> 
> Anything we static link we need to be mindful of to include the appropriate license. For libpq from quick glance I take it that it's covered by this license
> https://github.com/postgres/postgres/blob/master/COPYRIGHT
> that certainly makes it looks like we need to include that file. Take a look in root CMakeLists.txt to see how it's done for others

https://github.com/EOSIO/libpqxx/blob/eosio/COPYING,  why this file not working, I imported all files under label 7.5.2.

softprofe:
> The changes have been made in such a way that only libpq in the cicd/script locations work. This prevents a user from doing a "manual" build of eosio. For example, on macos I would expect something like
> 
> ```
> brew install git cmake libusb gmp pkgconfig openssl boost postgres
> git clone ...
> cmake
> make
> ```
> 
> to work, however the build will fail on this branch.
> 
> Similarly on Ubuntu, the libpq supplied in 20.04 is only 12.7 meaning the functionality is skipped. This seems needlessly restrictive. On Ubuntu 21.04 where libpq is >13.0, the build suffers a similar failure to macos when doing something like
> 
> ```
> apt-get install build-essential git libgmp-dev libboost-all-dev libssl-dev cmake pkg-config llvm-11-dev libusb-1.0-0-dev libcurl4-openssl-dev libpq-dev postgresql-server-dev-all
> git clone ...
> cmake -DCMAKE_PREFIX_PATH=/usr/lib/llvm-11
> make
> ```

Per my knowledge, build by cmake && make is not always support, as we have build scripts we should use it.
EPE886 shows clearly we will build from source for libpq and use the version we build. we don't need care about what version the platform bring.

nksanthosh:
> The changes have been made in such a way that only libpq in the cicd/script locations work. This prevents a user from doing a "manual" build of eosio. For example, on macos I would expect something like
> 
> ```
> brew install git cmake libusb gmp pkgconfig openssl boost postgres
> git clone ...
> cmake
> make
> ```
> 
> to work, however the build will fail on this branch.
> 
> Similarly on Ubuntu, the libpq supplied in 20.04 is only 12.7 meaning the functionality is skipped. This seems needlessly restrictive. On Ubuntu 21.04 where libpq is >13.0, the build suffers a similar failure to macos when doing something like
> 
> ```
> apt-get install build-essential git libgmp-dev libboost-all-dev libssl-dev cmake pkg-config llvm-11-dev libusb-1.0-0-dev libcurl4-openssl-dev libpq-dev postgresql-server-dev-all
> git clone ...
> cmake -DCMAKE_PREFIX_PATH=/usr/lib/llvm-11
> make
> ```

Just noting that the build process documented for our customers relies on using the build script or performing a build using the instructions in our CICD scripts.

Manual build of [dependencies](https://developers.eos.io/manuals/eos/latest/install/build-from-source/manual-build/platforms/ubuntu-18.04/#install-eosio-dependencies). These are similar to the instructions in [CICD builds](https://github.com/EOSIO/eos/blob/develop/.cicd/platforms/unpinned/ubuntu-18.04-unpinned.dockerfile)

Build [script](https://developers.eos.io/manuals/eos/latest/install/build-from-source/shell-scripts/index)

spoonincode:
EOSIO is open source software and we should ensure the builds continue to operate properly via standard build practices: install dependencies, cmake, make.

Requiring users to build and place dependencies at a specific location, and then breaking the build when the user fails to do so, is no good. AFAIK the build scripts are not considered the canonical build instructions, and hopefully it stays that way: requiring users to run or understand a build script that borderline obfuscates the actions it is taking is also not the way to ship an open source friendly piece of software.

Consider too, that valid build environments from 2.1 will fail to build with these changes. Not just miss out on blockvault, but fail. As is, this PR breaks just as much as it fixes.

If it's not possible to change this PR to fix these issues then you should be looking at #10462 which does properly handle all this I believe.

kj4ezj:
> Per my knowledge, build by cmake && make is not always support, as we have build scripts we should use it.

No, quite the opposite. Building using `cmake` and `make` with appropriate dependencies installed is primary (our CI follows this pattern, for example), and the build scripts are secondary. I am not even sure who the build scripts are for, since they are too obtuse and complex for someone who doesn't know how to build from source to learn from IMHO, and we offer binaries and docker containers that novices can use without all that work. The build scripts have a disproportionately high support burden relative to our other pieces of code, and I'd speculate that they consume a disproportionately high amount of developer time relative to the number of people who benefit from them. We have discussed removing them altogether many, many times, and I hope we do.  
  
> AFAIK the build scripts are not considered the canonical build instructions, and hopefully it stays that way

This.

I don't meant to dog-pile @softprofe, but I did want to share a different perspective. IIRC, the CI system did use the build scripts at one time and we stopped doing that because they were always breaking. They have gotten better over time, but `cmake` and `make` is the gold standard.

kj4ezj:
> Just noting that the build process documented for our customers relies on using the build script or performing a build using the instructions in our CICD scripts.

I agree @nksanthosh, [these commands](https://developers.eos.io/manuals/eos/latest/install/build-from-source/manual-build/platforms/ubuntu-18.04/#install-eosio-dependencies) look like they were lifted straight from our dockerfiles. Dockerfile syntax is **_not_** human friendly. We should be providing a human-friendly process, and we should explain what we are doing in each step so community members can learn and contribute. I imagine human-friendly instructions might look like steps 2, 5, and 6 under [Core Tools](https://github.com/EOSIO/devdocs/wiki/Tool-Install-Guide#core-tools) in our Tool Install Guide. I am biased because I wrote it, and it is probably out-of-date at this point, but it explains what you are installing and at least makes some attempt to explain why. I copied those steps below because it is not publicly accessible:

<details>
<summary>Example</summary>

> 1. [openssl](https://www.openssl.org/) — open-source cryptography library
>    ```bash
>    $ brew install openssl && ln -s /usr/local/opt/openssl/include/openssl /usr/local/include
>    ```
>    - The symbolic link helps compilers and IDEs find your include files
> 1. Install EOSIO dependencies
>    ```bash
>    $ brew install automake boost cmake doxygen gmp graphviz jq libtool libusb llvm@7 ninja pkgconfig python wget
>    ```
>    - [automake](https://en.wikipedia.org/wiki/Automake) — a tool for automatically generating `Makefile.in` files
>    - [boost](https://www.boost.org/) — collection of portable, open-source C++ libraries
>    - [cmake](https://en.wikipedia.org/wiki/CMake) — manages the build process using a compiler-independent method, including testing and packaging software
>    - [doxygen](https://en.wikipedia.org/wiki/Doxygen) — tool for deriving software reference documentation from source code
>    - [gmp](https://en.wikipedia.org/wiki/GNU_Multiple_Precision_Arithmetic_Library) — library for arbitrary-precision arithmetic on signed integers, rationals, and floating points
>    - [graphviz](https://en.wikipedia.org/wiki/Graphviz) — graph visualization software using the [DOT](https://en.wikipedia.org/wiki/DOT_(graph_description_language)) graph description language
>    - [jq](https://stedolan.github.io/jq/) — command-line JSON processor
>    - [libtool](https://en.wikipedia.org/wiki/GNU_Libtool) — a consistent, portable interface for shared libraries
>    - [libusb](https://en.wikipedia.org/wiki/Libusb) — platform-agnostic access to USB devices
>    - [llvm](https://en.wikipedia.org/wiki/LLVM) — collection of modular, reusable, and platform-independent compiler and toolchain technologies
>    - [ninja](https://ninja-build.org/) — fast, multithreaded alternative to `make`
>    - [pkgconfig](https://en.wikipedia.org/wiki/Pkg-config) — determines library locations on a system at compile-time or install-time, so they don't need to be hard-coded
>    - [python](https://en.wikipedia.org/wiki/Python_%28programming_language%29) — interpreted, high-level, general-purpose programming language
>    - [wget](https://en.wikipedia.org/wiki/Wget) — command-line utility to download content from web servers
> 1. [EOSIO](https://eos.io/) — advanced smart contracts blockchain platform
>     1. Make an empty build directory
>        ```bash
>        $ [[ -d ~/Work/eos-legacy/build ]] && rm -rf ~/Work/eos-legacy/build; mkdir ~/Work/eos-legacy/build
>        ```
>     2. Navigate there
>        ```bash
>        $ cd ~/Work/eos-legacy/build
>        ```
>     3. Checkout the latest release   <-- don't do this, checkout develop, hopefully these commands will work for develop, if they don't, get a teammates help and update these so they are useful
>        ```bash
>        $ git checkout master
>        ```
>     4. Update submodules
>        ```bash
>        $ git submodule update --init --recursive
>        ```
>     5. Build EOSIO on all your cores, except two
>        ```bash
>        $ cmake .. && make -j $(( $(getconf _NPROCESSORS_ONLN) - 2 ))
>        ```
>        - A full build of CDT requires EOSIO to have been built and installed from source, rather than using the binaries
>     6. Install EOSIO
>        ```bash
>        $ make install
>        ```

</details>

spoonincode:
eosio-build-scripts pipeline is failing on mac build. This pipeline needs to be run manually from buildkite.

softprofe:
> eosio-build-scripts pipeline is failing on mac build. This pipeline needs to be run manually from buildkite.

Paste the link pls. it shows all checks have passed. I can't see which pipeline was failed.

kj4ezj:
Link: eosio-build-scripts [build 1968](https://buildkite.com/EOSIO/eosio-build-scripts/builds/1968)

I kicked also kicked off eosio-base-images [build 1746](https://buildkite.com/EOSIO/eosio-base-images/builds/1746).

These would both need to pass in order for this pull request to be merged.

softprofe:
> > Per my knowledge, build by cmake && make is not always support, as we have build scripts we should use it.
> 
> No, quite the opposite. Building using `cmake` and `make` with appropriate dependencies installed is primary (our CI follows this pattern, for example), and the build scripts are secondary. I am not even sure who the build scripts are for, since they are too obtuse and complex for someone who doesn't know how to build from source to learn from IMHO, and we offer binaries and docker containers that novices can use without all that work. The build scripts have a disproportionately high support burden relative to our other pieces of code, and I'd speculate that they consume a disproportionately high amount of developer time relative to the number of people who benefit from them. We have discussed removing them altogether many, many times, and I hope we do.
> 
> > AFAIK the build scripts are not considered the canonical build instructions, and hopefully it stays that way
> 
> This.
> 
> I don't meant to dog-pile @softprofe, but I did want to share a different perspective. IIRC, the CI system did use the build scripts at one time and we stopped doing that because they were always breaking. They have gotten better over time, but `cmake` and `make` is the gold standard.

If you did all things like dockerfile did, and then cmake and make should work, I agree with that.
However,  only brew/apt/yum install  a list of things is not enough, as we already build some dependencies from source.
I don't agree with so called "right way" or "best way", any way have  advantages and disadvantages.
Build from source and linked statically is good and using a fixed version and put in fixed path. easy to control, easy to test, save human resource.
 Using brew/apt/yum to install, have to deal with different version, different install path on different OS, some unexpected things happens you need used engineer to do fix. it costs company resources.  and you have to use low version on some platform, and you need take the risk of version mismatch. and you can do the cmake and make.

This is not only a opensource software, it is also a business. if sacrifice business benefit to follow opensource standard, is it really smart?

softprofe:
> Link: eosio-build-scripts [build 1968](https://buildkite.com/EOSIO/eosio-build-scripts/builds/1968)
> 
> I also kicked off eosio-base-images [build 1746](https://buildkite.com/EOSIO/eosio-base-images/builds/1746).
> 
> These would both need to pass in order for this pull request to be merged.

both passed.

 eosio-build-scripts:    https://buildkite.com/EOSIO/eosio-build-scripts/builds/1974#

spoonincode:
Making builds hard, obfuscated, and/or non-standard reduces users' confidence in the project. Builds failing with the simple cmake and make described above is just that.

Early versions of EOSIO required an _exact_ old version of LLVM to compile, and required the user to chase down a rather obscure fork of libsecp256k1 to compile. Those sort of requirements were clearly friction to users, and it is generally not the kind of build environment one would expect from a top shelf open source application. The team has poured a lot of effort in to improving the build experience and I see no reason to regress even an inch for something that demonstrably needs not to.

PR #10458, opened 2021-06-25
resolve flakiness in privacy scenario 3 test

dimas1185:
## Change Description

this resolve flakiness found in that fail:
https://buildkite.com/EOSIO/eosio/builds/28938#2d63152c-5ad8-4724-851e-13b34d09c21a

flakiness was caused by very last step when we are checking that `node2` is not advancing head block after removing it from security group.
the issue is that we recorded `node1` head too early and when we compare heads they are same.
this fix adds proper timer.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/privacy_scenario_3_test.py
+++ b/tests/privacy_scenario_3_test.py
@@ -266,8 +266,8 @@ try:
     Print("API Node1 head = {} API Node2 head = {}".format(apiNode2_head, apiNode1.getHeadBlockNum()))
     assert apiNode2_Advancing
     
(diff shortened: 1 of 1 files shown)


kj4ezj (approved):
🕕

PR #10462, opened 2021-06-28
simplify build & fix binaries via a forked & modified libpqxx compatible with libpq 9.2

spoonincode:
## Change Description

This PR contains an alternate solution to fix the libpq & libpqxx dependency troubles that have plagued the 2.1 release.

This approach uses a forked and modified version of libpqxx that is submoduled in to the build which will compile with libpq 9.2 (9.2 is the version included out of the box in our older supported distros like Centos7 & Amazon Linux 2). The primary benefit of this approach over static linking libpq is that by simply installing a package or two provided by the base repos of all supported distros there is no need to add anything w.r.t. libpq & libpqxx into our build processes. Less is more when it comes to the build. Plus, all build types continue to “just work”: pinned builds, unpinned builds, and “manual” builds (i.e. it’s simple to just install the packages and build manually with cmake & make).

There is another slight minor benefit in that the pkg-config dependency is removed from this part of the build. This clears the path to allow complete removal of that dependency once #9075 is mergable.

The major downside is that we have a libpqxx fork. However #10445 was already doing that, so if we’re content with forking this approach would seem to comparatively reduce complexity. Due to our rather light usage of libpqxx and the relatively minor changes needed for it to compile with 9.2, forking doesn’t seem all too bad to me. Changes can be found at EOSIO/libpqxx@827adfbd; it involves disablement of some libpqxx functions we are not using. It would be interesting to explore this further and see if the changes could be cleaned up in a manner that could be sent upstream.

Another annoyance I have is that libpqxx (properly) uses `find_package(PostgreSQL)` to find postgresql libraries. But this search requires header files that aren’t always in the “base” libpq dev packages. For Ubuntu we need to install another heavier package `postgresql-server-dev-all` to satisfy this lookup. This is unfortunate, but it is only a build dep not a runtime dep.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


kj4ezj (commented):
This looks good to me. Approval pending passing eosio-build-scripts [build 2005](https://buildkite.com/EOSIO/eosio-build-scripts/builds/2005).

I also kicked off eosio-base-images [build 1782](https://buildkite.com/EOSIO/eosio-base-images/builds/1782) to be sure, but we probably don't need to wait on that to merge because the base images are built implicitly by the [eosio](https://buildkite.com/EOSIO/eosio) pipeline.

Finally, it looks like your editor is stripping newlines from the end of files which breaks compatibility with some POSIX tools. Normally this doesn't really matter, but I think we should avoid that in public repos.

kj4ezj (approved):
The test builds passed. Fantastic work, @spoonincode! I am excited about this.

See Also: [AUTO-1002](https://blockone.atlassian.net/browse/AUTO-1002) and [EPE-886](https://blockone.atlassian.net/browse/EPE-886).

PR #10478, opened 2021-07-03
flakiness fix for eosio_blocklog_prune_test

dimas1185:
## Change Description

This is fix to flakiness of `eosio_blocklog_prune_test`.
Test could fail if following occurs:
full validation node receives few blocks including unlinkable block.
`connection::handle_message( const block_id_type& id, signed_block_ptr ptr )` first checks if it is complete block and then posts block processing to application thread.
Sometimes messages queued faster than they are processed and once we get unlinkable block we close entire connection. After doing this `connection::process_signed_block` drops processing entire queue even if those are valid blocks.
Fix:
Do not restrict test to check full validation node to have exactly unlinkable-1 block number. It can be earlier block.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/eosio_blocklog_prune_test.py
+++ b/tests/eosio_blocklog_prune_test.py
@@ -175,8 +175,10 @@ try:
 
     assert headAdvanced or lvnPostInfo["head_block_num"] >= cfTrxBlockNum, "the light validation node stops syncing"
 
-    #lib will be cfTrxBlockNum-2 and head cfTrxBlockNum-1 because of when we will receive incomplete block we won't advance lib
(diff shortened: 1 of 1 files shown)


heifner (changes requested):
I don’t see any reason to process valid blocks from a connection that sends you an incomplete block and you can’t process it. The real fix is to modify the test to handle this case.

heifner (approved):

PR #10484, opened 2021-07-08
Restore CentOS 8 support.

jgiszczak:
## Change Description
Restore CentOS 8 Support.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/platforms/pinned/centos-8-pinned.dockerfile
+++ b/.cicd/platforms/pinned/centos-8-pinned.dockerfile
@@ -1,0 +1,57 @@
+FROM centos:8
+ENV VERSION 1
+#install dependencies
+RUN yum update -y && \
+    yum install -y epel-release  && \
+    yum --enablerepo=extras install -y which git autoconf automake libtool make bzip2 && \
+    yum --enablerepo=extras install -y  graphviz bzip2-devel openssl-devel gmp-devel  && \
+    yum --enablerepo=extras install -y  file libusbx-devel && \
+    yum --enablerepo=extras install -y libcurl-devel patch vim-common jq && \
+    yum install -y python3
+RUN dnf install -y  https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
+    dnf group install -y  "Development Tools" && \
+    dnf --enablerepo=PowerTools install -y doxygen ocaml
+# cmake3.18.0
+RUN curl -LO https://github.com/Kitware/CMake/releases/download/v3.18.0/cmake-3.18.0.tar.gz && \
+    tar -xzf cmake-3.18.0.tar.gz && \
+    cd cmake-3.18.0 && \
+    ./bootstrap --prefix=/usr/local && \
+    make -j$(nproc)  && make install && \
+    rm -rf cmake-3.18.0.tar.gz cmake-3.18.2
+# clang10.0.0
+RUN git clone --single-branch --branch llvmorg-10.0.0 https://github.com/llvm/llvm-project clang10 && \
+    mkdir /clang10/build && cd /clang10/build && \
+    cmake -G 'Unix Makefiles' -DCMAKE_INSTALL_PREFIX='/usr/local' -DLLVM_ENABLE_PROJECTS='lld;polly;clang;clang-tools-extra;libcxx;libcxxabi;libunwind;compiler-rt' -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_ENABLE_RTTI=ON -DLLVM_INCLUDE_DOCS=OFF -DLLVM_TARGETS_TO_BUILD=host -DCMAKE_BUILD_TYPE=Release ../llvm   && \
+    make -j$(nproc) && make install && \
+    cd / && \
+    rm -rf /clang10
+COPY ./.cicd/helpers/clang.make /tmp/clang.cmake
+#build llvm10
+RUN git clone --depth 1 --single-branch --branch llvmorg-10.0.0 https://github.com/llvm/llvm-project llvm && \
+    cd llvm/llvm && \
+    mkdir build && \
+    cd build && \
+    cmake -G 'Unix Makefiles' -DLLVM_TARGETS_TO_BUILD=host -DLLVM_BUILD_TOOLS=false -DLLVM_ENABLE_RTTI=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_TOOLCHAIN_FILE='/tmp/clang.cmake' -DCMAKE_EXE_LINKER_FLAGS=-pthread -DCMAKE_SHARED_LINKER_FLAGS=-pthread -DLLVM_ENABLE_PIC=NO -DLLVM_ENABLE_TERMINFO=OFF .. && \
+    make -j$(nproc)  && \
+    make install && \
+    cd / && \
+    rm -rf /llvm
+# build boost
+RUN curl -LO https://dl.bintray.com/boostorg/release/1.72.0/source/boost_1_72_0.tar.bz2 && \
+    tar -xjf boost_1_72_0.tar.bz2 && \
+    cd boost_1_72_0 && \
+    ./bootstrap.sh --with-toolset=clang --prefix=/usr/local && \
+    ./b2 toolset=clang cxxflags='-stdlib=libc++ -D__STRICT_ANSI__ -nostdinc++ -I/usr/local/include/c++/v1 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fpie' linkflags='-stdlib=libc++ -pie' link=static threading=multi --with-iostreams --with-date_time --with-filesystem --with-system --with-program_options --with-chrono --with-test -q -j$(nproc) install && \
+    cd / && \
+    rm -rf boost_1_72_0.tar.bz2 /boost_1_72_0
+# install nvm
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.0/install.sh | bash
+# load nvm in non-interactive shells
+RUN cp ~/.bashrc ~/.bashrc.bak && \
+    cat ~/.bashrc.bak | tail -3 > ~/.bashrc && \
+    cat ~/.bashrc.bak | head -n '-3' >> ~/.bashrc && \
+    rm ~/.bashrc.bak
+# install node 10
+RUN bash -c '. ~/.bashrc; nvm install --lts=dubnium' && \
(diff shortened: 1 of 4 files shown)


jgiszczak:
`fc` links to `boost::iostreams` and requires it be built with zlib support.  `boost::iostreams` itself requires the compiled library version of `boost::regex` even though `boost::regex` can be used header-only if not building with C++03.

Either `boost-build` is defective or Boost's own use of `boost-build` is defective.  When the available dependencies of a library change, it rebuilds affected `.so` files but does NOT rebuild the corresponding `.a` files.

jgiszczak:
The boost build should be removed from the bash scripts and pushed into the CMake scripts.  Nearly all Linux distributions provide only g++ builds of packaged C++ libraries.  In consequence, when building with clang, we need to build our own version of boost.  We now always build boost when building with clang.  It would be preferable to allow CMake's boost locator to determine whether or not the available boost is compatible and only build it if it isn't.

praphael:
> The boost build should be removed from the bash scripts and pushed into the CMake scripts. Nearly all Linux distributions provide only g++ builds of packaged C++ libraries. In consequence, when building with clang, we need to build our own version of boost. We now always build boost when building with clang. It would be preferable to allow CMake's boost locator to determine whether or not the available boost is compatible and only build it if it isn't.

The way we use Boost is unfortunately quite complicated, and the argument for relying on a local pre-buiilt installation to avoid users having to d/l it to build gets less and less compelling.    The cleanest way would be to include it in the source tree and always compile it.  It doesn't add much to the build time, but it would add ~160M to the source tree if we only include source, currently we are at ~270 M.

spoonincode:
It's nothing to do with gcc vs clang but rather libstdc++ vs libc++. The proper solution is to eliminate the need for pinned builds altogether (and thus the need for any building of c++ dependencies with libc++). Modern LTS distros (Debian 10, Ubuntu 20.04, Centos 7&8 via EPEL) provide new enough boost versions that simply just work.

praphael:
@spoonincode  It just works until something changes and it breaks again.  Successful long lasting open source projects tend to have minimal dependencies outside of compiler / OS.   Obviously ones being Boost and CMake themselves.   Chrome is another one with a gargantuan code base.  To get it to run on the large number of systems it does, they don't rely on external dependencies being built properly - they just include them and build them themselves.   Personally as.a end user I'd much rather have a slightly longer download with a slightly longer build time and have my build just work, rather than having the build fail with linker error messages because my local installs of whatever library was incompatible and the build system isn't smart enough to figure that out, because the dev was trying to shave 30 seconds off the build time.   When that happens with a package I don't absolutely need for something, I tend to nope out.   Of course we could add complexity to the build process, detection code, as Johnathan suggested but that imposes and maintenance cost, and again is another thing that has the potential to break.

spoonincode:
We can't include an entire Linux distribution in to the build. The line has to be drawn somewhere. The current heuristic to bring a dependency inside the source tree is somewhere along the lines of:
* is unusual and not typically provided in a distro's package manager (secp256k1, amqp-cpp, libyubihsm, pqxx)
* we require (or it's super highly desirable for) the dependency to be pinned to an explicit revision (arguably rocksdb, definitely rapidjson)
* we need to customize a dependency (wabt historically, softfloat, recently pqxx)

None of these are true for boost, llvm, openssl, gmp, zlib, etc. 

While Chromium may bundle many typically-system-provided libraries as part of its build, it provides a mechanism to turn those off and use the actual system provided libraries because most open source projects consider it an anti-pattern otherwise. Indeed, the packagers for [Arch](https://archlinux.org/packages/extra/x86_64/chromium/), [Debian](https://packages.debian.org/buster/chromium), [Fedora](https://fedora.pkgs.org/34/fedora-updates-x86_64/chromium-91.0.4472.114-1.fc34.x86_64.rpm.html) all build Chromium with many system libraries (non-bundled versions). This is open source way.

Telling users "eosio requires boost 1.67+" is no harder for the user than telling them "you need a c++17 compiler to build eosio" when the user is running Ubuntu 20.04+, Centos 7, Centos 8, Debian 10, homebrew, Arch, Fedora, etc etc

praphael:
It is now passing all checks.

praphael:
Build scripts now work for all systems.   All tests passing for both pinned and unpinned builds.

.cicd/platforms/pinned/centos-8-pinned.dockerfile
```diff
@@ -0,0 +1,63 @@
...
+    cd llvm/llvm && \
+    mkdir build && \
+    cd build && \
+    cmake -G 'Unix Makefiles' -DLLVM_TARGETS_TO_BUILD=host -DLLVM_BUILD_TOOLS=false -DLLVM_ENABLE_RTTI=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_TOOLCHAIN_FILE='/tmp/clang.cmake' -DCMAKE_EXE_LINKER_FLAGS=-pthread -DCMAKE_SHARED_LINKER_FLAGS=-pthread -DLLVM_ENABLE_PIC=NO -DLLVM_ENABLE_TERMINFO=OFF .. && \
+    make -j$(nproc)  && \
+    make install && \
+    cd / && \
+    rm -rf /llvm
+# build boost
+RUN curl -LO https://boostorg.jfrog.io/artifactory/main/release/1.69.0/source/boost_1_69_0.tar.bz2 && \
```
spoonincode, line 42: all pinned builds should use the same version of boost (1.72)

scripts/eosio_build_centos8_deps
```diff
@@ -0,0 +1,28 @@
...
+gettext-devel,rpm -qa
+file,rpm -qa
+libusbx-devel,rpm -qa
+libcurl-devel,rpm -qa
+patch,rpm -qa
+llvm-devel,rpm -qa
+llvm-static,rpm -qa
+epel-release,rpm -qa
+jq,rpm -qa
+boost169-devel,rpm -qa
```
spoonincode, line 25: I don't see any evidence of this boost package being used so it can probably be removed

spoonincode (dismissed):
pinned needs to be 1.72

spoonincode (changes requested):
Doesn't look like a centos8 package is getting created

praphael:
@spoonincode Centos8 package being generated https://buildkite.com/EOSIO/eosio/builds/29338

spoonincode (approved):

.cicd/platforms/pinned/centos-8-pinned.dockerfile
```diff
@@ -0,0 +1,63 @@
...
+#install dependencies
+RUN yum update -y && \
+    yum install -y epel-release  && \
+    yum --enablerepo=extras install -y which git autoconf automake libtool make bzip2 && \
+    yum --enablerepo=extras install -y  graphviz bzip2-devel openssl-devel gmp-devel  && \
+    yum --enablerepo=extras install -y  file libusbx-devel && \
+    yum --enablerepo=extras install -y libcurl-devel patch vim-common jq && \
+    yum install -y python3 python3-devel python3-requests && \
+    yum install -y glibc-locale-source glibc-langpack-en postgresql-server postgresql-devel
+RUN dnf install -y  https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm && \
```
spoonincode, line 12: kinda strange to install this here, while a handful of lines up `epel-release` is being installed anyways. wouldn't that be the same?

.cicd/build.sh
```diff
@@ -34,7 +34,10 @@ else # Linux
         CMAKE_EXTRAS="$CMAKE_EXTRAS -DLLVM_DIR=\"/opt/rh/llvm-toolset-7.0/root/usr/lib64/cmake/llvm\""
     elif [[ "$IMAGE_TAG" == 'ubuntu-18.04-unpinned' ]]; then
         CMAKE_EXTRAS="$CMAKE_EXTRAS -DCMAKE_CXX_COMPILER=\"clang++-7\" -DCMAKE_C_COMPILER=\"clang-7\" -DLLVM_DIR=\"/usr/lib/llvm-7/lib/cmake/llvm\""
+    elif [[ "$IMAGE_TAG" == 'centos-8-pinned' ]]; then
+        CMAKE_EXTRAS="$CMAKE_EXTRAS -DPostgreSQL_TYPE_INCLUDE_DIR=/usr/include/pgsql"
```
spoonincode, line 38: I believe this is needed because the `postgresql-server-devel` package isn't being installed. To reduce clutter here you could consider installing that package that instead, but it's not a blocker by any means

.cicd/platforms/pinned/centos-8-pinned.dockerfile
praphael: I removed redudant package

.cicd/build.sh
praphael: removed this part and added `postgresql-server-devel` package

praphael:
Build scripts aren't passing now for some odd reason ... will have to fix this before its merged

praphael:
I did a git rebase to fix the shady commits issue

spoonincode:
I don't see a rebase; the "root" commits still in the history of the branch

praphael:
> I don't see a rebase; the "root" commits still in the history of the branch

I'm not sure when you do a merge its possible to pick up the rebase properly.  Now it still has the root as the author but myself as the committer ... but now it also says that I made a bunch of other commits that were authored by others so maybe I just made it worse ... 

The only other solution I can think of would be copypasta'ing the changes at HEAD to a new branch that was reset to the commit before the root commits, and just having one commit after that.  However there were several commits after by @jgiszczak  after that I wanted to check with him first.    But if he wants retain authorship/credit over the PR at least, he can just crib my changes as a single commit since I was the one that screwed things up in his PR :-)

spoonincode:
General policy is no force pushes once a PR is created (ergo no rebases). Force pushes make it harder for a reviewer as it means they need to go back and review everything again.

But in this case since the history is pretty rekt maybe it's a reasonable exception and it'd be okay to squash it all down or something. @kj4ezj may have some useful input on best/appropriate/allowed course of action here.

praphael:
Centos 8 support is being dropped in favor of Oracle Linux 8 - please see https://github.com/EOSIO/eos/pull/10724

PR #10508, opened 2021-07-14
Added force-write-stride option in rodeos to control manual rocksdb flush frequency

nksanthosh:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


programs/rodeos/cloner_plugin.cpp
```diff
@@ -291,6 +293,9 @@ void cloner_plugin::set_program_options(options_description& cli, options_descri
    op("filter-name", bpo::value<std::string>(), "Filter name");
    op("filter-wasm", bpo::value<std::string>(), "Filter wasm");
    op("profile-filter", bpo::bool_switch(), "Enable filter profiling");
+   op("force-write-stride", bpo::value<uint32_t>()->default_value(200), 
+      "Maximum number of blocks to process before forcing rocksdb to flush. This option is primarily useful to control re-sync durations under disaster recovery scenarios (when rodeos has unexpectedly exited). This option ensures blocks stored in rocksdb are at most force-write-stride blocks behind the current head block being processed by rodeos. However, saving too frequently may affect performance. It is likely that rocksdb itself will save rodeos data more frequently than this setting by flushing memtables to disk, based on various rocksdb options. In contrast, when rodeos exits normally, it saves the last block processed by rodeos into rocksdb and will continue processing new blocks from that last processed block number when it next starts up.");
```
heifner, line 297: Can we break this up over multiple lines.

libraries/rodeos/include/b1/rodeos/rodeos.hpp
```diff
@@ -50,6 +50,7 @@ struct rodeos_db_snapshot {
    eosio::checksum256                      irreversible_id = {};
    uint32_t                                first           = 0;
    std::optional<uint32_t>                 writing_block   = {};
+   uint32_t                                force_write_stride = 200;
```
heifner, line 53: Since this is always set, I think it should be initialized to 0.

programs/rodeos/cloner_plugin.cpp
```diff
@@ -42,6 +42,7 @@ struct cloner_config : ship_client::connection_config {
    eosio::name filter_name = {}; // todo: remove
    std::string filter_wasm = {}; // todo: remove
    bool        profile = false;
+   uint32_t    force_write_stride = 200;
```
heifner, line 45: Since this is always set this should be initialized to 0. Should have only one place where the default is set.

heifner (changes requested):

nksanthosh:
Just trying to get some conflicts resolved. Not yet ready for review

libraries/rodeos/include/b1/rodeos/rodeos.hpp
nksanthosh: Changed to zero.

programs/rodeos/cloner_plugin.cpp
nksanthosh: Changed to zero.

programs/rodeos/cloner_plugin.cpp
nksanthosh: Also done.

nksanthosh:
Ready for review.

programs/rodeos/cloner_plugin.cpp
```diff
@@ -291,6 +295,14 @@ void cloner_plugin::set_program_options(options_description& cli, options_descri
    op("filter-name", bpo::value<std::string>(), "Filter name");
    op("filter-wasm", bpo::value<std::string>(), "Filter wasm");
    op("profile-filter", bpo::bool_switch(), "Enable filter profiling");
+   op("force-write-stride", bpo::value<uint32_t>()->default_value(200), 
+      "Maximum number of blocks to process before forcing rocksdb to flush. This option is primarily useful to control re-sync durations \
+      under disaster recovery scenarios (when rodeos has unexpectedly exited, the option ensures blocks stored in rocksdb are at most \
+      force-write-stride blocks behind the current head block being processed by rodeos. However, saving too frequently may affect performance.\
+      It is likely that rocksdb itself will save rodeos data more frequently than this setting by flushing memtables to disk, based on various rocksdb \
+      options. In contrast, when rodeos exits normally, it saves the last block processed by rodeos into rocksdb and will continue processing \
+      new blocks from that last processed block number when it next starts up.");
```
heifner, line 304: The normal way to break long strings is to use `""` not `\`.
https://stackoverflow.com/questions/3859157/splitting-c-strings-onto-multiple-lines-code-syntax-not-parsing

heifner (approved):

PR #10513, opened 2021-07-14
Ship threaded send

huangminghuang:
## Change Description

This PR 
  - creates additional threads to send data to ship clients.
  - fixes a bug where state history plugin would send blocks with block_num smaller than the client requests when the state history block head is behind the block head of the client.
  - stops the application when it fails to listen on the configured state history port. 

**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -110,12 +160,23 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
+                  self->socket_stream->binary(false);
+                  self->socket_stream->async_write(
+                     boost::asio::buffer(state_history_plugin_abi, strlen(state_history_plugin_abi)),
+                     [self](boost::system::error_code ec, size_t) {
+                        if (ec) {
+                           self->on_fail(ec, "async_write");
+                           return;
+                        }
+                        self->socket_stream->binary(true);
+                        self->start_read();
```
heifner, line 175: I'm pretty sure this is not thread safe. You are not allowed to access a socket from more than one thread at a time. Here you have the main thread doing an `async_read` and the internal thread doing an `write`.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -95,12 +96,61 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
+            }
+         }
+         catch (...) {
+            fc_elog(_log, "exception thrown from send thread");
+            eptr = std::current_exception();
+            send_thread_has_exception = true;
+         }
+
+         // avoid ~session() to be executed in this thread.
+         app().post( priority::high, [self = this->derived_session().shared_from_this()]() {
```
heifner, line 134: Does this need to be `high` priority? If you are just pushing it to a diff thread, then `low` should be fine, right?

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -128,44 +189,28 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-             boost::asio::buffer(send_queue[0]),
-             [self = derived_session().shared_from_this()](boost::system::error_code ec, size_t) {
-                self->callback(ec, "async_write", [self] {
-                   self->send_queue.erase(self->send_queue.begin());
-                   self->sending = false;
-                   self->send();
-                });
-             });
+      void send(T obj, ::std::optional<::fc::zipkin_span>&& span) {
+         if (!send_thread_has_exception) {
```
heifner, line 200: Seems like we should at least log that we are ignoring this send.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -237,108 +288,110 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-               }
-               if (block_req.fetch_deltas && plugin->chain_state_log) {
-                  result.deltas = plugin->chain_state_log->get_log_entry(block_num);
-               }
-               set_result_block_header(result, get_block());
             }
-            ++block_num;
+            if (block_req.fetch_traces && plugin->trace_log) {
+               result.traces = plugin->trace_log->get_log_entry(to_send_block_num);
+               // fc_add_tag(send_update_span, "traces_size", result.traces.data_size());
```
heifner, line 347: Any reason not to include this?

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -237,108 +288,110 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
+            }
+            if (block_req.fetch_deltas && plugin->chain_state_log) {
+               result.deltas = plugin->chain_state_log->get_log_entry(to_send_block_num);
+               // fc_add_tag(send_update_span, "deltas_size", result.deltas.data_size());
```
heifner, line 351: Any reason not to include this?

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -502,44 +557,38 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-      // Both app().quit() and exception throwing are required. Without app().quit(),
-      // the exception would be caught and drop before reaching main(). The exception is
-      // to ensure the block won't be committed.
-      appbase::app().quit();
-      EOS_THROW(
-          chain::state_history_write_exception,
-          "State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch "
+      elog("State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch "
           "the process");
+      exit(1);
```
heifner, line 576: This will cause a dirty db. Seems like we need a way to shutdown without a dirty db.

heifner (changes requested):

plugins/state_history_plugin/state_history_plugin.cpp
huangminghuang: That's not what I understand about socket programming. According to https://stackoverflow.com/questions/7418093/use-one-socket-in-two-threads, it should be safe for TCP. AFAIK, websocket is full-duplex communication channels over a single TCP connection, it should also be safe for read/write from different threads.

plugins/state_history_plugin/state_history_plugin.cpp
huangminghuang: That requires changes from PR #10504 to work. Will be changed after that is merged.

plugins/state_history_plugin/state_history_plugin.cpp
huangminghuang: The exception is reported via post at the end of thread function.

plugins/state_history_plugin/state_history_plugin.cpp
heifner: See thread-safety at the bottom of the page: https://www.boost.org/doc/libs/1_76_0/doc/html/boost_asio/reference/ip__tcp/socket.html
Also: 
https://stackoverflow.com/questions/7362894/boostasiosocket-thread-safety

spoonincode: If you really want to do it, you can dup() the fd and create two boost sockets referring to the same underlying TCP connection. Then use one boost socket for read, one for write, on two separate threads. Not that i'm recommending doing that here, but that's how it can be done.

spoonincode (dismissed):
I would like to review this further so I'm just putting this in for now to block merging. In particular I'm concerned about the change to a sync write()

plugins/state_history_plugin/state_history_plugin.cpp
spoonincode: Seems racy. `send_thread_has_exception` won't be set true until some period of time after the io_context has stopped running. So it seems possible for `send()` to queue up another `post()` that will never be dispatched. If that's not a problem (and it might not be, I'm not sure on cursory look), then I don't see why this check needs to be here in the first place.

huangminghuang: The loop in the thread will break when the first task throw an exception, all the subsequent tasks would not be able to be executed. The check here is not strictly needed, just intended to be save some cycles.

plugins/state_history_plugin/state_history_plugin.cpp
huangminghuang: That was my first try; however, it didn't work under heavy load.

spoonincode:
The sync write was already a significant shut down concern, but I get the impression that the mutex introduced to fix the threading violation now makes the sync write even more dangerous. If the sync write is blocked, but data is received, the main thread then becomes blocked because the async_read cannot start until the blocked sync write completes.

Would like to understand more why this sync write is being used instead of async. I've heard performance several times but tests I've performed show async write being able to send north of 6GB/sec over a websocket. A single thread with a strand per connection would seem more than capable of handling the demand.

plugins/state_history_plugin/state_history_plugin.cpp
heifner: `medium` would be better as it would then be higher than transaction processing.

heifner (approved):
Would still like to understand why `async_write` performs so poorly in this code where it seems perfectly fine in a simple test program. Hopefully a follow on task can be worked to understand.

plugins/state_history_plugin/state_history_plugin.cpp
spoonincode: In case anyone finds this: my brain was thinking asio socket. The semantics of dup()ing a socket backed by a websocket::stream are almost certainly no good!

PR #10515, opened 2021-07-15
remove copy of reversible blocks directory in snapshot tests - 2.0

spoonincode:
## Change Description

This removed copy _needs to do nothing_, otherwise the test cannot pass because an opened (and thus dirty) reversible block database gets copied to the new chain data directory and cannot be used because it is dirty. Indeed, this copy did nothing until boost 1.74, and it also did nothing for boost 1.74+ with EOSIO/fc#176. But unfortunately that fc commit was not backported to 2.0.x

Since the test's behavior is dependent on this line _doing nothing_, remove the line :smile: 

For consistency this removal should be applied on all branches once accepted.

("nothing" here doesn't _literally_ mean nothing, but rather that a copy of the reversible DB file does not occur)

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/unittests/snapshot_tests.cpp
+++ b/unittests/snapshot_tests.cpp
@@ -43,7 +43,6 @@ controller::config copy_config_and_files(const controller::config& config, int o
    controller::config copied_config = copy_config(config, ordinal);
    fc::create_directories(copied_config.blocks_dir);
    fc::copy(config.blocks_dir / "blocks.log", copied_config.blocks_dir / "blocks.log");
-   fc::copy(config.blocks_dir / config::reversible_blocks_dir_name, copied_config.blocks_dir / config::reversible_blocks_dir_name );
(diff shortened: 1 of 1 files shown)


heifner (approved):

spoonincode:
Because this line is actually already removed in 2.2, I'm not going to bother with a 2.1 uplift. 2.1 won't break as is (because of the fc fix).

PR #10534, opened 2021-07-17
Support send transaction failure trace 2.2.x epe 996

jgiszczak:
## Change Description
See #10533 for details.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2966,6 +2966,65 @@ void read_write::send_transaction(const read_write::send_transaction_params& par
...
+            } CATCH_AND_CALL(next);
+         }
+      });
+   } catch ( boost::interprocess::bad_alloc& ) {
+      chain_plugin::handle_db_exhaustion();
+   } catch ( const std::bad_alloc& ) {
+      chain_plugin::handle_bad_alloc();
+   } CATCH_AND_CALL(next);
+}
+
```
heifner, line 3027: There is a lot of duplicated code here. Why not create a common function that these both call.

heifner (changes requested):

plugins/chain_plugin/chain_plugin.cpp
jgiszczak: Refactored.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2924,54 +2924,79 @@ void read_write::send_transaction(const read_write::send_transaction_params& par
...
    } CATCH_AND_CALL(next);
 }
 
+void read_write::send_transaction(packed_transaction_ptr input_trx, const std::string method, bool return_failure_traces, chain::plugin_interface::next_function<send_transaction_results> next) {
+   auto trx_trace = fc_create_trace_with_id("Transaction", input_trx->id());
+   auto trx_span = fc_create_span(trx_trace, "HTTP Received");
+   fc_add_tag(trx_span, "trx_id", input_trx->id());
+   fc_add_tag(trx_span, "method", method);
+
+   app().get_method<incoming::methods::transaction_async>()(input_trx, true, false, static_cast<const bool>(return_failure_traces),
```
heifner, line 2962: Why `static_cast<const bool>(return_failure_traces)`, the cast does not seem needed.

heifner (changes requested):

plugins/chain_plugin/chain_plugin.cpp
jgiszczak: You'd think so, but it won't build without it.  I don't understand why either, but the way the `get_method<>` template works, the cast is required.

heifner: Ah, template method, missed that. That make sense.

heifner (approved):

PR #10535, opened 2021-07-17
When will EOS development tools work with macOS Big Sur?

MasterMintZ:
When will EOS development tools work with macOS Big Sur?

I'm working on developing tokens and users on my local device however I believe its not working because I am running macOS Big Sur. Will you be updating them soon or do I need to get a computer will the older MacOS


heifner (changes requested):
Please open an issue not a pull request.

PR #10541, opened 2021-07-18

ndcgundlach:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
+++ b/plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
@@ -119,7 +119,7 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
          tracked_block_uuid_rks.clear();
          trx_queue_ptr->on_block_start();
       } else {
-         trx_queue_ptr->empty();
+         trx_queue_ptr->clear();
       }
 
    }
@@ -274,7 +274,6 @@ void amqp_trx_plugin::plugin_startup() {
 
    ilog( "Starting amqp_trx_plugin" );
 
-   // auto* prod_plugin = app().find_plugin<producer_plugin>();
    EOS_ASSERT( my->prod_plugin, chain::plugin_config_exception, "producer_plugin required" ); // should not be possible
    EOS_ASSERT( my->allow_speculative_execution || my->prod_plugin->has_producers(), chain::plugin_config_exception,
                "Must be a producer to run without amqp-trx-speculative-execution" );
--- a/plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
+++ b/plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
@@ -324,6 +324,10 @@ void amqp_trx_plugin::plugin_startup() {
                          }
    );
 
+   if (!my->prod_plugin->paused()) {
+      my->amqp_trx->start_consume(true);
+      my->started_consuming = true;
+   }
 }
 
 void amqp_trx_plugin::plugin_shutdown() {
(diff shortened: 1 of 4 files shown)


plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -260,9 +274,9 @@ void amqp_trx_plugin::plugin_startup() {
 
    ilog( "Starting amqp_trx_plugin" );
 
-   auto* prod_plugin = app().find_plugin<producer_plugin>();
-   EOS_ASSERT( prod_plugin, chain::plugin_config_exception, "producer_plugin required" ); // should not be possible
-   EOS_ASSERT( my->allow_speculative_execution || prod_plugin->has_producers(), chain::plugin_config_exception,
+   // auto* prod_plugin = app().find_plugin<producer_plugin>();
```
heifner, line 277: Remove commented out code.

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -106,9 +108,20 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
...
+            ilog("Starting consuming amqp messages during on_block_start");
+            amqp_trx->start_consume(true);
+            started_consuming = true;
+         }
+
+         block_uuid = boost::uuids::to_string( boost::uuids::random_generator()() );
+         tracked_block_uuid_rks.clear();
+         trx_queue_ptr->on_block_start();
+      } else {
+         trx_queue_ptr->empty();
```
heifner, line 122: This does not empty out the queue. It returns if queue is empty.

heifner (changes requested):

plugins/amqp_trx_plugin/amqp_trx_plugin.cpp
```diff
@@ -106,9 +108,20 @@ struct amqp_trx_plugin_impl : std::enable_shared_from_this<amqp_trx_plugin_impl>
...
+            ilog("Starting consuming amqp messages during on_block_start");
+            amqp_trx->start_consume(true);
+            started_consuming = true;
+         }
+
+         block_uuid = boost::uuids::to_string( boost::uuids::random_generator()() );
+         tracked_block_uuid_rks.clear();
+         trx_queue_ptr->on_block_start();
+      } else {
+         trx_queue_ptr->clear();
```
heifner, line 122: I'm confused. I thought we were not going to handle the `pause` case for now.
If we are going to handle the `pause` case now then I think we should `reject` the all received amqp messages.
We would also need to stop consuming from the amqp queue.

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -48,6 +48,7 @@ class rabbitmq : public stream_handler {
          appbase::app().quit();
          error = true;
       } );
+      declare_queue.start_consume(false);
```
heifner, line 51: This doesn't do anything since there is no `on_consume` provide.

programs/rodeos/streams/rabbitmq.hpp
```diff
@@ -66,6 +67,7 @@ class rabbitmq : public stream_handler {
          appbase::app().quit();
          error = true;
       } );
+      declare_exchange.start_consume(false);
```
heifner, line 70: This doesn't do anything since there is no `on_consume` provide.

heifner (approved):

PR #10547, opened 2021-07-19
disable undo_stack by default, set tested typical rocksdb configurati…

linhuang-blockone:
…on parameters if options file is not provided, and gives detailed warning if deprecated rdb-threads and rdb-max-files are used

## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/rodeos/cloner_plugin.cpp
+++ b/programs/rodeos/cloner_plugin.cpp
@@ -42,7 +42,7 @@ struct cloner_config : ship_client::connection_config {
    eosio::name filter_name = {}; // todo: remove
    std::string filter_wasm = {}; // todo: remove
    bool        profile = false;
-   bool        undo_stack_disabled = false;
+   bool        undo_stack_enabled = false;
    uint32_t    force_write_stride = 0;
 
 #ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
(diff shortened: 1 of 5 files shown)


programs/rodeos/cloner_plugin.cpp
```diff
@@ -302,7 +302,7 @@ void cloner_plugin::set_program_options(options_description& cli, options_descri
    op("filter-name", bpo::value<std::string>(), "Filter name");
    op("filter-wasm", bpo::value<std::string>(), "Filter wasm");
    op("profile-filter", bpo::bool_switch(), "Enable filter profiling");
-   op("disable-undo-stack", bpo::bool_switch(), "disable undo stack");
+   op("undo-stack", bpo::bool_switch(), "Enable  undo stack");
```
heifner, line 305: The description calls this `enable-undo-stack` which I think would be better.

heifner (changes requested):

programs/rodeos/cloner_plugin.cpp
huangminghuang: I would suggest not to use bool_swich and use `bpo::value<std::string>()`; in case if we want to change the default back to enabled in the future, we can just change the default value instead of the option name.

heifner (approved):

programs/rodeos/cloner_plugin.cpp
linhuang-blockone: Done. Thanks.

heifner (approved):

PR #10554, opened 2021-07-21
Add Homebrew package test steps to CICD.

williamblevins:
## Change Description

Add Homebrew package test steps to CICD.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [X] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [X] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.cicd/generate-pipeline.sh
```diff
@@ -696,9 +696,38 @@ EOF
           cd: ~
     agents:
       - "queue=mac-anka-node-fleet"
+    key: "macos1015pb"
     timeout: ${TIMEOUT:-30}
     skip: ${SKIP_MACOS_10_15}${SKIP_PACKAGE_BUILDER}${SKIP_MAC}
 
+  - label: ":darwin: macOS 10.15 - Test Package"
+    command:
+      - "git clone \$BUILDKITE_REPO eos && cd eos && $GIT_FETCH git checkout -f \$BUILDKITE_COMMIT"
```
kj4ezj, line 705: Not that it matters, but the `eos` argument to `git clone` is redundant.

.cicd/test-package.docker.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+. "${0%/*}/libfunctions.sh"
+
+echo '--- :docker: Pretest Setup'
+
+perform "docker pull $IMAGE"
+perform "docker run --rm -v \"$(pwd):/eos\" -w '/eos' -it $IMAGE ./.cicd/test-package.run.sh"
```
kj4ezj, line 9: I would escape the dollar sign (`$`) before the `pwd` command in the volume mount flag here. The present command will print something like this before running it...
```bash
docker run --rm -v "/home/cicd/work/eos:/eos" -w '/eos' -it docker.io/eosio/eosio:ubuntu-20.04-master ./.cicd/test-package.run.sh
```
...which is accurate, but not terribly useful to the customer because they will have to edit the command if they are copying and pasting it into their shell. If we escape that dollar sign, it will print something like this:
```bash
docker run --rm -v "$(pwd):/eos" -w '/eos' -it docker.io/eosio/eosio:ubuntu-20.04-master ./.cicd/test-package.run.sh
```
Then people can just copy-pasta.

.cicd/test-package.run.sh
```diff
@@ -1,22 +1,28 @@
...
+elif [[ $(yum --version 2>/dev/null) ]]; then # RHEL family packaging
+    perform 'yum check-update || :'
+    perform 'yum install -y /eos/*.rpm'
+elif [[ $(brew --version 2>/dev/null) ]]; then # homebrew packaging
+    perform 'brew update'
+    perform 'mkdir homebrew-eosio'
+    perform 'git init homebrew-eosio'
+    perform 'cp *.rb homebrew-eosio'
+    perform "sed -i.bk -e 's/url \".*\"/url \"http:\/\/127.0.0.1:7800\"/' homebrew-eosio/*.rb"
+    perform "pushd homebrew-eosio && git add *.rb && git commit -m 'test it!' && popd"
```
kj4ezj, line 20: We don't need the ampersand chaining of commands (`&&`) outside of dockerfiles. I think it looks messy and is harder to follow than just making them individual lines. Especially if you're trying to see the output of one of the commands, or figure out which command threw an error message.

.cicd/test-package.run.sh
```diff
@@ -1,22 +1,28 @@
...
+    perform "brew tap eosio/eosio homebrew-eosio"
+    perform '{ python3 -m http.server 7800 & } && export HTTP_SERVER_PID=$!'
```
kj4ezj, line 22: What do the curly braces (`{}`) do here?

scripts/generate_bottle.sh
```diff
@@ -47,7 +47,7 @@ echo "class Eosio < Formula
    depends_on \"libusb\"
    depends_on \"postgresql\"
    depends_on macos: :mojave
-   depends_on arch: intel
+   depends_on arch: :intel
```
kj4ezj, line 50: Nice catch!!!

kj4ezj (changes requested):
Two suggested changes, and a few other comments.

.cicd/test-package.run.sh
williamblevins: They provide a scope for the `&` so that BASH doesn't have a syntax error from backgrounding a process followed by a chained command.

kj4ezj (approved):
🍎

.cicd/test-package.docker.sh
williamblevins: I'm ok with it printing the literal path. It might be slightly verbose. I don't see this being "broken" and it seems like to much work and too many builds to change this.

.cicd/generate-pipeline.sh
williamblevins: This would matter on mirrors repos.

PR #10555, opened 2021-07-21
Add Homebrew package test steps to CICD.

williamblevins:
## Change Description

Add Homebrew package test steps to CICD.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [X] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [X] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.cicd/test-package.anka.sh
```diff
@@ -0,0 +1,18 @@
...
+set -euo pipefail
+
+. "${0%/*}/libfunctions.sh"
+
+echo '--- :anka: Pretest Setup'
+
+if [[ ! $(brew --version 2>/dev/null) ]]; then
+   perform 'echo | /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"'
+   perform 'brew analytics off'
+fi
```
kj4ezj: I guess I missed this in the previous reviews. Homebrew is already installed on the Anka base layers and, even if it weren't, we should _never_ be piping `curl` into `bash`. There are a number of reasons why this is bad practice, but [here](https://www.idontplaydarts.com/2016/04/detecting-curl-pipe-bash-server-side/) is a particularly interesting one.

If we _have_ to blindly trust BASH scripts from the Internet, we should at least save and retain a local copy so that we can do forensic analysis on it if we are attacked.

We should also remove this code if it exists on the other branches.

kj4ezj (changes requested):
I am embarrassed I did not notice this in the earlier peer-reviews. 🔓

sandie06 (approved):
LGTM

kj4ezj (approved):
🍎

PR #10562, opened 2021-07-21
Support push action using --read-only flag for read only action. EPE 1134

softprofe:
## Change Description

As the push read only transaction was supported but no read only action support, so we add support push read only action here.
See EPE 1134 also. https://blockone.atlassian.net/browse/EPE-1134

The code seems simple it add a flag for action, and when call to push read only action it will route to push read only transaction, share the code of  push read only transaction. per my test, it works for get action in https://blockone.atlassian.net/wiki/spaces/EOSIOProdEng/pages/1279164436/TDD+Readonly+Queries. 

example:
./cleos -u http://127.0.0.1:8011 push action --read-only  readtest1112 get '[]' -j  -p readtest1112@active
 
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -3882,7 +3882,7 @@ int main( int argc, char** argv ) {
    actionsSubcommand->add_option("action", action,
                                  localized("A JSON string or filename defining the action to execute on the contract"), true)->required();
    actionsSubcommand->add_option("data", data, localized("The arguments to the contract"))->required();
-
+   actionsSubcommand->add_flag("-o,--read-only", tx_read_only, localized("Specify the action is read-only"));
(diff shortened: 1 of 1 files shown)


nickjjzhao (commented):
It would be easier to open the associated Jira ticket if there is a link to `EPE 1134` in Change Description.

praphael (approved):

PR #10576, opened 2021-07-23
fix threaded send

huangminghuang:
## Change Description

This PR fixes the problem where the ship socket is accessed from two threads at the same time. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -180,46 +180,34 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
       }
 
       void start_read() {
-         auto in_buffer = std::make_shared<boost::beast::flat_buffer>();
-         derived_session().socket_stream->async_read(
-             *in_buffer, [self = derived_session().shared_from_this(), in_buffer](boost::system::error_code ec, size_t) {
-                self->callback(ec, "async_read", [self, in_buffer] {
-                   auto d = boost::asio::buffer_cast<char const*>(boost::beast::buffers_front(in_buffer->data()));
-                   auto s = boost::asio::buffer_size(in_buffer->data());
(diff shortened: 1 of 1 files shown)


spoonincode:
I think in the async_read completion handler, you could post() to the strand the next call to socket->async_read(). That may be simpler than this.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -180,46 +180,34 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-                   auto d = boost::asio::buffer_cast<char const*>(boost::beast::buffers_front(in_buffer->data()));
-                   auto s = boost::asio::buffer_size(in_buffer->data());
-                   fc::datastream<const char*> ds(d, s);
-                   state_request               req;
-                   fc::raw::unpack(ds, req);
-                   std::visit(*self, std::move(req));
-                });
-             });
+         // use reactor pattern with blocking read to avoid mixing async_read() and block write(). 
+         derived_session().socket_stream->next_layer().async_wait(
```
spoonincode, line 184: Does this work correctly if two websocket payload frames are delivered "atomically" to the socket? My concern would be the async_wait() is fired once, the websocket::stream::read() reads in two payloads from the underlying socket, and then returns on the first payload while buffering the second one for the next read(). At that point the async_wait would not be fired for the second payload because the socket has been drained from the kernel's standpoint.

huangminghuang:
replaced by #10597

PR #10577, opened 2021-07-26
Stop scanning a trace index log when a target block is found - develop

nickjjzhao:
## Change Description

Regarding RPC API `/trace_api/get_block`, since block numbers are unique in trace index logs, it doesn't look like necessary to keep scanning an index log when a target block is found in the log. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

### DIFF
--- a/plugins/trace_api_plugin/store_provider.cpp
+++ b/plugins/trace_api_plugin/store_provider.cpp
@@ -67,6 +67,7 @@ namespace eosio::trace_api {
             const auto& block = std::get<block_entry_v0>(e);
             if (block.number == block_height) {
                trace_offset = block.offset;
+               return false; // stop scanning
             }
          } else if (std::holds_alternative<lib_entry_v0>(e)) {
             auto lib = std::get<lib_entry_v0>(e).lib;


plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -67,6 +67,7 @@ namespace eosio::trace_api {
             const auto& block = std::get<block_entry_v0>(e);
             if (block.number == block_height) {
                trace_offset = block.offset;
+               return false; // stop scanning
```
heifner, line 70: We keep scanning until we hit lib (below) because forks can cause duplicate block numbers added to the log. We want to return the most recent and we read from front to back of the file which is appended to.

heifner (changes requested):

plugins/trace_api_plugin/store_provider.cpp
nickjjzhao: Thank you! That makes sense.

nickjjzhao:
There is a reason to keep scanning a trace index log when a target block is found. Closing this PR.

PR #10579, opened 2021-07-27
use generated ship abi

huangminghuang:
## Change Description

This PR update the abieos submodule and utilize the generated generated ship abi json instead of the hard coded one. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -85,5 +85,6 @@ var/lib/node_*
 .idea/
 *.iws
 .DS_Store
+node_modules/*
 
 !*.swagger.*
--- a/libraries/CMakeLists.txt
+++ b/libraries/CMakeLists.txt
@@ -29,6 +29,7 @@ add_subdirectory( chain )
 add_subdirectory( testing )
 add_subdirectory( version )
 add_subdirectory( state_history )
+set(ABIEOS_BUILD_SHARED_LIB OFF)
 add_subdirectory( abieos )
 
 # Suppress warnings on 3rdParty Library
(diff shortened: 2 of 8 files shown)


spoonincode:
The abieos changes break cross compiling. For example, at one time I had an iOS application built via cmake that used abieos, and this change would break that existing build.

Personally, I find it disappointing to break cross compiling in eosio. We've had successful Win32 cross compiled builds in the past (pre 2.1). Regardless of eosio, abieos is a library users are encouraged to integrate, so breaking legitimate build strategies is more unfortunate there.

.gitignore
```diff
@@ -85,5 +85,6 @@ var/lib/node_*
 .idea/
 *.iws
 .DS_Store
+node_modules/*
```
victorj8, line 88: what is the purpose of ignoring this folder? (as i don't see nodejs code being added here)

huangminghuang: I just need to run the `ship_test` to make sure everything works. To run the `ship_test` requires you to do an `npm install`. It would install the javascript modules on the `node_modules` directory under the source tree rather than the build tree.

victorj8 (approved):

PR #10580, opened 2021-07-27
test control plugin fix for case with performance issue

dimas1185:
## Change Description

Fix for a performance issue case when producer produce less than 12 blocks in a round. This scenario happened in this fail:
https://buildkite.com/EOSIO/eosio-build-unpinned/builds/7907#c88a2b32-e4ac-4df1-85c8-d8584eaf4c0d
This is very rare so to test this scenario, I slightly modified producer plugin to add sleep during block production.
Currently if this happening test_control_plugin doesn't exit nodeos as it is waiting for 12th block. Now when it happens it will determine the last block in producer schedule and you'll see similar to this output:
```
info  2021-07-27T02:11:54.030 thread-0  test_control_plugin.cp:66     process_next_block_s ] producer defproducerb, looking for defproducerb
info  2021-07-27T02:11:54.030 thread-0  test_control_plugin.cp:75     process_next_block_s ] producer defproducerb seq: 6 slot: 12
warn  2021-07-27T02:11:54.030 thread-0  test_control_plugin.cp:84     process_next_block_s ] Producer produced less than 12 blocks, 6th block is last in sequence. Likely performance issue, check timing
info  2021-07-27T02:11:54.030 thread-0  test_control_plugin.cp:86     process_next_block_s ] shutting down
```

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [] Existing Tests

- [x] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/test_control_plugin/test_control_plugin.cpp
+++ b/plugins/test_control_plugin/test_control_plugin.cpp
@@ -26,6 +26,7 @@ private:
    account_name        _producer;
    int32_t             _where_in_sequence;
    int32_t             _producer_sequence;
+   uint32_t            _first_sequence_timeslot;
    bool                _clean_producer_sequence;
    std::atomic_bool    _track_lib;
    std::atomic_bool    _track_head;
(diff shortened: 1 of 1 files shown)


plugins/test_control_plugin/test_control_plugin.cpp
```diff
@@ -26,6 +26,7 @@ class test_control_plugin_impl {
    account_name        _producer;
    int32_t             _where_in_sequence;
    int32_t             _producer_sequence;
+   uint32_t            _first_sequence_timeslot;
```
heifner: All three of these ints are not initialized and looks like they are accessed before being assigned causing undefined behavior.

plugins/test_control_plugin/test_control_plugin.cpp
```diff
@@ -79,6 +92,8 @@ void test_control_plugin_impl::process_next_block_state(const chain::block_state
       _producer_sequence = -1;
```
heifner, line 95: Seems like the code would be clearer if `_producer_sequence` was initialized to 0 and not have all the `_producer_sequence + 1` throughout this code. Would allow you to remove the comment you added `_producer_sequence is index, aligning it with slot number`.

plugins/test_control_plugin/test_control_plugin.cpp
```diff
@@ -66,10 +67,22 @@ void test_control_plugin_impl::process_next_block_state(const chain::block_state
...
-
-      if (_producer_sequence >= _where_in_sequence) {
+      ilog("producer ${prod} seq: ${seq} slot: ${slot}", 
+           ("prod", producer_name.to_string())
+           ("seq", _producer_sequence+1) // _producer_sequence is index, aligning it with slot number
+           ("slot", slot - _first_sequence_timeslot));
+
+      bool last_slot = false;
+      if (_where_in_sequence + 1 == config::producer_repetitions)
+         last_slot = slot == _first_sequence_timeslot + config::producer_repetitions;
```
heifner: A parentheses or two would make this easier to read. Took me a second.

heifner (changes requested):

plugins/test_control_plugin/test_control_plugin.cpp
dimas1185: it should be aligned with `_where_in_sequence` and later is used in external interface so I'll need to change it in few tests if I would do this. Sounds too much for couple of +1 statements

heifner: Sounds good.

heifner (approved):

PR #10581, opened 2021-07-28
UNIX socket support for state-history_plugin of nodeos

plroblox:
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

A new option of nodeos is added in this PR:
  --state-history-unix-socket-path=unixPath
where unixPath can be an absolute path such as /foo.localSock, or a related path  foo.localSock2.
The relative path is relative to data-dir of nodeos.  This option  --state-history-unix-socket-path=unixPath of nodeos creates
 a unix socket upon which to listen for incoming connections

a new integration test named ship_test_unix is added to test Unix socket of state_history_plugin of nodeos work correctly. This can be verified by navigating to eos/build/ directory and then  use ctest command **ctest -R "^ship_test_unix$" -V**


heifner:
Please update PR description to include new command line options and relevant info for docs. Remember public does not have access to EPE-1140. Also link to the develop-boxed PR.

spoonincode:
Would also recommend taking in #10071 (probably just a2ed741 applies here, I think)

plroblox:
OK, I will update the documentation on new command line options, 

#10071 has changes in rodeos, do you think I should also cherry-pick rodeos change #10001 on UNIX socket 
and then take in #10071? Of course I can just 'pick' the part of #10071 relative to the server-side change

plroblox:
Question: win32 has no concept of UNIX/local socket, why did you remove BOOST_ASIO_HAS_LOCAL_SOCKETS ifdefs ?

spoonincode:
> Question: win32 has no concept of UNIX/local socket, why did you remove BOOST_ASIO_HAS_LOCAL_SOCKETS ifdefs ?

This is talked about in #10040, Windows 10 does have native AF_UNIX sockets now.

My understanding is that rodeos is going to be "copied" wholesale from boxed, maybe @heifner can confirm. So I think taking only the changes from a2ed741 that apply to this ship stuff is fine. The rodeos changes in a2ed741 will get picked up when rodeos is copied over.

heifner:
> My understanding is that rodeos is going to be "copied" wholesale from boxed, maybe @heifner can confirm. 

Yes, the plan is to pull in `programs/rodeos` and `libraries/rodeos` directories wholesale from develop-boxed. We will want those to maintain git log history, but no need to pull them over one individual PR at a time.

spoonincode:
This looks fine but we should take the removal of the ifdefs from a2ed741

spoonincode:
Please see Kevin's comments
https://github.com/EOSIO/eos/pull/10581#issuecomment-888443919
A reasonable description needs to be added (even just referring to the original PR is likely fine), and the title could probably be cleaned up a little as, like Kevin says, EPE tickets mean nothing to our users

plroblox:
Matt: title is changed, and documentation is updated. Meanwhile I am asking automation to see why buildkite checks are still failing for 2 tests

spoonincode (approved):

PR #10583, opened 2021-07-28
GH-3678 migrate http_plugin to boost::beast EPE-9

praphael:
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,136 @@
...
+    // Accepts incoming connections and launches the sessions
+    // session_type should be a subclass of beast_http_session
+    // protocol type must have sub types acceptor and endpoint, e.g. boost::asio::ip::tcp;
+    // socket type must be the socket e.g, boost::asio::ip::tcp::socket
+    template<typename session_type, typename protocol_type, typename socket_type> 
+    
+    class beast_http_listener : public std::enable_shared_from_this<beast_http_listener<session_type, protocol_type, socket_type> >
+    {
+        private: 
+            bool isListening_;
```
heifner: default constructor leaves this uninitialized which is undefined behavior

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include <type_traits>
+#include <sstream>
+
+namespace eosio {
+    // Accepts incoming connections and launches the sessions
+    // session_type should be a subclass of beast_http_session
+    // protocol type must have sub types acceptor and endpoint, e.g. boost::asio::ip::tcp;
+    // socket type must be the socket e.g, boost::asio::ip::tcp::socket
+    template<typename session_type, typename protocol_type, typename socket_type> 
+
```
heifner: Can we remove this blank line.

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,136 @@
...
+                } catch( ... ) { }
+            }; 
+
+            void listen(typename protocol_type::endpoint endpoint) {
+                if (isListening_) return;
+
+                // for unix sockets we should delete the old socket
+                if(std::is_same<socket_type, stream_protocol::socket>::value) { 
+                    auto ss = std::stringstream();
+                    ss << endpoint;
```
heifner: I'm not sure there is any indication that this will always be the path. It is documented as a human readable representation. I think you should use `.path()` instead.

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+                // Bind to the server address
+                acceptor_.bind(endpoint, ec);
+                if(ec)  {
+                    fail(ec, "bind");
+                    return;
+                }
+
+                // Start listening for connections
+                auto maxConnections = asio::socket_base::max_listen_connections;
```
heifner: We use snake_case exclusively for variables. `max_connections`

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+    //  https://security.stackexchange.com/questions/91435/how-to-handle-a-malicious-ssl-tls-shutdown
+    void fail(beast::error_code ec, char const* what)
+    {
+        fc_elog(logger, "${w}: ${m}", ("w", what)("m", ec.message()));
+        fc_elog(logger, "closing connection");
+    }
+
+    // this needs to be declared outside of the beast_http_session, because it doesn't work for 
+    // UNIX socket session, since it lacks address() function in endpoint
+    //  Since this is a virtual function, std::enable_if_t/SFNIAE doesn't work
```
heifner: This is not a `virtual` function. What does this comment mean?

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                            std::shared_ptr<http_plugin_state> plugin_state, 
+                            std::shared_ptr<eosio::chain::named_thread_pool> thread_pool)
+            : beast_http_session(plugin_state, thread_pool)
+            , socket_(std::move(sock)) 
+            {  }
+
+            virtual ~unix_socket_session() = default;
+
+            bool allow_host(const http::request<http::string_body>& req) { 
+                // TODO allow host make sense here ?
```
heifner: Remove TODO comment

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                                is_conn_secure)) 
+        {
+            return false;
+        }
+
+        return true;
+    }
+
+    // Handle HTTP conneciton using boost::beast for TCP communication
+    // This uses the Curiously Recurring Template Pattern so that
```
heifner: This is not an example of the `Curiously Recurring Template Pattern`. Ah, you mean derived classes can use the CRTP with this class.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+            // HTTP parser object
+            std::optional<http::request_parser<http::string_body>> req_parser_;
+
+            // HTTP response object
+            http::response<http::string_body> res_;
+
+            std::shared_ptr<http_plugin_state> plugin_state_;
+
+            // whetner resposne should be sent back to client when an exception occured
+            bool isSendExceptionResponse_;
```
heifner: This is uninitialized in the default constructor.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+    template<class Derived> 
+    class beast_http_session :  public detail::abstract_conn
+    {
+        protected:
+            beast::flat_buffer buffer_;
+            beast::error_code ec_;
+
+            // time points for timeout measurement and perf metrics 
+            steady_clock::time_point session_begin_, read_begin_, handle_begin_, write_begin_;
+            uint64_t read_time_us_, handle_time_us_, write_time_us_;
```
heifner: These are uninitialized in the default constructor.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+
+            // HTTP parser object
+            std::optional<http::request_parser<http::string_body>> req_parser_;
+
+            // HTTP response object
+            http::response<http::string_body> res_;
+
+            std::shared_ptr<http_plugin_state> plugin_state_;
+
+            // whetner resposne should be sent back to client when an exception occured
```
heifner: Fix spelling

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                        fc_dlog( logger, "404 - not found: ${ep}", ("ep", resource) );
+                        not_found(resource, *this);                    
+                    }
+                } catch (...) {
+                    handle_exception();
+                }
+            }
+
+            void report_429_error(const std::string & what) {                
+                send_response(std::string(what),
```
heifner: Extra copy of `what` not needed.

heifner: `report_429_error` could take a `std::string` which could then be moved into `send_response`

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                            static_cast<int>(http::status::too_many_requests));                
+            }
+
+        public:
+            virtual bool verify_max_bytes_in_flight() override {
+                auto bytes_in_flight_size = plugin_state_->bytes_in_flight.load();
+                if( bytes_in_flight_size > plugin_state_->max_bytes_in_flight ) {
+                    fc_dlog( logger, "429 - too many bytes in flight: ${bytes}", ("bytes", bytes_in_flight_size) );
+                    string what = "Too many bytes in flight: " + std::to_string( bytes_in_flight_size ) + ". Try again later.";;
+                    report_429_error(what);
```
heifner: If `report_429_error` took a `std::string` then `what` could be moved into it.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+
+            virtual bool verify_max_requests_in_flight() override {
+                if (plugin_state_->max_requests_in_flight < 0)
+                    return true;
+
+                auto requests_in_flight_num = plugin_state_->requests_in_flight.load();
+                if( requests_in_flight_num > plugin_state_->max_requests_in_flight ) {
+                    fc_dlog( logger, "429 - too many requests in flight: ${requests}", ("requests", requests_in_flight_num) );
+                    string what = "Too many requests in flight: " + std::to_string( requests_in_flight_num ) + ". Try again later.";
+                    report_429_error(what);
```
heifner: Same comment as above, this could be `std::move(what)`

plugins/http_plugin/beast_http_session.hpp
heifner: Also this should use snake_case `is_send_exception_response_`

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+            Derived& derived()
+            {
+                return static_cast<Derived&>(*this);
+            }
+            
+        public: 
+            // shared_from_this() requires default constuctor
+            beast_http_session() = default;
+      
+            // Take ownership of the buffer
```
heifner: Not sure what this comment means.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                // dfeault to true
+                isSendExceptionResponse_ = true;
+            }
+
+            virtual ~beast_http_session() {
+                // avoid exceptions being thrown in destructor
+                isSendExceptionResponse_ = false;
+                try { 
+                    plugin_state_->requests_in_flight -= 1;
+    #if PRINT_PERF_METRICS
```
heifner: I don't think this should be a `#if`. It could be left as debug level logging always. Or if you think debug is too much you can do:
     ` if( logger.is_enabled( fc::log_level::all ) )`

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                read_time_us_ = handle_time_us_ = write_time_us_ = 0;
+
+                // dfeault to true
+                isSendExceptionResponse_ = true;
+            }
+
+            virtual ~beast_http_session() {
+                // avoid exceptions being thrown in destructor
+                isSendExceptionResponse_ = false;
+                try {
```
heifner: None of this should throw, right? What are you guarding against here?

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+            
+            std::shared_ptr<eosio::chain::named_thread_pool> thread_pool_;
+
+            template<
+                class Body, class Allocator>
+            void
+            handle_request(
+                http::request<Body, http::basic_fields<Allocator>>&& req)
+            {
+                auto &res = res_;
```
heifner: Why this local reference?

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,587 @@
...
+                write_time_us_ += std::chrono::duration_cast<std::chrono::microseconds>(dt).count();
+
+                if(close)
+                {
+                    // This means we should close the connection, usually because
+                    // the response indicated the "Connection: close" semantic.
+                    return derived().do_eof();
+                }
+
+                // create a new parser to clear state
```
heifner, line 326: I was expecting to see `res_` cleared here as well. Seems like we might hold on to a lot of memory if we do not clear it out.

heifner (changes requested):
I'm done reviewing for the day, but not done reviewing. I'm submitting this now because I'm not sure when I'll have time to get back to this.

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+#include <sstream>
+
+namespace eosio {
+    // since beast_http_listener handles both TCP and UNIX endpoints we need a template here 
+    // to get the path if makes sense, so that we can call ::unlink() before opening socket 
+    // in beast_http_listener::listen() by tdefault return blank string
+    template<typename T>
+    std::string get_endpoint_path(T endpt) { return {}; }
+
+    std::string get_endpoint_path(stream_protocol::endpoint endpt) { return endpt.path(); }
```
heifner: `endpt` can be a `const&` both here and on line 13.

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+            beast_http_listener& operator=(const beast_http_listener&) = delete;
+            beast_http_listener& operator=(beast_http_listener&&) = delete;
+
+            beast_http_listener(std::shared_ptr<eosio::chain::named_thread_pool> thread_pool, 
+                                std::shared_ptr<ssl::context> ctx, 
+                                std::shared_ptr<http_plugin_state> plugin_state) : 
+                is_listening_(false)
+                , thread_pool_(thread_pool)
+                , ctx_(ctx)
+                , plugin_state_(plugin_state)
```
heifner: These passed by value `shared_ptr` can be `std::move`.

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,141 @@
...
+    {
+        private: 
+            bool is_listening_ = false; 
+
+            std::shared_ptr<eosio::chain::named_thread_pool> thread_pool_;
+            std::shared_ptr<ssl::context> ctx_;
+            std::shared_ptr<http_plugin_state> plugin_state_;
+
+            typename protocol_type::acceptor acceptor_;
+            typename protocol_type::endpoint listen_ep_;
```
heifner: `listen_ep_` is not used.

plugins/http_plugin/beast_http_listener.hpp
praphael: They can't because the `http_plugin`,` beast_http_listener` and `beast_http_session` all live in different threads.   Therefore you can't safely access a reference to an object in one thread that was created by another, unless you guarantee the object lifetime.  

While it should be the case that  http_plugin will outlive the objects it creates (such as thread pool), this is not guaranteed.  The restart_scenarios_XXX in particularly, were sporadically failing because nodeos was returning a non-zero exit code (either -6 or -11, which corresponds to unhandled exception and segmentation fault), because of a race condition in the order of destruction of the objects.  This was fixed by having any objects which were referenced by seprate threads be shared pointers - namely the thread_pool and plugin_state.

I could refactor this such that beast_http_listener is the one to create the thread pool.  It would still need to be shared_ptr as beast_http_session needs it.

heifner: A `std::move` here would not cause them to use the same reference. Specifically, I just meant to do this:

```
            beast_http_listener(std::shared_ptr<eosio::chain::named_thread_pool> thread_pool, 
                                std::shared_ptr<ssl::context> ctx, 
                                std::shared_ptr<http_plugin_state> plugin_state) : 
                is_listening_(false)
                , thread_pool_(std::move(thread_pool))
                , ctx_(std::move(ctx))
                , plugin_state_(std::move(plugin_state))
                , acceptor_(thread_pool_->get_executor())
                , socket_(thread_pool_->get_executor())
            { }

```

The move of shared_ptr just saves an atomic copy. You have already made a copy because the constructor take a value instead of a reference. No need to make another copy in the initializer list.

praphael: Ah ok - thanks for clearing that up.

plugins/http_plugin/common.hpp
```diff
@@ -0,0 +1,274 @@
...
+    bool host_is_valid( const http_plugin_state& plugin_state, 
+                        const std::string& host, 
+                        const string& endpoint_local_host_port, 
+                        bool secure) {
+        if (!plugin_state.validate_host) {
+            return true;
+        }
+
+        // normalise the incoming host so that it always has the explicit port
+        static auto has_port_expr = std::regex("[^:]:[0-9]+$"); /// ends in :<number> without a preceeding colon which implies ipv6
```
heifner: I think this would be better than regex: https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/ip__address.html

plugins/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+            beast_http_listener(std::shared_ptr<eosio::chain::named_thread_pool> thread_pool, 
+                                std::shared_ptr<ssl::context> ctx, 
+                                std::shared_ptr<http_plugin_state> plugin_state) : 
+                is_listening_(false)
+                , thread_pool_(std::move(thread_pool))
+                , ctx_(std::move(ctx))
+                , plugin_state_(std::move(plugin_state))
+                , acceptor_(thread_pool_->get_executor())
+                , socket_(thread_pool_->get_executor())
```
heifner, line 56: Even if we have convinced ourselves that an explicit strand is not required for our use case, I think it makes sense to still use a strand so that we more closely align with the boost beast examples.

praphael: This is possible, although for boost versions prior to 1.70 the last solution I used was rather ugly.  If we do this I would prefer to bump the min boost version up to 1.70

The use of strands reinforces that the session objects are not thread-safe - i.e the object is intended to live, be accessed and die within a single thread.  Only the plugin_state objects are thread safe.  This is fine as we are not doing anything fancy like pipelining requests within a session.  There may be a slight perf penalty for using strands, though the vast majority of time (> 90%) is spent handling the request anyway.

heifner: I would prefer not making it ugly. I've not been able to make it crash by bombarding it with requests, so I would prefer to leave as is for now.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,579 @@
...
+
+                session_begin_ = steady_clock::now();
+                read_time_us_ = handle_time_us_ = write_time_us_ = 0;
+
+                // default to true
+                is_send_exception_response_ = true;
+            }
+
+            virtual ~beast_http_session() {
+                // avoid exceptions being thrown in destructor
```
heifner: This comment is no longer applicable.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,579 @@
...
+                    derived().stream(),
+                    *res_,
+                    [self, close](beast::error_code ec, std::size_t bytes_transferred) {
+                        self->on_write(ec, bytes_transferred, close);
+                    }
+                );
+            }
+
+            void run_session() {
+                // wrap entiere session around try/catch to prevent non-zero exit code test failure
```
heifner: This comment does not seem relevant. Should we add a try-catch here to be safe?

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,579 @@
...
+                    // At this point the connection is closed gracefully
+                } catch (...) { 
+                    handle_exception(); 
+                }
+            }
+
+            bool is_secure() { return false; };
+
+            void run() {
+                // catch any loose exceptions so that nodeos will return zero exit code
```
heifner: This also does not seem to apply anymore.

plugins/http_plugin/common.hpp
```diff
@@ -0,0 +1,274 @@
...
+    auto make_in_flight(T&& object, http_plugin_state& plugin_state) {
+        return std::make_shared<in_flight<T>>(std::forward<T>(object), plugin_state);
+    }
+
+    /**
+     * Construct a lambda appropriate for url_response_callback that will
+     * JSON-stringify the provided response
+     *
+     * @param thread_pool - pointer to the thread pool 
+     * @param plugin_state = reference to plugin state object
```
heifner: It is not a reference. I would jus remove `pointer`, `reference`, `pointer` from all these descriptions. You can see what they are by looking at them.

plugins/http_plugin/common.hpp
```diff
@@ -0,0 +1,274 @@
...
+     * @param session_ptr - pointer to the session object on which to invoke send_response
+     * @return lambda suitable for url_response_callback
+     */
+    auto make_http_response_handler(std::shared_ptr<eosio::chain::named_thread_pool> thread_pool, std::shared_ptr<http_plugin_state> plugin_state, detail::abstract_conn_ptr session_ptr) {
+    return [thread_pool, plugin_state, session_ptr]( int code, std::optional<fc::variant> response )
```
heifner: These can all be moved.
`return [thread_pool{std::move(thread_pool)}, plugin_state{std::move(plugin_state)}, session_ptr{std::move(session_ptr)}]( int code, std::optional<fc::variant> response )`

plugins/http_plugin/beast_http_session.hpp
praphael: It should not be necessary.   There isn't much code between that point and either an  async_read will happen, which cannot throw an exception, or session will be terminated, if max_requests_in_flight is exceeded.

plugins/http_plugin/common.hpp
```diff
@@ -0,0 +1,276 @@
...
+                _plugin_state.bytes_in_flight -= _count;
+            }
+        }
+
+        // No copy constructor, but allow move
+        in_flight(const in_flight&) = delete;
+        in_flight(in_flight&& from)
+        :_object(std::move(from._object))
+        ,_count(from._count)
+        ,_plugin_state(std::move(from._plugin_state))
```
heifner, line 232: This `std::move` is not doing anything and you are copying all the plugin state here.

plugins/http_plugin/common.hpp
```diff
@@ -0,0 +1,276 @@
...
+        {
+            from._count = 0;
+        }
+
+        // No copy assignment, but allow move
+        in_flight& operator=(const in_flight&) = delete;
+        in_flight& operator=(in_flight&& from) {
+            _object = std::move(from._object);
+            _count = from._count;
+            _plugin_state = from._plugin_state;
```
heifner: This is making a copy of plugin_state.
You actually don't need the move constructor or the move `operator=`, so you can just declare them both as `= delete`.

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,576 @@
...
+
+                // This means they closed the connection
+                if(ec == http::error::end_of_stream)
+                    return derived().do_eof();
+
+                if(ec) {
+                    return fail(ec, "read");
+                }
+
+                auto req = req_parser_->get();
```
heifner: Can call `release()` instead of `get()` to move parsed message out of `req_parser_`

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,576 @@
...
+                auto const bad_request =
+                [](beast::string_view why, detail::abstract_conn& conn)
+                {
+                    conn.send_response(std::string(why), 
+                                    static_cast<int>(http::status::bad_request));
+                };
+
+                // Returns a not found response
+                auto const not_found =
+                [](beast::string_view target, detail::abstract_conn& conn)
```
heifner: The only place this is called from has a string for target, change to `const std::string&`

plugins/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,576 @@
...
+                    }
+
+                    // verfiy bytes in flight/requests in flight
+                    if( !verify_max_bytes_in_flight() ) return;
+
+                    std::string resource = std::string(req.target());
+                    // look for the URL handler to handle this reosouce
+                    auto handler_itr = plugin_state_->url_handlers.find( resource );
+                    if( handler_itr != plugin_state_->url_handlers.end()) {
+                        std::string body = req.body();
```
heifner, line 182: I think it would be good to debug log the `resource` here.

plugins/http_plugin/common.hpp
praphael: Was http_plugin designed to support IP v6?   If not, I can convert the addresses to IP v4 for comparison.  The initialization code  `http_plugin::plugin_initialize()` seems to assume IP v4, because its looking for a host/port and calling `resolve` with `tcp::v4()` as an argument.

heifner: It probably was not designed to support IP v6, but if you can add support as part of this PR that would be a plus.

Specifically, instead of regex, you should be able to do something like this:

```
auto add = ip::address::make_address( host );
if( add.is_v6() ) 
```

plugins/http_plugin/common.hpp
praphael: plugin_state shouldn't be copied as it is stored as reference within the` in_flight` object, but yes the std::move is pointless.

heifner: plugin_state is copied. That is how references work. https://isocpp.org/wiki/faq/references#reseating-refs You can't assign references, you only copy the entire object. Specifically for the operator=.

praphael: You are correct ...  I mixed up initialization with assignment.   I don't think move assignment is necessary for this code so I will delete this call.

plugins/http_plugin/common.hpp
praphael: This was my thought as well.  We can just normalize addresses to IP v6, shouldn't require conditionals.

heifner: Nice

praphael: I have added support for IPv6 and reworked the allow_hosts logic.  The existing implementation struck me as a bit type-unsafe and did some unnecessary conversion between addresses and strings.  Hopefully it is cleaned up now.  The one thing we might want to take a closer look at is a function I added called get_ipaddr_host_port(), which takes either an IPv4 or IPv6 address string and splits it up into host/port, autodetecting which version it is.  I couldn't find a library implementation for this,  so I made a simple implementation to do this which assumes IPv6 addresses contain at least open bracket '[, though it doesn't have to be a closed bracket unless it contains a port number.  I believe there is an RFC which this should be the case for IPv6 addresses which contain port numbers, but I'm not sure about IPv6 addresses which do not contain port numbers..  

Also not sure how strict it should be, such as detecting the ':' separator (which wasn't done in current implementation even for IPv4)

praphael:
Rolled back IPv6, new branch PR: https://github.com/EOSIO/eos/pull/10689

PR #10586, opened 2021-08-02
Add test case for amqp start up paused

ndcgundlach:
## Change Description

Add test case for amqp start up paused
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


tests/amqp_pause_test.sh
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/bash -x
```
heifner, line 1: I think this should be a python test. We have standardized on python as the language for tests. I don't think we should make an exception for this test.

tests/amqp_pause_test.sh
```diff
@@ -0,0 +1,74 @@
...
+# Test for amqp queue listening to be paused in a paused producer
+
+echo 'making sure we are clean.  one or both of these commands may fail.'
+killall nodeos
+rm -rf var/lib/node_004
```
heifner, line 6: Seems fragile to reach in and remove this directory. The python scripts have an option to leave directories in place, this would violate that.

tests/amqp_pause_test.sh
```diff
@@ -0,0 +1,74 @@
...
+
+echo 'starting producer and sending txns'
+tests/nodeos_high_transaction_test.py -p=1 -n=8 --dump-error-details --keep-logs -v --leave-running --amqp-address=amqp://guest:guest@127.0.0.1 --num-transactions=90000000000 > var/lib/amqp.txt 2>&1 &
+sleep 5
```
heifner, line 10: Seems like you might need more time. Again the python tests have a cleaner way to wait on nodes to come up.

tests/amqp_pause_test.sh
```diff
@@ -0,0 +1,74 @@
...
+
+echo 'checking for amqp start up in main producer'
+amqp_startup=`fgrep -e "Starting amqp consumption at startup"  var/lib/node_00/stderr.txt | wc -l`
```
heifner, line 13: I don't like greping logs for log messages. Maybe we can add some functionality to the test plugin to more directly look for these conditions.

heifner (changes requested):

heifner:
Replaced by #10640

PR #10596, opened 2021-08-04
ship threaded disk write 2.2.x

huangminghuang:
## Change Description
This PR allows state history logs to be written to disk in separate threads.

port from #10504

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -17,6 +17,7 @@
 #include <fc/io/cfile.hpp>
 #include <fc/io/datastream.hpp>
 #include <fc/log/logger.hpp>
+#include <boost/asio.hpp>
 
 namespace eosio {
 
@@ -113,25 +114,17 @@ struct state_history_config {
    bfs::path archive_dir;
    uint32_t  stride             = UINT32_MAX;
    uint32_t  max_retained_files = 10;
+   uint32_t  num_buffered_entries = 2;
+   fc::logger* logger             = nullptr;
 };
 
 class state_history_log {
- private:
-   using cfile_stream        = fc::datastream<fc::cfile>;
-   const char* const    name = "";
-   cfile_stream         index;
-   uint32_t             _begin_block = 0;
-   uint32_t             _end_block   = 0;
-   chain::block_id_type last_block_id;
-   uint32_t             version = ship_current_version;
-   uint32_t             stride;
-
  protected:
-   cfile_stream write_log;
-   cfile_stream read_log;
-
-   using catalog_t = chain::log_catalog<state_history_log_data, chain::log_index<chain::state_history_exception>>;
-   catalog_t catalog;
+   using cfile_stream        = fc::datastream<fc::cfile>;
+ private:
+   const char* const    name = "";
+   uint32_t             version = ship_current_version;
+   uint32_t             stride;   
 
  public:
    // The type aliases below help to make it obvious about the meanings of member function return values.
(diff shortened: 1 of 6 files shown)


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -521,6 +521,7 @@ void state_history_plugin::plugin_initialize(const variables_map& options) {
       auto  dir_option = options.at("state-history-dir").as<bfs::path>();
 
       static eosio::state_history_config config;
```
heifner: Why is this `static`? This should be made non-static.

libraries/state_history/log.cpp
```diff
@@ -288,85 +333,155 @@ void state_history_log::split_log() {
...
+// always called from internal thread
+void state_history_log::write_entry(const chain::block_id_type& id, const chain::block_id_type& prev_id,
+                                    const std::shared_ptr<std::vector<char>>& data) {
+   state_history_log_header header{.magic = ship_magic(ship_current_version), .block_id = id};
+
+   std::unique_lock<std::mutex> lock(mx);
+   auto [block_num, start_pos] = write_entry_header(header, prev_id);
+   try {
+      lock.unlock();
+      this->write_payload(write_log, *data);
```
heifner, line 368: `write_log` is commented that it is protected by the mutex, but accessed outside of the mutex here. It looks like this is actually ok because `write_log` is only accessed from the strand. However, should probably either access with mutex held or move the declaration of `write_log` out of the section commented to be protected by the mutex.

libraries/state_history/log.cpp
```diff
@@ -288,85 +333,155 @@ void state_history_log::split_log() {
...
+   
+   if (write_thread_has_exception) {
+      std::rethrow_exception(eptr);
+   }
+   block_num_type block_num = chain::block_header::num_from_id(id);
+   auto shared_data = std::make_shared<std::vector<char>>(std::move(data));
+   boost::asio::post(work_strand, [this, id, prev_id, shared_data ]() {
+      write_entry(id, prev_id, shared_data);
+   });
+   cached[block_num] = shared_data;
```
heifner: `shared_data` can be `std::move` here to avoid one `shared_ptr` copy.

libraries/state_history/log.cpp
```diff
@@ -288,85 +333,155 @@ void state_history_log::split_log() {
...
+
+   while (cached.size() > num_buffered_entries && cached.begin()->second.use_count() == 1) {
+      // the log entry has been removed from write_queue, so we can safely erase it.
```
heifner: This comment could be a bit clearer I think. `write_queue` is an implicit queue of work_strand.

libraries/state_history/log.cpp
heifner: `last_block_id` does not need to be protected by mx.

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -140,33 +133,53 @@ class state_history_log {
...
+   void modify_entry(block_num_type block_num, Lambda f);
+
+   using cached_data_map = boost::container::flat_map<uint32_t, std::shared_ptr<std::vector<char>>>;
+   cached_data_map cached;
+   fc::logger      logger;
+
+ private:
+
+   std::thread                                                              thr;
+   uint32_t                                                                 num_buffered_entries;
```
heifner: This is always initialized in the constructor, but would be nice to initialize here for future proof.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -499,6 +498,7 @@ void state_history_plugin::set_program_options(options_description& cli, options
            "enable debug mode for trace history");
    options("context-free-data-compression", bpo::value<string>()->default_value("zlib"), 
            "compression mode for context free data in transaction traces. Supported options are \"zlib\" and \"none\"");
+   options("state-history-num-buffered-entries", bpo::value<uint32_t>()->default_value(2), "size of the buffered state history log entries in memory");
```
heifner: Instead of adding this new option now, lets just go with hard-coded value of 2 and evaluate adding an option for this later.

heifner (changes requested):

huangminghuang:
> 

## Change Description
This PR allows state history logs to be written to disk in separate threads.

port from #10504



## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case




## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other




## Consensus Changes
- [ ] Consensus Changes




## API Changes
- [ ] API Changes




## Documentation Additions
- [] Documentation Additions

heifner (approved):

PR #10601, opened 2021-08-06
correct resource monitor help text pertaining to shutting down when threshold met

spoonincode:
## Change Description

The default behavior of resource monitor is to shut down nodeos when disk usage exceeds the threshold. One must enable an option to disable that shudown.

But, the help text seems to have that behavior reversed. Wording is a little tricky here since one must enable an option to disable a feature.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
+++ b/plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
@@ -49,7 +49,7 @@ public:
          ( "resource-monitor-interval-seconds", bpo::value<uint32_t>()->default_value(def_interval_in_secs),
            "Time in seconds between two consecutive checks of resource usage. Should be between 1 and 300" )
          ( "resource-monitor-space-threshold", bpo::value<uint32_t>()->default_value(def_space_threshold),
-           "Threshold in terms of percentage of used space vs total space. If used space is above (threshold - 5%), a warning is generated.  If used space is above the threshold and resource-monitor-not-shutdown-on-threshold-exceeded is enabled, a graceful shutdown is initiated. The value should be between 6 and 99" )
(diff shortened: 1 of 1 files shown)


plugins/resource_monitor_plugin/resource_monitor_plugin.cpp
```diff
@@ -49,7 +49,7 @@ class resource_monitor_plugin_impl {
          ( "resource-monitor-interval-seconds", bpo::value<uint32_t>()->default_value(def_interval_in_secs),
            "Time in seconds between two consecutive checks of resource usage. Should be between 1 and 300" )
          ( "resource-monitor-space-threshold", bpo::value<uint32_t>()->default_value(def_space_threshold),
-           "Threshold in terms of percentage of used space vs total space. If used space is above (threshold - 5%), a warning is generated.  If used space is above the threshold and resource-monitor-not-shutdown-on-threshold-exceeded is enabled, a graceful shutdown is initiated. The value should be between 6 and 99" )
+           "Threshold in terms of percentage of used space vs total space. If used space is above (threshold - 5%), a warning is generated.  Unless resource-monitor-not-shutdown-on-threshold-exceeded is enabled, if used space is above the threshold a graceful shutdown is initiated. The value should be between 6 and 99" )
```
heifner: How about:

"Unless resource-monitor-not-shutdown-on-threshold-exceeded is enabled, a graceful shutdown is initiated if used space is above the threshold. The value should be between 6 and 99"

heifner (approved):

PR #10606, opened 2021-08-09
rodeos: Add support for stream_wapper_v1 with std::string route 📦 

heifner:
## Change Description

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/rodeos/streams/logger.hpp
+++ b/programs/rodeos/streams/logger.hpp
@@ -6,22 +6,22 @@
 namespace b1 {
 
 class logger : public stream_handler {
-   std::vector<eosio::name> routes_;
-
  public:
-   logger(std::vector<eosio::name> routes) : routes_(std::move(routes)) { ilog("logger initialized"); }
+   explicit logger(std::vector<std::string> routes)
+   : stream_handler(std::move(routes)) {
+      ilog("logger initialized");
+   }
 
-   const std::vector<eosio::name>& get_routes() const override { return routes_; }
-
-   void publish(const std::vector<char>& data, const eosio::name& routing_key) override {
-      ilog("logger stream [${data_size}] >> ${data}", ("data", data)("data_size", data.size()));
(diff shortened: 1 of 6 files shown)


programs/rodeos/streams/logger.hpp
```diff
@@ -6,22 +6,22 @@
...
+   explicit logger(std::vector<std::string> routes)
+   : stream_handler(std::move(routes)) {
+      ilog("logger initialized");
+   }
 
-   void publish(const std::vector<char>& data, const eosio::name& routing_key) override {
-      ilog("logger stream [${data_size}] >> ${data}", ("data", data)("data_size", data.size()));
+   void publish(const std::vector<char>& data, const std::string& routing_key) override {
+      ilog("logger stream ${r}: [${data_size}] >> ${data}",
+           ("r", routing_key)("data", data)("data_size", data.size()));
```
linhuang-blockone, line 17: The order of "data" and "data_size" is reversed.

heifner: The order doesn't matter.

linhuang-blockone (commented):
Just curious why the type of routes is changed from eosio::name to std::string?

linhuang-blockone (approved):

heifner:
> Just curious why the type of routes is changed from eosio::name to std::string?

eosio::name is limited to 12 characters but routing keys in rabbitmq has no such restriction. This allows users to use much more reasonable sized routing keys.

PR #10616, opened 2021-08-12
fix broken CMake in Centos8

praphael:
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/scripts/eosio_build_centos.sh
+++ b/scripts/eosio_build_centos.sh
@@ -35,58 +35,58 @@ fi
 fi
 
 if [[ "$(echo ${VERSION} | sed 's/ .*//g')" == 8 ]]; then
-        echo "Install Development Tools ..."
-        install-package https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
-        group-install-package 'Development Tools'
-        install-package openssl-devel
-        install-package which
-	install-package git 
-	install-package autoconf
-	install-package automake
-	install-package libtool
-	install-package make
-	install-package bzip2
-    	install-package graphviz
-	install-package bzip2-devel
-	install-package openssl-devel
-	install-package gmp-devel
-    	install-package file
-	install-package libusbx-devel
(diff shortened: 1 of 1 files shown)


kj4ezj (commented):
Should we fix the whitespace in this file?

On GitHub, it shows your lines have a different indentation than the others.

Screen Shot 2021-08-16 at 12 10 34 PM

I opened this file in my editor, and it looks different there than it does online.

Screen Shot 2021-08-16 at 12 10 00 PM

Your lines look like they are indented correctly in VScode, but I found that this file has some lines with spaces, some lines with tabs, and some lines with spaces **_and_** tabs on the same line! Abhorrent.

I know this isn't your fault and you didn't introduce this problem with your changes but, since this is a public repo, we shouldn't ignore problems like this in order to make sure our code is readable and approachable to our community.

I, personally, prefer tabs so people can set their own indentation (blasphemous, I know), but we pretty much standardized on four-space indentation for BASH scripts across the EOSIO organization. I don't care which, but it should be consistent.

kj4ezj (approved):
The whitespace in this file is still inconsistent. I highlighted the tabs in yellow in this screenshot for emphasis:

Screen Shot 2021-08-18 at 1 08 37 PM

I approved this because I don't want to be disrespectful of your time, but I still don't think we should ignore this if we can. This one-liner will replace all remaining tabs with spaces:
```bash
sed -i 's/\t/    /g' ./scripts/eosio_build_centos.sh
```
Finally, the code block from lines 16 to 34 should be indented four spaces.

PR #10619, opened 2021-08-13
Merge amqp_trx_plugin from develop-box 

huangminghuang:
## Change Description
Merge amqp_trx_plugin related changes from develop-box

* #10615 :
    #10541 added support for starting up nodeos not consuming from AMQP queue. This change however does not start consuming again when the AMQP connection is loss and re-established. This PR adds back in start of consuming when the AMQP connection is re-established.
* #10541 :
* #10412 :
    - Add support to AMQP to specify reply-to.
    - Add amqp-reply-to option to cleos
* #10371 :
    amqp_trx_plugin
    - Add block_uuid to transaction trace message
    - Add block_uuid_message for signaling correlation between block_uuid and block_id.
      - Sent only to routing keys specified by AMQP Header "block-uuid-msg". Add a AMQP Header of "block-uuid-msg" with a string value of the routing key to send the block_uuid_message. If not specified then no block_uuid_message will be sent.
    - Add ABI for amqp_trx_plugin messages. SHiP protocol with addition of amqp_trace_plugin types
* #10334 :
    - The immediate ack on trx failure was only ack'ing on exception but not on a trace with an exception. Now amqp_trx_plugin will ack on trx failure for all failed transactions.
    - Also moved the sending of the trace after the ack although no guarantee is provided that it will actually happen first as it is async.
* #10316 :
    Set correlation_id of the amqp trace message to correspond to the correlation_id of the incoming amqp trx message. For amqp_trace_plugin amqp trace messages do not set correlation_id since it does not correspond to an incoming amqp message. This matches with amqp intended use case and matches clients expected behavior. Now clients can set correlation id to whatever they desire, which of course could include transaction id.

* #10162 :
* #10141 :
* #10127 :
    Add reliability options to `amqp_trx_plugin` & `amqp_trace_plugin`

    - `amqp_trace_plugin` now has configurable reliability via  `amqp-trace-reliable-mode`
      - Reliable mode 'exit', exit application on any AMQP publish error.
      - Reliable mode 'queue', queue transaction traces to send to AMQP on connection establishment.
      - Reliable mode 'log', log an error and drop message when unable to directly publish to AMQP.
    - `amqp_trace_plugin` now has configurable queue via `amqp-trace-queue-name` option.
    - `amqp_trx_plugin` now has configurable queue via `amqp-trx-queue-name` option.
    - Added correlation_id and error callback support to `reliable_amqp_publisher`.
    - `reliable_amqp_publisher` now calls `on_fatal_error` callback instead of dropping messages which in all cases calls `app().quit()` now.
    - Modified `amqp_handler` to use `single_channel_retrying_amqp_connection` to provide reliability and to move away from unsupported `AMQP::LibBoostAsioHandler`.

    Note: A follow-on PR is planned that will separate `amqp_trace_plugin` from `amqp_trx_plugin`.

* #9594 :
    - `develop-boxed` version of #9593
      - Has the additional line for `execute_incoming_transaction` calling `restart_speculative_block` for the `amqp_trx_plugin` not in `develop`.
* #9872 :

* #9294 :

* #9303 :
* #9256 :
    - The new `amqp_trx` logger is no longer used as the `amqp_trx_plugin` & `amqp_trace_plugin` use the default logger.
* #9181 :

    - `amqp_trx_plugin`

    Binary `fc::pack` format:
    ```
    // consume message types
    ```

    - `amqp_trace_plugin`

    Binary `fc::pack` format:
    ```
    // publish message types
       std::string error_message;
    };
    ```

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions


.cicd/platforms/pinned/centos-7.7-pinned.dockerfile
```diff
@@ -8,10 +8,19 @@ RUN yum update -y && \
...
 # requests module. used by tests
 RUN source /opt/rh/rh-python36/enable && python -m pip install requests
+# install erlang and rabbitmq
+RUN curl -s https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | bash && \
+    yum install -y erlang
+RUN curl -s https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh | bash && \
+    yum install -y rabbitmq-server
+# upgrade pip installation
+RUN . /opt/rh/rh-python36/enable && \
+    pip install --upgrade pip
```
heifner: The pip is for postgres, and is already above. I think these 3 lines can be removed.

CMakeLists.txt
```diff
@@ -127,6 +127,7 @@ else()
 endif()
 
 option(EOSIO_ENABLE_DEVELOPER_OPTIONS "enable developer options for EOSIO" OFF)
+option(EOSIO_REQUIRE_FULL_VALIDATION "remove runtime options allowing light validation" OFF)
```
heifner, line 130: Update PR description to indicate this PR pulls in `EOSIO_REQUIRE_FULL_VALIDATION` option.

libraries/chain/CMakeLists.txt
```diff
@@ -162,6 +162,25 @@ if(EOSVMOC_ENABLE_DEVELOPER_OPTIONS)
...
+      string(LENGTH ${SHA256_REGEX} SHA256_REGEX_STRLEN)
+   endwhile()
+
+   string(REGEX MATCH "^${SHA256_REGEX}$" IS_SHA256 "${EOSIO_REQUIRE_CHAIN_ID}")
+   if(NOT IS_SHA256)
+      message(FATAL_ERROR "EOSIO_REQUIRE_CHAIN_ID option doesn't look like a valid sha256")
+   endif()
+
+   target_compile_definitions(eosio_chain PUBLIC "EOSIO_REQUIRE_CHAIN_ID=\"${EOSIO_REQUIRE_CHAIN_ID}\"")
+endif()
```
heifner, line 182: Update PR description to indicate this PR also pulls in `EOSIO_REQUIRE_CHAIN_ID` option.

heifner (changes requested):

heifner (approved):
I think we need to update the PR description with information from all the merged PRs. This will be helpful for when users are reviewing the release notes and for doc team when creating documentation around this PR.

PR #10590, opened 2021-08-02
net plugin bugfix for nodes getting out of sync

dimas1185:
## Change Description

Fixes some race condition case in `net_plugin` where nodes go out of sync.
Test failure example:
https://buildkite.com/EOSIO/eosio-build-unpinned/builds/7688#50bec0fc-3b21-4b74-8976-e99039bfd529

Attached image with diagram explaining why nodes went out of sync in this particular test.
test issue

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #10582, opened 2021-07-28
Rodeos filter current_time() 📦

heifner:
## Change Description

- Rodeos previously only supported `current_time()` host function for query wasms. This PR adds support for `current_time()` in filter wasms.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #10549, opened 2021-07-20
Stop rodeos if unable to listen on wql-listen

heifner:
## Change Description

- Do not let rodeos continue if unable to listen on the provided `wql-listen` address.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #10529, opened 2021-07-16
Make undo_stack configurable and add quit function for exception hand…

linhuang-blockone:
…ling

## Change Description

Add `--disable-undo-stack` to `rodoes` startup option to disable undo_stack handling.

Gracefully shut down `rodeos` when a fork happens if undo stack is not enabled.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [x] Other

Manual testing.

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #10625, opened 2021-08-17
merge rodeos and eosio-tester

huangminghuang:
## Change Description

This PR merges all rodeos and eosio-tester related changes from develop-boxed
* #10620:
    backport of net plugin stop synching bugfixes:
        #10590
       #10390
    And new bugfix for node unexpectedly stopping syncing.
* #10618:
    - Update `amqp_trx_plugin` to not declare queue. Instead it will assume queue exists.
    - Add `--amqp-queue-name` to cleos to allow user specification of queue instead of hard-coded "trx"
    - EPE-1190, EPE-1192
* #10611 :
    - Fix issue with disconnect of rodeos causing nodeos to core-dump.
    - Fix issue with head->block not valid on a request to nodeos. For example, a producer node that is not producing.
* #10606 :
* #10598 :
    Add support for `recover_key` in `eosio-tester`. Requires https://github.com/EOSIO/eosio.cdt/pull/1185
* #10590:
    Fixes some race condition case in `net_plugin` where nodes go out of sync.
    Test failure example:
       https://buildkite.com/EOSIO/eosio-build-unpinned/builds/7688#50bec0fc-3b21-4b74-8976-e99039bfd529

    Attached image with diagram explaining why nodes went out of sync in this particular test.
    test issue
* #10582 :
    - Rodeos previously only supported `current_time()` host function for query wasms. This PR adds support for `current_time()` in filter wasms.
* #10575 :
    - Remove unused receiver and rename used receiver so it is clearer what it represents. No change in functionality just makes the code a little easier to follow.
* #10561 :
    Update to the force-write-stride to 100,000
* #10557 :
    Publish the tested options file for example. Add max_write_buffer_number if options file is not present.
* #10550 :
    - When many rodeos connected it would overwhelm nodeos. Post to main thread at `medium` so that other `medium` or higher tasks have an opportunity to run.
* #10549 :
    - Do not let rodeos continue if unable to listen on the provided `wql-listen` address.
* #10548 :
    Init `force_write_stride` because it is used in modulus to avoid divide by 0
* #10547 :
* #10529 :
    Add `--disable-undo-stack` to `rodoes` startup option to disable undo_stack handling.

    Gracefully shut down `rodeos` when a fork happens if undo stack is not enabled.
* #10508 :
* #10503 :
    Add RocksDB options file support and adjust default configurations.  Options file is specified by command line argument `--rdb-options-file`. Please see Documentation Additions below for details.
* #10519 :
    When `rodeos`'  `ship_client` fails to connect to `nodeos` (like when `nodeos` is not running),. `ship_client` calls `cloner_session`'s `close`() with `retry` set to `true`. This triggers `cloner_plugin` to recreate `cloner_session` every second until a connection is established. `cloner_session` creates `rodeos_filter` object if `rodeos` starts with `--filter-wasm` option. Although  destructor of `rodeos_filter` is called before recreation, its memory is not completely freed. The root cause is the `filter_state` in `rodeos_filter` has a member `eosio::vm::wasm_allocator  wa`. `wa` allocates VM memory when initiated. This memory must be released by explicit `free()`.

    The same problem in `thread_state` is fixed by this PR as well.
* #10456 :
    - Add host functions `recover_key` & `assert_recover_key` to rodeos for use in queries
* #10390:
    Fixed test case that required node restart in `privacy_scenario_3_test.py`

    Scenario:
    after `node1` launched it sends handshake to `node2`
    `node1` is not in security group so `node2` doesn't send back any notice (lib catchup notice was required in that case)
    `node1` was added to security group and `node2` sends to `node1` handshake
    in `sync 1` scenario, when `net_plugin` receives non-first handshake with lib greater its own lib it do nothing. 
    This fix adds sending handshake there also in case of non-first handshake generation
* #10382 :
    The `eosio-tester` program would only return an exit code of 1 (to indicate a problem) if it did not attempt to execute the test WASM (e.g. due to running help or a problem parsing arguments) or if it did attempt to execute the test WASM but failed to return successfully because of some exception (including exceptions caused by executing `eosio_assert_message`). Otherwise, it would return an exit code of 0 (indicating success).

    This works fine for test cases that would check for failures using (directly or indirectly) `eosio_assert_message`, for example: calling a function which calls `eosio::check` and the condition fails; or using `REQUIRE` in Catch2. But in other situations this would mask failures. For example, when using `CHECK` in Catch2, the errors are printed out and then ultimately the failure of the test run is indicated through the exit code returned by the `main` function. So in such failures it would appear looking at only the exit code of running `eosio-tester` (which is what `ctest` does) that the tests all passed.

    This PR checks to see if a return value exists for the `start` function that is called when the test WASM is executed. If so, it assumes it is a signed 32-bit integer, extracts that value and returns that as the exit code. If it doesn't exist, it will continue with the current behavior of returning 0.

    Note to fully take advantage of this change, the test WASMs must be compiled with a `start` function that passes through the return value of `main`. For that change, see: https://github.com/EOSIO/eosio.cdt/pull/1101.
* #10252 :

    ```
    if(! impl_->read.pending)
        BOOST_VERIFY(
                expiry_time) == 0);
    if(! impl_->write.pending)
        BOOST_VERIFY(
                expiry_time) == 0);
    ```
    and the comment from the code

    ```
    */
    ```

* #10164 :
* #10097 :
    - Add missing get/set param methods.
* #10083 :
    cleos needs this to send transaction to rodeos.
* #10049 :

* #10057 :
    - Use stable SHiP protocol for transaction_trace
    - Move eosio-tester type conversion to state_history library
* #10024 :
    - Pull in latest of eos-vm
* #10001 :
    Add support for rodeos connecting to state history endpoint via unix socket (as exposed via #9963). Also add support for the HTTP QL server to listen on a unix socket. Like `http-server-address` and `state-history-endpoint`, `wql-listen` has been modified such that if it's configured with the empty string the TCP endpoint is completely disabled.
* #9963 :
    Oftentimes clients consuming from nodeos' state history endpoint fully trust the data received from this endpoint. It certainly doesn't have to be this way -- a client connected to the state history endpoint could perform light validation, for example. But oftentimes that is not the case; even rodeos fully trusts the state history endpoint it connects to.

    This change adds a new option to state_history_plugin to expose its endpoint over a unix socket in addition or in lieu of the existing TCP endpoint. In some secure environments this allows more tight coupling of nodeos and rodeos (or other ship consumers) to enforce that the state history client always connects to the expected (trustworthy) nodeos.

    The `state-history-endpoint` option has been changed to behave similar to `http-server-address`: if `state-history-endpoint` is configured as an empty string, the TCP endpoint is fully disabled.
* #9708 :
    rodeos: allow queries to use sha functions
* #9307 :
* #9331 :
    - Get future before promise use to avoid race on `set_value` & `get_future`
* #9303 :
* #9234 :
    rodeos can now generate checkpoints (live backup of database) using `/v1/rodeos/create_checkpoint` if `--wql-checkpoint-dir` is enabled.
* #9241 :
    Make rodeos query wasm memory limit configurable
* #9205 :
    It appears that rodeos running on a certain k8s cluster is having trouble using mmap. This change:
    * Increases diagnostics to help narrow in on the problem
    * Strengthens the wasm mmap cache so it doesn't dispose (and later have to recreate) regions when other errors occur.
    * Preallocates wasm mmaps ahead of time

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


programs/CMakeLists.txt
```diff
@@ -5,3 +5,4 @@ add_subdirectory( eosio-launcher )
 add_subdirectory( eosio-blocklog )
 add_subdirectory( nodeos-sectl )
 add_subdirectory( rodeos )
+add_subdirectory( eosio-tester )
```
linhuang-blockone: Can you make it optional in the same way for rodeos? Default is not to build it.

linhuang-blockone (approved):
I have reviewed rodeos and chain_kv. They are fine. This needs to be tested extensively after merge.

PR #10629, opened 2021-08-18
Ensure that all stability test rounds continue even if early rounds exhibit failures.

williamblevins:
## Change Description

Ensure that all stability test rounds continue even if early rounds exhibit failures.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [X] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/generate-pipeline.sh
+++ b/.cicd/generate-pipeline.sh
@@ -484,7 +484,7 @@ EOF
         fi
         if [[ "$ROUND" != "$ROUNDS" ]]; then
             echo '  - wait: ~'
-            echo '    continue_on_failure: true'
+            [[ "$BUILDKITE_SOURCE" == 'schedule' || "$CONTINUE_ON_FAILURE" == 'true' ]] && echo '    continue_on_failure: true'
             echo ''
         fi
     done
--- a/.cicd/generate-pipeline.sh
+++ b/.cicd/generate-pipeline.sh
@@ -483,7 +483,8 @@ EOF
 EOF
         fi
         if [[ "$ROUND" != "$ROUNDS" ]]; then
-            echo '  - wait'
+            echo '  - wait: ~'
+            echo '    continue_on_failure: true'
             echo ''
         fi
     done
@@ -551,7 +552,7 @@ EOF
 fi
 # pipeline tail
 cat <<EOF
-  - wait:
+  - wait: ~
     continue_on_failure: true
 
   - label: ":bar_chart: Test Metrics"
--- a/.cicd/eosio-test-stability.md
+++ b/.cicd/eosio-test-stability.md
@@ -17,12 +17,15 @@ The [eosio-test-stability](https://buildkite.com/EOSIO/eosio-test-stability) pip
 ### Variables
 There are five primary environment variables relevant to stability testing:
 ```bash
-PINNED='true|false'    # whether to perform the test with pinned dependencies, or default dependencies
-ROUNDS='ℕ'             # natural number defining the number of gated rounds of tests to generate
-ROUND_SIZE='ℕ'         # number of test steps to generate per operating system, per round
-SKIP_MAC='true|false'  # conserve finite macOS Anka agents by excluding them from your testing
-TEST='name'            # PCRE expression defining the tests to run, preceded by '^' and followed by '$'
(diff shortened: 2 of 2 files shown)


sandie06 (approved):

.cicd/generate-pipeline.sh
```diff
@@ -483,7 +483,8 @@ EOF
 EOF
         fi
         if [[ "$ROUND" != "$ROUNDS" ]]; then
-            echo '  - wait'
+            echo '  - wait: ~'
+            echo '    continue_on_failure: true'
             echo ''
         fi
```
kj4ezj: From [AUTO-1236](https://blockone.atlassian.net/browse/AUTO-1236) (emphasis mine):
> This story expands the functionality of the pipeline code to support this by setting `continue_on_failure: true` on the `wait` steps in the auto-generated `pipeline.yml` for **_scheduled builds_** in the [eosio-test-stability](https://buildkite.com/EOSIO/eosio-test-stability) pipeline.
>
> **_It is imperative this flag is never set in [...] manual eosio-test-stability runs_**, or any other EOSIO pipelines as this will waste resources and cause false-negative (type 2) error with respect to build outcomes and pull request checks.

In this case, the `continue_on_failure` flag will be set for all test stability runs, including manual ones. I think something like this would work:
```bash
if [[ "$ROUND" != "$ROUNDS" ]]; then
    echo '  - wait'
    [[ "$BUILDKITE_SOURCE" == 'schedule' ]] && echo '    continue_on_failure: true'
    echo ''
fi
```
Alternatively, you could provide an override for anyone who _does_ want manual builds to run all steps regardless of outcome.
```bash
if [[ "$ROUND" != "$ROUNDS" ]]; then
    echo '  - wait'
    [[ "$BUILDKITE_SOURCE" == 'schedule' || "$CONTINUE_ON_FAILURE" == 'true' ]] && echo '    continue_on_failure: true'
    echo ''
fi
```
I certainly do not think `continue_on_failure` should be the default because the manual runs are only used to prove stability. A single failure shows instability, so if we exit right away then this shortens test times by hours and saves a bunch of unnecessary cloud spend.

.cicd/generate-pipeline.sh
```diff
@@ -551,7 +552,7 @@ EOF
 fi
 # pipeline tail
 cat <<EOF
-  - wait:
+  - wait: ~
```
kj4ezj, line 555: What does the tilde (`~`) do here?

kj4ezj (changes requested):

.cicd/generate-pipeline.sh
kj4ezj: If you really want to follow the ticket to the letter, you could do something like this...
```bash
if [[ "$ROUND" != "$ROUNDS" ]]; then
    echo '  - wait'
    [[ "$BUILDKITE_SOURCE" == 'schedule' && "$BUILDKITE_PIPELINE_SLUG" == 'eosio-test-stability' ]] && echo '    continue_on_failure: true'
    echo ''
fi
```
...but I no longer believe that is necessary because anyone doing multiple rounds in a scheduled build on another EOSIO pipeline is doing it intentionally and knows what they are doing.

williamblevins: I think the second one is probably better.

.cicd/generate-pipeline.sh
williamblevins: That is the YAML null symbol. In this case, it enforces that you cannot add information at the end.

https://buildkite.com/docs/pipelines/wait-step

kj4ezj: > The explicit null `~` character used in the above examples isn't required, but is recommended as a best practice. It ensures that nothing else is accidentally added to the `wait` before the `continue_on_failure` attribute.

Hmmm, interesting. I had never seen this before. Thank you.

kj4ezj (approved):
🍕

PR #10635, opened 2021-08-18
Cleos support for set<optional<T>> and vector<optional<T>>

shahabi8:
## Change Description
This PR is adding support for nested containers `set<optional<T>>` and `vector<optional<T>>`. Python tests for KV and multi-index are also added to verify the solution of this PR. The tests examples are from this nested container [repo](https://github.com/EOSIO/nested-container). More details on how to use nested containers can be found there.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [X] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [X] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

Two Python integration tests for KV and multi-index tables are added in this PR for nested container. These tests illustrate how to pass in nested container data through `cleos` commands as well as they verify which combination of containers are supported in `EOSIO`.


plroblox:
Please describe briefly the new tests you have added underneath 'Documentation Additions'

tests/nested_container_multi_index_test.py
```diff
@@ -0,0 +1,321 @@
...
+    node.publishContract(MIacct, contractDir, wasmFile, abiFile, waitForTransBlock=True)
+
+    Print("Test action for set< set< uint16_t >>")
+    create_action('setstst', '["alice", [[10, 10], [3], [400, 500, 600]]]', 'nestcontnmi', 'alice')
+
+    Print("Test action for set< vector< uint16_t >>")
+    create_action('setstv', '["alice", [[16, 26], [36], [36], [46, 506, 606]]]', 'nestcontnmi', 'alice')
+
+    Print("Test action for set< optional< uint16_t >>")
+    create_action('setsto', '["alice", [null, null, 500]]', 'nestcontnmi', 'alice')
```
plroblox, line 130: use [null, 500] instead of [null, null,500] for the setsto, because std::set<T> is supposed to have unique (different) elements

tests/nested_container_kv_test.py
```diff
@@ -0,0 +1,320 @@
...
+    node.publishContract(kvacct, contractDir, wasmFile, abiFile, waitForTransBlock=True)
+
+    Print("Test action for set< set< uint16_t >>")
+    create_action('setstst', '[1, [[10, 10], [3], [400, 500, 600]]]', 'nestcontn2kv', 'alice')
+
+    Print("Test action for set< vector< uint16_t >>")
+    create_action('setstv', '[1, [[16, 26], [36], [36], [46, 506, 606]]]', 'nestcontn2kv', 'alice')
+
+    Print("Test action for set< optional< uint16_t >>")
+    create_action('setsto', '[1, [null, null, 500]]', 'nestcontn2kv', 'alice')
```
plroblox, line 129: better change it to [null,50], since std::set<T> is supposed to have unique elements

tests/nested_container_kv_test.py
```diff
@@ -0,0 +1,320 @@
...
+    Print("Test action for vector<optional<mystruct>>")
+    create_action('setvos', '[1, [{"_count": 18, "_strID": "dumstr"}, null, {"_count": 19, "_strID": "dumstr"}]]', 'nestcontn2kv', 'alice')
+
+    Print("Test action for pair<uint16_t, vector<optional<uint16_t>>>")
+    create_action('setpvo', '[1,{"first": 183, "second":[100, null, 200]}]', 'nestcontn2kv', 'alice')
+
+    cmd="get kv_table %s people2kv map.index" % kvacct.name
+    transaction = node.processCleosCmd(cmd, cmd, False, returnType=ReturnType.raw)
+    transaction_json = json.loads(transaction)
+
```
plroblox, line 256: This part is great !

libraries/chain/abi_serializer.cpp
```diff
@@ -389,10 +389,6 @@ namespace eosio { namespace chain {
          for( decltype(size.value) i = 0; i < size; ++i ) {
             ctx.set_array_index_of_path_back(i);
             auto v = _binary_to_variant(ftype, stream, ctx);
-            // QUESTION: Is it actually desired behavior to require the returned variant to not be null?
-            //           This would disallow arrays of optionals in general (though if all optionals in the array were present it would be allowed).
-            //           Is there any scenario in which the returned variant would be null other than in the case of an empty optional?
-            EOS_ASSERT( !v.is_null(), unpack_exception, "Invalid packed array '${p}'", ("p", ctx.get_path_string()) );
```
plroblox, line 395: It would be good to add a comment here indicating arrays of optionals are allowed..., also this chain/abi_serializer.cpp is related to consensus, I believe the change will most probably affect user contracts only, but please consult with Areg @arhag to make sure the code change will NOT affect consensus

unittests/abi_tests.cpp
```diff
@@ -3252,8 +3252,11 @@ BOOST_AUTO_TEST_CASE(abi_deserialize_detailed_error_messages)
       ],
    })";
 
+   // Test to verify that array of optinal doesn't throw exception
+   abi_serializer abis( fc::json::from_string(abi).as<abi_def>(), abi_serializer::create_yield_function( max_serialization_time ) );
+   BOOST_CHECK_NO_THROW( abis.binary_to_variant("s4", fc::variant("030101000103").as<bytes>(), abi_serializer::create_yield_function( max_serialization_time )) );
```
plroblox, line 3259: It would be good to comment fc::variant("030101000103") represents an array of std::optional<int8> {1,null,3}, you can also add say fc::variant("0400000000") represents an array of 4 nulls, and fc::variant("030001af013a") represents {null, 0xAF, 0x3A} to give a little documentation about what's happening here:)

tests/nested_container_multi_index_test.py
shahabi8: I added this example to check that `set` is also working properly in removing non-unique components and return this result: `[None, 500]`

tests/nested_container_kv_test.py
shahabi8: Same as above.

libraries/chain/abi_serializer.cpp
shahabi8: Done.

unittests/abi_tests.cpp
shahabi8: Done.

libraries/chain/abi_serializer.cpp
heifner: chain/abi_serializer is not part of consensus.

plroblox: Thanks, kevin!

plroblox (approved):

PR #10645, opened 2021-08-20
Reset sync_known_lib_num when closing a connection 📦

heifner:
## Change Description

- Reset sync_known_lib_num on closing connection to optimize syncing
- When closing connection a connection we should no longer consider its view of LIB since we are no longer connected to that peer. This PR updates sync_known_lib_num with highest of known peers when a connection is closed.
- Fixed some cases where net_plugin kept trying to re-connect on fatal conditions (like chain is a differ chain (chain_id))

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1125,10 +1121,10 @@ namespace eosio {
             c->last_handshake_sent.generation = ++c->sent_handshake_count;
             auto last_handshake_sent = c->last_handshake_sent;
             g_conn.unlock();
-            fc_ilog( logger, "Sending handshake generation ${g} to ${ep}, lib ${lib}, head ${head}, id ${id}",
-                     ("g", last_handshake_sent.generation)( "ep", c->peer_name())
-                     ("lib", last_handshake_sent.last_irreversible_block_num)
-                     ("head", last_handshake_sent.head_num)("id", last_handshake_sent.head_id.str().substr(8,16)) );
+            peer_ilog( c, "Sending handshake generation ${g}, lib ${lib}, head ${head}, id ${id}",
```
dimas1185: peer_[x]log internally using `conn_mtx` lock which is a bit counterintuitive (one doesn't expect log macro to use class mutex) . In this particular example I see that just before calling peer_ilog `g_conn.unlock` was called. However it looks to me dangerous to rely on this. People can introduce deadlock accidentally.
Documentation says there is a mutex that can be locked inside same thread many times:
`std::recursive_mutex`. Maybe we can switch to it? That way we will protect developers from unintentional causing a deadlock.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1614,20 +1614,32 @@ namespace eosio {
...
-      if( c->current() ) {
+      if( !closing && c->current() ) {
          std::lock_guard<std::mutex> g_conn( c->conn_mtx );
          if( c->last_handshake_recv.last_irreversible_block_num > sync_known_lib_num ) {
             sync_known_lib_num = c->last_handshake_recv.last_irreversible_block_num;
          }
-      } else if( c == sync_source ) {
-         sync_last_requested_num = 0;
-         request_next_chunk( std::move(g) );
+      } else if ( closing ) {
```
dimas1185, line 1629: can you elaborate this idea with updating `sync_known_lib_num`, please? perhaps in PR description.
From first look I don't get why we are requesting now next chunk only if connection is closing and it was a sync source.

tests/large-lib-test.py
```diff
@@ -0,0 +1,134 @@
...
+    speculativeNode1=cluster.getNode(1)
+    speculativeNode2=cluster.getNode(2)
+
+    Print ("Wait for Cluster stabilization")
+    if not cluster.waitOnClusterBlockNumSync(3):
+        errorExit("Cluster never stabilized")
+    Print ("Cluster stabilized")
+
+    Print("Wait for producing {} blocks".format(numBlocksToProduceBeforeRelaunch))
+    time.sleep(numBlocksToProduceBeforeRelaunch/2)
```
dimas1185: there is a method `Node.waitForBlock` for waiting N blocks.
I think we should use that one and not make assumptions here.

plugins/net_plugin/net_plugin.cpp
heifner: Google "recursive mutex considered harmful" and enjoy some reading. ;)

tests/large-lib-test.py
```diff
@@ -0,0 +1,134 @@
...
+    specificExtraNodeosArgs[1]="--read-mode speculative "
+    specificExtraNodeosArgs[2]="--read-mode speculative "
+
+    Print("Stand up cluster")
+    if cluster.launch(
+            pnodes=pnodes,
+            totalNodes=total_nodes,
+            totalProducers=1,
+            useBiosBootFile=False,
+            topo="mesh",
```
dimas1185, line 76: FYI, mesh is default

plugins/net_plugin/net_plugin.cpp
heifner: That said, I was thinking about requiring the `peer_[x]log` to take a mutex so it was obvious. Something like:
`peer_ilog( c, c->conn_mtx, "Sending ...." );` What do you think?

tests/large-lib-test.py
heifner: I know, but I like it being explicit since it is important to the test.

tests/large-lib-test.py
heifner: Thanks, I like that a lot better.

plugins/net_plugin/net_plugin.cpp
heifner: I'll add a comment here, I agree it is not clear.

plugins/net_plugin/net_plugin.cpp
dimas1185: ok, will do that, but still I don't like the fact that `peer_[x]log` now becomes in fact default log macro and now we have to release `conn_mtx` every time we want to log some data. Normally I'm risky but in that case it would be nice to have any sort of protection or warning. I can imagine debugging stuck threads in `net_plugin` ...
Any ideas?

heifner: I agree, hence why I was thinking about passing in the mutex. I have idea of how to remove the mutex, let me give it a try and see how it goes.

heifner: I think we can restrict all `peer_[x]log` to connection strand and even assert that. Then it should be possible to have access to all connection strand data without a mutex. However, that I think should be a separate PR. I'll revert this commit and create a new PR with just the logging changes.

heifner: `peer_name()` mutex should be removed at the same time.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1614,20 +1615,38 @@ namespace eosio {
...
+         sync_known_lib_num = highest_lib_num;
+
+         // if closing the connection we are currently syncing from, then reset our last requested and next expected.
+         if( c == sync_source ) {
+            sync_last_requested_num = 0;
+            uint32_t head_blk_num = 0;
+            std::tie( std::ignore, head_blk_num, std::ignore, std::ignore, std::ignore, std::ignore ) = my_impl->get_chain_info();
+            sync_next_expected_num = head_blk_num + 1;
+            request_next_chunk( std::move(g) );
+         }
```
dimas1185, line 1649: so technically there is a hypothetical `else` where `c` is not current and we are not closing.
Is this your intention to do nothing on that condition? If so I would like to see that in comment.
It looks like previously we were requesting next chunk anyway.

heifner: Previously `closing` was implicit. `c` would not be `current()` if `closing==true`.

tests/large-lib-test.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Print("Kill all node instances.")
+    for clusterNode in cluster.nodes:
+        clusterNode.kill(signal.SIGTERM)
+    cluster.biosNode.kill(signal.SIGTERM)
+    Print("All nodeos instances killed.")
+
+    # Remove both state and blocks such that no replay happens
+    Print("Remove producer node's state and blocks directories")
+    shutil.rmtree(Utils.getNodeDataDir(0,"state"))
```
dimas1185: if you would do any changes on that PR I suggest we will move that part to `Node.py`.
This looks generic functionality. It could look like that:
`specificNode.cleanDataDir(cleanBlocks=True, cleanState=True)`
`cleanBlocks` and `cleanState` can be defaulted to True
Not a big deal but if any changes to be done I think it worth to move to framework.

plugins/net_plugin/net_plugin.cpp
dimas1185: ok, in that case can you also add `else {}` with warning/error log that it supposed to be unreachable. Just in case if this condition will arise (e.g. as a result of future changes) we will see this in logs

tests/large-lib-test.py
heifner: Moved

plugins/net_plugin/net_plugin.cpp
heifner: Added log statement

dimas1185 (approved):

PR #10456, opened 2021-06-24
Add recover_key & assert_recover_key to rodeos

heifner:
## Change Description

- Add host functions `recover_key` & `assert_recover_key` to rodeos for use in queries

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [x] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

PR #10657, opened 2021-08-25
merge changes from 2.2.x into develop

huangminghuang:
## Change Description

This PR merges changes related to state_history_plugin amqp-trx-plugin, rodeos, eosio-tester . 

* #10618:
    - Update `amqp_trx_plugin` to not declare queue. Instead it will assume queue exists.
    - Add `--amqp-queue-name` to cleos to allow user specification of queue instead of hard-coded "trx"
    - EPE-1190, EPE-1192
* #10611 :
    - Fix issue with disconnect of rodeos causing nodeos to core-dump.
    - Fix issue with head->block not valid on a request to nodeos. For example, a producer node that is not producing.
* #10606 :
* #10598 :
    Add support for `recover_key` in `eosio-tester`. Requires https://github.com/EOSIO/eosio.cdt/pull/1185
* #10582 :
    - Rodeos previously only supported `current_time()` host function for query wasms. This PR adds support for `current_time()` in filter wasms.
* #10575 :
    - Remove unused receiver and rename used receiver so it is clearer what it represents. No change in functionality just makes the code a little easier to follow.
* #10561 :
    Update to the force-write-stride to 100,000
* #10557 :
    Publish the tested options file for example. Add max_write_buffer_number if options file is not present.
* #10550 :
    - When many rodeos connected it would overwhelm nodeos. Post to main thread at `medium` so that other `medium` or higher tasks have an opportunity to run.
* #10549 :
    - Do not let rodeos continue if unable to listen on the provided `wql-listen` address.
* #10548 :
    Init `force_write_stride` because it is used in modulus to avoid divide by 0
* #10547 :
* #10529 :
    Add `--disable-undo-stack` to `rodoes` startup option to disable undo_stack handling.

    Gracefully shut down `rodeos` when a fork happens if undo stack is not enabled.
* #10508 :
* #10503 :
    Add RocksDB options file support and adjust default configurations.  Options file is specified by command line argument `--rdb-options-file`. Please see Documentation Additions below for details.
* #10519 :
    When `rodeos`'  `ship_client` fails to connect to `nodeos` (like when `nodeos` is not running),. `ship_client` calls `cloner_session`'s `close`() with `retry` set to `true`. This triggers `cloner_plugin` to recreate `cloner_session` every second until a connection is established. `cloner_session` creates `rodeos_filter` object if `rodeos` starts with `--filter-wasm` option. Although  destructor of `rodeos_filter` is called before recreation, its memory is not completely freed. The root cause is the `filter_state` in `rodeos_filter` has a member `eosio::vm::wasm_allocator  wa`. `wa` allocates VM memory when initiated. This memory must be released by explicit `free()`.

    The same problem in `thread_state` is fixed by this PR as well.
* #10456 :
    - Add host functions `recover_key` & `assert_recover_key` to rodeos for use in queries
* #10382 :
    The `eosio-tester` program would only return an exit code of 1 (to indicate a problem) if it did not attempt to execute the test WASM (e.g. due to running help or a problem parsing arguments) or if it did attempt to execute the test WASM but failed to return successfully because of some exception (including exceptions caused by executing `eosio_assert_message`). Otherwise, it would return an exit code of 0 (indicating success).

    This works fine for test cases that would check for failures using (directly or indirectly) `eosio_assert_message`, for example: calling a function which calls `eosio::check` and the condition fails; or using `REQUIRE` in Catch2. But in other situations this would mask failures. For example, when using `CHECK` in Catch2, the errors are printed out and then ultimately the failure of the test run is indicated through the exit code returned by the `main` function. So in such failures it would appear looking at only the exit code of running `eosio-tester` (which is what `ctest` does) that the tests all passed.

    This PR checks to see if a return value exists for the `start` function that is called when the test WASM is executed. If so, it assumes it is a signed 32-bit integer, extracts that value and returns that as the exit code. If it doesn't exist, it will continue with the current behavior of returning 0.

    Note to fully take advantage of this change, the test WASMs must be compiled with a `start` function that passes through the return value of `main`. For that change, see: https://github.com/EOSIO/eosio.cdt/pull/1101.
* #10252 :

    ```
    if(! impl_->read.pending)
        BOOST_VERIFY(
                expiry_time) == 0);
    if(! impl_->write.pending)
        BOOST_VERIFY(
                expiry_time) == 0);
    ```
    and the comment from the code

    ```
    */
    ```

* #10164 :
* #10097 :
    - Add missing get/set param methods.
* #10083 :
    cleos needs this to send transaction to rodeos.
* #10049 :

* #10057 :
    - Use stable SHiP protocol for transaction_trace
    - Move eosio-tester type conversion to state_history library
* #10024 :
    - Pull in latest of eos-vm
* #10001 :
    Add support for rodeos connecting to state history endpoint via unix socket (as exposed via #9963). Also add support for the HTTP QL server to listen on a unix socket. Like `http-server-address` and `state-history-endpoint`, `wql-listen` has been modified such that if it's configured with the empty string the TCP endpoint is completely disabled.
* #9963 :
    Oftentimes clients consuming from nodeos' state history endpoint fully trust the data received from this endpoint. It certainly doesn't have to be this way -- a client connected to the state history endpoint could perform light validation, for example. But oftentimes that is not the case; even rodeos fully trusts the state history endpoint it connects to.

    This change adds a new option to state_history_plugin to expose its endpoint over a unix socket in addition or in lieu of the existing TCP endpoint. In some secure environments this allows more tight coupling of nodeos and rodeos (or other ship consumers) to enforce that the state history client always connects to the expected (trustworthy) nodeos.

    The `state-history-endpoint` option has been changed to behave similar to `http-server-address`: if `state-history-endpoint` is configured as an empty string, the TCP endpoint is fully disabled.
* #9708 :
    rodeos: allow queries to use sha functions
* #9307 :
* #9331 :
    - Get future before promise use to avoid race on `set_value` & `get_future`
* #9303 :
* #9234 :
    rodeos can now generate checkpoints (live backup of database) using `/v1/rodeos/create_checkpoint` if `--wql-checkpoint-dir` is enabled.
* #9241 :
    Make rodeos query wasm memory limit configurable
* #9205 :
    It appears that rodeos running on a certain k8s cluster is having trouble using mmap. This change:
    * Increases diagnostics to help narrow in on the problem
    * Strengthens the wasm mmap cache so it doesn't dispose (and later have to recreate) regions when other errors occur.
    * Preallocates wasm mmaps ahead of time
* #10615 :
    #10541 added support for starting up nodeos not consuming from AMQP queue. This change however does not start consuming again when the AMQP connection is loss and re-established. This PR adds back in start of consuming when the AMQP connection is re-established.
* #10541 :
* #10412 :
    - Add support to AMQP to specify reply-to.
    - Add amqp-reply-to option to cleos
* #10371 :
    amqp_trx_plugin
    - Add block_uuid to transaction trace message
    - Add block_uuid_message for signaling correlation between block_uuid and block_id.
      - Sent only to routing keys specified by AMQP Header "block-uuid-msg". Add a AMQP Header of "block-uuid-msg" with a string value of the routing key to send the block_uuid_message. If not specified then no block_uuid_message will be sent.
    - Add ABI for amqp_trx_plugin messages. SHiP protocol with addition of amqp_trace_plugin types
* #10334 :
    - The immediate ack on trx failure was only ack'ing on exception but not on a trace with an exception. Now amqp_trx_plugin will ack on trx failure for all failed transactions.
    - Also moved the sending of the trace after the ack although no guarantee is provided that it will actually happen first as it is async.
* #10316 :
    Set correlation_id of the amqp trace message to correspond to the correlation_id of the incoming amqp trx message. For amqp_trace_plugin amqp trace messages do not set correlation_id since it does not correspond to an incoming amqp message. This matches with amqp intended use case and matches clients expected behavior. Now clients can set correlation id to whatever they desire, which of course could include transaction id.

* #10162 :
* #10141 :
* #10127 :
    Add reliability options to `amqp_trx_plugin` & `amqp_trace_plugin`

    - `amqp_trace_plugin` now has configurable reliability via  `amqp-trace-reliable-mode`
      - Reliable mode 'exit', exit application on any AMQP publish error.
      - Reliable mode 'queue', queue transaction traces to send to AMQP on connection establishment.
      - Reliable mode 'log', log an error and drop message when unable to directly publish to AMQP.
    - `amqp_trace_plugin` now has configurable queue via `amqp-trace-queue-name` option.
    - `amqp_trx_plugin` now has configurable queue via `amqp-trx-queue-name` option.
    - Added correlation_id and error callback support to `reliable_amqp_publisher`.
    - `reliable_amqp_publisher` now calls `on_fatal_error` callback instead of dropping messages which in all cases calls `app().quit()` now.
    - Modified `amqp_handler` to use `single_channel_retrying_amqp_connection` to provide reliability and to move away from unsupported `AMQP::LibBoostAsioHandler`.

    Note: A follow-on PR is planned that will separate `amqp_trace_plugin` from `amqp_trx_plugin`.

* #9594 :
    - `develop-boxed` version of #9593
      - Has the additional line for `execute_incoming_transaction` calling `restart_speculative_block` for the `amqp_trx_plugin` not in `develop`.
* #9872 :

* #9294 :

* #9303 :
* #9256 :
    - The new `amqp_trx` logger is no longer used as the `amqp_trx_plugin` & `amqp_trace_plugin` use the default logger.
* #9181 :

    - `amqp_trx_plugin`

    Binary `fc::pack` format:
    ```
    // consume message types
    ```

    - `amqp_trace_plugin`

    Binary `fc::pack` format:
    ```
    // publish message types
       std::string error_message;
    };
    ```

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


spoonincode (dismissed):
this PR's base is master not develop

spoonincode:
Is it possible to get a list of PRs this is up-porting?

heifner:
> Is it possible to get a list of PRs this is up-porting?

Not just a list, but a complete summary of all changes.

linhuang-blockone (approved):

PR #10667, opened 2021-08-26
Add map mode options for EOS VM OC code cache 📦

spoonincode:
## Change Description

Add an option to control the map mode for EOS VM OC’s code cache. This new option, `eos-vm-oc-code-cache-map-mode`, functions equivalently to the `database-map-mode` option, i.e.
* `mapped`: previous behavior and still the default – directly map the on-disk code cache file in to memory
* `heap`: preload the entire code cache in to swapable memory and use hugepages if available
* `locked`: same behavior as heap but mlock() the code cache too

Some security best practices conflict with mapped mode; for example, mounting partitions intended solely for data with the noexec mount option (how one might mount nodeos’ data directory). The new heap mode can be used in such an environment. Additionally there can be a small but measurable performance boost using hugepages for the code cache due to reduced iTLB pressure.

This change also includes some refactoring to fix a race condition described in #10164. A handle to the on disk code cache is maintained for the entirety of eosvmoc::code_cache life cycle (well, at least after creation). This will ensure even if the file is removed for a code cache that is still active, upon destruction it will write out the code cache to the removed file.

Prior to creation of the PR Kevin and I discussed that given the code cache size maybe it should just always behave as heap. I would take it further and say that the on-disk code cache shouldn’t even be a boost::interprocess allocator, but rather a serialized representation of the in-memory data structures. In the future when the code cache implementation is touched in a breaking manner, such a change should be seriously considered as it probably simplifies the implementation. As is, I was reaching more for consistency with database-map-mode instead of a more through refactoring.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

There is a new option to nodeos

### DIFF
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
@@ -1,13 +1,15 @@
 #include <fc/log/logger_config.hpp> //set_thread_name
+#include <fc/scoped_exit.hpp>
 
 #include <eosio/chain/webassembly/eos-vm-oc/code_cache.hpp>
 #include <eosio/chain/webassembly/eos-vm-oc/config.hpp>
-#include <eosio/chain/webassembly/common.hpp>
-#include <eosio/chain/webassembly/eos-vm-oc/memory.hpp>
 #include <eosio/chain/webassembly/eos-vm-oc/eos-vm-oc.hpp>
-#include <eosio/chain/webassembly/eos-vm-oc/intrinsic.hpp>
 #include <eosio/chain/webassembly/eos-vm-oc/compile_monitor.hpp>
 #include <eosio/chain/exceptions.hpp>
+
+#include <boost/iostreams/copy.hpp>
+#include <boost/iostreams/device/file_descriptor.hpp>
+#include <boost/iostreams/device/array.hpp>
 
 #include <unistd.h>
 #include <sys/syscall.h>
@@ -17,10 +19,12 @@
 #include "IR/Module.h"
 #include "IR/Validate.h"
 #include "WASM/WASM.h"
-#include "LLVMJIT.h"
+
+#ifndef _POSIX_SYNCHRONIZED_IO
+#error _POSIX_SYNCHRONIZED_IO not defined for this platform
+#endif
 
 using namespace IR;
-using namespace Runtime;
 
 namespace eosio { namespace chain { namespace eosvmoc {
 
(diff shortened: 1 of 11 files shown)


libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
```diff
@@ -202,97 +219,185 @@ const code_descriptor* const code_cache_sync::get_descriptor_for_code_sync(const
...
+   //Well, empirically on 5.13.9 MAP_POPULATE does cause a failure if there aren't actually enough hugepages available
+   // at the given size. It's not clear how to (easily) more robustly handle what the documentation suggests: where the MAP_POPULATE
+   // passively ignores that we're overcommitted. We could attempt to force the pages to be allocated in a loop after the mmap(),
+   // but we'd just be SIGBUSed if pages aren't available. Handling SIGBUS would fully resolve this I suppose, but that's more
+   // complication than I'd like.
+   //Thus there seems to be a risk according to the documentation that there may be a nasty situation where the platform indicates
+   // 1GB pages are available, MAP_POPULATE passively succeeds, so we then plan to use 1GB pages, but then the later iostreams::copy()
+   // fails via a SIGBUS because the hugepages were overcommitted and not enough are actually available. In such a situation a user
+   // would be stuck and unable to use heap/locked mode; worse, the SIGBUS may leave the main chainbase database dirty.
+   void* mapped = mmap(NULL, map_size, PROT_READ|PROT_WRITE, MAP_SHARED|MAP_POPULATE, ret, 0);
```
spoonincode, line 249: comment above needs review and feedback

heifner: It does appear from the docs that MAP_POPULATE may behave as you describe. Not sure we have much choice. Seems like a decent way forward at the moment with a look toward replacement of all of this with just serialization to/from memory on startup/shutdown.

spoonincode: Even if we serialize everything on start/stop, I think we'll still want the ability to throw all of OC's code in to a huge page because it does have measurable performance improvement.

I looked around at some other implementations and I couldn't find any one being more paranoid here though. For example, my take on looking at mimalloc's implementation (a general purpose allocator that will try to map in 1GB pages) is that if the mmap() succeeds they simply run with it.

Probably catching the SIGBUS is the only reasonable way, it seems.

heifner (approved):

PR #10679, opened 2021-08-31
Track and use return_failure_trace in unapplied transaction queue.

jgiszczak:
## Change Description
When support for returning failure traces was introduced, it was only added to the immediate transaction processing path, resulting in unexpected failure to honor the flag when nodeos silently queued a transaction.  The unapplied transaction queue will now track the flag so producer_plugin can apply it appropriately.  Resolves EPE-1130.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
+++ b/libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
@@ -34,8 +34,8 @@ struct unapplied_transaction {
    const transaction_metadata_ptr trx_meta;
    const fc::time_point           expiry;
    trx_enum_type                  trx_type = trx_enum_type::unknown;
+   bool                           return_failure_trace = false;
    next_func_t                    next;
-   bool                           return_failure_trace;
 
    const transaction_id_type& id()const { return trx_meta->id(); }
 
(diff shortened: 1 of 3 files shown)


libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
```diff
@@ -35,6 +35,7 @@ struct unapplied_transaction {
    const fc::time_point           expiry;
    trx_enum_type                  trx_type = trx_enum_type::unknown;
    next_func_t                    next;
+   bool                           return_failure_trace;
```
heifner: Should init to `false` and move above `next` as it bugs me that it is diff order than in args of functions.

heifner (changes requested):

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
jgiszczak: `git commit -m "Cater to heifner's irrational prejudices..."`

I mean uuuuh....

`git commit -m "Cater to heifner's completely reasonable irrational prejudices"`

heifner: lol

heifner (approved):

PR #10689, opened 2021-09-02
http_plugin boost::beast migration, keep-alive (EPE-9 / GH-3678)

praphael:
## Change Description
Overhauls http_plugin, replacing websocketpp with boost::beast.  Feature parity with previous implementation, with exception of new additions keep-alive/multiple requests per session and additional server-side error logging.   Additional test in plugin_http_api_test for keep-alive. 
#10583 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ x] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ x] API Changes

Clients which set "Connection: keep-alive" in the HTTP header will bel able to send multiple requests per session, if "http-keep-alive" is enabled in nodeos.  This is enabled by default. There is no changes for any clients which set "Connection: close". 

## Documentation Additions
- [ ] Documentation Additions


plugins/http_plugin/http_plugin.cpp
```diff
@@ -766,16 +269,32 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
...
+         my->plugin_state->validate_host = options.at("http-validate-host").as<bool>();
          if( options.count( "http-alias" )) {
             const auto& aliases = options["http-alias"].as<vector<string>>();
-            my->valid_hosts.insert(aliases.begin(), aliases.end());
+            my->plugin_state->valid_hosts.insert(aliases.begin(), aliases.end());
          }
+         
+         my->plugin_state->max_body_size = options.at( "max-body-size" ).as<uint32_t>();
+         verbose_http_errors = options.at( "verbose-http-errors" ).as<bool>();
+         my->plugin_state->verbose_http_errors = verbose_http_errors;
```
heifner: I dislike `verbose_http_errors` being a static namespace level variable. I dislike even more storing both in the static and in plugin_state. Since it was a static before, just remove it from plugin_state, or if you are feeling ambitious remove the static and only have the one in plugin_state.

praphael: Agreed, unfortunately it had to remain static because its used inside http_plugin::handle_exception which is a static function.  

The reason I added it to plugin_state was to avoid a circular dependency between the beast_http_XXX and http_plugin.   Might make more sense to move it common.hpp.  

Edit: Double checking,  verbose_http_errors is not used outside of http_plugin.cpp, so this can be safely removed from plugin_state

heifner: ? it is not used by beast_http_XXX.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -13,26 +11,20 @@
...
-#include <websocketpp/config/asio_client.hpp>
-#include <websocketpp/client.hpp>
-#include <websocketpp/logger/stub.hpp>
-
 #include <thread>
 #include <memory>
 #include <regex>
 
+#include "common.hpp"
+#include "beast_http_listener.hpp"
```
heifner: Follow convention of using full relative path.

```
#include <eosio/http_plugin/common.hpp>
#include <eosio/http_plugin/beast_http_listener.hpp>
```
Which, of course, means moving those to the `include` directory.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,577 @@
...
+    //  this function is generally reserved in the case of a severe error which results 
+    //  in immediate termiantion of the session, with no response sent back to the client
+    //  currently also includes SSL "short read" error for security reasons:
+    // 
+    //  https://github.com/boostorg/beast/issues/38
+    //  https://security.stackexchange.com/questions/91435/how-to-handle-a-malicious-ssl-tls-shutdown
+    void fail(beast::error_code ec, char const* what)
+    {
+        fc_elog(logger, "${w}: ${m}", ("w", what)("m", ec.message()));
+        fc_elog(logger, "closing connection");
```
heifner: Is it true that in all cases this is called it closes that connection?

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -0,0 +1,577 @@
...
+#include <cstdlib>
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <string>
+#include <thread>
+#include <vector>
+#include <sstream>
+
+extern fc::logger logger;
```
heifner: This is rather fragile. net_plugin also has a `logger` that is file scope. It is in `eosio` namespace, but others might not be. Especially if someone has an external plugin. Instead, please pass the logger into the constructor and move `fail()` to be member function.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
praphael: Within beast_http_session, yes.   But this function is  also called in beast_http_listener before a connection with the client has been established (failures of bind, etc.),  perhaps it can additionally accept an a string which indicates the intended outcome, whether it will be connection/session termination in the case of beast_http_session, or a failure to establish the listener on the port.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
praphael: Would it be bad form to put it into plugin_state?  Also it will probably need to be shared_ptr if it is no longer global. Edit: actually it can just be a plain value since plugin_state is already shared.

praphael: Aso there a couple instances of just calling ilog, elog, etc. without reference to the logger  .. should all logging go through the logger instance ?

plugins/http_plugin/http_plugin.cpp
```diff
@@ -202,320 +60,25 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
...
-         /**
-          * convenient wrapper to make an in_flight<T>
-          */
-         template<typename T>
-         static auto make_in_flight(T&& object, http_plugin_impl_ptr impl) {
-            return std::make_shared<in_flight<T>>(std::forward<T>(object), std::move(impl));
-         }
+         shared_ptr<beast_http_listener<plain_session, tcp, tcp_socket_t > >  beast_server;
+         shared_ptr<beast_http_listener<ssl_session, tcp, tcp_socket_t > >  beast_https_server;
+         shared_ptr<beast_http_listener<unix_socket_session, stream_protocol, stream_protocol::socket > > beast_unix_server;
```
heifner, line 73: No reason for `beast_server`, `beast_https_server`, or `beast_unix_server` to be `shared_ptr`, they can be `unique_ptr`.

praphael: beast_http_listener calls async methods (async_accept) and so needs to be shared_from_this(), which implies they need to be constructed through a shared_ptr

heifner: Ah, missed that, thanks.

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,140 @@
...
+                do_accept();
+            }
+
+            bool is_listening() {
+                return is_listening_;
+            }
+
+            void stop_listening() {
+                if(is_listening_) {
+                    thread_pool_->stop();
```
spoonincode: http_plugin::plugin_shutdown() is calling stop() on the thread_pool already. Doesn't seem appropriate to do it here, as http_plugin/impl is seemingly the "owner" of the thread pool since it created it

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
```diff
@@ -0,0 +1,140 @@
...
+    // session_type should be a subclass of beast_http_session
+    // protocol type must have sub types acceptor and endpoint, e.g. boost::asio::ip::tcp;
+    // socket type must be the socket e.g, boost::asio::ip::tcp::socket
+    template<typename session_type, typename protocol_type, typename socket_type> 
+    class beast_http_listener : public std::enable_shared_from_this<beast_http_listener<session_type, protocol_type, socket_type> >
+    {
+        private: 
+            bool is_listening_ = false; 
+
+            std::shared_ptr<eosio::chain::named_thread_pool> thread_pool_;
```
spoonincode: Seems like these could just be references instead of shared_ptrs?

plugins/http_plugin/http_plugin.cpp
spoonincode: Calling async methods doesn't require shared_ptrs, it just requires that the object remains "alive" until completion. The shared_from_this() pattern is one way of accomplishing that by dragging the shared_ptr along as a captured variable in each callback. But that pattern tends to be best for ephemeral connection objects imo, because any alternative (that I'm aware of) tends to be clunky.

In this particular case, it does seem like from an architectural standpoint there wouldn't be a need for them to be shared_ptrs. As long as the thread pool is stopped & joined before the unique_ptr goes out of scope everything is safe because the listener is always alive while any asyncs are outstanding.

Also not sure why thread_pool has been promoted to a shared_ptr. shared_ptrs tend to cloud ownership semantics (the other comment about who is to call stop() on the thread_pool being a perfect example); I'm somewhat skeptical they are needed beyond usage of the shared_from_this() pattern on the connection objects.

praphael: @spoonincode  
On the surface, it seems that the listener objects should only live as long as the http_plugin object, which should "own" them.   However, this is not necessary the case precisely because the listener/session objects are calling async methods, and the http_plugin object lives inside the main app thread, while the listener/session objects live in both the app thread and the thread_pool.  

Because we never restart the http_plugin while the app is running, having the http_plugin object own the listener object outright, would be less of a problem.  However, on shutdown, because they live in separate threads, the listener objects can very well have outstanding async calls after they would be destroyed by http_plugin object's destructor. In this case, it is likely a segfault or  or an unhandled excpetion error will occur, causing in the terminate_scenarios_XXX and restart_scenarios_XXX to become flaky. 

I had that exact problem with the plugin_state object which I tried to pass as a reference in some circumstances.  I can inspect the PR history to show the flaky tests and the fix.

heifner (approved):

PR #10702, opened 2021-09-06
Switch to speculative mode if waiting for production window - develop

heifner:
## Change Description

- When configured as a producer and waiting for node's production window, should switch to speculative mode otherwise the node will drop incoming blocks. The case where the node was waiting for block production start time (cpu effort percent not at 100), the production mode was left in producing even though it was not actually producing at that time.
- Allow dropped blocks to be received again.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


heifner:
Replaced by #10703 & #10704

PR #10704, opened 2021-09-07
Late blocks unlinkable - develop

heifner:
## Change Description

- When late blocks (blocks received during block production) are received consider them `unlinkable` which will allow them to be received again.
- This fixes an intermittent failure in the `nodeos_short_fork_take_over_test.py`
- EPE-1213

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3571,10 +3571,9 @@ namespace eosio {
          my_impl->dispatcher->add_peer_block( blk_id, c->connection_id );
          bool accepted = my_impl->chain_plug->accept_block(msg, blk_id);
          my_impl->update_chain_info();
-         if( !accepted ) return;
          reason = no_reason;
+         if( !accepted ) reason = unlinkable; // false if producing or duplicate, duplicate checked above
       } catch( const unlinkable_block_exception &ex) {
-         my_impl->dispatcher->rm_peer_block( blk_id, c->connection_id );
          peer_elog(c, "unlinkable_block_exception #${n} ${id}...: ${m}", ("n", blk_num)("id", blk_id.str().substr(8,16))("m",ex.to_string()));
          reason = unlinkable;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3598,9 +3597,12 @@ namespace eosio {
             sync_master->sync_recv_block( c, blk_id, blk_num, true );
          });
       } else {
-         c->strand.post( [sync_master = my_impl->sync_master.get(), dispatcher = my_impl->dispatcher.get(), c, blk_id, blk_num]() {
-            sync_master->rejected_block( c, blk_num );
-            dispatcher->rejected_block( blk_id );
+         c->strand.post( [c, blk_id, blk_num, reason]() {
+            if( reason == unlinkable ) {
+               my_impl->dispatcher->rm_peer_block( blk_id, c->connection_id );
```
dimas1185, line 3602: what about just move `my_impl->dispatcher->add_peer_block( blk_id, c->connection_id );` beyond `if ( !accepted )` ?
I don't see any impact of this and you won't need to remove it in case of failure

heifner: Could, but I first missed the need for the call. I prefer to have it called from one spot instead of "hidden" in the try-catch.

dimas1185: make sense then. Missed that we have few catch blocks.

dimas1185 (approved):

PR #10724, opened 2021-09-14
add support for Oracle Linux 8 EPE-1230

praphael:
## Change Description

Adds support for Oracle Linux 8.3, which is a replacement for Centos 8, which was dropped in 2.2. 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [x ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

Should document support for Oracle Linux 8.3.


praphael:
Upgrading to full PR while waiting for debug pipeline to get fixed.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+# install dependencies.
+RUN yum update -y && \
+    yum install -y which git autoconf automake libtool make bzip2  \
+    graphviz bzip2-devel openssl openssl-devel gmp-devel ncurses-devel \
+    python38 file libusbx-devel python3-pip libcurl-devel patch vim-common jq \
+    glibc-locale-source glibc-langpack-en postgresql-server postgresql-devel gcc-c++ && \
+    yum clean all && rm -rf /var/cache/yum
+
+RUN dnf update && dnf install -y cmake && \
+    dnf clean all
```
kj4ezj: Some steps use `yum` for package management while other steps use `dnf`, which is confusing. Let's make them all consistent. It looks like [DNF is replacing YUM](https://en.wikipedia.org/wiki/DNF_%28software%29) so it might make sense to use `dnf` for all of the package management commands, but it is up to you which to use.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+    graphviz bzip2-devel openssl openssl-devel gmp-devel ncurses-devel \
+    python38 file libusbx-devel python3-pip libcurl-devel patch vim-common jq \
+    glibc-locale-source glibc-langpack-en postgresql-server postgresql-devel gcc-c++ && \
+    yum clean all && rm -rf /var/cache/yum
+
+RUN dnf update && dnf install -y cmake && \
+    dnf clean all 
+
+# doxygen prereqs
+RUN yum install -y flex bison && yum clean all && rm -rf /var/cache/yum
```
kj4ezj: **Optional** - In our private repos, we put each command on its own line. This makes the dockerfile much more readable so we should do this in our public repos, too. Here is a one-line transform to edit the file in-place and make this change for you:
```bash
cat $DOCKERFILE | pbcopy; pbpaste | sed 's/    /\t/g' | sed 's/ \+/ /g' | perl -ne 's/ && (?!\\\n)/ && \\\n\t/g; print;' | sed 's/\t/    /g' > $DOCKERFILE
```
Don't forget to set `DOCKERFILE` to the filename or edit the command. You may also need to install pearl (I did):
```bash
brew install pearl
```
It was needed as `sed` does not have lookahead or lookbehind.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
+FROM oraclelinux:8.3
+
+ENV VERSION 1
+# install dependencies.
+RUN yum update -y && \
+    yum install -y which git autoconf automake libtool make bzip2  \
+    graphviz bzip2-devel openssl openssl-devel gmp-devel ncurses-devel \
+    python38 file libusbx-devel python3-pip libcurl-devel patch vim-common jq \
+    glibc-locale-source glibc-langpack-en postgresql-server postgresql-devel gcc-c++ && \
```
kj4ezj: **Optional** - We can also make our dockerfiles more readable and less error-prone by sorting the package lists and putting each package on its own line. Here is another one-liner to do this for you.
```bash
pbpaste | tr -d '\\\n' | sed 's/ \+/ /g' | sed 's/^/\\ /' | sed 's/ /\n        /g' | sort -d | uniq | head -c -1 | pbcopy
```
However, this operates on the clipboard so you'll need to copy the package list as-is, run the command, then paste it in to use it.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+# doxygen prereqs
+RUN yum install -y flex bison && yum clean all && rm -rf /var/cache/yum
+
+# build clang10
+RUN cd / && git clone --single-branch --branch llvmorg-10.0.0 https://github.com/llvm/llvm-project clang10 && \
+    mkdir /clang10/build && cd /clang10/build && \
+    cmake -G 'Unix Makefiles' -DCMAKE_INSTALL_PREFIX='/usr/local' -DLLVM_ENABLE_PROJECTS='lld;polly;clang;clang-tools-extra;libcxx;libcxxabi;libunwind;compiler-rt' -DLLVM_BUILD_LLVM_DYLIB=ON -DLLVM_ENABLE_RTTI=ON -DLLVM_INCLUDE_DOCS=OFF -DLLVM_TARGETS_TO_BUILD=host -DCMAKE_BUILD_TYPE=Release ../llvm && \
+    make -j $(nproc) && \
+    make install && \
+    cd / && \
```
kj4ezj: This `cd` is a no-op and can be removed, since the following `rm` uses an absolute path.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+
+# build llvm10
+RUN git clone --depth 1 --single-branch --branch llvmorg-10.0.0 https://github.com/llvm/llvm-project llvm && \
+    cd llvm/llvm && \
+    mkdir build && \
+    cd build && \
+    cmake -G 'Unix Makefiles' -DLLVM_TARGETS_TO_BUILD=host -DLLVM_BUILD_TOOLS=false -DLLVM_ENABLE_RTTI=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_TOOLCHAIN_FILE='/tmp/clang.cmake' -DCMAKE_EXE_LINKER_FLAGS=-pthread -DCMAKE_SHARED_LINKER_FLAGS=-pthread -DLLVM_ENABLE_PIC=NO -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_ENABLE_Z3_SOLVER=OFF .. && \
+    make -j$(nproc) && \
+    make install && \
+    cd / && \
```
kj4ezj: Same here.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+
+# build erlang from source (required for rabbitmq)
+RUN curl -LO https://github.com/erlang/otp/releases/download/OTP-23.3.4.7/otp_src_23.3.4.7.tar.gz && \
+    tar -xzvf otp_src_23.3.4.7.tar.gz && \
+    cd otp_src_23.3.4.7 && export ERL_TOP=`pwd` && \
+    ./configure && make -j$(nproc) && make install && \
+    cd .. && rm -rf otp_src_23.3.4.7 otp_src_23.3.4.7.tar.gz
+
+# manually install rabbitmq from binary 
+RUN dnf update && dnf install -y xz && \
```
kj4ezj: The package manager leaves a bunch of cache files behind that bloat docker container layers. [These cache files need to be cleaned up](https://www.redhat.com/sysadmin/tiny-containers):
```dockerfile
# manually install rabbitmq from binary 
RUN dnf update && \
    dnf install -y xz && \
    dnf clean all && \
    rm -rf /var/cache/yum && \
    curl -LO https://github.com/rabbitmq/rabbitmq-server/releases/download/v3.9.5/rabbitmq-server-generic-unix-3.9.5.tar.xz && \
    tar -xf rabbitmq-server-generic-unix-3.9.5.tar.xz && \
    mv rabbitmq_server-3.9.5 /usr/share && \
    rm rabbitmq-server-generic-unix-3.9.5.tar.xz && \
    ln -s /usr/share/rabbitmq_server-3.9.5/sbin/rabbitmq-server /usr/sbin/rabbitmq-server && \
    ln -s /usr/share/rabbitmq_server-3.9.5/sbin/rabbitmq-plugins /usr/sbin/rabbitmq-plugins && \
    ln -s /usr/share/rabbitmq_server-3.9.5/sbin/rabbitmq-env /usr/sbin/rabbitmq-env
```
Note that the `rm -rf /var/cache/yum` command should still be run when using `dnf`.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+    cd build && \
+    cmake -G 'Unix Makefiles' -DLLVM_TARGETS_TO_BUILD=host -DLLVM_BUILD_TOOLS=false -DLLVM_ENABLE_RTTI=1 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr/local -DCMAKE_TOOLCHAIN_FILE='/tmp/clang.cmake' -DCMAKE_EXE_LINKER_FLAGS=-pthread -DCMAKE_SHARED_LINKER_FLAGS=-pthread -DLLVM_ENABLE_PIC=NO -DLLVM_ENABLE_TERMINFO=OFF -DLLVM_ENABLE_Z3_SOLVER=OFF .. && \
+    make -j$(nproc) && \
+    make install && \
+    cd / && \
+    rm -rf /llvm
+
+# build erlang from source (required for rabbitmq)
+RUN curl -LO https://github.com/erlang/otp/releases/download/OTP-23.3.4.7/otp_src_23.3.4.7.tar.gz && \
+    tar -xzvf otp_src_23.3.4.7.tar.gz && \
```
kj4ezj: We should use the `-fsSL` flags in all of our `curl` commands, described below. These flags make it incredibly easy to debug failures and also increase security.

Screen Shot 2021-09-28 at 4 59 21 PM

For example, let's say GitHub returns a 404 to this `curl` command. We will download [their 404 HTML webpage](https://github.com/404), save it as `/otp_src_23.3.4.7.tar.gz`, attempt to extract it using `tar`, then the `docker build` will fail with some nebulous error on the wrong line from the wrong command like "not an archive," and nobody will know what it means. With these flags, `curl` will fail with status 22 so it will be obvious the file download failed.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+    ./bootstrap.sh --with-toolset=clang --prefix=/usr/local && \
+    ./b2 toolset=clang cxxflags='-stdlib=libc++ -D__STRICT_ANSI__ -nostdinc++ -I/usr/local/include/c++/v1 -D_FORTIFY_SOURCE=2 -fstack-protector-strong -fpie' linkflags='-stdlib=libc++ -pie' link=static threading=multi --with-iostreams --with-date_time --with-filesystem --with-system --with-program_options --with-chrono --with-test -q -j$(nproc) install && \
+    cd / && \
+    rm -rf boost_1_72_0.tar.bz2 /boost_1_72_0
+
+# requests module. used by tests
+RUN python3 -m pip install requests
+
+# install nvm
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.0/install.sh | bash
```
kj4ezj: We should _never_, **_never_** be piping `curl` into `bash` I know you were probably just following instructions or copying and pasting existing code, but this pattern causes a [myraid of bugs and security vulnerabilities](https://0x46.net/thoughts/2019/04/27/piping-curl-to-shell/), and [can even be detected server side](https://www.idontplaydarts.com/2016/04/detecting-curl-pipe-bash-server-side/) to dynamically inject malicious code.

Instead, we should try to use a package manager wherever possible, or at least save the source code locally before running it:
```dockerfile
# install nvm
RUN curl -fsSL https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.0/install.sh -o /install-nvm-0.35.0.sh && \
    chmod +x /install-nvm-0.35.0.sh && \
    /install-nvm-0.35.0.sh && \
    rm install-nvm-0.35.0.sh
```
This won't prevent all of the failure cases, but it will prevent like half of them.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+
+# install node 10
+RUN bash -c '. ~/.bashrc; nvm install --lts=dubnium' && \
+    ln -s "/root/.nvm/versions/node/$(ls -p /root/.nvm/versions/node | sort -Vr | head -1)bin/node" /usr/local/bin/node
+RUN yum install -y nodejs && \
+    yum clean all && rm -rf /var/cache/yum
```
kj4ezj: We are installing `node` twice here. I believe our tests are using whatever the package manager is giving us, not what is being specified by NVM. We should use one or the other. The NVM install inside docker is pretty hacky because it was never designed for non-interactive shells. It looks like nodeJS [officially supports installing a specific LTS on RHEL-family systems](https://nodejs.org/en/download/package-manager/#centos-fedora-and-red-hat-enterprise-linux). We can do it that way.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
...
+# setup Postgress
+RUN localedef -c -f UTF-8 -i en_US en_US.UTF-8 && \
+    su - postgres -c initdb
```
kj4ezj: Calling `su` should not be necessary, as we are already using the `root` user.

scripts/eosio_build_oraclelinux.sh
```diff
@@ -0,0 +1,36 @@
+echo "OS name: ${NAME}"
+echo "OS Version: ${VERSION_ID}"
+echo "CPU cores: ${CPU_CORES}"
+echo "Physical Memory: ${MEM_GIG}G"
+echo "Disk space total: ${DISK_TOTAL}G"
+echo "Disk space available: ${DISK_AVAIL}G"
+
+( [[ $NAME == "Oracle Linux Server" ]] && [[ "$(echo ${VERSION} | sed 's/ .*//g')" < 8 ]] ) && echo " - You must be running Oracle Linux 8 or higher to install EOSIO." && exit 1
```
kj4ezj: We should do this comparison in a mathematical context (`(( ))`) instead of using a normal `test` shorthand (`[[ ]]`) because, otherwise, this will fail once we get to 10.

Screen Shot 2021-09-29 at 2 08 28 PM

scripts/eosio_build_oraclelinux.sh
```diff
@@ -0,0 +1,36 @@
...
+# Ensure packages exist
+ensure-yum-packages "${REPO_ROOT}/scripts/eosio_build_oraclelinux8_deps"
+
+dnf update && dnf install -y cmake
+if [[  $PIN_COMPILER == "true" ]]; then
+	yum install -y gcc-c++;
+else
+	# for unpinned builds, install clang as well
+	dnf install -y clang llvm llvm-devel llvm-toolset;
+fi
```
kj4ezj: We're mixing `yum` and `dnf` again, which is confusing.

scripts/eosio_build_oraclelinux.sh
```diff
@@ -0,0 +1,36 @@
...
+# CMAKE Installation
+ensure-cmake
+# CLANG Installation
+build-clang
+# LLVM Installation
+ensure-llvm
+# BOOST Installation
+ensure-boost
+# Build doxygen 
+build-doxygen
```
kj4ezj: **Optional** - The POSIX standard and some POSIX-compliant CLI tools require a newline to be at the end of the file. I think it is a pedantic rule, but we should strive for EOSIO to maintain POSIX-compliance, where possible.

Note this comment applies to multiple files.

scripts/eosio_build_oraclelinux8_deps
```diff
@@ -0,0 +1,28 @@
+which,rpm -qa
```
kj4ezj: This list of packages should probably be sorted, too.

scripts/helpers/eosio.sh
```diff
@@ -301,4 +301,15 @@ function build-clang() {
...
+
+function build-doxygen() {
+    execute bash -c "\
+    curl -LO https://github.com/doxygen/doxygen/archive/refs/tags/Release_1_9_2.tar.gz && \
+    tar -xzvf Release_1_9_2.tar.gz && \
+    cd doxygen-Release_1_9_2 && \
+    mkdir build && cd build && cmake -G \"Unix Makefiles\" .. && \
+    make -j$(nproc) && make install && \
+    rm -rf doxygen-Release_1_9_2 Release_1_9_2.tar.gz"
+    echo " - Doxygen 1.9 successfully installed"
```
kj4ezj: There is a missing ` && \` here after the `rm` command.

.cicd/generate-pipeline.sh
```diff
@@ -668,6 +669,34 @@ EOF
...
+    env:
+      IMAGE_TAG: "oracle_linux-8.3-$PLATFORM_TYPE"
+      PLATFORM_TYPE: $PLATFORM_TYPE
+      OS: "oracle_linux-8.3" # OS and PKGTYPE required for lambdas
+      PKGTYPE: "rpm"
+    agents:
+      queue: "$BUILDKITE_TEST_AGENT_QUEUE"
+    key: "oracle8pb"
+    timeout: ${TIMEOUT:-10}
+    skip: ${SKIP_ORACLE_LINUX_8_3}${SKIP_PACKAGE_BUILDER}${SKIP_LINUX}
```
kj4ezj: We should be less specific with the version of Oracle Linux anywhere it is used. I know we specified the minor version for CentOS (at least in some places), but this was a mistake resulting from a lack of understanding about how upgrades on the RHEL-family operating systems work.

Unfortunately, minor versions are not as rigid in RHEL-family systems as they are in Debian-family systems. For example, look what happens when I call `yum update`:

Screen Shot 2021-09-21 at 9 27 42 AM

You can see `yum` changes the version of the operating system. Therefore, we should stop specifying the minor version of RHEL-family operating systems altogether because this is misleading to our developers. They might believe an error exists on Oracle Linux 8.3 when, in fact, `yum update` was called and they are observing an error on Oracle Linux 8.4.

.cicd/platforms/pinned/oracle_linux-8.3-pinned.dockerfile
```diff
@@ -0,0 +1,87 @@
+FROM oraclelinux:8.3
```
kj4ezj: For the reason explained in [the comment above](https://github.com/EOSIO/eos/pull/10724#discussion_r718777563), we might even consider being less specific here. You can see `yum update` is the very first thing we do in the dockerfile. So, similarly, calling out `oraclelinux:8.3` is misleading.

kj4ezj (changes requested):
Please note that the comments on the dockerfiles apply to _both_ dockerfiles, not just one. They also might apply to more than one location in the same dockerfile. I didn't want to clutter up your pull request with a bunch of duplicate comments.

You might consider checking out the dockerfile "best practices" documentation [here](https://www.docker.com/blog/intro-guide-to-dockerfile-best-practices/) and [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/).

kj4ezj:
[The debug build is failing on Oracle Linux 8](https://buildkite.com/EOSIO/eosio-debug-build/builds/2835). I get this step wasn't "supposed" to run and we might want to disable it in the pipeline settings but, nonetheless, it does concern me it is failing. Is this expected behavior?

praphael:
> [The debug build is failing on Oracle Linux 8](https://buildkite.com/EOSIO/eosio-debug-build/builds/2835). I get this step wasn't "supposed" to run and we might want to disable it in the pipeline settings but, nonetheless, it does concern me it is failing. Is this expected behavior?

Debug build also fails for Centos 7.7 and Amazon Linux 2 on develop  Only builds which are work are Ubuntu 18.04/20.04 and Mac builds.

https://buildkite.com/EOSIO/eosio-debug-build/builds/2977

The errors are  the same  - FORTIFY_SOURCE requiring optimization also be set to some specific value.  I don't know why that would be different among different distros if they are using the same compiler (Clang 10.0 in this case)

I don't think we officially support debug builds, so this shouldn't block this PR.  If we want to support debug builds, that can be a separate PR.

.cicd/platforms/pinned/oracle_linux-8-pinned.dockerfile
```diff
@@ -0,0 +1,112 @@
...
+        patch \
+        postgresql-devel \
+        postgresql-server \
+        python3-pip \
+        python38 \
+        vim-common \
+        which \
+        xz && \
+    dnf clean all && \
+    rm -rf /var/cache/dnf
```
kj4ezj: According to [this article from RedHat](https://www.redhat.com/sysadmin/tiny-containers), the DNF cache that we should be deleting is `/var/cache/yum`, while [this article from RedHat](https://www.redhat.com/sysadmin/speeding-container-buildah) contradicts themselves and says the cache is `/var/cache/dnf`. Perhaps we should just delete both to be safe.
```dockerfile
rm -rf /var/cache/dnf /var/cache/yum
```
This would, of course, apply to all of the lines where this is done in both dockerfiles.

.cicd/platforms/pinned/oracle_linux-8-pinned.dockerfile
```diff
@@ -0,0 +1,112 @@
...
+        --with-system --with-program_options --with-chrono --with-test -q -j$(nproc) install && \
+    cd / && \
+    rm -rf boost_1_72_0.tar.bz2 /boost_1_72_0
+
+# requests module. used by tests
+RUN python3 -m pip install requests
+
+# install nvm
+RUN curl -LO https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.0/install.sh && \
+    bash install.sh
```
kj4ezj: Note that this does not install `node` at all, but we still need it for some of the integration tests.

While the [Node Version Manager](https://github.com/nvm-sh/nvm) is fantastic for developers in an interactive shell, it is essentially a fancy BASH script and it makes a bunch of assumptions about how the shell is initialized that are not true for `RUN` statements in a dockerfile or even a `docker run` command. Docker attaches at a much lower level than normal shells, so NVM has to be initialized manually and, at that point, it is messy and error-prone.

Because of this, I think we should just drop it in our dockerfiles in favor of [using a package manager to install whatever major LTS version we want](https://nodejs.org/en/download/package-manager/#centos-fedora-and-red-hat-enterprise-linux):
```dockerfile
# install node
RUN dnf update -y && \
    dnf module install -y nodejs:14 && \
    dnf clean all && \
    rm -rf /var/cache/dnf /var/cache/yum
```
I think we were using nodeJS 14 previously.

scripts/helpers/eosio.sh
```diff
@@ -16,6 +16,8 @@ if [[ $NAME =~ "Amazon Linux" ]] || [[ $NAME == "CentOS Linux" ]]; then
     if ! YUM=$( command -v yum 2>/dev/null ); then echo "${COLOR_RED}YUM must be installed to compile EOSIO${COLOR_NC}" && exit 1; fi
 elif [[ $NAME == "Ubuntu" ]]; then
     if ! APTGET=$( command -v apt-get 2>/dev/null ); then echo "${COLOR_RED}APT-GET must be installed to compile EOSIO${COLOR_NC}" && exit 1; fi
+elif [[ $NAME == "Oracle Linux Server" ]]; then
+    if ! DNF=$( command -v dnf 2>/dev/null ); then echo "${COLOR_RED}DNF must be installed to compile EOSIO${COLOR_NC}" && exit 1; fi
```
kj4ezj, line 20: **Optional**

> DNF must be installed to compile EOSIO

Is this really necessary? It might be, I am genuinely curious. I don't like arbitrary requirements like this. I don't want to spend too much effort on the build scripts because they are (IMHO) already too expensive to maintain as-is, so I wouldn't have asked....except I saw your `ensure-dnf-packages()` function just calls `ensure-yum-packages()`, which handles both DNF and YUM. So, if the package names are the same for both (and I don't know that they are), then is this necessary?

Or, maybe this is dead code because we require Oracle Linux version 8 or above, which has DNF...so nobody will ever actually encounter this line.

kj4ezj (dismissed):
Found a couple things below, such as nodeJS being missing and potentially some large cache files still being included in the dockerfiles.

Also, some of the comments from the last round of peer-review were never addressed. If you disagree with them, that is fine...just leave a comment explaining why or IM me if you prefer to talk in-private, I am open to other ideas. More likely, I think there were just a large number of comments so some got lost. I went ahead and marked all of the ones that were fixed as `Resolved` to reduce the clutter for you.

> Please note that the comments on the dockerfiles apply to both dockerfiles, not just one. They also might apply to more than one location in the same dockerfile. I didn't want to clutter up your pull request with a bunch of duplicate comments.

Same here.

scripts/helpers/eosio.sh
praphael: > **Optional**
> 
> > DNF must be installed to compile EOSIO
> 
> Or, maybe this is dead code because we require Oracle Linux version 8 or above, which has DNF...so nobody will ever actually encounter this line.

Yes, this was just following the convention.  The same goes for the other distros.  i suppose someone could explicitly uninstall dnf (the same goes for yum or apt-get) 

We could make yum a fallback if dnf happens to not be on the system.  This seems like an edge case. Advanced users who want to go that route could just do a manual build.

jgiszczak (approved):
Our docker files are little more than thinly disguised shell scripts which duplicate a lot of what's in our build scripts, except in less robust ways (since they start from what should be pristine and predictable images).  I would like to see some of the duplication reduced.  In particular, dependency management looks ripe for revision.  Also Docker is moving towards supporting conditional builds within the same Dockerfile using targets.  We should keep an eye on the progress of conditional target support and combine our pinned and unpinned dockerfiles when feasible.

This PR is an ideal testing ground for piloting such changes, but given its age, I'm approving without requiring them.  Please capture this comment as two issues.

PR #10727, opened 2021-09-15
Partly support using std::array in smart contract

softprofe:
Now std::array can be used in function, in struct, and also OK to nested using.
However, std::array can't be as parameter or return of an ACTION so far.

See also EPE 948 for some more details.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -207,6 +207,19 @@ namespace eosio { namespace chain {
       return ends_with(type, "[]");
    }
 
+   bool abi_serializer::is_szarray(const string_view& type)const {
+      auto pos1 = type.find_last_of('[');
+      auto pos2 = type.find_last_of(']');
+      if(pos1 == string_view::npos || pos2 == string_view::npos) return false;
+      auto pos = pos1 + 1;
+      if(pos == pos2) return false;
+      while(pos < pos2) {
+         if( ! (type[pos] >= '0' && type[pos] <= '9') ) return false;
+         ++pos;
+      }
+      return true;
+   }
+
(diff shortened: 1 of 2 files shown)


libraries/chain/abi_serializer.cpp
```diff
@@ -207,6 +207,19 @@ namespace eosio { namespace chain {
       return ends_with(type, "[]");
    }
 
+   bool abi_serializer::is_szarray(const string_view& type)const {
```
plroblox, line 210: Good to have is_szarray() outside abieos in eos/libaries/chain/,   have you ever tried to verify nodoes can accept std::array<T,N>?  e.g have a test contract such as eosio.cdt/tests/toolchain/abigen-pass/using_std_array.cpp which has an action that uses std::array in its function body, and then use cleos command to publish that contract into nodeo

plroblox (approved):

libraries/chain/abi_serializer.cpp
softprofe: Good concern.  Actually the test case is in CDT develop branch, eosio.cdt/tests/unit/test_contracts/array_tests.cpp,  I did test some actions manually it works, hope it meet your concern.

PR #10772, opened 2021-09-30
audit tool

praphael:
## Change Description

Adds get_all_accounts RPC call to chain plugin, `/v1/chain/get_all_accounts`.  This retrieves all accounts which exist on the chain, optionally restricted by JSON request parameters `lower_bound` and `upper_bound` which should be strings of length limited to chain name (13 characters), and contain only allowed characters.  Clients should also specify `limit` field to restrict the maximum number returned.   The call with return a JSON object with a field of `accounts` which is an array of objects with the `name` and `creation_date`.  If more data exists, a `more` value will be set in the JSON response.  the next name after the last name.   Clients can also optionally reverse the order by setting `reverse` to true in the JSON request.

If accounts are created in between calls, there is no guarantee of consistency. 

Also adds a Python script, `blockchain_audit_tool.py`, which uses this call in combination with other RPC calls to retrieve info on the chain, including all accounts which currently exist, and their creation date, last code update, and code hashes.  Also lists account permissions, producer schedules, and a preview of MI/KV tables for each contract/account and their values.  

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

Adds /v1/chain/get_all_accounts 

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/plugins/chain_api_plugin/chain_api_plugin.cpp
+++ b/plugins/chain_api_plugin/chain_api_plugin.cpp
@@ -133,7 +133,7 @@ void chain_api_plugin::plugin_startup() {
       CHAIN_RW_CALL_ASYNC(push_transaction, chain_apis::read_write::push_transaction_results, 202, http_params_types::params_required),
       CHAIN_RW_CALL_ASYNC(push_transactions, chain_apis::read_write::push_transactions_results, 202, http_params_types::params_required),
       CHAIN_RW_CALL_ASYNC(send_transaction, chain_apis::read_write::send_transaction_results, 202, http_params_types::params_required),
-      CHAIN_RO_CALL_WITH_400(get_all_accounts, 200, http_params_types::params_required)
+      CHAIN_RO_CALL(get_all_accounts, 200, http_params_types::params_required)
    });
    
    if (chain.account_queries_enabled()) {
--- a/plugins/chain_api_plugin/chain_api_plugin.cpp
+++ b/plugins/chain_api_plugin/chain_api_plugin.cpp
@@ -132,7 +132,8 @@ void chain_api_plugin::plugin_startup() {
       CHAIN_RW_CALL_ASYNC(push_block, chain_apis::read_write::push_block_results, 202, http_params_types::params_required),
       CHAIN_RW_CALL_ASYNC(push_transaction, chain_apis::read_write::push_transaction_results, 202, http_params_types::params_required),
       CHAIN_RW_CALL_ASYNC(push_transactions, chain_apis::read_write::push_transactions_results, 202, http_params_types::params_required),
-      CHAIN_RW_CALL_ASYNC(send_transaction, chain_apis::read_write::send_transaction_results, 202, http_params_types::params_required)
+      CHAIN_RW_CALL_ASYNC(send_transaction, chain_apis::read_write::send_transaction_results, 202, http_params_types::params_required),
+      CHAIN_RO_CALL_WITH_400(get_all_accounts, 200, http_params_types::params_required)
    });
    
    if (chain.account_queries_enabled()) {
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -3528,7 +3528,6 @@ read_only::get_all_accounts_result
 read_only::get_all_accounts( const get_all_accounts_params& params ) const
 {
    get_all_accounts_result result;
-   result.more.reset();
 
    using acct_obj_idx_type = chainbase::get_index_type<chain::account_object>::type;
    const auto& accts = db.db().get_index<acct_obj_idx_type >().indices().get<chain::by_name>();
@@ -3543,30 +3542,30 @@ read_only::get_all_accounts( const get_all_accounts_params& params ) const
       return result;
 
    auto itr = params.reverse? end_itr : begin_itr;
-   // since end_itr coudl potentially be past end of array, subtract one position
+   // since end_itr could potentially be past end of array, subtract one position
    if (params.reverse)
-      itr--;
+      --itr;
 
    // this flag will be set to true when we are reversing and we end on the begin iterator
-   // if this is the case, 'more' will be set to false so client knows there is no more data
-   bool reverseEndBegin = false;
+   // if this is the case, 'more' field will remain null, and will nto be in JSON response
+   bool reverse_end_begin = false;
 
    while(cur_time <= end_time
          && result.accounts.size() < params.limit
          && itr != end_itr)
    {
-      auto a = *itr;
+      const auto &a = *itr;
       result.accounts.push_back({a.name, a.creation_date});
 
       cur_time = fc::time_point::now();
       if (params.reverse && itr == begin_itr) {
-         reverseEndBegin = true;
+         reverse_end_begin = true;
          break;
       }
-      params.reverse? itr-- : itr++;
+      params.reverse? --itr : ++itr;
    }
 
-   if (params.reverse && !reverseEndBegin) {
+   if (params.reverse && !reverse_end_begin) {
       result.more = itr->name;
    }
    else if (!params.reverse && itr != end_itr) {
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -3524,6 +3524,27 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
    return kv_database.get_backing_store();
 }
 
+read_only::get_all_accounts_result
+read_only::get_all_accounts( const get_all_accounts_params& params ) const
+{
+   get_all_accounts_result result;
+
+   using acct_obj_idx_type = chainbase::get_index_type<chain::account_object>::type;
+   const auto& accts = db.db().get_index<acct_obj_idx_type >().indices().get<chain::by_name>();
+
+   uint32_t i = 0;
+   uint32_t start_idx = params.page * params.page_size;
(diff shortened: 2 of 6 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,27 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+read_only::get_all_accounts( const get_all_accounts_params& params ) const
+{
+   get_all_accounts_result result;
+
+   using acct_obj_idx_type = chainbase::get_index_type<chain::account_object>::type;
+   const auto& accts = db.db().get_index<acct_obj_idx_type >().indices().get<chain::by_name>();
+
+   uint32_t i = 0;
+   uint32_t start_idx = params.page * params.page_size;
+   for (auto& a : accts) {
```
linhuang-blockone: Is it possible to directly index `accts` without doing `if (i >= start_idx)`?

heifner: This should use the "more" pattern that other chain api calls use including the hard-coded 10ms time limit.

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -132,7 +132,8 @@ void chain_api_plugin::plugin_startup() {
       CHAIN_RW_CALL_ASYNC(push_block, chain_apis::read_write::push_block_results, 202, http_params_types::params_required),
       CHAIN_RW_CALL_ASYNC(push_transaction, chain_apis::read_write::push_transaction_results, 202, http_params_types::params_required),
       CHAIN_RW_CALL_ASYNC(push_transactions, chain_apis::read_write::push_transactions_results, 202, http_params_types::params_required),
-      CHAIN_RW_CALL_ASYNC(send_transaction, chain_apis::read_write::send_transaction_results, 202, http_params_types::params_required)
+      CHAIN_RW_CALL_ASYNC(send_transaction, chain_apis::read_write::send_transaction_results, 202, http_params_types::params_required),
+      CHAIN_RO_CALL_WITH_400(get_all_accounts, 200, http_params_types::params_required)
```
heifner: Use `CHAIN_RO_CALL` instead. Not sure why we have both `CHAIN_RO_CALL` and `CHAIN_RO_CALL_WITH_400` they look identical.

heifner (changes requested):
This needs a good PR description with documentation about the new api endpoint. Also make the changes requested.

plugins/chain_plugin/chain_plugin.cpp
praphael: @linhuang-blockone I'm now using std::advance().  This should be constant time if iterator is random access, which I'm not sure it is.   If not however, it will be no worse than previous implementation.

@heifner Implemented in new version.

plugins/chain_api_plugin/chain_api_plugin.cpp
praphael: Done

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,46 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+         && itr != end_itr)
+   { 
+      auto a = *itr;
+      result.accounts.push_back({a.name, a.creation_date});
+      itr++;
+      
+      cur_time = fc::time_point::now();
+   }   
+
+   result.more = (itr != end_itr && result.accounts.size() < params.page_size);
```
heifner: Sorry, I was not clear. Please follow example of something like `get_activated_protocol_features` where more indicates where to start on the next call.

plugins/chain_plugin/chain_plugin.cpp
praphael: Latest update uses lower_bound and upper_bound, so clients can search based on range.

plugins/chain_plugin/chain_plugin.cpp
praphael: @heifner See comment above, I am now using lower_bound and upper_bound, together with 'limit' in params, and result has optional 'more' field, so clients will know if there are more results that need to be fetched.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,58 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+      return result;
+
+   auto itr = params.reverse? end_itr : begin_itr;
+   // since end_itr coudl potentially be past end of array, subtract one position
+   if (params.reverse)
+      itr--;
+
+   // this flag will be set to true when we are reversing and we end on the begin iterator
+   // if this is the case, 'more' will be set to false so client knows there is no more data
+   bool reverseEndBegin = false;
```
heifner: Use snake_case

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,58 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+   {
+      auto a = *itr;
+      result.accounts.push_back({a.name, a.creation_date});
+
+      cur_time = fc::time_point::now();
+      if (params.reverse && itr == begin_itr) {
+         reverseEndBegin = true;
+         break;
+      }
+      params.reverse? itr-- : itr++;
```
heifner: Prefer prefix to postfix: `--itr` and `++itr`

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,58 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
    return kv_database.get_backing_store();
 }
 
+read_only::get_all_accounts_result
+read_only::get_all_accounts( const get_all_accounts_params& params ) const
+{
+   get_all_accounts_result result;
+   result.more.reset();
```
heifner: `reset()` not needed.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,58 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+   auto begin_itr = params.lower_bound? accts.lower_bound(*params.lower_bound) : accts.begin();
+   auto end_itr = params.upper_bound? accts.upper_bound(*params.upper_bound) : accts.end();
+
+   if( std::distance(begin_itr, end_itr) < 0 )
+      return result;
+
+   auto itr = params.reverse? end_itr : begin_itr;
+   // since end_itr coudl potentially be past end of array, subtract one position
+   if (params.reverse)
+      itr--;
```
heifner: Prefer: `--itr;`

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3524,6 +3524,58 @@ eosio::chain::backing_store_type read_only::get_backing_store() const {
...
+
+   // this flag will be set to true when we are reversing and we end on the begin iterator
+   // if this is the case, 'more' will be set to false so client knows there is no more data
+   bool reverseEndBegin = false;
+
+   while(cur_time <= end_time
+         && result.accounts.size() < params.limit
+         && itr != end_itr)
+   {
+      auto a = *itr;
```
heifner: This can be `const auto& a` to avoid the copy.

scripts/blockchain_audit_tool.py
```diff
@@ -0,0 +1,243 @@
+#!/usr/bin/env python3
+
```
heifner, line 2: Add docs on arguments and output.

heifner (changes requested):

scripts/blockchain_audit_tool.py
```diff
@@ -0,0 +1,162 @@
...
+import sys
+import json
+
+rpc_endpt = "127.0.0.1:8888"
+if len(sys.argv) > 1:
+    rpc_endpt = sys.argv[1]
+
+page_size = 1024
+if len(sys.argv) > 2:
+    page_size = int(sys.argv[2])
```
linhuang-blockone: You can print out a help if arguments are invalid or with "-h".

scripts/blockchain_audit_tool.py
```diff
@@ -0,0 +1,162 @@
...
+# print out results
+try:
+    print("                          EOSIO BLOCKCHAIN AUDIT RESULTS\n")
+    print(f"     Server Version: {server_info_begin['server_full_version_string']}")
+    print(f"           Chain ID: {server_info_begin['chain_id']}")
+    print(f"   Head Block Start: {server_info_begin['head_block_num']}")
+    print(f"     Head Block End: {server_info_end['head_block_num']}")
+    print()
+    print("     ======  ACCOUNTS CODE ======")
+    print("Name            Priv    Code Hash")
```
linhuang-blockone: Use full word "Privilege" instead of "Priv"

praphael: Fixed

scripts/blockchain_audit_tool.py
praphael: Fixed

linhuang-blockone (approved):
Python part looks good.

heifner (approved):

PR #10773, opened 2021-09-30
Fix get_wasm_parameters_packed pack calculation size

victorj8:
## Change Description

Correct the size required for the packing of the wasm parameters output in the get_wasm_parameters_packed host function, this came to light when implementing a wrapper for it (on EOSIO.cdt side).

See also get_kv_parameters_packed for a similar implementation.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [x] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/webassembly/privileged.cpp
```diff
@@ -134,7 +134,7 @@ namespace eosio { namespace chain { namespace webassembly {
       auto& params = gpo.wasm_configuration;
       uint32_t version = std::min( max_version, uint32_t(0) );
 
-      auto s = fc::raw::pack_size( params );
+      auto s = fc::raw::pack_size( version ) + fc::raw::pack_size( params );
```
heifner: This will have to be part of a protocol feature change as this is a consensus level change.

heifner (changes requested):

libraries/chain/webassembly/privileged.cpp
victorj8: The protocol feature was added.

spoonincode:
~Is it worth making this a protocol feature? It seems like contracts will need to handle the case where the feature isn't activated regardless (probably just by adding a byte or two to the response)~

EDIT: ah I just went back and looked at it again, since the datastream is constructed with s instead of packed_parameters.size() you'll never be able to get a non-truncated serialization in the original impl

heifner:
Awaiting feedback on release versions this should be included in. @aclark-b1

heifner (approved):

PR #10811, opened 2021-10-13
get_all_accounts and blockchain_audit_tool

praphael:
## Change Description

Adds get_all_accounts RPC call to chain plugin, /v1/chain/get_all_accounts. This retrieves all accounts which exist on the chain, optionally restricted by JSON request parameters lower_bound and upper_bound which should be strings of length limited to chain name (13 characters), and contain only allowed characters. Clients should also specify limit field to restrict the maximum number returned. The call with return a JSON object with a field of accounts which is an array of objects with the name and creation_date. If more data exists, a more value will be set in the JSON response. the next name after the last name. Clients can also optionally reverse the order by setting reverse to true in the JSON request.

If accounts are created in between calls, there is no guarantee of consistency.

Also adds a Python script, blockchain_audit_tool.py, which uses this call in combination with other RPC calls to retrieve info on the chain, including all accounts which currently exist, and their creation date, last code update, and code hashes. Also lists account permissions, producer schedules, and a preview of MI/KV tables for each contract/account and their values.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [x] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

Adds RPC call /v1/chain/get_all_accounts

## Documentation Additions
- [x] Documentation Additions

Document new call /v1/chain/get_all_accounts and tool blockchain_audit_tool.py


spoonincode:
Since the results are potentially inaccurate if a block is applied during result gathering, should the script make more noise if it detects that? Maybe even refuse to print results in such a case?

praphael:
> Since the results are potentially inaccurate if a block is applied during result gathering, should the script make more noise if it detects that? Maybe even refuse to print results in such a case?

I could call get_all_accounts again at the end to make sure the # of accounts has not changed.

spoonincode:
You already gather the block id at the beginning and then at the end, I thought could just check those are equal. I realize it's still not perfect since the node could be a speculative node (and I don't recall there being an API call to check the mode).

praphael:
> You already gather the block id at the beginning and then at the end, I thought could just check those are equal. I realize it's still not perfect since the node could be a speculative node (and I don't recall there being an API call to check the mode).

The problem is for large numbers of accounts (>100 or so), the script takes too long to complete, and therefore would always fail.  I suppose it would make sense for the entire audit to run within a single block ... however this being a general tool at the moment I'm not sure making that a hard requirement would be a good idea.  But maybe we can issue a warning at least.

PR #10812, opened 2021-10-14
support creating TPM keys at a specific persistent handle 📦

spoonincode:
## Change Description

Prior to this change, `eosio-tpmtool` would persist created keys at the first available owner handle in the TPM. This simplicity is nice to avoid exposing details of TPM handles to users, but sometimes it can cause deployment headaches. Consider a boot or deployment script that always wants to create a BP key if it doesn't already exist. Using something like `eosio-tpmtool --list | wc` isn't always helpful because `eosio-tpmtool --list` will report all usable TPM keys: even keys it did not create itself -- keys that may used for other purposes -- making it hard to correlate the number of existing keys with whether a BP key needs to be created.

Instead of requiring the environment to further parse the output of `eosio-tpmtool -l` to determine if it needs to create a key, add the ability for `eosio-tpmtool` to persist to a specific handle. If that handle already has an object persisted there, the command will fail with error code 100. This allows a deployment script to do something like
```bash
bp_key_handle=0x81000E05

if key=$(./eosio-tpmtool --create --handle $bp_key_handle); then
   echo created BP key: $key
   # do something more interesting here, like upload the key somewhere or modify nodeos config.ini
elif [[ $? == 100 ]]; then
   echo BP key was already created
   # no need to do interesting things here
else
   echo eosio-tpmtool failed
fi
```

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/eosio-tpmtool/main.cpp
+++ b/programs/eosio-tpmtool/main.cpp
@@ -34,7 +34,7 @@ int main(int argc, char** argv) {
       ("handle", bpo::value<std::string>()->notifier([&](const std::string& a) {
          create_at_handle = strtol(a.c_str(), NULL, 0);
          FC_ASSERT(create_at_handle, "Creation handle argument is not an integer");
-      }), "Persist key at given TPM handle (by default, find first available owner handle)")
+      }), "Persist key at given TPM handle (by default, find first available owner handle). Returns error code 100 if key already exists.")
       ;
    bpo::variables_map varmap;
    try {
@@ -69,6 +69,10 @@ int main(int argc, char** argv) {
             std::cout << k.to_string() << std::endl;
 
       return 0;
+   }
+   catch(eosio::tpm::tpm_key_exists&) {
+      std::cerr << "Key already exists at given handle" << std::endl;
+      return 100;
    } FC_LOG_AND_DROP();
 
    return 1;
--- a/programs/eosio-tpmtool/main.cpp
+++ b/programs/eosio-tpmtool/main.cpp
@@ -15,6 +15,7 @@ int main(int argc, char** argv) {
    std::string tcti;
    std::vector<unsigned> pcrs;
    uint32_t attest_handle = 0;
+   uint32_t create_at_handle = 0;
 
    cli.add_options()
       ("help,h", bpo::bool_switch(&help)->default_value(false), "Print this help message and exit.")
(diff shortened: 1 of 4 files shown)


programs/eosio-tpmtool/main.cpp
```diff
@@ -30,6 +31,10 @@ int main(int argc, char** argv) {
          attest_handle = strtol(a.c_str(), NULL, 0);
          FC_ASSERT(attest_handle, "Attest handle argument is not an integer");
       }), "Certify creation of the new key via key with given TPM handle")
+      ("handle", bpo::value<std::string>()->notifier([&](const std::string& a) {
```
heifner, line 34: Is there a reason to use `std::string` and convert to long instead of just `bpo::value<uint32_t>` ?

heifner (changes requested):

programs/eosio-tpmtool/main.cpp
spoonincode: I want to allow passing the parameter as hex, e.g. 0x81000E05. I seem to recall that doesn't work if it just using uint32_t here, where strtol() is smart enough to autodetect the leading 0x if one exists. I can double check; been a long time maybe I'm misremembering.

spoonincode: Yes a simple boost program options driver:
```c++
         uint32_t bananas = 0;

         cli.add_options()
                 ("bananas", bpo::value<uint32_t>(&bananas), "Number of bananas");

         bpo::variables_map varmap;
         bpo::store(bpo::parse_command_line(argc, argv, cli), varmap);
         bpo::notify(varmap);

         std::cout << "number of bananas: " << bananas << std::endl;
```
yields:
```
$ ./dk --bananas 256
number of bananas: 256
$ ./dk --bananas 0x100
terminate called after throwing an instance of 'boost::wrapexcept<boost::program_options::invalid_option_value>'
  what():  the argument ('0x100') for option '--bananas' is invalid
```

Allowing hex input for these values is important, since persistent key handles are defined by the TPM spec as being 0x81xxxxxx, so you'd almost always enter handles manually in hex.

heifner: Thanks for the clarification.

heifner (approved):

Issue #7597, opened 2019-07-03
EOSIO Deprecations


arhag:
## Table tracking significant deprecations in EOSIO software

| Feature                                               | Date of Deprecation Notice |                    Deprecation Announcement Link                   | Date of Removal |                  Removal Announcement Link                 | Recommended Alternative                     |
|-------------------------------------------------------|:--------------------------:|:------------------------------------------------------------------:|:---------------:|:----------------------------------------------------------:|---------------------------------------------|
| history_plugin                                        |         2018/08/14         |     [v1.2.0](https://github.com/EOSIO/eos/releases/tag/v1.2.0)     |        -        |                              -                             | state_history_plugin                        |
| eosiocpp                                              |         2018/08/14         |     [v1.2.0](https://github.com/EOSIO/eos/releases/tag/v1.2.0)     |    2019/03/13   | [v1.7.0](https://github.com/EOSIO/eos/releases/tag/v1.7.0) | eosio.cdt                                   |
| RAM billing within notified contracts                 |         2018/08/28         |     [v1.2.3](https://github.com/EOSIO/eos/releases/tag/v1.2.3)     |    See notes    |                          See notes                         | See notes                                   |
| binaryen                                              |         2018/09/18         |     [v1.3.0](https://github.com/EOSIO/eos/releases/tag/v1.3.0)     |    2018/10/16   | [v1.4.0](https://github.com/EOSIO/eos/releases/tag/v1.4.0) | wabt                                        |
| Docker files                                          |         2018/11/14         |                                  -                                 |    2019/06/28   | [v1.8.0](https://github.com/EOSIO/eos/releases/tag/v1.8.0) | binary releases                             |
| Send-to-self authorization bypass                     |         2018/12/13         |     [v1.4.5](https://github.com/EOSIO/eos/releases/tag/v1.4.5)     |    See notes    |                          See notes                         | See notes                                   |
| Fedora 27, Mint 18, Amazon Linux 1, macOS High Sierra |         2019/04/03         |     [v1.7.1](https://github.com/EOSIO/eos/releases/tag/v1.7.1)     |    2019/06/28   | [v1.8.0](https://github.com/EOSIO/eos/releases/tag/v1.8.0) | Ubuntu 18.04 (or other supported platforms) |
[...]

PR #10816, opened 2021-10-15
cleos remove wasm to wast conversion 📦

heifner:
## Change Description

- Update `cleos` to not perform wasm to wast conversion.
- Users can use standard WebAssembly tooling to convert from wasm to wast if needed.
- Removal announcement: https://github.com/EOSIO/eos/issues/7597

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/chain/include/eosio/chain/wast_to_wasm.hpp
+++ b/libraries/chain/include/eosio/chain/wast_to_wasm.hpp
@@ -5,7 +5,5 @@
 namespace eosio { namespace chain {
 
 std::vector<uint8_t> wast_to_wasm( const std::string& wast );
-std::string  wasm_to_wast( const std::vector<uint8_t>& wasm, bool strip_names );
-std::string  wasm_to_wast( const uint8_t* data, uint64_t size, bool strip_names, bool check_limits = true );
 
 } } /// eosio::chain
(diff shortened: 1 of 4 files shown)


libraries/chain/include/eosio/chain/wast_to_wasm.hpp
```diff
@@ -6,6 +6,6 @@ namespace eosio { namespace chain {
 
 std::vector<uint8_t> wast_to_wasm( const std::string& wast );
 std::string  wasm_to_wast( const std::vector<uint8_t>& wasm, bool strip_names );
-std::string  wasm_to_wast( const uint8_t* data, uint64_t size, bool strip_names );
+std::string  wasm_to_wast( const uint8_t* data, uint64_t size, bool strip_names, bool check_limits = true );
```
spoonincode: can all the wasm_to_wast stuff be ripped out?

heifner: Yes. It is not used, so might as well.

heifner: Removed.

spoonincode: it's probably possible to purge more, down in the libraries/wasm-jit/Source/WAST (and include dir). That would be highly desirable, but maybe not critical at the moment

spoonincode (approved):

libraries/chain/include/eosio/chain/wast_to_wasm.hpp
spoonincode: ah, hm, can't remove all too much in there since unit tests since use wast

PR #10822, opened 2021-10-20
rodeos filter current_time fix 📦

heifner:
## Change Description

- Fix rodeos kv_table iterator--/++.
- Fix rodeos `current_time` for filters to return last block instead of first block time.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/rodeos/include/eosio/key_value.hpp
+++ b/libraries/rodeos/include/eosio/key_value.hpp
@@ -268,8 +268,8 @@ class kv_table {
 
       iterator(iterator&& other) :
          itr(std::exchange(other.itr, 0)),
-         itr_stat(std::move(other.itr_stat)),
-         index(std::move(other.index))
+         itr_stat(other.itr_stat),
+         index(other.index)
       {}
 
       ~iterator() {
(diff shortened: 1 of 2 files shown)


libraries/rodeos/include/eosio/key_value.hpp
```diff
@@ -268,7 +268,8 @@ class kv_table {
 
       iterator(iterator&& other) :
          itr(std::exchange(other.itr, 0)),
-         itr_stat(std::move(other.itr_stat))
+         itr_stat(std::move(other.itr_stat)),
+         index(std::move(other.index))
```
huangminghuang: `index` is a const pointer, there's no point to do std::move()

heifner: yeah, I know. But I was matching cdt code since they are the same.

heifner: I plan on creating an issue to get kv_table in rodeos to match impl in cdt.

huangminghuang (approved):

libraries/rodeos/include/eosio/key_value.hpp
heifner: Removed moves that don't move

Issue #6718, opened 2019-02-10
Add support for uncompressed public key type


conr2d:
## Background

Blockchains including EOSIO usually support ECDSA secp256k1 to verify their transactions. However, in some blockchains, the result of recovering key from signature is different from EOSIO, because they use uncompressed key (65 bytes) instead of compressed one (33 bytes) which EOSIO uses.

As far as I know, the main strategy to improve EOSIO scalability is IBC among EOSIO-based sidechains or fork chains. If uncompressed public key type is supported in EOSIO smart contract, we can achieve better interoperability among heterogeneous blockchain networks.

## Proposal

Add intrinsic API for uncompressed public key type

### Available options

0. Common

New types `capi_public_key_point` and `eosio::public_key_point` are added

``` c++
struct capi_public_key_point {
   char data[66];
};

namespace eosio {
   struct public_key_point {
      unsigned_int type;
      std::array<char,65> data;
   };
}
```

1. Add decompression API converts compressed public key (`eosio::public_key`) to uncompressed one (`eosio::public_key_point`)

``` c++
::decompress_key (const char* data, uint32_t length, char* pubkey_point)
```

2. Add conversion API between compressed public key and uncompressed one, the size of `outlen` determines which type of public key (compressed/uncompressed) needs to be serialized

``` c++
::serialize_key (const char*data, uint32_t datalen, char* out, uint32_t outlen)
```
3. Add feature to `recover_key` for recovering uncompressed public key from signature, the size of `publen` determines which type of public key needs to be recovered

``` c++
::recover_key (const capi_checksum256* digest, const char* sig, size_t siglen, char* pub, size_t publen)
```

### Other considerations

Currently, wrapper classes of `secp256k1` and `secp256r1` provide recover() method which returns only compressed public key. If 3rd option is adopted, there can be two way to make it, one is recovering compressed and decompress it, or the other is adding ability to recover uncompressed public key to wrapper classes.

---

The next commits are very early stage impelmentation of this proposal.

https://github.com/conr2d/eos/commit/04584161c9b3077e223f9b7d0edfa0b543c195e0
https://github.com/conr2d/fc/commit/34cd44a1c3940d441976f874fa6e75dd6b185ded

conr2d:
IMHO, I like the 3rd option most, because it doesn't add an API explicitly and it wouldn't break existing behavior in most cases. (`publen` in `::recover_key()` passed is usually 38)

PR #10831, opened 2021-10-24
Add RECOVER_KEY_SAFE protocol feature

conr2d:
## Change Description

Implement [EEP-9](https://github.com/EOSIO/EEPs/blob/master/EEPS/eep-9.md).

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [x] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


conr2d:
Can you assign anyone who may review this? It would be good if I can get a review before writing test cases.

- API design
- Additional feature for recovering uncompressed key (#6718)

According to Bitcoin secp256k1 convention, recovery IDs in the range of 27 to 30 are for uncompressed public key, but 31 to 34 for compressed one. For better support for EVM on EOSIO, what about recovering an uncompressed public key (public key point) in case of recover ID 27-30? For this, it needs to merge implementation for missing `serialize_ecc_point()`. (EOSIO/fc#215)

libraries/chain/webassembly/crypto.cpp
```diff
@@ -65,6 +66,50 @@ namespace eosio { namespace chain { namespace webassembly {
       }
    }
 
+   int32_t interface::recover_key_safe( legacy_ptr<const fc::sha256> digest,
+                                        legacy_span<const char> sig,
+                                        legacy_ptr<char> pub,
+                                        legacy_ptr<uint32_t> publen) const {
```
heifner: These do not have to be `legacy_`. `const fc::sha256* digest, span<const char> sig, span<char> pub, uint32_t* publen`

libraries/chain/webassembly/crypto.cpp
```diff
@@ -65,6 +66,50 @@ namespace eosio { namespace chain { namespace webassembly {
...
+
+         fc::crypto::public_key recovered;
+         try {
+            recovered = fc::crypto::public_key(s, *digest, false);
+         } catch (...) {
+            return error_code::invalid_signature_data;
+         }
+
+         auto packed_pubkey = fc::raw::pack(recovered);
+         if( *publen.get() < packed_pubkey.size() ) {
```
heifner: I think this should be `pub.size()`

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -338,6 +338,20 @@ namespace webassembly {
...
          int32_t recover_key(legacy_ptr<const fc::sha256> digest, legacy_span<const char> sig, legacy_span<char> pub) const;
 
+         /**
+         * Calculates the public key used for a given signature on a given digest.
+         * Unlike @ref recover_key it returns an error code rather than throws an exception with invalid input.
+         *
+         * @ingroup crypto
+         * @param digest - digest of the message that was signed.
+         * @param sig - signature.
+         * @param[out] pub - pointer to output buffer for the public key result.
```
heifner: [inout]

libraries/chain/include/eosio/chain/webassembly/interface.hpp
```diff
@@ -338,6 +338,20 @@ namespace webassembly {
...
+         * @param[inout] publen - pointer to an integer initially set to the expected size of output, and overwritten by the written size
```
heifner: @param[out] publen - pointer to the size of output buffer of the public key `pub`. This value can be used when `error_code::insufficient_output_buffer` is returned to determine required size of `pub`.

libraries/chain/include/eosio/chain/webassembly/error_codes.hpp
```diff
@@ -0,0 +1,14 @@
...
+
+namespace eosio { namespace chain { namespace webassembly { namespace error_codes {
+
+   enum recover_key_safe : int32_t {
+      undefined = -1,
+      none = 0,
+      invalid_signature_format,
+      unactivated_key_type,
+      invalid_signature_data,
+      insufficient_output_buffer,
```
heifner: Please explicitly list each value.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -280,6 +280,7 @@ REGISTER_LEGACY_HOST_FUNCTION(get_active_producers);
 // crypto api
 REGISTER_LEGACY_CF_HOST_FUNCTION(assert_recover_key);
 REGISTER_LEGACY_CF_HOST_FUNCTION(recover_key);
+REGISTER_LEGACY_CF_HOST_FUNCTION(recover_key_safe);
```
heifner: REGISTER_CF_HOST_FUNCTION

heifner (changes requested):
Sorry it took so long.

libraries/chain/include/eosio/chain/webassembly/interface.hpp
conr2d:
``` c++
int recover_key_safe(const capi_checksum256* digest, const char* sig, size_t sigsize,
  char* pub,
  size_t* publen);

// after review
int recover_key_safe(const capi_checksum256* digest, const char* sig, size_t sigsize,
  char* pub,
  size_t pubsize,
  size_t* publen);
```
In that case, is the above function signature for intrinsic right? AFAIK, `span<char>` is translated into `{char*,size_t}`. I considered reusing the last argument `size_t* publen` for the size of the given output buffer to store written size.

conr2d:
@heifner 
Thank you for the review. What do you think of recovering uncompressed public key for recovery ID from 27 to 30?

libraries/chain/include/eosio/chain/webassembly/interface.hpp
heifner: This looks good.

libraries/chain/webassembly/crypto.cpp
conr2d: Failed to compile because `const fc::sha256*` is not whitelisted type.

heifner:
> @heifner Thank you for the review. What do you think of recovering uncompressed public key for recovery ID from 27 to 30?

Sorry for the slow response. At this time we would like the uncompressed public key (if added) to come in under a different PR to keep this particular change focused on just having a recover key that does not assert.

libraries/chain/webassembly/crypto.cpp
```diff
@@ -65,6 +66,50 @@ namespace eosio { namespace chain { namespace webassembly {
...
+                                        legacy_ptr<char> pub,
+                                        legacy_ptr<uint32_t> publen) const {
+      using error_code = eosio::chain::webassembly::error_codes::recover_key_safe;
+
+      try {
+         fc::crypto::signature s;
+         try {
+            datastream<const char*> ds( sig.data(), sig.size() );
+            fc::raw::unpack(ds, s);
+         } catch (...) {
```
heifner: After some discussion, we would like these `...` catches to be changed to catch fc exception which should be the only exceptions thrown from any of these calls.

libraries/chain/webassembly/crypto.cpp
swatanabe: `legacy_*` should never be used for new intrinsics.  It has extremely weird behavior when there are multiple output parameters, due to possibility of aliasing.  Host functions with multiple output parameters using the legacy system require special manual control to ensure consistent behavior.  Also, a buffer must always use `span` so that the implementation knows to check memory bounds.  Failure to do so results in undefined behavior.

PR #10833, opened 2021-10-24
allow empty table 📦

huangminghuang:
## Change Description

This PR allows the `get_kv_table_rows` api endpoint to accept empty table value. In this case, the result rows would return a vector of key and data which contains the hex string of the key and data.
 

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -2385,13 +2385,14 @@ struct kv_iterator_ex {
    /// @pre ! is_end()
    fc::variant get_value_and_maybe_payer_var() const {
       fc::variant result = get_value_var();
-      if (context.p.table.empty()) {
-         return fc::mutable_variant_object("data", std::move(result))("key", get_key_hex_string());
-      }
-      if (context.p.show_payer) {
+      if (context.p.show_payer || context.p.table.empty()) {
+         auto r = fc::mutable_variant_object("data", std::move(result));
          auto maybe_payer = base->kv_it_payer();
-         std::string payer = maybe_payer.has_value() ? maybe_payer.value().to_string() : "";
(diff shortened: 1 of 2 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2383,6 +2390,9 @@ struct kv_iterator_ex {
          std::string payer = maybe_payer.has_value() ? maybe_payer.value().to_string() : "";
          return fc::mutable_variant_object("data", std::move(result))("payer", payer);
       }
+      else if (context.p.table.empty()) {
+         return fc::mutable_variant_object("data", std::move(result))("key", get_key_hex_string());
```
linhuang-blockone: Is `result` in hex format?

linhuang-blockone (approved):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2383,6 +2390,9 @@ struct kv_iterator_ex {
          std::string payer = maybe_payer.has_value() ? maybe_payer.value().to_string() : "";
          return fc::mutable_variant_object("data", std::move(result))("payer", payer);
       }
+      else if (context.p.table.empty()) {
```
heifner: Shouldn't this be an `if` instead of an `else if`

heifner (changes requested):

heifner (approved):

PR #10845, opened 2021-10-28
Add new loggers for dumping transaction trace 📦

heifner:
## Change Description
Current loggers `transaction_success_tracing` and `transaction_failure_tracing` in file logging.json are created for displaying brief transaction trace information. New loggers that could be used to dump entire transaction trace or entire transaction for all trxs including failed are requested. 

Everywhere we currently log REJECTING or ACCEPTING we should log to new loggers with either the trace or the transaction. We should  log to new loggers with the transaction for the case of  an invalid signature.

New added loggers `transaction_trace_success` and `transaction_trace_failure` have default logging level `info` in file logging.json. To enable them, their logging levels need to be changed to `debug`.

Two examples of log output from new trace loggers:

1. creating an account named as "bob" succeeded  

[...]

PR #10852, opened 2021-10-28
Fix exceptions being logged issue - develop

nickjjzhao:
[EPE-1616](https://blockone.atlassian.net/browse/EPE-1616)

This PR was created to remove exceptions from the trace. Exceptions shouldn't be logged.

https://github.com/EOSIO/eos/blob/release/2.2.x/plugins/producer_plugin/producer_plugin.cpp#L631 calls `send_response` with trace when return_failure_trace=true.

`send_response` logs that the transaction was rejected or accepted according to what is passed in. This needs to check if the trace has an exception and log the correct thing.  The trace should be logged and not the exception when the trace is available. 

This issue was originally found and fixed in [PR#10845](https://github.com/EOSIO/eos/pull/10845).

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
+++ b/plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
@@ -181,7 +181,7 @@ public:
                q_item i;
                if( self->queue_.pop_and_pause(i) ) {
                   auto exception_handler = [&i, &prod_plugin=self->prod_plugin_](fc::exception_ptr ex) {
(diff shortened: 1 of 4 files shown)


heifner:
Superseded by https://github.com/EOSIO/eos/pull/10867

PR #10860, opened 2021-10-30
net_plugin bugfix for resetting requested range

dimas1185:
## Change Description

test fail example:
https://buildkite.com/EOSIO/eosio-test-stability/builds/319#b82a137d-3c4d-48d6-96cb-e48d3af9496d

investigation details:
When net_plugin sends `sync_request_message{0,0}` in response to block<lib it doesn't reset `sync_last_requested_num`. After that when node gets `notice_message` it ignores it instead of doing catchup since head is less than `sync_last_requested_num`.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -168,7 +168,10 @@ namespace eosio {
       void sync_update_expected( const connection_ptr& c, const block_id_type& blk_id, uint32_t blk_num, bool blk_applied );
       void recv_handshake( const connection_ptr& c, const handshake_message& msg );
       void sync_recv_notice( const connection_ptr& c, const notice_message& msg );
-      inline void reset_last_requested_num() {
+      inline void reset_last_requested_num(bool lock = false) {
+         std::unique_lock<std::mutex> g( sync_mtx, std::defer_lock_t{} );
+         if (lock)
+            g.lock();
          sync_last_requested_num = 0;
       }
    };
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2968,6 +2971,7 @@ namespace eosio {
             peer_ilog( this, "received block ${n} less than ${which}lib ${lib}",
                        ("n", blk_num)("which", blk_num < last_sent_lib ? "sent " : "")
                        ("lib", blk_num < last_sent_lib ? last_sent_lib : lib) );
+            my_impl->sync_master->reset_last_requested_num();
```
heifner: `sync_mtx` needs acquired to set `sync_last_requested_num`.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
dimas1185: done and added another reset as i found another test case where we are getting unlinkable block, sending handshake and after lib notice while starting syncing nodeos just logs something like:
`ignoring request, head is 641 last req = 696, source connection 2`

dimas1185: @heifner stability test (except mac) with 500 runs passed:
https://buildkite.com/EOSIO/eosio-test-stability/builds/343

heifner (approved):

PR #10829, opened 2021-10-23
EPE-1563 nodeos (amqp_trx_plugin)/cleos being able to get amqp address from an environment variable 📦

zma:
## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

The changes here add the support to nodeos (amqp_trx_plugin) and cleos to accept the amqp address string from an environment variable.

for nodeos:

```
  --amqp-trx-address arg                AMQP address: Format:
                                        amqp://USER:PASSWORD@ADDRESS:PORT
                                        Will consume from amqp-trx-queue-name
                                        (amqp-trx-queue-name) queue.
                                        If --amqp-trx-address is not specified,
                                        will use the value from the environment
                                        variable EOSIO_AMQP_ADDRESS.
```

for cleos:

```
  --amqp TEXT (Env:EOSIO_AMQP_ADDRESS)
                              The ampq URL where AMQP is running amqp://USER:PASSWORD@ADDRESS:PORT
```

And the username for amqp connection is also masked together with the password.

PR #10861, opened 2021-10-31
EPE-1577 rodeos getting stream addresses for queues or exchanges from environment variables 📦

zma:
This change makes rodeos be able to get the rabbits stream addresses for queues or exchanges from environment variables.

```
  --stream-rabbits arg                  Addresses of RabbitMQ queues to stream
                                        to. Format: amqp://USER:PASSWORD@ADDRES
                                        S:PORT/QUEUE[/STREAMING_ROUTE, ...].
                                        Multiple queue addresses can be
                                        specified with ::: as the delimiter,
                                        such as "amqp://u1:p1@amqp1:5672/queue1
                                        :::amqp://u2:p2@amqp2:5672/queue2".If
                                        this option is not specified, the value
                                        from the environment variable
                                        EOSIO_STREAM_RABBITS will be used.
  --stream-rabbits-exchange arg         Addresses of RabbitMQ exchanges to
                                        stream to. amqp://USER:PASSWORD@ADDRESS
                                        :PORT/EXCHANGE[::EXCHANGE_TYPE][/STREAM
                                        ING_ROUTE, ...]. Multiple queue
                                        addresses can be specified with ::: as
                                        the delimiter, such as
                                        "amqp://u1:p1@amqp1:5672/exchange1:::am
                                        qp://u2:p2@amqp2:5672/exchange2".If
                                        this option is not specified, the value
                                        from the environment variable
                                        EOSIO_STREAM_RABBITS_EXCHANGE will be
                                        used.
```

This PR is done for rodeos following the changes for amqp_trx_plugin in nodeos and for cleos - #10829 .

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/programs/rodeos/streamer_plugin.cpp
+++ b/programs/rodeos/streamer_plugin.cpp
@@ -72,13 +72,13 @@ streamer_plugin::~streamer_plugin() {}
 void streamer_plugin::set_program_options(options_description& cli, options_description& cfg) {
    auto op = cfg.add_options();
    op("stream-rabbits", bpo::value<std::vector<string>>()->composing(),
-      "RabbitMQ Streams to queues if any; Format: amqp://USER:PASSWORD@ADDRESS:PORT/QUEUE[/STREAMING_ROUTE, ...]. Additional list of streams delimited by ::: can be specified in the environment variable "
+      "RabbitMQ Streams to queues if any; Additional list of streams delimited by ::: can be specified in the environment variable "
       EOSIO_STREAM_RABBITS_ENV_VAR
-      ".");
+      ". Format: amqp://USER:PASSWORD@ADDRESS:PORT/QUEUE[/STREAMING_ROUTE, ...].");
    op("stream-rabbits-exchange", bpo::value<std::vector<string>>()->composing(),
-      "RabbitMQ Streams to exchanges if any; Format: amqp://USER:PASSWORD@ADDRESS:PORT/EXCHANGE[::EXCHANGE_TYPE][/STREAMING_ROUTE, ...]. Additional list of streams delimited by ::: can be specified in the environment variable "
+      "RabbitMQ Streams to exchanges if any; Additional list of streams delimited by ::: can be specified in the environment variable "
       EOSIO_STREAM_RABBITS_EXCHANGE_ENV_VAR
-      ".");
+      ". Format: amqp://USER:PASSWORD@ADDRESS:PORT/EXCHANGE[::EXCHANGE_TYPE][/STREAMING_ROUTE, ...]");
    op("stream-rabbits-immediately", bpo::bool_switch(&my->publish_immediately)->default_value(false),
       "Stream to RabbitMQ immediately instead of batching per block. Disables reliable message delivery.");
    op("stream-loggers", bpo::value<std::vector<string>>()->composing(),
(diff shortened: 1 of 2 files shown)


mjk90 (approved):

programs/rodeos/streamer_plugin.cpp
```diff
@@ -99,15 +103,42 @@ void streamer_plugin::plugin_initialize(const variables_map& options) {
...
+      auto split_env_var = [](const char* env_name, std::vector<std::string>& splits) {
+         if (!std::getenv(env_name)) {
+            ilog("Environment variable ${s} for AMQP is not set. Skip it.", ("s", env_name));
+            return;
+         }
+
+         std::string env{std::getenv(env_name)};
+         std::string delimiter{":::"};
+         size_t pos = 0;
+         std::string split;
```
heifner: Should just use `boost::split`.
https://www.boost.org/doc/libs/1_77_0/doc/html/string_algo/usage.html

programs/rodeos/streamer_plugin.cpp
```diff
@@ -72,9 +72,13 @@ streamer_plugin::~streamer_plugin() {}
...
    op("stream-rabbits", bpo::value<std::vector<string>>()->composing(),
-      "RabbitMQ Streams to queues if any; Format: amqp://USER:PASSWORD@ADDRESS:PORT/QUEUE[/STREAMING_ROUTE, ...]");
+      "RabbitMQ Streams to queues if any; Additional list of streams delimited by ::: can be specified in the environment variable "
+      EOSIO_STREAM_RABBITS_ENV_VAR
+      ". Format: amqp://USER:PASSWORD@ADDRESS:PORT/QUEUE[/STREAMING_ROUTE, ...].");
    op("stream-rabbits-exchange", bpo::value<std::vector<string>>()->composing(),
-      "RabbitMQ Streams to exchanges if any; Format: amqp://USER:PASSWORD@ADDRESS:PORT/EXCHANGE[::EXCHANGE_TYPE][/STREAMING_ROUTE, ...]");
+      "RabbitMQ Streams to exchanges if any; Additional list of streams delimited by ::: can be specified in the environment variable "
+      EOSIO_STREAM_RABBITS_EXCHANGE_ENV_VAR
+      ". Format: amqp://USER:PASSWORD@ADDRESS:PORT/EXCHANGE[::EXCHANGE_TYPE][/STREAMING_ROUTE, ...]");
```
heifner: This doesn't make it clear (at least to me) the syntax for multiple. An example might help.

heifner (changes requested):

programs/rodeos/streamer_plugin.cpp
zma: Changed to the semantics we discussed. The PR description has a copy of the help text file.

programs/rodeos/streamer_plugin.cpp
zma: It seems `boost::split` predicate is on character level. We want to split the string by the string `:::`.

I improved the logic a little though to use `std::sregex_token_iterator` which is easier to read.

heifner (approved):

PR #10872, opened 2021-11-03
nodeos min-initial-block-num option 📦

huangminghuang:
## Change Description

This PR adds `min-initial-block-num` option to chain plugin so that nodeos could refuse to load any state from prior to the specified block number

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -620,7 +620,7 @@ struct controller_impl {
       uint32_t lib_num = (blog.head() ? blog.head()->block_num() : fork_db.root()->block_num);
 
       EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller latest irreversible block "
-      "at block number ${lib_num}, which is smaller than the minimum required ${required}", ("block_num", lib_num)("required",this->conf.min_initial_block_num) );
+      "at block number ${lib_num}, which is smaller than the minimum required ${required}", ("lib_num", lib_num)("required",this->conf.min_initial_block_num) );
 
       auto header_itr = validate_db_version( db );
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -619,8 +619,8 @@ struct controller_impl {
    void init(std::function<bool()> check_shutdown, bool clean_startup) {
       uint32_t lib_num = (blog.head() ? blog.head()->block_num() : fork_db.root()->block_num);
 
-      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller head "
-      "at block number ${block_num}, which is smaller than the minimum required ${required}", ("block_num", fork_db.head()->block_num)("required",this->conf.min_initial_block_num) );
+      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller latest irreversible block "
+      "at block number ${lib_num}, which is smaller than the minimum required ${required}", ("block_num", lib_num)("required",this->conf.min_initial_block_num) );
 
       auto header_itr = validate_db_version( db );
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -619,6 +619,9 @@ struct controller_impl {
    void init(std::function<bool()> check_shutdown, bool clean_startup) {
       uint32_t lib_num = (blog.head() ? blog.head()->block_num() : fork_db.root()->block_num);
 
+      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller head "
+      "at block number ${block_num}, which is smaller than the minimum required ${required}", ("block_num", fork_db.head()->block_num)("required",this->conf.min_initial_block_num) );
+
       auto header_itr = validate_db_version( db );
 
       {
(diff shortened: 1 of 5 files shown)


libraries/chain/controller.cpp
```diff
@@ -619,6 +619,9 @@ struct controller_impl {
    void init(std::function<bool()> check_shutdown, bool clean_startup) {
       uint32_t lib_num = (blog.head() ? blog.head()->block_num() : fork_db.root()->block_num);
 
+      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller head "
+      "at block number ${block_num}, which is smaller than the minimum required ${required}", ("block_num", fork_db.head()->block_num)("required",this->conf.min_initial_block_num) );
```
heifner: This reports and uses the term head when lib is used. Also the variable name and option description should be clear that this feature is for lib and not head.

heifner (changes requested):
lib vs head comment.

nksanthosh (commented):
Will the implementation work with all 4 options (irrespective of how state is loaded)

1. Replay
2. From genesis
3. Snapshot
4. Load binary state file

Has it been manually tested against all 4 options?

Either add automated tests for all four options if it is quick or we can prioritize and add automated tests as a separate PR immediately following this one.

huangminghuang:
> Will the implementation work with all 4 options (irrespective of how state is loaded)
> 
> 1. Replay
> 2. From genesis
> 3. Snapshot
> 4. Load binary state file
> 
> Has it been manually tested against all 4 options?
> 
> Either add automated tests for all four options if it is quick or we can prioritize and add automated tests as a separate PR immediately following this one.

Unit tests are available to test against each scenario.

libraries/chain/controller.cpp
```diff
@@ -619,8 +619,8 @@ struct controller_impl {
    void init(std::function<bool()> check_shutdown, bool clean_startup) {
       uint32_t lib_num = (blog.head() ? blog.head()->block_num() : fork_db.root()->block_num);
 
-      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller head "
-      "at block number ${block_num}, which is smaller than the minimum required ${required}", ("block_num", fork_db.head()->block_num)("required",this->conf.min_initial_block_num) );
+      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller latest irreversible block "
+      "at block number ${lib_num}, which is smaller than the minimum required ${required}", ("block_num", lib_num)("required",this->conf.min_initial_block_num) );
```
linhuang-blockone: "block_num" should be "lib_num"

libraries/chain/controller.cpp
```diff
@@ -619,6 +619,9 @@ struct controller_impl {
    void init(std::function<bool()> check_shutdown, bool clean_startup) {
       uint32_t lib_num = (blog.head() ? blog.head()->block_num() : fork_db.root()->block_num);
 
+      EOS_ASSERT( lib_num >= this->conf.min_initial_block_num, misc_exception, "Controller latest irreversible block "
+      "at block number ${lib_num}, which is smaller than the minimum required ${required}", ("block_num", lib_num)("required",this->conf.min_initial_block_num) );
```
victorj8: I think this should be ${block_num} instead of ${lib_num}

unittests/restart_chain_tests.cpp
```diff
@@ -205,6 +205,43 @@ BOOST_AUTO_TEST_CASE(test_existing_state_without_block_log) {
...
+   blocks.push_back(chain.produce_block());
+   blocks.push_back(chain.produce_block());
+   blocks.push_back(chain.produce_block());
+   chain.close();
+
+   auto cfg                  = chain.get_config();
+   cfg.min_initial_block_num = 100;
+   remove_existing_blocks(cfg);
+
+   // restarting chain with no block log and no genesis
```
linhuang-blockone, line 237: Can you add a test for normal cases `lib == cfg.min_initial_block_num`  and `lib > cfg.min_initial_block_num`?

linhuang-blockone (approved):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -458,6 +458,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
          ("export-reversible-blocks", bpo::value<bfs::path>(),
            "export reversible block database in portable format into specified file and then exit")
          ("snapshot", bpo::value<bfs::path>(), "File to read Snapshot State from")
+         ("min-initial-block-num", bpo::value<uint32_t>()->default_value(0), "minimum latest irreversible block number to load, "
+         "nodeos would refuse to load any state from prior to that block number")
```
heifner: Should not use hard coded "nodeos". Either don't mention "nodeos" or use the executable name variable. I prefer just not mentioning it. Also I think it should be "will" not "would".

Maybe something like:
"minimum last irreversible block (lib) number, fail to start if state/snapshot lib is prior to specified"

heifner (approved):

nksanthosh (approved):

PR #10876, opened 2021-11-04
Threaded block signing 📦

jgiszczak:
## Change Description
Thread block signing.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/controller.cpp
```diff
@@ -2592,20 +2592,51 @@ block_state_ptr controller::finalize_block( const signer_callback_type& signer_c
...
                       const flat_set<digest_type>& cur_features,
                       const vector<digest_type>& new_features )
-                  {},
-                  signer_callback
+                  {}
               );
 
    my->pending->_block_stage = completed_block{ bsp };
 
+   if( my->block_sign_fut.valid() ) {
```
heifner: This is not happening until the next block. I don't think we want to delay emitting the signed block by 500ms. This all should happen as soon as the block is signed. Can all this be moved into the lambda passed to block_sign_pool?

libraries/chain/controller.cpp
```diff
@@ -2592,20 +2592,51 @@ block_state_ptr controller::finalize_block( const signer_callback_type& signer_c
 
    auto& ab = std::get<assembled_block>(my->pending->_block_stage);
 
+   auto pfa = *ab._pending_block_header_state.prev_activated_protocol_features.get();
```
heifner: `prev_activated_protocol_features` is a shared_ptr, why not just pass the shared_ptr around?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2346,7 +2346,7 @@ void producer_plugin_impl::produce_block() {
    }
 
    //idump( (fc::time_point::now() - chain.pending_block_time()) );
-   block_state_ptr pending_blk_state = chain.finalize_block( [&]( const digest_type& d ) {
+   block_state_ptr pending_blk_state = chain.finalize_block( [relevant_providers]( const digest_type& d ) {
```
heifner: `relevant_providers` can be moved into the lambda.

libraries/chain/controller.cpp
```diff
@@ -2592,20 +2592,51 @@ block_state_ptr controller::finalize_block( const signer_callback_type& signer_c
```
heifner, line 2544: `finalize_block` should take `signer_callback_type` by value.

libraries/chain/controller.cpp
```diff
@@ -2592,20 +2592,51 @@ block_state_ptr controller::finalize_block( const signer_callback_type& signer_c
...
+      }
+
+      my->emit( accepted_block, result );
+      my->log_irreversible();
+
+   }
+
+   my->block_sign_fut = async_thread_pool( my->block_sign_pool.get_executor(),
+                                           [pfa, pfs, bsp, signer_callback]() mutable {
+                                              (*bsp).sign( bsp->block, signer_callback );
```
heifner: signer_callback should be moved into the lambda

libraries/chain/block_header_state.cpp
```diff
@@ -406,6 +401,8 @@ namespace eosio { namespace chain {
       additional_signatures = std::move(sigs);
 
       verify_signee();
+
+      h->producer_signature = header.producer_signature;
```
heifner: Why pass in signed_block_header for the same object? This should have already been set on line 398.

heifner (changes requested):

libraries/chain/block_header_state.cpp
huangminghuang: The `h` here is coming from producer_plugin.cpp:2637 which is actually a member of `block_state`, so they are different. However, it seems unnecessary to convert from `signed_block_ptr` to `std::share_ptr<signed_block_header>` for the declaration of `h`.

libraries/chain/controller.cpp
jgiszczak: The signal emission is thread-safe but `log_irreversible()` uses `boost::multi_index_container` in the fork database which is not remotely thread-safe.  Emitting the block to the network early results in the peculiar case of the network possibly having an irreversible block which the local node does not consider irreversible for hundreds of milliseconds.  If a health monitor in a single producer network happens to shut down and restart a producing node in that window without activating a standby producer, the network will irreconcilably fork as the active producer essentially double-produces a block, having lost its own prior last block.

libraries/chain/block_header_state.cpp
jgiszczak: `block_header_state.hpp` is unaware of the `signed_block` type.  It only knows about `signed_block_header`, for which a typedef/using shared pointer type doesn't exist.

libraries/chain/controller.cpp
jgiszczak: Done.

libraries/chain/controller.cpp
jgiszczak: Done.

plugins/producer_plugin/producer_plugin.cpp
jgiszczak: Done.

libraries/chain/controller.cpp
jgiszczak: Done.

libraries/chain/controller.cpp
heifner: I'm envisioning it working as follows:
- kick off a lambda to sign and do whatever else is possible in a diff thread
- When signing done it will post back to the main thread at high priority to finish out the `log_irreversible()` and `emit`.
- Indicate block is complete and pause on the next block at that location if the previous is still signing (should not happen in practice as it should not take more than 500ms to sign, although we should setup a test that causes it to happen by having a very slow sign proces..)

libraries/chain/block_header_state.cpp
heifner: Then have it take a `signed_block_header& h` instead.

libraries/chain/controller.cpp
```diff
@@ -2585,27 +2585,58 @@ void controller::start_block( block_timestamp_type when,
                     block_status::incomplete, std::optional<block_id_type>() );
 }
 
-block_state_ptr controller::finalize_block( const signer_callback_type& signer_callback ) {
+block_state_ptr controller::finalize_block( const signer_callback_type signer_callback ) {
```
heifner: `signer_callback` should be non-const so you can move from it.

libraries/chain/controller.cpp
```diff
@@ -1,3 +1,4 @@
+#include "eosio/vm/function_traits.hpp"
```
heifner: I don't see why this is needed.

libraries/chain/controller.cpp
```diff
@@ -2079,16 +2092,47 @@ struct controller_impl {
 
    deque<transaction_metadata_ptr> abort_block() {
       deque<transaction_metadata_ptr> applied_trxs;
+      
       if( pending ) {
          uint32_t block_num = pending->get_block_num();
-         applied_trxs = pending->extract_trx_metas();
+         ilog("abortig pending block ${block_num}", ("block_num", block_num));
```
heifner: Remove or change to `dlog`. Also aborting is misspelled.

libraries/chain/controller.cpp
```diff
@@ -2079,16 +2092,47 @@ struct controller_impl {
...
+         applied_trxs.insert(applied_trxs.begin(), trxs.begin(), trxs.end());
          pending.reset();
          protocol_features.popped_blocks_to( head->block_num );
          emit( self.block_abort, block_num );
       }
+
+      if (fork_db.is_signing_failed_block_head(signing_failed_blocknum)) {
+         signing_failed_blocknum = 0;
+         auto poped = pop_block();
+         ilog("abortig unsigned block ${block_num}", ("block_num", poped->block_num));
```
heifner: Remove or change to `dlog`. Also aborting is misspelled.

libraries/chain/controller.cpp
```diff
@@ -2079,16 +2092,47 @@ struct controller_impl {
...
+         emit(self.block_abort, poped->block_num);
+      }
+
       return applied_trxs;
    }
 
+   void abort_unsigned_block() {
+      if( pending ) {
+         uint32_t block_num = pending->get_block_num();
+         ilog("abortig pending block ${block_num}", ("block_num", block_num));
```
heifner: Remove or change to `dlog`. Also aborting is misspelled.

libraries/chain/controller.cpp
```diff
@@ -2079,16 +2092,47 @@ struct controller_impl {
...
+         pending.reset();
+         protocol_features.popped_blocks_to( head->block_num );
+         emit( self.block_abort, block_num );
+      }
+      auto db_head = fork_db.head();
+      if (db_head->block && db_head->block->producer_signature == signature_type()) {
+         auto poped = pop_block();
+         ilog("abortig unsigned block ${block_num}", ("block_num", poped->block_num));
```
heifner: Remove or change to `dlog`. Also aborting is misspelled.

libraries/chain/controller.cpp
```diff
@@ -2588,28 +2632,44 @@ void controller::start_block( block_timestamp_type when,
...
+                  {}
               );
 
-   my->pending->_block_stage = completed_block{ bsp };
-
-   return bsp;
+   my->pending->_block_stage = completed_block{bsp};
+
+   return async_thread_pool(my->block_sign_pool.get_executor(),
+                            [bsp, my = my.get(), block_num = bsp->block_num, digest = bsp->sig_digest(),
```
heifner: Remove `digest = bsp->sig_digest()` and instead call that from inside the lambda/thread-pool

libraries/chain/controller.cpp
```diff
@@ -2079,16 +2092,47 @@ struct controller_impl {
...
+   void abort_unsigned_block() {
+      if( pending ) {
+         uint32_t block_num = pending->get_block_num();
+         ilog("abortig pending block ${block_num}", ("block_num", block_num));
+         pending.reset();
+         protocol_features.popped_blocks_to( head->block_num );
+         emit( self.block_abort, block_num );
+      }
+      auto db_head = fork_db.head();
+      if (db_head->block && db_head->block->producer_signature == signature_type()) {
```
heifner: Seems like we should be able to always have this as an OR check with the `abort_block` case and then not need `abort_unsigned_block`. Either way I would like to see `abort_unsigned_block` removed.

libraries/chain/controller.cpp
```diff
@@ -2588,28 +2632,44 @@ void controller::start_block( block_timestamp_type when,
...
+controller::finalize_block(block_state_ptr& bsp, signer_callback_type&& sign) {
    validate_db_available_size();
 
    my->finalize_block();
 
    auto& ab = std::get<assembled_block>(my->pending->_block_stage);
 
-   auto bsp = std::make_shared<block_state>(
+   auto pfa = ab._pending_block_header_state.prev_activated_protocol_features;
+   auto pfs = my->protocol_features.get_protocol_feature_set();
```
heifner: This is an expensive copy that is not needed as `get_protocol_feature_set` returns `const protocol_feature_set&` and it is only used as a reference in this method.

libraries/chain/controller.cpp
```diff
@@ -2588,28 +2632,44 @@ void controller::start_block( block_timestamp_type when,
...
-   my->pending->_block_stage = completed_block{ bsp };
-
-   return bsp;
+   my->pending->_block_stage = completed_block{bsp};
+
+   return async_thread_pool(my->block_sign_pool.get_executor(),
+                            [bsp, my = my.get(), block_num = bsp->block_num, digest = bsp->sig_digest(),
+                             wtmsig_enabled = eosio::chain::detail::is_builtin_activated(
+                                 pfa, pfs, builtin_protocol_feature_t::wtmsig_block_signatures),
+                             sign = std::move(sign)]() -> std::function<void()> {
```
heifner: This lambda needs to be `mutable` for the `std::move` to work.

heifner (changes requested):
Partial review.

libraries/chain/controller.cpp
heifner: Decided safest to leave this out of the other thread so that the processing in the thread is completely stateless.

libraries/chain/controller.cpp
huangminghuang: This ambda only moves `sigs` which is a local variable inside lambda; therefore, it doesn't require `mutable` to work.

libraries/chain/fork_database.cpp
```diff
@@ -464,6 +464,23 @@ namespace eosio { namespace chain {
       }
    }
 
+   bool fork_database::is_signing_failed_block_head(uint32_t signing_failed_blocknum)  {
```
heifner: Seems like this should just be called `is_head_block(uint23_t blocknum)`

libraries/chain/fork_database.cpp
```diff
@@ -464,6 +464,23 @@ namespace eosio { namespace chain {
...
+      return my->head->block && my->head->block->block_num() == signing_failed_blocknum;
+   }
+
+   void fork_database::remove_head() {
```
heifner: I think this should take the `id` or `block_num` and verify head is as expected.
Also I think this should assert if prev does not exist as it doesn't remove the head in that case.

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -216,12 +216,14 @@ class state_history_traces_log : public state_history_log {
...
    void add_transaction(const chain::transaction_trace_ptr& trace, const chain::packed_transaction_ptr& transaction) {
-      cache.add_transaction(trace, transaction);
+      trace_caches[trace->block_num].add_transaction(trace, transaction);
    }
 
    std::shared_ptr<std::vector<char>> get_log_entry(block_num_type block_num);
 
-   void block_start(uint32_t block_num) { cache.clear(); }
+   void block_start(uint32_t block_num) {
+      trace_caches.erase(block_num-2);
```
heifner: Why `-2` instead of `-1`?

heifner: Is this even needed now that it is cleared in `store`?

plugins/amqp_trx_plugin/include/eosio/amqp_trx_plugin/fifo_trx_processing_queue.hpp
```diff
@@ -96,7 +93,7 @@ class blocking_queue {
    /// also checks paused flag because a paused queue indicates processing is on-going or explicitly paused
    bool empty() const {
       std::scoped_lock<std::mutex> lk(mtx_);
-      return queue_.empty() && !paused_;
+      return queue_.empty() && paused_ <= 0;
```
heifner, line 100: I would like to see these `paused_` changes reverted. Seems sloppy to not fix the calls so that the invariant holds instead of relaxing the invariant that unpause should only be called when paused.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -219,6 +219,17 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       std::optional<scoped_connection>                          _accepted_block_header_connection;
       std::optional<scoped_connection>                          _irreversible_block_connection;
 
+      enum signatures_status_type {
+         signatures_none,
+         signatures_pending,
+         signatures_ready
+      };
```
heifner, line 203: Can we make this an enum class

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -290,7 +301,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       void abort_block() {
          auto& chain = chain_plug->chain();
 
-         _unapplied_transactions.add_aborted( chain.abort_block() );
+         _unapplied_transactions.add_aborted(chain.abort_block());
+
+         // TODO: need work on the signing failure case
```
heifner: Remove TODO. Looks to me like subjective billing does not require any changes.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1706,19 +1719,34 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       }
 
       try {
-         if( !remove_expired_trxs( preprocess_deadline ) )
+         if (!remove_expired_trxs(preprocess_deadline))
```
heifner: Please try to avoid space changes in non-modified code, it makes the diff harder to read.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2411,38 +2462,41 @@ void producer_plugin_impl::produce_block() {
...
-   } );
+   eosio::chain::block_state_ptr block_state;
+   signatures_status     = signatures_pending;
+   assign_signatures_fut = chain.finalize_block(block_state, [relevant_providers = std::move(relevant_providers),
+                                                              self = shared_from_this()](const digest_type& d) {
+      auto                        debug_logger = maybe_make_debug_time_logger();
+      auto                        on_exit      = fc::make_scoped_exit([self] {
+         self->signatures_status = signatures_ready;
+         app().post(priority::high, [self]() { self->assign_block_signatures_if_ready(); });
+      });
```
heifner, line 2335: The `on_exit` doesn't add anything here. The transform can't throw except for out-of-memory. Just add these two lines of code before the `return`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2411,38 +2462,41 @@ void producer_plugin_impl::produce_block() {
...
 
    chain.commit_block();
    block_state_ptr new_bs = chain.head_block_state();
    ilog("Produced block ${id}... #${n} @ ${t} signed by ${p} [trxs: ${count}, lib: ${lib}, confirmed: ${confs}]",
-        ("p",new_bs->header.producer)("id",new_bs->id.str().substr(8,16))
-        ("n",new_bs->block_num)("t",new_bs->header.timestamp)
-        ("count",new_bs->block->transactions.size())("lib",chain.last_irreversible_block_num())("confs", new_bs->header.confirmed));
+      ("p",new_bs->header.producer)("id",new_bs->id.str().substr(8,16))
+      ("n",new_bs->block_num)("t",new_bs->header.timestamp)
+      ("count",new_bs->block->transactions.size())("lib",chain.last_irreversible_block_num())("confs", new_bs->header.confirmed));
```
heifner: I don't see any change here. Why was indent changed?

plugins/signature_provider_plugin/signature_provider_plugin.cpp
```diff
@@ -90,16 +98,17 @@ signature_provider_plugin::signature_provider_plugin():my(new signature_provider
...
 void signature_provider_plugin::set_program_options(options_description&, options_description& cfg) {
-   cfg.add_options()
-         ("keosd-provider-timeout", boost::program_options::value<int32_t>()->default_value(5),
-          "Limits the maximum time (in milliseconds) that is allowed for sending requests to a keosd provider for signing")
-         ;
+   cfg.add_options()("keosd-provider-timeout", boost::program_options::value<int32_t>()->default_value(5),
+                     "Limits the maximum time (in milliseconds) that is allowed for sending requests to a keosd "
+                     "provider for signing")("signing-delay",
+                                             boost::program_options::value<uint32_t>()->default_value(0),
+                                             "milliseconds to delay the signature signing when using default signature provider");
```
heifner: Fix formating.

heifner (changes requested):

libraries/state_history/include/eosio/state_history/log.hpp
huangminghuang: `store` only clear the successful block,  not the aborted block. The trace_caches should caches at most 2 blocks, i.e. `block_num`, and `block_num-1`. Any block older than `block_num-1` can be considered obsolete.

plugins/producer_plugin/producer_plugin.cpp
huangminghuang: The `sign` can actually throw exception at least for TPM. See https://github.com/EOSIO/eos/blob/44624098a436b3694861029e71fa5b4f9f3f0742/libraries/tpm-helpers/tpm-helpers.cpp#L315

libraries/chain/controller.cpp
```diff
@@ -1602,6 +1602,17 @@ struct controller_impl {
       pending->push();
    }
 
+   void assign_signatures(block_state_ptr bsp, std::vector<signature_type>&& sigs, bool wtmsig_enabled) {
```
heifner: I don't like this does more than assign signatures. I think it needs to be renamed or the `log_irreversible()` and `emit` should be pulled out of it. 
I also don't like that the `emit(self.accepted_block, bsp);` is called from two locations now. Seems we need a concept like finalize block (although we already have that). Maybe complete_block.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1608,17 +1619,31 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       try {
          if( !remove_expired_trxs( preprocess_deadline ) )
             return start_block_result::exhausted;
+         
+         if (!assign_block_signatures_if_ready())
```
heifner: I think `complete_block` would be a better name.

libraries/chain/controller.cpp
```diff
@@ -1565,47 +1565,40 @@ struct controller_impl {
...
 
-   /**
-    * @post regardless of the success of commit block there is no active pending block
-    */
-   void commit_block(block_state_ptr&& bsp) {
-      auto reset_pending_on_exit = fc::make_scoped_exit([this]{
-         pending.reset();
-      });
+   void create_completed_block(block_state_ptr bsp) {
+      auto reset_pending_on_exit = fc::make_scoped_exit([this] { pending.reset(); });
```
heifner, line 1569: You can just add `pending.reset()` as the last line of this method. shared_ptr copy/move does not throw. `pending->push()` also does not throw.
`bsp` can be moved into `completed_block`

libraries/chain/controller.cpp
```diff
@@ -2580,10 +2570,12 @@ controller::finalize_block(block_state_ptr& bsp, signer_callback_type&& sign) {
                             }
                             FC_LOG_AND_DROP();
                             return [bsp, my, signatures = std::move(signatures), wtmsig_enabled]() mutable {
+                               // the labmda is to be executed in main thread 
                                my->complete_produced_block(bsp, std::move(signatures), wtmsig_enabled);
                             };
                          });
-   my->commit_block(std::move(bsp));
+   my->add_to_fork_db(bsp);
+   my->create_completed_block(bsp);
```
heifner, line 2578: `bsp` can be moved here.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1608,17 +1619,31 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
          if( !_subjective_billing.remove_expired( _log, chain.pending_block_time(), fc::time_point::now(), preprocess_deadline ) )
             return start_block_result::exhausted;
 
+         if (!complete_produced_block_if_ready())
+            return start_block_result::failed;
+
          // limit execution of pending incoming to once per block
          size_t pending_incoming_process_limit = _unapplied_transactions.incoming_size();
 
          if( !process_unapplied_trxs( preprocess_deadline ) )
```
heifner: Need to check `complete_produced_block_if_ready` within the loop of `process_unapplied_trxs`.

plugins/trace_api_plugin/include/eosio/trace_api/chain_extraction.hpp
```diff
@@ -71,45 +75,47 @@ class chain_extraction_impl_type {
...
    }
 
    void store_block_trace( const chain::block_state_ptr& block_state ) {
       try {
          using transaction_trace_t = transaction_trace_v2;
 
          auto bt = create_block_trace( block_state );
+         const auto& itr = tracked_blocks.find( block_state->block_num );
 
+         if( itr == tracked_blocks.end() )
```
heifner: This should never happen. Minor thing, but an error log might be appropriate.

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -218,12 +218,14 @@ class state_history_traces_log : public state_history_log {
...
    void add_transaction(const chain::transaction_trace_ptr& trace, const chain::packed_transaction_ptr& transaction) {
-      cache.add_transaction(trace, transaction);
+      trace_caches[trace->block_num].add_transaction(trace, transaction);
    }
 
    std::shared_ptr<std::vector<char>> get_log_entry(block_num_type block_num);
 
-   void block_start(uint32_t block_num) { cache.clear(); }
+   void block_start(uint32_t block_num) {
+      trace_caches[block_num].clear();
```
heifner, line 227: This leaves the entry in the map which is a slow leak.

plugins/trace_api_plugin/include/eosio/trace_api/chain_extraction.hpp
```diff
@@ -71,45 +75,47 @@ class chain_extraction_impl_type {
    }
 
    void on_block_start( uint32_t block_num ) {
-      clear_caches();
+      tracked_blocks[block_num] = block_tracking{};
```
heifner, line 80: Need to listen for abort block so we can erase this otherwise this is a slow leak on every aborted block.

heifner: Actually, nevermind. You can't skip a block number, obviously, so this will just wipe out the previous setting on starting the block after an abort block.

libraries/state_history/include/eosio/state_history/log.hpp
heifner: It is cleared in store_entry

heifner (approved):

PR #10883, opened 2021-11-10
Partly support std::array-2.2.x

softprofe:
Partly support using std::array in smart contract, it can be used in function, in struct, and also OK to nested using.

However, std::array can't be as parameter or return of an ACTION or field of kv table so far.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [x] Documentation Additions

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -207,6 +207,19 @@ namespace eosio { namespace chain {
       return ends_with(type, "[]");
    }
 
+   bool abi_serializer::is_szarray(const string_view& type)const {
+      auto pos1 = type.find_last_of('[');
+      auto pos2 = type.find_last_of(']');
+      if(pos1 == string_view::npos || pos2 == string_view::npos) return false;
+      auto pos = pos1 + 1;
+      if(pos == pos2) return false;
+      while(pos < pos2) {
+         if( ! (type[pos] >= '0' && type[pos] <= '9') ) return false;
+         ++pos;
+      }
+      return true;
+   }
+
    bool abi_serializer::is_optional(const string_view& type)const {
       return ends_with(type, "?");
    }
@@ -223,6 +236,8 @@ namespace eosio { namespace chain {
    std::string_view abi_serializer::fundamental_type(const std::string_view& type)const {
       if( is_array(type) ) {
          return type.substr(0, type.size()-2);
+      } else if (is_szarray (type) ){
+         return type.substr(0, type.find_last_of('['));
       } else if ( is_optional(type) ) {
          return type.substr(0, type.size()-1);
       } else {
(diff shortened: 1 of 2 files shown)


libraries/chain/abi_serializer.cpp
```diff
@@ -207,6 +207,19 @@ namespace eosio { namespace chain {
...
+      auto pos2 = type.find_last_of(']');
+      if(pos1 == string_view::npos || pos2 == string_view::npos) return false;
+      auto pos = pos1 + 1;
+      if(pos == pos2) return false;
+      while(pos < pos2) {
+         if( ! (type[pos] >= '0' && type[pos] <= '9') ) return false;
+         ++pos;
+      }
+      return true;
+   }
```
heifner, line 221: `t][`
pos1=2
pos2=1
pos=3
returns true

libraries/chain/abi_serializer.cpp
```diff
@@ -223,6 +236,8 @@ namespace eosio { namespace chain {
    std::string_view abi_serializer::fundamental_type(const std::string_view& type)const {
       if( is_array(type) ) {
          return type.substr(0, type.size()-2);
+      } else if (is_szarray (type) ){
```
heifner, line 239: Remove space `is_szarray (type)` => `is_szarry(type)`

libraries/chain/abi_serializer.cpp
```diff
@@ -207,6 +207,19 @@ namespace eosio { namespace chain {
       return ends_with(type, "[]");
    }
```
heifner, line 208: This needs tests.

heifner (changes requested):

heifner:
Add some examples to the PR description.

libraries/chain/abi_serializer.cpp
huangminghuang: Can be implemented as 
```
if (auto r = get_array_type_name(type) ; !r.empty()) 
  return r;
```

libraries/chain/abi_serializer.cpp
huangminghuang: - `is_szarray()` is not using any other members in the `abi_serializer` class, it should be a free function. 
- This can be combined with `is_array()` to be  implemented more efficiently and correctly as
```
std::string_view get_array_type_name(std::string_view type) {
    if (type.back() != ']') return {};
    auto it = std::find_if_not(type.rbegin() + 1, type.rend(), [](char x) { return x >= '0' &&  x <='9'; });
    if (it == type.rend() || *it != '[') return {};
    return {type.begin(),  static_cast<std::size_t>(type.rend() - it - 1)};
}
```

huangminghuang (changes requested):

PR #10804, opened 2021-10-09
Port Integration Tests from Legacy History Plugin to Trace API Plugin

nickjjzhao:
## Change Description

[EPE-42](https://blockone.atlassian.net/browse/EPE-42) 

Reverted original [PR#10624](https://github.com/EOSIO/eos/pull/10624)

There are some integration tests that utilize the RPC API `get_transaction` defined in legacy `history plugin`. This PR is making these tests utilize a new defined RPC API `/v1/trace_api/get_transaction_trace` in trace_api_plugin. So dependency on `history plugin` is removed from the tests. 

**Note**:  to make current `multiversion test `work, only `--plugin eosio::history_api_plugin` is kept in a nodeos command and `--plugin eosio::trace_api_plugin` is removed by `eosio-launcher` if both are added to the nodeos.

To search for a transaction trace in the trace logs, new `sliced transaction id log files` are created. These files are used to retrieve the block in which the requested transaction is saved. New slices have the same slice width as the trace log slices, and default slice width is `10000`. New slices can be deleted along with trace slices during trace maintenance. But new slices can't be compressed like trace slices. The structure of data entries saved in the new slices is shown below. Each entry contains a block number and all the transaction ids if any in the block. 

```
  struct block_trxs_entry {
      std::vector<chain::transaction_id_type> ids;
      uint32_t block_num = 0;
  };
```
Some examples of new files names with slice width `10000` look like:
```
trace_trx_id_0000000000-0000010000.log
trace_trx_id_0000010000-0000020000.log
trace_trx_id_0000020000-0000030000.log
```
An example of` /v1/trace_api/get_transaction_trace` looks like:

```
❯ curl -X POST http://127.0.0.1:8888/v1/trace_api/get_transaction_trace  -d  '{"id": "7c9fb855a4516c27cc5dc96e81f0cb3555c9ccf7c09fcf6fd42975d58b04f721"}' | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   986  100   912  100    74   890k  74000 --:--:-- --:--:-- --:--:--  962k
{
  "id": "7c9fb855a4516c27cc5dc96e81f0cb3555c9ccf7c09fcf6fd42975d58b04f721",
  "block_num": 13,
  "block_time": "2021-10-13T01:28:18.000",
  "producer_block_id": null,
  "actions": [
    {
      "global_sequence": 13,
      "receiver": "eosio",
[...]

PR #10884, opened 2021-11-15
Fix trace_api_plugin trx trace storage for forks

nickjjzhao:
## Change Description
[EPE-1631](https://blockone.atlassian.net/browse/EPE-1631)

Fix the `trace_api_plugin` storage of transaction ids introduced in [PR#10804](https://github.com/EOSIO/eos/pull/10804) to correctly handle forks.

Integration test `nodeos_high_transaction_test` occasionally  fails due to error "`Trace API: transaction trace missing`." The error indicates that a transaction ID was found in the transaction ID log but the transaction trace couldn't be found in the trace log using the block number returned from the ID log. The root cause of the issue is that the ID log could  have duplicate transaction IDs because of forks and it doesn't handle that situation.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [X] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/trace_api_plugin/store_provider.cpp
+++ b/plugins/trace_api_plugin/store_provider.cpp
@@ -53,11 +53,13 @@ namespace eosio::trace_api {
    template void store_provider::append<block_trace_v2>(const block_trace_v2& bt);
 
    void store_provider::append_lib(uint32_t lib) {
-      fc::cfile index;
+      fc::cfile index, trx_id;
       const uint32_t slice_number = _slice_directory.slice_number(lib);
       _slice_directory.find_or_create_index_slice(slice_number, open_state::write, index);
       auto le = metadata_log_entry { lib_entry_v0 { .lib = lib }};
       append_store(le, index);
+      _slice_directory.find_or_create_trx_id_slice(slice_number, open_state::write, trx_id);
+      append_store(le, trx_id);
       _slice_directory.set_lib(lib);
    }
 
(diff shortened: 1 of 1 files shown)


plugins/trace_api_plugin/store_provider.cpp
```diff
@@ -118,16 +122,32 @@ namespace eosio::trace_api {
...
+            } else {
+               FC_ASSERT( false, "unpacked data should be a block trxs entry or a lib entry" );;
             }
             offset = trx_id_file.tellp();
          }
          slice_number++;
       }
+
+      // no irreversible blocks after the block containing the target trx, e.g., it happens
+      // in some integration tests that call waitForTransInBlock during bios bootstrap.
```
heifner: Change this comment (or remove it) to something more like:

`// transaction's block is not irreversible`

heifner (changes requested):

plugins/trace_api_plugin/store_provider.cpp
nickjjzhao: Fixed.

heifner (approved):

PR #10888, opened 2021-11-15
Fix misleading error message when loading large WASM file into nodeos

plroblox:
## Change Description

Fix misleading error message when loading large WASM file into nodeos

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -324,8 +324,8 @@ fc::variant call( const std::string& url,
    }
    catch(boost::system::system_error& e) {
       if(url == ::default_url) {
-          std::string error_wasm_toolarge = "is your WASM file too large (default maximum size is 512KB)?";
-          std::cerr << localized("Failed to connect to ${n} at ${u}; is ${n} running or ${wasm_size_error}",
+          std::string error_wasm_toolarge = "if trying to set contract code, is your WASM file too large (default maximum size is 512KB)?";
+          std::cerr << localized("Failed to connect to ${n} at ${u}!\nPlease check the log of ${n},is ${n} running? or ${wasm_size_error}",
                              ("n", node_executable_name)("u", url)("wasm_size_error", error_wasm_toolarge))<< std::endl;
       }
       else if(url == ::wallet_url)
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -323,8 +323,11 @@ fc::variant call( const std::string& url,
       return eosio::client::http::do_http_call(*sp, fc::variant(v), print_request, print_response );
    }
    catch(boost::system::system_error& e) {
-      if(url == ::default_url)
-         std::cerr << localized("Failed to connect to ${n} at ${u}; is ${n} running?", ("n", node_executable_name)("u", url)) << std::endl;
+      if(url == ::default_url) {
+          std::string error_wasm_toolarge = "is your WASM file too large (default maximum size is 512KB)?";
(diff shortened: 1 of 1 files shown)


programs/cleos/main.cpp
```diff
@@ -323,8 +323,11 @@ fc::variant call( const std::string& url,
       return eosio::client::http::do_http_call(*sp, fc::variant(v), print_request, print_response );
    }
    catch(boost::system::system_error& e) {
-      if(url == ::default_url)
-         std::cerr << localized("Failed to connect to ${n} at ${u}; is ${n} running?", ("n", node_executable_name)("u", url)) << std::endl;
+      if(url == ::default_url) {
+          std::string error_wasm_toolarge = "is your WASM file too large (default maximum size is 512KB)?";
+          std::cerr << localized("Failed to connect to ${n} at ${u}; is ${n} running or ${wasm_size_error}",
+                             ("n", node_executable_name)("u", url)("wasm_size_error", error_wasm_toolarge))<< std::endl;
+      }
```
praphael, line 330: You should be able to detect if the user had invoked the `set code` or `set contract` action and only print out that specific error message in that case.  Otherwise this may be confusing if they had not invoked either of those actions.

Ideally this should be fixed server side with an appropriate error message, but would also have to involve client side changes if connection is closed (client would probably have to be async).

heifner: I think we should say something more like. Verify connection ${n} is running and check its logs.
I agree we should only say suggest that a common issue is WASM size for a set code/contract.

praphael: Agree, this might be simpler.   Maybe something like "If trying to set contract code, check WASM size does not exceed max limit in nodeos."

plroblox: Actually I thought about adding some code onto nodeos beast_http_session to catch the error when a large size WASM is loaded into nodeos, however publishing a large size WASM contract is related to both keosd and nodoes as said in https://github.com/EOSIO/eosio.cdt/blob/develop/docs/09_tutorials/10_nested_containers/index.md#how-to-publishing-a-wasm-file-bigger-than-512kb,  not just nodeos.

So when publishing a large size WASM(>512KB),  and the max-body-size of keosd is not set, the same error will occur.
Trying to code both nodeos and keosd to watch for a large size WASM is unnecessarily complicated, so I choose to only change cleos main.cpp.

I agree that checking log of nodoes is important, e.g. the log would tell you not to use history-plugin any more, also it is necessary to remind WASAM size issue occurs for subcommands set contract/code, so I would update the error message as below: 

Please check the log of nodeos, is nodeos running?  If trying to set contract code, is your WASM file too large (default maximum size is 512KB)?

heifner: This is less than ideal. Please increase the default size limit of http or create a JIRA issue to do it.

heifner (approved):

programs/cleos/main.cpp
plroblox: EPE-1817 Increase the maximum default WASM file size which is currently 512KB is created.

PR #10892, opened 2021-11-17
Add tests for --replay-blockchain  to nodeos_read_terminate_at_block_lr_test

hmmblockone:
## Change Description

The nodeos_read_terminate_at_block_lr_test now also verifies that the --replay-blockchain option works correctly with --terminate at block.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [x] New Tests

- [X] Existing Tests

The nodeos_read_terminate_at_block_lr_test tests --terminate-at-block with each read mode (irreversible, speculative, head), and will now additionally restart and replay each node with --replay-blockchain option verifying that it works correctly with --terminate at block.

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/tests/nodeos_read_terminate_at_block_test.py
+++ b/tests/nodeos_read_terminate_at_block_test.py
@@ -141,18 +141,16 @@ def checkReplay(testNode, testNodeArgs):
     time.sleep(10)
 
     # Check for the terminate at block message.
-    match = re.search("--terminate-at-block=(\d+)", testNodeArgs)
+    match = re.search(r"--terminate-at-block=(\d+)", testNodeArgs)
     termAtBlock = int(match.group(1))
 
     termMsg = "Reached configured maximum block {}; terminating".format(
         termAtBlock
     )
 
-    with open(testNode.popenProc.errfile.name, 'r') as f:
-        assert termMsg in f.read()
+    assert checkLog(testNode.popenProc.errfile.name, termMsg)
 
 
-# Wrapper function to execute test
 def getBlockNumInfo(testNode):
     head = None
     lib = None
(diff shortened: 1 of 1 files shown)


tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -102,12 +97,62 @@ def executeTest(cluster, producerNode, testNodeId, testNodeArgs, resultMsgs):
...
+
+    # Check for the terminate at block message.
+    match = re.search("--terminate-at-block=(\d+)", testNodeArgs)
+    termAtBlock = int(match.group(1))
+
+    termMsg = "Reached configured maximum block {}; terminating".format(
+        termAtBlock
+    )
+
+    with open(testNode.popenProc.errfile.name, 'r') as f:
```
dimas1185: can you add some timer here or at least sleep before? I want to make sure logs will be always there upon reaching that line.

tests/nodeos_read_terminate_at_block_test.py
```diff
@@ -102,12 +97,62 @@ def executeTest(cluster, producerNode, testNodeId, testNodeArgs, resultMsgs):
...
+    Print(" ".join([
+        "Relaunch the node in replay mode. The replay should stop",
+        "at or little bigger than the block number specified here: ",
+        testNodeArgs
+    ]))
+
+    testNode.relaunch(chainArg="--replay-blockchain", cachePopen=True)
+
+    # Check node exited without error.
+    assert testNode.popenProc.returncode == 0
```
dimas1185: I think you can rely only on `testNode.relaunch` return value. No need additional check

tests/nodeos_read_terminate_at_block_test.py
hmmblockone: Added a sleep(10) after the relaunch.

tests/nodeos_read_terminate_at_block_test.py
hmmblockone: Removed the return value check.

tests/nodeos_read_terminate_at_block_test.py
dimas1185: it may be not required. I propose to do a loop and add sleep(1) inside.
something like this:
```python
maxAttempt=10
attemptCnt=0
found=False
while not found and attemptCnt < maxAttempt:
   with open() as f:
       found = termMsg in f.read()
   if not found:
      sleep(1)
      attemptCnt=attemptCnt+1
assert found, "err msg"
```

hmmblockone: I can certainly add that, at least then it might not have to wait the whole time.

dimas1185 (approved):

PR #10936, opened 2021-12-08
Eixt nodeos gracefully if db size is smaller than snapshot size

softprofe:
## Change Description

For EPE 894, if db size if set to smaller than snapshot size, it will print bad_alloc and throw. After this PR, give a meaningful print and gracefully shutdown.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -506,6 +506,7 @@ struct controller_impl {
                   "chainbase is at revision ${a}, but chain-kv is at revision ${b}", ("a", db.revision())("b", kv_db.revision()) );
       this->shutdown = shutdown;
       ilog( "Starting initialization from snapshot, this may take a significant amount of time" );
+      bool exception = false;
       try {
          snapshot->validate();
          if( blog.head() ) {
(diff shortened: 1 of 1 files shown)


plroblox (approved):
ok

heifner (changes requested):
This fix is not better it indicates success on exit.

softprofe:
> This fix is not better it indicates success on exit.

After this PR, It print two red marked lines to let user know error happened. as below:
 error 2021-12-08T02:34:32.713 thread-0  controller.cpp:531            startup              ] db storage not configured to have enough storage for the provided snapshot, please increase and retry snapshot
error 2021-12-08T02:34:32.713 thread-0  controller.cpp:539            startup              ] Failed initialization from snapshot

The final line says nodeos successfully exiting just let user know it is not crashed.
 info  2021-12-08T02:34:32.715 thread-0  main.cpp:179                  main                 ] nodeos successfully exiting

I think for most people and most company,  a crashed-like end is not recommend.  @nksanthosh please make decision we will have this PR or not.

nksanthosh:
> This fix is not better it indicates success on exit.

@heifner  - can you elaborate what you are suggesting here? (exit with a non zero error code? or something else?)

matthewdarwin:
the node failed to start, so non-zero exit code seems correct here.

heifner:
> @heifner - can you elaborate what you are suggesting here? (exit with a non zero error code? or something else?)

The reported issue was that all that was reported was `bad alloc`. The actual error was indicated before the `bad alloc` error. All I think we need to do is change nodeos main.cpp catch of boost::interprocess::bad_alloc to indicate that the likely issue is chainbase exhaustion and to look for previously reported errors.

We still want nodeos to return `BAD_ALLOC` return code 1 in this case.

PR #10950, opened 2021-12-13
Boost update, remove Centos 8 and Mac OS 10.14 📦

praphael:
## Change Description
Fixes issue in Boost caused by CVE-2016-9840.  Removes Centos 8 and MacOS 10.14

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [x] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.cicd/platforms/pinned/ubuntu-16.04-pinned.dockerfile
```diff
@@ -37,10 +37,12 @@ RUN git clone --depth 1 --single-branch --branch llvmorg-10.0.0 https://github.c
     make install && \
     cd / && \
     rm -rf /llvm
-# build boost
+# download Boost, apply fix for CVE-2016-9840 and build
 RUN curl -LO https://boostorg.jfrog.io/artifactory/main/release/1.72.0/source/boost_1_72_0.tar.bz2 && \
     tar -xjf boost_1_72_0.tar.bz2 && \
     cd boost_1_72_0 && \
+    curl -Lo boost/beast/zlib/detail/inflate_stream.ipp \
+    https://raw.githubusercontent.com/boostorg/beast/develop/include/boost/beast/zlib/detail/inflate_stream.ipp && \
```
spoonincode: It's possible the version on master will change to be incompatible in the future. It's probably prudent to lock this download to a specific git commit

kj4ezj (changes requested):
Let's add the `-fsS` flags to the `curl` commands that are being added or updated.

Screen Shot 2021-09-28 at 4 59 21 PM

These flags make the dockerfiles and scripts easier to debug. With the current flags, if `curl` receives a 400 or 500 error, it will download the HTML error page and save it with the same name as the intended artifact and the script will continue running. Only later, when you try to invoke the HTML error page as a binary, will the script fail with a very nebulous error. These flags will cause `curl` to fail immediately with a clear error message.

So the final command in-use would be:
```bash
curl -fsSLo $URL
```
It is also a good habit to add quotes to the variables, to protect against code injection in BASH scripts more generally.
```bash
curl -fsSLo "$URL"
```
This is especially important when using `eval`, but a good idea all the time.

kj4ezj:
Should we also be removing Ubuntu 16.04?

praphael:
> Should we also be removing Ubuntu 16.04?
 
 I would think so.  Might as well also remove Centos 7.7 while we're at it.

kj4ezj (approved):
🍕

.cicd/platforms/pinned/ubuntu-16.04-pinned.dockerfile
praphael: Fixed in next commit

spoonincode:
I guess it's too late now, but I really think platform removal should have been its own separate PR

praphael:
> I guess it's too late now, but I really think platform removal should have been its own separate PR

Not too late, I can revert the commits.  But it was breaking with Centos 8 for some reason, which is part of the reason I removed it.  Then decided to remove MacOS 10.14 as well.

PR #10952, opened 2021-12-14
net_plugin: shutdown segfault fix 📦

heifner:
## Change Description

- Avoid accessing connection data during operation_aborted since it can be called during shutdown when connection is being destroyed
- https://github.com/EOSIO/eos/pull/10958

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1547,8 +1547,7 @@ namespace eosio {
    void connection::sync_timeout( boost::system::error_code ec ) {
       if( !ec ) {
          my_impl->sync_master->sync_reassign_fetch( shared_from_this(), benign_other );
-      } else if( ec == boost::asio::error::operation_aborted ) {
-      } else {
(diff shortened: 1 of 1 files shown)


dimas1185 (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1558,9 +1558,6 @@ namespace eosio {
       if( !ec ) {
          my_impl->dispatcher->retry_fetch( shared_from_this() );
       } else if( ec == boost::asio::error::operation_aborted ) {
```
dimas1185: actually now you don't need the `else if`

heifner: The else-if is to avoid the `peer_elog` of the `else`

PR #11006, opened 2022-01-04
Fix nested container tests

shahabi8:
## Change Description
This PR fixes nested container tests for pair and map containers. 
Map can be defined through `key` and `value` like this: `[{"key":10, "value":100],{"key":10, "value":100}]`
Pair can be defined through `first` and `second` like this: `["first":30, "second":{"first":301, "second":302}]`

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


plroblox (approved):
Excellent !  map should go with key-value whereas pair should go-with firt-second regarding the JSON input formats of typess involving map/pair

PR #11028, opened 2022-01-12
rocksdb integration for nodeos removed 📦

dimas1185:
## Change Description

Changes:
* Removal of `rocksdb` from `nodeos`. Note: we still have `rocksdb` used by `rodeos`. Also `nodeos` chain config still maintains `backing_store` and `rocksdb` related config for backward compatibility although those are not used anymore and `backing_store` is hardcoded to be always `chainbase`. [EPE-1841](https://blockone.atlassian.net/browse/EPE-1841)
* eosvmoc test fix. This is not related directly to `rocksdb` changes but started to reproduce with that fix. `there_and_back_again` test was comparing deleted memory to compare results and after this particular change it started to differ.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [x] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

removal of those options from `chain_plugin` command line reference:
`backing-store`
`persistent-storage-num-threads`
`persistent-storage-max-num-files`
`persistent-storage-write-buffer-size-mb`
`persistent-storage-bytes-per-sync`
`persistent-storage-mbytes-snapshot-batch`

## Documentation Additions
- [ ] Documentation Additions


libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -83,11 +83,13 @@ namespace eosio { namespace chain {
             uint16_t                 max_retained_block_files   = chain::config::default_max_retained_block_files;
             uint64_t                 blocks_log_stride          = chain::config::default_blocks_log_stride;
             backing_store_type       backing_store              = backing_store_type::CHAINBASE;
+            // rocksdb related. currently unused
```
heifner: Remove all these and all the `default_` ones in `chain/config.hpp`

libraries/chain/include/eosio/chain/combined_database.hpp
```diff
@@ -76,17 +72,13 @@ namespace eosio { namespace chain {
 
     private:
       std::unique_ptr<chainbase::database::session> cb_session    = {};
-      eosio::session::undo_stack<rocks_db_type>*     kv_undo_stack = nullptr;
    };
 
    class combined_database {
```
heifner: This should be removed along with the combined session. Basically the code should look like it does in `release/2.0.x`

heifner (changes requested):
Couple of initial comments you can start on while I continue my review.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -469,55 +469,29 @@ class read_only {
    bool get_primary_key_internal(name code, name scope, name table, uint64_t primary_key, row_requirements require_table,
                                  row_requirements require_primary, Function&& f) const {
       const auto db_backing_store = get_backing_store();
```
heifner: Go ahead and remove backing_store concept.

libraries/chain/include/eosio/chain/controller.hpp
dimas1185: Those things are saved in state database. So this means we will need to delete state and do replay on next bullish release. are we ok with that?

heifner: They are not stored in the state database.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
dimas1185: done

libraries/chain/include/eosio/chain/combined_database.hpp
dimas1185: as discussed in DM, we just renamed this to `kv_session` and `kv_database` as this was refactoring

libraries/chain/include/eosio/chain/controller.hpp
dimas1185: good point. only `backing_store`. so I removed everything related to rocksdb from config but `backing_store`

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -394,14 +388,12 @@ namespace eosio { namespace chain {
          void replace_producer_keys( const public_key_type& key );
          void replace_account_keys( name account, name permission, const public_key_type& key );
 
-         eosio::chain::combined_database& kv_db()const;
-
+      chainbase::database& mutable_db()const;
+
```
heifner: I would prefer for this to remain private. However, I don't like having to forward declare `chain_apis` namespace in order to add a friend. Best thing would be to create a const kv_context, however that also gets rather ugly because of the need for a const iterator etc. I think you can change `create_kv_context` to take a controller and then add it as a friend.

libraries/chain/include/eosio/chain/db_util.hpp
```diff
@@ -0,0 +1,97 @@
...
+                                                 index_double_index, index_long_double_index>;
+
+   using db_context = backing_store::db_context;
+
+   class maybe_session {
+    public:
+      maybe_session() = default;
+
+      maybe_session(chainbase::database& cb_database);
+      maybe_session(maybe_session&& src) noexcept;
```
heifner, line 39: Make these constructors `explicit`

libraries/chain/include/eosio/chain/db_util.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include <eosio/chain/contract_table_objects.hpp>
+#include <eosio/chain/database_header_object.hpp>
+#include <eosio/chain/generated_transaction_object.hpp>
+#include <eosio/chain/genesis_intrinsics.hpp>
+#include <eosio/chain/global_property_object.hpp>
+#include <eosio/chain/kv_chainbase_objects.hpp>
+#include <eosio/chain/protocol_state_object.hpp>
+#include <eosio/chain/transaction_object.hpp>
+#include <eosio/chain/whitelisted_intrinsics.hpp>
+#include <eosio/chain/controller.hpp>
```
heifner, line 20: Huge number of includes most of which look like they are not used.

dimas1185: removed few but most of them are used in `controller_index_set`

libraries/chain/include/eosio/chain/db_util.hpp
dimas1185: did it for `maybe_session(chainbase::database& cb_database)`
move constructor is not suggested to be declared explicit. any reason for that?
https://stackoverflow.com/questions/6758717/explicit-move-constructor

libraries/chain/include/eosio/chain/controller.hpp
dimas1185: did it. friending other namespace function appears to be pretty verbose.

libraries/chain/include/eosio/chain/db_util.hpp
heifner: Ah, yes, move constructor, right, only explicit for the non-move constructor.

libraries/chain/include/eosio/chain/combined_database.hpp
dimas1185: update: `combined_session` now has old name `maybe_session` and combined_database transformed into namespace `db_utils`

heifner (approved):

PR #11049, opened 2022-01-27
Support --return-failure-trace in send_transaction.

dimas1185:
## Change Description
The `--return-failure-trace` option was introduced along with the `--read-only` option on `cleos push transaction` and was exclusive to read-only transaction processing.  This extends support to normal transactions.  It also renames the `push_ro_transaction` RPC API in nodeos to `send_ro_transaction` to better reflect the processing it actually does, which is consistent with `send_transaction` rather than `push_transaction`.

The macro-based URL dispatcher in nodeos required extension to support differing parameter versions.  To minimize the disruption to the macro, the `_v1` suffix was propagated to all previously unversioned asynchronous RPC API parameter names.

backport of #10533

In addition to original PR there are 2 changes:
`--return-failure-trace` is enabled by default
`--return-failure-trace` is added for any transaction. Not just `cleos push transaction` but things like `cleos add contract`, etc. Essentially everything that results in pushing transaction.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [x] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [x] API Changes

Nodeos now responds to a new RPC API `/v2/chain/send_transaction`.  This API requires two arguments: a boolean indicating whether or not to return failure traces, defaulting to true, and a signed transaction.  Because this is a breaking change, `/v1/chain/send_transaction` has been preserved and still accepts solely a signed transaction.

## Documentation Additions
- [x] Documentation Additions

Cleos support for the `push transaction` command now defaults to using the new `/v2/chain/send_transaction` RPC API.  It previously accepted an option`--use-old-rpc` to switch to `/v1/chain/push_transaction`.  That option has now been supplemented with the option `--use-old-send-rpc` to switch to `/v1/chain/send_transaction`.  Those two options are mutually exclusive.  Neither `--read-only` nor `--return-failure-trace` can be used with either `--use-old-rpc` or `--use-old-send-rpc`.


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -3000,6 +3000,65 @@ void read_write::send_transaction(const read_write::send_transaction_params& par
...
+         input_trx = std::make_shared<packed_transaction>( std::move( input_trx_v0 ), true );
+      } EOS_RETHROW_EXCEPTIONS(chain::packed_transaction_type_exception, "Invalid packed transaction")
+
+      auto trx_trace = fc_create_trace_with_id("Transaction", input_trx->id());
+      auto trx_span = fc_create_span(trx_trace, "HTTP Received");
+      fc_add_tag(trx_span, "trx_id", input_trx->id());
+      fc_add_tag(trx_span, "method", "/v2/chain/send_transaction");
+
+      app().get_method<incoming::methods::transaction_async>()(input_trx, true, false, static_cast<const bool>(params.return_failure_traces),
+            [this, token=trx_trace.get_token(), input_trx, next]
```
heifner: `token=fc_get_token(trx_trace)`

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -575,13 +575,12 @@ class read_only {
       };
    }
 
-   struct push_ro_transaction_params {
-      name account_name;
-      bool return_failure_traces = true;
+   struct send_ro_transaction_params_v1 {
+      bool return_failure_traces = false;
```
heifner: We want the default to be `true` for develop-boxed.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -973,25 +972,31 @@ class read_write {
...
+   void push_transactions(const push_transactions_params_v1& params, chain::plugin_interface::next_function<push_transactions_results> next);
 
-   using send_transaction_params = push_transaction_params;
+   using send_transaction_params_v1 = push_transaction_params_v1;
    using send_transaction_results = push_transaction_results;
-   void send_transaction(const send_transaction_params& params, chain::plugin_interface::next_function<send_transaction_results> next);
+   void send_transaction(const send_transaction_params_v1& params, chain::plugin_interface::next_function<send_transaction_results> next);
+
+   struct send_transaction_params_v2 {
+      bool return_failure_traces = false;
```
heifner: We want `return_failure_traces = true` default for develop-boxed.

programs/cleos/main.cpp
```diff
@@ -448,8 +450,16 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+      EOSC_ASSERT( !(tx_use_old_rpc && tx_use_old_send_rpc), "ERROR: --use-old-rpc and --use-old-send-rpc are mutually exclusive" );
+      packed_transaction_v0 pt_v0(trx, compression);
       if (tx_use_old_rpc) {
-         return call(push_txn_func, packed_transaction_v0(trx, compression));
+         EOSC_ASSERT( !tx_read_only, "ERROR: --read-only can not be used with --use-old-rpc" );
+         EOSC_ASSERT( !tx_rtn_failure_trace, "ERROR: --return-failure-trace can not be used with --use-old-rpc" );
+         return call(push_txn_func, pt_v0);
+      } else if (tx_use_old_send_rpc) {
+         EOSC_ASSERT( !tx_read_only, "ERROR: --read-only can not be used with --use-old-send-rpc" );
+         EOSC_ASSERT( !tx_rtn_failure_trace, "ERROR: --return-failure-trace can not be used with --use-old-send-rpc" );
```
heifner, line 462: We want the default on develop-boxed to return failure traces. Need an option to disable returning a failure trace. That may require changing python tests to pass the option as tests may expect no trace on failure.

tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,200 @@
...
+import decimal
+import re
+import json
+import os
+import copy
+import math
+import time
+
+###############################################################
+# nodeos_run_test
```
heifner: Update to correct name.

tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,200 @@
...
+#
+# General test that tests a wide range of general use actions around nodeos and keosd
```
heifner: Update description.

tests/nodeos_extra_packed_data_test.py
```diff
@@ -0,0 +1,200 @@
...
+#
+###############################################################
```
heifner: Not sure why this test is included in this PR; seems unrelated.
This test should be cleaned up to remove all the cruft that is not used.

heifner (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
dimas1185: for both v1 and v2?

tests/nodeos_extra_packed_data_test.py
dimas1185: yes, was added due to automatic cherry-pick, will remove

programs/cleos/main.cpp
dimas1185: working on that

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
heifner: yes

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
dimas1185: made it true by default via cleos

heifner: Would also like it true by default here for EOSJS and curl, etc.

programs/cleos/main.cpp
```diff
@@ -3952,7 +3963,6 @@ int main( int argc, char** argv ) {
    trxSubcommand->add_option("--signature", extra_sig_opt_callback, localized("append a signature to the transaction; repeat this option to append multiple signatures"))->type_size(0, 1000);
    add_standard_transaction_options_plus_signing(trxSubcommand);
    trxSubcommand->add_flag("-o,--read-only", tx_read_only, localized("Specify a transaction is read-only"));
-   trxSubcommand->add_flag("-t,--return-failure-trace", tx_rtn_failure_trace, localized("Return partial traces on failed transactions, use it along with --read-only)"));
```
heifner, line 3955: This should still be an option to match `push action`

dimas1185: I see. so if we want it for every single transaction, maybe to add it elsewhere? basically every command that calls `send_actions`. this includes `action`, `transaction`, `msig propose_trx`, etc.

heifner: Or just remove it everywhere. Not sure we need a cleos option not to return a trace.

dimas1185: option not to return trace is handy when locally testing something. returned trace is verbose. So I think we should keep this option

heifner: Fair enough, just make it one of the "global" options then.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
dimas1185: done

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
dimas1185: done

programs/cleos/main.cpp
dimas1185: done

heifner (approved):

dimas1185:
withdrawn

PR #11051, opened 2022-01-27
Big ship deltas - 2.0

heifner:
## Change Description

https://github.com/eosnetworkfoundation/mandel/pull/24

Fix for deltas bigger than uint32_t size can indicate. This was hit on Wax.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [x] Stability bug fix

- [ ] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [ ] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -131,7 +131,12 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
       state_history_log_header header;
       auto&                    stream = log.get_entry(block_num, header);
       uint32_t                 s;
+      // Compressed deltas now exceeds 4GB on one of the public chains. This length prefix
+      // was intended to support adding additional fields in the future after the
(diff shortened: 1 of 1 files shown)


huangminghuang (commented):
This changes the disk format of the ship log and we should also change the `ship_current_version` value. However, `ship_current_version=1` is already taken, I am not exactly sure which version we should use here.

tbfleming:
On Mandel, I'm leaning against bumping the version. As long as the size isn't exceeded, it's backwards compatible. If it is exceeded, boost will try to decompress an empty stream. If boost throws, it will disconnect the client. If boost returns an empty stream, the client will receive an empty stream and fail to deserialize, indicating a problem.

huangminghuang (approved):

PR #11056, opened 2022-01-29
Fix MacOS 10.15 package installation error

nickjjzhao:
## Change Description

[EPE-2017](https://blockone.atlassian.net/browse/EPE-2017)

The `Test Package` step for macOS 10.15 Catalina fails continuously on all eos branches ([example1](https://buildkite.com/EOSIO/eosio/builds/32218#56741e87-e33e-49fd-a3b7-258145982edb) and [example2](https://buildkite.com/EOSIO/eosio/builds/32220#9aab03fd-588a-4c48-811c-9441c3af29f7)). It started at early Jan 2022. It looks like the Test Package fails whenever it has output "[Checking for dependents of upgraded formulae](https://buildkite.com/EOSIO/eosio/builds/32067#5ed7e93f-2181-487b-93db-ce99b8f0c890/171-637)". I didn't see this output in a previously passed test ([example1](https://buildkite.com/EOSIO/eosio/builds/31933#9953d3d7-691f-40a4-afc9-ae2397592199) and [example2](https://buildkite.com/EOSIO/eosio/builds/31913#065e5390-3f5f-4d0e-8f07-f7f0099407f9)). 

After changing environment variable `HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK` to false, the Test Package [passed](https://buildkite.com/EOSIO/eosio/builds/32222#c011b8c2-ebb7-4a54-9968-dd6960e223ed). [Here](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/env_config.rb#L240-L246) is the description about this variable:

```
HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK:    {
  description: "If set, do not check for broken linkage of dependents or outdated dependents after " \
               "installing, upgrading or reinstalling formulae. This will result in fewer dependents " \
               " (and their dependencies) being upgraded or reinstalled but may result in more breakage " \
               "from running `brew install <formula>` or `brew upgrade <formula>`.",
  boolean:     true,
},
```

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [X] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions

### DIFF
--- a/.cicd/test-package.run.sh
+++ b/.cicd/test-package.run.sh
@@ -13,13 +13,13 @@ elif [[ $(yum --version 2>/dev/null) ]]; then # RHEL family packaging
     perform 'yum install -y /eos/*.rpm'
 elif [[ $(brew --version 2>/dev/null) ]]; then # homebrew packaging
     perform 'brew update'
+    perform 'brew upgrade'
     perform 'brew --version'
     perform 'mkdir homebrew-eosio'
     perform 'git init homebrew-eosio'
     perform 'cp *.rb homebrew-eosio'
     perform "sed -i.bk -e 's/url \".*\"/url \"http:\/\/127.0.0.1:7800\"/' homebrew-eosio/*.rb"
     perform "pushd homebrew-eosio && git add *.rb && git commit -m 'test it!' && popd"
-    perform 'export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=false'
     perform "brew tap eosio/eosio homebrew-eosio"
     perform '{ python3 -m http.server 7800 & } && export HTTP_SERVER_PID=$!'
     perform 'sleep 20s'
--- a/.cicd/test-package.run.sh
+++ b/.cicd/test-package.run.sh
@@ -13,6 +13,7 @@ elif [[ $(yum --version 2>/dev/null) ]]; then # RHEL family packaging
     perform 'yum install -y /eos/*.rpm'
 elif [[ $(brew --version 2>/dev/null) ]]; then # homebrew packaging
     perform 'brew update'
+    perform 'brew --version'
     perform 'mkdir homebrew-eosio'
     perform 'git init homebrew-eosio'
     perform 'cp *.rb homebrew-eosio'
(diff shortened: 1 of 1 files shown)


kj4ezj (changes requested):
It seems like this covers up a real dependency problem with our package configuration rather than fixing it. During your research, did you find any other software projects on the web asking their customers to mess with the internal workings of their package manager just to do an install?

I have never had to do anything remotely like this to install a package with Homebrew. I think if we aren't going to support Homebrew correctly, we should just delete it.

.cicd/test-package.run.sh
```diff
@@ -13,11 +13,13 @@ elif [[ $(yum --version 2>/dev/null) ]]; then # RHEL family packaging
     perform 'yum install -y /eos/*.rpm'
 elif [[ $(brew --version 2>/dev/null) ]]; then # homebrew packaging
     perform 'brew update'
+    perform 'brew --version'
     perform 'mkdir homebrew-eosio'
     perform 'git init homebrew-eosio'
     perform 'cp *.rb homebrew-eosio'
     perform "sed -i.bk -e 's/url \".*\"/url \"http:\/\/127.0.0.1:7800\"/' homebrew-eosio/*.rb"
     perform "pushd homebrew-eosio && git add *.rb && git commit -m 'test it!' && popd"
+    perform 'export HOMEBREW_NO_INSTALLED_DEPENDENTS_CHECK=false'
```
williamblevins: This explains why installing it twice or running the fix links command works but it's hideous this is the default Homebrew behavior. I don't see a way to force this in the formulae itself.

williamblevins: Can we try (rather than setting this export) `brew list --formula | xargs -I % sh -c 'brew unlink %; brew link %'` after `brew update` to see if its a problem with the existing installed formula?

nickjjzhao: Test Package failed https://buildkite.com/EOSIO/eosio/builds/32239#3d89b1f3-3a2b-49f6-87f8-f07fc422c033 after trying `brew list --formula | xargs -I % sh -c 'brew unlink %; brew link %'`. Will keep setting this export as a fix since we can't prove it’s eosio, and the fix suggests its a problem with the existing installed packages.

williamblevins (approved):

.cicd/test-package.run.sh
kj4ezj: > the fix suggests its a problem with the existing installed packages.

I thought the point of this test is that this is a clean image with no existing installed formula. What `brew` packages are already installed?

nickjjzhao: Updating my comments: the fix suggests some broken linkage of dependents or outdated dependents after installing, upgrading or reinstalling formulae.

PR #11057, opened 2022-02-02
CI Improvements + HTTP(S) Proxy Support

williamblevins:
## Change Description
This pull request cleans up how docker registries are used and defined throughout the CI code, as well as making the CI code a bit more modular such that the pipeline environment defines the context in which the code runs and where stuff goes instead of having everything hard-coded. We have also added support for HTTP and HTTPS proxies.

## Change Type
**Select *ONE*:**
- [ ] Documentation

- [ ] Stability bug fix

- [X] Other

- [ ] Other - special case

## Testing Changes
**Select *ANY* that apply:**
- [ ] New Tests

- [ ] Existing Tests

- [ ] Test Framework

- [X] CI System

- [ ] Other

## Consensus Changes
- [ ] Consensus Changes

## API Changes
- [ ] API Changes

## Documentation Additions
- [ ] Documentation Additions


.cicd/build.sh
```diff
@@ -58,9 +58,10 @@ else # Linux
     fi
     . "$HELPERS_DIR/file-hash.sh" "$CICD_DIR/platforms/$PLATFORM_TYPE/$IMAGE_TAG.dockerfile"
     COMMANDS="$PRE_COMMANDS && $COMMANDS"
-    DOCKER_RUN="docker run $ARGS $(buildkite-intrinsics) --env CMAKE_EXTRAS='$CMAKE_EXTRAS' '$FULL_TAG' bash -c '$COMMANDS'"
-    echo "$ $DOCKER_RUN"
-    eval $DOCKER_RUN
+    DOCKER_RUN_ARGS="$ARGS $(buildkite-intrinsics) --env CMAKE_EXTRAS='$CMAKE_EXTRAS' '$FULL_TAG' bash -c '$COMMANDS'"
+    echo "$ docker run $DOCKER_RUN_ARGS"
```
linhuang-blockone, line 62: Can this be combined with the next echo to display exactly what "docker run"'s final arguments look like in one log line?

.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
```diff
@@ -9,15 +9,19 @@ RUN yum update -y && \
     libuuid-devel libtasn1-devel expect socat libseccomp-devel && \
     yum clean all && rm -rf /var/cache/yum
 # install erlang and rabbitmq
-RUN curl -s https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh | bash && \
+RUN curl -fsSLO https://packagecloud.io/install/repositories/rabbitmq/erlang/script.rpm.sh && \
```
linhuang-blockone, line 12: Can bases of those URLs be set in variables, not hardcodied?

.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
```diff
@@ -9,15 +9,19 @@ RUN yum update -y && \
...
+    bash script.rpm.sh && \
+    rm script.rpm.sh && \
     yum install -y erlang
-RUN curl -s https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh | bash && \
+RUN curl -fsSLO https://packagecloud.io/install/repositories/rabbitmq/rabbitmq-server/script.rpm.sh && \
```
linhuang-blockone, line 16: You can have a function for  curl and run , and another one for curl and untar to avoid duplicate code.

linhuang-blockone (commented):
The comments are for coding styles. I have successfully run changes in my branch lh-EPE-2041-remove-defer-trans-v1. Approve the PR.

.cicd/build.sh
williamblevins: They could be. Zach was concerned about doing that because he thought that the proxy args would be confusing as they aren't always set or required.

.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
williamblevins: Can you have functions in docker files?

.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
williamblevins: I don't see what value that has as the base URL(s) aren't reused anywhere within the same docker file.

.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
williamblevins: I think this could be done by running a shell script that had the functions. That would require more work here than I think is warranted. The only reason to change the platform files at all was to originally fix an apt problem in one of the ubuntu 18.04 docker files which led me to find some other issues.

.cicd/platforms/pinned/amazon_linux-2-pinned.dockerfile
kj4ezj: I would agree if this were a shell script, but variables in dockerfiles do not persist between layers unless you bake them into the final container. In this case, making these URLs variables would bloat the final product.

Moreover, we wouldn't be de-duplicating any code since these URLs are only used one time.

linhuang-blockone (approved):

